PR #2584, opened 2020-04-03
State tree format conversion with the tree overlay method

gballet:
This EIP describes the tree overlay method to change the state trie format from hexary to binary, as presented in https://ethresear.ch/t/overlay-method-for-hex-bin-tree-conversion/7104


EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+When a miner has calculated Hᵣ², it proceeds to phase 2.
+
+### Phase 2
+
+Phase 2 is the same as phase 1, except for the following:
+
+  * Hᵣ' ≡ Hᵣ²
+
+
+Phase 2 ends when a sufficient number of subsequent blocks have reported the same value for Hᵣ².
```
holiman, line 64: Does this mean that before this happen, `hi2 = "donald_duck"` is a valid block? 
Shouldn't it say, instead

> .. when a sufficient number of subsequent blocks contain binary root hash. 

(assuming that only valid root hashes are 'valid')

Also, does this mean that it's acceptable for a miner to mine on top of a (non-verified) binary-hashed block, but omit the binary-root hash (a phase-one block)?

Could a valid block progression be: `p1, p1, p2, p2, p1, p2, p2,p1,p2` ?

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Phase 3
+
+The following changes occur in phase 3:
+
+  * N accounts are being deleted from the binary overlay trie and inserted into the binary base trie.
```
holiman, line 70: Will `N` be specified later? Or would it happen implicitly via the next bullet-item?

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Simple Summary
+
+This EIP proposes a method to convert the state trie format from hexary to binary: new values are directly stored in a binary trie “laid over” the hexary trie. Meanwhile, the hexary trie is converted to a binary trie in the background. When the process is finished, both layers are merged.
+
+## Abstract
+
+This EIP describes a four phase process to complete the conversion.
+
+  * In the first phase, all new state writes are made to an overlay binary trie, while the hexary trie is being converted to binary. The block format is changed to have two storage roots: the root of the nhexary trie and the root of the binary trie.
```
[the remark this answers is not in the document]
gballet: If the overlay is hexary, then it would have to be converted during the phase 3 merge. Having an overlay binary trie allows the clients to directly switch to binary tries. Then it gives clients time to perform the conversion in the background.

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+
+A conversion process occurs in the background to turn the hexary trie into its binary equivalent. The end goal of this process is the calculation of the _root hash of the converted binary base trie_, denoted Hᵣ². The root of the hexary base trie is hereafter called Hᵣⁱ⁶.
+
+When a miner has calculated Hᵣ², it proceeds to phase 2.
+
+### Phase 2
+
+Phase 2 is the same as phase 1, except for the following:
+
+  * Hᵣ' ≡ Hᵣ²
```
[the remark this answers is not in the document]
gballet: What I mean is that the new Header root is replaced with the binary root. I'm trying to follow the yp notation, in spite of unicode missing some of the symbols. Gotta rethink this then :+1:

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+discussions-to: https://ethresear.ch/t/overlay-method-for-hex-bin-tree-conversion/7104
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-04-03
+---
+
+## Simple Summary
+
+This EIP proposes a method to convert the state trie format from hexary to binary: new values are directly stored in a binary trie “laid over” the hexary trie. Meanwhile, the hexary trie is converted to a binary trie in the background. When the process is finished, both layers are merged.
```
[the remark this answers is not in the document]
gballet: That is voluntarily left vague at the moment, because I don't know at this time if both trees are going to be merged (it has been discussed at the 1x workshop and there's an EIP by Vitalik that I haven't found yet).

I currently assume that it will be both the account and storage tries, but the exact shape of things depend on how and when other teams work this out.

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+  * The series of bytes in I₀ is seen as a series of _bits_ and so ∀i≤256, I₀[i] is the ith bit in key I₀
+  * The first item of an **extension** or a **leaf** is replacing nibbles with bits;
+  * A **branch** is a 2 item structure in which both items correspond to each of the two possible bit values for the keys at this point in their traversal;
+  * c(𝕴,i) ≡ RLP((u(0), u(1)) at a branch, where u(j) = n({I : I ∈ 𝕴 ⋀ I₀[i] = j}, i+1)
+
+### Phase 1
+
+Let _H₁_ be the previously agreed-upon block height at which phase 1 starts. For each block of height H₁ ≤ _h_ < H₂:
+
+  * Block headers contain a block header has a new Hₒ which is the _root of the overlay binary trie_
```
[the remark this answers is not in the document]
gballet: That I should do even more proof-reading :wink:

EIPS/eip-overlay_tree.md
gballet: `N` still has to be specified. The way I see it, there are two approaches, and I'd like to get some feedback on that:

Approach 1: `N` is set to roughly 700. This is because a quick estimate based on my tests say it represents about 1 second. This number would be bigger if we use extensions in binary trees, which I will try when I complete the code that does this.

Approach 2: Dynamically calculate `N` based on the difficulty, to ensure it takes around 1 second.

Any thoughts?

EIPS/eip-overlay_tree.md
gballet: Any value for Hᵣ² is valid. What I'm trying to achieve here with the transition from phase 2 to phase 3 is some kind of vote: miners "vote" the value that they converted, through this Hᵣ² field. If there is an agreement on the same value of more than 51% of the network (the sequence of blocks is a proxy for that) then it is accepted as the "official" value.

If more than 51% of the network is advertising `donald_duck`, then a majority of the network has effectively killed it (let alone committed a copyright infringement, Disney's not kidding with these).

Can the client run by miner build on top of a non-verified Hᵣ²? Yes it can if they trust another party with their conversion. However, they would still need the state as soon as phase 3 starts, because it is needed to progressively merge the overlay back into the base trie. Since in phase 3 both Hᵣ² and Hₒ are part of the consensus, trusting a value without calculating the state is risky.

> Could a valid block progression be: p1, p1, p2, p2, p1, p2, p2,p1,p2 ?

Yes, in this case it means that clients don't agree on their conversion results. It will happen at the beginning when one miner is done converting. The votes should converge over time, unless one actor with most of the hashing power is malicious or made an error.

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
+---
+eip: <to be assigned>
```
axic, line 2:
```suggestion
eip: 2584
```

Also please rename the file to `eip-2584.md`

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+
+Let _H₁_ be the previously agreed-upon block height at which phase 1 starts. For each block of height H₁ ≤ _h_ < H₂:
+
+  * Block headers contain a new Hₒ field, which is the _root of the overlay binary trie_;
+  * Hᵣ ≡ P(H)ᵣ, i.e. the hexary trie root is the same as that of the block parent's. The hexary trie is referred to as the _hexary base trie_.
+
+The following is changed in the execution environment:
+
+  * Upon executing a _state read_, ϒ first searches for the address in the overlay trie. If the key can not be found there, ϒ then searches the hexary base trie as it did at block heights h' < H₁;
+  * Upon executing a _state write_, ϒ will insert or update the value into the overlay tree.
```
axic, line 51: Shouldn't this also clarify that:
1) Before every write a read also happens (needed to determine costs).
2) Upon writing into the overlay, it is deleted from the base trie? 

I'm not sure if 2) is proposed here or whether it is a good idea or not, but clarity would be nice.

EIPS/eip-overlay_tree.md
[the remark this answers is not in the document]
axic: >  I'm trying to follow the yp notation

Ah that clarifies my (not yet asked) question. Can you perhaps make this clear in the beginning of the specification section?

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+  * The third phase begins when a sufficient number of consecutive blocks report the same value for `binary_trie_root`: the overlay tree is progressively merged back into the newly converted binary base trie. A constant number of entries are deleted from the overlay and inserted into the base trie.
+  * The fourth and final phase begins when the overlay trie is empty. The field holding its root is removed from the block header.
+
+## Motivation
+
+There is a long running interest in switching the state trie from a hexary format to a binary format, for reasons pertaining to proof and storage sizes. The conversion process poses a catch-up issue, caused by the sheer size of the full state: it can not be translated in a reasonable time (i.e. on the same order of magnitude as the block time). 
+
+## Specification
+The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).
+
```
axic, line 31:
```suggestion
This specification follows the notation introduced by the [Yellow Paper](https://ethereum.github.io/yellowpaper). Prior to reading it is advisable to be familiar with the Yellow Paper.
```

EIPS/eip-overlay_tree.md
gballet: During phase 1, values are not deleted from the base trie, as the base trie is read-only. I'll make it clear :+1:

EIPS/eip-overlay_tree.md
```diff
@@ -0,0 +1,108 @@
...
+  * Hᵣ ≡ P(H)ᵣ, i.e. the hexary trie root is the same as that of the block parent's. The hexary trie is referred to as the _hexary base trie_.
+
+The following is changed in the execution environment:
+
+  * Upon executing a _state read_, ϒ first searches for the address in the overlay trie. If the key can not be found there, ϒ then searches the hexary base trie as it did at block heights h' < H₁;
+  * Upon executing a _state write_, ϒ will insert or update the value into the overlay tree.
+
+A conversion process occurs in the background to turn the hexary trie into its binary equivalent. The end goal of this process is the calculation of the _root hash of the converted binary base trie_, denoted Hᵣ². The root of the hexary base trie is hereafter called Hᵣⁱ⁶.
+
+When a miner has calculated Hᵣ², it proceeds to phase 2.
```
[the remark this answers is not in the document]
gballet: Done.

EIPS/eip-overlay_tree.md
gballet: So after discussing this on other media, there doesn't seem to be any advantages to the complexity of this approach. I have updated this document to present a simpler transition: it will unconditionally happen at block height h2, which is set far apart in the future to make sure that honest nodes are done with the conversion. Measurements indicate that this can be achieved in less than a day.

EIPS/eip-2584.md
```diff
@@ -0,0 +1,108 @@
...
+Let _h₁_ be the previously agreed-upon block height at which phase 1 starts, and _h₂_ the block at which phase 2 starts. For each block of height h₁ ≤ _h_ < h₂:
+
+  0. A conversion process is started in the background, to turn the hexary trie into its binary equivalent. The end goal of this process is the calculation of the _root hash of the converted binary trie_, denoted Hᵣ². The root of the hexary trie is hereafter called Hᵣ¹⁶. Formally, this process is written as Hᵣ² ≡ ß(Hᵣ¹⁶).
+  1. Block headers contain a new Hₒ field, which is the _root of the overlay binary trie_;
+  2. Hᵣ ≡ P(H)ᵣ¹⁶, i.e. as long as the conversion from hexary to binary is not complete, the hexary trie root is the same as that of its parent block.
+
+The following is changed in the execution environment:
+
+  * Upon executing a _state read_, ϒ first searches for the address in the overlay trie. If the key can not be found there, ϒ then searches the base trie as it did at block heights h' < h₁;
+  * Upon executing a _state write_, ϒ will insert or update the value into the overlay tree. The base tree is left untouched.
```
holiman, line 54: what happens on state deletions, when an account is deleted from the state trie. Do we insert some "temporary" null-value into the trie or do we need to maintain an explicit "deletion" thing?

holiman:
Alternative variant: 
During phase 1 , we don't stop the world, but we stop transactions. The side-effect would be, that the only things that goes into the overlay trie is the changes to the miner coinbases. Thus, the overlay trie will be very small (realistically a handful, theoretically the same as the number of blocks in phase 1). 
The result will be that the merging of overlay trie into base trie can happen in one go, and we can skip the move-700-accounts phase altogether. 

I think phase 1 is complex, as we need to check conversion-pointer, read from overlay trie, and _then_ read from main trie if the account was deleted from the bintrie due to an earlier write. 

This is not 'stop the world', just means that we won't process transactions for a couple of days, but we still pay miners for work performed.

holiman:
One more upside with no-tx-in-phase-1 is that we won't have to worry about deletions. 
So here's the problem with deletions: 
- The hexary trie is 'frozen', so if an account is deleted, we somehow have to place a deleted-marker into the "ephemeral" overlay binary trie. 
- However, if an account is created during this phase, and _then_ deleted, it must not get that same deleted-marker. 
- There's no precedent about a deleted-marker in tries, and there's no definitions on how to hash it. 

So you might run into the solution that we'll have to maintain an additional "deletion-trie", or "deletion-list", which _also_ must be in the blockhash. IT becomes a pretty hairy problem to solve. 

So if the only effect after a block is the addition of mining rewards to the miner and any ommer-miners, then that problem goes away. 

TLDR; - there are _two_ reasons for not allowing txs during phase 1: 
- Phase 2 can be done instantly, in one block
- No deletions need to be managed,

holiman:
> Could you elaborate on why a different marker is necessary?

No, actually, you're probably right, those could be treated the same way. But what we wind up with is a third trie variant, which functions slightly differently than the other tries.

EIPS/eip-2584.md
```diff
@@ -0,0 +1,116 @@
...
+There are three attack vectors that I can foresee:
+
+  * A targeted attack that would cause the overlay trie to be unreasonably large. Since gas costs will likely increase during the transition process, lengthening phase 2 will make Ethereum more expensive during an extended period of time. This could be solved by increasing the cost of `SSTORE` during phase 1.
+  * Conversely, if h₂ comes too soon, a majority of miners might not be able to produce the correct value for Hᵣ² in time.
+  * If a group of miners representing more than 51% of the network are reporting an invalid value, they could be stealing funds without anyone having a say.
+
+## Community feedback
+
+  * Preliminary tests indicate that a fast machine can perform the conversion in roughly 30 minutes.
+  * The initial version of this EIP expected miners to vote on the value of the binary base root. This has been removed because of the complexity of this process, and because this functionality is already guaranteed by the "longuest chain" rule.
```
axic, line 113:
```suggestion
  * The initial version of this EIP expected miners to vote on the value of the binary base root. This has been removed because of the complexity of this process, and because this functionality is already guaranteed by the "longest chain" rule.
```

axic:
@gballet when should this be merged as a draft?

gballet:
@axic it could be merged now.

axic (approved):

PR #2586, opened 2020-04-04
1193: `disconnect` event, connectivity errors

rekmarks:
https://github.com/rekmarks/EIPs/blob/1193-open/EIPS/eip-1193.md

- Adds the `disconnect` event, intended to:
  - supersede the `close` event
  - make it clear that it's the counterpart of the `connect` event
- Deprecates the `close` event

The above follows discussions with @ryanio 

In addition:

- Adds the following Provider errors:
  - `4900: Not Connected`
  - `4901: Chain Not Connected`
- Adds the section `Connectivity` for defining "connected" and "disconnected"

This allows consumers to tell whether their requests are failing because the Provider is disconnected.

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -32,7 +32,7 @@ type RequestParams = Array<any> | { [key: string]: any };
 ethereum.request(method: string, params?: RequestParams): Promise<unknown>;
 ```
 
-The Promise resolves with the method's result or rejects with an `Error`. For example:
+The Promise resolves with the method's result or rejects with a [`ProviderRpcError`](#errors). For example:
 
 ```javascript
 ethereum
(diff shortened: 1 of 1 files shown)


adrianmcli:
As much as I like the fact that `open` and `close` are symmetric. I think `connect` and `disconnect` makes a lot more intuitive sense.

rekmarks:
The only reason I picked `open` was because the `close` event uses the Web API `CloseEvent` status codes, but that's not really striking me as that relevant, on consideration. I'll change it.

eip-automerger:

rekmarks:
@ryanio, note that I added the `isConnected` method; please see description for justification.

Happy to roll it back if undesired, but, I obviously think it's justified 😛

ryanio:
I’d prefer not to introduce another method. Wouldn’t you be able to just send a json rpc request like eth_chainId to verify connection?

rekmarks:
I’m very sympathetic to that goal. `eth_chainId` is unfortunately of no help, because EIP 695 has nothing to say about the case where no chain ID is available.

On reflection, even though a method enabling consumers to tell if “the wire is hot” would be nice, we could just standardize this by means of an RPC error.

Since this will be specific to the Provider, that error should IMO be introduced here. What do you think?

rekmarks:
e0e52a3 Removed `isConnected`, added connectivity errors.

PR #2587, opened 2020-04-04
1193: add global initialization event

rekmarks:
_This PR is strictly additive; it does not break or modify any existing features of 1193._

_This PR is in draft awaiting #2586._

https://github.com/rekmarks/EIPs/blob/1193-global-event/EIPS/eip-1193.md

- Add a "Global Initialization Event": `ethereum#initialized`
  - This is intended for environments where the Provider is asynchronously injected
  - Today, in such environments, consumers have no recourse but to poll for `window.ethereum` via `setInterval` if the Provider is not synchronously available
  - This event is emitted using the event API of `globalThis`, if available
    - In browsers, this is the [`EventTarget` Web API](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget), which is supported by `window` (the browser `globalThis`)
- Various minor fixes (typos etc.)

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -20,6 +20,8 @@ The Provider's interface is designed to be minimal, preferring that features are
 
 The events `connect`, `disconnect`, `chainChanged`, and `accountsChanged` are provided as a convenience to help enable reactive dapp UIs.
 
+For usage, see [Examples](#examples).
+
 ## API
 
 ### request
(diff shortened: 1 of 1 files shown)


axic:
I am a bit of outsider here, but I wonder what is the point where it makes sense to set the "core functionality" of EIP-1193 and introduce extension ERCs for it?

Just as a bystander the amount of changes going on under 1193 seem to be confused given it is something already used in production?

rekmarks:
@axic 

>I am a bit of outsider here, but I wonder what is the point where it makes sense to set the "core functionality" of EIP-1193 and introduce extension ERCs for it?

Ideally this week, IMO. The only reason I'm suggesting this particular feature, is that it is my (and others at MetaMask) firm belief that 1193 would be underspecified without it.

>Just as a bystander the amount of changes going on under 1193 seem to be confused given it is something already used in production?

Anecdotally, by discussions with `web3` and `ethers` developers, and by the results of a recent survey conducted by MetaMask (results TBD), the current degree of standardization between providers is very poor. The changes recently introduced to 1193 were designed to be implementable without forcing wallets to break their existing APIs, hopefully encouraging its adoption.

rekmarks:
Superseded by #2588

PR #2577, opened 2020-03-30
Update EIP 1193

rekmarks:
Link to file: https://github.com/rekmarks/EIPs/blob/rekmarks-1193-update/EIPS/eip-1193.md

Makes the following updates to EIP 1193, in light of recent discussions in #2319:
- Deprecates `send` and `sendAsync`
  - Providers may continue to support these methods
- Specifies a new method, `request`, with the same Promise signature as the `send` currently at `ethereum/EIPs#master`
  - Per @MicahZoltu's [suggestion](https://github.com/ethereum/EIPs/issues/2319#issuecomment-591194700)
- Changes the interface of the `notification` event, such that a generalized object is emitted
  - This was done to support other notifications than `eth_subscription`
- Makes the spec transport- and RPC protocol-agnostic, per the contributions of @ryanio
- Adds the `message` event for generalized messaging, and deprecates the `notification` event
  - This is so that existing uses of the `notification` event do not break
  - All functionality previously handled by the `notification` event is now handled by the `message` event
- Various edits for clarity and stylistic consistency

PR #2588, opened 2020-04-07
1193: Make spec availability-agnostic

rekmarks:
https://github.com/rekmarks/EIPs/blob/1193-availability/EIPS/eip-1193.md

For the purpose of this PR, let's define the problem of how to expose the Provider API as "availability" or "the availability problem."

Per [this comment](https://github.com/ethereum/EIPs/issues/2319#issuecomment-544795458) by @alcuadrado, one of the concerns from the 1193 meeting at Devcon 5 was:

>1. EIP-1193 began as a spec on how to expose a provider in a dapp browser, but mutated into specifying its interface. It still depends on EIPs related to browser-specific stuff.
>2. The way a provider is exposed (i.e. `window.ethereum`) only supports a single provider. This is in conflict with users installing multiple wallet extensions, and may be problematic with a multi-chain future.

Meanwhile, the proposed solution was:

>1. EIP-1193 will only focus on the provider's interface. How to expose in browser environments will be part of another document, possibly an RFC. Any dependency on browser-specific EIPs will be removed.
>2. As this EIP won't dictate how to expose the provider, this item is outside of its scope.

The goal of stripping all (normative) availability-related language was never fully realized, and recently merged PRs (#2577, #2586) in fact introduced more such language.

This PR makes the spec completely availability-agnostic, while including a non-normative section noting that `window.ethereum` is, by convention, the de facto standard for exposing the Provider API. Said section notes some limitations of this convention, for the benefit of Provider implementers, dapp developers, and future authors of future EIPs.

Code examples remain unchanged, and use the `window.ethereum` convention (although note that it is merely a convention).

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -284,9 +284,9 @@ _This section is non-normative._
 
 _This section is non-normative._
 
-How the Provider is made available is beyond the scope of this specification.
-At the time of writing, there exists no specification for Provider availability.
-There is a convention, however, which is described here for the benefit of dapp developers and Provider implementers.
+How the Provider is made available to consumers is beyond the scope of this specification.
+At the time of writing, there exists no specification for Provider availability, merely a convention.
+This convention is described here for the benefit of dapp developers and Provider implementers.
 
 Historically, Providers have been injected into web pages as `window.ethereum` (more generally, `globalThis.ethereum`), such that they are available on page load.
 
@@ -295,7 +295,7 @@ In practice, this convention does not handle some situations, including:
 - Multiple Providers being injected into the same page, e.g. when the user has multiple wallets installed
 - Asynchronously injected Providers, whether by choice or due to platform limitations
 
-Provider implementers are encouraged to work with each other and with dapp developers to solve these problems until new standards emerge.
+Provider implementers are encouraged to work with each other and with dapp developers to solve these problems until standards emerge.
 
 ### Connectivity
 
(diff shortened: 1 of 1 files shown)


rekmarks:
cc: @adrianmcli @ryanio

eip-automerger:

EIPS/eip-1193.md
```diff
@@ -29,7 +29,7 @@ Makes an Ethereum RPC method call.
 ```typescript
 type RequestParams = Array<any> | { [key: string]: any };
 
-ethereum.request(method: string, params?: RequestParams): Promise<unknown>;
+Provider.request(method: string, params?: RequestParams): Promise<unknown>;
```
Gudahtt, line 32: There's another use of `ethereum` on line 38, just below this one

rekmarks: cab3d7a

adrianmcli:
The diff looks good to me! Nice work @rekmarks

frozeman:
I like it to be generic. Though we should talk about how a multi EVM chain future can be possible with this provider. E.g. `ethereum.chainX.request()`, or `ethereum.request('chainIDX', 'method', [params])`

rekmarks:
@frozeman I was thinking about how to change the `request` method signature to enable the targeting of a specific chain, or even another protocol entirely.

That'd be a different PR, of course, but the two things that come to mind are:

1. Just add a third `config` param, like so:
```typescript
request(method: string, params: Object | Array, config: Object): Promise<unknown>`
```

2. Change/overload the signature such that it can be passed a single object param:
```typescript
interface RequestArgs {
  method: string;
  params: Array | Object;
  chainId?: string;
  ...otherStuff: Array<any>;
}

request(args: RequestArgs): Promise<unknown>
```

PR #2590, opened 2020-04-09
1193: single object argument for `request` method

rekmarks:
https://github.com/rekmarks/EIPs/blob/1193-request-arg/EIPS/eip-1193.md

New signature for the `request` method:
```typescript
interface RequestArguments {
  method: string;
  params?: Array<any> | { [key: string]: any };
  [key: string]: any;
}

Provider.request(args: RequestArguments): Promise<unknown>;
```

Example call:
```javascript
Provider.request({ method: 'eth_accounts' })
  .then((accounts) => console.log(accounts))
  .catch((error) => console.error(error));
```

See document for further details.

This is motivated by discussions with @frozeman and others, and enables a future for the Provider interface generally and the `request` method specifically where any of the following holds:

- Multiple chains are served by a single Provider at once
- Multiple protocols are served by a single Provider at once
- The Provider needs some kind of metadata submitted along with `request` calls, for reasons we are unable to anticipate today

If we want EIP 1193 to function "in perpetuity", why not make the `request` method maximally flexible?

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -29,8 +29,8 @@ Makes an Ethereum RPC method call.
 ```typescript
 interface RequestArguments {
   method: string;
-  params?: any;
-  [key: string]: any;
+  params?: unknown;
+  [key: string]: unknown;
 }
 
 Provider.request(args: RequestArguments): Promise<unknown>;
@@ -73,7 +73,7 @@ Historically, they have followed the [Ethereum JSON-RPC specification](https://g
 This method is deprecated in favor of [`request`](#request).
 
 ```typescript
-Provider.send(...args: Array<any>): unknown;
+Provider.send(...args: Array<unknown>): unknown;
 ```
 
 ### Events
@@ -171,7 +171,7 @@ Historically, this event has returned e.g. `eth_subscribe` subscription updates 
 interface ProviderRpcError extends Error {
   message: string;
   code: number;
-  data?: any;
+  data?: unknown;
 }
 ```
 
@@ -328,8 +328,8 @@ The Provider **MAY** expose methods and properties not specified in this documen
 ```typescript
 interface RequestArguments {
   method: string;
-  params?: any;
-  [key: string]: any;
+  params?: unknown;
+  [key: string]: unknown;
 }
 
 Provider.request(args: RequestArguments): Promise<unknown>;
@@ -381,7 +381,7 @@ If an RPC method defined in a finalized EIP is not supported, it **SHOULD** be r
 interface ProviderRpcError extends Error {
   message: string;
   code: number;
-  data?: any;
+  data?: unknown;
 }
 ```
 
@@ -442,7 +442,7 @@ interface EthSubscription extends ProviderMessage {
   type: 'eth_subscription';
   data: {
     subscription: string;
-    result: any;
+    result: unknown;
   };
 }
 ```
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -27,16 +27,19 @@ The events `connect`, `disconnect`, `chainChanged`, and `accountsChanged` are pr
 Makes an Ethereum RPC method call.
 
 ```typescript
-type RequestParams = Array<any> | { [key: string]: any };
+interface RequestArguments {
+  method: string;
+  params?: Array<any> | { [key: string]: any };
+  [key: string]: any;
+}
 
-Provider.request(method: string, params?: RequestParams): Promise<unknown>;
+Provider.request(args: RequestArguments): Promise<unknown>;
 ```
 
 The Promise resolves with the method's result or rejects with a [`ProviderRpcError`](#errors). For example:
 
 ```javascript
-Provider
-  .request('eth_accounts')
+Provider.request({ method: 'eth_accounts' })
   .then((accounts) => console.log(accounts))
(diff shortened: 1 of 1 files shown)


eip-automerger:

EIPS/eip-1193.md
```diff
@@ -200,7 +203,10 @@ ethereum
 
 // Example 2: Log last block
 ethereum
-  .request('eth_getBlockByNumber', ['latest', 'true'])
+  .request({
+    method: 'eth_getBlockByNumber',
+    params: ['latest', 'true'],
```
adrianmcli, line 208: This line can also be written like this, right?

```js
params: {
  latest: true
}
```

adrianmcli: This example seems a little more straight-forward to understand when written as an object instead of an array.

rekmarks: Hm, I don't know if the params can be given in object form. I imagine it's just following: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbynumber

rekmarks: By the method's interface the params can certainly be given as an object, but whatever the specific RPC method's definition says, goes.

adrianmcli: Ah nevermind, I confused myself. However the type annotation (on line 32) here makes it seems like params can either be an array of anything or an object?

```ts
params?: Array<any> | { [key: string]: any };
```

adrianmcli: But in the official JSON-RPC, it seems like it is always an array.

EIPS/eip-1193.md
```diff
@@ -27,16 +27,19 @@ The events `connect`, `disconnect`, `chainChanged`, and `accountsChanged` are pr
 Makes an Ethereum RPC method call.
 
 ```typescript
-type RequestParams = Array<any> | { [key: string]: any };
+interface RequestArguments {
+  method: string;
+  params?: Array<any> | { [key: string]: any };
```
adrianmcli, line 32: This part confuses me because it seems to indicate that the `params` property can either be an array of stuff or an object. But the JSON-RPC only ever lists them out as arrays.

ryanio:
@adrianmcli the ethereum json rpc api has traditionally accepted an array for params, however the official json-rpc spec does specify that params can also be passed as a by-name object. The real underlying reason why I recently made it optional was this realization but also to support graphql queries in the future `method` can == `query` and `params` can == `variables`.

Hope that adds some clarity :)

rekmarks:
@adrianmcli @ryanio, I'm thinking we should just change `params` to `any` and refer to the RPC method's own documentation, adding that for Ethereum JSON RPC (by far of course the most important one right now), always expects an Array. Since we want the spec to be RPC protocol-agnostic, there's no real reason to follow its specification for `params`.

Any thoughts on that? I might just push a commit so we can get a feel for it.

EIPS/eip-1193.md
MicahZoltu, line 32: `any` should be `unknown`.  Never use `any` in TypeScript, it basically is a shorthand for "turn type checking off entirely for this variable".

EIPS/eip-1193.md
```diff
@@ -27,16 +27,19 @@ The events `connect`, `disconnect`, `chainChanged`, and `accountsChanged` are pr
...
+  [key: string]: any;
```
MicahZoltu, line 33: `unknown` instead of `any` here as well.

EIPS/eip-1193.md
rekmarks: Done in 7753ce8

EIPS/eip-1193.md
rekmarks: Done in 7753ce8

EIPS/eip-1193.md
```diff
@@ -314,18 +323,24 @@ The Provider **MAY** expose methods and properties not specified in this documen
 
 #### request
 
+> The `request` method is intended as a transport- and protocol-agnostic wrapper function for Remote Procedure Calls (RPCs).
+
 ```typescript
-type RequestParams = Array<any> | { [key: string]: any };
+interface RequestArguments {
+  method: string;
+  params?: Array<any> | { [key: string]: any };
```
MicahZoltu: Same comments as above about `any`.  Always prefer `unknown`.

rekmarks: My bad for not clarifying: all references to `any` were changed to `unknown` in 7753ce8

MicahZoltu:
I'm not personally a fan of this change as it hurts the ergonomics of using these functions and doesn't actually solve the versioning problem (which I do agree *is* a problem that should be solved).  While options objects give a bit more flexibility in terms of versioning, you *still* need some larger versioning mechanism in place to make this *useful* and if you have a versioning system you can just have versioned function signatures.

MicahZoltu:
That being said, ergonomics are not a blocker for me on 1193, so if that is what everyone else wants I'll capitulate.

rekmarks:
@MicahZoltu, I also find ergonomics important, although I personally find the difference negligible in this case, especially in light of the gains in extensibility.

On that note, this change is only meant to address extensibility, not versioning. I still firmly support the creation of that EIP, and may put down some time on that, too.

Thanks for the review; your input is always appreciated!

PR #2591, opened 2020-04-10
EIP-2315: fix opcodes in examples

holiman:
Ref: 
https://github.com/openethereum/openethereum/pull/11612#discussion_r404788297

Geth had the wrong opcodes defined, so the provided example bytecode/traces were wrong. Fixed with this PR


holiman:
Ah, the automerger merged too soon :)

axic:
Next time create it as draft (there is an option when creating the PR) if you want to avoid it

PR #2593, opened 2020-04-11
Gas price market change: Escalator Algorithm (alternative to 1559)

danfinlay:
Submitting a draft of a version of the escalator algorithm applied to the Ethereum network, which I suspect has superior qualities to EIP-1559 in several regards.

Opening this PR so that the algorithms can more easily be compared side by side.

Full text available for easy reading here:
https://github.com/danfinlay/EIPs/blob/Escalator/EIPS/eip-x.md

You can read some previous discussion on it here:
https://ethresear.ch/t/another-simple-gas-fee-model-the-escalator-algorithm-from-the-agoric-papers/6399


EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+**Proposal**
+
+For all blocks where `block.number >= INITIAL_FORK_BLKNUM`:
+
+When processing a transaction with the new pricing parameters, miners now receive a fee based off of the following linear function, where `BLOCK` is the current block number:
+
+* IF `BLOCK > MAX_BLOCK` then `TX_FEE = MAX_PRICE`.
+* `TX_FEE = START_PRICE + ((MAX_PRICE - START_PRICE) / (MAX_BLOCK - START_BLOCK) * (BLOCK - START_BLOCK))`
+
+As a JavaScript function:
```
moodysalem, line 106: Suggest this could be a smart contract that implements this function in the future, but by default linear?

danfinlay: That could be, I looked at some other EIPs and it seemed like pseudocode is fairly common for layer-1 proposals, but I'd gladly add specificity anywhere it's lacking.

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+* `MAX_PRICE`: The maximum price the sender would be willing to pay to have this transaction processed.
+* `MAX_BLOCK`: The last block that the user is willing to wait for the transaction to be processed in.
+
+**Proposal**
+
+For all blocks where `block.number >= INITIAL_FORK_BLKNUM`:
+
+When processing a transaction with the new pricing parameters, miners now receive a fee based off of the following linear function, where `BLOCK` is the current block number:
+
+* IF `BLOCK > MAX_BLOCK` then `TX_FEE = MAX_PRICE`.
```
moodysalem, line 103: The ideal behavior imo is that the transaction cannot be included after the `MAX_BLOCK`

EIPS/eip-x.md
moodysalem: Just to clarify, I mean that the EIP could specify that a linear 'escalator' algorithm is just one of a number of gas price functions, which could be implemented as smart contracts and then specified by the user in the transaction as an address, `gasPriceContractAddress`. E.g. the specified contract could produce gas prices as a step function, or decreasing gas prices to encourage immediate inclusion. This may be way out of scope for this EIP though, but just worth mentioning as a hint towards the extensibility of this feature

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+
+**Client-Wide Parameters**
+* `INITIAL_FORK_BLKNUM`: TBD
+
+**Transaction Parameters**
+The transaction `gasPrice` parameter is now optional, and if excluded can be replaced by these parameters instead:
+
+* `START_PRICE`: The lowest price that the user would like to pay for the transaction.
+* `START_BLOCK`: The first block that this transaction is valid at.
```
AlexeyAkhunov, line 93: How is the `START_BLOCK` actually defined? If a transaction could have been valid at block 1000, but it has been created at block 2000, what is the `START_BLOCK` in such case?

MicahZoltu: Similarly, what happens if `START_BLOCK` is in the future?  Can this EIP *also* solve the problem of delayed transactions?

AlexeyAkhunov: My question was basically a hint that `START_BLOCK` probably cannot be suitably defined (so that it can be efficiently calculated from consensus values, and not just from subjective tx pool timestamps), and this could be the main thing that makes this EIP hard or impossible to implement

moodysalem: What do you mean suitably defined? The user or user's wallet specifies the `START_BLOCK` and `MAX_BLOCK` in the transaction, just like with gas price and gas limit. If the miner is working on a block with number `BLOCK_NUMBER`, the block is valid iff _all_ transactions within the block have `START_BLOCK` <= `BLOCK_NUMBER` and `MAX_BLOCK` >= `BLOCK_NUMBER`

It's very easy for a wallet to set this parameter. Just choose the next block for `START_BLOCK` and the `MAX_BLOCK` to whatever delay the user is willing to tolerate divided by the 15 seconds average block times plus the start block.

This constraint can be cheaply validated by any node, and adds some compute time to ordering transactions into a block (must determine the gas price for each transaction for the given block, and whether it can be included) but also has the benefit that transactions can be dropped from the mempool as soon as the block number exceeds `MAX_BLOCK` by some threshold, and potentially if they are sent with a `START_BLOCK` too far in the future

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+* `MAX_PRICE`: The maximum price the sender would be willing to pay to have this transaction processed.
+* `MAX_BLOCK`: The last block that the user is willing to wait for the transaction to be processed in.
```
MicahZoltu, line 95: I'm not a fan of the transaction getting dropped when max-block is reached.  In most cases I would want to escalate up to target price and then sit at that target price until mined or cancelled.  I can *also* imagine situations where I want to "mine by this block or give up", so I can appreciate the desire but I don't think it is the common case.

EIPS/eip-x.md
MicahZoltu: Hmm, this wording conflicts with the wording above on `MAX_BLOCK`.  I disagree with @moodysalem here and think that the ideal behavior (if we have to choose between two options) is to have it sit at `MAX_PRICE` forever once `MAX_BLOCK` is reached.

EIPS/eip-x.md
MicahZoltu: In order to prevent DoS attacks against the network, nodes need to be able to *very quickly* check to see if the caller can afford to submit the transaction, before executing arbitrary user-provided code.  This ensures that someone can't waste node resources by submitting transactions to the network that are invalid.

EIPS/eip-x.md
moodysalem: I see people complain about transactions pending for days, blocking new things they want to do, in basically every dapp's support chat. Cancelling transactions via resubmitting with different gas price is a horrible experience, and has non-deterministic results. If you really want to wait up to 3 days for a transaction, set the MAX_BLOCK to last 3 days.

EIPS/eip-x.md
MicahZoltu, line 95: Consider changing this to `START_TIME` and `MAX_TIME` instead.  Miners can only fudge the block timestamp a little, and it is very costly for them to fudge the timestamp into the future.  Timestamps are much easier for humans, and also much more reliable (block times range from 1 second to 100 seconds) so using block numbers can result in very unexpected (for a human) wait times.

Also, by using timestamps here we will incentivize miners to more accurately set block timestamps than they do now.  Right now, many miners set timestamps far (30+ seconds) in the past because there is a *very marginal* advantage for them to have their clocks be off by a bit.  By having this be timestamp based, miners are strongly encouraged to set the timestamp as far forward as they can without risking significant profits.  It turns out, that is just about `now` because they start taking on huge risk by setting `block.timestamp` to the future.  This means that we can kill two birds with one stone with this PR by having gas price (or gas premium) be a function of time instead of block number.

EIPS/eip-x.md
moodysalem: I think the DoS consideration is fair, but there is an obvious solution: limit how much gas the computation requires and make it pure. You still have to look up the code, but that can be easily memoized.

EIPS/eip-x.md
MicahZoltu: The problem is that the fee algorithm is a function of `MAX_BLOCK, START_BLOCK`, so if you set `MAX_BLOCK` to 3 days in the future, your fee increase rate is *really slow*.  I would not be opposed to adding yet-another-variable for `TERMINATION_BLOCK`, so you can escalate up to a target fee over say 5 minutes, but then sit in the pending queue for up to 24 hours at that fee.

In other words, I agree that there is a problem with never-expiring transactions.  I just disagree that `MAX_BLOCK` in this specification is the right place to fix that.  In fact, one could create a totally separate EIP that just adds in a `NOT_AFTER` field to transactions.

EIPS/eip-x.md
moodysalem, line 95: `START_BLOCK`/`MAX_BLOCK` is inconsistent, prefer `MIN_BLOCK`/`MAX_BLOCK` or `START_BLOCK`/`END_BLOCK`

EIPS/eip-x.md
AlexeyAkhunov: Ok, I misunderstood, now I see that `START_BLOCK` is suggested to be a new transaction parameter. Thank you

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
+---
+eip: tbd
```
MicahZoltu, line 2:
```suggestion
eip: 2593
```
Also make sure to change the name of this file to `eip-2593.md`

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+author: Dan Finlay <dan@danfinlay.com>
+discussions-to: https://ethresear.ch/t/another-simple-gas-fee-model-the-escalator-algorithm-from-the-agoric-papers/6399
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-03-13
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
```
MicahZoltu, line 13:
```suggestion
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
```
MicahZoltu, line 15:
```suggestion
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+Based on [The Agoric Papers](https://agoric.com/papers/incentive-engineering-for-computational-resource-management/full-text/).
+
+Each transaction would have the option of providing parameters that specify an "escalating" bid, creating a time-based auction for validators to include that transaction.
+
+This creates highly efficient price discovery, where the price will always immediately fall to the highest bid price, which is not necessarily that user's highest price they would pay.
+
+![escalator algorithm price chart](https://ethresear.ch/uploads/default/original/2X/0/042795efa4c2680d644bc66386cd2984a70293f8.gif)
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
```
MicahZoltu, line 29:
```suggestion
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+
+EIP-1559 also performs well under normal conditions, but under conditions where blocks are regularly full, the price discovery mechanism breaks, and miners will fall back to the `TIP` parameter to choose the transactions to include, meaning that under network congestion, EIP-1559 forces users to _either_ choose efficient prices or certainty of next-block inclusion.
+
+EIP-1559 also has all the re-submission issues of the current model in situations where a user would like to pay under the current market rate, but has certain time constraints limiting their patience. The Escalator algorithm is the only strategy listed here that allows users to discover the lowest possible price given the network conditions and their time constraints.
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+
```
MicahZoltu, line 85:
```suggestion
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+  return startPrice + (slope * (blockNumber - startBlock))
+}
+```
+
+## Backwards Compatibility
+
+Since a current `gasPrice` transaction is effectively a flat-escalated transaction bid, it is entirely compatible with this model, and so there is no concrete requirement to deprecate current transaction processing logic, allowing cold wallets and hardware wallets to continue working for the forseeable future.
+
+## Test Cases
+<!--Test cases for an implementation are mandatory for EIPs that are affecting consensus changes. Other EIPs can choose to include links to test cases if applicable.-->
```
MicahZoltu, line 125:
```suggestion
TBD
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Implementation
+<!--The implementations must be completed before any EIP is given status "Final", but it need not be completed before the EIP is accepted. While there is merit to the approach of reaching consensus on the specification and rationale before writing code, the principle of "rough consensus and running code" is still useful when it comes to resolving many discussions of API details.-->
```
MicahZoltu, line 128:
```suggestion
TBD
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Security Considerations
+<!--All EIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. EIP submissions missing the "Security Considerations" section will be rejected. An EIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
```
MicahZoltu, line 131:
```suggestion
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+The security considerations for this EIP are:
+- None currently known.
+
+## Resources
+* [Original Magicians thread](https://ethresear.ch/t/another-simple-gas-fee-model-the-escalator-algorithm-from-the-agoric-papers/6399)
+
+## Copyright
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
```
MicahZoltu, line 140:
```suggestion
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,140 @@
...
+## Security Considerations
+<!--All EIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. EIP submissions missing the "Security Considerations" section will be rejected. An EIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
+The security considerations for this EIP are:
+- None currently known.
+
+## Resources
+* [Original Magicians thread](https://ethresear.ch/t/another-simple-gas-fee-model-the-escalator-algorithm-from-the-agoric-papers/6399)
+
```
MicahZoltu, line 137:
```suggestion
```

This isn't a standard EIP section, and while non-standard sections aren't prohibited they are discouraged.  This is already linked in the header, so it is redundant to link it here anyway.

MicahZoltu (changes requested):
I wasn't an editor back when this was proposed and so I never left actual editor feedback to get it merged as a draft.

This needs an EIP number to merge, and we have a new policy about leaving the comments in (it is to remove them) so I added suggestions for that as well.  Otherwise, I think we can merge this as draft.

EIPS/eip-x.md
danfinlay: Thanks for highlighting this distinction. I think I agree with you as the default case, especially since it resembles current behavior more, but I think there is a use case (for a future EIP, perhaps a new tx type) to allow TXs that become invalid over time or after some conditions. Making this change.

danfinlay: This is an interesting point, and I have no objection, so I've made this change. I hope this doesn't greatly complicate any analyses that are currently in progress.

EIPS/eip-2593.md
```diff
@@ -0,0 +1,131 @@
...
+  const blockRange = maxTime - startTime
+  const slope = priceRange / blockRange
+
+  return startPrice + (slope * (currentTime - startTime))
+}
+```
+
+## Backwards Compatibility
+
+Since a current `gasPrice` transaction is effectively a flat-escalated transaction bid, it is entirely compatible with this model, and so there is no concrete requirement to deprecate current transaction processing logic, allowing cold wallets and hardware wallets to continue working for the forseeable future.
```
axic, line 118: To make the spellcheck happy (and CI to pass):
```suggestion
Since a current `gasPrice` transaction is effectively a flat-escalated transaction bid, it is entirely compatible with this model, and so there is no concrete requirement to deprecate current transaction processing logic, allowing cold wallets and hardware wallets to continue working for the foreseeable future.
```

MicahZoltu (approved):
Good to merge as draft.  I'll commit the spelling fix just to speed the process along, if the authors think this is incorrect please open a PR to change it back so we can discuss (we will need to add a custom rule to the bot in such a case).

PR #2599, opened 2020-04-13
EIP-2315: updates

holiman:
This update to 2315 contains 

- One more example testcase which shows the behaviour when the `RETURNSUB` destination is off the end of the code. 
- A clarification about the implementation of the returnstack.


matkt:
LGTM. This clarification about the implementation of the returnstack and this new example seems interesting to add to the description of the PIE.

PR #2609, opened 2020-04-22
Reserve chainId param for request method arguments

rekmarks:
https://github.com/rekmarks/EIPs/blob/1193-request-chainId/EIPS/eip-1193.md

Per @frozeman's suggestion, reserve the `chainId` key on the `RequestArguments` interface for values per the `eth_chainId` Ethereum RPC method.

Intentionally does not specify what to do with this property, and does not include it in the API portion of the documentation, as its already implicitly allowed, and specifying what it does is beyond the scope of this EIP.

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -329,6 +329,7 @@ The Provider **MAY** expose methods and properties not specified in this documen
 interface RequestArguments {
   method: string;
   params?: unknown;
+  chainId?: string;
   [key: string]: unknown;
 }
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

EIPS/eip-1193.md
```diff
@@ -338,6 +339,12 @@ Provider.request(args: RequestArguments): Promise<unknown>;
 The `request` method **MUST** send a properly formatted request to the Provider's Ethereum client.
 Requests **MUST** be handled such that, for a given set of arguments, the returned Promise either resolves with a value per the RPC method's specification, or rejects with an error.
 
+If the `args` object includes the `chainId` property, its value **SHOULD** be the integer ID of a chain as a hexadecimal string, per the [`eth_chainId`](https://eips.ethereum.org/EIPS/eip-695) Ethereum RPC method.
+
+The Provider **MAY** ignore the `chainId` property or interpret it in any manner whatsoever.
```
adrianmcli, line 344: This sounds a little too prescriptive, I wonder if this could just be:

```suggestion
This EIP does not specify how the Provider should interpret the `chainId` property or whether or not it will be handled at all.
```

adrianmcli: I also understand the word **MAY** is supposed to say the same thing, but I guess it just reads too strongly to me. Feel free to use your own judgment.

rekmarks:
I've changed my mind about this, and I'd rather we didn't do it.

PR #2602, opened 2020-04-15
EIP-2602: Disable null hash message verification for ecrecover precompile

sorpaas:
Disable null hash message `ecrecover` operation in the precompile, due to it being potentially unsafe.

PR #2612, opened 2020-04-24
ERC for `permit`: 712-signed token approvals

MrChico:
An ERC for the `permit` token extension, as popularized by the Dai ERC20. Will update eip number and `discussions-to` once I find out which PR this ends up being

### DIFF
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -184,7 +184,7 @@ Additional test vectors for the Dai/Chai-like implementation are given at [chai.
 
 Note that the latter two implementations differ slightly from the presentation given here.
 
-## Security considerations
+## Security Considerations
 
 Though the signer of a `Permit` may have a certain party in mind to submit their transaction, another party can always front run this transaction and call `permit` before the intended party. The end result is the same for the `Permit` signer, however.
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -190,7 +190,7 @@ Some basic tests can be found here https://github.com/Uniswap/uniswap-v2-core/bl
 
 Though the signer of a `Permit` may have a certain party in mind to submit their transaction, another party can always front run this transaction and call `permit` before the intended party. The end result is the same for the `Permit` signer, however.
 
-Since the ECREC precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
+Since the ecrecover precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
 
 Signed `Permit` messages are censorable. The relaying party can always choose to not submit the `Permit` after having received it, withholding the option to submit it. The `deadline` parameter is one mitigation to this. If the signing party holds ETH they can also just submit the `Permit` themselves, which can render previously signed `Permit`s invalid.
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -155,6 +155,8 @@ In other words, the message is the ERC-712 typed structure:
 
 Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
 
+For smoother debugging, and to be able to validate `permit` signed messages in a STATIC environment, we also require the `DOMAIN_SEPARATOR` to be retrievable through `STATICCALL`. In solidity, this is easily achieved by declaring the `DOMAIN_SEPARATOR` to be `public`.
+
 ## Rationale
 The `permit` function is sufficient for enabling any operation involving ERC-20 tokens to be paid for using the token itself, rather than using ETH. 
 An example of a contract which enables gasless token transactions can be found [here](https://github.com/dapphub/ds-dach).
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -165,7 +165,7 @@ The `nonces` mapping is given for replay protection.
 
 A common use case of `permit` has a relayer submit a `Permit` on behalf of the `owner`. In this scenario, the relaying party is essentially given a free option to submit or withhold the `Permit`. If this is a cause of concern, the `owner` can limit the time a `Permit` is valid for by setting `deadline` to a value in the near future. The `deadline` argument can be set to `uint(-1)` to create `Permit`s that effectively never expire.
 
-ERC712 typed messages are included because of its wide spread adoption in many wallet providers.
+ERC-712 typed messages are included because of its wide spread adoption in many wallet providers.
 
 
 ## Backwards Compatibility
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -190,6 +190,8 @@ Though the signer of a `Permit` may have a certain party in mind to submit their
 
 Signed `Permit` messages are censorable. The relaying party can always choose to not submit the `Permit` after having received it, withholding the option to submit it. The `deadline` parameter is one mitigation to this. If the signing party holds ETH they can also just submit the `Permit` themselves, which can render previously signed `Permit`s invalid.
 
+The standard [ERC-20 race condition for approvals](https://swcregistry.io/docs/SWC-114) applies to `permit` as well.
+
 ## Copyright
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -192,7 +192,7 @@ Some basic tests can be found here https://github.com/Uniswap/uniswap-v2-core/bl
 
 Though the signer of a `Permit` may have a certain party in mind to submit their transaction, another party can always front run this transaction and call `permit` before the intended party. The end result is the same for the `Permit` signer, however.
 
-Since the ecrecover precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
+Since the ecrecover precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner != address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
 
 Signed `Permit` messages are censorable. The relaying party can always choose to not submit the `Permit` after having received it, withholding the option to submit it. The `deadline` parameter is one mitigation to this. If the signing party holds ETH they can also just submit the `Permit` themselves, which can render previously signed `Permit`s invalid.
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -155,7 +155,6 @@ In other words, the message is the ERC-712 typed structure:
 
 Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
 
-For smoother debugging, and to be able to validate `permit` signed messages in a STATIC environment, we also require the `DOMAIN_SEPARATOR` to be retrievable through `STATICCALL`. In solidity, this is easily achieved by declaring the `DOMAIN_SEPARATOR` to be `public`.
 
 ## Rationale
 The `permit` function is sufficient for enabling any operation involving ERC-20 tokens to be paid for using the token itself, rather than using ETH. 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -1,6 +1,6 @@
 ---
 eip: 2612
-title: permit: 712-signed approvals
+title: permit – 712-signed approvals
 author: Martin Lundfall (@Mrchico)
 discussions-to: https://github.com/ethereum/EIPs/issues/2613
 status: Draft
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -26,7 +26,7 @@ while adding as little as possible over the existing ERC20 standard.
 Arguably one of the main reasons for the success of ERC20 tokens lies in the interplay between `approve` and `transferFrom`, 
 which allows for tokens to not only be transferred between externally owned accounts (EOA), but to be used in other contracts under application specific conditions by abstracting away `msg.sender` as the defining mechanism for token access control.
 
-However, a limiting factor in this design stems from the fact that the ERC20 `approve` function itself is defined in terms of `msg.sender`. This means that users _initial action_ involving ERC20 tokens must be performed by an EOA <sup>[1]</sup>. This means that token holders need to hold ETH to pay for transaction gas costs and, if they want to use the token in another DeFi system, that more than one transaction is required.
+However, a limiting factor in this design stems from the fact that the ERC20 `approve` function itself is defined in terms of `msg.sender`. This means that user's _initial action_ involving ERC20 tokens must be performed by an EOA <sup>[1]</sup>. If the user needs to interact with a smart contract, then they need to make 2 transactions (`approve` and the smart contract call which will internally call `transferFrom`). Even in the simple use case of paying another person, they need to hold ETH to pay for transaction gas costs.
 
 This ERC extends the ERC20 standard with a new function `permit`, which allows users to modify the `allowance` mapping using a signed message, instead of through `msg.sender`. 
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -37,7 +37,7 @@ For an improved user experience, the signed data is structured following [ERC-71
 <!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
 While ERC-20 tokens have become ubiquotous in the Ethereum ecosystem, their status remains that of second class tokens from the perspective of the protocol. The ability for users to interact with Ethereum without holding any ETH has been a [long outstanding goal](https://github.com/ethereum/EIPs/blob/ed621645c8f3bc5756492f327cda015f35d9f8da/EIPS/eip-101.md) and the [subject](https://eips.ethereum.org/EIPS/eip-1077) [of](https://eips.ethereum.org/EIPS/eip-777) [many](https://github.com/ethereum/EIPs/issues/1776#) [EIPs](https://eips.ethereum.org/EIPS/eip-1271).
 
-So far, many of these proposals have seen very little adoption, and the ones that have been adopted (such as ERC777), introduce a lot of additional functionality, causing [unexpected behavior in mainstream contracts](https://medium.com/consensys-diligence/uniswap-audit-b90335ac007).
+So far, many of these proposals have seen very little adoption, and the ones that have been adopted (such as [ERC-777](https://eips.ethereum.org/EIPS/eip-777)), introduce a lot of additional functionality, causing [unexpected behavior in mainstream contracts](https://medium.com/consensys-diligence/uniswap-audit-b90335ac007).
 
 This ERC proposes an alternative solution which is designed to be as minimal as possible and to only address _one problem_: the lack of abstraction in the ERC-20 `approve` method. 
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -14,7 +14,7 @@ requires: ERC20, ERC712
 
 ## Simple Summary
 <!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
-A function `permit` extending ERC20 which allows for approvals to be made via `secp256k1` signatures. This kind "account abstraction for ERC20" brings about two main benefits:
+A function `permit` extending ERC20 which allows for approvals to be made via `secp256k1` signatures. This kind of "account abstraction for ERC20" brings about two main benefits:
 
 - transactions involving ERC20 operations can be paid using the token itself rather than ETH,
 - approve and pull operations can happen in a single transaction instead of two consecutive transactions,
@@ -188,4 +188,3 @@ Note that the latter two implementations differ slightly from the presentation g
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
 
 [1] -  Unless the address owning the token is actually a contract wallet. Although contract wallets solves many of the same problems that motivates this EIP, they are currently only scarcely adopted in the ecosystem. Contract wallets suffer from a UX problem -- since they separate the EOA `owner` of the contract wallet from the contract wallet itself (which is meant to carry out actions on the `owner`s behalf and holds all of their funds), user interfaces need to be specifically designed to support them. The `permit` pattern reaps many of the same benefits while requiring little to no change in user interfaces.
-
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -72,25 +72,25 @@ if and only if the following conditions are met:
 - `nonces[owner]` (before the state update) is equal to `nonce`.
 - `r`, `s` and `v` is a valid `secp256k1` signature from `owner` of the message:
 
+in solidity pseudocode:
 ```solidity
-keccak256(abi.encodePacked(
-   hex"1901",
-   keccak256(abi.encodePacked(
-            keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
-            keccak256(bytes(erc20name)),
-            keccak256(bytes(version)),
-            chainid,
-            tokenAddress)),
-   keccak256(abi.encodePacked(
-            keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)"),
-            owner,
-            spender,
-            value,
-            nonce,
-            deadline))
+keccak256(hex"1901"
+          ++ keccak256(
+                keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
+             ++ keccak256(bytes(erc20name))
+             ++ keccak256(bytes(version))
+             ++ bytes(chainid)
+             ++ bytes(tokenAddress))
+          ++ keccak256(
+                 keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)")
+              ++ bytes(owner)
+              ++ bytes(spender)
+              ++ bytes(value)
+              ++ bytes(nonce)
+              ++ bytes(deadline))
 ))
 ```
-where `tokenAddress` is the address of the token contract, `chainid` is the chain id of the network it is deployed to and `erc20name` is the name of the token as defined by `ERC-20`. `version` is a `string` defined at contract deployment which remains constant throughout the lifetime of the contract, but is otherwise unconstrained.
+where `++` denotes bytestring concatenation, `tokenAddress` is the address of the token contract, `chainid` is the chain id of the network it is deployed to and `erc20name` is the name of the token as defined by `ERC-20`. `version` is a `string` defined at contract deployment which remains constant throughout the lifetime of the contract, but is otherwise unconstrained.
 
 In other words, the message is the ERC-712 typed structure:
     
@@ -169,24 +169,28 @@ ERC-712 typed messages are included because of its wide spread adoption in many 
 
 
 ## Backwards Compatibility
-There are currently two slightly differing implementations of this ERC, and we are forced to make a choice here for specificity.
-This implies that the given ERC deviates from the implementation given in the Dai and Chai ERC-20 contracts. There, the `permit` method takes `nonce` as an additional argument, and the `uint256 value` argument is exchanged for `bool approval`, admitting binary approvals only. There is also a slight difference in argument names. The specification presented here is in line with the implementation in [Uniswap-v2](https://github.com/uniswap/uniswap-v2-core). This mismatch is a little unfortunate, but not very different from the variations found in ERC-20 contracts. 
+There are already a couple of `permit` functions in token contracts implemented in contracts in the wild, most notably the one introduced in the `dai.sol`.
+
+Its implementation differs slightly from the presentation here in that:
+- instead of taking a `value` argument, it takes a bool `allowed`, setting approval to 0 or `uint(-1)`.
+- the `deadline` argument is instead called `expiry`. This is not just a syntactic change, as it effects the contents of the signed message.
+
+There is also an implementation in the token [`Stake`](https://etherscan.io/address/0x0Ae055097C6d159879521C384F1D2123D1f195e6#code) with the same ABI as `dai` but with different semantics: it lets users issue "expiring approvals", that only allow `transferFrom` to occur while `expiry >= block.timestamp`.
+
+The specification presented here is in line with the implementation in [Uniswap-v2](https://github.com/uniswap/uniswap-v2-core).
 
 ## Test Cases
-Some basic test vector can be found here https://github.com/Uniswap/uniswap-v2-core/blob/master/test/UniswapV2ERC20.spec.ts.
-Additional test vectors for the Dai/Chai-like implementation are given at [chai.t.sol](https://github.com/dapphub/chai/blob/master/src/test/chai.t.sol).
 
+Some basic tests can be found here https://github.com/Uniswap/uniswap-v2-core/blob/master/test/UniswapV2ERC20.spec.ts.
 
 ## Implementation
 [UniswapV2ERC20.sol](https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2ERC20.sol)
-[Dai.sol](https://github.com/makerdao/dss/blob/master/src/dai.sol)
-[Chai.sol](https://github.com/makerdao/dss/blob/master/src/dai.sol)
-
-Note that the latter two implementations differ slightly from the presentation given here.
 
 ## Security Considerations
 
 Though the signer of a `Permit` may have a certain party in mind to submit their transaction, another party can always front run this transaction and call `permit` before the intended party. The end result is the same for the `Permit` signer, however.
+
+Since the ECREC precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
 
 Signed `Permit` messages are censorable. The relaying party can always choose to not submit the `Permit` after having received it, withholding the option to submit it. The `deadline` parameter is one mitigation to this. If the signing party holds ETH they can also just submit the `Permit` themselves, which can render previously signed `Permit`s invalid.
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -156,7 +156,7 @@ In other words, the message is the ERC-712 typed structure:
 Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
 
 ## Rationale
-The `permit` function is sufficient for enabling any operation involving erc20 tokens to be paid for using the token itself, rather than using ETH. 
+The `permit` function is sufficient for enabling any operation involving ERC-20 tokens to be paid for using the token itself, rather than using ETH. 
 An example of a contract which enables gasless token transactions can be found [here](https://github.com/dapphub/ds-dach).
 
 It avoids any calls to unknown code.
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -26,7 +26,7 @@ while adding as little as possible over the existing ERC-20 standard.
 Arguably one of the main reasons for the success of ERC-20 tokens lies in the interplay between `approve` and `transferFrom`, 
 which allows for tokens to not only be transferred between externally owned accounts (EOA), but to be used in other contracts under application specific conditions by abstracting away `msg.sender` as the defining mechanism for token access control.
 
-However, a limiting factor in this design stems from the fact that the ERC-20 `approve` function itself is defined in terms of `msg.sender`. This means that user's _initial action_ involving ERC20 tokens must be performed by an EOA <sup>[1]</sup>. If the user needs to interact with a smart contract, then they need to make 2 transactions (`approve` and the smart contract call which will internally call `transferFrom`). Even in the simple use case of paying another person, they need to hold ETH to pay for transaction gas costs.
+However, a limiting factor in this design stems from the fact that the ERC-20 `approve` function itself is defined in terms of `msg.sender`. This means that user's _initial action_ involving ERC-20 tokens must be performed by an EOA <sup>[1]</sup>. If the user needs to interact with a smart contract, then they need to make 2 transactions (`approve` and the smart contract call which will internally call `transferFrom`). Even in the simple use case of paying another person, they need to hold ETH to pay for transaction gas costs.
 
 This ERC extends the ERC-20 standard with a new function `permit`, which allows users to modify the `allowance` mapping using a signed message, instead of through `msg.sender`. 
 
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -1,8 +1,8 @@
 ---
-eip: <to be assigned>
+eip: 2612
 title: `permit`: 712-signed approvals
 author: Martin Lundfall (@Mrchico)
-discussions-to: <TBD>
+discussions-to: https://github.com/ethereum/EIPs/pull/2612
 status: Draft
 type: Standards Track
 category: ERC
@@ -24,7 +24,7 @@ while adding as little as possible over the existing ERC20 standard.
 ## Abstract
 <!--A short (~200 word) description of the technical issue being addressed.-->
 Arguably one of the main reasons for the success of ERC20 tokens lies in the interplay between `approve` and `transferFrom`, 
-which allows for tokens to not only be transfered between externally owned accounts (EOA), but to be used in other contracts under application specific conditions by abstracting away `msg.sender` as the defining mechanism for token access control.
+which allows for tokens to not only be transferred between externally owned accounts (EOA), but to be used in other contracts under application specific conditions by abstracting away `msg.sender` as the defining mechanism for token access control.
 
 However, a limiting factor in this design stems from the fact that the ERC20 `approve` function itself is defined in terms of `msg.sender`. This means that users _initial action_ involving ERC20 tokens must be performed by an EOA <sup>[1]</sup>. This means that token holders need to hold ETH to pay for transaction gas costs and, if they want to use the token in another DeFi system, that more than one transaction is required.
 
@@ -156,7 +156,6 @@ In other words, the message is the 712 typed structure:
 Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
 
 ## Rationale
-<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
 The `permit` function is sufficient for enabling any operation involving erc20 tokens to be paid for using the token itself, rather than using ETH. 
 An example of a contract which enables gasless token transactions can be found [here](https://github.com/dapphub/ds-dach).
 
@@ -164,7 +163,7 @@ It avoids any calls to unknown code.
 
 The `nonces` mapping is given for replay protection.
 
-A common use case of `permit` has a relayer submit a `Permit` on behalf of the `owner`. In this scenario, the relaying party is essentially given a free option to submit or withold the `Permit`. If this is a cause of concern, the `owner` can limit the time a `Permit` is valid for by setting `deadline` to a value in the near future. The `deadline` argument can be set to `uint(-1)` to create `Permit`s that effectively never expire.
+A common use case of `permit` has a relayer submit a `Permit` on behalf of the `owner`. In this scenario, the relaying party is essentially given a free option to submit or withhold the `Permit`. If this is a cause of concern, the `owner` can limit the time a `Permit` is valid for by setting `deadline` to a value in the near future. The `deadline` argument can be set to `uint(-1)` to create `Permit`s that effectively never expire.
 
 ERC712 typed messages are included because of its wide spread adoption in many wallet providers.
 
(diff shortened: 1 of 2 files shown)


EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+category: ERC
+created: 2020-04-13
+requires: ERC20, ERC712
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
+A function `permit` extending ERC20 which allows for approvals to be made via `secp256k1` signatures. This kind "account abstraction for ERC20" brings about two main benefits:
```
gakonst, line 17:
```suggestion
A function `permit` extending ERC20 which allows for approvals to be made via `secp256k1` signatures. This kind of "account abstraction for ERC20" brings about two main benefits:
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+- approve and pull operations can happen in a single transaction instead of two consecutive transactions,
+
+while adding as little as possible over the existing ERC20 standard.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Arguably one of the main reasons for the success of ERC20 tokens lies in the interplay between `approve` and `transferFrom`, 
+which allows for tokens to not only be transferred between externally owned accounts (EOA), but to be used in other contracts under application specific conditions by abstracting away `msg.sender` as the defining mechanism for token access control.
+
+However, a limiting factor in this design stems from the fact that the ERC20 `approve` function itself is defined in terms of `msg.sender`. This means that users _initial action_ involving ERC20 tokens must be performed by an EOA <sup>[1]</sup>. This means that token holders need to hold ETH to pay for transaction gas costs and, if they want to use the token in another DeFi system, that more than one transaction is required.
```
gakonst, line 29:
```suggestion
However, a limiting factor in this design stems from the fact that the ERC20 `approve` function itself is defined in terms of `msg.sender`. This means that user's _initial action_ involving ERC20 tokens must be performed by an EOA <sup>[1]</sup>. If the user needs to interact with a smart contract, then they need to make 2 transactions (`approve` and the smart contract call which will internally call `transferFrom`). Even in the simple use case of paying another person, they need to hold ETH to pay for transaction gas costs.
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+
+ - the desired specifics of such functions, such as decision regarding fees for `transfer_by_signature`, possible batching algorithms, varies depending on the use case, and,
+ - they can be implemented using a combination of `permit` and additional helper contracts without loss of generality.
+
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+A new method 
+```sol
+function permit(address owner, address spender, uint value, uint deadline, uint8 v, bytes32 r, bytes32 s)
```
gakonst, line 54: nit: maybe use `uint256` instead of `uint`

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Rationale
+The `permit` function is sufficient for enabling any operation involving erc20 tokens to be paid for using the token itself, rather than using ETH. 
+An example of a contract which enables gasless token transactions can be found [here](https://github.com/dapphub/ds-dach).
+
+It avoids any calls to unknown code.
+
+The `nonces` mapping is given for replay protection.
+
+A common use case of `permit` has a relayer submit a `Permit` on behalf of the `owner`. In this scenario, the relaying party is essentially given a free option to submit or withhold the `Permit`. If this is a cause of concern, the `owner` can limit the time a `Permit` is valid for by setting `deadline` to a value in the near future. The `deadline` argument can be set to `uint(-1)` to create `Permit`s that effectively never expire.
```
gakonst, line 166: 👍 this is a good point

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+
+ERC712 typed messages are included because of its wide spread adoption in many wallet providers.
+
+
+## Backwards Compatibility
+There are currently two slightly differing implementations of this ERC, and we are forced to make a choice here for specificity.
+This implies that the given ERC deviates from the implementation given in the Dai and Chai ERC20 contracts. There, the `permit` method takes `nonce` as an additional argument, and the `uint256 value` argument is exchanged for `bool approval`, admitting binary approvals only. There is also a slight difference in argument names. The specification presented here is in line with the implementation in [Uniswap-v2](https://github.com/uniswap/uniswap-v2-core). This mismatch is a little unfortunate, but not very different from the variations found in ERC20 contracts.
```
gakonst, line 173: Is this a big deal? Could potentially have a 1-way conversion contract for migrating Chai to a version which is compatible with this ERC. Also not sure how widespread usage of the DAI `permit` at the moment is

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
+---
+eip: 2612
+title: `permit`: 712-signed approvals
+author: Martin Lundfall (@Mrchico)
+discussions-to: https://github.com/ethereum/EIPs/pull/2612
```
axic, line 5: According to "rules" cant use the PR as the URL.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2020-04-13
+requires: ERC20, ERC712
```
axic, line 10:
```suggestion
requires: 20, 712
```

EIPS/eip-2612.md
axic, line 17: Please use `ERC-20` throughout and make sure the first use is a link to https://eips.ethereum.org/EIPS/eip-20

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+While it may be tempting to introduce `*_by_signature` counterparts for every ERC20 function, they are intentionally left out of this ERC20 for two reasons:
+
+ - the desired specifics of such functions, such as decision regarding fees for `transfer_by_signature`, possible batching algorithms, varies depending on the use case, and,
+ - they can be implemented using a combination of `permit` and additional helper contracts without loss of generality.
+
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+A new method 
+```sol
```
axic, line 53: Please use this throughout:

```suggestion
```solidity
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,191 @@
...
+Additional test vectors for the Dai/Chai-like implementation are given at [chai.t.sol](https://github.com/dapphub/chai/blob/master/src/test/chai.t.sol).
+
+
+## Implementation
+[UniswapV2ERC20.sol](https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2ERC20.sol)
+[Dai.sol](https://github.com/makerdao/dss/blob/master/src/dai.sol)
+[Chai.sol](https://github.com/makerdao/dss/blob/master/src/dai.sol)
+
+Note that the latter two implementations differ slightly from the presentation given here.
+
```
axic, line 186: Please add a "Security Considerations" section.

MrChico:
Thank you @axic and @gakonst for the review. Discussions about the ERC itself are now directed to https://github.com/ethereum/EIPs/issues/2613, leaving this PR for discussing EIP submission technicalities

EIPS/eip-2612.md
Nipol: I was looking forward to talking about this. I'd like to talk about the need for a "deadline"

What aspects of Tx Relayer do we need to take into account "deadline"? The Tx Relayer may have a response that causes tx while setting tx fee low. or no response.

And when a user signs for an interface, it requires more information. or more knowledge. Is usability not compromised? and more complexity for Tx Relayer.

So I think the deadline is unnecessary. 🤔

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+## Security considerations
```
axic, line 187:
```suggestion
## Security Considerations
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+A new method 
+```solidity
+function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+```
+and a new storage item
+```solidity
+mapping(address=>uint256) nonces;
```
axic, line 58: According to the Solidity style guide:
```suggestion
mapping(address => uint256) nonces;
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+if and only if the following conditions are met:
+
+
+- The current blocktime is less than or equal to `deadline`.
+- `owner` is not the zero address.
+- `nonces[owner]` (before the state update) is equal to `nonce`.
+- `r`, `s` and `v` is a valid `secp256k1` signature from `owner` of the message:
+
+```solidity
+keccak256(abi.encodePacked(
```
axic, line 76: Is there a nicer way to specify this in EIP712 "terms" ?

axic: This is too much compiler version dependant.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+            owner,
+            spender,
+            value,
+            nonce,
+            deadline))
+))
+```
+where `tokenAddress` is the address of the token contract, `chainid` is the chain id of the network it is deployed to and `erc20name` is the name of the token as defined by `ERC-20`. `version` is a `string` defined at contract deployment which remains constant throughout the lifetime of the contract, but is otherwise unconstrained.
+
+In other words, the message is the 712 typed structure:
```
axic, line 95:
```suggestion
In other words, the message is the ERC-712 typed structure:
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+The `permit` function is sufficient for enabling any operation involving erc20 tokens to be paid for using the token itself, rather than using ETH. 
+An example of a contract which enables gasless token transactions can be found [here](https://github.com/dapphub/ds-dach).
+
+It avoids any calls to unknown code.
+
+The `nonces` mapping is given for replay protection.
+
+A common use case of `permit` has a relayer submit a `Permit` on behalf of the `owner`. In this scenario, the relaying party is essentially given a free option to submit or withhold the `Permit`. If this is a cause of concern, the `owner` can limit the time a `Permit` is valid for by setting `deadline` to a value in the near future. The `deadline` argument can be set to `uint(-1)` to create `Permit`s that effectively never expire.
+
+ERC712 typed messages are included because of its wide spread adoption in many wallet providers.
```
axic, line 168:
```suggestion
ERC-712 typed messages are included because of its wide spread adoption in many wallet providers.
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+author: Martin Lundfall (@Mrchico)
+discussions-to: https://github.com/ethereum/EIPs/issues/2613
+status: Draft
+type: Standards Track
+category: ERC
+created: 2020-04-13
+requires: 20, 712
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
```
axic, line 13: Nitpick: all these comments are not needed.

EIPS/eip-2612.md
MrChico: It is specified in EIP712 terms below. I'm happy to change this definition to be pseudocode instead of solidity if you prefer

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+    "nonce": nonce,
+    "deadline": deadline
+  }
+}}
+```
+
+Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
+
+## Rationale
+The `permit` function is sufficient for enabling any operation involving erc20 tokens to be paid for using the token itself, rather than using ETH.
```
axic, line 159:
```suggestion
The `permit` function is sufficient for enabling any operation involving ERC-20 tokens to be paid for using the token itself, rather than using ETH. 
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+- approve and pull operations can happen in a single transaction instead of two consecutive transactions,
+
+while adding as little as possible over the existing ERC-20 standard.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Arguably one of the main reasons for the success of ERC-20 tokens lies in the interplay between `approve` and `transferFrom`, 
+which allows for tokens to not only be transferred between externally owned accounts (EOA), but to be used in other contracts under application specific conditions by abstracting away `msg.sender` as the defining mechanism for token access control.
+
+However, a limiting factor in this design stems from the fact that the ERC-20 `approve` function itself is defined in terms of `msg.sender`. This means that user's _initial action_ involving ERC20 tokens must be performed by an EOA <sup>[1]</sup>. If the user needs to interact with a smart contract, then they need to make 2 transactions (`approve` and the smart contract call which will internally call `transferFrom`). Even in the simple use case of paying another person, they need to hold ETH to pay for transaction gas costs.
```
axic, line 29:
```suggestion
However, a limiting factor in this design stems from the fact that the ERC-20 `approve` function itself is defined in terms of `msg.sender`. This means that user's _initial action_ involving ERC-20 tokens must be performed by an EOA <sup>[1]</sup>. If the user needs to interact with a smart contract, then they need to make 2 transactions (`approve` and the smart contract call which will internally call `transferFrom`). Even in the simple use case of paying another person, they need to hold ETH to pay for transaction gas costs.
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+While ERC-20 tokens have become ubiquotous in the Ethereum ecosystem, their status remains that of second class tokens from the perspective of the protocol. The ability for users to interact with Ethereum without holding any ETH has been a [long outstanding goal](https://github.com/ethereum/EIPs/blob/ed621645c8f3bc5756492f327cda015f35d9f8da/EIPS/eip-101.md) and the [subject](https://eips.ethereum.org/EIPS/eip-1077) [of](https://eips.ethereum.org/EIPS/eip-777) [many](https://github.com/ethereum/EIPs/issues/1776#) [EIPs](https://eips.ethereum.org/EIPS/eip-1271).
+
+So far, many of these proposals have seen very little adoption, and the ones that have been adopted (such as ERC777), introduce a lot of additional functionality, causing [unexpected behavior in mainstream contracts](https://medium.com/consensys-diligence/uniswap-audit-b90335ac007).
```
axic, line 40:
```suggestion
So far, many of these proposals have seen very little adoption, and the ones that have been adopted (such as [ERC-777](https://eips.ethereum.org/EIPS/eip-777)), introduce a lot of additional functionality, causing [unexpected behavior in mainstream contracts](https://medium.com/consensys-diligence/uniswap-audit-b90335ac007).
```

EIPS/eip-2612.md
MrChico: I have to respectfully disagree with the style guide here

axic: It is only a suggestion, feel free to ignore.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+A new method 
+```solidity
+function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+```
+and a new storage item
+```solidity
+mapping(address=>uint256) nonces;
```
Nipol, line 58: Is it fine to set this option explicitly to `public`?

This means to verify that a `permit` request has been accepted.

MrChico: Yes. The erc specifies that the nonce container should be accompanied with a getter function, which can be achieved by marking it with the `public` keyword

EIPS/eip-2612.md
```diff
@@ -169,25 +169,29 @@ ERC-712 typed messages are included because of its wide spread adoption in many
...
-[Dai.sol](https://github.com/makerdao/dss/blob/master/src/dai.sol)
-[Chai.sol](https://github.com/makerdao/dss/blob/master/src/dai.sol)
-
-Note that the latter two implementations differ slightly from the presentation given here.
 
 ## Security Considerations
 
 Though the signer of a `Permit` may have a certain party in mind to submit their transaction, another party can always front run this transaction and call `permit` before the intended party. The end result is the same for the `Permit` signer, however.
 
+Since the ECREC precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
```
axic, line 193: I think `ecrecover precmopile` is a more commonly used term.

Also could refer to https://github.com/ethereum/EIPs/pull/2602 here.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,202 @@
...
+Some basic tests can be found here https://github.com/Uniswap/uniswap-v2-core/blob/master/test/UniswapV2ERC20.spec.ts.
+
+## Implementation
+[UniswapV2ERC20.sol](https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2ERC20.sol)
+
+## Security Considerations
+
+Though the signer of a `Permit` may have a certain party in mind to submit their transaction, another party can always front run this transaction and call `permit` before the intended party. The end result is the same for the `Permit` signer, however.
+
+Since the ECREC precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
```
MrChico, line 193:
```suggestion
Since the ecrecover precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
```

petejkim:
`nonces` should be something less generic like `permitNonces`, the name `nonces` may already be in use.

NoahZinsmeister:
> nonces should be something less generic like permitNonces, the name nonces may already be in use.

1) i'd argue the most natural use for a `nonces` mapping in a smart contract is in fact to ensure signature uniqueness.
2) it's likely that an existing `nonces` mapping could be use dually by `permit` and another pattern, as the use of an incrementing nonce is a fairly generic behavior.
3) `nonces` is already used in the wild in various `permit` implementations. are there any known uses outside of these?

MrChico:
@axic initial discussions seem to have settled down. Is this good to merge?

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+Some basic tests can be found here https://github.com/Uniswap/uniswap-v2-core/blob/master/test/UniswapV2ERC20.spec.ts.
+
+## Implementation
+[UniswapV2ERC20.sol](https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2ERC20.sol)
+
+## Security Considerations
+
+Though the signer of a `Permit` may have a certain party in mind to submit their transaction, another party can always front run this transaction and call `permit` before the intended party. The end result is the same for the `Permit` signer, however.
+
+Since the ecrecover precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner =/= address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
```
lightclient, line 195:
```suggestion
`owner != address(0)`
```

lightclient, line 195:
```suggestion
Since the ecrecover precompile fails silently and just returns the zero address as `signer` when given malformed messages, it is important to ensure `owner != address(0)` to avoid `permit` from creating an approval to spend "zombie funds" belong to the zero address.
```

petejkim:
> > nonces should be something less generic like permitNonces, the name nonces may already be in use.
> 
> 1. i'd argue the most natural use for a `nonces` mapping in a smart contract is in fact to ensure signature uniqueness.
> 2. it's likely that an existing `nonces` mapping could be use dually by `permit` and another pattern, as the use of an incrementing nonce is a fairly generic behavior.
> 3. `nonces` is already used in the wild in various `permit` implementations. are there any known uses outside of these?

I found one instance where `permitNonces` is used as opposed to `nonces`: https://github.com/ourzora/zora-contracts/blob/master/contracts/Token.sol#L21

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+    "value": value,
+    "nonce": nonce,
+    "deadline": deadline
+  }
+}}
+```
+
+Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
+
+For smoother debugging, and to be able to validate `permit` signed messages in a STATIC environment, we also require the `DOMAIN_SEPARATOR` to be retrievable through `STATICCALL`. In solidity, this is easily achieved by declaring the `DOMAIN_SEPARATOR` to be `public`.
```
eMarchenko, line 158:
```
In solidity, this is easily achieved by declaring the `DOMAIN_SEPARATOR` to be `public`.
```
probably it should be `view` instead, isn't it?

MrChico: The getters for `public` variables are always interpreted as `view`

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+ - the desired specifics of such functions, such as decision regarding fees for `transfer_by_signature`, possible batching algorithms, varies depending on the use case, and,
+ - they can be implemented using a combination of `permit` and additional helper contracts without loss of generality.
+
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+A new method 
+```solidity
+function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
```
cleanunicorn, line 54: Is this function missing the `uint256 nonce` argument?`

MrChico: the `nonce` does not need to be given as an argument to the `permit` function, even though it needs to be defined in the `Permit` message. To check the validity of the `Permit` message you can simply let `nonce = nonces[owner]`. If the nonce is correct, the signed message will be correct. If the nonce is wrong, the signature will be off

cleanunicorn: Just to be clear, the nonce which is currently saved in the contract storage represents the next nonce to be used. This is why the nonce does not need to be specified, because we already know the next nonce. 

This also means that nonces have to be consecutive, nonces SHOULD not be skipped (otherwise a lot of gas needs to be used to validate the nonce).

cleanunicorn: Do you think it makes sense to add a note about the missing `nonce` argument?

Nipol: In my opinion, we have considered nonce verification throughout the specification, so we don't need to use it as far as the interface.

cleanunicorn: I agree it should not be used. It doesn't need to be there.

EIPS/eip-2612.md
eMarchenko: I see it now, ty!

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
+---
+eip: 2612
+title: permit: 712-signed approvals
```
MicahZoltu, line 3: I *think* the bot isn't happy with the `:` here.  Try this and see if it satisfies the bot:
```suggestion
title: permit - 712-signed approvals
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
+A function `permit` extending [ERC-20](https://eips.ethereum.org/EIPS/eip-20) which allows for approvals to be made via `secp256k1` signatures. This kind of "account abstraction for ERC-20" brings about two main benefits:
+
+- transactions involving ERC-20 operations can be paid using the token itself rather than ETH,
+- approve and pull operations can happen in a single transaction instead of two consecutive transactions,
+
+while adding as little as possible over the existing ERC-20 standard.
```
MicahZoltu, line 22: Most of this should be moved to the motivation section.  The Simple Summary section is meant to be a very terse/pithy "what", while the motivation section is for "why".

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+category: ERC
+created: 2020-04-13
+requires: 20, 712
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
+A function `permit` extending [ERC-20](https://eips.ethereum.org/EIPS/eip-20) which allows for approvals to be made via `secp256k1` signatures. This kind of "account abstraction for ERC-20" brings about two main benefits:
```
MicahZoltu, line 17: Use relative path links with exact matches:
```
[ERC-20](./eip-20.md)
```

MicahZoltu, line 17: Grammar:
```
A `permit` function that can be added to an ERC20 token which allows users to transfer tokens while someone else pays for gas.
```

EIPS/eip-2612.md
MicahZoltu, line 3: I *suspect* EIP-2711 is going to go through eventually (my hope is in the HF after Berlin).  You may want to consider whether this EIP is still valuable after that, and if not whether it is worth standardizing if it will only be valuable for a relatively short amount of time.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+which allows for tokens to not only be transferred between externally owned accounts (EOA), but to be used in other contracts under application specific conditions by abstracting away `msg.sender` as the defining mechanism for token access control.
+
+However, a limiting factor in this design stems from the fact that the ERC-20 `approve` function itself is defined in terms of `msg.sender`. This means that user's _initial action_ involving ERC-20 tokens must be performed by an EOA <sup>[1]</sup>. If the user needs to interact with a smart contract, then they need to make 2 transactions (`approve` and the smart contract call which will internally call `transferFrom`). Even in the simple use case of paying another person, they need to hold ETH to pay for transaction gas costs.
+
+This ERC extends the ERC-20 standard with a new function `permit`, which allows users to modify the `allowance` mapping using a signed message, instead of through `msg.sender`. 
+
+For an improved user experience, the signed data is structured following [ERC-712](https://eips.ethereum.org/EIPS/eip-712), which already has wide spread adoption in major RPC providers.
```
MicahZoltu, line 33:
```suggestion
For an improved user experience, the signed data is structured following [ERC-712](./eip-712.md), which already has wide spread adoption in major RPC providers.
```

MicahZoltu, line 33: The wording of this section feels more like an argument for why this should be standardized rather than an abstract that explains the specification in plain English.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+While ERC-20 tokens have become ubiquotous in the Ethereum ecosystem, their status remains that of second class tokens from the perspective of the protocol. The ability for users to interact with Ethereum without holding any ETH has been a [long outstanding goal](https://github.com/ethereum/EIPs/blob/ed621645c8f3bc5756492f327cda015f35d9f8da/EIPS/eip-101.md) and the [subject](https://eips.ethereum.org/EIPS/eip-1077) [of](https://eips.ethereum.org/EIPS/eip-777) [many](https://github.com/ethereum/EIPs/issues/1776#) [EIPs](https://eips.ethereum.org/EIPS/eip-1271).
+
+So far, many of these proposals have seen very little adoption, and the ones that have been adopted (such as [ERC-777](https://eips.ethereum.org/EIPS/eip-777)), introduce a lot of additional functionality, causing [unexpected behavior in mainstream contracts](https://medium.com/consensys-diligence/uniswap-audit-b90335ac007).
```
MicahZoltu, line 40:
```suggestion
So far, many of these proposals have seen very little adoption, and the ones that have been adopted (such as [ERC-777](./eip-777.md)), introduce a lot of additional functionality, causing [unexpected behavior in mainstream contracts](https://medium.com/consensys-diligence/uniswap-audit-b90335ac007).
```

MicahZoltu, line 40: I recommend against including external links in EIPs whenever possible.  If someone looks at this specification in 10 years, it is incredibly likely that the links will no longer work.  In this case, instead include a short summary of the problem rather than a link to an article describing the problem.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+For an improved user experience, the signed data is structured following [ERC-712](https://eips.ethereum.org/EIPS/eip-712), which already has wide spread adoption in major RPC providers.
+
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+While ERC-20 tokens have become ubiquotous in the Ethereum ecosystem, their status remains that of second class tokens from the perspective of the protocol. The ability for users to interact with Ethereum without holding any ETH has been a [long outstanding goal](https://github.com/ethereum/EIPs/blob/ed621645c8f3bc5756492f327cda015f35d9f8da/EIPS/eip-101.md) and the [subject](https://eips.ethereum.org/EIPS/eip-1077) [of](https://eips.ethereum.org/EIPS/eip-777) [many](https://github.com/ethereum/EIPs/issues/1776#) [EIPs](https://eips.ethereum.org/EIPS/eip-1271).
```
MicahZoltu, line 38:
```suggestion
While ERC-20 tokens have become ubiquotous in the Ethereum ecosystem, their status remains that of second class tokens from the perspective of the protocol. The ability for users to interact with Ethereum without holding any ETH has been a [long outstanding goal](./eip-101.md) and the [subject](./eip-1077.md) [of](./eip-777.md) [many](https://github.com/ethereum/EIPs/issues/1776#) [EIPs](./eip-1271.md)
```
Also, recommend removing the link to the GitHub issue since that isn't guaranteed to resolve if we ever move off GitHub.

MicahZoltu, line 38: An EIP is meant to be a standard, not an argument/essay.  You can merely describe the problem, you don't need to defend it with lots of reference material supporting your position.  Convincing people to use the standard is something that should occur outside of the standardization process, rather than part of the standard itself.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+So far, many of these proposals have seen very little adoption, and the ones that have been adopted (such as [ERC-777](https://eips.ethereum.org/EIPS/eip-777)), introduce a lot of additional functionality, causing [unexpected behavior in mainstream contracts](https://medium.com/consensys-diligence/uniswap-audit-b90335ac007).
+
+This ERC proposes an alternative solution which is designed to be as minimal as possible and to only address _one problem_: the lack of abstraction in the ERC-20 `approve` method. 
+
+While it may be tempting to introduce `*_by_signature` counterparts for every ERC-20 function, they are intentionally left out of this ERC-20 for two reasons:
+
+ - the desired specifics of such functions, such as decision regarding fees for `transfer_by_signature`, possible batching algorithms, varies depending on the use case, and,
+ - they can be implemented using a combination of `permit` and additional helper contracts without loss of generality.
```
MicahZoltu, line 47: This block would be a better fit in the `## Rationale` section rather than the `## Motivation` section.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+A new method 
+```solidity
+function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+```
+and a new storage item
+```solidity
+mapping(address=>uint256) nonces;
+```
```
MicahZoltu, line 59: Standards of this sort should not specify implementation details.  It is certainly reasonable to provide an example implementation (in the `## Implementations` section) that uses this, but as far as standards go they should focus solely on the public facing API.  In this case, it sounds like what you want is:
```solidity
function nonces(address holder) external view returns (uint256 nonce);
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+with accompanying getter function are introduced, the semantics of which are as follows:
+
+For all addresses `owner`, `spender`, uint256s `value`, `deadline` and `nonce`, uint8 `v`, bytes32 `r` and `s`, 
+a call to `permit(owner, spender, value, deadline, v, r, s)` will set 
+`approval[owner][spender]` to `value`,
+increment `nonces[owner]` by 1,
```
MicahZoltu, line 65: As indicated above, this internal implementation detail should be left up to the contract.  Some may want to implement this with some novel proof system, or a private storage variable, etc.  What matters is that every call to `permit` should result in all future calls to `nonces(owner)` to return a value that is one greater than it was before the call to `permit` (similar for `approval`).

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+    "owner": owner,
+    "spender": spender,
+    "value": value,
+    "nonce": nonce,
+    "deadline": deadline
+  }
+}}
+```
+
+Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
```
MicahZoltu, line 156: Pithy is good in specifications.
```suggestion
The caller of the `permit` function can be any address.
```

EIPS/eip-2612.md
MicahZoltu, line 158: This doesn't feel like something that needs to be standardized.  While I agree it is a good idea, not all good ideas need to be standardized.  If it is standardized, it should be standardized by its API surface, not implementation (like suggestions above).

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+For smoother debugging, and to be able to validate `permit` signed messages in a STATIC environment, we also require the `DOMAIN_SEPARATOR` to be retrievable through `STATICCALL`. In solidity, this is easily achieved by declaring the `DOMAIN_SEPARATOR` to be `public`.
+
+## Rationale
+The `permit` function is sufficient for enabling any operation involving ERC-20 tokens to be paid for using the token itself, rather than using ETH. 
+An example of a contract which enables gasless token transactions can be found [here](https://github.com/dapphub/ds-dach).
```
MicahZoltu, line 162: Include example implementations in the `## Implementations` section (inline), rather than linking to external sites/code, whenever possible.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+It avoids any calls to unknown code.
```
MicahZoltu, line 164: Consider expanding on this a bit and explaining how it achieves this.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+
+## Backwards Compatibility
+There are already a couple of `permit` functions in token contracts implemented in contracts in the wild, most notably the one introduced in the `dai.sol`.
+
+Its implementation differs slightly from the presentation here in that:
+- instead of taking a `value` argument, it takes a bool `allowed`, setting approval to 0 or `uint(-1)`.
+- the `deadline` argument is instead called `expiry`. This is not just a syntactic change, as it effects the contents of the signed message.
+
+There is also an implementation in the token [`Stake`](https://etherscan.io/address/0x0Ae055097C6d159879521C384F1D2123D1f195e6#code) with the same ABI as `dai` but with different semantics: it lets users issue "expiring approvals", that only allow `transferFrom` to occur while `expiry >= block.timestamp`.
```
MicahZoltu, line 180: As above, remove external links in favor of summaries, inline code, etc.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+The specification presented here is in line with the implementation in [Uniswap-v2](https://github.com/uniswap/uniswap-v2-core).
+
+## Test Cases
+
+Some basic tests can be found here https://github.com/Uniswap/uniswap-v2-core/blob/master/test/UniswapV2ERC20.spec.ts.
```
MicahZoltu, line 186: Can these test cases be formatted in a language neutral way and inlined here rather than an external link?

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Implementation
+[UniswapV2ERC20.sol](https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2ERC20.sol)
```
MicahZoltu, line 189: Recommend inlining this, and consider stripping out all of the code unrelated to this specification.

EIPS/eip-2612.md
MrChico, line 3:
```suggestion
title: permit – 712-signed approvals
```

EIPS/eip-2612.md
MrChico: Better to link to eips.ethereum.org as per @axic s comment https://github.com/ethereum/EIPs/pull/2612#discussion_r414488488

EIPS/eip-2612.md
MrChico: I think standards should specify the intended semantics, not just the interface. Implementation details have been left out.

EIPS/eip-2612.md
MrChico: I aim for specificity and minimalism with this EIP. Although one could certainly imagine replay protection being brought about in many ways, I think that would be a premature optimization at this point. I'm open for future EIPs to innovate or generalize here, but I want to constrain the scope of this one.

EIPS/eip-2612.md
MicahZoltu: I agree, they definitely should specify intended behavior, but they should not do so by specifying implementation details.  Someone may, for example, use a proxy that contract that manages storage manually rather than through the storage slot mechanism, or they may use some fancy internal merkle tree for storage, or they may just want to do some internal accounting when the nonce is set, or they may want to write a base contract that can be derived from but they don't want the derived contracts to be able to mutate the value, or they may be using a different language that has a totally different mechanism for storage and mappings.

EIPS/eip-2612.md
MrChico: I'm inclined to agree

EIPS/eip-2612.md
```diff
@@ -0,0 +1,204 @@
...
+    "value": value,
+    "nonce": nonce,
+    "deadline": deadline
+  }
+}}
+```
+
+Note that nowhere in this definition we refer to `msg.sender`. The caller of the `permit` function can be any address.
+
+For smoother debugging, and to be able to validate `permit` signed messages in a STATIC environment, we also require the `DOMAIN_SEPARATOR` to be retrievable through `STATICCALL`. In solidity, this is easily achieved by declaring the `DOMAIN_SEPARATOR` to be `public`.
```
MrChico, line 158:
```suggestion
```

EIPS/eip-2612.md
```diff
@@ -0,0 +1,203 @@
...
+- The current blocktime is less than or equal to `deadline`.
+- `owner` is not the zero address.
+- `nonces[owner]` (before the state update) is equal to `nonce`.
+- `r`, `s` and `v` is a valid `secp256k1` signature from `owner` of the message:
+
+in solidity pseudocode:
+```solidity
+keccak256(hex"1901"
+          ++ keccak256(
+                keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
```
wighawag, line 79: A note should be added here that the EIP712Domain does not need to be exactly as specified here.
An implementation might want to add or remove field. The current EIP-712 spec stipulate that "eip712Domain is a struct named EIP712Domain with one or more of the below fields"

EIPS/eip-2612.md
```diff
@@ -0,0 +1,203 @@
...
+          ++ keccak256(
+                 keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)")
+              ++ bytes(owner)
+              ++ bytes(spender)
+              ++ bytes(value)
+              ++ bytes(nonce)
+              ++ bytes(deadline))
+))
+```
+where `++` denotes bytestring concatenation, `tokenAddress` is the address of the token contract, `chainid` is the chain id of the network it is deployed to and `erc20name` is the name of the token as defined by `ERC-20`. `version` is a `string` defined at contract deployment which remains constant throughout the lifetime of the contract, but is otherwise unconstrained.
```
wighawag, line 93: Similar to my comment above, a note should be added that nonce of the `chainId`, `version` or even `name` are not required.

if the EIP intend to have name always be the erc20 name then this should be formulated more clearly.

MicahZoltu:
Sorry for not merging once the bot passed.  I usually check back when I see commits, but for some reason I didn't on this one.

I strongly recommend heeding the feedback left in this PR as it is going to come up again when it comes time for Last Call and the requirements for merging to Last Call (and then final) are much more strict.

EIPS/eip-2612.md
MicahZoltu: I believe we have since discussed this and relative links are better as they work both in GitHub and on the EIPs site and any mirror sites as well.  cc @axic in case you still disagree or I am misremembering having a conversation somewhere with you about this.

EIPS/eip-2612.md
MrChico: I think this is good to emphasize

EIPS/eip-2612.md
axic: Only seen this ping now. We have since discussed and it seems the current recommendation is the absolute link, but perhaps this can be changed in the future.

EIPS/eip-2612.md
wschwab: It's kind of hard to inline ERC tests imho, and the general reader will be expecting a functional test base on the ERC - even if we move towards removing external links, I think a link to a working test base for ERCs should be encouraged.

MicahZoltu: `../assets/eip-####/files` can be used if your test cases are complex.  Note that the purpose of the test cases isn't meant to serve the same purpose as a test suite in an application.  It is meant to exemplify expected input and output combinations, which should be able to be written as a table or simple JSON test cases.  People *should not* be writing a full test suite implementation in this section.

EIPS/eip-2612.md
```diff
@@ -0,0 +1,203 @@
...
+ERC-712 typed messages are included because of its wide spread adoption in many wallet providers.
+
+
+## Backwards Compatibility
+There are already a couple of `permit` functions in token contracts implemented in contracts in the wild, most notably the one introduced in the `dai.sol`.
+
+Its implementation differs slightly from the presentation here in that:
+- instead of taking a `value` argument, it takes a bool `allowed`, setting approval to 0 or `uint(-1)`.
+- the `deadline` argument is instead called `expiry`. This is not just a syntactic change, as it effects the contents of the signed message.
+
```
a6-dou, line 178: `value` and `deadline` are not the only differences, `dai.sol` uses `holder` instead of `owner` also.
in total it's 3 points:
- instead of taking a `value` argument, it takes a bool `allowed`, setting approval to 0 or `uint(-1)`.
- the `deadline` argument is instead called `expiry`. This is not just a syntactic change, as it effects the contents of the signed message.
- `holder` is used instead of `owner`, similar to`deadline` <> `expiry` this will affect the contents of the signed message


@MrChico

a6-dou: https://etherscan.io/token/0x6B175474E89094C44Da98b954EedeAC495271d0F#code#L107

PR #2614, opened 2020-04-24
Add KEthereum implementation to EIP634

ligi:
@decanus mentioned you could use content in the implementation section to move this on from draft. As I had an implementation already I link this here.

### DIFF
--- a/EIPS/eip-634.md
+++ b/EIPS/eip-634.md
@@ -70,7 +70,8 @@ Not applicable.
 TBD
 
 ## Implementation
-None yet.
+
+ * Kotlin implementation in the [ENS module of KEthereum](https://github.com/komputing/KEthereum/tree/master/ens#getting-text-records)
 
 ## Copyright
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


eip-automerger:

 - EIP 634 requires approval from one of (@ricmoo)

axic:
@ricmoo can this be merged?

ligi:
> or add a comment indicating that you have addressed all required feedback and are ready for a review.

this

ricmoo:
I'll look into it soon.

I was originally thinking the implementation would like to the TextProfile on the ENS repo.

But I can add this link too, in which case would it be inappropriate/conflict-of-interest-esque for me to add a link to ethers?

ligi:
> I'll look into it soon

:+1: 


> But I can add this link too, in which case would it be inappropriate/conflict-of-interest-esque for me to add a link to ethers?

I do not think so

MicahZoltu:
Ideally, implementations would not link out to third party libraries but instead would include a reference implementation inline or in `../assets/eip-634/`.  The problem we have run into is two fold:
1. External resources tend to move with time and eventually 404, and we do not have the manpower to maintain dead links in every single EIP.
2. People have begun using EIPs as a form of advertising for their projects, sometimes creating an EIP just so they can drop links to their project in the EIP.  By having a general policy of "please don't link to external resources" we can mostly prevent people from using the EIP system for advertising (which is a huge drain on EIP editors).

If you want to make your implementation discoverable, I recommend SEO for "eip-634 implementation" or something.

MicahZoltu:
Oh, also:
3. EIPs are CC0 license, and linking out to code that isn't CC0 licensed can cause problems.  If someone clicks through a link to GPL code (without realizing it is an external link perhaps) then they lose plausible deniability that they never saw the GPL code and thus they didn't copy it when their code happens to look similar.

ligi:
3. I am pretty sure a link is fine - same as the references in Wikipedia - that said IANAL
2. Not the case here - the EIP was there before and just needed implementations to move out of draft
1. is a fair point

so do you add an implementation to this EIP so it can move out of draft?

ricmoo:
@MicahZoltu Your point 2 is why I was apprehensive about linking to ethers.

I think linking to the ENS implementation still makes sense though. It is certainly not feasible for an EIP to always include complete implementations inline... I like the sentiment of point 1, but I also think the benefits would outweigh the cost. 

I don’t worry too much about your point 3. Plausible deniability is not an excuse for violating license agreements. There should be no assumption that a CC0 document does not contain references to non-CC0 content.

ricmoo:
FYI: this is likely short enough to copy over, however copying *would* be an issue since this code is not public domain.

https://github.com/ensdomains/resolvers/blob/master/contracts/profiles/TextResolver.sol

MicahZoltu:
An implementation is not required an EIP.  There are many EIPs that have no reference implementation and this is fine.  It is often *useful* to have a reference implementation, but not always.  If that is the only reason this EIP is not being proposed for moving forward, I recommend just removing the `Implementation` section entirely.

PR #2617, opened 2020-04-26
ERC-1363: Move to Last Call

vittominacori:
Moves ERC-1363 to Final.

### DIFF
--- a/EIPS/eip-1363.md
+++ b/EIPS/eip-1363.md
@@ -3,7 +3,7 @@ eip: 1363
 title: ERC-1363 Payable Token
 author: Vittorio Minacori (@vittominacori)
 discussions-to: https://github.com/ethereum/eips/issues/1363
-status: Last Call
+status: Final
 type: Standards Track
 category: ERC
 created: 2020-01-31
--- a/EIPS/eip-1363.md
+++ b/EIPS/eip-1363.md
@@ -3,7 +3,7 @@ eip: 1363
 title: ERC-1363 Payable Token
 author: Vittorio Minacori (@vittominacori)
 discussions-to: https://github.com/ethereum/eips/issues/1363
-status: Final
+status: Last Call
 type: Standards Track
 category: ERC
 created: 2020-01-31
--- a/EIPS/eip-1363.md
+++ b/EIPS/eip-1363.md
@@ -4,7 +4,7 @@ title: ERC-1363 Payable Token
 author: Vittorio Minacori (@vittominacori)
 discussions-to: https://github.com/ethereum/eips/issues/1363
 status: Last Call
-review-period-end: 2020-09-20
+review-period-end: 2020-09-30
 type: Standards Track
 category: ERC
 created: 2020-01-31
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 1363 state from Draft to Last Call

axic:
It is not possible to move from draft to final in a single step, please refer to EIP-1.

vittominacori:
Updated to Last Call.

It was Last Call first but automerger fails. So I moved to final.

EIPS/eip-1363.md
```diff
@@ -3,7 +3,7 @@ eip: 1363
 title: ERC-1363 Payable Token
 author: Vittorio Minacori (@vittominacori)
 discussions-to: https://github.com/ethereum/eips/issues/1363
-status: Draft
+status: Last Call
```
MicahZoltu, line 6:
```suggestion
status: Last Call
review-period-end: 2020-09-20
```

MicahZoltu (changes requested):
Last Call must have a `review-period-end` of at least 2 weeks in the future.

MicahZoltu:
`## Security Considerations` section is required for all EIPs.  It comes right before the Copyright section.  If you believe there are no security considerations, the section can just include one-sentence saying as much.

MicahZoltu:
I would also recommend changing:
```
Every Payable Token compliant contract MUST implement the ERC-1363 interface other than ERC-20 and ERC-165 interfaces.
```
to
```
Every Payable Token compliant contract **MUST** implement the ERC-1363 interface as well as the ERC-20 and ERC-165 interfaces.
```

MicahZoltu:
I recommend using the RFC 2119 style for `**MUST**` rather than `MUST`.

MicahZoltu:
Recommend changing:
```
Description of a Payable Token compatible with the ERC-20 definition. Also description of a Token Receiver and/or Spender that accepts Payable Token payments.
```
to
```
Defines a token interface for ERC-20 tokens that supports executing recipient code on transfer.
```

MicahZoltu:
Recommend changing the abstract to be a bit more descriptive, maybe mentioning the highlights of the new functions.  I generally recommend the abstract contain a human-readable and terse version of the specification.  Something like, "Adds a `transferAndCall` method that allows the caller to specify that the that the recipient contract will be called with calldata supplied by the caller. ...".  Maybe add a sentence or two about some other features, but I recommend leaving out the specific details like all of the overloads, and the exact mechanism for callbacks, etc.

Basically, state the core protocol in human readable words, but don't cover all of the edge cases and little details.

MicahZoltu (changes requested):
The Security Considerations section is a hard requirement.  The rest are feedback that I encourage applying, but aren't strictly required to move to Last Call.

MicahZoltu:
I know I said that the other changes were optional, but on doing a final review I think this change is necessary afterall because the current wording doesn't align with the rest of the document:
> Every Payable Token compliant contract **MUST** implement the ERC-1363 interface other than ERC-20 and ERC-165 interfaces.

to
> Every Payable Token compliant contract **MUST** implement the ERC-1363 interface as well as the ERC-20 and ERC-165 interfaces.

"other than" implies that someone implementing this standard **MUST NOT** implement ERC-20 and ERC-165, which I don't think was the intent.

Along with that change, please also set the review period end to 2 weeks from the date of that change as I *hope* this is the last change before Last Call.  😄

MicahZoltu (approved):

PR #2357, opened 2019-11-06
Revert EIP-875 to draft

fulldecent:
Last call is long overdue

PR #2622, opened 2020-04-29
EIP 2622 - Total Difficulty in Block Header

tkstanczak:
This document proposes to replace Dififculty with Total Difficulty in the Ethereum block headers..


EIPS/eip-2357.md
```diff
@@ -0,0 +1,38 @@
...
+
+This simplifies the selection of the best peer, acceptance of the NewBlock broadcasts and stateless client execution without verification from genesis.
+
+## Backwards Compatibility
+
+Clients will have to understand that headers before the EIP will have Difficulty and not total difficulty field.
+
+## Copyright
+Copyright and related rights waived via
+[CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```
MicahZoltu, line 38: Nitpick: These last two lines should probably be one line.

tkstanczak: I wonder, just copied it from other sample EIP

MicahZoltu: I'm not sure where you might have copied it from, but the template has it one-lined.  :)  https://github.com/ethereum/EIPs/blob/master/eip-template.md

edsonayllon: In Markdown, this would show up as one line. You need 2 linebreaks to make one linebreak in Github Markdown.

EIPS/eip-2357.md
```diff
@@ -0,0 +1,37 @@
+---
+eip: 2357
```
axic, line 2: Not sure where 2357 came from, but that seems be something else: https://github.com/ethereum/EIPs/pull/2357

I suggest to use 2622 which is the number of this PR>

EIPS/eip-2357.md
```diff
@@ -0,0 +1,37 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-2357-replace-difficulty-with-total-difficulty-in-block-headers/4233
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-04-29
+---
+
+## Simple Summary
+
+This document proposes to replace Dififculty with Total Difficulty in the Ethereum block headers..
```
axic, line 14:
```suggestion
This document proposes to replace Difficulty with Total Difficulty in the Ethereum block headers.
```

EIPS/eip-2357.md
```diff
@@ -0,0 +1,37 @@
...
+title: Total Difficulty in Block Header
+author: Tomasz K. Stanczak (@tkstanczak)
+discussions-to: https://ethereum-magicians.org/t/eip-2357-replace-difficulty-with-total-difficulty-in-block-headers/4233
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-04-29
+---
+
+## Simple Summary
```
axic, line 12: No need for both `Simple Summary` and `Abstract`. This requirement has been abolished a while ago.

EIPS/eip-2357.md
```diff
@@ -0,0 +1,37 @@
...
+
+Difficulty field will be replaced with TotalDifficulty (sum of difficulties since genesis) in the Ethereum block header.
+
+## Rationale
+
+This simplifies the selection of the best peer, acceptance of the NewBlock broadcasts and stateless client execution without verification from genesis.
+
+## Backwards Compatibility
+
+Clients will have to understand that headers before the EIP will have Difficulty and not total difficulty field.
```
axic, line 34: I suggest to consistently use `TotalDifficulty`, ` Total Difficulty` or `total difficulty`.

EIPS/eip-2357.md
```diff
@@ -0,0 +1,37 @@
...
+
+Replace field Difficulty with TotalDifficulty (sum of difficulties since genesis) in the Ethereum block header.
+
+## Motivation
+
+Simplify total difficulty verification in stateless clients.
+
+## Specification
+
+Difficulty field will be replaced with TotalDifficulty (sum of difficulties since genesis) in the Ethereum block header.
```
axic, line 26: Is that really all what needs to be in the spec? (I'm not sure)

axic:
Missing the `Security Considerations` section.

EIPS/eip-2357.md
MicahZoltu: I suspect the spec will need something like:
> on and after block FORK_BLOCK_NUMBER, the block header will include `totalDifficulty` which is a uint256 value that represents the sum of difficulties from of all blocks since genesis and `difficulty` field will no longer be present in the block header.

I'm just guessing though... I'm assuming by "the spec" you are referring to the Yellow Paper?

tkstanczak: @axic I understand that you point at implications in header validation and so on - I think this is worth discussing. Header validation requires parent anyway but the seal validation currently does not and it will after thic change. The implication is that invalid seal blocks may invoke one header lookup in the DB before the block is rebroadcasted but I think that all clients do a full header validation before broadcasting.

tkstanczak: The moment the parent is loaded the difficulty field can be calculated as the difference between parent and child. Worth reviewing if that works for AuRa chains but I think this will be ok. It does work fine for Clique.

MicahZoltu: It may be worth finding out from @AlexeyAkhunov whether this change would work for TurboGeth.  The only reason I call it out specifically is because I know their DB structure is *significantly* different from other clients so anything that requires lookups of previous blocks isn't necessarily the same for them.  I don't *think* it matters in this case because their differences are primarily around the state DB, not the block headers...

axic: It is also a once-off look up at the transition block, otherwise it is just an increment on the previous `block.total_difficulty`.

tkstanczak: @axic there is an issue now that when a NewBlock message is sent over Eth62 it contains a TotalDifficulty field which is currently ignored by the clients because you cannot verify it, also when you exchange StatusMessage in Eth62 you mention total difficulty and then request headers and if someone is on a branch that is unknown to you you would need to ask for the entire branch headers to switch

EIPS/eip-2622.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Abstract
+
+Replace field Difficulty with TotalDifficulty (sum of difficulties since genesis) in the Ethereum block header.
+
+## Motivation
+
+Simplify total difficulty verification in stateless clients (no need to hold all headers from genesis to verify total difficulty claim).
+
+In Eth62 StatusMessage a TotalDifficulty field is included and when nodes are on different branches they need to resync from the branching point to verify such TotalDifficulty claim. A false claim may lead to long range reorg assumption and force clients to try to resync with new entrants many blocks back.
```
axic, line 24: Should link to the Eth62 spec here. Preferably to the `StatusMessage` part.

PR #2629, opened 2020-05-06
ERC-695: Specify behavior in absence of chainId

rekmarks:
This PR specifies that `eth_chainId` should "return a suitable error" if the latest head block "has no chain id".


eip-automerger:

 - EIP 695 requires approval from one of (isaac.ardis@gmail.com, @tcz001, @sorpaas)

PR #2632, opened 2020-05-08
Update eip-721.md

ilanolkies:
Add ERC-721 implementation.


axic:
I think this will be unlikely to be merged, given it is final, but @fulldecent it is up to you.

fulldecent:
Thank you for reaching out. Only corrections and fixes are accepted.

The inclusion of implementations at the time of publication were to show: 

1. The standard was based on existing work at the time of publication; and
2. The standard has interoperability with existing work at the time of publication.

Therefore it is not appropriate for that list to include work after the time of standardization.

PR declined.

---

But if your goal was to reach out to me, you have my attention. You know how to get a hold of me.

PR #2634, opened 2020-05-10
Rename and clarify disconnection errors

rekmarks:
[Link to rendered file with updates.](https://github.com/rekmarks/EIPs/blob/1193-clarify-disconnect-errors/EIPS/eip-1193.md)

The errors related to connectivity were somewhat underspecified, and oddly named. Their names and descriptions have been updated, and a non-normative comment has been added.

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -407,7 +407,7 @@ interface ProviderRpcError extends Error {
 | 4100        | Unauthorized          | The requested method and/or account has not been authorized by the user. |
 | 4200        | Unsupported Method    | The Provider does not support the requested method.                      |
 | 4900        | Disconnected          | The Provider is disconnected from all chains.                            |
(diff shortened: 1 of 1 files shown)


adrianmcli:
Just curious, is this to pave the way for a multi-chain provider? I'm really opening up to that idea.

rekmarks:
@adrianmcli, these particular changes are more about clarifying the distinctions between `4900` and `4901`.

The provider doesn’t have to be “multi-chain” (scare quotes because we haven’t really defined the concept) in order for `4901` to be useful. Imagine a provider exposes a custom RPC method for switching the chain from the dapp. (Not saying that’s a good idea, but just as an example.) If the provider doesn’t recognize a requested chain ID, it can return `4901`.

That being said, as a side-effect, if wallet developers start experimenting with multi-chain providers, there’s now already a standardized way to handle a particular chain not being serviced. :)

PR #2647, opened 2020-05-15
Jumpsub

gcolvin:
Clearer motivation, cleaner opcode specs.


eip-automerger:

 - EIP 1057 is in state Accepted, not Draft or Last Call

PR #2656, opened 2020-05-19
EIP-2315: updated testcases + clarifications

holiman:
This PR contains the updated testcases with the recent changes. Also some clarifications and I re-added cost of `BEGINSUB`. All opcodes should have a defined cost, IMO.

### DIFF
--- a/EIPS/eip-2315.md
+++ b/EIPS/eip-2315.md
@@ -60,10 +60,8 @@ pushes: `0`
 
 **Note 3:** A value popped from `return_stack` _may_ be outside of the code length if the last `JUMPSUB` was the last byte of the `code`. In this case the next opcode is implicitly a `STOP`, which is not an error.
 
-**Note 4:** The description above lays out the _semantics_ of this feature in terms of a `return_stack`.  But the actual state of the `return_stack` is not observable to EVM code or consensus-critical to the protocol.
-> For example, a node implementor may code `JUMPSUB` to unobservably push `pc` on the `return_stack` rather than `pc+1`, which is allowed so long as `RETURNSUB` observably returns control to the `pc+1` location.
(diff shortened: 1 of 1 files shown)


EIPS/eip-2315.md
```diff
@@ -149,11 +147,23 @@ Bytecode: `0x600556b2b75b6003b3`
...
+
+In this example, the code 'walks' into a subroutine, which is not allowed, and causes an error
+
+|  Pc   |      Op     | Cost |   Stack   |   RStack  |
+|-------|-------------|------|-----------|-----------|
+|    0  |   BEGINSUB  |    1 |        [] |        [] |
+
+
+```
+Error: at pc=0, op=BEGINSUB: invalid subroutine entry
```
matkt, line 165: Do we want an OOG message or another like `invalid subroutine entry` ? 
If we want to use this new message maybe we need to update this part https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2315.md#beginsub ?

holiman:
The actual error message is not part of consensus

matkt:
> The actual error message is not part of consensus

yes indeed it's just to have a consistency :)

holiman:
Well, I think saying "it goes OOG", means that it will "consensus-wise" behave as an OOG, which is a well-understood failure mode at this time. Whether the human-readable strings that a client spits out says that or something else is not something that should be defined in the EIP, imo.

If we want more consistency, it would be more suited for another eip (not `core`), to standardize stuff that are non-consensus-critical. 

But I guess it's un-intuitive that the EIP says OOG error but the EIP-examples says something else. SO may be I should clarify in the EIP that the testcases/examples are based on geth and the message itself is an implementation detail.

PR #2046, opened 2019-05-18
Reduced gas cost for static calls made to precompiles

axic:

PR #2666, opened 2020-05-22
Precompiles and Keccak256 repricing

shamatar:
Main points:
- Reprice sha256 and Ripemd160 precompiles after #2046 changes and to better reflect current performance and internal structures of the computations performed
- Reprice Keccak256 native function and reflect internal structures of the computations performed

Edits:
- BNADD and BNMUL are no longer required to reprice after faster inversion is implemented in Geth
- Error handling is out of scope for now


axic:
Didn't we agree on ACD#84 to split these: https://github.com/ethereum/pm/issues/162#issuecomment-608471710 ?

Also this seems to conflate even more EIPs into one, including [EIP-1352](https://eips.ethereum.org/EIPS/eip-1352) which has been discussed separately.

shamatar:
Here proposal is much more radical than EIP-2046: `STATICCALL` price is set to zero when calling a precompile.

I can set EIP-1352 as a requirement instead instead of talking about it explicitly.

Otherwise it's all merged into one EIP to reflect that all these changes should (at least in my mind) happen simultaneously to have logical correctness.

axic:
It was also mentioned multiple times that the figure in 2046 is not final and should be adjusted based on benchmarking.

shamatar:
Do you mind to update it to zero and I'll add it as a requirement instead of replacement?

axic:
I can update stuff next week (trying to wind down now).

I think it could help to have perhaps a semi-ad-hoc chat early next week, with some others interested in this topic, to get aligned, as it seems the ACD format doesn't really facilitates a good discussion. I'm sure it would be possible to resolve these small conflicts.

shamatar:
Error handling was removed for now

EIPS/eip-2666.md
```diff
@@ -0,0 +1,136 @@
...
+created: 2020-05-22
+requires: 1352, 2046, 2565
+---
+
+## Simple Summary
+This EIP tries to fix the problems with Ethereum precompiles and built-in EVM function:
+- EIP-2046 changes a `STATICCALL` cost to precompile and it may be necessary to adjust costs of some precompiles that have taken old large cost (`700` gas) into account 
+- Some precompiles are overpriced and their pricing formulas to not reflect the structure of underlying function
+- Keccak256 built-in function in EVM has pricing that does not reflect underlying hash function structure
+- Better specify behavior of how gas handled if the precompile call fails (DDoS handling behavior)
```
MicahZoltu, line 18:
```suggestion
Change gas price of `BNADD`, `BNMUL`, `SHA256`, `RIPEMD` and `Keccak256`.
```
Simple summary should be very terse, one or two sentences at most, and should cut straight to the heart of this specification with no dilly dallying on the "why" or the "how".

EIPS/eip-2666.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Abstract
+Costs of many precompiles and built-in functions are invalid at the current state of the clients. This EIP contains a list of changes to the pricing formulas to better reflect underlying computations' structure.
```
MicahZoltu, line 21: Consider including the list of precompiles/built-ins affected and their new gas prices.  Abstract should ideally be a human friendly version of the specification, and in this case IIUC the specification is just a price change for a few things.

EIPS/eip-2666.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Motivation
+Motivation is simple: make pricing formulas accurately reflect resources (CPU time) requires to actually perform the called computations.
```
MicahZoltu, line 24:
```suggestion
Historic pricing for these functions is not in line with their actual costs relative to other operations in the EVM.
```
Current motivation is fine if you want to keep it, but I recommend using a less conversational style and being a bit more to the point.

EIPS/eip-2666.md
```diff
@@ -0,0 +1,136 @@
...
+| Besu  |   | 19  | 28  |
+| Nethermind  |   | 16  | 3  |
+
+EIP 2666 proposes `A = 15`, `B = 13`. There are no large one-off costs in this precompile, so it's EIP 2046 - safe. Besu expects to have performance improvements by the end of the year.
+
+### Inclusion plan
+
+EIP 2046 can not be accepted without at least part of the repricings from this EIP (namely for `BNADD` and `BNMUL`). Even though work necessary to activate this EIP is very small (adjust few constants), one can also consider two step plan that can be implemented as the following:
+- Include EIP 2046 and `BNADD`/`BNMUL` repricings
+- Include repricings for `SHA256`, `RIPEMD` and `Keccak256`
```
MicahZoltu, line 110:
```suggestion
As of FORK_BLOCK_NUMBER, `SHA256` precompile (address `0x02`) will costs `10 + ((len(input) + 8)/64 + 1) * 9` gas.
As of FORK_BLOCK_NUMBER, `RIPEMD` precompile (address `0x03`) will cost `6 + ((len(input) + 8)/64 + 1) * 12` gas.
As of FORK_BLOCK_NUMBER, `BNADD` precompile (address `0x06`) will cost `350` gas.
As of FORK_BLOCK_NUMBER, `BNMUL` precompile (address `0x07`) will cost `6300` gas.
As of FORK_BLOCK_NUMBER,  `SHA3` (`0x20`) will cost `13 + (len(input)/136 + 1)*15` gas.
```
Specification section should focus entirely on the specification, and not discuss rationale, motivation, inclusion process, etc.  Someone trying to build a client should be able to read the specification and get exactly what they need without having to dig through a bunch of other information.

A lot of the information included in this section is valuable, but it should be moved to the Rationale section, which describes *why* you chose the values you did.  The specification section just asserts the values.  As far as the historic values, this isn't particularly relevant to the specification itself so I would include them in either the motivation section or just remove that data all together (someone who wants that information can find it if they need to by looking at old specifications or the yellow paper.

MicahZoltu (approved):
I recommend making the suggested changes (or something similar) prior to Last Call, but this looks well formatted enough to merge as a draft.

MicahZoltu:
Note: I agree with @axic that this EIP should really be split up into one EIP per repricing so they can be easily discussed independently, rather than people having to agree on all or nothing.  We don't currently have a hard rule against monolithic EIPs though so I'll stick with my approval for now, but I'll wait to see if the author would prefer to split this up before merging.

@shamatar If you want to merge this without splitting it up, let me know and I'll click the button.

shamatar:
More EIPs is more scary than less EIPs, so I'll keep it as it is for now. Let me refactor with your suggestions in mind and it will be ready to merge. 

Since I'll be pushing further for this EIP and 2046 it may be good to return back to having this one supersede the 2046.

MicahZoltu:
> More EIPs is more scary than less EIPs

It should be the opposite.  You are much more likely to see things go through with more EIPs because people cannot hamstring the entire set of changes by only complaining about one thing.  e.g., someone has some problem with the way you calculated SHA256 costs, so the entire EIP gets stuck while you debate about the merits of your strategy vs theirs.  If they were separate EIPs then such a debate would only block SHA256, but not all of the other changes.

shamatar:
@MicahZoltu Last two commits should make it much more readable and simple

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+created: 2020-05-22
+requires: 1352, 2046, 2565
+---
+
+## Simple Summary
+This EIP tries to fix the problems with Ethereum precompiles and built-in EVM function:
+- EIP-2046 changes a `STATICCALL` cost to precompile and it may be necessary to adjust costs of some precompiles that have taken old large cost (`700` gas) into account 
+- Some precompiles are overpriced and their pricing formulas do not reflect the structure of underlying function
+- Keccak256 built-in function in EVM has pricing that does not reflect underlying hash function structure
+- Better specify behavior of how gas handled if the precompile call fails (DDoS handling behavior)
```
MicahZoltu, line 18: Simple summary is akin to an email subject line, forum title, or GitHub PR title.  Just a one sentence description that can be used in conversation like, "EIP 2666... you know, the one that ..."

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Abstract
+Costs of many precompiles and built-in functions are invalid at the current state of the clients. This EIP contains a list of changes to the pricing formulas to better reflect underlying computations' structure.
+
+## Motivation
+Historic pricing for these functions in EVM does not reflect inner structure of the underlying computations (inner structure of the hash functions).
+
+## Specification
+
+If `block_number >= X` set the following prices to precompiles and Keccak256 opcode:
```
MicahZoltu, line 28:
```suggestion
If `block_number >= X` set the gas cost of the following precompiles and Keccak256 opcode:
```

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+  - `SHA256` precompile (address `0x02`) should be priced as `10 + ((len(input) + 8)/64 + 1) * 9`
+  - RIPEMD precompile (address `0x03`) should be priced as `6 + ((len(input) + 8)/64 + 1) * 12`
+  - Keccak256 built-in function should be priced as `13 + (len(input)/136 + 1)*15`
```
MicahZoltu, line 31:
```suggestion
  - `SHA256` precompile (address `0x02`) **MUST** have a gas cost of `10 + ((len(input) + 8)/64 + 1) * 9`
  - RIPEMD precompile (address `0x03`) **MUST** have a gas cost of `6 + ((len(input) + 8)/64 + 1) * 12`
  - Keccak256 built-in function **MUST** have a gas cost of `13 + (len(input)/136 + 1)*15`
```

Alternatively, consider formatting this as a list of pairs:
```suggestion
  - SHA256 (precompile 0x02): `10 + ((len(input) + 8)/64 + 1) * 9`
  - RIPEMD (precompile 0x03): `6 + ((len(input) + 8)/64 + 1) * 12`
  - KECCAK256 (0x20): `13 + (len(input)/136 + 1)*15`
```

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+
+This EIP *ideally* requires that `MODEXP` repricing is [implemented](https://eips.ethereum.org/EIPS/eip-2565) to also accurately reflect that there is implicit compensation for an old `STATICCALL` cost (pre-2046).
```
MicahZoltu, line 33:
```suggestion
This EIP *ideally* requires that `MODEXP` repricing is [implemented](./eip-2565.md) to also accurately reflect that there is implicit compensation for an old `STATICCALL` cost (pre-2046).
```

MicahZoltu: Also, all opcodes must be in the form `STATICCALL (0xfa)` the first time they are referenced in the EIP at least (optional for further references).  I think we should do something similar for precompiles as well.

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+
+|   |   | A  | B  |
+|---|---|---|---|
+| Geth  |   | 13  | 13  |
+| OE  |   | 15  | 2  |
+| Besu  |   | 19  | 28  |
+| Nethermind  |   | 16  | 3  |
+
+EIP-2666 proposes `A = 15`, `B = 13`. There are no large one-off costs in this precompile, so it's EIP-2046 - safe. Besu expects to have performance improvements by the end of the year.
+
```
MicahZoltu, line 83: All of this should go in the Rationale.  The Specification should *only* contain the specification, not the rationale behind *why* particular values were chosen.  This content is all quite appropriate for Rationale section though.

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+There are no explicit test cases.
+
+## Implementation
+
+Reference material with raw data can be found [here](https://docs.google.com/spreadsheets/d/1aCQnk7prrp3Mbcf011BE5zZnkbc3Iw7QAixn6mLbKS0/edit?usp=sharing).
+
+There is a repository available with inputs for benchmarking and precompiles testing [here](https://github.com/shamatar/bench_precompiles).
+
+Raw Besu [benchmarks](https://gist.github.com/shemnon/0ddba91be501fa23291bdec9107fe99a).
+
```
MicahZoltu, line 101: The Implementation section is optional, and should ideally contain either an inline reference implementation or a link to something in `../assets/eip-2666/`.  For tutorials, examples, and other supplementary material a blog article or forum thread is generally a better place for that kind of stuff.

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+## Rationale
+Cost of functions being executed must accurately reflect real CPU time spent on computations, so benchmarking was performed for current precompiles and Keccak256 function to measure running time versus input parameters. 
+
+## Backwards Compatibility
+Precompile repricings has happened in a past and can be considered standard procedure. Gas costs of many contracts is expected to reduce that may break re-entrancy protection measures based on fixed gas costs. In any case, such protection should have never been considered good and final.
+
+## Test Cases
+
+There are no explicit test cases.
```
MicahZoltu, line 92: Not an immediate issue, but it feels like there should be at least a handful of example calls and their expected gas costs so that clients can do some basic consensus testing prior to integration testing.

MicahZoltu (changes requested):
The only thing that must be fixed prior to merging as a draft is that all opcodes must include their numeric code in parenthesis after as described here: https://eips.ethereum.org/EIPS/eip-1#special-requirements-for-core-eips.  All of the other feedback should probably be addressed/considered prior to moving this to Review, but isn't a blocker for Draft.

EIPS/eip-2666.md
```diff
@@ -0,0 +1,107 @@
...
+  - RIPEMD precompile (address `0x03`) should be priced as `6 + ((len(input) + 8)/64 + 1) * 12`
+  - Keccak256 built-in function should be priced as `13 + (len(input)/136 + 1)*15`
+
+This EIP *ideally* requires that `MODEXP` repricing is [implemented](https://eips.ethereum.org/EIPS/eip-2565) to also accurately reflect that there is implicit compensation for an old `STATICCALL` cost (pre-2046).
+
+### Detailed summary of repricing approach
+
+This EIP relies on two facts:
+- apriori knowledge of the inner strucute of the hash functions
+- benchmarks provided by the client teams for some reasoable range of input lengths for random inputs (random byte strings of a given length)
```
MicahZoltu, line 39:
```suggestion
- benchmarks provided by the client teams for some reasonable range of input lengths for random inputs (random byte strings of a given length)
```
CI won't let this through with spelling errors.  😬

shamatar:
Should be in better shape now

PR #2667, opened 2020-05-23
Update eip-2458.md

edsonayllon:
Updated spec to match other headers, as suggested by Axic.


edsonayllon:
@axic any clue as to why the `codespell` check is failing?

PR #2672, opened 2020-05-26
695: Errata: remove EIP-1474 from dependencies

rekmarks:
EIP-695 still listed EIP-1474 as a dependency when it was finalized. This was an oversight on the part of the authors. 695 never required 1474. Rather, 1474 could be said to require 695. I'm author of both, and 1474 is still in draft.

Apologies to @axic and @Arachnid for missing this.

### DIFF
--- a/EIPS/eip-695.md
+++ b/EIPS/eip-695.md
@@ -7,7 +7,7 @@ type: Standards Track
 category: Interface
 status: Final
 created: 2017-08-21
-requires: 155, 1474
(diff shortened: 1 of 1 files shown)


rekmarks:
@Arachnid @axic this is rebased on latest master and ready to go.

axic:
While I wish 1474 would be final and we could reference it for the core definitions, it would be wrong to keep a final EIP referencing a draft.

PR #2673, opened 2020-05-27
Update interfaces, clarify type intentions

rekmarks:
Link to file: https://github.com/rekmarks/EIPs/blob/1193-update-interfaces/EIPS/eip-1193.md

### Summary of Changes

- Remove index signatures from interfaces
  - i.e. no more `[key: string]: unknown`; the ability to add properties is implied
- `RequestArguments`
  - `params?: unknown` -> `params?: unknown[] | object`
  - Explicitly define relationship between `RequestArguments` fields and RPC method names and parameters
    - You could say that relationship was heavily implied, but 🤷 
- Remove redundant **MAY** statements.
- Various minor fixups and clarifications
  - Clarify author intentions in certain places

### A Note on `request`

After review, here's where I currently stand on the signature of `request`.
- `request` takes a single, options bag argument, to make it as extensible as possible. We are extremely unlikely to have to revisit it ever again. This is a wonderful thing.
  - If devs find the options bag unergonomic, they can define their own wrapper functions. We, however, should not bloat the spec with syntactic sugar.
- Provider implementers should type `RequestArguments` and its `params` property however it makes practical sense to do so. Presumably, they'll type it such that it supports whatever RPC methods the implementation supports.
  - _That said_, I have not found an example of an RPC protocol that _does not_ take its parameters by name (object) or position (array). As such, we set the type to `params?: unknown[] | object`.
  - This is not normative for future RPC specifications. This only normative for how those methods are called via this API. If RPC standards emerge that map poorly to this concept, that's IMO an acceptable use case for errata.
- I recommend that implementers implement `request` in this way:
  - Do _basic_ input validation (i.e. `typeof args?.method === 'string'`).
  - Try to find a method handler.
  - If no handler, return error. If handler, hand off to handler; return result or error from handler.
    - Do detailed type checking of `params` in handler.

--------------------------

Ref #2319

cc: @alcuadrado @ryanio @MicahZoltu

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -30,7 +30,7 @@ Makes an Ethereum RPC method call.
 ```typescript
 interface RequestArguments {
   method: string;
-  params?: unknown;
+  params?: unknown[] | object;
 }
 
 Provider.request(args: RequestArguments): Promise<unknown>;
@@ -317,7 +317,7 @@ The Provider **MUST** implement and expose the API defined in this section. All 
 
 The Provider **MAY** expose methods and properties not specified in this section.
 
-The Provider **MAY** extend TypeScript `interface` declarations to include fields not listed in this specification.
+The Provider **MAY** extend TypeScript `interface` declarations to include properties not listed in this specification.
 
 The Provider **MAY** replace references to `unknown` with more specific types, if the possible values can be known at compile time.
 
@@ -334,7 +334,7 @@ The Provider **MAY** replace references to `unknown` with more specific types, i
 ```typescript
 interface RequestArguments {
   method: string;
-  params?: unknown;
+  params?: unknown[] | object;
 }
 
 Provider.request(args: RequestArguments): Promise<unknown>;
@@ -376,7 +376,13 @@ All supported RPC methods **MUST** be identified by unique strings.
 
 Providers **MAY** support whatever RPC methods required to fulfill their purpose, standardized or otherwise.
 
-If a Provider supports a method defined in a finalized EIP, the Provider's implementation of this method **MUST** adhere to the method's specification.
+If a Provider supports a method defined in a finalized EIP, the Provider's implementation of this method **MUST** adhere to the method's specification. This includes:
+
+- The method name, given as the string value of `RequestArguments.method`
+- The method parameters, given as the value of `RequestArguments.params`, either:
+  - By position, in an array
+  - By name, in an object
+- Any additional properties on `RequestArguments`, as necessary
 
 If an RPC method defined in a finalized EIP is not supported, it **SHOULD** be rejected with a `4200` error per the [Provider Errors](#provider-errors) section below, or an appropriate error per the RPC method's specification.
 
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -311,15 +311,10 @@ The Provider is said to be "disconnected" when it cannot service RPC requests to
 
 ### API
 
-The Provider API is specified using TypeScript.
+> The Provider API is specified using TypeScript.
+> The authors encourage implementers to declare their own types and interfaces, using the ones in this section as a basis.
 
 The Provider **MUST** implement and expose the API defined in this section. All API entities **MUST** adhere to the types and interfaces defined in this section.
-
-> The TypeScript `interface` declarations are designed to be minimal and extensible.
-> The authors encourage implementers to make syntactically valid extensions of interfaces when necessary.
->
-> Furthermore, an `unknown` type indicates that the possible values depend on information external to this specification.
-> The authors encourage implementers to replace instances of `unknown` with more specific types in their own type definitions, if the possible values can be known at compile time.
 
 #### request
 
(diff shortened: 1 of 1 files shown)


rekmarks:
@ryanio after further discussion, I expanded these changes, and changed the type of `RequestArguments.params`.

EIPS/eip-1193.md
```diff
@@ -312,9 +311,21 @@ The Provider is said to be "disconnected" when it cannot service RPC requests to
...
 ### API
 
-The Provider **MUST** expose the API defined in this section. All API entities **MUST** adhere to the types and interfaces defined in this section.
+The Provider API is specified using TypeScript.
+
+The Provider **MUST** implement and expose the API defined in this section. All API entities **MUST** adhere to the types and interfaces defined in this section.
+
+The Provider **MAY** expose methods and properties not specified in this section.
+
+The Provider **MAY** extend TypeScript `interface` declarations to include properties not listed in this specification.
```
MicahZoltu, line 320: I don't think this is necessary for a specification.  Since TypeScript types are erased at runtime, whether the provider does this or not is totally transparent to the app so this is just needless noise in the specification.

EIPS/eip-1193.md
```diff
@@ -312,9 +311,21 @@ The Provider is said to be "disconnected" when it cannot service RPC requests to
...
+
+The Provider **MAY** replace references to `unknown` with more specific types, if the possible values can be known at compile time.
```
MicahZoltu, line 322: Same comment as above, type erasure results in this being an empty statement.

EIPS/eip-1193.md
```diff
@@ -366,7 +376,13 @@ All supported RPC methods **MUST** be identified by unique strings.
 
 Providers **MAY** support whatever RPC methods required to fulfill their purpose, standardized or otherwise.
 
-If a Provider supports a method defined in a finalized EIP, the Provider's implementation of this method **MUST** adhere to the method's specification.
+If a Provider supports a method defined in a finalized EIP, the Provider's implementation of this method **MUST** adhere to the method's specification. This includes:
```
MicahZoltu, line 379: This statement feels cyclical?  I read it as "If a provider supports X, then a provider must support X".  Perhaps I'm misunderstanding something?

EIPS/eip-1193.md
```diff
@@ -366,7 +376,13 @@ All supported RPC methods **MUST** be identified by unique strings.
...
+
+- The method name, given as the string value of `RequestArguments.method`
+- The method parameters, given as the value of `RequestArguments.params`, either:
+  - By position, in an array
+  - By name, in an object
```
MicahZoltu, line 384: This feels like it is redundant with the initial specification.  Earlier in the specification we have already asserted that `method`'s type is `string` and `params` type is `unknown[] | object | undefined`.

EIPS/eip-1193.md
```diff
@@ -366,7 +376,13 @@ All supported RPC methods **MUST** be identified by unique strings.
...
+- Any additional properties on `RequestArguments`, as necessary
```
MicahZoltu, line 385: I don't think it is necessary for this to be part of the specification.  The specification only needs to describe the set of things required for the two systems to communicate, it doesn't need to describe all possible things that one may wish to have.  For example, an HTTP specification may state that certain headers are required, but it needn't specify that all other headers are _allowed_ (though not relevant to the specification).

EIPS/eip-1193.md
rekmarks: It is a bit cyclical, although I go through the trouble of explicitly defining what we mean by "supported RPC method" above:
> A "supported RPC method" is any RPC method that may be called via the Provider.

If you have any ideas for how to express it in a less cyclical fashion, I'm all ears!

MicahZoltu: My vote is to just remove it, as I don't understand the value it adds to the specification.

EIPS/eip-1193.md
rekmarks: Very good. I removed all references to types from these lines, and just explicitly state how the RPC method specification relates to the fields of the `RequestArguments` interface.

Unless I'm missing something, that was previously only implied, and I'd rather it be explicit.

rekmarks: I removed this from the `Supported RPC Methods` section and added a more concise version to the `request` section, which has been made more concise overall.

alcuadrado:
These changes look good. Thanks @rekmarks.

rekmarks:
FYI: I'm not making any changes to this PR anymore, I'm just force-pushing to cause CI to rerun.

frozeman:
Thanks for that @rekmarks!
I would though like to see at least one example in written out JS as I know how much easier it will make it to impotenteres and how it will prevent misunderstandings. 99% developers don’t read the spec they simply copy the example abs change it up.

rekmarks:
MetaMask has a working implementation here: https://github.com/MetaMask/inpage-provider/blob/master/src/MetamaskInpageProvider.js

frozeman:
I mean a one liner:

`ethereum.request({method: 'eth_exampleRequest', params: ['param1']});`

Not a full implementation. We spent need to go to one companies implantation for reference, as this could be biased. (Don’t get me wrong, I like MM)

rekmarks:
Ah, got it! Haha, our implementation is definitely biased.

There are a number of examples in the `Examples` section already, see: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#examples

frozeman:
Oh yes,
I forgot about those. Sorry I am on my phone.

PR #2677, opened 2020-05-28
EIP-2677: Cap initcode (draft)

holiman:
Propose a cap on initcode. 

:shipit:


EIPS/eip-draft_cap_initcode.md
```diff
@@ -0,0 +1,70 @@
...
+author: Martin Holst Swende (@holiman), Pawel Bylica (@chfast), Alex Beregszazi (@axic)
+discussions-to: <URL>
+status: Draft
+type: Core
+created: 2020-05-18
+---
+
+## Simple Summary
+
+Enforce a maximum size limit (`max_initcode_size`) of `49152` (`0x6000`) for `initcode`.
```
chfast, line 13:
```suggestion
Enforce a maximum size limit (`max_initcode_size`) of `49152` (`0xc000`) for `initcode`.
```

s1na, line 13:
```suggestion
Enforce a maximum size limit (`max_initcode_size`) of `49152` (`0xc000`) for `initcode`.
```

holiman: Thanks!

holiman: Thanks! @s1na was faster :)

axic:
@holiman merged in master and added a discussion URL so this can be merged.

axic (approved):

EIPS/eip-2677.md
```diff
@@ -0,0 +1,76 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-2677-limit-size-of-initcode/4550
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-05-18
+---
+
+## Simple Summary
+
+Enforce a maximum size limit (`max_initcode_size`) of `49152` (`0xc000`) for `initcode`.
```
MicahZoltu, line 14:
```suggestion
Set a maximum size limit of 49152 for `initcode`.
```

EIPS/eip-2677.md
```diff
@@ -0,0 +1,76 @@
...
+## Abstract
+
+Enforce a maximum size limit (`max_initcode_size`) for `initcode`. If the size of `initcode` exceeds `max_initcode_size`, then contract creation fails with an out of gas error.
+
+Since [EIP-170](https://eips.ethereum.org/EIPS/eip-170) was implemented, there has been a size limit of `24576` (`0x6000`) on contract code. We propose to also limit the size of executable code to `2x` the above limit, i.e. `49152` (`0xc000`).
+
+This also leads to two nice properties:
+
+- instruction offset in code fits 16-bit value,
+- code size fits 16-bit value.
```
MicahZoltu, line 25: Consider moving all of this to the motivation section.  Abstract should be a human-friendly and terse version of the specification, while the motivation section is where you state why this is a good idea, how these changes make things better, etc.

EIPS/eip-2677.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Specification
+
+There are three situations where this is applicable:
+
+* `CREATE`,
+* `CREATE2`,
+* creation using a transaction with empty receiver. 
+
+In all these (and future) cases, the EVM should fail with Out Of Gas error if the code has a length more than `max_initcode_size`.
```
MicahZoltu, line 48: A user should be able to implement an EIP from the specification section alone.  Currently, this specification section is not self contained as it appears to be referring to information in other sections (e.g., the `this` reference in the first sentence).

Also it should make assertions, not suggestions, **especially** for consensus code.  "should fail" will result in total consensus failure if some clients fail and some clients don't fail.  My guess is that you meant **MUST** fail here.

Also, the specification section should not be discussing hypothetical future cases here.  It may be worth mentioning in the security considerations or backward compatibility section, but discussion about the behavior of future hypothetical cases are not really relevant for the specification section of a technical document.

MicahZoltu (approved):
Good enough to merge as a draft, but I wouldn't be willing to merge this to Last Call as it is (even ignoring the incomplete sections).  I recommend reviewing the feedback provided and making adjustments before submitting to last call.

Also consider adjusting the tense of the writing here.  I very much prefer to write EIPs as technical documents that target a reader base of future specification readers, rather than current/past readers.  This means instead of saying "there has been a size limit of ..." you might say "prior to this EIP there was a size limit of ..." (as a simple example).

EIPs are assertions, not proposals.  This means the EIP shouldn't say stuff like "we propose you do X", it should instead say "do X".  whether someone *follows* the EIP or not is a totally separate conversation that should not be part of the EIP itself, and once someone has decided to follow the EIP everything in it should be an assertion unless it is intentionally left ambiguous (e.g., **SHOULD** or **MAY**).

PR #2678, opened 2020-05-28
EthPM v3 EIP

njgheorghita:

 - The build passes.


EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+number of areas of improvement found for the previous version (defined
+in
+[EIP-1123](https://eips.ethereum.org/EIPS/eip-1123)).
+This version:
+
+
+-   Updates the schema for a *package manifest* to be compatible with
+	the [metadata](https://solidity.readthedocs.io/en/latest/metadata.html) output for compilers.
+-   Updates the `"sources"` object definition to support a wider range of source file types and serve as [JSON input](https://solidity.readthedocs.io/en/latest/using-the-compiler.html#compiler-input-and-output-json-description) for a compiler.
+-   Moves compiler definitions to a top-level `"compilers"` array to:
```
fubuloubu, line 53: weird sentence, I think this?
```suggestion
-   Moves compiler definitions to a top-level `"compilers"` array, in order to:
```

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+	-   Simplify the links between a compiler version, sources, and its
```
fubuloubu, line 54:
```suggestion
	-   Simplify the links between a compiler version, sources, and the
```

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+    package management software as part of the release process.
+
+2.  Package manifests will be consumed by package managers during tasks
+    like installing package dependencies or building and deploying new
+    releases.
+
+3.  Package manifests will typically **not** be stored alongside the
+    source, but rather by package registries *or* referenced by package
+    registries and stored in something akin to IPFS.
+
```
fubuloubu, line 93: Maybe add another item:
"Package manifests can be used to verify public deployments of source contracts" a la Etherscan verification

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+<td><p><code>deployments</code></p></td>
+</tr>
+<tr class="odd">
+<td><p>Type</p></td>
+<td><p>Object (String: Object(String: <a href="#contract-instance-object">Contract Instance Object</a>))</p></td>
+</tr>
+<tr class="even">
+<td><p>Format</p></td>
+<td><p>Keys <strong>must</strong> be a valid BIP122 URI chain definition.</p>
+<p>Values <strong>must</strong> be objects which conform to the following format.</p>
```
fubuloubu, line 456:
```suggestion
<p>Values <strong>must</strong> be objects which conform to the following format:</p>
```

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+#### Install Path: `installPath`
+
+Filesystem path of source file.  
+-   **Must** be a relative filesystem path that begins with a `./`.
+
+-   **Must** resolve to a path that is within the current virtual
+    working directory.
+
+-   **Must** be unique across all included sources.
+
```
fubuloubu, line 1135: Cannot contain `'../`, so as to avoid accessing files outside of the source folder in improper implementations

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+<td><p>Required</p></td>
+<td><p>Yes</p></td>
+</tr>
+<tr class="even">
+<td><p>Type</p></td>
+<td><p>String</p></td>
+</tr>
+</tbody>
+</table>
+
```
fubuloubu, line 1241: Mention possible algorithms (`sha3`, `sha256`, `md5`, etc.)

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+</tr>
+<tr class="odd">
+<td><p>Format</p></td>
+<td><p><code>0x</code> prefixed hex encoded block hash.</p></td>
+</tr>
+</tbody>
+</table>
+
+
+#### Runtime Bytecode: `runtimeBytecode`
```
fubuloubu, line 1604: What is the use case for this? Could this be replaced with `deploymentArguments` (ABI-encoded arguments used during deployment transaction)

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+<td><p>Object</p></td>
+</tr>
+</tbody>
+</table>
+
+
+#### Contract Types: `contractTypes`
+
+A list of the [Contract Alias](#term-contract-alias) or [Contract Types](#term-contract-type) in this package
+that used this compiler to generate its outputs.
```
fubuloubu, line 1739: Why is this done via `CompilerInformation` -> `ContractType`, instead of the other way around? Would make more sense if `ContractType` linked to `CompilerInformation` object instead, ensuring no duplication

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+Work](#further-work), specified below.
+
+
+Further Work
+============
+
+This EIP addresses only the data format for package descriptions.
+Excluded from the scope of this specification are:
+
+-   Package registry interface definition (ERC 1319)
```
fubuloubu, line 2110: Link to ERC 1319

EIPS/eip-2678.md
```diff
@@ -0,0 +1,2132 @@
...
+This version:
+
+
+-   Updates the schema for a *package manifest* to be compatible with
+	the [metadata](https://solidity.readthedocs.io/en/latest/metadata.html) output for compilers.
+-   Updates the `"sources"` object definition to support a wider range of source file types and serve as [JSON input](https://solidity.readthedocs.io/en/latest/using-the-compiler.html#compiler-input-and-output-json-description) for a compiler.
+-   Moves compiler definitions to a top-level `"compilers"` array to:
+	-   Simplify the links between a compiler version, sources, and its
+		compiled assets.
+	-   Simplify packages that contain multiple compiler versions.
```
chriseth, line 56:
```suggestion
	-   Simplify packages that use multiple compiler versions.
```

EIPS/eip-2678.md
njgheorghita: Yeah this makes sense, though in the spirit of not enforcing unnecessary requirements, is it necessary?

1. Just a couple of lines above we say that every path must resolve to the virtual directory (aka source folder).

2. A path like `contracts/wallet/../safe-math-lib/SafeMathLib.sol` would be dumb but valid.

We could word it like `Cannot start with "../" ..` but I'm not sure it's necessary to explicitly state this restriction given point #1. Thoughts?

EIPS/eip-2678.md
njgheorghita: Probably, but given that `Compilers` is an array of `CompilerInformationObject`s, I'm not sure of any reliable method to link `ContractType` -> `CompilerInformation`.

EIPS/eip-2678.md
fubuloubu: I think requiring that it "only access inside it's own directory" is important, but you can probably relax not having `../` to be a simpler check if it's all self-containted

EIPS/eip-2678.md
fubuloubu: By index? Both should be handled by a framework, so I see no issue in them being able to rectify that

EIPS/eip-2678.md
njgheorghita: I believe it's also to support a deployment that references any combination of linked libraries (via `runtimeBytecode`).

EIPS/eip-2678.md
njgheorghita: I feel somewhat uneasy about relying on the index. It doesn't seem like a reliable strategy to maintain the links, and it's not as explicit as the current strategy, which while clunky in terms of enforcing no duplication, it's more obvious how the links work without a thorough knowledge of the spec.

EIPS/eip-2678.md
njgheorghita: Just to be clear, imo `**Must** resolve to a path that is within the current virtual working directory.` serves this restriction. Do you interpret this differently?

fubuloubu: Must have missed that

MicahZoltu:
@njgheorghita This EIP doesn't follow the [Template](https://github.com/ethereum/EIPs/blob/master/eip-template.md) formatting.  Can you update it to match the formatting found there?  The most noticeable difference is the section headings, which should match the formatting found in the Template (leading `##` before line).  I didn't look too closely, so there may be other differences that need to be addressed.

njgheorghita:
@MicahZoltu Thanks for the heads up on the formatting! I've updated the pr to match the formatting now.

PR #2679, opened 2020-05-29
Upgrade to latest github-pages and enable Solidity syntax highlighting

axic:
After https://github.com/github/pages-gem/pull/683 / https://github.com/github/pages-gem/pull/690 was merged this is possible now.

This allows EIPs to use solidity syntax highlighting. Currently there are 330 instances of it:

```
$ grep "\`solidity" * | wc -l
     330
```

And some EIPs used `typescript` or `javascript` as an alternative, because that was highlighted already:
```
$ grep "\`typescript" * | wc -l
      16
$ grep "\`javascript" * | wc -l
      25
```

Thanks @veox again in adding highlighting to rouge.


eip-automerger:

 - File .travis.yml is not an EIP
 - File Gemfile is not an EIP
 - File Gemfile.lock is not an EIP
 - File _config.yml is not an EIP
 - EIP 2309 requires approval from one of (@pizza-r0b)

axic:
This should be merged after review, cc @Souptacular @gcolvin @nicksavers @Arachnid

MicahZoltu:
@axic This now has conflicts and cannot be merged.  Seeing as how it has been 1.5 months since you opened this and no one has commented, I would propose we just merge it and if things break we can fix it with a follow-up PR.  Worst case scenario if things go south isn't *that* bad, and I think having this sit open for 6 months while we wait for all of the editors to have time to thoroughly review it isn't ideal.

It *would* be nice to get a review from at least one person that understands Ruby/Jekyll just to make sure things are sane, but otherwise I don't think it is critical to wait for total consensus.

> And some EIPs used typescript or javascript as an alternative, because that was highlighted already:

Some of these may be actual TS or JS, so we should take care not to automatically convert all of them to solidity automatically when the time comes.

axic:
> > And some EIPs used typescript or javascript as an alternative, because that was highlighted already:
>
> Some of these may be actual TS or JS, so we should take care not to automatically convert all of them to solidity automatically when the time comes.

I have manually checked and changed the appropriate ones -- did not use any regex to do this.

EIPS/eip-2309.md
```diff
@@ -121,8 +121,6 @@ This extension was written to allow for the smallest change possible to the orig
 
 [Cargo](https://cargo.build)
 
-<!-- ## Implementation
```
axic, line 124: Semi-invalid HTML right here.

PR #2681, opened 2020-05-29
Add EIP for limiting account nonce

axic:


MicahZoltu:
@axic Is there a reason you haven't merged this?  The only requirement I know of for merging a draft PR is that it matches the proper EIP format, which this does.  As an editor, you have merge rights so it is unclear why this is sitting open for a month.  Perhaps you just forgot to merge?

axic:
> Is there a reason you haven't merged this?

As a courtesy editors don't merge their own PRs (with perhaps the exception of simple typo changes).

MicahZoltu:
> As a courtesy editors don't merge their own PRs (with perhaps the exception of simple typo changes).

Ah, this doesn't align with what @gcolvin told me the other day.

I'm personally fine with editors merging their own PRs until we run into a problem with editors merging PRs they shouldn't be (and not correcting their own behavior upon it being mentioned).  Ideally, editors will follow normal rules for soliciting feedback and won't just willy-nilly merge PRs, but I have seen quite a few PRs from editors sitting for months in here because no other editor takes the time to approve/merge.  Given that editors are already generally over-taxed, I think allowing them to self merge is worth the risk at the moment.

PR #2706, opened 2020-06-09
Moves 2696 to last call.

MicahZoltu:


eip-automerger:

 - Trying to change EIP 2696 state from Draft to Last Call

PR #2709, opened 2020-06-10
EIP-2334 small fix: specifically describe deriving multi validation keys 

alonmuroch:
This is a small enhancement for EIP-2334.

EIP2334 should be the recipe for anyone that wants to implement an eth 2.0 compatible wallet, to create interoperability. 
There is one thing it doesn't specify which is a deterministic way of creating validator and withdrawal keys for a multi validator setup. That is, should a wallet implementation create 1 withdrawal key for multiple validation keys? should it be a new pair every time (and if so how to derive them)?

This fix does 2 things
1) makes it easy and simple to restore withdrawal and validation keys from a seed.
2) preserve the user's withdrawal credentials privacy


alonmuroch:
@CarlBeek would love your thoughts on this

eip-automerger:

 - EIP 2334 requires approval from one of (@carlbeek)

alonmuroch:
@MicahZoltu @CarlBeek just synced it with master, please review.

MicahZoltu:
@CarlBeek If you do a PR review in GitHub just choose Approve, Reject, or Comment and we can merge, close, or iterate (respectively).

rauljordan:
Raul from @prysmaticlabs here. This is the derivation format we are using for our eth2 client's built-in wallet. Looks good @alonmuroch

MicahZoltu:
There is talk of a new policy for dealing with abandoned EIPs, but at the moment we don't have any solution for it.  If someone wants to try to reach out to Carl via email or other mechanisms to get him to review this PR that would help move things along.  Alternatively, if Carl can show up and approve a change in authors that would work as well.  Otherwise, someone should create a new EIP that is a copy of this one but with a new author so we can continue to iterate.

alonmuroch:
@MicahZoltu I don't mind creating a new EIP, how should it be named?

MicahZoltu:
@alonmuroch You can title EIPs however you like, it is your EIP.  😄  If you mean how should it be numbered, I recommend using the number of the PR it is introduced in.  My process is usually to put `TBD` for the name initially and submit the PR, then once I have a PR number I go backfill the EIP number as a commit on that branch.

alonmuroch:
@MicahZoltu oh I thought of copying this EIP and "replacing" it. 
We can definitely create another EIP built on that.

CarlBeek:
I disagree with this approach here @alonmuroch. Now you need to track each of the latest indicies for every validator client you wish to run. What happens is a user splits up their validators between phase 0 and 1 when all of a sudden a lot more hardware is required per validator.

alonmuroch:
@CarlBeek not sure I understood... what do you mean by splitting up validators?

CarlBeek:
What I am trying to say here is that IMO the complexity/benefit trade off is not good enough. In the happy case, you have nice separation between VCs, but that quickly degrades when you try to change your validator configuration. If you have 2 VCs with 4 validators each and you wish to rebalance them, then this indexing no longer makes sense. At that point, you need to start tracking 2 indices (`i`, & `n` in your example) for each of your validators. I think this is a brittle configuration in which the indicies will quickly no-longer reflect which keys are where.

axic:
@alonmuroch @CarlBeek can this discussion be taken to the discussion-to-url and this PR be closed? That seems to be the next sensible step to me, given the author (@CarlBeek) disagreed with merging this.

Issue #232, opened 2017-03-21
New transaction formats


vbuterin:
### Specification

When `block.number >= FORK_BLKNUM`, allow two new types of transaction:

    [0, network_id, nonce, gasprice, startgas, to, value, data, v, r, s]
    [1, network_id, startgas, to, data]

The first type of transaction replicates the functionality of existing regular transactions, with the sender equal to `ecrecover(sha3(rlp.encode([0, network_id, nonce, gasprice, startgas, to, value, data])), v, r, s)`, and the second type replicates the functionality of EIP #208 (formerly EIP #86) transactions, where the sender is set to `NULL_SENDER` (a constant equal to 2^160 - 1).

When `block.number >= DEPRECATION_BLKNUM` (set `DEPRECATION_BLKNUM = FORK_BLKNUM + 500000`), disallow the older transaction types.

### Specification, version 2

Have a hard "meta-standard" that ALL present and future transactions will have the following form:

    [version_number, network_id, protocol_specific_params, data]

Under this standard, implement the above as:

    [0, network_id, [nonce, gasprice, startgas, to, value, v, r, s], data]
    [1, network_id, [startgas, to], data]

### Specification, version 3

Add a valid_until field to type 0 transactions (and also type 1 transactions, or not? Should discuss) immediately before the "nonce" argument in type 0 transactions, and immediately before "startgas" in type 1 transactions.

### Rationale

Creates a cleaner and more forward-compatible way of representing transactions, as well as removing technical debt created by EIP #208 and EIP #155 and replacing it with more explicit mechanisms that achieve the same function.

Also, upgrades replay protection with ETC and other alt-ethereum chains by making it compulsory and two-way.

Version 2 has the benefit that it makes a principled distinction between parameters that the protocol cares about and parameters that account code cares about, and makes an explicit 2-layer data structure when with EIP 86 the data structure will have to become 2-layer anyway with nonces, gasprices and the like being part of the transaction data field.

In future transaction versions, arguments could be added such as:

* Subsets of address space that the transaction has read and write permission to (eg. four fields could be added for READSTART, READEND, WRITESTART, WRITEEND, which would accept 0-20 bytes as arguments, and would, with zero-byte-right-padding, represent the minimum and maximum addresses that the tx can read and write to). This would improve parallelizability.
* Information about what shard the transaction is intended for.
[...]

PR #2718, opened 2020-06-13
Creates EIP-2718: Typed Transaction Envelope

MicahZoltu:


vbuterin:
I tried to do something similar before!

https://github.com/ethereum/EIPs/issues/232

Definitely support such initiatives.

luoyhang003:
What is the impact on receipt root hash calculation caused by this EIP?

I found I cannot calculate the correct `receiptsRoot` of block 2358629 (Ethereum Mainnet) because there is a transaction `0xd1a7f9562e58ac9a911eac197f6ef0460edd20b560253be7ada6991f2188c01e` with transaction type `0x1`

MicahZoltu:
https://eips.ethereum.org/EIPS/eip-2711#new-transaction-receipt

luoyhang003:
> https://eips.ethereum.org/EIPS/eip-2711#new-transaction-receipt

thx, already figured it out.

PR #2724, opened 2020-06-15
EIP - estimate gas return data

julesGoullee:
Link to file: [https://github.com/julesGoullee/EIPs/blob/estimate-gas-return-data/EIPS/eip-draft_estimate_gas_return_data.md](https://github.com/julesGoullee/EIPs/blob/estimate-gas-return-data/EIPS/eip-draft_estimate_gas_return_data.md)

### DIFF
--- a/EIPS/eip-draft_estimate_gas_return_data.md
+++ b/EIPS/eip-draft_estimate_gas_return_data.md
@@ -5,7 +5,7 @@ author: Jules Goullée (@julesGoullee)
 discussions-to: _
 status: Draft
 type: Standards Track
-category Interface
+category: Interface
 created: 2020-06-15
 ---
 
--- a/EIPS/eip-draft_estimate_gas_return_data.md
+++ b/EIPS/eip-draft_estimate_gas_return_data.md
@@ -1,0 +1,41 @@
+---
+eip: _
+title: Estimate gas return data
+author: Jules Goullée (@julesGoullee)
+discussions-to: _
+status: Draft
+type: Standards Track
+category Interface
+created: 2020-06-15
+---
+
+## Simple Summary
+This EIP intends to modify RPC to return transaction data for `eth_estimateGas` method. When transaction required state modification, (not a Solidity view or pure, it can't be execute without a transaction), the estimate endpoint returns the amount of gas needed but omit the returned value.
+
+## Abstract
+The estimate endpoint returns the amount of gas needed but omit the returned value.
+
+## Motivation
+Pure of view function can be call without cost to "read" information and free of cost. State modifier function can be estimate to know the cost and if it will not fail, but there is no way to know the return value before executing the transaction in live. 
+
+## Specification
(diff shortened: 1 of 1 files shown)


EIPS/eip-draft_estimate_gas_return_data.md
```diff
@@ -0,0 +1,39 @@
+---
+eip: _
+title: Estimate gas return data
+author: Jules Goullée (@julesGoullee)
+discussions-to: _
```
MicahZoltu, line 5: You'll need a discussion-to link to get this merged.  The two most common places to do this are either a GitHub issue  or a thread in Ethereum Magicians forums.

EIPS/eip-draft_estimate_gas_return_data.md
```diff
@@ -0,0 +1,39 @@
...
+type: Standards Track
+category: Interface
+created: 2020-06-15
+---
+
+## Abstract
+This EIP intends to modify RPC to return transaction data for `eth_estimateGas` method. When transaction required state modification, (not a Solidity view or pure, it can't be executed without a transaction), the estimate endpoint returns the amount of gas needed but omit the returned value. The propose of this change is to add the return value in the response.
+
+## Motivation
+Pure of view function can be call without cost to "read" information and free of cost. State modifier function can be estimate to know the cost and if it will not fail, but there is no way to know the return value before executing the transaction in live.
```
MicahZoltu, line 16: This isn't true currently with any clients I'm familiar with.  You can always do an `eth_call` of any transaction that you would have done an `eth_estimateGas` or `eth_sendTransaction` for.  The only motivation I can see here is that instead of having to do an `eth_call` *and* an `eth_estimateGas`, you can instead just do one of them.

That being said, getting the result of an on-chain transaction like this is *generally* a bad practice since the transaction may not execute the same on-chain as it does off-chain, so I'm generally against a proposal to encourage that behavior.  Really, contracts should be emitting events with results, or we need to include results in transaction receipts.

EIPS/eip-draft_estimate_gas_return_data.md
```diff
@@ -0,0 +1,39 @@
...
+Returns:
+```json
+{"result": { "gasUsed": 123456, "returnData": "0x" } }
+```
+
+Where `returnData` is typed `Bytes`
+
+## Backwards Compatibility
+Sadly this modification required a breaking change in the RPC in the response format.
+Can also be a new endpoint to avoid breaking change.
```
MicahZoltu, line 35: I recommend a new endpoint.  Backward incompatible changes like this are *very* hard to push through.

3esmit:
Can you see https://github.com/ethereum/go-ethereum/issues/21746

We have this problem with gas estimation don't allows us to set a desired outcome, so the gas estimation goes the lower possible regardless of inner transactions.

Maybe you can also include a "errors" output, which can include all errors (call revert, create fails, etc) that happened within the call and return the one that has less errors?

Issue #1464, opened 2018-10-01
Style Guide -- EIP123 or EIP-123?


fulldecent:
Let us canonicalize the references for EIPs. They are inconsistently referred to as EIP-123 and EIP123 in the EIP documentation as well as the EIPs themselves. Another option is EIP 123. 

# Workplan

- [ ] Decide preferred citation format
- [ ] Update all text that is in the voice of the EIP authors (README, HTML files, EIP-1, etc.)

# Motivation

EIPs are much higher visibility than BIPs or PEPs. To reduce reader and community confusion, we should promote a consistent reference to these standards. 

It is unavoidable that end-users of Ethereum-based systems will be seeing EIP numbers. It will improve the professional image of these projects if we can have these look consistent.

In other words, seeing "You have 5 ERC-20 tokens" in MetaMask and "You have 5 ERC20 tokens" in Etherscan and "These are ERC 20 tokens" in the actual specification is less than optimal. 

# Background / data

EIP is inspired heavily from [BIPS](https://github.com/bitcoin/bips) which draws inspiration from [PEPS](https://github.com/python/peps). Both of those projects refer to proposals like BIP 1234.

Current usage of these words is tracked by Google. "ERC20" is a clear winner versus "ERC-20" and "ERC 20". 

https://trends.google.com/trends/explore?geo=US&q=%22erc%2020%22,%22erc-20%22,%22erc20%22

However this includes `ERC20`, the slug used in contracts.

fulldecent:
Pinging esteemed top contributors to this repository for your thoughts please. @Arachnid @jamesray1 @cdetrio @axic @pirapira @chriseth @wanderer

fulldecent:
I agree with all these points and do support ERC-20 naming for clarity. Previously I did not consider that ERC can be interpreted as a currency symbol. But with all the crazy currency symbols coming out this will now be my primary arguing point.

Let's not cite Wikipedia in this debate. I am an editor on that article, I think there is an XKCD for that. I stopped editing and came where when I realized this is an explicit decision we need to make before cleaning up external references.

jamesray1:
OK, we agree then. Since you are actively involved with Ethereum it's probably best not to edit Wikipedia articles related to Ethereum. I've been told off for doing so.

fulldecent:
[...]

PR #2734, opened 2020-06-17
Update to canonical eip-X format

lightclient:
It appears that the format `EIP-X` is preferred to `EIPX`:

```console
$ grep "EIP-[0-9]+" * | wc -l
373

$ grep "EIP[0-9]+" * | wc -l
118
```

Also note that several of the `EIPX` references are coming from `index.html` and `EIPS/eip-1.md`.

```console
$ grep "EIP[0-9]+" index.html | wc -l
4

$ grep "EIP[0-9]+" EIPS/eip-1.md | wc -l
13
```

I have updated `index.html` and `eip-1` accordingly.


lightclient:
Pinging @Souptacular @axic @nicksavers @Arachnid @gcolvin for thoughts.

lightclient:
Also see: https://github.com/ethereum/EIPs/pull/2737

axic:
Copying this from #2737:
> This has been discussed in #1464. I see little reason having a meta issue, rather than adding this into EIP-1 itself.

This is rather confusing to have two PRs. I suggest to close #2737 and merge this.

lightclient:
I have added the style guide to EIP-1 itself. Please review the copy and let me know if it is acceptable. A potential next task is to comb through the repo and expunge all `EIPX` references and replace them with `EIP-X`. I'd be happy to do this if desired.

MicahZoltu:
I didn't read any dissent in the meeting, and some agreement, with the only real strongly held (from the wording) opinion being that we should pick a standard and go with it.  We have used EIP-X historically, and that is what this PR proposes, so I'm going to merge it.  We can always change it later if people disagree.

PR #2746, opened 2020-06-21
Initial Draft for EIP-Rules-Engine

jaerith:

 - The build passes.


EIPS/eip-draft_rules_engine.md
```diff
@@ -0,0 +1,186 @@
+---
+eip: <to be assigned>
```
lightclient, line 2:
```suggestion
eip: 2746
```

EIPS/eip-draft_rules_engine.md
```diff
@@ -0,0 +1,186 @@
...
+title: Rules Engine Standard
+author: Aaron Kendall (@jaerith), Juan Blanco <@juanfranblanco>
+discussions-to: https://ethereum-magicians.org/t/eip-rules-engine-interface/4374
+status: Draft
+type: Standards Track
+category : ERC
+created: 2020-06-20
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
```
lightclient, line 12: nitpick: you can remove the comments

EIPS/eip-draft_rules_engine.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Specification
+For the clarification of terminology, an Attribute is a registered data point within the data domain, representing data that exists either in the rules engine contract or elsewhere.  A Rule is an predefined action that occurs upon a single data point (i.e., Attribute) in the predefined data domain.  For example, a Rule could check whether the Attribute 'TokenAmt' has a value less than the RHL (i.e., right-hand value) of 10.   A RuleSet is a collection of Rules, where their collection invocation creates a boolean result that determines the navigational flow of execution between RuleSets.  A RuleTree is a collection of RuleSets that are organized within a hierarchy, where RuleSets can contain other RuleSets.
+
+```solidity
+pragma solidity ^0.6.0;
+
+/**
+    @title ERC-#### Rules Engine Standard
```
lightclient, line 30:
```suggestion
    @title ERC-2746 Rules Engine Standard
```

EIPS/eip-draft_rules_engine.md
```diff
@@ -0,0 +1,186 @@
...
+    @dev See https://eips.ethereum.org/EIPS/eip-####
```
lightclient, line 31:
```suggestion
    @dev See https://eips.ethereum.org/EIPS/eip-2746
```

EIPS/eip-draft_rules_engine.md
```diff
@@ -0,0 +1,186 @@
...
+    */
+    function removeRuleTree(address _owner) external returns (bool);    
+}
+```
+
+### Considerations
+
+An argument could be made for interface functions that allow a RuleTree's owner to include others users as executors of the RuleTree.
+
+Another argument could be made for interface functions that allow an administrator to configure the origin point of an Attribute, such as whether the Attribute's value comes from a data structure (internal to the rules engine contract) or from calling a contract method (like an implementation of the Diamond Standard).
```
lightclient, line 143: It would be helpful to link to the Diamond Standard EIP here in addition to the references section.

jaerith: Thanks for the help, @lightclient!

lightclient:
@jaerith I'm not sure why the CI failed here. Could you force push your last changes to retrigger it?

jaerith:
> @jaerith I'm not sure why the CI failed here. Could you force push your last changes to retrigger it?
> Edit: Actually it may not retrigger with the same hash, but I believe [closing and reopening the PR](https://stackoverflow.com/a/17624403/13421890) will retrigger.

I made another small change to trigger the integration.  Still want me to try closing and reopening the PR?

lightclient:
@jaerith I ran the `eip-validator` locally and I realized that you also need to update the file name of your EIP to `eip-2746.md`. I think that will solve the CI issues (although still not sure why there was no output).

MicahZoltu:
I think the idea here is good, but it is unclear to me why this needs to be a standard.  It seems that this would be better suited as a contract/dapp.  Generally speaking, the benefit of having an ERC standard defined in an EIP is when you have a many-to-many relationship, meaning you'll have many contracts interfacing with many different clients.  In this case, it seems like there will likely be a single contract that may have many clients talking to it.  When the relationship is one to many, you can simply publish a contract to the blockchain, publish some documentation on how to interface with it, and then people can build clients against it.

The above is my personal opinion on EIPs (I generally prefer to standardize less, as standardization leads to calcification).  As an EIP editor however, these changes look fine.  If you still want to move forward with an ERC given the above mention me in a comment and I'll merge this for you.

jaerith:
> The above is my personal opinion on EIPs (I generally prefer to standardize less, as standardization leads to calcification). As an EIP editor however, these changes look fine. If you still want to move forward with an ERC given the above mention me in a comment and I'll merge this for you.

Hi @MicahZoltu,

Thanks for the compliments about the idea!  In general, I'm also laissez-faire when it comes to the idea of standardizing, especially since most standards become unused anyway (due to usually being too abstract).  But in this case, I think that such a standard could help with mainstream consideration of Ethereum, especially with enterprises.  As the case with the Baseline protocol, it's my hope that any proposals that focus on business transactions could attract the involvement of enterprise, even before Ethereum 2.0.  Hopefully, this standard (or another like it) would be implemented and used by both the enterprise companies themselves (like a supplier in a pipeline) and software vendors who build solutions as services (as you had suggested).  In the case of the latter, I could see a many-to-many scenario as part of a vendor solution, where each client has its own deployed instance that can interact with other client instances (i.e., one company's RuleTree calls another).  Also, by having a standard interface in the case of vendor solutions, it might help to prevent vendor lock-in, so that the clients could easily move to a different vendor.  Perhaps clients might even standardize and reuse the same rules (like validating a VAT calculation) by sharing them via IPFS or FileCoin in a standard markup form.

Most importantly, though, I don't expect this interface to be the final version of what's really needed.  Instead, I think that it can help serve as the starting point for discussion, to talk about introducing the possibility of a rules engine into the Ethereum space.  It's my hope, actually, that others will become involved to make this proposal better than I ever could!  So, yes, please move this forward with an ERC.  And if you have any ideas for this proposal, I would love to hear them!

PR #2751, opened 2020-06-25
Disable SELFDESTRUCT EIP

AlexeyAkhunov:
Make `SELFDESTRUCT` EVM operation no-op (effectively disable). Contracts will not be able to delete themselves.

### DIFF
--- a/EIPS/eip-draft_disable_selfdestruct.md
+++ b/EIPS/eip-draft_disable_selfdestruct.md
@@ -2,7 +2,7 @@
 eip: <to be assigned>
 title: Disable SELFDESTRUCT
 author: Alexey Akhunov (@AlexeyAkhunov)
-discussions-to: 
+discussions-to: https://ethereum-magicians.org/t/eip-for-disabling-selfdestruct-opcode/4382
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-draft_disable_selfdestruct.md
+++ b/EIPS/eip-draft_disable_selfdestruct.md
@@ -35,8 +35,8 @@ will reduce complexity of EVM going forward, disable GasToken2 (but not GasToken
 polymorthic contracts impossible.
 ## Specification
 <!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
-After certain block number, the semantics of `SELFDESTRUCT` becomes the same as the combination of `POP` followed by `STOP`. Gas cost is the same as the gas cost
-of `POP`, which is 2 gas. No value transfer occurs and no gas refund is given.
+After certain block number, the semantics of `SELFDESTRUCT` becomes the same as the combination of `POP`, followed by transferring remaining ETH
+to the address popped from the stack, followed by `STOP`. Gas cost is the same as the gas cost of non-zero value transfer, which is 9000 gas. No gas refund is given.
 
 ## Rationale
 <!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
--- a/EIPS/eip-draft_disable_selfdestruct.md
+++ b/EIPS/eip-draft_disable_selfdestruct.md
@@ -1,0 +1,63 @@
+---
+eip: <to be assigned>
+title: Disable SELFDESTRUCT
+author: Alexey Akhunov (@AlexeyAkhunov)
+discussions-to: 
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-06-25
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+Disable SELFDESTRUCT
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
+Make `SELFDESTRUCT` EVM operation no-op (effectively disable). Contracts will not be able to delete themselves.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Althouth `SELFDESTRUCT` originally came with EVM to help clean up the state, we learnt that in practice it did not achieve this objective on
+sufficient scale, and is on balance a burden of extra complexity and unintended effects (GasToken2 and polymorphic contracts).
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+Operation `SELFDESTRUCT` (formerly `SUICIDE`) was in the EVM from the beginning, and its purpose was to incentivise clearing the state, by giving the
+caller gas refund. In practice, this incentivisation turned out to be very limited, and the designed purpose was not archieved. However, `SELFDESTRUCT`
+brings significant amount of complexity to the EVM. It is responsible for some of the most arcane edge cases in the EVM semantics.
+It is also used as a vehicle to run an efficient arbitrage of gas prices (GasToken2), which
+ironically lead to the increase use of the state. Introduction of `CREATE2` opcode in Constantinople upgrade created a new phenomenom of
+polymorthic contracts, i.e. contracts that can change their bytecode over time, while residing on the same address. Polymorphic contracts are limited
+in their use, because changing the bytecode via `SELFDESTRUCT` + `CREATE2` clears all the contract storage, making contract lose all its data,
+and making it unsuitable to replace Proxy Pattern as a technique for upgradable contracts. Removing the effect of `SELFDESTRUCT`
+will reduce complexity of EVM going forward, disable GasToken2 (but not GasToken1, which is based on storage clearing refunds), and make
+polymorthic contracts impossible.
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+After certain block number, the semantics of `SELFDESTRUCT` becomes the same as the combination of `POP` followed by `STOP`. Gas cost is the same as the gas cost
+of `POP`, which is 2 gas. No value transfer occurs and no gas refund is given.
+
+## Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+Disabling `SELFDESTRUCT` is the simplest and most effective way to remove its negative effects.
+
+## Backwards Compatibility
+<!--All EIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The EIP must explain how the author proposes to deal with these incompatibilities. EIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+Backwards incompatible and requres a hard fork.
+
+## Test Cases
+<!--Test cases for an implementation are mandatory for EIPs that are affecting consensus changes. Other EIPs can choose to include links to test cases if applicable.-->
+There are a lot of test in the standards suite related to `SELFDESTRUCT` and its edge cases. These need to be modified for the new no-op semantics.
+
+## Implementation
+<!--The implementations must be completed before any EIP is given status "Final", but it need not be completed before the EIP is accepted. While there is merit to the approach of reaching consensus on the specification and rationale before writing code, the principle of "rough consensus and running code" is still useful when it comes to resolving many discussions of API details.-->
+Implementation is likely to be trivial - introducing the eip flag and replacing the old, complex semantics, with the new, simple semantics.
+
+## Security Considerations
+<!--All EIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. EIP submissions missing the "Security Considerations" section will be rejected. An EIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
+At this point, author of this EIP is not aware of any class of smart contracts that rely on `SELFDESTRUCT` for their functionality and security, with the exception of
+GasToken2 contract and similar.
+
+## Copyright
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


MrChico:
An alternative to POP semantics would be to turn its semantics into a TRANSFER. Side effect free tranfers is a legitimate use case of SELFDESTRUCT imo

sorpaas:
> However, `SELFDESTRUCT` brings significant amount of complexity to the EVM.

In practice, you cannot remove any complexity already introduced. The *future* complexity of EVM may be reduced, but all EVMs have to support past block executions, so they will still have the complexity.

> Introduction of `CREATE2` opcode in Constantinople upgrade created a new phenomenom of polymorthic contracts, i.e. contracts that can change their bytecode over time, while residing on the same address.

I think a better semantic to fix this would be to just leave a marker on-chain but keep everything else of `SELFDESTRUCT`. Something like this:

> At the end of the transaction, instead of removing the contract from state, replace the to-be-removed contract with an empty account with nonce `1`.

As @MrChico said, otherwise you'd risk breaking those time-lock contract that relies on `SELFDESTRUCT` to send funds. I remember they're actually quite common.

lightclient:
FYI in case you missed it, the discussion form is: https://ethereum-magicians.org/t/eip-for-disabling-selfdestruct-opcode/4382

AlexeyAkhunov:
> In practice, you cannot remove any complexity already introduced. The _future_ complexity of EVM may be reduced, but all EVMs have to support past block executions, so they will still have the complexity.

You can, if you change the architecture of the client implementation. It is possible to construct the implementation that only uses EVM (simpler) for the current rule set, and special historical EVM (more complex) to re-trace historical transactions, as I mentioned here: https://ethereum-magicians.org/t/backwards-forwards-sync-of-ethereum-clients/2258

And yes, this discussion will disappear when PR is merged, so better discussed on the FEM link

vbuterin:
> but all EVMs have to support past block executions, so they will still have the complexity

Is this actually true? Why can't a future ethereum client just refuse to support blocks older than block N, with its mainnet config pretending that block N is the genesis?

There are things that removing the SELFDESTRUCT refund *will* break; particularly, the CREATE -> do stuff -> SELFDESTRUCT workflow for combining multiple operations into one transaction. That said, that may be fine as there are other EIPs to more "naturally" support such functionality.

EIPS/eip-draft_disable_selfdestruct.md
```diff
@@ -0,0 +1,63 @@
...
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Althouth `SELFDESTRUCT` originally came with EVM to help clean up the state, we learnt that in practice it did not achieve this objective on
+sufficient scale, and is on balance a burden of extra complexity and unintended effects (GasToken2 and polymorphic contracts).
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+Operation `SELFDESTRUCT` (formerly `SUICIDE`) was in the EVM from the beginning, and its purpose was to incentivise clearing the state, by giving the
+caller gas refund. In practice, this incentivisation turned out to be very limited, and the designed purpose was not archieved. However, `SELFDESTRUCT`
+brings significant amount of complexity to the EVM. It is responsible for some of the most arcane edge cases in the EVM semantics.
```
vbuterin, line 28: Strongly agree on this. In particular, the existence of this opcode significantly increases the complexity of many forms of caching, as SELFDESTRUCT can potentially delete an unbounded number of storage slots.

EIPS/eip-draft_disable_selfdestruct.md
```diff
@@ -35,8 +35,8 @@ will reduce complexity of EVM going forward, disable GasToken2 (but not GasToken
 polymorthic contracts impossible.
 ## Specification
 <!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
-After certain block number, the semantics of `SELFDESTRUCT` becomes the same as the combination of `POP` followed by `STOP`. Gas cost is the same as the gas cost
-of `POP`, which is 2 gas. No value transfer occurs and no gas refund is given.
+After certain block number, the semantics of `SELFDESTRUCT` becomes the same as the combination of `POP`, followed by transferring remaining ETH
+to the address popped from the stack, followed by `STOP`. Gas cost is the same as the gas cost of non-zero value transfer, which is 9000 gas. No gas refund is given.
```
sorpaas, line 39: This gas scheme will make it cheaper to create new account via `SELFDESTRUCT` (which costs 30000 right now). Better either:

* Keep the current `SELFDESTRUCT` gas cost: `5000 + 25000 if new account`.
* Or, increase the base to `9000` like this but also with the 25000 new account penalty: `9000 + 25000 if new account`.

Also note that `9000` comes from `CALL` opcode, but `SELFDESTRUCT` is not exactly the same as `CALL` transfer because the former will never execute the target if it's a contract.

EIPS/eip-draft_disable_selfdestruct.md
wjmelements: If selfdestruct did not delete those storage slots it would empower a new kind of upgrade pattern (create2 reincarnations) while reducing SELFDESTRUCT complexity.

I don't think geth ever clears those storage slots from the state trie; it checks if there was a recent self-destruct. A reincarnation could fork the network if not enough nodes remain aware of the selfdestruct and reference a quantum storage slot.

sorpaas:
> Is this actually true? Why can't a future ethereum client just refuse to support blocks older than block N, with its mainnet config pretending that block N is the genesis?

I'm merely speaking as an eth1.x client implementor when talking about this complexity. *Practically*, *right now*, we cannot remove any past complexity under the assumption that we want to support all past block executions. Re-defining genesis or like Alexey proposed doing the backwards-forwards sync works around this, of course, but that does mean we drop support for past block execution.

However, putting off the hat of an Ethereum client implementor, I do agree with you and I believe this problem is already solved nicely. For example, in Substrate, all runtime execution logic is written in wasm and recorded on-chain. So client can re-define logic and remove any accidental complexity at will while still being able to execute all past blocks. I believe there's something similar in eth2 EEs, but I'm not really familiar with that.

wjmelements:
As a gastoken whale I'm strongly opposed. We provide stability to the gasprice and provide more space when blocks are full. Removing the SELFDESTRUCT refund would force us to use SSTORE, which is less-efficient. You would see larger swings in the gasPrice and we would consume more gas overall.

JayeHarrill:
Really happy people are finally taking this seriously.

Been shouting this for 6 months now.

How can I help as a Comms person?

wjmelements:
If you were to break tokenized refunds, you should also fork the state to cleanup every contract they have created and compensate tokenholders at some "fair" gasPrice to limit migratory congestion. I'm aware of the following refund contract creators, but there may be more.

- CHI 0x0000000000004946c0e9F43F4Dee607b0eF1fA1c
- GST2 0x0000000000004946c0e9F43F4Dee607b0eF1fA1c
- TrueUSD 0x0000000000085d4780B73119b644AE5ecd22b376
- TrueCAD 0x00000100F2A2bd000715001920eB70D229700085
- TrueAUD 0x00006100F7090010005F1bd7aE6122c3C2CF0090
- TrueGBP 0x00000000441378008EA67F4284A57932B1c000a5
- TrueHKD 0x0000852600CEB001E08e00bC008be620d60031F2

EIPS/eip-draft_disable_selfdestruct.md
wjmelements: It's currently cheapest to create a new account with a regular 21000 gwei transfer

wjmelements: It's currently cheapest to create an account using a regular 21000 transfer

EIPS/eip-draft_disable_selfdestruct.md
```diff
@@ -0,0 +1,63 @@
...
+Implementation is likely to be trivial - introducing the eip flag and replacing the old, complex semantics, with the new, simple semantics.
+
+## Security Considerations
+<!--All EIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. EIP submissions missing the "Security Considerations" section will be rejected. An EIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
+At this point, author of this EIP is not aware of any class of smart contracts that rely on `SELFDESTRUCT` for their functionality and security, with the exception of
```
Agusx1211, line 59: Some projects are using a `ephemeral contracts` pattern to generate payment accounts, to withdraw from such accounts a contract is deployed on them (using CREATE2), the withdrawal takes place and the contract it's self-destruct.

Because of the `ephemeral` nature of those accounts, authentication can be skipped (either the transaction fails and the account is never created, or the transaction success and the account is emptied). This would still be a safe pattern as long as the project never reuses those addresses, but it would endanger funds otherwise.

A good example of this pattern can be found on UniswapEX (https://github.com/UniswapEx/exchange/blob/master/contracts/libs/Fabric.sol), I think that UniswapEX is not affected, because those accounts are never re-used, but we would need to check if other projects are using the same pattern while re-using accounts.

Agusx1211, line 59: Some projects are using a `ephemeral contracts` pattern to generate payment accounts, to withdraw from such accounts a contract is deployed on them (using CREATE2), the withdrawal takes place and the contract it's self-destruct.

Because of the `ephemeral` nature of those accounts, authentication can be skipped (either the transaction fails and the account is never created, or the transaction success and the account is emptied). This would still be a safe pattern as long as the project never reuses those addresses, but it would endanger funds otherwise.

A good example of this pattern can be found on UniswapEX (https://github.com/UniswapEx/exchange/blob/master/contracts/libs/Fabric.sol), I think that UniswapEX is not affected, because those accounts are never re-used, but we would need to check if other projects are using the same pattern while re-using accounts.

Agusx1211:
> As a gastoken whale I'm strongly opposed. We provide stability to the gasprice and provide more space when blocks are full. Removing the SELFDESTRUCT refund would force us to use SSTORE, which is less-efficient. You would see larger swings in the gasPrice and we would consume more gas overall.

I wonder if a built-in gasToken could act as an alternative to EIP-1556, it could mimic the behaviour of GasToken without actually storing anything on the state (only mimic the refunds), so we would get the "gas stabiliser" effect of GasToken without bloating the storage.

EIPS/eip-draft_disable_selfdestruct.md
```diff
@@ -0,0 +1,63 @@
...
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+Disable SELFDESTRUCT
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
+Make `SELFDESTRUCT` EVM operation no-op (effectively disable). Contracts will not be able to delete themselves.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Althouth `SELFDESTRUCT` originally came with EVM to help clean up the state, we learnt that in practice it did not achieve this objective on
```
AndreaLanfranchi, line 21: `Althouth` should read `although`

EIPS/eip-draft_disable_selfdestruct.md
```diff
@@ -0,0 +1,63 @@
...
+in their use, because changing the bytecode via `SELFDESTRUCT` + `CREATE2` clears all the contract storage, making contract lose all its data,
+and making it unsuitable to replace Proxy Pattern as a technique for upgradable contracts. Removing the effect of `SELFDESTRUCT`
+will reduce complexity of EVM going forward, disable GasToken2 (but not GasToken1, which is based on storage clearing refunds), and make
+polymorthic contracts impossible.
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
+After certain block number, the semantics of `SELFDESTRUCT` becomes the same as the combination of `POP`, followed by transferring remaining ETH
+to the address popped from the stack, followed by `STOP`. Gas cost is the same as the gas cost of non-zero value transfer, which is 9000 gas. No gas refund is given.
```
axic, line 39: > the address popped from the stack, followed by `STOP`.

This suggests selfdestruct can be executed multiple times. Is that the goal?

axic, line 39: I don't think there's any point introducing new pricing here, if the specification is that `SELFDESTRUCT` is replaced with `CALL(<addr>, <value>, gas=0, ..`) and `STOP`, then just state that and apply the gas rules of those two.

wjmelements:
> I wonder if a built-in gasToken could act as an alternative to EIP-1556, it could mimic the behaviour of GasToken without actually storing anything on the state (only mimic the refunds), so we would get the "gas stabiliser" effect of GasToken without bloating the storage.

The convenience of the storage overhead is that we free the space when the network is congested, so in exchange for the additional gas we free more storage than the additional usage fills. By outbidding low-priority transactions in low congestion, we consume less space than usage that would not zero-out, thereby smoothing the growth of the state trie.

EIPS/eip-draft_disable_selfdestruct.md
```diff
@@ -0,0 +1,63 @@
...
+sufficient scale, and is on balance a burden of extra complexity and unintended effects (GasToken2 and polymorphic contracts).
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+Operation `SELFDESTRUCT` (formerly `SUICIDE`) was in the EVM from the beginning, and its purpose was to incentivise clearing the state, by giving the
+caller gas refund. In practice, this incentivisation turned out to be very limited, and the designed purpose was not archieved. However, `SELFDESTRUCT`
+brings significant amount of complexity to the EVM. It is responsible for some of the most arcane edge cases in the EVM semantics.
+It is also used as a vehicle to run an efficient arbitrage of gas prices (GasToken2), which
+ironically lead to the increase use of the state. Introduction of `CREATE2` opcode in Constantinople upgrade created a new phenomenom of
```
lightclient, line 30:
```suggestion
ironically lead to the increase use of the state. Introduction of `CREATE2` opcode in Constantinople upgrade created a new phenomenon of
```

EIPS/eip-draft_disable_selfdestruct.md
```diff
@@ -0,0 +1,63 @@
...
+to the address popped from the stack, followed by `STOP`. Gas cost is the same as the gas cost of non-zero value transfer, which is 9000 gas. No gas refund is given.
+
+## Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+Disabling `SELFDESTRUCT` is the simplest and most effective way to remove its negative effects.
+
+## Backwards Compatibility
+<!--All EIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The EIP must explain how the author proposes to deal with these incompatibilities. EIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+Backwards incompatible and requres a hard fork.
```
lightclient, line 47:
```suggestion
Backwards incompatible and requires a hard fork.
```

EIPS/eip-draft_disable_selfdestruct.md
lightclient, line 21:
```suggestion
Although `SELFDESTRUCT` originally came with EVM to help clean up the state, we learnt that in practice it did not achieve this objective on
```

EIPS/eip-draft_disable_selfdestruct.md
wjmelements: This opertation should be cheaper than CALL because it doesn't need to load code or create an execution context.

axic: > because it doesn't need to load code or create an execution context.

This EIP is changing the semantics of SELFDESTRUCT. From the current description it is not clear whether it turns into a non-executing transfer or not.

carver: Hm, turning it into an executing transfer would be a nice EVM unification/simplification. `SELFDESTRUCT` is currently one of the "sneakiest" ways to add funds to an address; a capability that people often forget about.

wjmelements: @carver Developers only "forget" about it because of [deceptive Solidity syntax](https://github.com/ethereum/solidity/issues/7727). Nobody should assume their contract is "nonpayable". This does not change that because miners can still break that assumption via coinbase.

wjmelements:
Polymorphic contracts are the superior upgrade model. If you have thousands of token balances and allowances it's much cheaper to upgrade the code in-place than to migrate to a new contract. If it's too much work for the evm to ensure all of the storage slots are cleared, just don't do it; it would save me the trouble of resetting all contract state. Most self-destructs are for contracts without state; very few contracts self-destruct with remaining state.

Instead you could incentivize other contracts to clear it out of the trie manually. A new opcode could point out external storage corresponding to an account without code and clear it for a gas refund.

EIPS/eip-draft_disable_selfdestruct.md
MicahZoltu: Ooh, how did you do this @lightclient?

lightclient: There should be a plus/minus button as part of the comment interface :)

PR #2764, opened 2020-07-03
Update word choice

fulldecent:
++

### DIFF
--- a/EIPS/eip-1613.md
+++ b/EIPS/eip-1613.md
@@ -68,12 +68,12 @@ Implementing a `RelayRecipient` contract:
 * Use `getSender()` and `getMessageData()` instead of `msg.sender` and `msg.data`, everywhere. `RelayRecipient` provides these functions and gets the information from `RelayHub`.
 * Implement a `acceptRelayedCall(address relay, address from, bytes memory encodedFunction, uint gasPrice, uint transactionFee, bytes memory approval)` view function that returns **zero** if and only if it is willing to accept a transaction and pay for it. 
   `acceptRelayedCall` is called by `RelayHub` as a view function when a `Relay` inquires it, and also during the actual transaction. Transactions are reverted if **non-zero**, and `Relay` only gets compensated for transactions (whether successful or reverted) if `acceptRelayedCall` returns **zero**. Some examples of `acceptRelayedCall()` implementations:
-    * Whitelist of trusted dapp members.
+    * Allowlist of trusted dapp members.
     * Balance sheet of registered users, maintained by the dapp owner. Users pay the dapp with a credit card or other non-ETH means, and are credited in the `RelayRecipient` balance sheet. 
       Users can never cost the dapp more than they were credited for.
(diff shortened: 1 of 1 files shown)


lightclient:
@fulldecent thoughts on `denylist` instead of `blocklist`?

fulldecent:
Both are 100% fine. Probably blocklist is more popular lately. And that is not a definitive reason to choose one word over another.

tjayrush:
`blockList` is a really bad choice in my opinion. It sounds pretty much like a list of `blocks`. `block` is already quite heavily laden with meaning, so I would suggest against piling on yet one more.

fulldecent:
This PR is proposing "blocklist" as a normal English word which, along with allowlist, are trading to become words of the year.

"blockList" (with a capital "L") and `blocklist` (as the name of a variable) are not proposed.

---

And of course there is no discussion necessary to agree that "blacklist" is not a term we want to use to refer to contracts that are being "[refused] to serve transactions" and "malicious users in their system".

tjayrush:
> And of course there is no discussion necessary to agree that "blacklist" is not a term we want to use to refer to contracts that are being "[refused] to serve transactions" and "malicious users in their system".

I agree with this, of course.

I was only suggesting that the word 'block' is already quite laden with meaning in a BLOCKchain, and perhaps adding additional connotations is ill-advised.

fulldecent:
"Chain reaction" is also a term that includes the word "chain". This term could cause confusion in the context of blockchains. However, because "chain reaction" has such a well-known meaning in chemistry, everyone will immediately recognize this chemistry usage over any more specific context-dependent interpretation.

Likewise, "blocklist" is quickly being recognized to mean "a list of bad things". (And we are all certain that this momentum will continue.) If QuickBlocks or any other project is using a variable named `blockList` to refer to something other than "a list of bad things", then it would be prudent to rename _those_ cases to "listOfBlocks", "blockArray", "blocks" or similar.

And besides, the style guide written by the author of C++ [recommends against this usage](https://www.stroustrup.com/bs_faq2.html?#Hungarian) where the variable type (e.g. "list") is included in the variable name (e.g. `blockList`), this is called Hungarian notation.

ricky-sb:
I propose `denylist`. `denylist` reduces the cognitive complexity in identifying the functionality.
- The antonym of `allow` is `deny` 
- `deny` has a single lexical category, and a single meaning, making it more explicit
- `block` has 2 lexical categories, and a total of 11 meanings (according to Google)

lightclient:
I also prefer `denylist` to `blocklist`.

axic:
@yoav-tabookey @drortirosh @forshtat can you approve?

MicahZoltu:
To be clear (and hopefully unbiased) you can also deny.  As the authors of the spec you have deciding power over the chosen terminology.  Regardless of how you decide it would be nice to close out this PR one way or another.

axic:
Yes, I meant to write "review", but got lazy.

Issue #2766, opened 2020-07-03
ERC: Contract Ownership Governance Standard


zemse:
Opening this thread to discuss a construction that can be standardized.

Traditionally, many contracts that require that they be owned or controlled in some way use `EIP-173` which standardizes the use of ownership in the smart contracts. For example to withdraw funds or perform administrative actions.

```solidity
contract dApp {
  function doSomethingAdministrative() external onlyOwner {
    // admin logic that can be performed by a single wallet
  }
}
```

Often, such administrative rights for a single wallet are written for maintainance purpose, but it overpowers the owner wallet and users need to trust the owner. Rescue operations by owner wallet have raised questions on decentralised nature of the dApp projects. Also, there is a possibility of compromise of owner's private key. All it needs is a single wallet to do a single transaction in a dApp to do misuse of the powers.

This can be solved by having the `dApp` smart contract to consider address of another smart contract `Governance` as it's owner (instead of a wallet address).

```solidity
contract Governance {
  function makeGovernedCall(
    uint256 _nonce,
    address _to,
    bytes memory _data,
    bytes[] memory _signatures
  ) external payable {
    // performs initial checks like enough signatures

    // once all checks are done, the governance makes
    //  an internal call to the application contract
    (bool _success, ) = _to.call{value: msg.value}(_data);
  }
}
```

Calls can be made by this contract to `EIP-173` compatible comtracts. This means existing contracts with a centralized wallet controlling the administrative ownership can be decentralized to multiple wallets with minimum complicated setup.

The reference implementation is available [here](https://github.com/zemse/smart-contract-governance/blob/master/contracts/SimpleGovernance/Governance.sol) in the [repository](https://github.com/zemse/contract-ownership-governance#contract-ownership-governance). I have also written some test cases in the same repo.

This is still very nascent. I believe improvising this would benefit users of existing and future dApps on Ethereum by not requiring to trust owners of dApps. Also do any similar constructions exists?

zemse:
Multi sig wallets do have the same functionality. My bad, closing it.

mudgen:
Hey, this is not just useful for wallets.  I think this would be very useful for governance of diamonds.  It would decentralized the ability to add new functions to diamonds.   See the Diamond Standard here: https://github.com/ethereum/EIPs/issues/2535

zemse:
[...]

PR #2767, opened 2020-07-04
Create EIP-2767: Contract Ownership Governance

zemse:
Discussions thread: https://github.com/ethereum/EIPs/issues/2766

### DIFF
--- a/EIPS/eip-2767.md
+++ b/EIPS/eip-2767.md
@@ -68,7 +68,7 @@ interface ERC2767 {
     /// @return Sum of the power of all governors
     function totalPower() external view returns (uint256);
 
-    /// @notice Gets number votes required for acheiving consensus
+    /// @notice Gets number votes required for achieving consensus
     /// @return Required number of consensus votes
     function required() external view returns (uint256);
 
@@ -88,7 +88,7 @@ interface ERC2767 {
 
 Implementations that require equal governance rights between the governors can simply have equal power values for every governor.
 
-Implementations can have a simple number of required votes value or any complex logic that depends on other factors (for e.g. `50%+` or `66%+` of `totalPower()`). The `required()` should return the calculated number of votes using the complex logic.
+There is a possibility that `totalPower()` changes rapidly due to adding or removing governors. This can result in having a fixed `required()` return value easier or difficult to achieve. Implementations can have custom logic that affect the return value of `required()`. For example the return value of `required()` can be changed by custom logic in order to maintain `51%` of `totalPower()`. So if more governors are added in such implementations, the `required()` return value would automatically take that into account.
 
 ### Interface Identification
 
--- a/EIPS/eip-2767.md
+++ b/EIPS/eip-2767.md
@@ -47,82 +47,15 @@ At present, there are various implementation ideas for handling governance. Adhe
 #### Notes
 
 - The following specifications use syntax from Solidity 0.6.0 (or above)
+- Governor is the board member who has voting right.
+- Power represents the voting privilege of a governor.
 
-#### Terminology
-
-- Governor: The board member who has voting rights.
-
-### Governance Type
-
-A governance contract can have equal or privileged voting rights.
-
-#### Equal voting power
-
-```solidity
-/// @title ERC-2767 Governance with Equal Voting Rights
-/// @dev ERC-165 InterfaceID: 0xbfca4246
-interface IGovernanceEqual {
-    /// @dev This emits when a governor is added
-    event GovernorAdded(address indexed governor);
-
-    /// @dev This emits when a governor is removed
-    event GovernorRemoved(address indexed governor);
-
-    /// @notice Gets list of governors
-    /// @return List of governors
-    function getGovernors() external view returns (address[] memory);
-
-    /// @notice Gets whether an address is governor or not
-    /// @param _governor Address of governor
-    /// @return governor status
-    function isGovernor(address _governor) external view returns (bool);
-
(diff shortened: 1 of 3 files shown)


mudgen:
@MicahZoltu Can you please merge this pull request?

EIPS/eip-2767.md
```diff
@@ -0,0 +1,239 @@
...
+
+    /// @notice Gets number votes required for achieving consensus
+    /// @return Required number of consensus votes
+    function required() external view returns (uint256);
+
+    /// @notice Updates governor statuses
+    /// @param _governor Governor address
+    /// @param _newPower New power for the governor
+    /// @dev Only Governance can call
+    function setGovernor(address _governor, uint256 _newPower) external;
```
MicahZoltu, line 79: This function doesn't seem like it needs to be standardized.

EIPS/eip-2767.md
```diff
@@ -0,0 +1,239 @@
...
+    /// @return The governor's voting power
+    function powerOf(address _governor) external view returns (uint256);
+
+    /// @notice Gets the sum of the power of all governors
+    /// @return Sum of the power of all governors
+    function totalPower() external view returns (uint256);
+
+    /// @notice Gets number votes required for achieving consensus
+    /// @return Required number of consensus votes
+    function required() external view returns (uint256);
```
MicahZoltu, line 73: Not all governed contracts will have voting.  Having voting be part of the interface may restrict its application.  A common (and IMO better) solution to governance is to have a single governor but that governor would be a multisig or something.  This means that the contract in question would *not* conform to this specification, and thus wouldn't show as having any governance.

EIPS/eip-2767.md
```diff
@@ -0,0 +1,239 @@
...
+    ///  uses less than 30,000 gas.
+    /// @return `true` if the contract implements `interfaceID` and
+    ///  `interfaceID` is not 0xffffffff, `false` otherwise
+    function supportsInterface(bytes4 interfaceID) external view returns (bool);
+}
+```
+
+### Transaction execution (Optional)
+
+There are two optional interfaces for transaction execution: On-chain and Off-chain. A `ERC-2767` compatible governance contract can optionally implement one of them or both. The contract should specify the interfaceID(s) using `ERC-165`'s `supportsInterface` method.
```
MicahZoltu, line 105: This whole section feels out of place in a standard.  I see very little value in standardizing how contracts are called by their owners.  Each owner can choose to use different tools/mechanisms and that is OK.  Recommend including a section in the rationale why you think this is something that should be standardized.

If you do think standardization of contract execution is valuable, then I think it should be a separate EIP.  As far as standards go, many short standards tend to be better than one monolithic standard.  It lets people pick and choose what they want to implement.

EIPS/eip-2767.md
```diff
@@ -0,0 +1,239 @@
+---
+eip: 2767
+title: Contract Ownership Governance
+author: Soham Zemse (@zemse), Nick Mudge (@mudgen)
+discussions-to: https://github.com/ethereum/EIPs/issues/2766
+status: Draft
+type: Standards Track
+category: ERC
+created: 2020-07-04
+requires: 173, 165, 191
```
MicahZoltu, line 10: Note: Two of these are still draft, and they'll need to go to final before this goes to final.

EIPS/eip-2767.md
zemse: I agree, users should be able to customize the way they want to do this. Any UI tool wouldn't care about this method at all.

EIPS/eip-2767.md
zemse: I think this EIP is more of a multi-sig interface. You are correct that this should not always have to be voting. Will improvise on this.

zemse:
@MicahZoltu Thanks for the valuable review! Just a quick question, are sub-specifications under same EIP number allowed? Or always a seperate EIP number will be needed per specification?

MicahZoltu:
@zemse New EIP number for each EIP.

EIPS/eip-2767.md
mudgen: I disagree. User interfaces and tools can be developed to work with any EIP-2767 contract to add new governors and set the power. Having this function makes that possible.

EIPS/eip-2767.md
mudgen: @MicahZoltu I am not following.  How can you have decentralized governing without some kind of voting?   The IGovernanceOffchain interface in the standard provides a multisignature function for voting.

EIPS/eip-2767.md
mudgen: Yes, agree there is some merit in not standardizing how contracts are called by their owners.  But there is also some usefulness to it.  That's why these voting interfaces are optional.

1. Standardizing the voting makes it easier to make user interfaces and tools without contract integration problems.
2. Standardizing the voting makes it easier to understand and adopt implementations because they follow the standard.
3. Every choice a standard makes is a choice that implementors don't have to figure out and choose which makes the implementor's job easier as long as the standard is flexible and correct enough not to get in the way.

EIPS/eip-2767.md
mudgen: Good catch.

mudgen:
I appreciate @MicahZoltu's feedback and it is valuable to consider different angles and perspectives of what a standard should do and to question things.  I do think this standard is shaping up and I think the things in it belong together in the same standard.  I would like to see more feedback from more people.

PR #2768, opened 2020-07-04
remove circular dependency in parameter definitions

i-norden:


EIPS/eip-1559.md
```diff
@@ -48,15 +48,15 @@ The transition to this gas price system will occur in two phases, in the first p
 * `FINAL_FORK_BLKNUM`: `INITIAL_FORK_BLKNUM + MIGRATION_DURATION_IN_BLOCKS`
 * `BLOCK_GAS_TARGET`: The gas limit field set in the block header, controlled by miner voting.
 * `EIP1559_INITIAL_GAS_TARGET`: `BLOCK_GAS_TARGET / 2`
-* `LEGACY_INITIAL_GAS_LIMIT`: `BLOCK_GAS_TARGET - EIP1559_GAS_TARGET`
+* `LEGACY_INITIAL_GAS_LIMIT`: `BLOCK_GAS_TARGET / 2`
```
MicahZoltu, line 51: The "bug" here was that `BLOCK_GAS_TARGET - EIP1559_GAS_TARGET` should have been `BLOCK_GAS_TARGET - EIP1559_INITIAL_GAS_TARGET`.  In theory, by having it that way it would be "easier" to update the EIP with a different initial distribution (e.g., if we wanted to start with 10% of block switched over initially).

EIPS/eip-1559.md
```diff
@@ -48,15 +48,15 @@ The transition to this gas price system will occur in two phases, in the first p
...
 * `EIP1559_GAS_TARGET`:
   ```
   if CURRENT_BLKNUM >= FINAL_FORK_BLOKNUM then
     BLOCK_GAS_TARGET
   elif CURRNT_BLKNUM < INITIAL_FORK_BLKNUM then
     0
   else
-    EIP1559_INITIAL_GAS_TARGET + LEGACY_INITIAL_GAS_LIMIT * (CURRENT_BLKNUM - INITIAL_FORK_BLKNUM) / MIGRATION_DURATION_IN_BLOCKS
+    EIP1559_INITIAL_GAS_TARGET + (BLOCK_GAS_TARGET / 2) * (CURRENT_BLKNUM - INITIAL_FORK_BLKNUM) / MIGRATION_DURATION_IN_BLOCKS
```
MicahZoltu, line 59: Same comment as above, by spreading out the `/2` rather than using variables, if we decide we want to change the initial distribution we have to update more parts of the EIP (more likely to miss/forget something and introduce a bug in the spec).  This change isn't incorrect, it just has *potential* to make the EIP a tad harder to maintain.

PR #2711, opened 2020-06-11
Creates EIP-2711: Separate gas payer from msg.sender

MicahZoltu:

PR #2771, opened 2020-07-06
Recipient eip

forshtat:

 - The build passes.


lightclient:
Hi @forshtat, it looks like the `author` field here is blank. This is causing CI to fail.

MicahZoltu:
@forshtat It appears that the Abstract and Motivation sections are missing.

Also, you may want to check out https://eips.ethereum.org/EIPS/eip-2711 which separates "sender" from "gas payer" at the protocol layer, which I believe will solve this problem (and others) in a more generalized way.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,129 @@
+---
+eip: 2771
+title: Secure Protocol for Native Meta Transactions
+author: Ronan Sandford (@wighawag), Liraz Siri (@lirazsiri),  Dror Tirosh (@drortirosh), Yoav Weiss (@yoavw), Alex Forshtat (@forshtat), Hadrien Croubois (@Amxx), Tomar Sachin (@tomarsachin2271), Patrick McCorry (@stonecoldpat), Nicolas Venturo (@nventuro), Fabian Vogelsteller (@frozeman) 
+discussions-to: https://github.com/ethereum/EIPs/pull/2771
+status: Draft
+type: Standards Track
+category: ERC
+created: 2020-07-01
+requires: 2770
```
MicahZoltu, line 10: Why does this require 2770?  I don't see any references to 2770 anywhere else in this EIP.

MicahZoltu: Also, we can't merge this until #2770 is merged, so will have to wait a bit.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,185 @@
...
+discussions-to: https://github.com/ethereum/EIPs/pull/2771
+status: Draft
+type: Standards Track
+category: ERC
+created: 2020-07-01
+---
+
+## Simple Summary
+
+We define a minimal contract-level protocol for receiving a meta transaction
```
MicahZoltu, line 14:
```suggestion
A minimal contract-level protocol for receiving a meta transaction
```

EIPS/eip-2771.md
```diff
@@ -0,0 +1,185 @@
...
+through a trusted forwarder contract that helps a recipient contract avoid
+confusing the address of the transaction relayer with the address of the
+transaction signer.
```
MicahZoltu, line 17: You probably should shorten this up a bit and leave some content for the abstract.  😄 
```suggestion
A contract interface for receiving meta transactions through a trusted forwarder.
```

EIPS/eip-2771.md
```diff
@@ -0,0 +1,185 @@
...
+            bytes memory data = msg.data;
+            uint256 length = msg.data.length;
+            assembly { signer := mload(add(data, length))) }
+        }    
+    }
+
+}
+```
+
+## Security Considerations
```
MicahZoltu, line 170: Missing Rationale section.  This is where you indicate why you made specific decisions you did in the specification.  Like why choose name X over name Y, why fire events at point A instead of point B, etc.
```suggestion
## Rationale
TBD

## Security Considerations
```

EIPS/eip-2771.md
lirazsiri: The dependency on 2770 was a mistake. Removed!

EIPS/eip-2771.md
lirazsiri: Added Rationale section.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,204 @@
...
+        }    
+    }
+
+}
+```
+
+## Rationale
+
+* Without support for meta transactions in the recipient contract, an externally owned 
+  account can not use meta transactions to interact with the receipient contract.
```
MicahZoltu, line 171:
```suggestion
account can not use meta transactions to interact with the recipient contract.
```

MicahZoltu:
Since you didn't respond to my comment about #2711 I'm going to assume that means you want to move forward with this.  Will merge as draft as soon as the spelling error is fixed (currently blocking CI).

EIPS/eip-2771.md
lirazsiri: fixed typo

lirazsiri:
Yes Micah, we would still like to move forward. This ERC is backwards compatible with the existing EVM and will not require any protocol changes, so there's no reason to block on a future protocol change. Sometimes worse is better. In the longer term, I believe proposing protocol changes like #2711  to better support gas abstraction are desirable, but on the other hand, I recognize that building up consensus towards including things in the next hard fork takes time and is more of a marathon effort than a sprint. In the meantime, we can move forward pragmatically with what is achievable without that. FWIW, reaching agreement between the authors on even on this seemingly trivial proposal was months in the making. We started working towards that a couple of weeks before EthCC.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,211 @@
+---
+eip: 2771
+title: Secure Protocol for Native Meta Transactions
+author: Ronan Sandford (@wighawag), Liraz Siri (@lirazsiri),  Dror Tirosh (@drortirosh), Yoav Weiss (@yoavw), Alex Forshtat (@forshtat), Hadrien Croubois (@Amxx), Tomar Sachin (@tomarsachin2271), Patrick McCorry (@stonecoldpat), Nicolas Venturo (@nventuro), Fabian Vogelsteller (@frozeman) 
+discussions-to: https://github.com/ethereum/EIPs/pull/2771
```
axic, line 5: @MicahZoltu EIP-1 states that pull requests cannot be used as discussion URLs. I'm not here to judge whether that is a good or bad idea, but this draft was merged against EIP-1 rules.

@lightclient is that enforced in eipv?

EIPS/eip-2771.md
```diff
@@ -0,0 +1,211 @@
...
+put in place. 
+
+A Gas limit of 10k is enough for making  the decision either inside the
+contract, or delegating it to another contract and doing some memory access
+calculations, like querying a mapping.
+
+###  Recipient example 
+
+```solidity
+contract RecipientExample {
```
axic, line 140: There is no version pragma, which would be verify useful for anyone reading this standard in a few months/years.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,211 @@
...
+
+    function purchaseItem(uint256 itemId) external {
+        address sender = _msgSender();
+        ... perform the purchase for sender
```
axic, line 144: Seems to be invalid solidity code.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,211 @@
...
+    function isTrustedForwarder(address forwarder) external returns(bool) {
+        return forwarder == _trustedForwarder;
+    }
+
+    function _msgSender() internal view returns (address payable signer) {
+        signer = msg.sender;
+        if (isTrustedForwarder(signer)) {
+            bytes memory data = msg.data;
+            uint256 length = msg.data.length;
+            assembly { signer := mload(add(data, length))) }
```
axic, line 161: It should be possible to use slicing on `msg.data` to avoid this mess.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,211 @@
...
+        _trustedForwarder = trustedForwarder;
+    }
+
+    function isTrustedForwarder(address forwarder) external returns(bool) {
+        return forwarder == _trustedForwarder;
+    }
+
+    function _msgSender() internal view returns (address payable signer) {
+        signer = msg.sender;
+        if (isTrustedForwarder(signer)) {
```
axic, line 158: Since `isTrustedForwarder` is marked external, this expression here likely will create an external functional call to do the comparison. This is quite expensive. Probably you do not want to suggest this to your users.

EIPS/eip-2771.md
```diff
@@ -0,0 +1,211 @@
...
+    Relay**
+*   **Gas Relay** - receives a signed request off-chain from **Transaction
+    Signer** and pays gas to turn it into a valid transaction that goes through
+**Trusted Forwarder**
+*   **Trusted Forwarder** - a contract that is trusted by the `Recipient` to
+    correctly verify the signature and nonce before forwarding the request from
+**Transaction Signer**
+*   **Recipient** - a contract that can securely accept meta-transactions
+    through a **Trusted Forwarder** by being compliant with this standard.
+
```
axic, line 73: It seems that the specification section does not explicitly state the specification: the sender address is appended to the end of the calldata.

It is only explained under the "extracting" and "example" subsections.

EIPS/eip-2771.md
MicahZoltu: Missed that, good catch!  Definitely something that would be nice for the bot to catch since it is easy to miss (valid Issue URL looks very similar to invalid PR URL).

lightclient: I have listed this as a [requirement](https://github.com/lightclient/eipv#requirements) for `eipv`, but it has not been implemented yet.

lirazsiri: I issued another PR to redirect the discussion to the following Ethereum Magicians post:

https://ethereum-magicians.org/t/eip-2771-secure-protocol-for-native-meta-transactions/4488

EIPS/eip-2771.md
lirazsiri: There is a pragma in the linked implementation. I don't mind adding it to the example if you think it would be very useful, but I am worried that it will add a bit of unnecessary clutter and possibly confuse with regards to the compatibility of the code with earlier versions.

EIPS/eip-2771.md
lirazsiri: How would you rewrite this for clarity? This is what this function looks like in the reference base class implementation:

    function _msgSender() internal override virtual view returns (address payable ret) {
        if (msg.data.length >= 24 && isTrustedForwarder(msg.sender)) {
            // At this point we know that the sender is a trusted forwarder,
            // so we trust that the last bytes of msg.data are the verified sender address.
            // extract sender address from the end of msg.data
            assembly {
                ret := shr(96,calldataload(sub(calldatasize(),20)))
            }
        } else {
            return msg.sender;
        }
    }

EIPS/eip-2771.md
lirazsiri: We don't expect users to use the examples in the EIP as they are optimized for readability. We expect users to use a reference base class such as this one:

https://github.com/opengsn/forwarder/blob/master/contracts/BaseRelayRecipient.sol

EIPS/eip-2771.md
lirazsiri: "The Trusted Forwarder is responsible for calling the Recipient contract and MUST append the address of the Transaction Signer (20 bytes of data) to the end of the call data."

Is it unclear? Would duplicating the above paragraph or parts of it help? Or are you suggesting to move it somewhere else? The specification section is divided into subsections for clarity and context. 

The rationale for having it below the "Extracting The Transaction Signer address" heading, is that explaining what needs to happen in terms extraction of the signer helps give additional context that may help explain why the signer was appended to the calldata.

EIPS/eip-2771.md
lirazsiri: Yes, it is supposed to be **obviously** invalid. Unless we actually want to implement the logic for purchaseItem we have either invalid syntax or invalid semantics.

EIPS/eip-2771.md
MicahZoltu: The `Extracting The Transaction Signer Address` is a subsection of the `Specification` section, so I think it is fine there.

MicahZoltu: You may want to put the above definitions into a `### Definitions` subsection though, just to help with automated table of contents generation.

PR #2772, opened 2020-07-06
[EIP-1193] Strict readonly typing for typescript definitions

bogdan:
Added `readonly` keyword to all typescript interface definitions.

I decided to avoid `readonly` on the `ProviderRpcError` because it extends `Error` which does not have its properties as readonly. We can assume that this is a really bad Javascript standard that we have to follow. In addition `message` property was removed from the interface because it is defined in original `Error` interface:  https://github.com/microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L972

https://github.com/ethereum/EIPs/issues/2319#issuecomment-654343547

### DIFF
--- a/EIPS/eip-1193.md
+++ b/EIPS/eip-1193.md
@@ -71,8 +71,8 @@ All API entities **MUST** adhere to the types and interfaces defined in this sec
 
 ```typescript
 interface RequestArguments {
-  method: string;
-  params?: unknown[] | object;
+  readonly method: string;
+  readonly params?: readonly unknown[] | object;
 }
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - EIP 1193 requires approval from one of (@rekmarks, @maiavictor, @frozeman, @marcgarreau, @ryanio)

MicahZoltu:
This feedback was integrated into #2696 (which is a break out of 1193).  @rekmarks indicated that they were in favor of this, you can auto-merge by just doing a PR review and choosing approve.

PR #2774, opened 2020-07-07
EIP 2774 - A price-sensitive challenge process 

stonecoldpat:


stonecoldpat:
I'm not really sure what to do next.

EIPS/eip-2774.md
```diff
@@ -0,0 +1,129 @@
+---
+eip: 2774
+title: A price-sensitive challenge process 
+author: Patrick McCorry <@stonecoldpat>
+discussions-to: https://github.com/ethereum/EIPs/issues/2775
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-07-07
```
lightclient, line 9: I believe you need specify that you are depending on EIP-1559 -- in that case, you'd need to add `requires: 1559` to your preamble.

EIPS/eip-2774.md
```diff
@@ -0,0 +1,129 @@
...
+
+**Defining max fee.** We define the ```MAXFEE``` as the maximum gas price the user is willing to pay for their transaction to be included in the blockchain. We re-use the ```BASEFEE``` from EIP-1559 to determine the minimum gas price a transaction must include in order to be accepted into the respective block. As long as ```MAXFEE >= BASEFEE```, then we know the transaction could have been accepted into the block. 
+
+**Sliding time window.** We change the challenge time period from a fixed deadline ```T``` to a sliding window such that the blockchain checks if ```K``` of the previous ```N``` blocks can include transactions that paid at most ```MAXFEE```. If that condition is not satisfied, then the sliding window progresses until either 1) the ```BASEFEE``` comes down or 2) it reverts to a distant fixed deadline ```T```. 
+
+Note: If we assume the basefee is publicly accessible in a smart contract, then a new opcode is not required. However if it is implemented as a pre-compile, then the gas-costs for the lookup can potentially be cheaper.
```
lightclient, line 28: Have you contacted anyone involved with EIP-1559 to determine if `BASEFEE` will be accessible in the EVM? If they are planning for it not to be, they may have a reason for doing so. Otherwise, adding that opcode would probably be better suited to occur in EIP-1559 itself.

PR #2777, opened 2020-07-09
747: Cleanup

rekmarks:
- Change required EIP from 1474 to 1193
  - 1193 is more apt, 1474 is just trying to specify the Ethereum client JSON-RPC interface, it's not really relevant (and the client teams also don't seem to care about it)
- Make some clarifying edits to language
- Specify that it's an RPC method, and that its name is `wallet_watchAsset`, not `metamask_watchAsset` (gifs must be updated separately)
- Update examples per EIP-1193
- Update RPC method documentation per e.g. EIP-695
- Change return in case of failure from `false` to an error


eip-automerger:

 - EIP 747 requires approval from one of (@danfinlay, @estebanmino)

EIPS/eip-747.md
```diff
@@ -31,68 +38,105 @@ In the second case, the user experience is terrible.
...
-) {
-/* Implementation would go here */
-}
+### wallet_watchAsset
 
-// Sample usage:
-web3.wallet.watchAsset({ type, options })
-```
-As there are several types of different assets, this method has to provide support for each of them in a separate way. If it doesn't, it should give a response according to that.
+Requests that a specified asset be added to the user's wallet, and returns `true` if the asset was successfully added, or an error if it was not.
```
Gudahtt, line 48: The current implementation in the MetaMask extension returns `false` if the user rejects the suggestion. Was changing the return in this case intentional? Which error would be returned if the user rejected the suggestion?

rekmarks: Hm, we should refer to 1193 for the errors: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#provider-errors

I'll update that.

MicahZoltu:
@danfinlay @estebanmino you can trigger an automerge of this by doing a PR review and approving.  If you are not interested in maintaining this EIP, I recommend assigning a new author or deleting the EIP (if you desire the latter, I can take care of that for you).

axic:
@danfinlay @estebanmino you can approve/reject this at your convenience

chtaili77:
> مرحبا! أنا روبوت ، وأردت دمج العلاقات العامة الخاصة بك تلقائيًا ، لكن لم أستطع بسبب المشكلة (المشكلات) التالية:
> 
> * يتطلب EIP 747 موافقة أحد ( danfinlay ، estebanmino )

3esmit:
Would make sense to add here the preferred method of using decentralized storage for image, as suggested by @danfinlay  `image: `cid:${CID}`. 

> I totally agree that a CID is better, so if you want to go that path, my request would be that Status implements a version that can be compatible with the current http-based method that MetaMask is using.
> 
> For example, maybe just a type prefix for the `image:string` parameter:
> ```javascript
> ethereum.request({
>     method: 'wallet_watchAsset',
>     params: {
>       type: 'ERC20', // Initially only supports ERC20, but eventually more!
>       options: {
>         address: tokenAddress, // The address that the token is at.
>         symbol: tokenSymbol, // A ticker symbol or shorthand, up to 5 chars.
>         decimals: tokenDecimals, // The number of decimals in the token
>         image: `cid:${CID}`, // By adding a type prefix, we can differentiate this type of image from a URL, and support both while migrating and suggesting the more secure approach.
>       },
>     },
>   });
> ```
> 
> This would allow you to only support `cid:` if you want, and MetaMask can then introduce compatible support with yours and recommend that approach instead.

_Originally posted by @danfinlay in https://github.com/status-im/status-react/issues/10036#issuecomment-715603867_

danfinlay:
I've also added an additional proposed change to EIP-747, doesn't need to be made here, but could be: https://github.com/status-im/status-react/issues/10036#issuecomment-716679899

danfinlay (approved):
Looks good, thank you!

PR #2780, opened 2020-07-11
Intial draft of eip to reduce intrinsic gas cost of transactions

lightclient:
rendered: https://github.com/lightclient/EIPs/blob/reduce-intrinsic-cost/EIPS/eip-2780.md

### DIFF
--- a/EIPS/eip-2780.md
+++ b/EIPS/eip-2780.md
@@ -1,5 +1,5 @@
 ---
-eip: <to be assigned>
+eip: 2780
 title: Reduce intrinsic transaction gas
 author: Matt Garnett (@lightclient)>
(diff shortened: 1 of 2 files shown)


ritzdorf:
Hi @lightclient,

I have a few questions.

1. What about the cost for hashing the transaction?
2. Should the inclusion into the merkle tree of the block be considered? This seems to be a per-transaction cost. (And there are probably more like that.)
3. What should happen in case the account is uninitialized?
4. Costs for SLOAD are expected to rise in the future (see EIP-1884). Would you then adjust the intrinsic cost also?
5. Do you assume that the SLOAD and SSTORE operations cover balance and nonce?

PR #2782, opened 2020-07-12
Update ERC725 with new format and sub standards ERC725X and ERC725Y

frozeman:
This updates ERC725 to the latest version, which separates the standard into two sub standards: `ERC725X` and `ERC725Y`

### DIFF
--- a/EIPS/eip-725.md
+++ b/EIPS/eip-725.md
@@ -1,5 +1,5 @@
 ```
-eip: <to be assigned>
+eip: 725
 title: ERC-725 Smart Contract Based Account
 author: Fabian Vogelsteller <fabian@lukso.network>, Tyler Yasaka (@tyleryasaka)
 discussions-to: https://github.com/ethereum/EIPs/issues/725
--- a/EIPS/eip-725.md
+++ b/EIPS/eip-725.md
@@ -1,55 +1,93 @@
----
-eip: 725
-title: Proxy Account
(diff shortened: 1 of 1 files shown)


EIPS/eip-725.md
```diff
@@ -1,55 +1,93 @@
----
```
lightclient, line 1: Please note that it is required for the preamble to start and end with `---` rather than ` ``` `.

EIPS/eip-725.md
```diff
@@ -1,55 +1,93 @@
...
 eip: 725
-title: Proxy Account
-author: Fabian Vogelsteller (@frozeman), Tyler Yasaka (@tyleryasaka)
+title: ERC-725 Smart Contract Based Account
+author: Fabian Vogelsteller <fabian@lukso.network>, Tyler Yasaka (@tyleryasaka)
 discussions-to: https://github.com/ethereum/EIPs/issues/725
 status: Draft
 type: Standards Track
 category: ERC
+requires: ERC165, ERC173, ERC1271 (optional)
```
lightclient, line 9: Please list ERCs and EIPs using only their number. E.g. `requires: 165, 173`. It isn't clear via EIP-1 how to denote optional EIPs, but considering it doesn't appear to be core to this EIP I recommend omitting it from the list of required EIPs.

axic:
I wonder what "sub standards" mean and whether it would make sense to actually separate them into their own EIPs?

PR #1337, opened 2018-08-18
EIP 1337 -- Subscriptions on the blockchain.

owocki:
# What

This is a PR that contains the Pending standard for recurring payments on the blockchain (subscriptions) coming from the [ERC 948 Working Group](https://github.com/ethereum/EIPs/issues/948)

# Why 

Monthly subscriptions are a key monetization channel for legacy web, and arguably they are the most healthy monetization channel for businesses on the legacy web (especially when compared to ad/surveillance) based models. They are arguably more healthy than a token based economic system (depending upon the vesting model of the ICO) because

For a user:

* you don't have to read a complex whitepaper to use a dapps utility (as opposed to utility tokens)
* you don't have to understand the founder's vesting schedules
* you can cancel anytime

For a dapp founder:

* since you know your subscriber numbers, churn numbers, conversion rate, you get consistent cash flow
* you get to focus on making your customers happy (as opposed to having two actors: speculators & users)

For these reasons, we think it's worth creating a standard way to do 'subscriptions' on Ethereum.

# Abstract
A standardized interface for recurring subscriptions on the blockchain.

# How / Logistics

This PR is not ready to be merged yet.  It will be updated as the working group comes to consensus on the EIP standard.

# Further Reading 
* [ERC 948](https://github.com/ethereum/EIPs/issues/948)
* [Architecture's we are considering](https://medium.com/gitcoin/technical-deep-dive-architecture-choices-for-subscriptions-on-the-blockchain-erc948-5fae89cabc7a)
* [Github Account for the Working Group](https://github.com/EthereumOpenSubscriptions/standard)
* We have weekly calls where we discuss and get aligned.  Go to the [Gitcoin Slack](https://gitcoin.co/slack) and join <i>#proj-subscriptions</i> to join the chat / ask for a calendar invite.

PR #2789, opened 2020-07-15
Add EIP-1337 draft

lightclient:
Per [this comment](https://github.com/ethereum/EIPs/pull/1337#issuecomment-506554727) and the tweet that recently put EIP-1337 [back under the spotlight](https://twitter.com/balajis/status/1283404484631117824), it would be nice for this to be merged as a draft. [Rendered](https://github.com/ethereum/EIPs/blob/b43e387c1b0e74c14913a16b271c738893c2e9c9/EIPS/eip-1337.md).


lightclient:
Stuck again.

Aeroxander:
Is there any time frame when this moves to the Final stage?

lightclient:
That's up to @owocki et al. If they're comfortable with where the standard is at, they are free to move it forward to last call. Two weeks after that, it will be eligible to be considered final.

owocki:
given that gitcoin grants has moved from subscriptions being a key feature to quadratic funding being the key feature, (+ we've actually removed subscription support bc it wasnt really used much) i dont see myself allocating more cycles to this.  

when we wrote the draft EIP, there was a whole working group (@androolloyd was my partner in crime + like 6 other ppl), but that has since stopped being active too.

sorry i know thats not the answer you were looking to hear, but im just trying to be deliberate about setting expectations.

would be happy to pass the baton to someone new!

davidtai:
Hello @owocki,

I have a team that is interested in subscription payments.  I would love to have a chat with picking up the baton since this is so close to the finish line.

MicahZoltu:
@davidtai Please use the discussions-to link in the EIP for further discussion.

PR #1605, opened 2018-11-20
Create contract ABI encoding draft 

axic:

PR #2792, opened 2020-07-17
Adapted ABIv2 to EIP

fubuloubu:
Moved:
- All examples to "Test Cases" section

Removed:
- Links that don't link anywhere (Markdown makes these automatically for sections)
- "Mapping Solidity to ABI Types" Section (specific to Solidity)
- "Strict Encoding Mode" Section (specific to Solidity)
- "Non-standard Packed Mode" Section (out of scope for EIP)


EIPS/eip-xxx.md
```diff
@@ -0,0 +1,779 @@
+---
+eip: XXX
```
fubuloubu, line 2:
```suggestion
eip: 2792
```

EIPS/eip-2792.md
```diff
@@ -0,0 +1,779 @@
+---
+eip: 2792
+title: ABI Specification, Version 2
+author: Alex Beregszaszi (@axic), Christian Reitwiessner <chris@ethereum.org>;
+discussions-to: <this PR>
```
fubuloubu, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/abi-specification/4429
```

MicahZoltu:
### As an Editor
Consider breaking this up into multiple EIPs.  Perhaps one for the ABI JSON definition, one for the function selector algorithm, and one for parameter encoding.  This isn't critical, but it will allow things like the function selector algorithm to remain fixed over time (I think it has changed once in its history, when tuples were added) while the parameter encoding may grow at a higher rate.  Separating out the algorithm for calculating the function selector from the list of valid parameter types may also be valuable.  Keep in mind that EIPs are immutable, so even just adding a single new type (e.g., bytes33) would require an entire new version of the specification, or at least a new EIP that adds to the previous.

The only problem I see with this PR as it stands is that @fubuloubu set the authors to two other people.  I would say put yourself down as the author, and then if @axic or @chriseth indicate they want to be added as authors then we can add them.  I'm generally uncomfortable with assigning EIP authorship to people without their agreement.  Once that is addressed I'll merge this as draft as it appears to be structurally sound other than potentially the authors issue.

### As a Contract Developer
I mentioned this in the Gitter discussion and wanted to bring it up here just to have it "on record".  While I appreciate the desire to have an immutable specification to reference, additional bureaucracy added to the ABI change process will increase the difficulty in iterating on the language.  Solidity is still pretty young, so I'm hesitant to support an increase in bureaucracy to solve a problem that I haven't witnessed being particularly impactful to developers.  I think that there will come a time in Solidity's life cycle where having a change control process on the ABI will be valuable, I just don't think we are there yet so I would *rather* this not be turned into an EIP **yet**.

chriseth:
Thanks for doing this! @axic proposed to make this an EIP a long time ago, but we never got around to doing this.

I'll take a deeper look next week, just two comments for now:

"v2" is just the version of Solidity's implementation, I would consider the standard still version 1 (or strictly, the padding and the function selector were both introduced at some point, so it could be version 3, but I guess nobody cares about those old versions).

I assume the fixed point types are specified as implemented in Vyper - is that correct?

Splitting up this EIP is probably a good idea - I would not extract the function selector calculation, but the json abi description is probably a good candidate to put in its own EIP.

Concerning slowing Solidity down by making this an EIP: My personal impression is that Solidity can never make a breaking change to the ABI because it would mean that the calling semantics silently change. What could be slowed down is introducing new types, but we are very reluctant to do this anyway because of the aforementioned reason, so I personally do not see this as a big problem.

fubuloubu:
@chriseth yes, we implement `fixed168x10` for our `decimal` type, but would probably look to expand our use of those series of types in the future. They work just fine with web3py, and I think web3js too (although we don't test with that one)

fubuloubu:
The other comments:
- add myself as author? If @axic and @chriseth don't have a problem, then sure
- split up the EIP? I agree with @chriseth, the selector and parameter encoding live better together, the JSON format can be separate
- make it V1 instead. Yes, this makes sense because it's never been formalized before.

EIPS/eip-2792.md
```diff
@@ -0,0 +1,779 @@
+---
+eip: 2792
+title: ABI Specification, Version 2
```
fubuloubu, line 3:
```suggestion
title: ABI Specification
```

EIPS/eip-2792.md
```diff
@@ -0,0 +1,779 @@
...
+author: Alex Beregszaszi (@axic), Christian Reitwiessner <chris@ethereum.org>;
```
fubuloubu, line 4:
```suggestion
author: Alex Beregszaszi (@axic), Christian Reitwiessner <chris@ethereum.org>, Bryant Eisenbach (@fubuloubu);
```

MicahZoltu:
@chriseth @axic Are you both OK being listed as author on this?  If so we can merge it as a draft.

axic:
Here's my earlier PR but nobody reviewed it: #1605

fubuloubu:
@axic lol, I tried looking to see if there was one! Yes, we should probably adopt the licensing, but are the docs licensed the same as the software? It's not recommended that they be: https://www.gnu.org/licenses/gpl-faq.en.html#GPLOtherThanSoftware

axic:
It may be recommended to use something else, but in the case of the Solidity tree everything is under the terms of GPL-3.0 with the exception of some clearly marked source code.

MicahZoltu:
Anything other than CC0 will require winning a holy war.  I recommend figuring out a way to keep this CC0.

axic:
> Anything other than CC0 will require winning a holy war. I recommend figuring out a way to keep this CC0.

That means acquiring permission from everything who contributed to it in the Solidity repository. Especially as CC0 is practically "public domain", one cannot claim it is a compatible license to GPL-3.0 :)

MicahZoltu:
One could argue that extracting a non-patented specification out of the codebase would qualify as fair use, especially given how small and insignificant it is compared to the Solidity codebase overall.  Also, I think you would only need to get permission from those who have contributed to those docs specifically, and even then I bet you could easily defend "only those who have contributed in a meaningful way" (e.g., not just typo corrections).  I suspect that set is starting to get pretty small.

axic:
Since Solidity is GPL-3.0 it should avoid patent encumberment (if my understanding is correct, but I must highlight I lack any training in law), as such it should be possible to create a brand new specification based on understanding the source code. In this case one should not need to ask anyone for permission.

However taking the currently written specification is already licensed under GPL-3.0. I did mean only getting permission from those who contributed to the file itself. ("everyone who contributed to **it** in the Solidity repository")

It is still couple of dozen people on the list and my experience with getting permissions for changing the license shows that there is always at least a single person who opposes and at least a single person who can not be reached through any means, and in many cases that turns out to be the case once bulk of the work is done. I have done two relicensings myself and taken part of several (mostly granting permission myself).

gnidan:
@fubuloubu to follow here from [our discussion on Twitter](https://twitter.com/fubuloubu/status/1284177658616717313) - you suggested making the JSON specification be a separate EIP. Looking at it more, I'm not sure I agree - it seems to fit here just fine.

But, what are others' thoughts on separating the binary spec from the JSON spec?

Also, Truffle's been maintaining [this formal ABI JSON spec](https://github.com/trufflesuite/truffle/blob/develop/packages/contract-schema/spec/abi.spec.json), and @fubuloubu suggested that it could serve as reference implementation. What's EIP best practice nowadays? Use natural language specification and treat formal schemas as implementations of the natural language?

fubuloubu:
Splitting the spec in this way was @chriseth's idea

gnidan:
> Splitting the spec in this way was @chriseth's idea

Oops, my bad.

Anyway, thinking about it some more, and I'm more keen to support separate specs so we can version them separately. I'll start looking at getting another PR for the JSON spec this week.

fubuloubu:
> I'm more keen to support separate specs so we can version them separately

Yeah, I am very keen to do the same. The ABI Encoding Spec (this EIP) and the JSON Formatting Spec definitely should live separately.

gnidan:
So what's the path forward, re: license issues?

MicahZoltu:
Options (each of which I'm sure someone will think is an absolutely terrible idea and outright veto given opportunity):
A. Ignore the fact that this was copied and merge it with CC0.
B. Ignore the EIP rules that everything must be public domain and merge it with GPL.
C. Relicense the original to something compatible with public domain and merge this with CC0.
D. Rewrite the EIP without using the language copied from GPL documentation.
E. Convince EIP editors that documentation falls under acceptable use policies in sufficient number of jurisdictions and thus GPL doesn't apply and merge this as CC0.
F. Leave the ABI specification in the Solidity docs and close this EIP.

tjayrush:
Just to extend item D. above. We could pull together some cash for a GitCoin bounty to get a professional technical writer to re-write the document. I'd be willing to put up $500.00 US. I've spent an ungodly amount of time trying to parse that nearly unreadable document, so it would be of great use to me (although I do admit that it seems much better now than it used to two years ago).

gnidan:
Regarding item C., [the ABI spec documentation source](https://github.com/ethereum/solidity/blob/develop/docs/abi-spec.rst) currently has 34 contributors. It might be hard to get permission from each of them to relicense.

fubuloubu:
Yet another reason why this Specification deserves more attention, making it a good candidate for an EIP: https://github.com/ethereum/web3.py/issues/1634#issuecomment-663779883

PR #2803, opened 2020-07-19
Creates EIP-2803: Rich Transactions

MicahZoltu:
Support 'rich transactions' by allowing transactions from externally owned accounts to execute bytecode directly.

### DIFF
--- a/eip-2801.md
+++ b/eip-2801.md
@@ -1,0 +1,58 @@
+---
+eip: 2801
+title: Rich Transactions
+author: Micah Zoltu (@MicahZoltu)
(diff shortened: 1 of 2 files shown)


Neurone:
I think there's an error, this EIP file is in the root and not inside `EIPS` folder

k06a:
Hi, we can generalize this EIP to extend CREATE functionality. Because it already executes calldata as EVM bytecode but then charge 32k gas for contract creation + 200 gas for every byte. Let's disable charging 32k of gas for the cases when returned smart contract size is 0.

PR #2805, opened 2020-07-19
Update eip-1538.md

mudgen:
Fixed link to EIP-2535.

### DIFF
--- a/EIPS/eip-1538.md
+++ b/EIPS/eip-1538.md
@@ -9,7 +9,7 @@ category: ERC
 created: 2018-10-31
 ---
 
-Replaced by [EIP-2535 Diamond Standard](https://eips.ethereum.org/EIPS/eip-2535).
+Replaced by [EIP-2535 Diamond Standard](./eip-2535.md).
 
 ## Simple Summary
 <!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
--- a/EIPS/eip-1538.md
+++ b/EIPS/eip-1538.md
@@ -9,7 +9,7 @@ category: ERC
 created: 2018-10-31
 ---
 
-Replaced by [EIP-2535 Diamond Standard](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2535.md).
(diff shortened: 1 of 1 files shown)


EIPS/eip-1538.md
```diff
@@ -9,7 +9,7 @@ category: ERC
 created: 2018-10-31
 ---
 
-Replaced by [EIP-2535 Diamond Standard](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2535.md).
+Replaced by [EIP-2535 Diamond Standard](https://eips.ethereum.org/EIPS/eip-2535).
```
MicahZoltu, line 12:
```suggestion
[EIP-2535 Diamond Standard](./eip-2535)
```

mudgen:
Good idea, thanks @MicahZoltu.  It is done.

EIPS/eip-1538.md
```diff
@@ -9,7 +9,7 @@ category: ERC
 created: 2018-10-31
 ---
 
-Replaced by [EIP-2535 Diamond Standard](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2535.md).
+Replaced by [EIP-2535 Diamond Standard](./eip-2535.md).
```
lightclient, line 12: Why is this in the body of the EIP rather than as the [preamble field](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md#eip-header-preamble) `superceded-by`?

mudgen: Because EIP-2535 isn't final yet.   But I would be happy to make this change if it is okay.

lightclient: Ah, I see. I suppose this is okay for now then. Looking over EIP-1 now and I don't believe it covers this sort of situation very well.

MicahZoltu: Also, there is talk of removing the superceded by concept, since deciding what supercedes what is a governance nightmare.  Take for example the 15 different token standards that all claim to supersede ERC-20.

lightclient: I'm in favor of removing it.

PR #2812, opened 2020-07-21
EIP-1559 Improves clarity on the new transaction format.

MicahZoltu:


eip-automerger:

KaycoinsInc:
Very good, thanks.

i-norden:
Looks good, should we go ahead and get rid of that 10 mil reference here?

MicahZoltu:
@i-norden Yes we should fix the 10 million thing, but in a separate PR I think.  I don't know when I or someone else will get around to fixing the 10 million thing and I don't want to hold this PR up waiting for that.

PR #2814, opened 2020-07-21
Switch JSON-LD from Article to TechArticle

fulldecent:


MicahZoltu:
Would be nice to address the other issues with missing JSON-LD fields (the ones we can) in this PR, though I am OK with this PR as is since it is an incremental improvement.

fulldecent:
Thank you. I'm all about the increment.

Added separate item here to discuss for that: https://github.com/ethereum/EIPs/pull/2816

I could do either way on that PR. But at least there is a PR that can merge in one click if needed.

PR #2796, opened 2020-07-17
Add JSON-LD for EIPS

fulldecent:
Extracted from PR at https://github.com/ethereum/EIPs/pull/2738

PR #2816, opened 2020-07-21
Fill in extra JSON-LD fields

fulldecent:
This fills in extra JSON-LD fields, addressing Google errors raised here:

https://github.com/ethereum/EIPs/pull/2796#issuecomment-661912226

---

All this information is duplicative to other properties in the JSON-LD already. And I don't see that these fields are *required* by Schema.org. But the Google validator flagged them, so perhaps it is worth merging this.

### DIFF
--- a/_layouts/eip.html
+++ b/_layouts/eip.html
@@ -50,9 +50,11 @@ https://schema.org/TechArticle
   {
     "@context": "http://schema.org",
     "@type": "TechArticle",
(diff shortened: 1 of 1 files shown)


MicahZoltu:
While the *values* may be duplicated, the *keys* are not.  Some data aggregator may filter or sort by `datePublished` for example and if you are missing a `datePublished` field then the aggregator may fail to properly filter/sort, even though the `dateCreated` is present.

PR #2817, opened 2020-07-22
Trying to fix 2-EIP-mixed problem

zhous:
So EIP-2794 will be changed to EIP-2787.
Oh no, I guess this is a wrong operation.


MicahZoltu:
This is definitely not what you want.  This changes the eip-template itself.

MicahZoltu:
I think the changes you were trying to apply have since been applied, so I believe you can close this PR now.

PR #2821, opened 2020-07-22
Update eip-2569.md

zhous:
OMG, is this time succeeded to make it seperate?

### DIFF
--- a/EIPS/eip-2569.md
+++ b/EIPS/eip-2569.md
@@ -92,6 +92,8 @@ There are no backward compatibility issues.
 ## Test Cases
 [DHonor](https://dhonor.io/)
 
+[1155](http://dhonor.io/1155)
+
 [DeMe](https://www.deme.app/)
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-2569.md
```diff
@@ -92,6 +92,8 @@ There are no backward compatibility issues.
 ## Test Cases
 [DHonor](https://dhonor.io/)
 
+[1155](http://dhonor.io/1155)
+
 [DeMe](https://www.deme.app/)
```
MicahZoltu, line 97: Test cases should be included directly in the specification, rather than links out to external documents.  Ideally, they are in some general purpose format like JSON or YML or just markdown with something like input => expected_output mapping.

whtyfhas: I removed this "Test Cases" section and appended a code section in "Specification" to show how to display an SVG image whose source code is saved on Ethereum

PR #2826, opened 2020-07-25
EIP-234: Moves to Last Call

MicahZoltu:
This has been implemented and live in multiple clients for quite some time now, I just lost track of it.


eip-automerger:

 - Trying to change EIP 234 state from Draft to Last Call

PR #2829, opened 2020-07-25
Mark EIP-2159 Common Metrics as final

ajsutton:
Add a note to EIP-2159 about the potential for existing metrics with the same names but different meanings and move it to final.


eip-automerger:

 - Trying to change EIP 2159 state from Last Call to Final

PR #1681, opened 2019-01-08
add temporal replay protection

holiman:
This PR adds an EIP about temporal replay protection, basically `valid-until` in the form of a timestamp to transactions.

PR #2831, opened 2020-07-26
EIP-2831 Transaction Replacement Message Type

GregTheGreek:

 - The build passes.

### DIFF
--- a/EIPS/eip-2831.md
+++ b/EIPS/eip-2831.md
@@ -12,7 +12,7 @@ requires: 1193
 
 ## Summary
 
-An extension to the JavaScript Ethereum Proider API ([EIP-1193](https://eips.ethereum.org/EIPS/eip-1193)) this creates a new message type in the event a transaction replacement occurs.
+An extension to the JavaScript Ethereum Proider API ([EIP-1193](./eip-1193.md)) this creates a new message type in the event a transaction replacement occurs.
 
 ## Abstract
 
--- a/EIPS/eip-2831.md
+++ b/EIPS/eip-2831.md
@@ -16,6 +16,10 @@ An extension to the JavaScript Ethereum Proider API ([EIP-1193](./eip-1193.md)) 
 
 ## Abstract
 
+
+
+## Motivation
+
 Exert from EIP-1193
 > A common convention in the Ethereum web application ("dapp") ecosystem is for key management software ("wallets") to expose their API via a JavaScript object in the web page.
 This object is called "the Provider".
@@ -26,13 +30,9 @@ Transaction replacement is not a new concept, but unfortunately causes major use
 
 This EIP formalizes a way for both providers and dapp developers to track transaction replacements seamlessly.
 
-This specification does not cover how one would utilize the transaction replacement message.
-
 ## Specification
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
-
-> Comments like this are non-normative.
 
 ### Definitions
 
@@ -40,18 +40,14 @@ _This section is non-normative._
 
 - Provider
   - A JavaScript object made available to a consumer, that provides access to Ethereum by means of a Client.
-- Client
-  - An endpoint that receives Remote Procedure Call (RPC) requests from the Provider, and returns their results.
 - Wallet
   - An end-user application that manages private keys, performs signing operations, and acts as a middleware between the Provider and the Client.
-- Remote Procedure Call (RPC)
-  - A Remote Procedure Call (RPC), is any request submitted to a Provider for some procedure that is to be processed by a Provider, its Wallet, or its Client.
 - Transaction Replacement
   - Submitting a transaction with both: the same nonce and a 10% increase in the gas price, of a previous transaction which a user no longer wishes to send. This must occur before the original transaction is included in the blockchain.
 
 ### Interface
 
-The `data` parameter contains two values: `oldTx` and `newTx`,the transaction hash that's being superseded and the transaction hash of the successor, respectively. 
+The `data` parameter contains two values: `oldTx` and `newTx`,the transaction hash that's being superseded and the transaction hash of the successor, respectively. Both `oldTx` and `newTx` should be encoded hex numbers as strings with the `0x` prefixed.
 
 ```JavaScript
 interface TxReplacement extends ProviderMessage {
@@ -65,7 +61,7 @@ interface TxReplacement extends ProviderMessage {
 
 ## Rationale
 
-The purpose of `tx_replacement` is to provide a user experience improvement to both wallet users and dapp developers. It can be implemented without breaking changes and leverages the existing framework set out by EIP-1193.
+The implementation was chosen to help the ease of implementation for both providers and dapp developers. Since `ProviderMessage` is widely used by dapp developers already it means that the implementation path would be as trivial as adding and additional `if` clause to their existing message listener. This also provides a benefit to dapps in the event that a provider has not yet implemented the `tx_replacement` and will not cause the dapp panic with `undefined` should it be implemented natively (eg: `ethereum.txReplacement(txHash, cb())` which would error with `ethereum.txReplacement()` is not a function).
 
 ## Backwards Compatibility
 
@@ -74,6 +70,10 @@ Many Providers adopted EIP-1193, as this EIP extends the arbitrary `message` eve
 ## Implementations
 
 At the time of writing, no projects have working implementations.
+
+## Security Considerations
+
+None at the current time.
 
 ## References
 
@@ -84,10 +84,7 @@ At the time of writing, no projects have working implementations.
 
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
 
-## Appendix I: Consumer-Facing API Documentation
-
-
-## Appendix II: Examples
+## Appendix I: Examples
 
 These examples assume a web browser environment.
 
--- a/EIPS/eip-2831.md
+++ b/EIPS/eip-2831.md
@@ -16,7 +16,7 @@ An extension to the JavaScript Ethereum Proider API ([EIP-1193](./eip-1193.md)) 
 
 ## Abstract
 
-
+The current communication between providers and consumers of providers are fundamentally broken in the event that a transaction in the mempool has been superseded by a newer transactions. Providers currently have no way of communicating a transaction replacement, and consumers are required to poll block by block for the resulting transaction.
 
 ## Motivation
 
--- a/EIPS/eip-2831.md
+++ b/EIPS/eip-2831.md
@@ -1,0 +1,121 @@
+---
+eip: 2831
+title: Transaction Replacement Message Type
+author: Gregory Markou (@GregTheGreek)
+discussions-to: [eth magicians](https://ethereum-magicians.org/t/eip-2831-transaction-replacement-message-type/4448)
+status: Draft
+type: Standards Track
+category: Interface
+created: 2020-07-26
+requires: 1193
+---
+
+## Summary
+
+An extension to the JavaScript Ethereum Proider API ([EIP-1193](https://eips.ethereum.org/EIPS/eip-1193)) this creates a new message type in the event a transaction replacement occurs.
+
+## Abstract
+
+Exert from EIP-1193
+> A common convention in the Ethereum web application ("dapp") ecosystem is for key management software ("wallets") to expose their API via a JavaScript object in the web page.
+This object is called "the Provider".
+
+Many ingenious developments have been made by wallet developers to improve the overall user experience while interacting with the Ethereum blockchain. One specific innovation was the tx replacement, offering users the ability to effectively cancel a previously sent transaction.
+
+Transaction replacement is not a new concept, but unfortunately causes major user experience problems for dapp developers as the replaced transaction is near impossible to track. 
+
+This EIP formalizes a way for both providers and dapp developers to track transaction replacements seamlessly.
+
+This specification does not cover how one would utilize the transaction replacement message.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+> Comments like this are non-normative.
+
+### Definitions
+
+_This section is non-normative._
+
+- Provider
+  - A JavaScript object made available to a consumer, that provides access to Ethereum by means of a Client.
+- Client
+  - An endpoint that receives Remote Procedure Call (RPC) requests from the Provider, and returns their results.
+- Wallet
+  - An end-user application that manages private keys, performs signing operations, and acts as a middleware between the Provider and the Client.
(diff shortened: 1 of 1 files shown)


EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
+---
+eip: 2831
+title: Transaction Replacement Message Type
+author: Gregory Markou (@GregTheGreek)
+discussions-to: [eth magicians](https://ethereum-magicians.org/t/eip-2831-transaction-replacement-message-type/4448)
```
MicahZoltu, line 5: EIPs normally just have a raw link here.  The final EIP renderer may try to pretty this at some point in the future, so I think it is good to keep it as a raw link everywhere.
```suggestion
 discussions-to: https://ethereum-magicians.org/t/eip-2831-transaction-replacement-message-type/4448
```

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+status: Draft
+type: Standards Track
+category: Interface
+created: 2020-07-26
+requires: 1193
+---
+
+## Summary
+
+An extension to the JavaScript Ethereum Proider API ([EIP-1193](https://eips.ethereum.org/EIPS/eip-1193)) this creates a new message type in the event a transaction replacement occurs.
```
MicahZoltu, line 15:
```suggestion
An extension to the JavaScript Ethereum Proider API ([EIP-1193](./eip-1193.md)) this creates a new message type in the event a transaction replacement occurs.
```

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Abstract
+
+Exert from EIP-1193
+> A common convention in the Ethereum web application ("dapp") ecosystem is for key management software ("wallets") to expose their API via a JavaScript object in the web page.
+This object is called "the Provider".
+
+Many ingenious developments have been made by wallet developers to improve the overall user experience while interacting with the Ethereum blockchain. One specific innovation was the tx replacement, offering users the ability to effectively cancel a previously sent transaction.
```
MicahZoltu, line 23:
```suggestion
Many ingenious developments have been made by wallet developers to improve the overall user experience while interacting with the Ethereum blockchain. One specific innovation was transaction replacement, offering users the ability to effectively cancel a previously sent transaction.
```

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+
+Transaction replacement is not a new concept, but unfortunately causes major user experience problems for dapp developers as the replaced transaction is near impossible to track.
```
MicahZoltu, line 25: This all feels like it should be in the `## Motivation` section, which appears to be missing.

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+> A common convention in the Ethereum web application ("dapp") ecosystem is for key management software ("wallets") to expose their API via a JavaScript object in the web page.
+This object is called "the Provider".
+
+Many ingenious developments have been made by wallet developers to improve the overall user experience while interacting with the Ethereum blockchain. One specific innovation was the tx replacement, offering users the ability to effectively cancel a previously sent transaction.
+
+Transaction replacement is not a new concept, but unfortunately causes major user experience problems for dapp developers as the replaced transaction is near impossible to track. 
+
+This EIP formalizes a way for both providers and dapp developers to track transaction replacements seamlessly.
+
+This specification does not cover how one would utilize the transaction replacement message.
```
MicahZoltu, line 29: I recommend adding to the abstract a very brief code description of the event being added, like the TypeScript signature

MicahZoltu, line 29: Consider removing the statement that says what the EIP doesn't do.   The EIP also doesn't make me coffee.  😉

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+### Interface
+
+The `data` parameter contains two values: `oldTx` and `newTx`,the transaction hash that's being superseded and the transaction hash of the successor, respectively. 
+
+```JavaScript
+interface TxReplacement extends ProviderMessage {
+  readonly type: 'tx_replacement';
+  readonly data: {
+    readonly oldTx: string;
+    readonly newTx: string;
```
MicahZoltu, line 61: Recommend defining more clearly how these are encoded.  Is it a number as a string, a hex number as a string, is 0x prefixing mandatory, optional, or disallowed?

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+  };
+}
+```
+
+## Rationale
+
+The purpose of `tx_replacement` is to provide a user experience improvement to both wallet users and dapp developers. It can be implemented without breaking changes and leverages the existing framework set out by EIP-1193.
+
+## Backwards Compatibility
```
MicahZoltu, line 70: Recommend explaining why `ProviderMessage` was used rather than a new event type.  A new event type seems much more appropriate here than throwing it into the generic `ProviderMethod` (which IMO is intended for random non-standard stuff).

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+interface TxReplacement extends ProviderMessage {
+  readonly type: 'tx_replacement';
+  readonly data: {
+    readonly oldTx: string;
+    readonly newTx: string;
+  };
+}
+```
+
+## Rationale
```
MicahZoltu, line 66: Recommend explaining why ProviderMessage was used rather than a new event type. A new event type seems much more appropriate here than throwing it into the generic ProviderMethod (which IMO is intended for random non-standard stuff).

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+
+The purpose of `tx_replacement` is to provide a user experience improvement to both wallet users and dapp developers. It can be implemented without breaking changes and leverages the existing framework set out by EIP-1193.
```
MicahZoltu, line 68: This statement feels like a repeat of the motivation/abstract.  The purpose of the Rationale section is (usually) to describe why you made certain decisions within the specification, like why you decided to use `ProviderMessage` instead of a new message type.

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+_This section is non-normative._
+
+- Provider
+  - A JavaScript object made available to a consumer, that provides access to Ethereum by means of a Client.
+- Client
+  - An endpoint that receives Remote Procedure Call (RPC) requests from the Provider, and returns their results.
+- Wallet
+  - An end-user application that manages private keys, performs signing operations, and acts as a middleware between the Provider and the Client.
+- Remote Procedure Call (RPC)
+  - A Remote Procedure Call (RPC), is any request submitted to a Provider for some procedure that is to be processed by a Provider, its Wallet, or its Client.
```
MicahZoltu, line 48: These definitions are not present in the rest of the specification, so they probably shouldn't be included in the EIP.

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
...
+- Transaction Replacement
+  - Submitting a transaction with both: the same nonce and a 10% increase in the gas price, of a previous transaction which a user no longer wishes to send. This must occur before the original transaction is included in the blockchain.
```
MicahZoltu, line 50: This section header says it is non-normative, but this definitely seems like something that should be normative.

EIPS/eip-2831.md
GregTheGreek: LOOL Love ok :)

I took eip 1193 as a template so it may be outdated for that reason

EIPS/eip-2831.md
```diff
@@ -0,0 +1,121 @@
+---
+eip: 2831
+title: Transaction Replacement Message Type
+author: Gregory Markou (@GregTheGreek)
+ discussions-to: https://ethereum-magicians.org/t/eip-2831-transaction-replacement-message-type/4448
```
lightclient, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-2831-transaction-replacement-message-type/4448
```

JakobP:
I have spent a fair amount of time think about replacement transactions from an user experience prespective as well as how they might be dealt with from a technical perspective. I were unable to come up with a satisfactory technical solution.

I can't speak on the exact implementation suggested in the EIP, but from a frontend development perspective this functionality is very much something I would like to see.

moodysalem:
With meta transactions where multiple relayers are able to submit a transaction (and potentially decentralized infrastructure where any account or contract can submit the transaction), is this change the most generally useful?

I understand this is useful for EOAs but I would prefer prioritizing efforts towards separating the identifier of a transaction from its hash, and making replacement less likely (e.g. deadlines as in https://github.com/ethereum/EIPs/pull/1681 and escalator gas prices as in https://github.com/ethereum/EIPs/pull/2593)

That said I do think this would improve the UX of the Uniswap Interface for many users if implemented by all providers.

lcdupree:
I arrived at this EIP via https://github.com/ethereum/web3.js/issues/3585. In my comment there, I describe how this has been a known issue for us since launching [Burner Machine](https://burnermachine.com) last October, and I describe some of the gymnastics where polling code was written in places where the promise semantics offered by the library are basically broken.

I appreciate this EIP, but agree with @moodysalem, in terms of looking at a more fundamental solution. Coming from the big-tech/enterprise software space, there are a lot of flaws found at design-time or during development that will stop a product from being shipped (show-stoppers), even if they are edge-cases where no one can articulate when, or if, the flaw will actually become a problem in production. Replace-by-fee presents edge-cases one could have reasoned about all the way back to the launch of mainnet. MetaMask shipped its speed-up feature and no one stopped to make sure that there'd be a solution to dApp breakage if it was used. It's only now that gas prices are 10-20x higher than they were a year ago due to DeFi, that people want to look at trying to fix this. And since the methodology in the blockchain space tends to be more about fixing something once it becomes a known problem, fixes tend to look more like bandaids, than design and architecture changes that would ameliorate the problem so it wouldn't have existed in the first place.

I don't know what to advocate here, but I do know that there are a lot of shipped dApps that are relying on convenience aync/promise semantics in web3.js/ethers.js today, and with this EIP, they will still be broken. I know how to write the polling code to work around this problem, but then that means throwing away the convenience features of these libraries anyway. And I also know that the more of this code that I write, the closer I get to manually parsing transactions myself, running them in a local implementation of the EVM, or even pulling them out of the mempool to inspect them directly. That all sounds pretty far gone from where we started; namely having some example javascript code on a page that shows anyone how to build their own Ethereum dApp using a web3 library by writing only a few lines of code...

GregTheGreek:
> With meta transactions where multiple relayers are able to submit a transaction (and potentially decentralized infrastructure where any account or contract can submit the transaction), is this change the most generally useful?
>
> I understand this is useful for EOAs but I would prefer prioritizing efforts towards separating the identifier of a transaction from its hash, and making replacement less likely (e.g. deadlines as in #1681 and escalator gas prices as in #2593)
> 
> That said I do think this would improve the UX of the Uniswap Interface for many users if implemented by all providers.

Right, so as you mentioned this is more useful EOA, and I would argue that #1681 is a great starting point, I'm going to look at creating a protocol level EIP that could compliment this one. Ultimately I believe this issue is still that we have a fair amount of systems using this, workflow - and this could provide immediate benefits.

Thank you for the feedback.

GregTheGreek:
@lcdupree 
> I don't know what to advocate here, but I do know that there are a lot of shipped dApps that are relying on convenience aync/promise semantics in web3.js/ethers.js today, and with this EIP, they will still be broken. I know how to write the polling code to work around this problem, but then that means throwing away the convenience features of these libraries anyway.

Do you mind expanding on this? What is broken? I dont see how this EIP would require a dapp to developer to do any polling, its a subscription from a provider.

Much like in the example I provided, all a dapp would need to do is subscribe to an additional message before requesting a tx. I'm also open to alternative solutions.

moodysalem:
> Right, so as you mentioned this is more useful EOA, and I would argue that #1681 is a great starting point, I'm going to look at creating a protocol level EIP that could compliment this one. Ultimately I believe this issue is still that we have a fair amount of systems using this, workflow - and this could provide immediate benefits.

The other point I would like to elaborate is that the ID of a transaction does not need to be the hash, and probably shouldn't be the hash given the hash can change for the same DAPP initiated action (e.g. due to speed up, cancel)

I would recommend exploring an option where the DAPP specifies the ID of the transaction sent to the ethereum provider, and additional methods for querying the state of the transaction by ID (e.g. latest hash, cancelled, confirmed, gas price) + the ability to request the provider cancels the transaction, for cases where it will always fail eg due to price limits or deadlines encoded in the transaction data

GregTheGreek:
> The other point I would like to elaborate is that the ID of a transaction does not need to be the hash, and probably shouldn't be the hash given the hash can change for the same DAPP initiated action (e.g. due to speed up, cancel)

Right so, i think this case is covered via the specification hence why there is an `old`/`new` tx. Ultimately a provider would need to implement a queue of some sorts to ensure that the transactions are being tracked.

> I would recommend exploring an option where the DAPP specifies the ID of the transaction sent to the ethereum provider, and additional methods for querying the state of the transaction by ID (e.g. latest hash, cancelled, confirmed, gas price) + the ability to request the provider cancels the transaction, for cases where it will always fail eg due to price limits or deadlines encoded in the transaction data

This could be an option for sure. I have a few ideas how this could work.

GregTheGreek:
@MicahZoltu I believe i've correctly updated this eip

MicahZoltu:
@GregTheGreek It is still missing a `## Motivation` and `## Security Considerations` section: https://github.com/ethereum/EIPs/pull/2831#pullrequestreview-455579177

GregTheGreek:
> @GregTheGreek It is still missing a `## Motivation` and `## Security Considerations` section: [#2831 (review)](https://github.com/ethereum/EIPs/pull/2831#pullrequestreview-455579177)

Forgot to push ...

lcdupree:
I want to respond to @GregTheGreek, but is the conversation over here now that the PR received two approvals and was merged? Do I need to move (copy/paste) my comments over to the discussion thread on eth-magicians to have them considered?

@rstormsf, why did you approve without participating in the dialogue?

GregTheGreek:
> I want to respond to @GregTheGreek, but is the conversation over here now that the PR received two approvals and was merged? Do I need to move (copy/paste) my comments over to the discussion thread on eth-magicians to have them considered?
> 
> @rstormsf, why did you approve without participating in the dialogue?

Please move the conversation to the Eth magicians thread. A merged EIP does not mean that it is being enforced, it simply means that the EIP is formatted correctly. This is still in a DRAFT state.

miohtama:
I have built a dApp where the transaction tracking was a problem as well. While tracking transaction is a larger problem and probably will need a complete revisit in the future, I feel that this spec would address some shortcomings in dApp UX in the next 2-3 years. 

Have a hook when a wallet uses to speed up or cancel transaction feature: The current spec look sufficient enough for this minor feature. 

What can still break and might be missing

* The replacement transaction is a different smart contract call what dApp originally intended (possible, though not sure why anyone would do this, so maybe not even worthy of discussion)

* Tracking transactions with wallet specific ids instead of transaction hashes: e.g. the user interface can associate the transaction in the notification menu easily as the same transaction regardless of how many times speed up feature is used for the same transaction

* Requesting transaction speed up (gas top up) from the dApp user interface itself

MicahZoltu:
@miohtama Please keep discussion in the discussion-to link: https://ethereum-magicians.org/t/eip-2831-transaction-replacement-message-type/4448

huahuayu:
@ Gregthegreek hey, how's everything going? Currently, what's the best way to detect replaced tx (cancel / speed up/ replaced)?

htadashi:
ethers.js has implemented EIP- 2831 Signer transaction.wait() style rejections since v5.2.0: https://github.com/ethers-io/ethers.js/issues/1477

PR #2832, opened 2020-07-27
Update eip-173.md, last call

mudgen:
Set status to last call.

### DIFF
--- a/EIPS/eip-173.md
+++ b/EIPS/eip-173.md
@@ -5,7 +5,8 @@ author: Nick Mudge <nick@perfectabstractions.com>, Dan Finlay <dan@danfinlay.com
 discussions-to: https://github.com/ethereum/EIPs/issues/173
 type: Standards Track
 category: ERC
-status: Draft
+status: Last Call
+review-period-end: 2020-08-10
 created: 2018-06-07
 ---
 
--- a/EIPS/eip-173.md
+++ b/EIPS/eip-173.md
@@ -16,7 +16,7 @@ A standard interface for ownership of contracts.
 
 ## Abstract
 
-The following standard allows for the implementation of a standard API for getting the owner address of a contract and transferring contract ownership to a different address.
+This standard defines a standard API for getting the owner address of a contract and transferring contract ownership to a different address.
 
 Key factors influencing the standard: 
 - Keeping the number of functions in the interface to a minimum to prevent contract bloat.
@@ -26,22 +26,21 @@ Key factors influencing the standard:
 
 ## Motivation
 
-Many smart contracts require that they be owned or controlled in some way. For example to withdraw funds or perform administrative actions. It is so common that the contract interface used to handle contract ownership should be standardized to allow compatibility with contracts that manage contracts.
+Many smart contracts require that they be owned or controlled in some way. For example to withdraw funds or perform administrative actions. It is so common that the contract interface used to handle contract ownership should be standardized to allow compatibility with user interfaces and contracts that manage contracts.
 
 Here are some examples of kinds of contracts and applications that can benefit from this standard:
 1. Exchanges that buy/sell/auction ethereum contracts. This is only widely possible if there is a standard for getting the owner of a contract and transferring ownership.
 2. Contract wallets that hold the ownership of contracts and that can transfer the ownership of contracts.
 3. Contract registries. It makes sense for some registries to only allow the owners of contracts to add/remove their contracts. A standard must exist for these contract registries to verify that a contract is being submitted by the owner of it before accepting it.
+4. User interfaces that show and transfer ownership of contracts.
 
 ## Specification
 
 Every ERC-173 compliant contract must implement the `ERC173` interface. Contracts should also implement `ERC165` for the ERC-173 interface.
 
 ```solidity
-pragma solidity ^0.4.24;
 
 /// @title ERC-173 Contract Ownership Standard
-/// @dev See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-173.md
 ///  Note: the ERC-165 identifier for this interface is 0x7f5828d0
 interface ERC173 /* is ERC165 */ {
     /// @dev This emits when ownership of a contract changes.    
@@ -49,7 +48,7 @@ interface ERC173 /* is ERC165 */ {
 
     /// @notice Get the address of the owner    
     /// @return The address of the owner.
-    function owner() view external;
+    function owner() view external returns(address);
 	
     /// @notice Set the address of the new owner of the contract
     /// @dev Set _newOwner to address(0) to renounce any ownership.
@@ -74,7 +73,7 @@ The `transferOwnership(address _newOwner)` function may be implemented as `publi
 
 To renounce any ownership of a contract set `_newOwner` to the zero address: `transferOwnership(address(0))`. If this is done then a contract is no longer owned by anybody.
 
-The OwnershipTransferred event does not have to be emitted when a contract is created.
+The OwnershipTransferred event should be emitted when a contract is created.
 
 ## Rationale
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 173 state from Draft to Last Call

mudgen:
> Hi! I'm a bot, and I wanted to automerge your PR, but couldn't because of the following issue(s):
> 
> * Trying to change EIP 173 state from Draft to Last Call

@MicahZoltu Help with this please.

MicahZoltu:
> The following standard allows for the implementation of a standard API for getting the owner address of a contract and transferring contract ownership to a different address.

This statement is in the middle of the standard, so it really isn't "the following standard" but instead "this standard".

MicahZoltu:
Most of the content currently in the abstract should be moved to the motivation section or the rationale section.  The motivation section is "why does this exist", the rationale section is "why did we make these particular choices" and the abstract is a "what is this".  Explaining why you included only a couple functions is a good candidate for the rationale section.  Maybe keep the "gas efficient" bit in the abstract, since that may be a good "what is this" statement: "A mechanism for establishing ownership and ownership transfer in a gas efficient way".

MicahZoltu:
Consider putting **MUST** and **SHOULD** in bold and capitalized in the specification section, and also consider putting the RFC blurb at the beginning of the specification section:
> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.

MicahZoltu:
> The OwnershipTransferred event does not have to be emitted when a contract is created.

This has caused problems in the past with ERC20 due to the way tools aggregate data.  I recommend changing this to assert that the owner transferred event **MUST** be fired on contract creation if an owner is assigned (anything other than 0 address).

MicahZoltu:
Consider having a separate event for ownership destruction rather than using the same event with a sentinel value.  This will make it a bit easier to find instances of ownership destruction due to the way events are indexed in databases.

MicahZoltu:
> /// @dev See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-173.md

Recommend removing this line as specifications ideally shouldn't have external links in them.

MicahZoltu:
Recommend removing all of the comments from the specification actually, they are not part of the specification.  If you want to include some interfaces that people can copy/paste into their Solidity project I recommend including them in a code block in the Implementations section of the EIP, rather than the specification section.  Ideally, the specification section *ONLY* includes the actual specification (stuff that must be standardized on) and doesn't contain superfluous material.

MicahZoltu:
Security Considerations section is missing and is required for all EIPs.  If there are no known security related issues then you can just say that, but the section must be present.

MicahZoltu:
Nitpick: This EIP has a dozen or two blank lines at the end of the document that should probably be cleaned up.

mudgen:
> > The following standard allows for the implementation of a standard API for getting the owner address of a contract and transferring contract ownership to a different address.
> 
> This statement is in the middle of the standard, so it really isn't "the following standard" but instead "this standard".

Understood. I updated this.

>Consider putting MUST and SHOULD in bold and capitalized in the specification section, and also consider putting the RFC blurb at the beginning of the specification section:

I understand and considered this. I feel that the standard is clear without this.

> > The OwnershipTransferred event does not have to be emitted when a contract is created.
> 
> This has caused problems in the past with ERC20 due to the way tools aggregate data. I recommend changing this to assert that the owner transferred event **MUST** be fired on contract creation if an owner is assigned (anything other than 0 address).

I understand.  I changed the text to say that the OwnershipTransferred event should be emitted.  I don't say must in order to maintain backwards compatibility.

> Consider having a separate event for `OwnershipRenounced` rather than using the same event with a sentinel value. This will make it a bit easier to find instances of ownership destruction due to the way events are index

Thank you, I considered this. I'm not adding this because it is not backwards compatible with existing implementations.  New implementations are free to implement this.

> > /// @dev See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-173.md
> 
> Recommend removing this line as specifications ideally shouldn't have external links in them.

Thank you, I removed it.

> Recommend removing all of the comments from the specification actually, they are not part of the specification. If you want to include some interfaces that people can copy/paste into their Solidity project I recommend including them in a code block in the Implementations section of the EIP, rather than the specification section. Ideally, the specification section _ONLY_ includes the actual specification (stuff that must be standardized on) and doesn't contain superfluous material.
> 
> Note: this includes removing the `pragma solidity` at the top, since that isn't part of the specification. You may also want to consider removing the interfaces and just having a function list instead (though from an organizational standpoint, they may be valuable)

I understand.  I want to keep the comments and interfaces. Good point about the `pragma solidity`; I removed that.

> Nitpick: This EIP has a dozen or two blank lines at the end of the document that should probably be cleaned up.

Thank you, I removed the blank lines.

> Most of the content currently in the abstract should be moved to the motivation section or the rationale section. The motivation section is "why does this exist", the rationale section is "why did we make these particular choices" and the abstract is a "what is this". Explaining why you included only a couple functions is a good candidate for the rationale section. Maybe keep the "gas efficient" bit in the abstract, since that may be a good "what is this" statement: "A mechanism for establishing ownership and ownership transfer in a gas efficient way".

Thank you for these clarifications. I understand that there is some overlap. I feel that the abstract how it is now helps the reader understand the standard clearly.

> Security Considerations section is missing and is required for all EIPs. If there are no known security related issues then you can just say that, but the section must be present.

Understood. I added a Security Considerations section.

MicahZoltu:
I have been trying to get another editor to review this but at the moment they are all busy.  My current reluctance to merge is because currently, I *personally* believe that the sections are not being used correctly and I'm a bit of a stickler for that sort of thing.  I suspect once another editor gets a chance to review they probably will care less (or openly disagree with me) and merge without any changes.

* **Simple Summary**: One-liner "what is this".
* **Abstract**: What this EIP does (human readable).
* **Motivation**: Why this EIP exists.
* **Specification**: Technical (almost computer readable).
* **Rationale**: Why specific decisions were made.

At the moment, you have rationale and motivation stuff in the abstract, and the abstract doesn't (IMO) go into quite enough detail about how the specification works (basically human readable high level overview of the specification section).  Something like this is what I would prefer to see in an abstract:
> This specification defines standard functions for a single-owner contract.  An implementation allows reading the current owner (`owner() returns (address)`) and transferring ownership (`transferOwnership(address newOwner)`) along with standardized events for when ownership is changed (`OwnershipTransferred(address indexed previousOwner, address indexed newOwner)`).

----

Again, I want to stress that these are my personal thoughts and not the thoughts of all of the editors (we don't agree on everything), but I wanted to let you know what was holding *me* back from merging at the moment so you didn't think I had simply forgotten about this.

mudgen:
@MicahZoltu Thank you.  I appreciate your feedback and help.  I think the abstract you wrote is good and I added it.  I made the changes.  Can we merge it?

MicahZoltu:
@mudgen Sadly, at the moment *nothing* is getting merged in this repository as CI is broken across the board.  I'll snooze the email reminder for next week so I don't forget to get this merged once CI is fixed (and continue to snooze until CI is fixed), though feel free to ping me if you see this go green and I'll take a look!

On a personal note, I feel bad that I could get any other editor to merge this previously (before CI broke).  I dislike the idea that I'm currently the gatekeeper for all EIPs and my opinion is the only one that matters.  I liked it better when you just had to get one of `n` editors to agree to your EIP's layout/wording to get merged.

Sadly, `n` seems to be 1 at the moment.  😢

mudgen:
Thanks @MicahZoltu 
I understand. One good editor is better than none.  I appreciate your good work and help.

EIPS/eip-173.md
```diff
@@ -5,7 +5,8 @@ author: Nick Mudge <nick@perfectabstractions.com>, Dan Finlay <dan@danfinlay.com
 discussions-to: https://github.com/ethereum/EIPs/issues/173
 type: Standards Track
 category: ERC
-status: Draft
+status: Last Call
+review-period-end: 2020-08-10
```
MicahZoltu, line 9: Would you mind bumping this up to 2020-09-03?  I know it isn't your fault, but this should really be 2 weeks after merged so people have enough time after automated notification goes out.

PR #2835, opened 2020-07-29
Rename tx.gas -> tx.gasLimit

minaminao:
It clarifies the meaning of the terms :)

Discussion log:
image

### DIFF
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -93,7 +93,7 @@ For the gas price:
 * During transaction execution, for EIP1559 transactions we calculate the cost to the `tx.origin` and the gain to the `block.coinbase` as follows:
   * Set `GASPRICE = min(BASEFEE + tx.GasPremium, tx.fee_cap)`
   * Let `GASUSED` be the gas used during the transaction execution/state transition
-  * The `tx.origin` initially pays `GASPRICE * tx.gas`, and gets refunded `GASPRICE * (tx.gas - GASUSED)`
+  * The `tx.origin` initially pays `GASPRICE * tx.gasLimit`, and gets refunded `GASPRICE * (tx.gasLimit - GASUSED)`
   * The `block.coinbase` gains `(GASPRICE - BASEFEE) * GASUSED`.
     * If `GASPRICE < BASEFEE` (due to the `FEECAP`), this means that the `block.coinbase` _loses_ funds from this operation; in this case, we check that the post-balance is non-negative and throw an exception if it is negative.
 


eip-automerger:

axic:
Why is this EIP using a mixture of EVM opcodes and Solidity notation? Why not use just EVM opcodes?

MicahZoltu:
Is there an opcode for `GAS_LIMIT`?

MicahZoltu:
A quick glance at the yellow paper suggests there isn't an opcode for transaction gas limit.  There is one for gas remaining, and one for block gas limit, but not one for transaction gas limit from what I can tell.

axic:
Solidity can only expose things which have an opcode :wink:

The "aptly" named opcode `GASLIMIT` is the block gas limit, and not transaction gas limit as imagined by many. There is no way to know the transaction gas limit from within the EVM, apart from running `GAS` as the very first instruction.

axic:
Btw that's why [EIP-1803]()https://eips.ethereum.org/EIPS/eip-1803 proposes to rename those opcodes for clarity.

MicahZoltu:
While I agree that using a mix of opcodes, definitions, and solidity is bad, I still think this PR is an iterative improvement over the previous state of things so I stand by my approval of it.

As far as how to improve things further, I think the best option is to just have every term be added to the definitions list in the specification.  In this case, we would need to add a `TRANSACTION_GAS_LIMIT` (or similar name).

axic:
I should have read the EIP carefully before commenting. I agree there's no need to even use EVM opcodes here as this is functionality outside of the EVM. And this also means using Solidity notiation is very confusing to the reader.

I'd agree to use only definitions and avoid EVM/Solidity here.

axic:
As for merging this, I think the actual authors shall approve and merge, and it should not be forcefully merged by an editor. Shall the authors not want to merge this, the disagreement with the notation can still be voiced during the last call process.

MicahZoltu:
> As for merging this, I think the actual authors shall approve and merge, and it should not be forcefully merged by an editor. Shall the authors not want to merge this, the disagreement with the notation can still be voiced during the last call process.

Totally agree, the only reason I approved it is because I have been pretty active with 1559 and I wanted to signal my agreement.  It was an approval from Micah EIP Author not Micah EIP Editor.  😄

AFDudley:
This change moves the EIP terminology away from what's currently in the code, right?

MicahZoltu:
> This change moves the EIP terminology away from what's currently in the code, right?

What code?  Besu and Geth you mean?  I don't know what either are using for internal variable naming, but it feels like internal naming conventions for clients shouldn't play a role in deciding how to best word an EIP?

minaminao:
Sorry to interrupt you. Maybe we should make the following case style changes for consistency: 
- `tx.GasPremium` -> `tx.gasPremium`
- `tx.fee_cap` -> `tx.feeCap`

PR #2839, opened 2020-07-30
2786: Cleanup, edit 'connect' params, add examples

rekmarks:
cc: @MicahZoltu 

This is a draft PR with some suggested changes.

- Lint using `prettier` (added blank lines after headings)
- Remove comment under `Specification` heading
- Change `connect` event parameters from `('connected', { chainId })` to `({ chainId })`
  - Since the event is named `connect`, I don't think the listeners need to be called with the string `connected`. `connect` and `disconnect` are also already in production per 1193, which specifies the single parameter as `{ chainId }`
- Add examples appendix
  - Examples only include `connect` for now, with `disconnect` to be added in a follow-up PR

### DIFF
--- a/EIPS/eip-2786.md
+++ b/EIPS/eip-2786.md
@@ -12,38 +12,60 @@ requires: 2700
 ---
 
 ## Simple Summary
+
 When an Ethereum Provider becomes connected or disconnected, it will emit a `connect`/`disconnect` event.
 
 ## Abstract
+
 The Provider is said to be “connected” when it can service RPC requests to at least one chain.
 The Provider is said to be “disconnected” when it cannot service RPC requests to any chain at all.
 When the Provider switches from a "connected" state to a "disconnected" state, it will emit a `connect` event.
 When the Provider switches from a "disconnected" state to a "connected" state, it will emit a `disconnect` event.
 
 ## Motivation
+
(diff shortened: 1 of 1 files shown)


EIPS/eip-2786.md
```diff
@@ -12,38 +12,60 @@ requires: 2700
...
+
 The Provider is considered `disconnected` when it is unable to service RPC requests to any chain.
+
 ### Events
+
 #### `connect`
+
 The Provider **MUST** emit a `connect` event to all attached [EIP-2700](./eip-2700) listeners if it transitions from a `disconnected` state to a `connected` state.
-All attached listeners **MUST** be called with the parameters `('connected', { chainId })`.
+All attached listeners **MUST** be called with the parameter `{ chainId }`.
```
MicahZoltu, line 47: I don't think this change is correct.  The attached listener is a function that takes two parameters, first being a string second being dependent on the first.  If you were to call an attached listener with just `{ chainId }` that would be incorrect behavior.

MicahZoltu: Bleh, you were right, I was wrong.

rekmarks:
For posterity: I squashed the original commits and force-pushed to trigger the automatic merge. No changes were made to the content.

Issue #2787, opened 2020-07-15
Discussion for EIP-2786: Ethereum Provider Connect/Disconnect Events


MicahZoltu:
When an Ethereum Provider becomes connected or disconnected, it will emit a `connect`/`disconnect` event.

MicahZoltu:
@rekmarks Would be great to get a review of this before last call ends to make sure it aligns with MetaMask's implementation.

MicahZoltu:
I think we should remove the `chainId` from this event.  If a provider supports multiple chains, and it connects to them in parallel, then this event will be fired for a random chain and will not be fired for any other chains.  This makes relying on this event a bad idea, yet it is a trap that I suspect almost every dapp developer will fall into.  Instead, we should have a separate event for `chainConnected` and `chainDisconnected` or we should update this spec such that `connected` and `disconnected` are called once for every chain that a provider connects to/disconnects from and the `chainId` would be included in both `connected` and `disconnected`.

At the moment, I'm struggling to understand how this event is useful.  You could get a `connected` event for `chainId` `1` and then connect to `chainId` `2` but not receive an event for it, then disconnect from `chainId` `1` without receiving an event for it.  As currently designed, I don't think any dapp should subscribe to this event because it simply doesn't give them enough information to do anything useful.

rekmarks:
I believe the original intention of this event was to cover cases when providers become disconnected from their RPC endpoints / whatever APIs they're using to communicate with the blockchain. The `chainId` was tacked on as an afterthought.

Given the points you've raised, I'm not against removing the `chainId` from the spec, and just referring to `chainChanged` for that. Indeed, MetaMask already emits `chainChanged` when a chain is first connected, making a `chainId` on `connect` redundant. Whether we need the event at all is a different question.

There's some previous discussion about this on the [1193 discussion issue](https://github.com/ethereum/EIPs/issues/2319). @alcuadrado [noted that some providers "need" `connect`/`disconnect` methods](https://github.com/ethereum/EIPs/issues/2319#issuecomment-544795458) (previously called `open`/`close`). I weighed in on the matter [here](https://github.com/ethereum/EIPs/issues/2319#issuecomment-561028949), referencing a PR by @pedrouid.

IIRC, the `connect`/`disconnect` events were intended to be paired with the corresponding methods. If the consumers can cause the provider to become connected or disconnected, we can imagine that the provider can become connected or disconnected for reasons beyond the consumer's control. Hence the need for the events for such providers.

[...]

PR #2840, opened 2020-07-30
2786: Add disconnect event to specification

rekmarks:
- Add the `disconnect` event to the specification section of 2786.
- Add `ConnectInfo` interface for `connect` event listener parameter (for consistency with `DisconnectError` interface for `disconnect` event)
- Makes all interface properties `readonly`
- Cleans up some faulty formatting

### DIFF
--- a/EIPS/eip-2786.md
+++ b/EIPS/eip-2786.md
@@ -49,6 +49,22 @@ All attached listeners **MUST** be called with the parameter `{ chainId }`.
 If the Provider supports the `eth_chainId` JSON-RPC method or a derivation of it, then the `chainId` **MUST** match the return value of `eth_chainId`.
 The Provider **MAY** call the attached listeners in any order.
 
+#### `disconnect`
+
+The Provider **MUST** emit a `disconnect` event to all attached [EIP-2700](./eip-2700) listeners if it transitions from a `connected` state to a `disconnected` state.
+All attached listeners **MUST** be called with the parameter `error: DisconnectError`.
+`error` **MUST** adhere to the `ProviderError` interface (see below).
+`error.code` **MUST** be an integer `number`.
+`error.message` **MUST** be a human-readable `string`.
+The Provider **MAY** call the attached listeners in any order.
+
+````ts
+interface DisconnectError extends Error {
+  code: number,
+  message: string,
+  data?: unknown,
+}
+
 ## Rationale
 
 This EIP is mostly a retrospective EIP meaning it codifies an already existing specification so there isn’t a lot of room for improving things such as by having a connect/disconnect event per chain.
@@ -66,6 +82,11 @@ Copyright and related rights waived via [CC0](https://creativecommons.org/public
 ```javascript
 // connect
 provider.on('connect', ({ chainId }) => {
-  console.log(`Provider connected to: ${chainId}`);
+  console.log(`Provider became connected to: ${chainId}`);
 });
-```
+
+// disconnect
+provider.on('disconnect', ({ code, message, data }) => {
+  console.log(`Provider became disconnected due to: ${code}: ${message}`);
+});
+````
--- a/EIPS/eip-2786.md
+++ b/EIPS/eip-2786.md
@@ -96,5 +96,8 @@ provider.on('connect', ({ chainId }) => {
 // disconnect
 provider.on('disconnect', ({ code, message, data }) => {
   console.log(`Provider became disconnected due to: ${code}: ${message}`);
+  if (data) {
+    console.log(`Additional error data:`, data);
+  }
 });
 ```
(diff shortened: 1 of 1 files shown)


EIPS/eip-2786.md
```diff
@@ -12,38 +12,81 @@ requires: 2700
...
 #### `connect`
+
 The Provider **MUST** emit a `connect` event to all attached [EIP-2700](./eip-2700) listeners if it transitions from a `disconnected` state to a `connected` state.
-All attached listeners **MUST** be called with the parameters `('connected', { chainId })`.
+All attached listeners **MUST** be called with the parameter `{ chainId }`.
 `chainId` **MUST** specify the integer ID of the connected chain encoded as a hexadecimal string.
 If the Provider supports the `eth_chainId` JSON-RPC method or a derivation of it, then the `chainId` **MUST** match the return value of `eth_chainId`.
 The Provider **MAY** call the attached listeners in any order.
 
+#### `disconnect`
```
MicahZoltu, line 52: This is why it is good to have someone review your work.  😬  I totally forgot to include disconnect in the connect/disconnect spec!

EIPS/eip-2786.md
```diff
@@ -12,38 +12,81 @@ requires: 2700
...
+`error` **MUST** adhere to the `ProviderError` interface (see below).
+`error.code` **MUST** be an integer `number`.
+`error.message` **MUST** be a human-readable `string`.
+The Provider **MAY** call the attached listeners in any order.
+
+````ts
+interface DisconnectError extends Error {
+  code: number,
+  message: string,
+  data?: unknown,
```
MicahZoltu, line 65: `readonly`?

rekmarks:
8585880
- Rebased following the merging of #2839 
- Extract `connect` listener param into interface
- Made all interface properties `readonly`

axic:
@MicahZoltu I think this is back to you for review.

MicahZoltu:
Ah, I was waiting for this PR to leave draft state before doing a final review and merge.  @rekmarks is this ready for review?

It is worth noting that I have a pretty strong distaste for this EIP as written (by me!) for reasons expressed in https://github.com/ethereum/EIPs/issues/2787#issuecomment-675208764.  As mentioned in that comment a possible path forward for this EIP may be to mark it as withdrawn in which case this PR would probably just get closed.

rekmarks:
Ah, many thanks to the GitHub actions bot. @MicahZoltu I will review it and get back to you tomorrow.

MicahZoltu:
@rekmarks Is it tomorrow yet?  😬 

Also, I still vote for marking this EIP as withdrawn.  I think it is a bad idea for reasons outlined elsewhere (see previous comments).

rekmarks:
Oof, sorry about that.

I agree that this EIP has issues. Before continuing, I'll summarize the needs I believe that we want to address:

1. An event pair to signal when the provider is operating nominally and when it cannot service RPC requests to any chain due to circumstances beyond its control. In other words, when the provider is "broken." This is what the  `connected`/`disconnected` event pair was supposed to address.
2. For providers that are connected to multiple chains at once, an event pair signaling when a specific chain becomes disconnected. This would be addressed by the `chainConnected`/`chainDisconnected` events you have suggested elsewhere.
    - I'll note that those events do not conflict with `chainChanged`, since that event is about which chain is "currently selected" by the provider.

I agree that emitting the `chainId` with the `connected` event is a bad idea.

Now, all that being said, do you agree that `1` is a real need? If so, I believe this EIP can be amended to a more sensible state. Otherwise, withdrawing it seems like the best option.

MicahZoltu:
@rekmarks Do you have an example of a scenario where a provider would be unable to connect to any chain and know it?  I'm one can imagine a provider which can somehow detect when it is not connected to the internet, but even in that case it would still be connected to localhost and would be able to connect to any chain that was local.  I'm struggling to imagine a scenario where a provider is in a situation where no chains are available?

rekmarks:
It's definitely unlikely, but we could imagine a situation where the wallet's Ethereum client service is down. To take MetaMask as an example, that would of course be if Infura were down.

However, that's to my knowledge never happened, and people would certainly figure out what the problem is if it did, with or without a `disconnected` event.

At the end of the day, this would most likely happen if the user's Internet connection is down, which they'll notice anyway. This is all to say, there are edge cases where these events would be useful, but there isn't exactly an urgent need for them.

MicahZoltu:
> At the end of the day, this would most likely happen if the user's Internet connection is down, which they'll notice anyway.

Even in this case though, you still may not be disconnected from all Ethereum networks.  If you have an Ethereum client running on localhost you will *always* be connected to it unless your local computer is off, in which case you certainly won't get a disconnected event.  😉 

In the Infura example you gave, I suppose that is true but it would only apply to providers that cannot multi-connect, and only apply if the provider doesn't support custom RPC endpoints.

I feel like the case here is edgy enough that it isn't worth going out of our way to support (increasing API complexity), especially when the user will receive a bunch of `disconnected from chain` events for each of the chains in a case where the provider is globally disconnected.

rekmarks:
Haha, yeah, the fact that `disconnected from chain` effectively covers the cases this event would cover, is a big argument against increasing API complexity with these events.

I support withdrawing this EIP!

rekmarks:
2786 is withdrawn.

PR #2842, opened 2020-07-30
Update eip-2780.md

uri-bloXroute:

 - The build passes.

### DIFF
--- a/EIPS/eip-2780.md
+++ b/EIPS/eip-2780.md
@@ -15,7 +15,7 @@ Reduce the intrinsic cost of a transaction from `21,000` to `7,000` gas.
 ## Motivation
 The current `21,000` gas intrinsic cost of a transaction makes sending ETH very costly, often times prohibitively costly for small amounts (tens of USD).
 While broad changes to the gas price and first price auction are being considerted in other EIPs (like EIP-1559),
-substantialy reducing the cost of sending ETH and enabling higher volumes of such transactions would be a net positive if done in a safe manner,
+substantially reducing the cost of sending ETH and enabling higher volumes of such transactions would be a net positive if done in a safe manner,
 and without imposing negative externalities, as outlined below.
 
 
--- a/EIPS/eip-2780.md
+++ b/EIPS/eip-2780.md
@@ -1,7 +1,7 @@
 ---
 eip: 2780
 title: Reduce intrinsic transaction gas
-author: Matt Garnett (@lightclient)>
+author: Matt Garnett (@lightclient), Uri Klarman (@uriklarman)
 discussions-to: https://ethereum-magicians.org/t/eip-2780-reduce-intrinsic-cost-of-transactions/4413
 status: Draft
 type: Standards Track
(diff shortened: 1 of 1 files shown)


EIPS/eip-2780.md
```diff
@@ -10,54 +10,107 @@ created: 2020-07-11
...
 ## Abstract
-Reduce the intrinsic cost of a transaction to `16,360` gas.
+Reduce the intrinsic cost of a transaction from `21,000` to `7,000` gas.
 
 ## Motivation
-The current `21,000` gas intrinsic cost of a transaction does not accurately
-represent the true cost of the transaction.
+The current `21,000` gas intrinsic cost of a transaction makes sending ETH very costly, often times prohibitively costly for small amounts (tens of USD).
+While broad changes to the gas price and first price auction are being considerted in other EIPs (like EIP-1559),
+substantialy reducing the cost of sending ETH and enabling higher volumes of such transactions would be a net positive if done in a safe manner,
```
lightclient, line 18:
```suggestion
substantially reducing the cost of sending ETH and enabling higher volumes of such transactions would be a net positive if done in a safe manner,
```

lightclient:
Oops, I missed that there was a spelling error. @uri-bloXroute you'll need to fix that before it will be allow to be merged.

EIPS/eip-2780.md
```diff
@@ -10,54 +10,107 @@ created: 2020-07-11
 ---
 
 ## Abstract
-Reduce the intrinsic cost of a transaction to `16,360` gas.
+Reduce the intrinsic cost of a transaction from `21,000` to `7,000` gas.
```
axic, line 13: How was this value 7000 selected? There does not seem to be an explanation how it was derived, while the previous version was very clear about that.

lightclient: My understanding is that @uri-bloXroute plans to do some tests to determine if lower intrinsic gas costs will increase the uncle rate, and `7,000` is only a starting point. In that case, he is taking an empirical approach -- where as the original draft was an attempt to "deduce" the fair cost a tx based on known resource costs. 

While I don't prefer the strictly empirical approach, I also don't have the time to champion this and want Uri to feel empowered to experiment and gather data. Hopefully the result of that will be beneficial for everyone!

uri-bloXroute: @axic - @lightclient is correct.
I don't believe the approach of calculating the cost of running a Tx on the EVM (under the worst-case assumption that all transactions are sent to a new account) is the right one. Why does it matter how much it would have cost?

My argument is:
- Sending Tx on chain became very expensive due to DeFi
- It would be valuable to allow for as many Tx as possible, but we must consider the negative externalities
- We should enable as low gas cost as possible (I believe we can do under 7,000) as long as we do not  create such externalities (or they are negligible)
- I consider the implications on uncle-rate (to be verified on testnet), state-size, and gas-token. I should probably add measurements of block validation on commodity PCs, to show they are not in danger of being pushed off the network.

Does that rationale make sense to you? I wanted to use this EIP as a starting point for a discussion with the different devs.

PR #2847, opened 2020-08-02
Moves /eip-2803.md to EIPS/eip-2803.md

MicahZoltu:
Total fail on my part.


axic:
One of the reasons for the courtesy that editors don't merge their own new EIPs, these simple things are caught by a second eye :wink:

PR #2853, opened 2020-08-03
1559 Cleanup

MicahZoltu:
I did a lot of EIP cleanup, I'll try to list some of the major changes here:
1. Simplify the simple summary.  This should be one or two sentences stating what this standard does.  The previous simple summary was a bit more wordy than necessary and discussed the standard as a comparison of previous state rather than on its own.
1. Removed the rest of the references to hard coded block gas limits that were missed in their previous removal.
1. Changed `gwei` to `nanoeth` to make this standard more accessible to regular educated folk who are not familiar with Ethereum's idiosyncratic jargon.  (I'm aware this may be contentious and I'm willing to remove it if there is significant pushback).
1. Removed external links in favor of inline data where possible and moved the rest to the Resources section.  Ideally, EIPs stand on their own without external references.
1. Took an initial stab at cleaning up the specification by moving more stuff to the definitions and removing some redundancy between definitions and specification sections.  I plan on doing another PR after this to turn the specification into a Python specification because https://www.commitstrip.com/en/2016/08/25/a-very-comprehensive-and-precise-spec/
1. Tried to bring a little consistency to the naming convention/style of the specification section.
1. I reformatted the security considerations and inlined the arguments rather than linking out to them.  I also removed some that were just people misunderstanding the specification rather than actual security issues.


eip-automerger:

EIPS/eip-1559.md
```diff
@@ -9,130 +9,107 @@ category: Core
...
+* **Inefficiencies of first price auctions**: The current approach, where transaction senders publish a transaction with a bid a maximum fee, miners choose the highest-paying transactions, and everyone pays what they bid. This is well-known in mechanism design literature to be highly inefficient, and so complex fee estimation algorithms are required. But even these algorithms often end up not working very well, leading to frequent fee overpayment.
+* **Instability of blockchains with no block reward**: In the long run, blockchains where there is no issuance (including Bitcoin and Zcash) at present intend to switch to rewarding miners entirely through transaction fees. However, there are known issues with this likely leads to a lot of instability, incentivizing mining "sister blocks" that steal transaction fees, opening up much stronger selfish mining attack vectors, and more. There is at present no good mitigation for this.
```
i-norden, line 28: Small typo needs fixed here "there are known issues with this likely leads to", I can get it on my next PR if you want to merge now.

EIPS/eip-1559.md
```diff
@@ -9,130 +9,107 @@ category: Core
...
   * `FEECAP` serves as the absolute maximum that the transaction sender is willing to pay
 * During transaction execution, for EIP1559 transactions we calculate the cost to the `tx.origin` and the gain to the `block.coinbase` as follows:
-  * Set `GASPRICE = min(BASEFEE + tx.GasPremium, tx.fee_cap)`
+  * Set `GASPRICE = min(BASE_FEE + tx.GasPremium, tx.fee_cap)`
   * Let `GASUSED` be the gas used during the transaction execution/state transition
   * The `tx.origin` initially pays `GASPRICE * tx.gasLimit`, and gets refunded `GASPRICE * (tx.gasLimit - GASUSED)`
-  * The `block.coinbase` gains `(GASPRICE - BASEFEE) * GASUSED`.
-    * If `GASPRICE < BASEFEE` (due to the `FEECAP`), this means that the `block.coinbase` _loses_ funds from this operation; in this case, we check that the post-balance is non-negative and throw an exception if it is negative.
+  * The `block.coinbase` gains `(GASPRICE - BASE_FEE) * GASUSED`.
+    * If `GASPRICE < BASE_FEE` (due to the `FEECAP`), this means that the `block.coinbase` _loses_ funds from this operation; in this case, we check that the post-balance is non-negative and throw an exception if it is negative.
```
i-norden, line 78: This should be removed and instead we assert gas_price >= base_fee

EIPS/eip-1559.md
```diff
@@ -9,130 +9,107 @@ category: Core
...
+* `INITIAL_BASE_FEE` : 0.000000001 ETH (1 nanoeth)
+* `BASE_FEE_CHANGE`: `CURRENT_BLOCK.parent.BASE_FEE`. The amount that the BASE_FEE will increase in `CURRENT_BLOCK` relative to `CURRENT_BLOCK.parent`.
+
+### Process
+* At `block.number == INITIAL_FORK_BLOCK_NUMBER` we set `BASE_FEE = INITIAL_BASE_FEE`
+* `BASE_FEE` is set as follows
   * Let `delta = block.gas_used - TARGET_GASUSED` (possibly negative).
-  * Set `BASEFEE = PARENT_BASEFEE + PARENT_BASEFEE * delta // TARGET_GASUSED // BASEFEE_MAX_CHANGE_DENOMINATOR`
-  * Clamp the resulting `BASEFEE` inside of the allowable bounds if needed, where a valid `BASEFEE` is one such that `abs(BASEFEE - PARENT_BASEFEE) <= max(1, PARENT_BASEFEE // BASEFEE_MAX_CHANGE_DENOMINATOR)`
+  * Set `BASE_FEE = PARENT_BASE_FEE + PARENT_BASE_FEE * delta // TARGET_GASUSED // BASE_FEE_MAX_CHANGE_DENOMINATOR`
```
i-norden, line 66: We should explicitly define `//` as euclidean division somewhere

EIPS/eip-1559.md
MicahZoltu: Fixed.

EIPS/eip-1559.md
MicahZoltu: Hmm, current implementations do not follow this rule?  I'm not a big fan of it so would be happy to delete it, but I left it in because my goal was to not change the spec at all in this PR.

MicahZoltu: Specifically, the specification does *NOT* assert that `fee_cap >= base_fee`.  This was intentional in the original design, but I would support removing it, especially if implementations currently do not implement this per spec.

EIPS/eip-1559.md
MicahZoltu: I have a follow-up PR where I change the entire specification section into fully functional Python, so I'm hesitant to bother trying to define `//` here at the moment.  If my follow-up PR doesn't get merged (e.g., people don't like the change) then I agree, we should define that (I actually had to look it up myself).

EIPS/eip-1559.md
i-norden: I would need to dig up or ask @vbuterin for the specifics but there was a very good reason to make this change- miners can potentially set the coinbase address to someone else's and burn the funds by this mechanism.

The geth implementation does assert that gas_price >= base_fee, and I believe the Besu one does as well as they have been tested to work together (although, perhaps not under those conditions).

EIPS/eip-1559.md
i-norden: Sounds good! Heads up, both the Besu and Geth imps are using euclidean.

EIPS/eip-1559.md
MicahZoltu: We definitely should generate a test case that covers that condition.  😄  Once I get the spec converted to Python, I have a personal task to create some tests (which would pass in the python spec code).  Though, that may come *after* I get 2718 tests generated so could be a while.

MicahZoltu: I would propose that we merge this as-is, and we have a separate PR to add the assertion.  This is primarily because I dislike mixing normative and non-normative changes in a PR, especially when the non-normative changes are as sweeping as the ones here.

i-norden: Great point, let's go ahead and merge this. I'm willing to bet Besu already has that test condition I just don't know for certain.

PR #2859, opened 2020-08-06
EIP-1559: Changes specification to functional Python code

MicahZoltu:
This may be contentious, or it may not be, but I wanted to get a feel for other people's thoughts on just having code for the specification.  My reasoning for this change can be summed up nicely in the following comic, and I was finding myself frustrated trying to make sure that the specification was clear enough while using an arbitrary language that was being made up on the fly:
image

I did my best to try to include comments for all of the important bits, but I'm certainly open to adding more comments for the human readers.  This code should run out-of-the-box in Python 3.6+ and it should type check with MyPy (you'll need to install one pip package).  In a future PR, I would like to include test cases that can actually run against this code (I currently have two that I used just to validate that the code compiled and ran, but their format isn't very portable yet).  These test cases would, presumably, go in the Tests section of the EIP though rather than in the specification.

Please note: I have tried not to apply any normative changes to the EIP in this PR (such as removing the "miner pays fees" branch that has been discussed) as I would like to keep this PR focused purely on how the specification is written, not what it specifies.  However, there were some problems with the spec that had to be addressed before I could write valid code (part of the reasoning why I switched to a code spec instead of an English spec) but the python code **SHOULD** align with the current implementations.


eip-automerger:

gakonst:
I think that while a reference implementation is useful, the text version is useful for non coders

i-norden:
I share @gakonst's sentiment. The Python code adds a ton of helpful details from an implementers perspective, but it also adds some superfluous ones and the additional cognitive overhead is such that I think it will make it more difficult to understand the EIP for people not familiar with Python.

MicahZoltu:
> what about both?

My concern with what about both is that the "human readable" spec is very likely to be incorrect, as we saw with the previous version where I didn't even realize it was incorrect until after I tried to implement it in Python, and even then it turns out that current implementations didn't even match the spec (further implying that it was hard to follow).

Do you all think that perhaps if the abstract were expanded to give more detail it would be useful?  I do think there is value in getting the *gist* across in the abstract and right now the abstract is a very high level overview.  Perhaps if the abstract went into a bit more technical detail everyone could be happy with the outcome?

> it takes a bit more work to dig out the terms of the spec from within the Python code

This suggests to me that I have failed with comments or coding style.  I would *hope* that the python spec would be easier for a developer to understand/read/follow than the "human readable spec" simply because it follows a clear logical flow, unlike the human readable version which is hard to do that with without turning it into code.

Would it perhaps help if some of the "boilerplate" was hidden away and the core of the spec was present a bit separate?  e.g., hide away the block and transaction definitions rather than them taking up a bunch of space at the top?

i-norden:
The implementations don't match the spec not because it was hard to follow the spec but because we realized there was an issue with the spec, and we didn't update it. Matching to the specs is as simple as adding the assertion that `gas_price >= base_fee`, removing the `If `GASPRICE < BASE_FEE` (due to the `FEECAP`), this means that the `block.coinbase` _loses_ funds from this operation; in this case, we check that the post-balance is non-negative and throw an exception if it is negative.` and explicitly defining // as euclidean division somewhere, as mentioned on the previous PR.

MicahZoltu:
> explicitly defining // as euclidean division somewhere

One of the advantages of using a language (such as Python) is that `//` is already [defined](https://www.python.org/dev/peps/pep-0238/) as part of the language so we can save space in the spec not having to define it again.  This is the same as how we don't define what `**` is or what `if` and `return` are, because the language we are using already defines them for us elsewhere.

> I'm not sure what is wrong with having both.

It is more work to write, maintain, and ensure correctness of two copies of one specification especially when one of those copies is written in an *incredibly* unspecific language (English).  Something to keep in mind is that I actually started by trying to fix the specification section (retaining the plain English style).  I got about half way through the process when I realized that I was basically just defining my own custom language that was difficult to understand because it didn't have well defined rules, which lead to defining the rules (like what does `//` mean), which lead to a lot more content than the Python code.  The problem is that English is just *really* ill suited for specifications.  We need to pick a specification language (e.g., homebrew, math, python, Go, K, whatever) and it felt like Python was the *closest* to English of those options.

> Also It's not only devs that should be able to read and understand the spec, so I think it makes sense to have an explanation in plain terms for those people and the more detailed code for developers.

I suspect the disagreement is stemming from a disagreement about who the target audience of this specification is.  IMO, the target audience of this specification is Ethereum developers, all of whom I *believe* will have no problem reading the Python specification and I *suspect* will have an easier time implementing from it than trying to implement from a plain English specification.  IMO, non-developers who want to learn about this specification will get their knowledge from blog posts, YouTube videos, Twitter threads, Reddit threads, chat rooms, etc. where we can be a whole lot more loose with our terminology and give people "the gist" in a form that is appropriate for the particular audience that is consuming that content.  I do support the abstract being a relatively high level explanation, but no one should be integrating with this standard via the abstract (or any other plain English description), they should be integrating with the "so explicit a computer can interpret it" version of the specification.

MicahZoltu:
For the Python code, I chose to implement the validation function.  The plain English specification on the other hand described transaction/block construction.  Would people feel better about the Python specification if instead of a validation function, it had a block construction function?

I went with validation because it ends up being slightly simpler as you don't have to try to explain in the code that miner gas_target voting and transaction selection are "up to the individual clients" but "constrained in these ways".  Also, I suspect that most tests will be of block validity, rather than block construction (again, since block construction has wiggle room for implementations).

MicahZoltu:
I have rewritten the abstract to try to include more technical depth.  I intentionally left out the specific formula/values, as I am still concerned with the abstract being out of sync with the specification, and I still do believe that anyone trying to implement this **really** should be reading the specification and not the English text.  However, I do agree with others that there should be a "plain English description" of what this specification does that will allow a more casual reader to fully understand how things work without having to dive into the specification itself.

Would like to hear feedback from @i-norden and @gakonst to see if this addresses your concern at all, or if your position remains unchanged after the aforementioned changes.

EIPS/eip-1559.md
```diff
@@ -34,48 +38,168 @@ An important aspect of this fee system is that miners only get to keep the gas p
...
-* To produce an EIP1559 transactions, the new `GAS_PREMIUM` and `FEECAP` fields are set as follows:
-  * `GAS_PREMIUM` serves as a "tip" to the miner
-  * `FEECAP` serves as the absolute maximum that the transaction sender is willing to pay
-* During transaction execution, for EIP1559 transactions we calculate the cost to the `tx.origin` and the gain to the `block.coinbase` as follows:
-  * Set `GASPRICE = min(BASE_FEE + tx.GasPremium, tx.fee_cap)`
-  * Let `GASUSED` be the gas used during the transaction execution/state transition
-  * The `tx.origin` initially pays `GASPRICE * tx.gasLimit`, and gets refunded `GASPRICE * (tx.gasLimit - GASUSED)`
-  * The `block.coinbase` gains `(GASPRICE - BASE_FEE) * GASUSED`.
-    * If `GASPRICE < BASE_FEE` (due to the `FEECAP`), this means that the `block.coinbase` _loses_ funds from this operation; in this case, we check that the post-balance is non-negative and throw an exception if it is negative.
+*Note: `//` in Python is integer division with rounding toward infinity (not truncation).*
```
i-norden, line 41: Should further clarify here that Python `//` rounds negative results to negative infinity and positive results to 0.

EIPS/eip-1559.md
```diff
@@ -34,48 +38,168 @@ An important aspect of this fee system is that miners only get to keep the gas p
...
+				assert signer.balance >= 0, 'invalid transaction: signer does not have enough ETH to cover gas'
+				gas_used = self.execute_transaction(transaction)
+				gas_refund = transaction.gas_limit - gas_used
+				cumulative_1559_transaction_gas_used += gas_used
+				# signer gets refunded for unused gas
+				signer.balance += gas_refund * block.base_fee
+				signer.balance += gas_refund * premium_per_gas
+				# miner only receives the premium; note that the base fee is not given to anyone (it is burned)
+				self.account(block.author).balance += gas_used * premium_per_gas
+				# if the fee cap is lower than the base fee, the miner can choose to pay for the transaction which could result in miner balance going negative
```
i-norden, line 152: This needs to be changed such that we assert `gas_price >= base_fee` and there is no deducting of coinbase balance allowed.

MicahZoltu: I have been trying to avoid normative changes when doing large revisions like this, but since it is difficult to get engagement on these PRs I'll start bundle them unless others complain.

EIPS/eip-1559.md
MicahZoltu: Fixed this to indicate that Python `//` rounds toward `0`.

For those not following along in Discord, here is a synopsis of the rounding issue:

Besu originally implemented truncation and Geth implemented round toward negative infinity.  Upon noticing consensus failure Besu changed to match Geth.  However, every modern language (C++11, Java, C#, JavaScript, Python, Go, Rust) uses truncation for integer division (round toward 0) *including Go* (which Geth is written in) and the only reason Geth initially rounded toward negative infinity is because the big number library that Geth uses rounds toward negative infinity (for reasons that I'm not familiar with).  As far as we know, this is the first time integer division with negative numbers has been part of consensus, so this is the first time this has come up.

Since every language rounds toward 0 (many of which natively support large integers), I argue that this spec should specify that we round toward 0.

@i-norden argues (please correct me if I'm misrepresenting your position) that since Geth is the oldest and most calcified codebase, we should do what is easiest for Geth since it doesn't matter *functionally* which direction we round (at least in this case).

i-norden: Python `//` rounds to 0 only when the result is positive, it rounds to negative infinity when it is negative. I have confirmed this in a Python interpreter. Both besu and geth are conformed to this, I think we should remain this way.

EIPS/eip-1559.md
```diff
@@ -38,7 +38,7 @@ An important aspect of this fee system is that miners only get to keep the gas p
 The transition to this gas price system will occur in two phases, in the first phase both legacy and EIP1559 transactions will be accepted by the protocol. Over the course of this first phase the amount of gas available for processing legacy transactions will decrease while the amount of gas available for processing EIP1559 transactions will increase, moving gas from the legacy pool into the EIP1559 pool until the legacy pool is depleted and the EIP1559 pool contains the entire gas maximum. After all of the gas has transitioned to the EIP1559 pool, the second, finalized, phase is entered and legacy transactions will no longer be accepted on the network.
 
 ## Specification
-*Note: `//` in Python is integer division with rounding toward infinity (not truncation).*
+*Note: `//` in Python is integer division with truncation (round toward zero).*
```
i-norden, line 41: This still isn't quite correct. It rounds to negative infinity when the result is negative, it rounds to 0 when the result is positive. This has been confirmed in a Python interpreter.

MicahZoltu: Bleh, I must he been drunk when I edited this last.  I know what the right comment is but for some reason when I type it my fingers refuse (I have failed on the comment 3 times now).  I'm AFK for the night but will fix this in the morning (hopefully 4th time is a charm).

MicahZoltu: Fixed it to read that Python rounds toward negative infinity.  I'm typing out this description to help ensure that my fingers didn't deceive me again:
```
>>> 9/4
2.25
>>> 9//4
2
>>> # 2 is closer to negative infinity than 3
>>> (-9)/4
-2.25
>>> (-9)//4
-3
>>> # -3 is closer to negative infinity than -2
```

I am in favor of ETH 2.x's strategy of "never use negative integers anywhere in the code", but I'm not sure how to write it well for round toward infinity (it is a bit easier for round toward zero behavior, which would be consensus incompatible).  Once we are ready for a consensus change, I would like to change the comment to the following because I think it is simpler to understand if you don't have to worry about negative numbers:
> `//` is the integer division operator, which truncates

And then change the code to:
```
if parent_gas_used >= parent_gas_target:
	excess_gas_used = parent_gas_used - parent_gas_target
	expected_base_fee = parent_base_fee + parent_base_fee * excess_gas_used // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
else:
	extra_gas_available = parent_gas_target - parent_gas_used
	expected_base_fee = parent_base_fee - parent_base_fee * extra_gas_available // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
assert expected_base_fee == block.base_fee, 'invalid block: base fee not correct'
```

MicahZoltu:
Updated this PR to include the change to make the base fee change by *at least* 1 (to deal with the problem of what happens around 0 base fee).

i-norden:
Thanks @MicahZoltu ! When you get a chance could you also rebase this ontop of the new master state? Maybe auto-merge-bot will handle the conflict but it's not clear to me if that's the case.

Unless people come and voice their objections I am inclined to go ahead and merge this, I will interpret the indifference as passive approval. Will let it sit over the weekend and see if we can collect more feedback but otherwise will merge on Monday. Does that sound fair to you?

EIPS/eip-1559.md
```diff
@@ -114,7 +114,7 @@ class World(ABC):
 		# check if the base fee is correct
 		if parent_gas_used >= parent_gas_target:
 			gas_delta = parent_gas_used - parent_gas_target
-			fee_delta = parent_base_fee * gas_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
+			fee_delta = max(parent_base_fee * gas_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR)
```
i-norden, line 117: Think we're missing a `, 1` in those max parentheses?

MicahZoltu: Doh!  Totally.  My docker is broken at the moment which is how I test Python stuff... so I just winged it thinking, "I can't possibly get this one line change wrong..."

MicahZoltu:
> Unless people come and voice their objections I am inclined to go ahead and merge this, I will interpret the indifference as passive approval.

I'm not a huge fan of winning our debate by default.  😢 While I won't fight against your proposal to merge after the weekend, it does leave me with a bit of a sour taste in my mouth knowing that my PR got merged just because the other party was a bigger person and was willing to back down when I wasn't.  😬

EIPS/eip-1559.md
i-norden: Haha oh yes, I can relate :) it's often the small things which demand the least amount of attention that get messed up!

MicahZoltu:
> When you get a chance could you also rebase this ontop of the new master state?

I believe I have addressed this.  I am no longer getting a merge conflict from GitHub.

i-norden:
> > Unless people come and voice their objections I am inclined to go ahead and merge this, I will interpret the indifference as passive approval.
> 
> I'm not a huge fan of winning our debate by default. 😢 While I won't fight against your proposal to merge after the weekend, it does leave me with a bit of a sour taste in my mouth knowing that my PR got merged just because the other party was a bigger person and was willing to back down when I wasn't. 😬

I don't want this to leave a bad taste in your mouth. Let me be more clear, I think this PR is a vast improvement on the current spec so I am approving it. I would prefer to add/retain a header section with the conditions spelled out in regular language, but I don't want that to hold up merging this any longer. If I or somebody else really wants to have that section, we can open up a PR to add it and argue its merits there.

EIPS/eip-1559.md
```diff
@@ -34,48 +38,174 @@ An important aspect of this fee system is that miners only get to keep the gas p
...
+
+class World(ABC):
+	def validate_block(self, block: Block) -> None:
+		parent_base_fee = self.parent(block).base_fee
+		parent_gas_used = self.parent(block).gas_used
+		parent_gas_target = self.parent(block).gas_target
+		transactions_all = self.transactions(block)
+
+		# check if the block changed the gas target too much
+		assert block.gas_target <= parent_gas_target + parent_gas_target // 1024, 'invalid block: gas target increased too much'
```
holgerd77, line 111: Hi there, we've started with the EIP-1559 [TypeScript implementation](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1148), will drop some comments here and there if I stumble upon something in the spec (not sure where to do best TBH, does this make sense (as I've done now) to search for the old PRs with the respective changes?).

So first thing here: where is this `1024` coming from? Should this eventually also get a name and a dedicated constant in the section above (`INITIAL_FORK_BLOCK_NUMBER`,...)?

MicahZoltu: It is from before EIP-1559, nothing changed.  It is included here simply because we are touching code related to gas limits and wanted to make it clear that this isn't changing behavior.

MicahZoltu: It is better to ask in the Ethereum R&D Discord server in #eip-1559-dev channel or in the `discussions-to` link on this EIP.

holgerd77: Ok, thanks, I will do that.

PR #2865, opened 2020-08-07
Update eip-2803.md

vporton:
Drawbacks and Alternative Proposals

This EIP is considered not a good idea by me, I have an alternative proposal.

### DIFF
--- a/EIPS/eip-2803.md
+++ b/EIPS/eip-2803.md
@@ -54,5 +54,11 @@ TBD.
 ## Implementation
 None yet.
 
+## Drawbacks and Alternative Proposals
+
+See https://github.com/ethereum/EIPs/issues/2863 for an alternative proposal.
+
+Particularly https://github.com/ethereum/EIPs/issues/2863#issuecomment-670551944 explains why this EIP is considered not a good idea.
+
 ## Copyright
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - EIP 2803 requires approval from one of (@micahzoltu)

EIPS/eip-2803.md
```diff
@@ -54,5 +54,11 @@ TBD.
 ## Implementation
 None yet.
 
+## Drawbacks and Alternative Proposals
+
+See https://github.com/ethereum/EIPs/issues/2863 for an alternative proposal.
+
+Particularly https://github.com/ethereum/EIPs/issues/2863#issuecomment-670551944 explains why this EIP is considered not a good idea.
```
MicahZoltu, line 61: In general, external links shouldn't be included in an EIP and instead statements should be included inline.

Also, EIPs are not really the place for discussion on whether an EIP is a good idea or not, that goes better in the discussion-to link (see the header of the EIP).

Also, `## Drawbacks and Alternative Proposals` isn't one of the sections specified in the EIP template.  While the template doesn't have to be followed exclusively, it is generally desirable to ensure EIPs follow that layout.

MicahZoltu:
I'm going to close this for now since it isn't really appropriate for an EIP as it isn't directly related to the specification.  You can, of course, open a separate EIP that proposes the changes you would like to see.

PR #2867, opened 2020-08-08
2309 to Final

pizzarob:
Review period has ended. Moving this EIP to Final! This EIP was originally proposed in October of 2019. It has been refined and tested in production environments. It recently helped facilitate the creation of 20,000 NFT vouchers where 500 of these vouchers were sold in a pre-sale. By including the event mentioned in this EIP we are able to effectively track the creation and distribution of these vouchers. Vouchers could be purchased in bulk and tracked effectively - saving consumers massive gas costs for bulk purchases.

### DIFF
--- a/EIPS/eip-2309.md
+++ b/EIPS/eip-2309.md
@@ -61,7 +61,7 @@ Contracts that implement the `ConsecutiveTransfer` event **MAY** still use the o
   event ConsecutiveTransfer(uint256 indexed fromTokenId, uint256 toTokenId, address indexed fromAddress, address indexed toAddress);
 ```
 
-## Examples
+### Examples
 
 The `ConsecutiveTransfer` event can be used for a single token as well as many tokens:
 
@@ -80,7 +80,6 @@ The `ConsecutiveTransfer` event can be used for a single token as well as many t
 **Burn**
 
 `emit ConsecutiveTransfer(1, 100000, from, address(0));`
-
 
 
 ## Rationale
@@ -114,6 +113,8 @@ As the NFT market continues to grow so does the need for the ability to scale th
 
 This extension was written to allow for the smallest change possible to the original ERC-721 spec while still providing a mechanism to track the creation, transfer, and deletion of a massive amount of tokens. While it is a minimal change the effects on platforms that only use the original `Transfer` event to index token ownership would be severe. They would not be properly recording token ownership information that could be known by listening for the `ConsecutiveTransfer` event. For platforms that wish to support the `ConsecutiveTransfer` event it would be best to support both the original `Transfer` event and the `ConsecutiveTransfer` event to track token ownership. 
 
+## Security Considerations
+There are no security considerations related directly to the implementation of this standard.
 
 ## Copyright
 
--- a/EIPS/eip-2309.md
+++ b/EIPS/eip-2309.md
@@ -20,7 +20,7 @@ The optional ERC-721 Consecutive Transfer Extension provides a standardized even
 
 ## Motivation
 
-This extension provides even more scalibility of the ERC-721 specification. It is possible to create, transfer, and burn 2^255 non-fungible tokens in one transaction. However, it is not possible to emit that many `Transfer` events in one transaction. The `Transfer` event is part of the original specification which states:
+This extension provides even more scalibility of the [ERC-721 specification](./eip-721.md). It is possible to create, transfer, and burn 2^255 non-fungible tokens in one transaction. However, it is not possible to emit that many `Transfer` events in one transaction. The `Transfer` event is part of the original specification which states:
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 2309 state from Last Call to Final

EIPS/eip-2309.md
```diff
@@ -3,8 +3,7 @@ eip: 2309
 title: ERC-721 Consecutive Transfer Extension
 author: Sean Papanikolas (@pizzarob)
 discussions-to: https://github.com/ethereum/EIPs/issues/2309
-status: Last Call
-review-period-end: 2020-09-08
```
MicahZoltu, line 7: Review period doesn't end for this until September (ISO-8601 date format).

pizzarob: @MicahZoltu gah my mistake - it was supposed to be August 9 - two weeks from the start of last call status. Is there anything we can do here?

MicahZoltu: Unfortunately, I think the best you can do is move back to DRAFT, then to Last Call again with a date that is 2 weeks from that date.  The reason for this is that it will trigger pushing the Last Call EIP to an RSS feed with a new correct date, and give people ample time to review before it is over.  This does mean that you'll have to wait at least 2 more weeks though.  😢

MicahZoltu: If you decide to do this, ping me and I'll do my best to merge both (to Draft and then to Last Call) quickly.

pizzarob: @MicahZoltu that's unfortunate. I will just leave as is. Would you recommend closing this PR and re-opening September 8?

MicahZoltu: Either that or just mention me on September 8th.  :)

pizzarob: @MicahZoltu today is the day! Can this be merged?

axic:
Sorry for the late review here. I think at least the first ERC-721 and ERC-1155 mention should link to the specification.

Why is `fromTokenId` indexed, but `toTokenId` is not?

pizzarob:
@axic good call re: links. Added in latest commit. Regarding indexed - You can only add the `indexed` attribute to up to three parameters. The rest of the data can be read from the data part of the log.

axic:
> You can only add the indexed attribute to up to three parameters. The rest of the data can be read from the data part of the log.

Yep, but what I miss is the explanation in the rationale why indexed vs. non-indexed, why these particular arguments, etc. And I would say more emphasis on the indexed attribute would be useful, because it is easy to miss, and Solidity had issues with it in during inheritance in the past.

pizzarob:
Good call @axic - updated.

EIPS/eip-2309.md
```diff
@@ -46,6 +45,10 @@ The address executing the transaction **MUST** own all the tokens within the ran
 
 The `fromTokenId` and `toTokenId` **MUST** be a consecutive range of tokens IDs.
 
+The `fromTokenId`, `fromAddress`, and `toAddress` **MUST** be indexed parameters
```
axic, line 48: Nitpicking:
```suggestion
The `fromTokenId`, `fromAddress`, and `toAddress` **MUST** be indexed parameters.
```

EIPS/eip-2309.md
```diff
@@ -46,6 +45,10 @@ The address executing the transaction **MUST** own all the tokens within the ran
...
+
+The `toTokenId` **MUST NOT** be an indexed parameter
```
axic, line 50: Nitpicking:
```suggestion
The `toTokenId` **MUST NOT** be an indexed parameter.
```

Given the rest of the statements were terminated with a period.

EIPS/eip-2309.md
axic, line 48: Side-comment: Solidity calls them "indexed", but in the protocol they are called "topics". Anyway, since this standard targets mostly Solidity developers, this is the better option.

axic (approved):
Looks good to me!

MicahZoltu (approved):

MicahZoltu (changes requested):
Missing `## Security Considerations` section (comes right before `## Copyright`) which is required.  If there are none, then you can assert as much in that section.

`## Examples` isn't a standard section, and while non-standard sections are not prohibited, we prefer to avoid them whenever possible.  Could that perhaps be moved into another section like a subsection in `## Specification`?

MicahZoltu (approved):

EIPS/eip-2309.md
axic: @pizzarob realised that half of the sentences were terminated with a period, the other half weren't. If you still want to fix that feel free to open a PR, it can be merged as it is only a stylistic change.

PR #2860, opened 2020-08-06
Switch validator to eipv

lightclient:
As discussed in the [EIPIP](https://github.com/ethereum-cat-herders/EIPIP/issues/25) meetings, I've been working on a new validator called [`eipv`](https://github.com/lightclient/eipv). It isn't feature complete yet, but it should be comparable to the existing validator.

PR #2873, opened 2020-08-11
Fix eip-1679 rendering

lightclient:
The HTML generator is having trouble with the double colon in the title line of EIP-1679. I removed the quotes when adding the new validator, and for some reason the HTML generator raised it as a warning rather than an error. I've logged [this issue](https://github.com/lightclient/eipv/issues/6) with `eipv`, but we should go ahead and fix this ASAP.

### DIFF
--- a/EIPS/eip-1679.md
+++ b/EIPS/eip-1679.md
@@ -1,6 +1,6 @@
 ---
 eip: 1679
-title: Hardfork Meta: Istanbul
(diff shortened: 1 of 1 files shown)


lightclient:
I also just realized there was a typo on this EIP updating the ordering of the required EIPs. I have fixed this and I rereviewed #2860 for any other typos and it looks good.

PR #2876, opened 2020-08-13
Add ERC for deposit address and contract interface

junderw:
This ERC defines a simple contract interface for managing deposits. It also defines a new address format that encodes the extra data passed into the interface's main deposit function.

### DIFF
--- a/EIPS/eip-2876.md
+++ b/EIPS/eip-2876.md
@@ -1,0 +1,188 @@
+---
+eip: 2876
+title: Deposit contract and address standard
+author: Jonathan Underwood (@junderw)
+discussions-to: https://github.com/junderw/deposit-contract-poc/issues/1
+status: Draft
+type: Standards Track
+category: ERC
+created: 2020-08-13
+---
+
+<!--You can leave these HTML comments in your merged EIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new EIPs. Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
+This ERC defines a simple contract interface for managing deposits. It also defines a new address format that encodes the extra data passed into the interface's main deposit function.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Centralized exchanges and merchants (Below: "apps") require an address format for accepting deposits. Currently the address format used refers to an account (external or contract), but this creates a problem. It requires that apps create a new account for every invoice / user. If the account is external, that means the app must have the deposit addresses be hot wallets, or have increased workload for cold wallet operators (as each deposit account will create 1 value tx to sweep). If the account is contract, generating an account costs at least 60k gas for a simple proxy, which is cost-prohibitive.
+
+This ERC seeks to create a standard for deposit systems (and how wallets interface with those systems) that will alleviate these problems.
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+As stated in the abstract, merchant and centralized exchange apps are forced between taking on one of the following:
+
+- Large security risk (deposit accounts are hot wallets)
+- Large manual labor cost (cold account manager spends time sweeping thousands of cold accounts)
+- Large service cost (deploying a contract-per-deposit-address model).
+
+The timing of this proposal is within the context of increased network gas prices. During times like this, more and more services who enter the space are being forced into hot wallets for deposits, which is a large security risk.
+
+The motivation for this proposal is to lower the cost of deploying and managing a system that accepts deposits from many users, and by standardizing the methodology for this, services across the world can easily use this interface to send value to and from each other without the need to create multiple accounts.
+
+## Specification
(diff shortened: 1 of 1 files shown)


junderw:
Question: Should I explicitly mention that `address(0)` is not a valid base address and explicitly require an address validator to fail if the base address portion is `address(0)`? Or is it sufficient enough to assume as a given?...

EIPS/eip-2876.md
```diff
@@ -0,0 +1,188 @@
...
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.-->
+This ERC defines a simple contract interface for managing deposits. It also defines a new address format that encodes the extra data passed into the interface's main deposit function.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Centralized exchanges and merchants (Below: "apps") require an address format for accepting deposits. Currently the address format used refers to an account (external or contract), but this creates a problem. It requires that apps create a new account for every invoice / user. If the account is external, that means the app must have the deposit addresses be hot wallets, or have increased workload for cold wallet operators (as each deposit account will create 1 value tx to sweep). If the account is contract, generating an account costs at least 60k gas for a simple proxy, which is cost-prohibitive.
```
MicahZoltu, line 19: Consider phrasing this with a target audience of someone reading it in 5 to 10 years.  Instead of saying stuff like "Currently the address format used..." and instead focus on what this standard describes and mostly ignore how things worked in the past.  The motivation section is usually the best place for describing the problems that are caused by *not* having this standard.

EIPS/eip-2876.md
```diff
@@ -0,0 +1,188 @@
...
+
+This ERC seeks to create a standard for deposit systems (and how wallets interface with those systems) that will alleviate these problems.
```
MicahZoltu, line 21: The first paragraph should probably be removed or moved to motivation (see previous comment) which leaves only this sentence as the abstract.  Consider expanding it a bit to try to give more detail on how you achieve that goal, though without going all the way into the specification. I like to use the abstract as a place to provide a human readable description of the specification for people who are trying to evaluate if this standard is what they want, and then the specification section is highly technical and targeted at people who are trying to implement this specification.

EIPS/eip-2876.md
```diff
@@ -0,0 +1,188 @@
...
+- `deposit(bytes8)` MUST revert if the deposit somehow failed and the contract does not need to keep the value sent.
+- `deposit(bytes8)` MUST return `true` if the value will be kept and the payment is logically considered complete by the parent application (exchange/merchant).
+- `deposit(bytes8)` SHOULD check the checksum contained within the 8 byte id. (See Part 1 for an example)
+- `The parent application` SHOULD return any excess value received if the deposit id is a one-time-use invoice that has a set value and the value received is higher than the set value. However, this SHOULD NOT be done by sending back to `msg.sender` directly, but rather should be noted in the parent application and the depositor should be contacted out-of-band to the best of the application manager's ability.
+
+### Part 3: Depositing Value to the Contract from a Wallet
```
MicahZoltu, line 118: Consider removing the `Part #` bit off of these subsections.  The title is likely enough, and the order of implementation isn't important for the specification so it isn't necessary to enumerate the sections.

junderw:
Thanks for the comments, I made a fix here #2877 

As for eip-681, from the perspective of an exchange looking for a solution to a problem (which I am), here are the reasons why I found it insufficient:

1. Adoption rate is low among other exchanges and wallets.
2. Designing a UI/UX around it is nearly impossible, since any transaction we make on behalf of our user we need to get confirmation "is this what you want to do?"... but if we scan a URI that says "call gorbldeyGoopedy(uint256,bytes8) with this number and these bytes" how do I confirm that with my user without confusing the heck out of them? If we assume the users all know the abi of the contracts and what they do/mean, yeah maybe we could make a UI that just says the function name and shows the parameters... but this assumption is not safe imo.
3. Implementation is difficult. Not only due to UX/UI issues above, but also conceptually for developers to comprehend the vast possibility space of what a scanned URI could mean for the app they are developing. It is extremely open ended, and if they think "well, our exchange only needs to deposit funds, why do I need to support every possible abi under the sun? They will likely avoid using it for their ETH wallet. Perhaps they'll use it if they are super-into ETH development and want to support some super complicated DeFi app, but exchanges especially want to have the smallest exposure possible to possible issues.

From an exchange perspective, the best UX and dev experience for account-based-blockchain deposit systems has been ripple, thanks to their destination tag feature. However, that is not without its headaches (there is no checksum nor is the dest_tag embedded in the address, so people forget it or mistype it all the time... the deposit address format laid out in the EIP fixes those problems).

It would also be easy to fit support into existing APIs for my proposal. ie. `web3.eth.sendTransaction` could accept deposit address format in the `to` field, and throw an error if `to` is a deposit address and `data` is not null. Then exchanges will automatically be able to support this deposit address format if they use `web3.eth.sendTransaction` internally.

These are the reasons why I did not decide to just use EIP-681.

I really look forward to evangelizing this standard.

MicahZoltu:
Note: The following is personal opinion, unrelated to my job as an EIP editor:

For (1), it feels like you'll end up with the same problem with this standard.  Anything you can do to push adoption of this EIP you can do to push adoption of 681.

It feels like there isn't much of a difference to end-users between "put 0xabcd1234... into your wallet" and "put eth://abcd1234 into your wallet".  In both cases the user will get an unhelpful error if their wallet doesn't support the standard, and in the case of this EIP the thing they paste into their wallet *looks* like a regular address (just a hex string) so it is less likely they'll ask the right questions or a wallet support representative who doesn't know about this EIP will realize the user is following a new standard.

A wallet could, if they so chose, only implement ETH transfers via 681 and give the user a "useful error" if they try to use a payment URL that does anything else.  Alternatively, you could standardize a particular contract endpoint (like this EIP does) that is used specifically for ETH transfers but have the wallet integration protocol be 681.  This way wallets could have a smaller set of 681 things to support, rather than having to support arbitrary method calls.

PR #2881, opened 2020-08-15
eip-1559: fix comparison error in abstract.

raulk:
I may be getting this entirely wrong, but it seems like the abstract flips around the compensatory actions when blocks are under/over the target gas limit?


eip-automerger:

PR #2890, opened 2020-08-20
Move `eips.ethereum.org` to Zola

lightclient:
I'm working on migrating away from Jekyll. I believe that [Zola](https://github.com/getzola/zola) is a more robust alternative, and by transitioning to it will begin to be able to take advantage of more static checks. This will improve the quality of EIPs and reduce the likelihood of invalid changes being merged.

Although this work is still in progress, it would be great for people to begin looking at this PR and sharing their thoughts.

The goal of this PR is transition to Zola with a nearly identical site. Further improvements can be added once this is complete. One of the largest changes is the migration to `toml`-based front matter (the part at the top of the EIP). Zola doesn't support the `yaml` variant that has been used previously. Although this will add a bit of a learning curve for authors, it is still relatively straightforward and the downstream gains should outweigh the initial cost. I would recommend auditing [the script](https://gist.github.com/lightclient/d8525a97f0b3fa798377705263f63f5c) I wrote to convert the `yaml` front matter to `toml`.

#### Why Move to Zola?
* ~10x faster compilation than Jekyll
* much easier for authors / editors to preview work before it reaches `eips.ethereum.org` since there is no need to rely on a Ruby environment
* better static analysis of EIP metadata
* errors are much clearer than in Jekyll

#### Todo
- [ ] Update `eipv` to support `toml` front matter
- [ ] Ensure assets load for EIPs
- [ ] Use Zola link checker to replace `html-proofer`
- [ ] Setup codespell again
- [ ] Push zola build to github pages
- [ ] Update readme instructions
- [ ] Update template to have the new `toml` front matter

#### Running Locally
See how easy it is to run locally:
1. Download Zola: https://www.getzola.org/documentation/getting-started/installation/
2. Clone this branch `git clone -b move-to-zola https://github.com/lightclient/EIPs.git && cd EIPs/www`
3. Run `zola serve`
4. Goto http://localhost:1111


Souptacular:
Love the idea. Will you need admin access to accomplish this? If so I can talk with the other editors about giving you temp. access so you don't have to rely on others to install.

lightclient:
Thanks, @Souptacular. I will need admin access to setup the deployment to github pages, but I will ping you when everything else is complete (probably a couple weeks) and we can figure that out.

lightclient:
@MicahZoltu thanks for finding that! I'll take a look into it. Skipping the TOML change at first seems like reasonable approach.

axic:
I have been following this thread since it was created and the change from yaml to toml seemed quite excessive, without a strong benefit this provides. I am still unsure what is a strong benefit here after looking at the initial points:

> Why Move to Zola?
> ~10x faster compilation than Jekyll

Where is this speed difference showing up as an issue? Why is it a problem that Jekyll is slower? Can Jekyll's speed be improved?

> much easier for authors / editors to preview work before it reaches eips.ethereum.org since there is no need to rely on a Ruby environment

Jekyll needs Ruby, Zola needs Rust. Both need an environment set up, and none of them come entirely standard on any system. (Though a version of Ruby comes by default with macOS and perhaps some Linux distributions too.) Is this really only about preferences?

> better static analysis of EIP metadata

Is that not the duty of eip-checker / eipv?

> errors are much clearer than in Jekyll

Do you have an example where this caused an issue in the past?

axic:
Additionally this PR seems to mix **a lot of** changes. Among other things, renames the template and the EIPS directory name.

There are bunch of outstanding PRs improving the current set up (adding anchor links, table of contents, more RSS feeds), all of which are pending on review from other editors. It would be sad to just ignore those PRs and the efforts of their authors, if they are just ignored, and essentially closed after this PR. Unless all those proposals are also covered with this switch.

lightclient:
@axic, this PR is very much WIP and a lot of the removal was for my own sanity while getting everything setup. Travis would be readded, along with HTML link checks (zola has a built in checker, would have to see how it compares to the existing setup).

That's a good point on the Solidity syntax highlighting. Zola supports Sublime syntax, so I would guess that adding Solidity will be relatively straightforward. I can open an issue on Zola. So far, the maintainer has been very responsive.

There is a [Github action](https://github.com/marketplace/actions/zola-deploy-to-pages) to deploy Zola sites to Github pages. It's a bit more cumbersome, but once it is setup it shouldn't be a problem. All the endpoints are intact (or will be, if I missed anything). I haven't worked on the RSS yet, but it will also be intact.

It is unfortunate that there are a lot of pending PRs. I'll take a look and see what has been glossed over and see if I can integrate them into this PR. I believe that moving to Zola will make it easier for people to build and view the site, which I *hope* will mean more people will review site changes, instead of allowing them to collect dust.

MicahZoltu:
> Jekyll needs Ruby, Zola needs Rust. Both need an environment set up, and none of them come entirely standard on any system. (Though a version of Ruby comes by default with macOS and perhaps some Linux distributions too.) Is this really only about preferences?

Jekyll needs Ruby to *run*, Zola needs Rust to *compile* (compiles to native).  Whether this matters or not is certainly a valuable question to ask, but I do think there is a difference that may matter here.

lightclient:
Fortunately Zola provides pre-built binaries that are available on most popular package managers: https://www.getzola.org/documentation/getting-started/installation/

Souptacular:
+1 to avoiding a TOML conversion. I think you've covered @axic's concerns (unless I missed any) so I am pro going ahead with this until we find a reason not to.

axic:
> Jekyll needs Ruby to run, Zola needs Rust to compile (compiles to native). Whether this matters or not is certainly a valuable question to ask, but I do think there is a difference that may matter here.

>  Fortunately Zola provides pre-built binaries that are available on most popular package managers: https://www.getzola.org/documentation/getting-started/installation/

I assumed Zola is like any other Rust program, relying on cargo. Don't get me wrong, I am a big proponent of Rust and write a lot of Rust code myself. (Likewise I was working with Ruby for years, long time ago.)

I am still not convinced what is the main benefit and reason for this rather massive amount of work. None of the questions from https://github.com/ethereum/EIPs/pull/2890#issuecomment-682254975 are answered.

lightclient:
@axic sorry, I responded on mobile and I guess I missed your first post.

> Where is this speed difference showing up as an issue? Why is it a problem that Jekyll is slower? 
It's not a major issue, but it doesn't hurt to have it build faster.

> Can Jekyll's speed be improved?
I haven't looked deeply into this, but I assume that the Ruby runtime will continue to block significant performance improvements.

> Jekyll needs Ruby, Zola needs Rust. Both need an environment set up, and none of them come entirely standard on any system. (Though a version of Ruby comes by default with macOS and perhaps some Linux distributions too.) Is this really only about preferences?

Zola releases [pre-built binaries](https://github.com/getzola/zola/releases), so there is no need to setup an environment. I think this is a great advantage.

> better static analysis of EIP metadata -- Is that not the duty of eip-checker / eipv?
It is, but the current `yaml` format requires much more processing since everything is generally considered a string instead of using native `yaml` elements. This isn't so much of an advantage of Zola, we could just update the existing front-matter.

> errors are much clearer than in Jekyll -- Do you have an example where this caused an issue in the past?
e.g. when `eipv` was merged, I removed the quotes from EIP-1679 and it broke rendering (although for some reason it didn't break it in the PR?): https://travis-ci.org/github/ethereum/EIPs/jobs/717001998

I think I responded to your questions in your second post though. IMO, the biggest benefits are i) no need to setup an environment just to preview the site and ii) we get a stack with better static validation.

axic:
> IMO, the biggest benefits are i) no need to setup an environment just to preview the site and ii) we get a stack with better static validation.

I still debate these.

> i) no need to setup an environment just to preview the site 

One still needs to install Zola. It does not come without installing anything. I think what is a better direction is providing web based tools for EIP authors. I doubt anyone today or in the future will install anything. Most of the people submitting are not fluent with git, and use Github's "edit file" feature.

> ii) we get a stack with better static validation.

I still do not understand what do you mean by this.

**To summarise**: I fear this project will take weeks, if not months, and the benefits will not be as big as imagined here.

I am strongly on the opinion that this work is better spent on
1) a web based interface for EIP authors,

2) or adding bots which comment on the problematic lines and leave a summary. The Solidity repository has such a bot for coding style issues, if there is a need for inspiration.

eips/eip-1588.md
```diff
@@ -19,7 +21,7 @@ This meta-EIP specifies the changes included in the alternative Ethereum hardfor
 - Activation:
   - `Block >= 7280000` on the Ethereum mainnet
 - Included EIPs:
-  - [EIP 1057](./eip-1057.md): ProgPoW, a Programmatic Proof-of-Work
+  - [EIP 1057](@/EIPS/eip-1057.md): ProgPoW, a Programmatic Proof-of-Work
```
axic, line 24: What kind of syntax is this? is it markdown compatible? Should we consider this for Jekyll? (See #2947)

lightclient: I believe this is a Zola specific thing. The `@/*` style denotes an "internal" link, which as I understand it is mapped to an internal post / asset. I believe that style can be manipulated by Zola to achieve certain permalink schemes. The `./*` style denotes a "relative" link and is treated as a simple relative link.

lightclient: The CHANGELOG on Zola for more info: https://github.com/getzola/zola/blob/2d1c95432290478547316522d1f9219ac38e5b13/CHANGELOG.md#breaking-4

MicahZoltu: As with my other comments above, I'm moderately against any change that breaks formatting when viewed in GitHub.

PR #2910, opened 2020-08-26
eip-2535.md

mudgen:


MicahZoltu:
@lightclient Do you know how hard it would be to make it so the automerge bot can automerge asset changes to eip folders when the author is an author of the EIP associated with that folder?

lightclient:
It shouldn't be too difficult. I'd like to do some work on the bot (like) this in the near future. Just been a bit bogged down the last couple weeks.

PR #2911, opened 2020-08-26
EIP779 (DAO HF meta EIP): add the `extraData` requirement

jochem-brouwer:
EIP779 (DAO HF meta EIP) "summarizes" what the DAO fork should do. However, it does not state that the `extraData` field of the blocks 1920000-1920009 ~~should~~ must be "dao-hard-fork". If someone wishes to implement their own Ethereum client, this is very important, because this gets explicitly tested in [ethereum/tests](https://github.com/ethereum/tests/tree/develop/BlockchainTests/TransitionTests/bcHomesteadToDao) and thus should be put in the "DAO fork summary" EIP.


EIPS/eip-779.md
```diff
@@ -54,6 +54,8 @@ contract WithdrawDAO {
 }
 ```
 
+The 9 blocks including and after the activation block (so block 1,920,000 - 1,920,009) should have "dao-hard-fork" (`0x64616f2d686172642d666f726b`) in the `extraData` field of the block.
```
MicahZoltu, line 57: If this was required, then this should be reworded to be a **MUST** statement.

MicahZoltu: If this was not a consensus requirement, then it should not be part of the specification.

jochem-brouwer: This is a requirement per the linked article in the EIP: https://blog.slock.it/hard-fork-specification-24b889e70703

Thanks, I reworded it to must =)

EIPS/eip-779.md
```diff
@@ -54,6 +54,8 @@ contract WithdrawDAO {
 }
 ```
 
+The 9 blocks including and after the activation block (so block 1,920,000 - 1,920,009) must have "dao-hard-fork" (`0x64616f2d686172642d666f726b`) in the `extraData` field of the block.
```
MicahZoltu, line 57:
```suggestion
Blocks with block numbers in the range [1,920,000, 1,920,009] **MUST** have `0x64616f2d686172642d666f726b` (hex encoded ASCII string `dao-hard-fork`) in the `extraData` field of the block.
```

MicahZoltu:
I'm currently waiting on confirmation from one or more (ideally more) client dev teams to assert that this is in fact implemented as part of consensus in their client.

lightclient:
Not a client dev, but a brief peak into `go-ethereum` does show that the `extraData` field **MUST** be equal to `0x64616f2d686172642d666f726b`.

* Verification from `ethash` module -- https://github.com/ethereum/go-ethereum/blob/4f2784b38f59066913bbaf1e6fa531ec724b3650/consensus/ethash/consensus.go#L296-L299
* Actual check -- https://github.com/ethereum/go-ethereum/blob/4f2784b38f59066913bbaf1e6fa531ec724b3650/consensus/misc/dao.go#L58-L62
* `extraData` constant definition -- https://github.com/ethereum/go-ethereum/blob/461291882edce0ac4a28f64c4e8725b7f57cbeae/params/dao.go#L25-L28

MicahZoltu:
That is convincing enough for me for Geth.  I'll give a few days to see if @cdetrio shows up to merge this, or if one of the other client developers pipes up and says their clients do not do that check for consensus.  Otherwise I'll merge.

ajsutton:
Besu also includes this check: https://github.com/hyperledger/besu/blob/a9bdd29128eca70d1922b282aa2b479ab492b563/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockHeaderValidator.java#L47

axic:
Since the goal is to clarify this EIP, I'd suggest to also explicitly include the list of accounts, e.g. what is being described in
> At block 1880000, the following accounts are encoded into a list `L`:

The complete list can be found here: https://github.com/ethereum/go-ethereum/blob/461291882edce0ac4a28f64c4e8725b7f57cbeae/params/dao.go#L41-L156

@jochem-brouwer what do you think?

axic:
> At the beginning of block 1920000, all ether throughout all accounts in `L` will be transferred to a contract deployed at `0xbf4ed7b27f1d666546e30d74d50d173d20bca754`. The contract was created from the following Solidity code (compiler version `v0.3.5-2016-07-01-48238c9`):

I'd also either include the contract bytecode or specify the block and transaction hash it was deployed at, for reference.

EIPS/eip-779.md
```diff
@@ -54,6 +54,8 @@ contract WithdrawDAO {
 }
 ```
 
+Blocks with block numbers in the range [1,920,000, 1,920,009] **MUST** have `0x64616f2d686172642d666f726b` (hex encoded ASCII string `dao-hard-fork`) in the `extraData` field of the block.
```
axic, line 57:
```suggestion
Blocks with block numbers in the range [1,920,000 ... 1,920,009] **MUST** have `0x64616f2d686172642d666f726b` (hex encoded ASCII string `dao-hard-fork`) in the `extraData` field of the block.
```

Otherwise the punctuation conflicts with the separator.

axic: Or the alternative:
```suggestion
Blocks with block numbers in the range [1_920_000, 1_920_009] **MUST** have `0x64616f2d686172642d666f726b` (hex encoded ASCII string `dao-hard-fork`) in the `extraData` field of the block.
```

MicahZoltu: I mildly favor the second since `[1234, 5678]` is standard math syntax for inclusive set range.

jochem-brouwer:
@axic Have added the list L, the contract creation transaction hash and the contract bytecode.

axic:
Nice! Personally I think it is better to include the list of accounts as opposed to linking geth, because those links can break. Less likely with geth, but still possible that links break.

jochem-brouwer:
You are right, I will add the list in the EIP.

EIPS/eip-779.md
```diff
@@ -54,6 +177,14 @@ contract WithdrawDAO {
...
...ffffffffffffffffffffff1681526020019150506020604051808303816000876161da5a03f11561000257505050604051805190602001503073ffffffffffffffffffffffffffffffffffffffff16310103604051809050600060405180830381858888f19350505050505b565b600073bb9bc244d798123fde783fcc1c72d3bb8c18941373ffffffffffffffffffffffffffffffffffffffff166370a0823133604051827c0100000000000000000000000000000000000000000000000000000000028152600401808273ffffffffffffffffffffffffffffffffffffffff1681526020019150506020604051808303816000876161da5a03f1156100025750505060405180519060200150905073bb9bc244d798123fde783fcc1c72d3bb8c18941373ffffffffffffffffffffffffffffffffffffffff166323b872dd333084604051847c0100000000000000000000000000000000000000000000000000000000028152600401808473ffffffffffffffffffffffffffffffffffffffff1681526020018373ffffffffffffffffffffffffffffffffffffffff16815260200182815260200193505050506020604051808303816000876161da5a03f1156100025750505060405180519060200150158061041657503373ffffffffffffffffffffffffffffffffffffffff16600082604051809050600060405180830381858888f19350505050155b1561042057610002565b5b50565b73bb9bc244d798123fde783fcc1c72d3bb8c18941381565b600060009054906101000a900473ffffffffffffffffffffffffffffffffffffffff168156
```
axic, line 183: Is this the deployment/creation code (e.g. what is sent in that transaction) or the bytecode deployed on the chain?

jochem-brouwer: Clarified this.

MicahZoltu:
I haven't heard any objections, and we received at least a couple confirmations from clients, so I'm going to merge this.

I'm going against normal policy here and not waiting on the original author to approve primarily because this is so old that it is quite likely the author is no longer available to comment.  Also, while this is a normative change, it can be confirmed by looking at client source code what reality is.

jochem-brouwer:
Great, thanks a lot @MicahZoltu. This will be very helpful for anyone implementing DAO-fork clients. 

I think in general these (fork) Meta-EIPs should be checked to see if they are complete. Building clients from scratch is extremely hard because I have not yet found a complete, readable spec of what a client should do. To do so, one should probably first create a client which supports Byzantium and from that point on re-implement the older hard forks, as it does not seem that there is any (readable) spec around in the open for Frontier clients.

axic:
> as it does not seem that there is any (readable) spec around in the open for Frontier clients.

archive.org probably has a version of the YP from 2016 which was not updated to Homestead and is *mostly* complete :wink:

PR #2915, opened 2020-08-28
Fix ambiguity in syncing definition for eth_syncing (EIP-1474)

hu55a1n1:
Previously we only said `eth_syncing` should return `false if this client is not syncing with the network`, but we didn't define what syncing meant. I have taken the liberty to define `not syncing` to mean `currentBlock == highestBlock`.
Note: client implementations seem to diverge from each other over the interpretation of what `syncing` means. With the change that is proposed here, clients would be continuously switching b/w syncing and not-syncing, because there will always be a newer block in the network.

### DIFF
--- a/EIPS/eip-1474.md
+++ b/EIPS/eip-1474.md
@@ -2147,7 +2147,7 @@ _(none)_
 
 ##### Returns
 
-{`boolean|object`} - `false` if this client is not syncing with the network, otherwise an object with the following members:
+{`boolean|object`} - `false` if `currentBlock` is equal to `highestBlock` (as defined below), otherwise an object with the following members:
 
 - {[`Quantity`](#quantity)} `currentBlock` - number of the most-recent block synced
 - {[`Quantity`](#quantity)} `highestBlock` - number of latest block on the network


eip-automerger:

 - EIP 1474 requires approval from one of (@bitpshr, rekmarks@protonmail.com)

EIPS/eip-1474.md
```diff
@@ -2147,7 +2147,7 @@ _(none)_
 
 ##### Returns
 
-{`boolean|object`} - `false` if this client is not syncing with the network, otherwise an object with the following members:
+{`boolean|object`} - `false` if `currentBlock` is equal to `highestBlock` (as defined below), otherwise an object with the following members:
```
MicahZoltu, line 2150: We should figure out what all of the clients do to determine the return value here.  Does someone want to volunteer to do this research?

I suspect it has more to do with whether syncing is forcibly stopped by the user, or if there is something preventing further syncing (e.g., no peers) than just being transiently behind.  That is, a client can be syncing:true when it starts syncing and is way behind, or when a new block has just arrived and it hasn't processed it yet.

hu55a1n1: Apparently, Nethermind doesn't returns false even when `currentBlock` and `highestBlock` are equal. So clients implementations seem to differ from each other on this.
Another possibility to fix the ambiguity is to completely remove the `boolean` return type.

hu55a1n1: So, here's what Nethermind is currently doing. The RPC method is implemented in [EthModule.cs#L77](https://github.com/NethermindEth/nethermind/blob/eed064c478f53994c8590b5d2203479389ed92c3/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthModule.cs#L77) -> 
```csharp
public ResultWrapper<SyncingResult> eth_syncing()
{
    SyncingResult result;
    if (_blockchainBridge.IsSyncing)
    {
        result = new SyncingResult
        {
            CurrentBlock = _blockchainBridge.Head.Number,
            HighestBlock = _blockchainBridge.BestKnown,
            StartingBlock = 0L,
            IsSyncing = true
        };
    }
    else
    {
        result = SyncingResult.NotSyncing;
    }

    return ResultWrapper<SyncingResult>.Success(result);
}
```
And here's the `IsSyncing` property definition in [BlockchainBridge.cs#L124](https://github.com/NethermindEth/nethermind/blob/eed064c478f53994c8590b5d2203479389ed92c3/src/Nethermind/Nethermind.Facade/BlockchainBridge.cs#L124) -> 
```csharp
public bool IsSyncing => _blockTree.BestSuggestedHeader.Hash != _blockTree.Head.Hash;
```
The actual calculations for `BestSuggestedHeader` and `Head` are somewhat complicated and implemented in [BlockTree.cs](https://github.com/NethermindEth/nethermind/blob/eed064c478f53994c8590b5d2203479389ed92c3/src/Nethermind/Nethermind.Blockchain/BlockTree.cs). I'll continue to post my findings here.

hu55a1n1:
Unfortunately, I am not able to contribute to this anymore. But I strongly feel that this is an ambiguity in the spec that needs to be fixed. If two implementations of the same spec diverge from each other, that can only mean one of two things - either one (or both) implementation(s) are wrong, or the spec is ambiguous.

MicahZoltu:
I'm going to close this PR since it currently has no champion, even though I agree that *someone* should try to fix this.

PR #2919, opened 2020-08-28
Run codespell on everything

axic:
I started pushing out a lot of old pending changes I did locally, this is the first one.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -17,5 +17,5 @@ elif [[ $TASK = 'htmlproofer-external' ]]; then
 elif [[ $TASK = 'eip-validator' ]]; then
   eipv EIPS/ --ignore=title_max_length,missing_discussions_to --skip=eip-20-token-standard.md
 elif [[ $TASK = 'codespell' ]]; then
-  codespell -q4 -I .codespell-whitelist eip-X.md EIPS/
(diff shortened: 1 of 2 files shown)


.travis-ci.sh
```diff
@@ -17,5 +17,5 @@ elif [[ $TASK = 'htmlproofer-external' ]]; then
 elif [[ $TASK = 'eip-validator' ]]; then
   eipv EIPS/ --ignore=title_max_length,missing_discussions_to --skip=eip-20-token-standard.md
 elif [[ $TASK = 'codespell' ]]; then
-  codespell -q4 -I .codespell-whitelist eip-X.md EIPS/
+  codespell -q4 -I .codespell-whitelist -S ".git,Gemfile.lock,**/*.png,**/*.gif,**/*.jpg,**/*.svg,.codespell-whitelist,vendor,_site,_config.yml,style.css"
```
axic, line 20: The main benefit here is that the template is also checked, while it was not because when `eip-X.md` was renamed to `eip-template.md`, nobody changed this line :wink:

MicahZoltu (approved):
Assuming this does what the author claims it does, I'm fine with it.  I know very little about Travis, but this doesn't look like an attack so I am OK with merging it and seeing what happens.

PR #2920, opened 2020-08-28
Remove social media links from the page (but keep Github)

axic:
Not sure whether it is useful. The Youtube channel has ACD, Eth2.0 calls, Eth1.x calls, which have relevance here.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,13 +19,7 @@ description: >-
   platform, including core protocol specifications, client APIs, and contract
   standards.
 url: "https://eips.ethereum.org"
-twitter_username: ethereum
 github_username:  ethereum
-youtube_username: channel/UCNOfzGXD_C9YMYmnefmPH0g
-
-twitter:
-  card: summary
-  username: ethereum
 
 header_pages:
  - all.html
--- a/_config.yml
+++ b/_config.yml
@@ -21,6 +21,12 @@ description: >-
 url: "https://eips.ethereum.org"
 twitter_username: ethereum
 github_username:  ethereum
+youtube_username: channel/UCNOfzGXD_C9YMYmnefmPH0g
+
+twitter:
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -21,6 +21,12 @@ description: >-
 url: "https://eips.ethereum.org"
 twitter_username: ethereum
 github_username:  ethereum
+youtube_username: channel/UCNOfzGXD_C9YMYmnefmPH0g
```
MicahZoltu, line 24: This is the EF's channel, not an EIPIP specific channel.  Same with GitHub username and Twitter username though... what is the purpose of all of these things being in here?

axic: I'd either add the youtube link as that includes ACD, Eth2 calls, etc. or remove github/twitter/youtube.

axic: @Souptacular @nicksavers any opinions?

Souptacular: I think we should remove GitHub/YouTube/Twitter until we have a more neutral place that the videos belong to. Linking official EF with EIPs has bad optics at minimum. I do understand my previous sentence has the implication that the ACD calls should be moved off of the EF YouTube channel and I agree with that, but it hasn't happened yet. As far as the `ethereum` repo being primarily controlled by the EF and having a mix of EF and non-EF repos, we are in the process of rectifying that.

axic: Reopened this PR because it seems it was never followed up on.

1. Should all these social media links removed them?
2. With the exception of GitHub since the repo is there?

axic:
Following up https://github.com/ethereum/EIPs/pull/2920#discussion_r479645994, this just removes all links bar github.

MicahZoltu (approved):

Souptacular:
I'm fine with the links being removed :)

PR #2923, opened 2020-08-29
Update html title to be EIP-N

lightclient:
Title before: `EIP 1: EIP Purpose and Guidelines`
Title after: `EIP-1: EIP Purpose and Guidelines`

Reflects the style guide in EIP-1:

> When referring to an EIP by number, it should be written in the hyphenated form EIP-X where X is the EIP’s assigned number.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -3,7 +3,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   {% if page.layout == "eip" %}
-    <title>EIP {{ page.eip }}: {{ page.title | escape }}</title>
+    <title>EIP-{{ page.eip }}: {{ page.title | escape }}</title>
     <meta property="og:title" content="EIP {{ page.eip }}: {{ page.title | escape }}" />
     <meta
       name="description"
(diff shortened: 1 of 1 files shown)


axic (approved):
Is there any other place which was missed? Template, readme, the frontpage of eips.ethereum.org?

_includes/head.html
```diff
@@ -3,7 +3,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   {% if page.layout == "eip" %}
-    <title>EIP {{ page.eip }}: {{ page.title | escape }}</title>
+    <title>EIP-{{ page.eip }}: {{ page.title | escape }}</title>
     <meta property="og:title" content="EIP {{ page.eip }}: {{ page.title | escape }}" />
```
axic, line 7:
```suggestion
    <meta property="og:title" content="EIP-{{ page.eip }}: {{ page.title | escape }}" />
```

axic: And more instances of this to be found in the metadata below.

lightclient: Should be update now, thanks for pointing those out.

MicahZoltu (approved):

PR #2924, opened 2020-08-29
Changes 1559 to have a single shared pool for all transactions.

MicahZoltu:
Rather than having two pools and a migration, we instead have a single transaction pool and legacy transactions are "upgraded" into 1559 transactions using a *very* naive strategy of just setting `fee_cap == gas_bribe == gas_price`.

While in here, I also made a few other minor changes since this is a consensus change and will require all clients update:
1. Renamed premium to bribe.  I think this makes it more clear what this is, it is a bribe for the miner to convince them to include your transaction.
2. Changed the gas target calculation to be a bit simpler (per prior conversation with @i-norden)
3. Added special handling for when the gas_used == gas_target (no change in target for next block).
4. Adds missing assertion that gas_used wasn't more than 2x the gas target.
5. Added clarity to some names like `fee_cap` to `fee_cap_per_gas` and `miner_bribe` to `miner_bribe_per_gas`.

### DIFF
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -119,7 +119,7 @@ class World(ABC):
 			expected_base_fee = parent_base_fee + fee_delta
 		else:
 			gas_delta = parent_gas_target - parent_gas_used
-			fee_delta = parent_base_fee * gas_delta // parent_gas_used // BASE_FEE_MAX_CHANGE_DENOMINATOR
+			fee_delta = parent_base_fee * gas_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
 			expected_base_fee = parent_base_fee - fee_delta
 		assert expected_base_fee == block.base_fee, 'invalid block: base fee not correct'
 
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -16,7 +16,8 @@ A transaction pricing mechanism that includes fixed-per-block network fee that i
 There is a base fee per gas in protocol, which can move up or down each block according to a formula which is a function of gas used in parent block and gas target (formerly known as gas limit) of parent block.
 The algorithm results in the base fee per gas increasing when blocks are above the gas target, and decreasing when blocks are below the gas target.
 The base fee per gas is burned.
-Transactions specify both the maximum network fee per gas (aka: base fee) they are willing to pay, as well as the maximum fee per gas they are willing to give to miners to incentivize them to include their transaction (aka: miner bribe).
+Transactions specify the maximum fee per gas they are willing to give to miners to incentivize them to include their transaction (aka: miner bribe).
+Transactions also specify the maximum fee per gas they are willing to pay total (aka: fee cap), which covers both the miner bribe and the block's network fee per gas (aka: base fee).
 The transaction will always pay the base fee per gas of the block it was included in, and they will pay the bribe per gas set in the transaction, as long as the combined amount of the two fees doesn't exceed the transaction's maximum fee per gas.
 
 ## Motivation
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -19,12 +19,8 @@ The base fee per gas is burned.
 Transactions specify both the maximum network fee per gas (aka: base fee) they are willing to pay, as well as the maximum fee per gas they are willing to give to miners to incentivize them to include their transaction (aka: miner bribe).
 The transaction will always pay the base fee per gas of the block it was included in, and they will pay the bribe per gas set in the transaction, as long as the combined amount of the two fees doesn't exceed the transaction's maximum fee per gas.
 
-This specification also includes a mechanism for migrating from the current transaction/block fee system to the new one.
-Initially, half of a block can be legacy transactions, and half of a block can be new transactions.
-Over time, the percentage of the block that can contain legacy transactions decreases, as more room for new transactions is made.
-
 ## Motivation
-Ethereum currently prices transaction fees using a simple auction mechanism, where users send transactions with bids ("gasprices") and miners choose transactions with the highest bids, and transactions that get included pay the bid that they specify. This leads to several large sources of inefficiency:
+Ethereum historically priced transaction fees using a simple auction mechanism, where users send transactions with bids ("gasprices") and miners choose transactions with the highest bids, and transactions that get included pay the bid that they specify. This leads to several large sources of inefficiency:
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

EIPS/eip-1559.md
```diff
@@ -16,8 +16,8 @@ A transaction pricing mechanism that includes fixed-per-block network fee that i
 There is a base fee per gas in protocol, which can move up or down each block according to a formula which is a function of gas used in parent block and gas target (formerly known as gas limit) of parent block.
 The algorithm results in the base fee per gas increasing when blocks are above the gas target, and decreasing when blocks are below the gas target.
 The base fee per gas is burned.
-Transactions specify both the maximum fee per gas they are willing to pay, as well as the maximum fee per gas they are willing to give to miners to incentivize them to include their transaction (aka: gas premium).
-The transaction will always pay the base fee per gas of the block it was included in, and they will pay the premium per gas set in the transaction, as long as the combined amount of the two fees doesn't exceed the transaction's maximum fee per gas.
+Transactions specify both the maximum network fee per gas (aka: base fee) they are willing to pay, as well as the maximum fee per gas they are willing to give to miners to incentivize them to include their transaction (aka: miner bribe).
```
bluepintail, line 19: `FEECAP` was previously the absolute maximum that the transaction sender is willing to pay. I think setting the overall maximum and the miner bribe maximum (as previously specified) makes more sense, otherwise the user has to add the two together to work out what the overall maximum fee is.

MicahZoltu: There was no intention to change the behavior here (this particular change was meant to be non-normative).  Do you have a suggestion for better wording than what I have that would make that more clear (ideally without going *too* in depth)?

bluepintail: Might be easier to specify the other way around - "transactions specify the maximum fee per gas they are willing to give to miners to incentivize them to include their transaction (aka: miner bribe), they also specify a maximum overall fee per gas they are willing to pay, including both the miner bribe and the network fee (aka: base fee)"

MicahZoltu:
```suggestion
Transactions specify the maximum fee per gas they are willing to give to miners to incentivize them to include their transaction (aka: miner bribe).
Transactions also specify the maximum fee per gas they are willing to pay total (aka: fee cap), which covers both the miner bribe and the block's network fee per gas (aka: base fee).
```
How does this sound @bluepintail

i-norden:
I think instead of `fee_cap == gas_bribe == gas_price`  

It is more accurate to say `fee_cap == gas_price; gas_bribe = fee_cap - base_fee`

In practice we should set `gas_bribe = gas_price` in the tx struct, but the amount of `gas_price` actually available for the bribe after burning the` base_fee` will be `fee_cap - base_fee`.

Thanks for opening this! I haven't had a chance to look into the rest of this PR in more depth, will do so soon.

MicahZoltu:
@i-norden I think what you are getting at is that `miner_bribe_per_gas` is a bit of a misnomer and really it is `maximum_miner_bribe_per_gas`.  If so, I agree with you and have updated the EIP to reflect that.

Now that we have merged the other PR, I have rebased this and it is ready to merge once the authors approve it.  Per discussion in Discord, we should make note that `711474afe36a9d778f7e1858b2c9a88b0bf146d0` is the commit hash for EIP-1559 that is currently on the integration testnet, and whatever the merge commit hash for this is will be the "target" EIP we are working toward and probably what new clients should implement.

@i-norden, @mslipper, @AFDudley, @vbuterin, @econoar

EIPS/eip-1559.md
```diff
@@ -96,97 +96,87 @@ class Account:
 	code_hash: int = 0
 
 INITIAL_FORK_BLOCK_NUMBER = 10 # TBD
-MIGRATION_DURATION_IN_BLOCKS = 800000
-FINAL_FORK_BLOCK_NUMBER = INITIAL_FORK_BLOCK_NUMBER + MIGRATION_DURATION_IN_BLOCKS
 BASE_FEE_MAX_CHANGE_DENOMINATOR = 8
+ELLASTICITY_MULTIPLIER = 2
```
vbuterin, line 100:
```suggestion
ELASTICITY_MULTIPLIER = 2
```

EIPS/eip-1559.md
```diff
@@ -96,97 +96,87 @@ class Account:
...
 class World(ABC):
 	def validate_block(self, block: Block) -> None:
 		parent_base_fee = self.parent(block).base_fee
 		parent_gas_used = self.parent(block).gas_used
 		parent_gas_target = self.parent(block).gas_target
-		transactions_all = self.transactions(block)
+		transactions = self.transactions(block)
+
+		# check if the block used too much gas
+		assert block.gas_used <= block.gas_target * ELLASTICITY_MULTIPLIER, 'invalid block: too much gas used'
```
vbuterin, line 110:
```suggestion
		assert block.gas_used <= block.gas_target * ELASTICITY_MULTIPLIER, 'invalid block: too much gas used'
```

MicahZoltu:
I have removed the prose text that referred to the transition as well (missed it in original PR).

EIPS/eip-1559.md
```diff
@@ -96,97 +93,87 @@ class Account:
...
+			expected_base_fee = parent_base_fee
+		elif parent_gas_used > parent_gas_target:
 			gas_delta = parent_gas_used - parent_gas_target
 			fee_delta = max(parent_base_fee * gas_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR, 1)
 			expected_base_fee = parent_base_fee + fee_delta
 		else:
 			gas_delta = parent_gas_target - parent_gas_used
-			fee_delta = (parent_base_fee * gas_delta - 1) // parent_gas_used // BASE_FEE_MAX_CHANGE_DENOMINATOR + 1
+			fee_delta = parent_base_fee * gas_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
 			expected_base_fee = parent_base_fee - fee_delta
```
tkstanczak, line 123: need to address cases when expected_base_fee ends up being < 0

EIPS/eip-1559.md
```diff
@@ -96,97 +93,87 @@ class Account:
...
 		assert block.gas_target <= parent_gas_target + parent_gas_target // 1024, 'invalid block: gas target increased too much'
 		assert block.gas_target >= parent_gas_target - parent_gas_target // 1024, 'invalid block: gas target decreased too much'
 
 		# check if the base fee is correct
-		if parent_gas_used >= parent_gas_target:
+		if parent_gas_used == parent_gas_target:
+			expected_base_fee = parent_base_fee
+		elif parent_gas_used > parent_gas_target:
 			gas_delta = parent_gas_used - parent_gas_target
 			fee_delta = max(parent_base_fee * gas_delta // parent_gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR, 1)
```
tkstanczak, line 118: max not needed here, it is fine to move by near 0 wei gas prices

EIPS/eip-1559.md
MicahZoltu: Discussed this in Discord, recording result here for thoroughness: In that particular conditional branch, there is no way for `gas_delta` to be negative, thus there is no way for `fee_delta` to be negative.  The same is true for the other branch.

abdelhamidbakhta (approved):
Looks good to me.

PR #2926, opened 2020-08-31
Chunk-Based Code Merkleization

s1na:
This EIP specifies a chunk-based approach to code merkleization with the goal of reducing block witness size in a stateless/partial-stateless ethereum.


MicahZoltu (approved):
🚢

PR #2932, opened 2020-09-01
Check root for eip files

lightclient:
Occasionally, people accidentally create EIP files in the root dir. This can be missed by editors very easily. This PR updates the CI script to check if any files in the root directory match the pattern `eip-*`. This should help reduce the number of accidental merges of EIP in the wrong location.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -15,8 +15,7 @@ elif [[ $TASK = 'htmlproofer-external' ]]; then
   bundle exec jekyll build
   bundle exec htmlproofer $HTMLPROOFER_OPTIONS --external_only
 elif [[ $TASK = 'eip-validator' ]]; then
-  ls | grep 'eip-' | grep -v 'eip-template.md' || EXIT=$?
-  if [ -z $EXIT ]; then
+  if [[ $(find . -maxdepth 1 -name 'eip-*' | wc -l) -ne 1 ]]; then
     echo "only 'eip-template.md' should be in the root"
     exit 1
   fi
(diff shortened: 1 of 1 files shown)


.travis-ci.sh
```diff
@@ -15,6 +15,11 @@ elif [[ $TASK = 'htmlproofer-external' ]]; then
   bundle exec jekyll build
   bundle exec htmlproofer $HTMLPROOFER_OPTIONS --external_only
 elif [[ $TASK = 'eip-validator' ]]; then
+  ls | grep 'eip-' | grep -v 'eip-template.md' || EXIT=$?
```
adlerjohn, line 18: This looks like it works, but generally piping the output of `ls` isn't a great idea. If you're trying to count the number of files with some name, it makes more sense to just do that directly semantically:

```sh
if [[ $(find . -name 'eip-*' | wc -l) -ne 1 ]]; then
  echo "only 'eip-template.md' should be in the root"
  exit 1
fi
```

(Also prefer ``[[ ]]` over `[ ]` [in bash](https://stackoverflow.com/a/669486).)

lightclient: This looks great, thanks @adlerjohn. Will update tomorrow.

lightclient:
cc @axic @MicahZoltu @Souptacular @adlerjohn

MicahZoltu:
I'm fine with the change, but I didn't want to approve since I don't know bash.

.travis-ci.sh
```diff
@@ -15,6 +15,10 @@ elif [[ $TASK = 'htmlproofer-external' ]]; then
   bundle exec jekyll build
   bundle exec htmlproofer $HTMLPROOFER_OPTIONS --external_only
 elif [[ $TASK = 'eip-validator' ]]; then
+  if [[ $(find . -maxdepth 1 -name 'eip-*' | wc -l) -ne 1 ]]; then
```
axic, line 18: I don't particularly think this is a good check (it is really fragile and not specific), but it is a net improvement compared to today.

axic (approved):

.travis-ci.sh
lightclient: Agreed, I would prefer to do as you recommended and scan for front-matter outside the `EIPS`. But for now this is a net improvement.

PR #2935, opened 2020-09-03
Save historical block hashes in state

vbuterin:
Store historical block hashes in a contract, and modify the `BLOCKHASH (0x40)` opcode to read this contract.


MicahZoltu (approved):

deauna (commented):
should commit?

PR #2955, opened 2020-09-08
eip-2565: rework it

holiman:
The EIP-2565 is totally out of sync with the spec on the discussions-to, and also does not make sense, as 

- the suggested option 2 actually makes the precompile more expensive (divisor `20` to `3`),
- The quoted formula from `198` is actually not correct, 
- It has two recommended options, but we're actually only going to implement one
- It has a third option which is not consensus-related, and shouldn't be part of the EIP

Discussion at https://ethereum-magicians.org/t/eip-2565-big-integer-modular-exponentiation-eip-198-gas-cost/4150/5  
This PR is not ready to merge, just a starting point for rewriting it. 

Primarily, it needs filling on the gas-table of testcases, and I'd rather have the authors do that, to ensure that the values used for the benchmarkings/charts matches with what's in the testcases.

### DIFF
--- a/EIPS/eip-2565.md
+++ b/EIPS/eip-2565.md
@@ -15,22 +15,62 @@ requires: 198
 The [EIP-198](https://eips.ethereum.org/EIPS/eip-198) ‘big integer modular exponentiation’, or `ModExp`, precompile is currently overpriced. Re-pricing this precompile will enable more cost efficient verification of RSA signatures, verifiable delay functions (VDFs), primality checks, and more.
 
 ## Abstract
-After benchmarking the ModExp precompile, we discovered that it is ‘overpriced’ relative to other precompiles. We also discovered that the current gas pricing formula could be improved to better estimate the computational complexity of various ModExp input variables. To improve the gas cost pricing for this precompile the following options are available:
+After benchmarking the ModExp precompile, we discovered that it is ‘overpriced’ relative to other precompiles. We also discovered that the current gas pricing formula could be improved to better estimate the computational complexity of various ModExp input variables. 
+To improve the gas cost pricing for this precompile, this EIP specifies
 
-1. Modifying the gas pricing formula to better reflect the computational complexity of ModExp operations
-2. Changing the value of the `GQUADDIVISOR` parameter in the ModExp pricing formula to bring its costs more in-line with other precompiles
-3. Improving the underlying libraries beneath the ModExp Precompile
-4. Any combination of (1), (2), and (3)
-
-We recommend **Option (1) and (2)** which provides a large practical improvement to gas estimation while keeping implementation complexity low. Option (3) can be implemented at a later point to improve the gas pricing even further.
+* Changing the value of the `GQUADDIVISOR` parameter in the ModExp pricing formula to bring its costs more in-line with other precompiles
 
 ## Motivation
 Modular exponentiation is a foundational arithmetic operation for many cryptographic functions including signatures, VDFs, SNARKs, accumulators, and more. Unfortunately, the ModExp precompile is currently over-priced, making these operations inefficient and expensive. By reducing the cost of this precompile, these cryptographic functions become more practical, enabling improved security, stronger randomness (VDFs), and more.
 
 ## Specification
-The current gas pricing formula is defined in EIP-198. This formula divides a ‘computational complexity’ function by a ‘gas conversion’ parameter called `GQUADDIVISOR` to arrive at a gas cost: `floor(mult_complexity(x)/GQUADDIVISOR)`
 
-### **Recommended** Option (1): Modify ‘computational complexity’ function and add minimum gas cost
+The current gas pricing formula is defined in [EIP-198](eip-198.md) as: 
+
+```
+floor(mult_complexity(max(length_of_MODULUS, length_of_BASE)) * max(ADJUSTED_EXPONENT_LENGTH, 1) / GQUADDIVISOR)
+```
+
+At `fork_block`, 
+ - set `GQUADDIVISOR` to `200`.
+
+
+## Rationale
+
+With this change, the cost of the ModExp precompile will have a higher cost (gas/second) than other precompiles such as ECRecover.
+
+![Option 2 Graph](../assets/eip-2565/GQuad_Change.png)
+
+## Test Cases
+
+There are no changes to the underlying interface or arithmetic algorithms, so the existing test vectors can be reused. Gas values will need to be updated in the existing unit tests based on the final pricing decided in this EIP. This will ensure that the updated gas calculations are done correctly.
+
+Existing testvectors from [go-ethereum](https://github.com/ethereum/go-ethereum/blob/master/core/vm/testdata/precompiles/modexp.json):
+
+| testcase | EIP-198 gas| 2565 opt  1 gas| 2565 opt 2 gas | 
+| -|--|--|--
+|modexp_nagydani_1_square| 204 | 100| 100?|
+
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - EIP 2565 requires approval from one of (@simonatsn, @dankrad, @justindrake, @sean-sn, @ineffectualproperty)

EIPS/eip-2565.md
```diff
@@ -15,22 +15,62 @@ requires: 198
...
 
 ## Specification
-The current gas pricing formula is defined in EIP-198. This formula divides a ‘computational complexity’ function by a ‘gas conversion’ parameter called `GQUADDIVISOR` to arrive at a gas cost: `floor(mult_complexity(x)/GQUADDIVISOR)`
 
-### **Recommended** Option (1): Modify ‘computational complexity’ function and add minimum gas cost
+The current gas pricing formula is defined in [EIP-198](eip-198.md) as: 
+
+```
+floor(mult_complexity(max(length_of_MODULUS, length_of_BASE)) * max(ADJUSTED_EXPONENT_LENGTH, 1) / GQUADDIVISOR)
+```
```
MicahZoltu, line 32: Defining old behavior feels out of scope for a specification.  Can we just leave this bit out, or move it to the rationale section or something?

EIPS/eip-2565.md
```diff
@@ -15,22 +15,62 @@ requires: 198
...
+| -|--|--|--
+|modexp_nagydani_1_square| 204 | 100| 100?|
+
+TODO fill this table
+
+
+
+## Security Considerations
+
+The biggest security consideration for this EIP is  creating a potential DoS vector by making ModExp operations too inexpensive relative to their computation time.
```
MicahZoltu, line 60:
```suggestion
The biggest security consideration for this EIP is creating a potential DoS vector by making ModExp operations too inexpensive relative to their computation time.
```

EIPS/eip-2565.md
```diff
@@ -15,22 +15,62 @@ requires: 198
...
-### **Recommended** Option (1): Modify ‘computational complexity’ function and add minimum gas cost
+The current gas pricing formula is defined in [EIP-198](eip-198.md) as: 
+
+```
+floor(mult_complexity(max(length_of_MODULUS, length_of_BASE)) * max(ADJUSTED_EXPONENT_LENGTH, 1) / GQUADDIVISOR)
+```
+
+At `fork_block`, 
+ - set `GQUADDIVISOR` to `200`.
+
```
MicahZoltu, line 36:
```suggestion
```

EIPS/eip-2565.md
```diff
@@ -15,22 +15,62 @@ requires: 198
...
+Existing testvectors from [go-ethereum](https://github.com/ethereum/go-ethereum/blob/master/core/vm/testdata/precompiles/modexp.json):
+
+| testcase | EIP-198 gas| 2565 opt  1 gas| 2565 opt 2 gas | 
+| -|--|--|--
+|modexp_nagydani_1_square| 204 | 100| 100?|
+
+TODO fill this table
+
+
```
MicahZoltu, line 56:
```suggestion
```

EIPS/eip-2565.md
```diff
@@ -15,22 +15,62 @@ requires: 198
...
+## References
+[EIP-198](https://eips.ethereum.org/EIPS/eip-198) 
+
+## Copyright
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+## Appendix
+
+### Abandoned option : Modify ‘computational complexity’ function and add minimum gas cost
+
```
MicahZoltu, line 73: Consider moving this to the rationale (that section is intended to represent why certain choices were made in this EIP) rather than a non-standard Appendix section.

EIPS/eip-2565.md
```diff
@@ -53,42 +93,11 @@ You can find comparison of these two complexity fomulas for the current test vec
...
+#### Rationale
 
-While alternative libraries can provide improved a further 2-5x improvement in performance, this option is not recommended at this time.
-
-## Rationale
-
-### **Recommended** Option (1): Modify ‘computational complexity’ formula
 A comparison of the current ‘complexity’ function and the proposed function described above can be found at the following [spreadsheet](https://docs.google.com/spreadsheets/d/1-xBzA-2-l2ZQDQ1eh3XXGZjcRSBQ_Hnp7NubXpbiSUY/edit?usp=sharing).
 
 ![Option 1 Graph](../assets/eip-2565/Complexity_Regression.png)
```
MicahZoltu, line 100: There are no longer options, so the asset title "Option 1 Graph" no longer makes sense.

EIPS/eip-2565.md
```diff
@@ -15,22 +15,62 @@ requires: 198
...
+* Changing the value of the `GQUADDIVISOR` parameter in the ModExp pricing formula to bring its costs more in-line with other precompiles
 
 ## Motivation
 Modular exponentiation is a foundational arithmetic operation for many cryptographic functions including signatures, VDFs, SNARKs, accumulators, and more. Unfortunately, the ModExp precompile is currently over-priced, making these operations inefficient and expensive. By reducing the cost of this precompile, these cryptographic functions become more practical, enabling improved security, stronger randomness (VDFs), and more.
 
 ## Specification
-The current gas pricing formula is defined in EIP-198. This formula divides a ‘computational complexity’ function by a ‘gas conversion’ parameter called `GQUADDIVISOR` to arrive at a gas cost: `floor(mult_complexity(x)/GQUADDIVISOR)`
 
-### **Recommended** Option (1): Modify ‘computational complexity’ function and add minimum gas cost
+The current gas pricing formula is defined in [EIP-198](eip-198.md) as:
```
MicahZoltu, line 28:
```suggestion
The current gas pricing formula is defined in [EIP-198](./eip-198.md) as: 
```

MicahZoltu (commented):
Merging is up to the authors, but I left some general comments.  I'm a fan of the removal of multiple options, EIPs should generally just be technical specifications, not discussions (that is what the discussions-to link is for).

EIPS/eip-2565.md
holiman: As the original specification in this case is rather small, I think it makes sense to include it here. It puts the specified change in context, so the reader can understand what the effect of the EIP is.

holiman: Also, the description of the previous behaviour was present before this PR (but it was not correct). I'll leave it in, for now, but I guess the final say won't be up to me

holiman:
Also, if the option that is currently "the one" is indeed the way I parsed and rewrote it, without a minimum `100` gascost, then geth seems to get borderline DoS behaviour on `nagydani-1-qube`, at `6mgas/s`, and `nagydani-2-qube` at `13mgas/s` (where a more appropriate value, on this particular machine would be `30-40 mgasps`). 

So please advice...

holiman:
If I use a minimum value of `100`, the worst case becomes `nagydani-3-qube`, at `15.9mgas/s`, with `189 gas` per op (well, assuming `189` is correct)

axic:
Maybe worth checking the feedback on #2892 also.

holiman:
Ah, yes, indeed relevant!

ineffectualproperty:
Thanks for the start to re-working this EIP @holiman . If you could please keep this open for the next couple days I will incorporate your feedback into #2892 . Once that is done we can close this issue and focus on that PR. Appreciate you taking the time to provide some proposed fixes and helping with the process.

PR #2972, opened 2020-09-12
Wrapped Legacy Transactions

MicahZoltu:
Two new transaction types for wrapping legacy transactions with and without a chain ID.


MicahZoltu:
Requesting a review from @axic just to get a second set of eyes, but going to merge this since it already has received a review by an editor (me).  😄

PR #2975, opened 2020-09-13
Add IS_STATIC opcode

vbuterin:
Add a `IS_STATIC (0x4A)` opcode that pushes `1` if the current context is static (ie. the execution is in a `STATICCALL` or a descendant thereof, so state-changing operations are not possible), and `0` if it is not.

### DIFF
--- a/EIPS/eip-2970.md
+++ b/EIPS/eip-2970.md
@@ -1,0 +1,39 @@
+---
+eip: 2970
+title: IS_STATIC opcode
+author: Vitalik Buterin (@vbuterin)
+discussions-to: https://ethereum-magicians.org/t/is-static-opcode-useful-for-aa/4609
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-09-13
+---
+
+## Simple Summary
+
+Add a `IS_STATIC (0x4A)` opcode that pushes `1` if the current context is static (ie. the execution is in a `STATICCALL` or a descendant thereof, so state-changing operations are not possible), and `0` if it is not.
+
+## Abstract
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


EIPS/eip-2970.md
```diff
@@ -0,0 +1,39 @@
...
+type: Standards Track
+category: Core
+created: 2020-09-13
+---
+
+## Simple Summary
+
+Add a `IS_STATIC (0x4A)` opcode that pushes `1` if the current context is static (ie. the execution is in a `STATICCALL` or a descendant thereof, so state-changing operations are not possible), and `0` if it is not.
+
+## Abstract
```
MicahZoltu, line 16:
```suggestion
## Simple Summary

Adds an opcode to detect if the current call context is limited to read-only operations.

## Abstract

Add a `IS_STATIC (0x4A)` opcode that pushes `1` if the current context is static (ie. the execution is in a `STATICCALL` or a descendant thereof, so state-changing operations are not possible), and `0` if it is not.
```
The current simple summary feels more like an abstract due to the level of detail it goes into and how it basically mirrors the specification.

MicahZoltu (approved):

EIPS/eip-2970.md
chfast: It simply does not make sense to have both.

MicahZoltu: The purpose of the simple summary is to give a one sentence overview of what this EIP is.  You can imagine it showing up in some places as a sub-title or inline description.  Think "email subject line".  Abstracts purpose is to describe the specification in human readable terms, with perhaps some of the minor details and edge cases left out.

chfast: Great explanation, thank.

PR #2976, opened 2020-09-13
2976: eth/##: Typed Transactions over Gossip

MicahZoltu:
Adds support for transmission of typed transactions over devp2p.

### DIFF
--- a/EIPS/eip-2976.md
+++ b/EIPS/eip-2976.md
@@ -1,0 +1,90 @@
+---
+eip: 2976
+title: eth/##: Typed Transactions over Gossip
+author: Micah Zoltu (@MicahZoltu)
+discussions-to: https://ethereum-magicians.org/t/eip-2976-eth-typed-transactions-over-gossip/4610
+status: Draft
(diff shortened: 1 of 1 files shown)


MicahZoltu:
Going to merge this as it has received a review from an editor (me), but since it is my own PR I requested a review from @axic to get a second set of eyes.  Would also love to get eyes on this by someone familiar with the networking protocol like @fjl

EIPS/eip-2976.md
```diff
@@ -0,0 +1,90 @@
+---
+eip: 2976
+title: eth/##: Typed Transactions over Gossip
```
lightclient, line 3: I believe this needs to be wrapped with double quotes to render properly on the eips site. This is because the front matter parser thinks "eth/##:" is a key like "title:". This is on the backlog to validate with eipv.

MicahZoltu: Will do.  I copied the name from the other networking EIPs, looking again it does appear they have quotes.

PR #2986, opened 2020-09-18
2938: Update discussion link to new ethmagicians thread

adietrichs:


adietrichs:
I was under the assumption that as an EIP author my PR would be auto-merged. Maybe the failing greeting job prevented that? @MicahZoltu could you have a look?

MicahZoltu:
@adietrichs Yeah, that is exactly the problem.  Going to close and re-open to see if that kicks the bot.

PR #2119, opened 2019-06-14
Add EIP-1654: Off-chain dapp-wallet authentication process

pazams:
The "request for comments" phase on EIP 1654 has converged to a stable spec on  https://github.com/ethereum/EIPs/issues/1654 .

Additionally, the process in this spec has been implemented successfully by a large number of dapps, including:
Crypto Kitties - https://www.cryptokitties.co/
Cheeze Wizards - https://www.cheezewizards.com/
MyCryptoHeroes - https://www.mycryptoheroes.net/
Etheremon - https://www.etheremon.com/
Megacryptopolis - https://www.megacryptopolis.com/
Blockchain Cuties Universe - https://blockchaincuties.com/
Hyperdragons - https://hyperdragons.alfakingdom.com/

Issue #1654, opened 2018-12-13
ERC-1654 Dapp-wallet authentication process with contract wallets support


pazams:
---
eip: <to be assigned>
title: Dapp-wallet authentication process with contract wallets support
author: Maor Zamski (@pazams)
discussions-to: <URL>
status: Draft
type: Meta
created: 2018-12-12
---

<details><summary>First draft</summary>

## Simple Summary

An off-chain process for dapps to prove actionable control (informally, "ownership") over a public Ethereum address using `eth_sign`. Supports both external wallets and contract wallets.

## Definitions
- `contract wallet` A contract [account](https://github.com/ethereum/wiki/wiki/White-Paper#ethereum-accounts) deployed with the intent to be used as the ownership address for on-chain assets (including ether, ERC-20 tokens, and ERC-721 NFTs). It has the ability to transfer ether or dynamically execute actions on other contracts (acting as the owner of assets controlled by those contracts). Common examples of contract wallets are `multisig wallets` (such as the ones provided by [Gnosis](https://github.com/Gnosis/MultiSigWallet) and [Parity](https://github.com/ConsenSys/MultiSigWallet)) and `identity contracts`, as defined in [ERC-725](https://github.com/ethereum/EIPs/issues/725).
- `external wallet` An externally owned [account](https://github.com/ethereum/wiki/wiki/White-Paper#ethereum-accounts), controlled by a private key. Currently, most on-chain assets are owned by such accounts. A common example for an external wallet are the wallets generated by MetaMask.
- `actionable control` A public key is defined to have actionable control over an address if either:
  - It is an external wallet AND the key is determined to correspond to the address.
  - It is a contract wallet AND the key exists in the contract account state and has a purpose of `ACTION` as defined in EIP-725.

## Abstract

The authentication process starts with the dapp client component requesting a message signature from the wallet.
The client then proceeds to send the result to the dapp backend component along with the requested address to be used for authentication. The dapp backend recovers a public key from the signature, and checks if it has actionable control over the requested address. This check is done under consideration that the address may represent either an external wallet or a contract wallet. This process works with external wallets and EIP-725 contract wallets. For this process to be compatible with any other contract wallet, it requires the wallet to implement a small subset of EIP-725.

## Motivation

[...]

PR #2995, opened 2020-09-22
Add EIP-1654: Off-chain dapp-wallet authentication process

pazams:
## Replacing and updating a stale PR

This is a new PR following the kind closure of https://github.com/ethereum/EIPs/pull/2119. This PR is now updated to reflect the recent changes to [EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md) and adds it as a required dependency with this commit https://github.com/ethereum/EIPs/commit/a1e98ed7d89e47644c06a83cf5e51c70ae638200

## Context

The "request for comments" phase on EIP 1654 has converged to a stable spec on #1654 .

Additionally, the process in this spec has been implemented successfully by a large number of dapps, including:
Crypto Kitties - https://www.cryptokitties.co/
Cheeze Wizards - https://www.cheezewizards.com/
MyCryptoHeroes - https://www.mycryptoheroes.net/
Etheremon - https://www.etheremon.com/
Megacryptopolis - https://www.megacryptopolis.com/
Blockchain Cuties Universe - https://blockchaincuties.com/
Hyperdragons - https://hyperdragons.alfakingdom.com/


EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+- External wallets are backwards compatible with this process.
+- Contract wallets need to support `isValidSignature()` as specified in this document.
+
+## Implementation
+Packages implementing the purposed algorithm: 
+- Javascript: https://github.com/dapperlabs/dappauth.js
+- Go: https://github.com/dapperlabs/dappauth
+
+
+## Copyright
```
MicahZoltu, line 103:
```suggestion
## Security Considerations
TBD

## Copyright
```

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Backwards Compatibility
+- External wallets are backwards compatible with this process.
+- Contract wallets need to support `isValidSignature()` as specified in this document.
+
+## Implementation
+Packages implementing the purposed algorithm: 
+- Javascript: https://github.com/dapperlabs/dappauth.js
+- Go: https://github.com/dapperlabs/dappauth
+
```
MicahZoltu, line 101: I believe this section is no longer required, and I encourage removing it if you don't have an inline reference implementation.

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Copyright
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+---
+Thanks to @dete @Arachnid @igorbarbashin @turbolent @jordanschalm @hwrdtm for feedback and suggestions
```
MicahZoltu, line 107:
```suggestion
```
EIPs are not the right forum for giving thanks/kudos/props/shout-outs.  If people want, they can look at the discussion-to link for some out of band history.

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+##### User agent
+
+A user agent intended to work with the contract MUST generate signatures over a EIP-191 hash of a regular Keccak256 hash of the challenge message.
+
+## Rationale
+[EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md) has done a great work with standardising the signature validation method for contracts. This proposal builds on top of it a process for dapp-wallet authentication. 
+
+## Backwards Compatibility
+- External wallets are backwards compatible with this process.
+- Contract wallets need to support `isValidSignature()` as specified in this document.
```
MicahZoltu, line 95: Why not EIP-1820 so any contract that supports arbitrary execution can also work?

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+##### Contract
+
+Must implement latest [EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md) with `MAGICVALUE = 0x1626ba7e`.
+
+##### User agent
+
+A user agent intended to work with the contract MUST generate signatures over a EIP-191 hash of a regular Keccak256 hash of the challenge message.
+
+## Rationale
+[EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md) has done a great work with standardising the signature validation method for contracts. This proposal builds on top of it a process for dapp-wallet authentication.
```
MicahZoltu, line 91:
```suggestion
TBD
```
Rationale is a section for explaining why you made some specific choices that you made.  This statement seems closer to a motivation... though it doesn't really add any value so I recommend deleting it entirely.

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+### Wallet
+
+#### External wallet
+Any software agents managing external wallets are not required to make any changes to continue to work with this process.
+
+#### Contract wallet
+
+##### Contract
+
+Must implement latest [EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md) with `MAGICVALUE = 0x1626ba7e`.
```
MicahZoltu, line 84: Consider mentioning in the rationale where this number comes from and why it was chosen (could just be a one-liner if it is obvious).

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Abstract
+The authentication process starts with the dapp client component requesting a message signature from the wallet.
+The client then proceeds to send the result to the dapp backend component along with the requested address to be used for authentication. The dapp backend recovers a public key from the signature, and checks if it has authorized control over the requested address. This check is done under consideration that the address may represent either an external wallet or a contract wallet. This process works with external wallets, as well with contract wallets that support the `isValidSignature()` method specified in this document.
+
+## Motivation
+Dapps frequently offer a customised off-chain user experience in addition to their smart-contract interface. For example, a dapp may provide a push notification feature to their users, allowing them to stay notified about successful state changes associated with their public addresses. For these type of features, a dapp needs a way to assert that a user has authorized control over the public address associated with their account.
```
MicahZoltu, line 27: I would like to see a section added to the motivation section that explains why it is reasonable for a dapp to "authenticate" a user before letting them view things.  If the data is coming from Ethereum, then it is all public and it gives the user a false sense of security.

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
...
+- `external wallet` An externally owned [account](https://github.com/ethereum/wiki/wiki/White-Paper#ethereum-accounts), controlled by a private key. Currently, most on-chain assets are owned by such accounts. A common example for an external wallet are the wallets generated by MetaMask.
+- `authorized signer` An entity is considered to have authorized control over a wallet if either:
+  - It produced a signature of which the recovered address matches the wallet address.
+  - It produced a signature of which the contract at the wallet address responds with the magic value `0x1626ba7e` to the `isValidSignature()` contract call of [EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md).
+
```
MicahZoltu, line 21: This isn't a section in the [template](https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md).  Recommend moving it to the `## Specification`.

EIPS/eip-1654.md
```diff
@@ -0,0 +1,107 @@
+---
+eip: 1654
+title: Off-chain dapp-wallet authentication process with contract wallets support
+author: Maor Zamski (@pazams), Christopher Scott (@chrisaxiom)
+discussions-to: https://github.com/ethereum/EIPs/issues/1654
+status: Draft
+type: Meta
```
MicahZoltu, line 7: At best this is `Informational`.  I'm personally pretty against Informational EIPs in general though and think this is just a Good Idea which isn't appropriate for an EIP.

MicahZoltu (changes requested):
Most of the content in this specification doesn't seem like stuff that needs to be a standard.  While it may be a Good Idea, not all Good Ideas need to be standardized.  If two dapps each utilize a different mechanism for signature validation that is OK.

Why is more than just EIP-1271 required here?  If a contract implements 1271 then it can validate arbitrary signatures and there is no need to have all apps verify the same signature.  Each one can verify something else and that is OK.

pazams:
@MicahZoltu 

> Most of the content in this specification doesn't seem like stuff that needs to be a standard. While it may be a Good Idea, not all Good Ideas need to be standardized. If two dapps each utilize a different mechanism for signature validation that is OK.

It is ok, but since a SC needs to know which to support in advance, it's better if everyone can agree instead of having separate clusters of dapps and wallets that work together for the purpose of authentication. That will be the blockchain equivalent of the "browser wars" era of the web.

> Why is more than just EIP-1271 required here? If a contract implements 1271 then it can validate arbitrary signatures and there is no need to have all apps verify the same signature. Each one can verify something else and that is OK.

Indeed this proposal is different than most other EIPs and the debate of its worthiness is important.
This is foremost an offchain algorithm that has been established as a de-facto practice by a wide range of apps listed at the top of this pull request.

The reason why so many apps were able to come together and use the same process was because of this draft.

Now, when a new app developer reaches out to Etheremon to enquire "Hey Etheremon, how did you guys manage to have offchain authentication to work both with Smart Contract wallets and "regular" EOA wallets".

Then Etheremon devs can reply with either:

- Oh well, we recovered the public key from signature, then if it matches with authentication address we are like "great", but if it doesn't, we take the wallet address of the public key, and make a contract call to isValidSignature as specified by EIP 1271, and see if it matches there.

- We use EIP 1654.

If Ethereum is a dev-first community, the second option would make sense as it will make life easier for them. If this proposal needs to be closed, I just hope people can still reach the information needed via the issue https://github.com/ethereum/EIPs/issues/1654

MicahZoltu:
> Now, when a new app developer reaches out to Etheremon to enquire "Hey Etheremon, how did you guys manage to have offchain authentication to work both with Smart Contract wallets and "regular" EOA wallets".

The third option, which I recommend, is "See https://awesome.blog.example.com/how-to-have-off-chain-authentication-work-with-both-contract-and-eoa"

Souptacular:
I agree with @MicahZoltu. This EIP delves beyond what Meta EIPs traditionally are. When we talk about "process" in Meta EIPs we are talking about human processes and changes in how things are determined, primarily from a hard fork and EIP process perspective (and we are starting to move away from Meta EIPs that involve hard forks now that we have https://github.com/ethereum/eth1.0-specs/ and https://github.com/ethereum/eth2.0-specs). This EIP is more technical than "process oriented" based on the EIP repo definition of process as it is implied by previous EIPs so I don't think we should merge this.

PR #2998, opened 2020-09-24
EIP-2929: clarifications

holiman:
Contains a few clarifications

- Request from @shamatar  to be more clear in the abstract that some ops become cheaper, 
- Clarification (maybe unnecessary), due to an error that was present in geth, found by @shemnon  
- Finding by @shemnon that geth incorrectly had not updated a composite variable


vbuterin:
Looks good to me! :+1:

PR #3001, opened 2020-09-24
EIP2333 - ETH2 keygen: test vector 0 wasn't updated in #2985

mratsim:
## What was wrong

Test vector 0 was not updated in https://github.com/ethereum/EIPs/pull/2985

https://github.com/ethereum/EIPs/pull/2985/files#diff-8caacf51cadcfbe94a39cf606351c359R243

## Correction

Test vector 0 is actually detailed with intermediate value at the bottom of the EIP so this merely copies that information over: https://github.com/ethereum/EIPs/pull/2985/files#diff-8caacf51cadcfbe94a39cf606351c359R294

The correction was checked against the reference implementation at eth2.0-deposits-cli which uses the correct vector:
https://github.com/ethereum/eth2.0-deposit-cli/pull/108/files#diff-909cbf3d1a7e55f1731e90bc1f9a7ad9R5

### DIFF
--- a/EIPS/eip-2333.md
+++ b/EIPS/eip-2333.md
@@ -241,9 +241,9 @@ There are no major backwards compatibility issues brought upon by this EIP as it
 
 ```text
 seed = 0xc55257c360c07c72029aebc1b53c05ed0362ada38ead3e3e9efa3708e53495531f09a6987599d18264c1e1c92f2cf141630c7a3c4ab7c81b2f001698e7463b04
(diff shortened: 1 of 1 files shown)


CarlBeek (approved):
Thanks @mratsim! As you mentioned this test is a restatement of the `intermediate test` further down. The idea is that if you past `test case 0`, you needn't implement tests for all of the intermediate values.

MicahZoltu:
Kicking CI to try to fix the greeter problem.

PR #3013, opened 2020-09-30
Update eip-173.md, makes its status Final

mudgen:


eip-automerger:

 - Trying to change EIP 173 state from Last Call to Final

lightclient (approved):

MicahZoltu (approved):
Can you remove this from the 165 interface comment:
>  This function uses less than 30,000 gas.
Gas costs can change (and sometimes dramatically) so there is no guarantee that this statement will remain true forever.

Otherwise I think this is good to go as final.

PR #3023, opened 2020-10-04
Informational: Contract Wallet Detection

3esmit:
## Simple Summary

Provides a registry for checking if a contract is a user wallet.

## Abstract

The wallet software should warn users of common mistakes, such as transferring ERC20 to a contract which is not intended to handle it. User wallet contracts can be detected off-chain through analysis of common code hashes used for most users. 

## Motivation

Huge amounts of value have been lost due simple mistakes as sending all your tokens to the token contract itself, a common mistake by new (or tired) users, to paste the token address in the destination field, instead of pasting it on the wallet configuration. 

Also, new users, coming from less advanced blockchains where all addresses are controlled by a person, willing to participate in a ethereum project, ends up sending value to the wrong contract, and/or directly (without using ERC20 approval design) to the project's contract.

## Specification

The contract should returns "Wallet" if the contract is registered as a "user wallet contract", and thus be capable move tokens received on that address.

Undefined code hashes are considered as proxies by default, as proxies can have unique code per implementation (see ERC1167).

The registry could also check for certain ERC-1820 signature to accept as a contract wallet.

A democracy or a multisig between the ethereum developer community (Status, MetaMask, MyCrypto, MyEtherWallet, Argent, Gnosis, Aragon, Decentraland, and others), would control the registry, and would be able to approve the inclusion in a codehash->type map for defining all commonly used wallet contracts. 

A script that connects to a full node would scan for combinations of function signatures inside contract code in state to guide controllers to the inclusion of all possible user wallets, even those with small tweaks that make them have a different code hash. 

## Rationale

A smart contract would allow wallets in a single request to find out if a destination address is safe to handle tokens. 
Sharing the smart contract between all major wallets creates a standard that contributes with a better ethereum UX. 

A democracy could replace the multisig, where the voters could be the former participants of the multisig, or any other more democratic method. 
To provide backwards compatibility with all possible wallet contracts, a codehash map is the best solution. 

## Backwards Compatibility

At the current time, externally owned addresses (code size == 0) would still be considered safe, but with the mature of this specification, and the future of ethereum with account abstraction, all wallets would be a smart contract, and sending tokens to a mistyped address would also be covered. 

Some Proxy standards are not compatible by default, as eip-1822, however this is easily manageable by the implementation contract, which can expose the function, or register itself as a token receiver through EIP-820.

## Security Considerations

The worst case scenario for this tool would be false positives (warning about risk of loss on a wallet contract) or false negatives (not warning when user is about to send tokens to an unrecoverable address), which would still be an improvement over current situation

## Test Cases

TBD

## Implementation

WIP
TODO ERC1820 support or similar

```solidity
pragma solidity ^0.7.2;

interface ERC897 {
    function implementation() external view returns (address);
}

interface NonStandard {
    function masterCopy() external view returns (address); //gnosis
    function currentContract() external view returns (address); //decentraland
    function contractImplementation() external view returns (address); //lime
    //add others
}

/**
 * @author Ricardo Guilherme Schmidt (Status Research & Development GmbH)
 * @notice Registry of Wallet Codehashes and Wallet capability checker
 */
contract CodehashType {
    event ControllerChanged(address newController);
    event CodehashTypeSet(bytes32 codehash, CodeType indexed codeType);
    enum CodeType { Undefined, Unknown, NoCode, Wallet, Proxy }
    
    address public controller;

    mapping(bytes32 => CodeType) public codeTypeMap;

    modifier onlyController(){
        require(msg.sender == controller, "Unauthorized");
        _;
    }

    /**
     * @param nullHash codehash of a externally owned address (EOA)
     */
    constructor(bytes32 nullHash) {
        codeTypeMap[nullHash] = CodeType.NoCode; 
    }

    /** 
     * @param sample A deployed contract to define codehash type.
     */
    function setCodehash(address sample, CodeType codeType) external onlyController {
        require(isContract(sample), "No code");
        bytes32 codeHash = getCodeHash(sample);
        codeTypeMap[codeHash] = codeType;
        emit CodehashTypeSet(codeHash, codeType);
    }

    function changeController(address newController) external onlyController {
        controller = newController;
    }

    /** 
     * @notice Search for thecode type of current installed logic
     * @param test An address in test.
     */
    function searchCodeType(address test) public view returns(CodeType codeType) {
        codeType = codeTypeMap[getCodeHash(test)];
        if(codeType == CodeType.Proxy || codeType == CodeType.Undefined){
            address implementation = getImplementation(test);
            if(implementation == address(0) || implementation == test){
                return CodeType.Unknown;
            }
            codeType = searchCodeType(implementation);
        }
    }

    /** 
     * @notice Search for proxy implementation
     * @param proxy Proxy in lookup.
     */
    function getImplementation(address proxy) public view returns(address) {
        try ERC897(proxy).implementation() returns (address implementation) {
            return implementation;
        } catch {
            
        } try NonStandard(proxy).masterCopy() returns (address implementation) {
            return implementation;
        } catch {
            
        } try NonStandard(proxy).currentContract() returns (address implementation) {
            return implementation;
        } catch {
            
        } try NonStandard(proxy).contractImplementation() returns (address implementation) {
            return implementation;
        } catch {
            
        } //** other proxy detection */
        try this.ERC1167implementation(proxy) returns (address implementation) {
            return implementation;
        }  catch {

        }
        return address(0);
    }

    /** 
     * @notice Extracts implementation of a ERC1167 address
     * @param proxy in lookup
     */
    function ERC1167implementation(address proxy) public view returns (address implementation) {     
        assembly {
            extcodecopy(proxy, implementation, 10, 20)
        }
        require(isContract(implementation), "Not ERC1167");
    } 

    /** 
     * @notice Get current codehash of source
     * @param source address to load code from
     */
    function getCodeHash(address source) public view returns (bytes32 codeHash) {
        assembly { codeHash := extcodehash(source) }
    }

    function isContract(address test) public view returns (bool){
        uint256 size;
        assembly {
            size := extcodesize(test)
        }
        return size > 0;
    }
}
```

## Copyright

### DIFF
--- a/EIPS/eip-3023.md
+++ b/EIPS/eip-3023.md
@@ -21,7 +21,7 @@ The wallet software should warn users of common mistakes, such as transferring E
 
 Huge amounts of value have been lost due simple mistakes as sending all your tokens to the token contract itself, a common mistake by new (or tired) users, to paste the token address in the destination field, instead of pasting it on the wallet configuration. 
 
-Also, new users, coming from less advanced blockchains where all addresses are controlled by a person, willing to participate in a ethereum project, ends up sending value to the wrong contract, and/or directly (without using ERC20 approval design) to the project's contract.
+New users end up sending value to the wrong contract, and directly (without using ERC20 approval design) to the project's contract.
 
 ## Specification
 
--- a/EIPS/eip-3023.md
+++ b/EIPS/eip-3023.md
@@ -6,7 +6,7 @@ discussions-to: https://github.com/ethereum/EIPs/pull/3023
 status: Draft
 type: Informational
 created: 2020-10-04
-requires: 20, 897, 1820 
+requires: 20, 897, 1820
 ---
 
 ## Simple Summary
--- a/EIPS/eip-3023.md
+++ b/EIPS/eip-3023.md
@@ -2,7 +2,7 @@
 eip: 3023
 title: Contract Wallet Detection
 author: Ricardo Guilherme Schmidt (@3esmit)
-discussions-to: https://github.com/ethereum/EIPs/pull/3023
+discussions-to: https://ethereum-magicians.org/t/informational-3023-contract-wallet-detection/4831
 status: Draft
 type: Informational
 created: 2020-10-04
--- a/EIPS/eip-3023.md
+++ b/EIPS/eip-3023.md
@@ -1,8 +1,8 @@
 ---
-eip: <to be assigned>
+eip: 3023
 title: Contract Wallet Detection
 author: Ricardo Guilherme Schmidt (@3esmit)
-discussions-to: <to be assigned>
+discussions-to: https://github.com/ethereum/EIPs/pull/3023
 status: Draft
 type: Informational
 created: 04-Oct-2020
--- a/EIPS/eip-3023.md
+++ b/EIPS/eip-3023.md
@@ -5,7 +5,8 @@ author: Ricardo Guilherme Schmidt (@3esmit)
 discussions-to: https://github.com/ethereum/EIPs/pull/3023
 status: Draft
 type: Informational
-created: 04-Oct-2020
+created: 04-10-2020
+requires: 20, 897, 1820 
 ---
 
 ## Simple Summary
--- a/EIPS/eip-3023.md
+++ b/EIPS/eip-3023.md
@@ -5,7 +5,7 @@ author: Ricardo Guilherme Schmidt (@3esmit)
 discussions-to: https://github.com/ethereum/EIPs/pull/3023
 status: Draft
 type: Informational
-created: 04-10-2020
+2020-10-04
 requires: 20, 897, 1820 
 ---
 
@@ -22,7 +22,6 @@ The wallet software should warn users of common mistakes, such as transferring E
 Huge amounts of value have been lost due simple mistakes as sending all your tokens to the token contract itself, a common mistake by new (or tired) users, to paste the token address in the destination field, instead of pasting it on the wallet configuration. 
 
 Also, new users, coming from less advanced blockchains where all addresses are controlled by a person, willing to participate in a ethereum project, ends up sending value to the wrong contract, and/or directly (without using ERC20 approval design) to the project's contract.
-
 
 ## Specification
 
--- a/EIPS/eip-3023.md
+++ b/EIPS/eip-3023.md
@@ -5,7 +5,7 @@ author: Ricardo Guilherme Schmidt (@3esmit)
 discussions-to: https://github.com/ethereum/EIPs/pull/3023
 status: Draft
 type: Informational
-2020-10-04
+created: 2020-10-04
 requires: 20, 897, 1820 
 ---
 
--- a/EIPS/eip-contract-wallet-detection.md
+++ b/EIPS/eip-contract-wallet-detection.md
@@ -1,0 +1,193 @@
+---
+eip: <to be assigned>
+title: Contract Wallet Detection
+author: Ricardo Guilherme Schmidt (@3esmit)
+discussions-to: <to be assigned>
+status: Draft
+type: Informational
+created: 04-Oct-2020
+---
+
+## Simple Summary
+
+Provides a registry for checking if a contract is a user wallet.
+
+## Abstract
+
+The wallet software should warn users of common mistakes, such as transferring ERC20 to a contract which is not intended to handle it. User wallet contracts can be detected off-chain through analysis of common code hashes used for most users. 
+
+## Motivation
+
+Huge amounts of value have been lost due simple mistakes as sending all your tokens to the token contract itself, a common mistake by new (or tired) users, to paste the token address in the destination field, instead of pasting it on the wallet configuration. 
+
+Also, new users, coming from less advanced blockchains where all addresses are controlled by a person, willing to participate in a ethereum project, ends up sending value to the wrong contract, and/or directly (without using ERC20 approval design) to the project's contract.
+
+
+## Specification
+
+The contract should returns "Wallet" if the contract is registered as a "user wallet contract", and thus be capable move tokens received on that address.
+
+Undefined code hashes are considered as proxies by default, as proxies can have unique code per implementation (see ERC1167).
+
+The registry could also check for certain ERC-1820 signature to accept as a contract wallet.
+
+A democracy or a multisig between the ethereum developer community (Status, MetaMask, MyCrypto, MyEtherWallet, Argent, Gnosis, Aragon, Decentraland, and others), would control the registry, and would be able to approve the inclusion in a codehash->type map for defining all commonly used wallet contracts. 
+
+A script that connects to a full node would scan for combinations of function signatures inside contract code in state to guide controllers to the inclusion of all possible user wallets, even those with small tweaks that make them have a different code hash. 
+
+## Rationale
+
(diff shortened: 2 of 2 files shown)


3esmit:
## Discussion: Should this EIP also cover this functionality: https://info.etherscan.com/ethprotect/ ?

It's simple to include another CodeTypes for things like "Phishing", "Honey Pot", "Pyramid" and more types, however its also easy for this type of attackers evade this detection, as the detection rules are public, they can simple make their contract/address change, being a for ever cat-mouse chase.

Personally I don't think is the scope of this EIP, however I think is something interesting to have, I am just not sure if this detection mechanism would work well being open source/public.

3esmit:
This twitter account posts everytime someone do this type of mistakes https://twitter.com/TokenOops

MicahZoltu:
This feels very much like a Good Idea rather than a Standard.

It is well known that I'm generally against Informational EIPs because it is a horribly slippery slope into "EIPs repository as documentation for random Ethereum applications/projects".  Not all editors agree with me on this stance, but I'm pretty strongly against using Informational EIPs really at all.

3esmit:
@MicahZoltu 

I think this is more than a documentation for random Ethereum projects. Account contacts are a common design for most Ethereum projects, and right now wallets have no standard way of identify what is a wallet contract. 

I think is informational because it informs wallet developers, both contract wallet developers and wallet software developers, a method to identify and define wallet contracts. 

Also, I wanted it to be part of EIP because I consider it an "Ethereum Improvement Proposal", as it improves the Ethereum ecossystem to avoid common mistakes related to ERC20, and therefore something that should be followed by other wallets as well. Also because I think that other Ethereum projects might want to collaborate with this definition.  

Do you think this would fit better as an ERC?

MicahZoltu:
Sorry, I didn't mean this EIP was "documentation for a random project".  I meant that this EIP is somewhere on the fuzzy spectrum between "technical standard" and "personal project documentation".  I was using it as an example of the slippery slope I see between those two extremes.

You have been caught up in my failing quest to get rid of informational EIPs entirely.  I *personally* want the EIPs repository to be a standards repository, not a Good Idea repository.  The reason for this is a fear of EIP repository scope creep and an increased drain on the already incredibly scarce resource that is EIP editors.  There is no schelling fence between Informational EIPs like this one, or others currently proposed, and "I want to store my project's documentation in the EIP repository".

The best schelling fence would be to assert that EIPs repository is for technical standards only, which would mean dropping the Informational category.

You are correct that per current repository rules this is reasonable and valid, I just dislike the current rules and when I see EIPs like this I bring up my distaste for the current rules as a reminder to everyone involved in the process of my preference to change these rules.

3esmit:
@MicahZoltu I think that you should create a Meta EIP to discuss the concerns on Informational EIPs, and what would be done with the current 4 Informational EIPs and this one. 
Let me know if you think there is a better way of classifying this idea than Informational EIP.

MicahZoltu:
My preference is that Good Ideas are documented in a book, blog post, reddit thread, Stack Exchange Q&A, news article, etc.  I'm not a fan of standards repositories like RFCs, IEEE, W3C, etc. including Good Ideas.

Keep in mind, while I'm the editor that reviews the most PRs currently, my say is certainly not final and on this topic especially I have been overruled many times in the past.  😄 If I ignore my personal bias, I would say that you can probably get something like this included by another editor without too much issue.

> I think that you should create a Meta EIP to discuss the concerns on Informational EIPs

I also have a dislike for Meta EIPs.  😬

EIPS/eip-3023.md
```diff
@@ -0,0 +1,194 @@
+---
+eip: 3023
+title: Contract Wallet Detection
+author: Ricardo Guilherme Schmidt (@3esmit)
+discussions-to: https://github.com/ethereum/EIPs/pull/3023
+status: Draft
+type: Informational
+created: 04-10-2020
```
MicahZoltu, line 8:
```suggestion
2020-10-04
```
ISO-8601

EIPS/eip-3023.md
```diff
@@ -0,0 +1,194 @@
...
+
+The wallet software should warn users of common mistakes, such as transferring ERC20 to a contract which is not intended to handle it. User wallet contracts can be detected off-chain through analysis of common code hashes used for most users. 
+
+## Motivation
+
+Huge amounts of value have been lost due simple mistakes as sending all your tokens to the token contract itself, a common mistake by new (or tired) users, to paste the token address in the destination field, instead of pasting it on the wallet configuration. 
+
+Also, new users, coming from less advanced blockchains where all addresses are controlled by a person, willing to participate in a ethereum project, ends up sending value to the wrong contract, and/or directly (without using ERC20 approval design) to the project's contract.
+
+
```
MicahZoltu, line 26:
```suggestion
```
Formatting consistency.

EIPS/eip-3023.md
3esmit: Ah, great catch. I used my native format instead of the universal format. Thank you.

EIPS/eip-3023.md
```diff
@@ -0,0 +1,193 @@
+---
+eip: 3023
+title: Contract Wallet Detection
+author: Ricardo Guilherme Schmidt (@3esmit)
+discussions-to: https://github.com/ethereum/EIPs/pull/3023
+status: Draft
+type: Informational
+created: 2020-10-04
+requires: 20, 897, 1820
```
lightclient, line 9:
```suggestion
requires: 20, 897, 1820
```

lightclient (changes requested):

EIPS/eip-3023.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Abstract
+
+The wallet software should warn users of common mistakes, such as transferring ERC20 to a contract which is not intended to handle it. User wallet contracts can be detected off-chain through analysis of common code hashes used for most users. 
+
+## Motivation
+
+Huge amounts of value have been lost due simple mistakes as sending all your tokens to the token contract itself, a common mistake by new (or tired) users, to paste the token address in the destination field, instead of pasting it on the wallet configuration. 
+
+Also, new users, coming from less advanced blockchains where all addresses are controlled by a person, willing to participate in a ethereum project, ends up sending value to the wrong contract, and/or directly (without using ERC20 approval design) to the project's contract.
```
Souptacular, line 24: I would change to "end up sending" instead of "ends up sending". If you take all the extras away from the sentence it would say "New users end up sending value to the wrong contract" which sounds better than "ends up".

Souptacular (commented):
I agree with @MicahZoltu that this particular EIP falls into a weird spot. I don't think this is in the spirit of an Informational EIP for a few reasons:

- This seems more like an EIP because it is something that different groups could implement at a smart contract level, making it an ERC.
- This line stuck out to me from [this comment](https://github.com/ethereum/EIPs/pull/3023#issuecomment-703582960): `I think is informational because it informs wallet developers, both contract wallet developers and wallet software developers, a method to identify and define wallet contracts.` I feel like ERC's are informing developers in similar ways.

I predict Informational EIPs will be dropped altogether in the next few months after a few more EIPIP meetings since I've only met a few strong proponents for them, and therefore highly recommend you move this to a blog post, or _maybe_ an ERC. I'm open to hearing from other editors.

EIPS/eip-3023.md
3esmit: Thanks, I did as you suggested.

PR #3024, opened 2020-10-05
EIP 1271 Suggest Adding `getSignatureDataHash()`

Muhammad-Altabba:
Actually, in a project we are working on, in which we use a modified version of Gnosis Safe, need the hash to be created in a specific way.

And for this, we suggest that this EIP should probably include the method `getSignatureDataHash()` that the Smart Contract should implement for hashing. So, the caller will call this hashing function first to get the bytes32 hash out of the data. And then pass this hash to `isValidSignature()`.

### DIFF
--- a/EIPS/eip-1271.md
+++ b/EIPS/eip-1271.md
@@ -43,7 +43,7 @@ contract ERC1271 {
    * MUST NOT modify state (using STATICCALL for solc < 0.5, view modifier for solc > 0.5)
    * MUST allow external calls
    */ 
-    function getDataHash(bytes memory _data)
+    function getSignatureDataHash(bytes memory _data)
       public
       view
       returns (bytes32);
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - EIP 1271 requires approval from one of (@izqui, @phabc, @abandeali1, @catageek, @shrugs, @frangio)

3esmit:
You can retrieve codehash on other contract. 

```solidity
    /** 
     * @notice Get current codehash of source
     * @param source address to load code from
     */
    function getCodeHash(address source) public view returns (bytes32 codeHash) {
        assembly { codeHash := extcodehash(source) }
    }
```
This could be a library implemented somewhere. Maybe you would like to add comments to the EIP im working about detecting user contract wallets, see #3023

Muhammad-Altabba:
Sorry @3esmit, this is a totally different something. 

What we are trying to add to this EIP is a method name that can get the Hash of some arbitrary **data**, not the Hash of the smart contract code.

Thanks,

Muhammad-Altabba:
Hello @izqui, @PhABC, @abandeali1, @catageek, @shrugs and @frangio,
What do you think about this?
Many thanks,

Muhammad-Altabba:
The pull request is ready for review.

ytbigvinny27 (approved):

Muhammad-Altabba:
Hello @izqui, @PhABC, @abandeali1, @catageek, @shrugs and @frangio,
What do you think about this?
Many thanks,

MicahZoltu:
@Muhammad-Altabba You may want to try reaching out ot some of the authors in other channels like Discord or Twitter.  I am beginning to suspect that all of them have GitHub notifications muted.

MicahZoltu:
Alternatively, you may want to consider just creating your own IP that is a copy of this one but with the changes you desire in it.  You'll need to give it a new number, but you can then be an (or the only) author and push the EIP through the process.

frangio:
I haven't been deeply involved in this EIP even though I'm listed as one of the authors. Thus, I wouldn't feel comfortable either accepting or rejecting this proposal.

That said, I'll give my opinion.

First of all I think this suggestion is missing an explanation and discussion about the requirements that motivated it. In the case that you mentioned, what is the specific way that the hash needs to be created? Is it really necessary to add a function in the target contract? Could this be solved by encoding something into the signature?

In any case, I think EIP1271 has had some adoption mainly because it is extremely simple and covers a very basic need. Being this simple, it also has gaps and flaws, and you may have run into one here. But from the information currently available it doesn't look like it warrants changing the EIP, especially given that it's already quite an old one, even though it's still technically a Draft.

I would instead suggest, as @MicahZoltu said, to create a new EIP that is an extension of this one, explains the motivation, and addresses backwards compatibility concerns.

PR #3026, opened 2020-10-06
BW6-761 curve operations

yelhousni:
This EIP adds a set of precompiles to efficiently implement operations over BW6-761 curve, which is the most efficient outer curve to BLS12-377 (EIP-2539).


yelhousni:
Note: gas cost is still WIP as there are many libs implementing the curve and we want to see and compare timings.

EIPS/eip-3026.md
```diff
@@ -0,0 +1,287 @@
...
+
+Also referred as the "zero point". For BW6-761 (`y^2=x^3-1`) and its M-twisted curves (`y^3=x^3+4`), the point with coordinates `(0, 0)` (formal zeros in Fp) is *not* on the curve, and so the encoding of `(0, 0)` is used as a convention to encode the point at infinity.
+
+##### Encoding of scalars for multiplication and multiexponentiation operations:
+
+For multiplication and multiexponentiation operations, a scalar is encoded as `64` bytes by performing BigEndian encoding of the corresponding (unsigned) integer.
+
+Note that the main subgroup order for BW6-761 is actually only `377` bits (`48` bytes), but an encoding of `64` bytes has been chosen to have a `32`-byte-aligned ABI (representable as e.g. `bytes32[2]` or `uint256[2]`).
+
+The corresponding integer is **not** required to be less than or equal to the main subgroup order.
```
MicahZoltu, line 106:
```suggestion
The corresponding integer **MAY** be greater than the main subgroup order.
```
RFC-2119

EIPS/eip-3026.md
```diff
@@ -0,0 +1,287 @@
...
+
+Error cases:
+- Any of the G1 or G2 points being not on the curve
+- Any of the G1 or G2 points being not in the correct subgroup
+- Input has invalid length
+- Field elements encoding rules apply (obviously)
+
+#### Prevention of DDoS on error handling
+
+This precompile performs extensive computations and in case of any errors during execution it MUST consume all gas from the the gas schedule for the corresponding operation.
```
MicahZoltu, line 183:
```suggestion
This precompile performs extensive computations and in case of any errors during execution it **MUST** consume all gas from the the gas schedule for the corresponding operation.
```

EIPS/eip-3026.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+`max_discount = 174`
+
+##### Pairing operation
+
+Average timing of 1000 random samples (random points in G1 and G2) for different number of pairs with linear lifting:
+Base cost of the pairing operation is `XXXXX*k + XXXXX` where `k` is a number of pairs.
+
+Each point (either G1 or G2) for which subgroup check is requested and performed adds the corresponding G1/G2 multiplication cost to it.
```
MicahZoltu, line 230: Most of this content should be in the Rationale section.  The specification section should only have the final list of gas costs and/or the algorithm for gas cost calculation (assuming the algorithm is part of consensus like for some of the other cryptographic precompiles).  How you decided on the gas costs is great content for the Rationale section.

EIPS/eip-3026.md
```diff
@@ -0,0 +1,287 @@
...
+- Golang implementation (Consensys/gurvy): https://github.com/ConsenSys/gurvy
+
+**Stand-alone implementation:**
+- Golang implementation with Intel assembly (Onur Kilic): https://github.com/kilic/bw6
+
+**Precompiles:**
+- OpenEthereum (EY/Parity): https://github.com/EYBlockchain/solidity-elliptic-curves
+
+**Sripts:**
+- SageMath and Magma scripts: https://gitlab.inria.fr/zk-curves/bw6-761/
```
MicahZoltu, line 279: This section is no longer required.  Consider just including a reference implementation written in a language that is easy for a human to read and is self contained.  We try to avoid external links whenever possible since they tend to 404 with time, so an inline or `../assets/eip-3026/reference_implementation.whatever` would be preferred to an external implementation.

EIPS/eip-3026.md
```diff
@@ -0,0 +1,287 @@
...
+- Subgroup check: `group_order * P = 0`
+- Trivial multiplication check: `1 * P = P`
+- Multiplication by zero: `0 * P = 0`
+- Multiplication by the unnormalized scalar `(scalar + group_order) * P = scalar * P`
+
+Required properties for pairing operation:
+- Degeneracy `e(P, 0*Q) = e(0*P, Q) = 1`
+- Bilinearity `e(a*P, b*Q) = e(a*b*P, Q) = e(P, a*b*Q)` (internal test, not visible through ABI)
+
+Test vector for all operations are expanded in this [gist](https://gist.github.com/shamatar/506ab3193a7932fe9302a2f3a31a23e8) until it's final.
```
MicahZoltu, line 263: Before this makes it to last call, the test cases should be inlined if at all possible.

MicahZoltu (approved):
I think this is generally fine to be merged as a draft.  Before it makes it to the next phase I recommend addressing the feedback I have left here.

Let me know if you want me to merge this as a draft now, or wait until after you have had a chance to apply the feedback.

EIPS/eip-3026.md
yelhousni: This sounds good, but I would like to keep the implementations links for now until we agree upon a reference implementation. There are few implementations with different performances.

EIPS/eip-3026.md
yelhousni: ok I'll provide that inlined before last call.

yelhousni:
@MicahZoltu I applied changes based on your feedback. I think we can merge as a Draft.

PR #3027, opened 2020-10-06
ERC-1363: Move to Final

vittominacori:
Moves ERC-1363 to Final.

### DIFF
--- a/EIPS/eip-1363.md
+++ b/EIPS/eip-1363.md
@@ -11,21 +11,24 @@ requires: 20, 165
 ---
 
 ## Simple Summary
-Defines a token interface for ERC-20 tokens that supports executing recipient code on `transfer` or spender code on `approve`.
+Defines a token interface for [ERC-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve`.
 
 ## Abstract
-The following describes standard functions a token contract and contracts working with specified token can implement to make a Token Payable.
+Standard functions a token contract and contracts working with tokens can implement to make a token Payable.
+
+`transferAndCall` and `transferFromAndCall` will call an `onTransferReceived` on a `ERC1363Receiver` contract.  
+
+`approveAndCall` will call an `onApprovalReceived` on a `ERC1363Spender` contract.
 
 ## Motivation
-This proposal allows to implement an ERC-20 compatible token that can be used for payments.
+There is no way to execute code after a [ERC-20](./eip-20.md) transfer or approval (i.e. making a payment), so to make an action it is required to send another transaction and pay GAS twice.
 
-It defines `transferAndCall` and `transferFromAndCall` functions that will call an `onTransferReceived` on a `ERC1363Receiver` contract.
-It defines `approveAndCall` functions that will call an `onApprovalReceived` on a `ERC1363Spender` contract.
+This proposal wants to make token payments easier and working without the use of any other listener. It allows to make a callback after a transfer or approval in a single transaction.
 
-This proposal is inspired by the ERC-721 `onERC721Received` and `ERC721TokenReceiver` behaviours. 
+This proposal has been inspired by the [ERC-721](./eip-721.md) `onERC721Received` and `ERC721TokenReceiver` behaviours. 
 
 ## Specification
-Every Payable Token compliant contract **MUST** implement the ERC-1363 interface as well as the ERC-20 and ERC-165 interfaces.
+Implementing contracts **MUST** implement the [ERC-1363](./eip-1363.md) interface as well as the [ERC-20](./eip-20.md) and [ERC-165](./eip-165.md) interfaces.
 
 ```solidity
 pragma solidity ^0.7.0;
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 1363 state from Last Call to Final

MicahZoltu (changes requested):
Abstract should describe the specification in a human readable way in relatively few words:
> Standard functions a token contract and contracts working with tokens can implement to make a token Payable.
> 
> `transferAndCall` and `transferFromAndCall` will call an `onTransferReceived` on a `ERC1363Receiver` contract.
> `approveAndCall` will call an `onApprovalReceived` on a `ERC1363Spender` contract.

----

Motivation should describe *why* this standard is valuable and how it can be used.  Most of what you have in there at the moment is more appropriate for Abstract.  Also, other EIPs should be linked to (see below).
> This proposal is inspired by the [ERC-721](./eip-721.md) `onERC721Received` and `ERC721TokenReceiver` behaviors.
> 
> There are many proposed uses of Ethereum smart contracts that can accept ERC-20 payments (i.e. to create a token payable crowdsale, selling services for tokens, paying invoices, making subscriptions, use them for a specific utility and many other purposes). 
> 
> This proposal wants to make token payments easier and working without the use of any other listener. It also allows to make a callback after a transfer in a single transaction.

----

In Specification section change
> Every Payable Token compliant contract **MUST** implement the ERC-1363 interface as well as the ERC-20 and ERC-165 interfaces.

to
> Implementing contracts **MUST** implement the [ERC-1363](./eip-1363.md) interface as well as the [ERC-20](./eip-20.md) and [ERC-165](eip-165.md) interfaces.

----

The Rationale section should describe why you made specific choices you did in the EIP.  For example, why did you choose a constant value of X, or why did you choose function name Y.  The content you have in there currently is more appropriate for the Motivation section and should be removed from the Rationale section.

----

In the backward compatibility section, be sure to link to external EIPs you reference.

----

Whenever possible, we prefer to avoid linking to external resources.  Can you inline the test cases and reference implementation into this EIP directly or include them in `../assets/eip-1363/whatever.ext` rather than linking to an external repository?

MicahZoltu:
Sorry for the delay on this, I like to be much more thorough with EIP reviews when things are moving to final and this kept getting pushed down my list due to an expectation that it would be a big time commitment.

vittominacori:
Just made the requested changes and added some esplicative texts.

Regarding tests and implementation i followed what done by EIP777. 
Tests and implementations are multiple files. Have you a sample of how they can be included in assets?

MicahZoltu:
> Regarding tests and implementation i followed what done by EIP777.
> Tests and implementations are multiple files. Have you a sample of how they can be included in assets?

You can put any files you want into a folder named after your EIP here: https://github.com/ethereum/EIPs/tree/master/assets  In your case, you would put all the files you want into https://github.com/ethereum/EIPs/tree/master/assets/eip-1363/.  There is an upper limit of course, we don't want a 10,000 file repository embedded into this project.  Generally speaking, I recommend removing any build tooling and *only* put the source files in here.

EIPS/eip-1363.md
```diff
@@ -184,23 +186,33 @@ interface ERC1363Spender {
...
+
+Examples could be 
+* to create a token payable crowdsale
+* selling services for tokens 
+* paying invoices
+* making subscriptions
+
+For these reasons it was named as **"Payable Token"**.
+
+Anyway you can use it for specific utilities or for any other purposes who require the execution of a callback after a transfer or approval received.
```
MicahZoltu, line 199: This should be moved to Motivation.  Motivation is where you describe why someone might want to implement this specification/standard.  Rationale is where you describe *specific* decisions made in the specification like "why use the constant 5 instead of 7" or "why name the function foo instead of bar".

Note: It is OK to have a mostly empty Rationale section if everything is straightforward and no one expressed any confusion over anything.  I think the comment you have about naming of the functions is enough.  Regarding the name "Payable Token", that isn't actually part of the specification (just the title of the EIP) so it isn't relevant to the rationale section.  :)

EIPS/eip-1363.md
```diff
@@ -184,23 +186,33 @@ interface ERC1363Spender {
...
 
 ## Test Cases
-[https://github.com/vittominacori/erc1363-payable-token/tree/master/test](https://github.com/vittominacori/erc1363-payable-token/tree/master/test)
+The [repository with the reference implementation](https://github.com/vittominacori/erc1363-payable-token) contains all the [tests](https://github.com/vittominacori/erc1363-payable-token/tree/master/test).
 
 ## Implementation
-[https://github.com/vittominacori/erc1363-payable-token/tree/master/contracts/token/ERC1363](https://github.com/vittominacori/erc1363-payable-token/tree/master/contracts/token/ERC1363)
+The GitHub repository [vittominacori/erc1363-payable-token](https://github.com/vittominacori/erc1363-payable-token) contains the reference implementation. The reference implementation is also available via [npm](https://www.npmjs.com/package/erc-payable-token) and can be installed with `npm install erc-payable-token`.
```
MicahZoltu, line 210: In case you weren't aware, Implementation and Test Cases sections are not required.  For a token like this, the value add the provide is limited IMO, so don't feel like you *must* get these integrated into this repository.  Instead, you could just remove these sections.  😄

MicahZoltu (changes requested):

EIPS/eip-1363.md
vittominacori: Can I add them as references?

MicahZoltu: We generally prefer to avoid that where possible.  The problems we are trying to solve here are:
1. **Advertising**.  People try to use EIPs to advertise their project by linking to their project from the EIP, hoping that this will drive traffic to their project.  By putting everything inside this repository, we don't have to worry about that problem.
2. **Licensing**.  Everything in the EIPs repository is CC0.  If you link out to something that is a different license (i.e., not public domain) it can create licensing problems for people down the road who aren't careful.  For example, if the CC0 licensed EIP links out to a GPL licensed implementation and in the course of reading the EIP the reader clicks through to the GPL code and reads it, there is now grounds for an argument that they have violated copyright when their code ends up looking like the GPL code they saw.

By having everything inside this repository (and thus licensed CC0), or simply left out of the EIP, we can avoid both of these problems.  It is valuable to have example code and test cases available to readers, but we don't need an entire test project or demo project.  The code alone is usually enough, and if someone really wants they can do a quick search online for EIP-1363 to find other people's implementations and test cases.

vittominacori:
Test and implementation removed.
Moved section to motivation.

For some reason the task failed.

PR #3030, opened 2020-10-08
EIP-3030: BLS Remote Signer HTTP API Standard

hermanjunge:
## Summary

This EIP defines a HTTP API standard for a BLS remote signer, consumed by validator clients to sign block proposals and attestations in the context of Ethereum 2.0 (eth2).

## Discussion Thread

https://ethereum-magicians.org/t/eip-3030-bls-remote-signer-http-api-standard/4810


hermanjunge:
Adding to the discussion:

- @AgeManning.
- @benjaminion.
- @djrtwo.
- @holiman.
- @paulhauner.
- @protolambda.
- @terencechain.
- @zedt3ster.


Hello! Please, do feel free to include more people ito this conversation :slightly_smiling_face:

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,404 @@
+---
+eip: <to be assigned>
+title: BLS Remote Signer HTTP API Standard
+author: Herman Junge <herman@sigmaprime.io>
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: ERC
```
MicahZoltu, line 8:
```suggestion
category: Interface
```

MicahZoltu (changes requested):
The frontmatter stuff must be changed, the rest is general feedback left here since there is no discussion link yet.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,404 @@
...
+--- | ---
+Code | `200`
+Content | `{"public_keys": "[public_key_hex_string_without_0x]"}`
+
+_or_
+
+Error | <br>
+--- | ---
+Code | `404`
+Content | `{"error": "No keys found in storage."}`
```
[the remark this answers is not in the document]
hermanjunge: Makes sense. Removed.

EIPS/eip-3030.md
```diff
@@ -0,0 +1,397 @@
...
+
+#### Authentication
+
+Can be accomplished through the use of an HTTP Request Header. There are several ways to negotiate and issue a valid token to authenticate the communication between the validator client and the remote signer, each of them with potential drawbacks (e.g replay attacks, challenges in distributing the token to the validator client, etc.). In general, any method of authentication must be combined with transport encryption to be effective.
+
+The operator can also implement network Access Control Lists (ACLs) between the validator client's network and the remote signer's network, reducing the attack surface by requiring a potential attacker to be positioned in the same network as the validator client.
+
+#### Pre-image validation
+
+A key feature for a remote signer, pre-image validation, implies that not only the `signingRoot`, but all the required elements needed to perform complete validation of the message, are sent through the wire to obtain a signature.
```
holiman, line 121: I'd argue that pre-image validation is such a core feature that the post-image shouldn't even be sent to the signer. The signer should always be forced to calculate the hash-to-sign locally, given the input + info about signing request (maybe using mimetypes to signal what type of signing is requested)

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,404 @@
...
+---
+
+### `GET /publicKeys`
+
+_**Responses**_
+
+Success | <br>
+--- | ---
+Code | `200`
+Content | `{"public_keys": "[public_key_hex_string_without_0x]"}`
```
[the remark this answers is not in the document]
hermanjunge: Yes. Replaced. Using `identifier` will clarify a lot the idea we are trying to convey.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,404 @@
...
+Code |  `404`
+Content | `{"error": "Key not found: <public_key_hex_string_without_0x>"}`
+
+---
+
+## Rationale
+
+### URL parameter `:public-key` without `0x`
+
+By relaxing the constraint of requiring hexadecimal input, clients can opt-in to using custom ids for their private keys, and perform the adequate retrieving and matching with the secret key within the signer. That is, if the remote signer implementer allows for this feature.
```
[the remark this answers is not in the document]
hermanjunge: Solved.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,404 @@
...
+
+URL Parameter | <br>
+--- | ---
+`:public_key` | `public_key_hex_string_without_0x`
+
+_**Request**_
+
+JSON Body | <br> | <br>
+--- | --- | ---
+`signingRoot` | **Required** | A string representation of the hexadecimal value to be signed
```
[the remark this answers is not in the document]
hermanjunge: I think this is the improvement needed.

JSON Body | <br> | <br>
--- | --- | ---
`signingRoot` | **Required** | A string representation of a SHA256 Hash.<br>The result of the Ethereum 2.0 function [`compute_signing_root`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#compute_signing_root).

EIPS/eip-3030.md
holiman: With this spec, as is, you cement the behaviour of a standard-compliant signer to sign blind checks. Which seems like a very bad idea, IMO

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,404 @@
...
+<br> | Optional | Any other field will be ignored by the signer
```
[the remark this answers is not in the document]
hermanjunge: Replaced by the snippet below.

### `GET /keys`

Returns the identifiers of the keys available to the signer.

_**Responses**_

Success | <br>
--- | ---
Code | `200`
Content | `{"keys": "[identifier]"}`

---

EIPS/eip-3030.md
```diff
@@ -0,0 +1,397 @@
...
+
+The following subsections discuss aspects to be considered by the client implementers relative to these subjects.
+
+#### Implementation of additional features
+
+From an API pipeline view, we have two nodes: The validator client (1) that makes requests to the remote signer (2). A more sophisticated chain can be built by introducing elements between these two nodes. Either by setting up reverse proxy services, or by adding plugins to the remote signer implementation.
+
+#### Authentication
+
+Can be accomplished through the use of an HTTP Request Header. There are several ways to negotiate and issue a valid token to authenticate the communication between the validator client and the remote signer, each of them with potential drawbacks (e.g replay attacks, challenges in distributing the token to the validator client, etc.). In general, any method of authentication must be combined with transport encryption to be effective.
```
holiman, line 114: There is no need for authentication if the requestor is not privileged. And it wouldn't be privileged if you removed the need for the signer to trust the `signingRoot`. 
So in essence, you now force the validator to store some sealer-credential, instead of a sealing-key -- just swapping one set of secrets for another.

hermanjunge: I agree. This rationale should be more emphatic on the point about swapping of secrets, since it beats the purpose of this API. Please notice that the following paragraph advises the reader to explore the use of network Access Control Lists (ACLs).

EIPS/eip-3030.md
hermanjunge: We can do the following then:

- Add as **required** fields `domain`, the `type` of message to be produced (randao, block, or attestation), and the specific `data` per type of message.
- It is responsibility of the Remote Signer Client to perform the adequate checks on the received payload in order to admit it and process the signature.

holiman: You can also _remove_ the `signingRoot` parameter. Btw, can you point me towards info about what the `signingRoot` consists of? Is it the ssz-hash of a full block? or a header? How large is the preimage?

hermanjunge: The `signingRoot` is the string representation of a SHA256 Hash. The result of the Ethereum 2.0 function [`compute_signing_root`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#compute_signing_root).

holiman: I see. So, in practice, what's the size of the `ssz_object` ? Why not send it in it's entirety? 
And depending on the size of that -- if it's extremely large, perhaps only portions + proofs/hashes for missing parts?

terencechain: Fyi this is Prysm's definition. We have a field for the beacon object for additional verifications such as slashing prevention.

https://github.com/prysmaticlabs/prysm/blob/master/proto/validator/accounts/v2/keymanager.proto#L53

It's not a mandatory field, the remote can ignore it

EIPS/eip-3030.md
```diff
@@ -0,0 +1,371 @@
...
+```
+
+## Implementation
+
+Repository Url | Language | Organization | Commentary
+--- | --- | --- | ---
+[BLS Remote Signer](https://github.com/sigp/rust-bls-remote-signer) | Rust | Sigma Prime | Supports proposed specification.
+[Web3signer](https://github.com/PegaSysEng/web3signer) | Java | PegaSys | Supports proposed specification, although with [slightly different methods](https://pegasyseng.github.io/web3signer/web3signer-eth2.html):<br>{`/sign` => `/api/v1/eth2/sign`, `/publicKeys` => `/api/v1/eth2/publicKeys`}.
+
+The Prysm client supports a [Remote Signing Wallet](https://docs.prylabs.network/docs/wallet/remote/), however its API requires using gRPC as transport.
```
terencechain, line 350: Prysm supports both gRPC and JSON over HTTP. There's a grpc<->json/http proxy. We will update our documentations

EIPS/eip-3030.md
```diff
@@ -0,0 +1,371 @@
...
+An attacker can repudiate a sent message. | Implement logging in the signer. Enhance it by sending logs to a syslog box.
+An attacker can disclose the contents of a private key by retrieving the key from storage. | Storage in Hardware security module (HSM).<br>_or_<br>Storage in Secrets management applications (e.g. Hashicorp Vault).
+An attacker can eavesdrop on the uploading of a secret key. | Upload the keys using a secure channel, based on each storage specification.
+An attacker can eavesdrop on the retrieval of a key from the remote signer. | Always pass the data between storage and remote signer node using a secure channel.
+An attacker can dump the memory in the remote signer to disclose a secret key. |  Prevent physical access to the node running the remote signer.<br>_or_<br>Prevent access to the terminal of the node runnig the remote signer: Logs being sent to a syslog box. Deployments triggered by a simple, non-parameterized API.<br>_or_<br>Implement zeroization of the secret key at memory.<br>_or_<br>Explore the compilation and running of the remote signer in a Trusted execution environment (TEE).
```
MicahZoltu, line 367:
```suggestion
An attacker can dump the memory in the remote signer to disclose a secret key. |  Prevent physical access to the node running the remote signer.<br>_or_<br>Prevent access to the terminal of the node running the remote signer: Logs being sent to a syslog box. Deployments triggered by a simple, non-parameterized API.<br>_or_<br>Implement zeroization of the secret key at memory.<br>_or_<br>Explore the compilation and running of the remote signer in a Trusted execution environment (TEE).
```
Spelling errors block merges: `runnig ==> running`

EIPS/eip-3030.md
```diff
@@ -0,0 +1,371 @@
+---
+eip: 3030
+title: BLS Remote Signer HTTP API Standard
```
MicahZoltu, line 3:
```suggestion
title: BLS Remote Signer HTTP API
```
Putting the word "standard" in the title of a standard is redundant.  😄

EIPS/eip-3030.md
tzapu: there s one scenario in my head where having the signingRoot together with all the needed fields to recompute it might come in handy, tamper detection. if the sent root and the calculated root mismatch then you have a big problem

holiman: In that case, someone is tampering with the traffic. So if they can modify one (e.g. the signingRoot, to sign something else), why wouldn't they tamper with both? 
On the one hand, there's a vague argument "it might be handy". So the arguments _for_ including it are: 

- It may be handy, against a sloppy attack where the attacker modifies one but not both fields

The arguments _against_ including it are: 

- If it's included, a lazy implementation or insecurely configured signer can choose to blindly sign anything. 
- If we do not include it, we _force_ to sealer to _at least_ verify that it's a block being sealed, and not something else. 
- If we do not include it, since the selaer needs to parse the ssz-object, it becomes a very small step to also do at least some basic anti-slashing validations, such as "do not seal same height twice".

tzapu: well, you are right of course :D thanks , 
pushing for everything to get signer to not sign anything blindly should be the the goal

MicahZoltu (approved):
This looks good to merge as a draft.  @hermanjunge I encourage you to heed the advice given by Martin and continue to iterate on this in its draft form before submitting a move to Review.

PR #3041, opened 2020-10-13
Add basefee to eth_getBlockByHash

abdelhamidbakhta:
### Description
Add basefee field to `eth_getBlockByHash`.


EIPS/eip-draft_getBlockByHash_add_basefee.md
```diff
@@ -0,0 +1,94 @@
+---
+eip: <to be assigned>
```
MicahZoltu, line 2:
```suggestion
eip: 3041
```
Be sure to rename the file to `EIPS/eip-3041.md` as well.

EIPS/eip-draft_getBlockByHash_add_basefee.md
```diff
@@ -0,0 +1,94 @@
...
+title: Add basefee in eth_getBlockByHash response
+author: Abdelhamid Bakhta (@abdelhamidbakhta)
+discussions-to:
```
MicahZoltu, line 5: Once you update the number, I recommend creating a new discussion topic over at https://ethereum-magicians.org/ and pasting the link here.

abdelhamidbakhta: Done

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
+---
+eip: 3041
+title: Add basefee in eth_getBlockByHash response
+author: Abdelhamid Bakhta (@abdelhamidbakhta)
+discussions-to: https://ethereum-magicians.org/t/eip-3041-add-basefee-in-eth-getblockbyhash-response/4825
+status: Draft
+type: Standards Track
+category : Interface
+created: 2020-10-13
+requires: 1559
```
MicahZoltu, line 10:
```suggestion
requires: 1474, 1559
```
We should also try to get @rekmarks and @bitpshr to get 1474 into final so it doesn't block EIPs like this from becoming final.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+## Simple Summary
+Add basefee field in `eth_getBlockByHash` RPC endpoint response.
```
MicahZoltu, line 14:
```suggestion
Add basefee field to `eth_getBlockByHash` RPC endpoint response.
```

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Abstract
+[EIP-1559](/EIPS/eip-1559) introduces a base fee per gas in protocol.
+This value is maintained under consensus as a new field in the block header structure.
+This change should be reflected in RPC methods that return block header information.
```
MicahZoltu, line 19:
```suggestion
Adds `baseFee` property to the `eth_getBlockByHash` JSON-RPC request `result` object.  This property will contain the value of the base fee for any block after the EIP-1559 fork.
```
Abstract should just be a human-readable version of the specification that is particularly terse.  It won't be as detailed as the specification section, and it likely won't mention edge cases or discuss minor details about the specification.  It should target a human reader trying to get a high level understanding of the specification, while the specification section is targeting implementers.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Motivation
+Users may need value of the base fee at a given block. Base fee value is important to make gas price predictions more accurate.
```
MicahZoltu, line 22:
```suggestion
[EIP-1559](/EIPS/eip-1559) introduces a base fee per gas in protocol.
This value is maintained under consensus as a new field in the block header structure.
Users may need value of the base fee at a given block. Base fee value is important to make gas price predictions more accurate.
```
Consider moving these sentences from the Abstract to here.  I think they are valuable, just not Abstract material is all.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+
+### eth_getBlockByHash
+
+#### Description
+
+Returns information about a block specified by hash.
+
+#### Parameters
+
+Parameters remain unchanged.
```
MicahZoltu, line 34: It *may* be valuable to include the whole spec for `eth_getBlockByHash` here.  You can copy it from eip-1474 and just add in base fee.  This is optional feedback, I'm on the fence as to whether it is better to just include the change, or include the whole spec here.  When implementing from scratch, having everything in one place is very convenient.  When implementing on top of pre-existing code, just the changeset is convenient.  I'm not sure if we want to optimize for new clients or optimize for updating existing clients.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Backwards Compatibility
+Backwards compatible. Calls related to block prior to [EIP-1559](/EIPS/eip-1559) fork block will omit the base fee field in the response.
+
+
+## Implementation
+
+### Besu
+Available in master branch: https://github.com/hyperledger/besu
+
```
MicahZoltu, line 89:
```suggestion
```
Stylistic consistency.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+        "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
+        "uncles": []
+    }
+}
+```
+---
+
+## Backwards Compatibility
+Backwards compatible. Calls related to block prior to [EIP-1559](/EIPS/eip-1559) fork block will omit the base fee field in the response.
+
```
MicahZoltu, line 83:
```suggestion
```

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+        "transactions": []
+        "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
+        "uncles": []
+    }
+}
+```
+---
+
+## Backwards Compatibility
+Backwards compatible. Calls related to block prior to [EIP-1559](/EIPS/eip-1559) fork block will omit the base fee field in the response.
```
MicahZoltu, line 82: This should probably be part of the specification since interface changes are not necessarily in sync with consensus forks.  Something like this in the Specification section:
```
Every block returned by this endpoint whose block number is before the [EIP-1559](./eip-1559.md) fork block **MUST NOT** include a `baseFee` field.
Every block returned by this endpoint whose block number is on or after the [EIP-1559](./eip-1559.md) fork block **MUST** include a `baseFee` field.
```
It may still be worth mentioning this in the backward compatibility section, as it *is* related to backward compatibility, even if it is also explicitly part of the specification.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+#### Description
+
+Returns information about a block specified by hash.
+
+#### Parameters
+
+Parameters remain unchanged.
+
+#### Returns
+Add a new JSON field in the `result` object for block headers containing a base fee (post [EIP-1559](/EIPS/eip-1559) fork block).
```
MicahZoltu, line 37:
```suggestion
Add a new JSON field in the `result` object for block headers containing a base fee (post [EIP-1559](./eip-1559.md) fork block).
```
Please use this format for links to other EIPs, as it works in all known environments while other forms of the path break in some environments.

EIPS/eip-3041.md
MicahZoltu, line 89: My personal recommendation here is to not include the `Implementation` section.  We generally want to avoid linking to external resources in EIPs as they tend to go stale with time and eventually 404 (e.g., if besu is renamed or moves to a different GitHub organization, or switches to private repository).  Normally I would suggest including a reference implementation inline or as one or more files in `../assets/eip-3041/...` but in this case I think that would not be particularly useful and would be a lot of work, so just leaving the section out is probably the best course of action.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
+---
+eip: 3041
+title: Add basefee in eth_getBlockByHash response
```
MicahZoltu, line 3:
```suggestion
title: Adds `baseFee` to `eth_getBlockByHash`
```
The shorter the title the better!

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+[EIP-1559](/EIPS/eip-1559) introduces a base fee per gas in protocol.
+This value is maintained under consensus as a new field in the block header structure.
+This change should be reflected in RPC methods that return block header information.
+
+## Motivation
+Users may need value of the base fee at a given block. Base fee value is important to make gas price predictions more accurate.
+
+## Specification
+
+### eth_getBlockByHash
```
MicahZoltu, line 26:
```suggestion
### `eth_getBlockByHash`
```
This should render a bit nicer in most places.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+        "totalDifficulty":  "0x027f07",
+        "transactions": []
+        "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
+        "uncles": []
+    }
+}
+```
+---
+
+## Backwards Compatibility
```
MicahZoltu, line 81:
```suggestion
## Rationale
TBD

## Backwards Compatibility
```
Rationale is a required section.  It is the appropriate place to discuss the *why* of specific decisions that were made such as why a particular constant value was chosen, a particular name used, a particular layout picked, etc..  In this case, there probably isn't much interesting to talk about, but the section is still required I believe.

Maybe just mention that the addition of a single parameter instead of introducing a whole new endpoint was the simplest change that would be easiest to get integrated?  You could also (or alternatively) discuss why it was decided to have `baseFee` be omitted from the result for pre-1559 blocks rather than be included with a null or 0 value?

MicahZoltu (approved):
Only thing blocking this getting merged as draft is the inclusion of a `Rationale` section (which is required in all EIPs).  The rest of the feedback can be applied after it is merged to draft (though some is probably required prior to moving this EIP into Review).

EIPS/eip-3041.md
abdelhamidbakhta: I would personally optimize for updating existing clients. I will refer EIP-1474. If someone thinks i should put the full spec i will

EIPS/eip-3041.md
abdelhamidbakhta: Ok i will remove it

EIPS/eip-3041.md
abdelhamidbakhta: Sounds good.

EIPS/eip-3041.md
abdelhamidbakhta: Done

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Parameters
+
+Parameters remain unchanged.
+
+#### Returns
+For the full specification of `eth_getBlockByHash` see [EIP-1474](./eip-1474.md).
+Add a new JSON field to the `result` object for block headers containing a base fee (post [EIP-1559](./eip-1559.md) fork block).
+
+- {[`Quantity`](#quantity)} `baseFee` - base fee for this block
```
MicahZoltu, line 42:
```suggestion
- {[`Quantity`](./eip-1474.md#quantity)} `baseFee` - base fee for this block
```
This is being interpreted as a link `[title](link)` and it isn't a valid link because there is no `#quantity` anchor.  I'm guessing you copied this from 1474, so I think the fix here is to just change it to link to that other spec.

EIPS/eip-3041.md
```diff
@@ -0,0 +1,95 @@
+---
+eip: 3041
+title: Adds `baseFee` to `eth_getBlockByHash`
+author: Abdelhamid Bakhta (@abdelhamidbakhta)
+discussions-to: https://ethereum-magicians.org/t/eip-3041-add-basefee-in-eth-getblockbyhash-response/4825
+status: Draft
+type: Standards Track
+category : Interface
```
MicahZoltu, line 8:
```suggestion
category: Interface
```
CI is very picky about formatting of the frontmatter.  😬

PR #3072, opened 2020-10-26
Update eip-2612

MrChico:
Incorporates the suggestion of @wighawag to leave DOMAIN_SEPARATOR out of the scope of this standard, and points out a commonly voiced safety concern with respect to `permit` and future chain forks

### DIFF
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -43,15 +43,13 @@ While it may be tempting to introduce `*_by_signature` counterparts for every ER
 
 
 ## Specification
-A new method 
-```solidity
-function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+Three new functions are added to the ERC20 ABI:
+```sol
+function permit(address owner, address spender, uint value, uint deadline, uint8 v, bytes32 r, bytes32 s) external
+function nonces(address owner) external view returns (uint)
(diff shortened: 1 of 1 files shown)


EIPS/eip-2612.md
```diff
@@ -43,15 +43,13 @@ While it may be tempting to introduce `*_by_signature` counterparts for every ER
...
-function permit(address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
-```
-and a new storage item
-```solidity
-mapping(address=>uint256) nonces;
+Three new functions are added to the ERC20 ABI:
+```sol
+function permit(address owner, address spender, uint value, uint deadline, uint8 v, bytes32 r, bytes32 s) external
+function nonces(address owner) external view returns (uint)
+function DOMAIN_SEPARATOR() external view returns (bytes32)
```
sohkai, line 50: I don't know if it makes sense to add `DOMAIN_SEPARATOR()` as a standard interface here, especially as it's more relevant to ERC-712.

MrChico: The `DOMAIN_SEPARATOR()` interface function is exposed in order to allow people to verify valid `permit` messages

sohkai: I agree that it should be exposed by the contract in one form or another to help create and verify signatures, but I think this part should be standardized separately from ERC-2612 because it relates more to ERC-712's interface (which currently does not specify a particular on-chain interface).

PR #3076, opened 2020-10-27
EIP-3076: Validator client interchange format (slashing protection)

sachayves:
A JSON interchange format for Serenity (eth2) that contains the necessary slashing protection information required to safely migrate keys between clients.

### DIFF
--- a/EIPS/eip-3076.md
+++ b/EIPS/eip-3076.md
@@ -230,7 +230,7 @@ The specification is designed to be flexible enough to support the full variety 
 
 An earlier design encompassed two separate formats to achieve the same goal: `complete` and `minimal` (for the complete changelog [see here](https://hackmd.io/@sproul/Bk0Y0qdGD)).
 
-The `complete` format was essentially the same as the format outlined in this document (without [conditions](#conditions) 2, 4 and 5). Its goal was to mimick the structure of a slashing protection database that prevents slashing by recording every message signed.
+The `complete` format was essentially the same as the format outlined in this document (without [conditions](#conditions) 2, 4 and 5). Its goal was to mimic the structure of a slashing protection database that prevents slashing by recording every message signed.
 
 The purpose of the `minimal` format was to allow for a simpler slashing protection strategy, whereby the only thing that needed to be recorded was the `slot` number of the most recently signed `block`, along with the `source` and `target` `epoch` numbers of the most recently signed attestation (while refusing to sign messages prior to these).
 
@@ -290,4 +290,3 @@ For implementers who wish to implement their slashing protection database by sto
 
 ## Copyright
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
-
--- a/EIPS/eip-3076.md
+++ b/EIPS/eip-3076.md
@@ -234,7 +234,7 @@ The `complete` format was essentially the same as the format outlined in this do
 
 The purpose of the `minimal` format was to allow for a simpler slashing protection strategy, whereby the only thing that needed to be recorded was the `slot` number of the most recently signed `block`, along with the `source` and `target` `epoch` numbers of the most recently signed attestation (while refusing to sign messages prior to these).
 
-> The advantage of the `complete` format was that it prevented false positives (meaning that it only prevented a validator from signing if it is guarranteed to incur a slashing otherwise). `complete` also supported republishing of identical messages if the `signing_root` is saved. The advantage of the `minimal` format was simplicity of implementation and succinctness of representation.
+> The advantage of the `complete` format was that it prevented false positives (meaning that it only prevented a validator from signing if it is guaranteed to incur a slashing otherwise). `complete` also supported republishing of identical messages if the `signing_root` is saved. The advantage of the `minimal` format was simplicity of implementation and succinctness of representation.
 
 However, in the process of fleshing out this EIP, we realised that we could add a couple of [conditions](#conditions) to `complete` that would allow it to cover the `minimal` case entirely.
 
--- a/EIPS/eip-3076.md
+++ b/EIPS/eip-3076.md
@@ -36,7 +36,7 @@ For example, say client A's time is accidentally set to a day in the future (225
 
 The validator in question now runs the risk of voting for two different blocks in the same epoch (a slashable offence) for the next 225 epochs (since they've already voted on these epochs with client A, and now stand to vote on them again with client B).
 
-The above situation may sound far-fetched, but it's actually very similar to the incident that occurred on the eth2 Medalla testnet on August 14th, where a time skew that affected all Prysm clients spiraled into a series of [cascading failures](https://medium.com/prysmatic-labs/eth2-medalla-testnet-incident-f7fbc3cc934a). It resulted in the testnet experiencing several days without finality -- an outcome that could have resulted in tens of millions of dollars in collective penalties had it occurred on mainnet. 
+The above situation may sound far-fetched, but it's actually very similar to the incident that occurred on the eth2 Medalla testnet on 2020-08-14, where a time skew that affected all Prysm clients spiraled into a series of [cascading failures](https://medium.com/prysmatic-labs/eth2-medalla-testnet-incident-f7fbc3cc934a). It resulted in the testnet experiencing several days without finality -- an outcome that could have resulted in tens of millions of dollars in collective penalties had it occurred on mainnet.
 
 
 ## Specification
--- a/EIPS/eip-3076.md
+++ b/EIPS/eip-3076.md
@@ -1,0 +1,293 @@
+---
+eip: 3076
+title: Validator client interchange format (slashing protection database)
+author: Michael Sproul (@michaelsproul), Sacha Saint-Leger (@sachayves), Danny Ryan (@djrtwo)
+discussions-to: https://discord.com/channels/595666850260713488/598292067260825641/
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-10-27
+---
+
+## Simple Summary
+
+A JSON interchange format for Eth2 that contains the necessary slashing protection information required to safely migrate keys between clients.
+
+## Abstract
+
+ A standard format for transferring a key's signing history allows validators to easily switch between clients without the risk of signing conflicting messages. While a [common keystore format](https://eips.ethereum.org/EIPS/eip-2335) provides part of the solution, it does not contain any information about a key's signing history. For a validator moving their keys from client A to client B, this can lead to scenarios in which client B inadvertently signs a message that conflicts with an earlier message signed with client A.  
+ 
+ We propose a database interchange format that helps solve this problem. The format contains a record of all blocks and attestations signed by a set of validators. 
+
+
+
+## Motivation
+
+Eth2 penalises validators for voting in ways that could result in two different versions of the chain being finalised. These types of penalties are called slashings.
+
+> The two main ways a validator can be slashed are 
+>
+> 1. **Double voting**: voting for two different blocks during the same epoch.
+> 2. **Surround voting**: voting for one version of reality, and later voting for another version in a way that doesn't make it clear that they no longer believe in the first.
+
+For a validator following the protocol correctly, there is, in principle, no risk of being slashed. However, changing clients (from client A to client B, say) can result in a slashing risk if client B is unaware of the blocks and attestations that were signed with client A. 
+
+Specifically, problems can occur if client A and client B do not agree on what the present time is (i.e if their clocks are not synchronised within a certain bound).
+
+For example, say client A's time is accidentally set to a day in the future (225 epochs), and a validator switches from client A to client B without giving B a record of the blocks and attestations signed with A.
+
+The validator in question now runs the risk of voting for two different blocks in the same epoch (a slashable offence) for the next 225 epochs (since they've already voted on these epochs with client A, and now stand to vote on them again with client B).
+
+The above situation may sound far-fetched, but it's actually very similar to the incident that occurred on the eth2 Medalla testnet on August 14th, where a time skew that affected all Prysm clients spiraled into a series of [cascading failures]((https://medium.com/prysmatic-labs/eth2-medalla-testnet-incident-f7fbc3cc934a)). It resulted in the testnet experiencing several days without finality -- an outcome that could have resulted in tens of millions of dollars in collective penalties had it occurred on mainnet. 
+
+
+## Specification
+
+
+### JSON Schema
+
+A valid interchange file is one that adheres to the following JSON schema:
+
+```json
+{
+  "title": "Signing history",
+  "description": "This schema provides a record of the blocks and attestations signed by a set of validators",
+  "type": "object",
+  "properties": {
+    "metadata": {
+      "type": "object",
+      "properties": {
+        "interchange_format_version": {
+          "type": "string",
+          "description": "The version of the interchange format that this document adheres to"
+        },
+        "genesis_validators_root": {
+          "type": "string",
+          "description": "Calculated at Genesis time; serves to uniquely identify the chain"
+        }
+      },
+      "required": [
+        "interchange_format_version",
+        "genesis_validators_root"
+      ]
+    },
+    "data": {
+      "type": "array",
+      "items": [
+        {
+          "type": "object",
+          "properties": {
+            "pubkey": {
+              "type": "string",
+              "description": "The BLS public key of the validator (encoded as a 0x-prefixed hex string)"
+            },
+            "signed_blocks": {
+              "type": "array",
+              "items": [
+                {
+                  "type": "object",
+                  "properties": {
+                    "slot": {
+                      "type": "string",
+                      "description": "The slot number of the block that was signed"
+                    },
+                    "signing_root": {
+                      "type": "string",
+                      "description": "The output of compute_signing_root(block, domain)"
+                    }
+                  },
+                  "required": [
+                    "slot"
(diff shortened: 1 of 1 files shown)


EIPS/eip-3076.md
```diff
@@ -0,0 +1,293 @@
...
+
+Specifically, problems can occur if client A and client B do not agree on what the present time is (i.e if their clocks are not synchronised within a certain bound).
+
+For example, say client A's time is accidentally set to a day in the future (225 epochs), and a validator switches from client A to client B without giving B a record of the blocks and attestations signed with A.
+
+The validator in question now runs the risk of voting for two different blocks in the same epoch (a slashable offence) for the next 225 epochs (since they've already voted on these epochs with client A, and now stand to vote on them again with client B).
+
+The above situation may sound far-fetched, but it's actually very similar to the incident that occurred on the eth2 Medalla testnet on August 14th, where a time skew that affected all Prysm clients spiraled into a series of [cascading failures]((https://medium.com/prysmatic-labs/eth2-medalla-testnet-incident-f7fbc3cc934a)). It resulted in the testnet experiencing several days without finality -- an outcome that could have resulted in tens of millions of dollars in collective penalties had it occurred on mainnet.
```
michaelsproul, line 41: There's an extra pair of parens on this URL causing CI to fail

EIPS/eip-3076.md
```diff
@@ -0,0 +1,293 @@
...
+[cd]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#compute_domain
+
+
+## Rationale
+
+The specification is designed to be flexible enough to support the full variety of slashing protection strategies that clients may implement.
+
+An earlier design encompassed two separate formats to achieve the same goal: `complete` and `minimal` (for the complete changelog [see here](https://hackmd.io/@sproul/Bk0Y0qdGD)).
+
+The `complete` format was essentially the same as the format outlined in this document (without [conditions](#conditions) 2, 4 and 5). Its goal was to mimick the structure of a slashing protection database that prevents slashing by recording every message signed.
```
michaelsproul, line 233:
```suggestion
The `complete` format was essentially the same as the format outlined in this document (without [conditions](#conditions) 2, 4 and 5). Its goal was to mimic the structure of a slashing protection database that prevents slashing by recording every message signed.
```

EIPS/eip-3076.md
```diff
@@ -0,0 +1,293 @@
...
+
+The purpose of the `minimal` format was to allow for a simpler slashing protection strategy, whereby the only thing that needed to be recorded was the `slot` number of the most recently signed `block`, along with the `source` and `target` `epoch` numbers of the most recently signed attestation (while refusing to sign messages prior to these).
+
+> The advantage of the `complete` format was that it prevented false positives (meaning that it only prevented a validator from signing if it is guarranteed to incur a slashing otherwise). `complete` also supported republishing of identical messages if the `signing_root` is saved. The advantage of the `minimal` format was simplicity of implementation and succinctness of representation.
```
michaelsproul, line 237:
```suggestion
> The advantage of the `complete` format was that it prevented false positives (meaning that it only prevented a validator from signing if it is guaranteed to incur a slashing otherwise). `complete` also supported republishing of identical messages if the `signing_root` is saved. The advantage of the `minimal` format was simplicity of implementation and succinctness of representation.
```

EIPS/eip-3076.md
sachayves: talk about cascading failures...

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
+---
+eip: 3076
+title: Validator client interchange format (slashing protection)
+author: Michael Sproul (@michaelsproul), Sacha Saint-Leger (@sachayves), Danny Ryan (@djrtwo)
+discussions-to: https://discord.com/channels/595666850260713488/598292067260825641/
```
MicahZoltu, line 5: This is the first time someone has proposed a Discord channel as a discussion-to link, so I'll need to talk with the other editors to see if that is allowed.  Normally, an https://ethereum-magicians.org/ forum post is the preferred discussion-to location.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+
+A JSON interchange format for Serenity (eth2) that contains the necessary slashing protection information required to safely migrate keys between clients.
+
+## Abstract
+
+ A standard format for transferring a key's signing history allows validators to easily switch between clients without the risk of signing conflicting messages. While a [common keystore format](https://eips.ethereum.org/EIPS/eip-2335) provides part of the solution, it does not contain any information about a key's signing history. For a validator moving their keys from client A to client B, this can lead to scenarios in which client B inadvertently signs a message that conflicts with an earlier message signed with client A.  
+ 
+ We propose a database interchange format that helps solve this problem. The format contains a record of all blocks and attestations signed by a set of validators. 
+
+
```
MicahZoltu, line 22:
```suggestion
```
Stylistic consistency.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+created: 2020-10-27
+---
+
+## Simple Summary
+
+A JSON interchange format for Serenity (eth2) that contains the necessary slashing protection information required to safely migrate keys between clients.
+
+## Abstract
+
+ A standard format for transferring a key's signing history allows validators to easily switch between clients without the risk of signing conflicting messages. While a [common keystore format](https://eips.ethereum.org/EIPS/eip-2335) provides part of the solution, it does not contain any information about a key's signing history. For a validator moving their keys from client A to client B, this can lead to scenarios in which client B inadvertently signs a message that conflicts with an earlier message signed with client A.
```
MicahZoltu, line 18: Keep in mind that linking to a draft EIP creates a weak dependency here that may cause problems if you want to move this to final before 2335 becomes final.  If the only reason for mentioning it is as comparison, I recommend just leaving it out as it isn't really meaningful information for developers 5-10 years in the future who are reading this EIP (target audience).

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Simple Summary
+
+A JSON interchange format for Serenity (eth2) that contains the necessary slashing protection information required to safely migrate keys between clients.
+
+## Abstract
+
+ A standard format for transferring a key's signing history allows validators to easily switch between clients without the risk of signing conflicting messages. While a [common keystore format](https://eips.ethereum.org/EIPS/eip-2335) provides part of the solution, it does not contain any information about a key's signing history. For a validator moving their keys from client A to client B, this can lead to scenarios in which client B inadvertently signs a message that conflicts with an earlier message signed with client A.  
+ 
+ We propose a database interchange format that helps solve this problem. The format contains a record of all blocks and attestations signed by a set of validators.
```
MicahZoltu, line 20: This content currently feels much more like motivation rather than an abstract.  Ideally, the abstract is a human readable and terse version of the specification.  A couple paragraphs that gives the reader the highlights of the specification without getting bogged down in the edge cases and minutia.  Someone may read the abstract to try to "get the gist" of the specification and decide whether this is an EIP that they want to implement.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+
+Specifically, problems can occur if client A and client B do not agree on what the present time is (i.e if their clocks are not synchronised within a certain bound).
+
+For example, say client A's time is accidentally set to a day in the future (225 epochs), and a validator switches from client A to client B without giving B a record of the blocks and attestations signed with A.
+
+The validator in question now runs the risk of voting for two different blocks in the same epoch (a slashable offence) for the next 225 epochs (since they've already voted on these epochs with client A, and now stand to vote on them again with client B).
+
+The above situation may sound far-fetched, but it's actually very similar to the incident that occurred on the eth2 Medalla testnet on August 14th, where a time skew that affected all Prysm clients spiraled into a series of [cascading failures](https://medium.com/prysmatic-labs/eth2-medalla-testnet-incident-f7fbc3cc934a). It resulted in the testnet experiencing several days without finality -- an outcome that could have resulted in tens of millions of dollars in collective penalties had it occurred on mainnet.
```
MicahZoltu, line 41:
```suggestion
The above situation may sound far-fetched, but it's actually very similar to the incident that occurred on the eth2 Medalla testnet on 2020-08-14, where a time skew that affected all Prysm clients spiraled into a series of [cascading failures](https://medium.com/prysmatic-labs/eth2-medalla-testnet-incident-f7fbc3cc934a). It resulted in the testnet experiencing several days without finality -- an outcome that could have resulted in tens of millions of dollars in collective penalties had it occurred on mainnet.
```

Dates should include years, since we don't know what year a reader will be reading this in.  Also, I have a personal preference for ISO-8601 formatted dates which is why this suggestion switches it, but feel free to stick with some other format if you prefer.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+  },
+  "required": [
+    "metadata",
+    "data"
+  ]
+}
+```
+### Example JSON Instance
+
+
```
MicahZoltu, line 149:
```suggestion
```
Style consistency.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+        }
+      ]
+    }
+  },
+  "required": [
+    "metadata",
+    "data"
+  ]
+}
+```
```
MicahZoltu, line 146:
````suggestion
```

````
Style consistency.

File: EIPS/eip-3076.md [hunk 11]
```diff
@@ -0,0 +1,292 @@
...
+
+- The `signed_blocks` `signing_root`s are calculated using [`compute_signing_root(block, domain)`][csr]: where `block` is the block (of type `BeaconBlock` or `BeaconBlockHeader`) that was signed, and `domain` is equal to [`compute_domain(DOMAIN_BEACON_PROPOSER, fork, metadata.genesis_validators_root)`][cd].
+
+- The `signed_attestations` `signing_root`s are calculated using [`compute_signing_root(attestation, domain)`][csr]: where `attestation` is the attestation (of type `AttestationData`) that was signed, and `domain` is equal to [`compute_domain(DOMAIN_BEACON_ATTESTER, fork, metadata.genesis_validators_root)`][cd].
+
+[pps]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#proposer-slashings
+[isad]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#is_slashable_attestation_data
+[csr]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#compute_signing_root
+[cd]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#compute_domain
```
MicahZoltu, line 224: Are these links likely to go stale?  Normally we prefer to have everything inlined and avoid linking externally partially to avoid having to deal with dead links.  Linking to the ETH2 specs repository seems a bit special, but I still have a concern over links breaking and EIP maintainers having to then deal with that down the road.  This is especially true with the links being to `v1.0.0-rc.0` which certainly doesn't sound like a very permanent location....

Ideal situation would be that the contents at the other end of these links be inlined, but I don't know if that is reasonable or not in this context.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+
```
MicahZoltu, line 225:
```suggestion
```
Style consistency.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Rationale
+
+The specification is designed to be flexible enough to support the full variety of slashing protection strategies that clients may implement.
+
+An earlier design encompassed two separate formats to achieve the same goal: `complete` and `minimal` (for the complete changelog [see here](https://hackmd.io/@sproul/Bk0Y0qdGD)).
```
MicahZoltu, line 231: It is best to just explain the rationale for a decision and leave it at that.  Linking out to previous implementations will almost certainly result in dead links and for future readers it can be confusing.  In this case, something like this seems like it would be fine:
```suggestion
We considered using two separate formats (`complete` and `minimal`) rather than a single consolidated format but found that ran into problems.
```

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+The relevant test cases can be found in [this repository](https://github.com/eth2-clients/slashing-protection-interchange-tests).
+
+
+## Implementation
+
+Implementations exist for the following clients:
+- [Teku](https://github.com/ConsenSys/teku/pull/2820)
+- [Lighthouse](https://github.com/sigp/lighthouse/pull/1544)
+- [Prysm](https://github.com/prysmaticlabs/prysm/pull/7518)
+- [Nimbus](https://github.com/status-im/nimbus-eth2/pull/1643)
```
MicahZoltu, line 267: This is currently a bit of a hot topic among the editors at the moment, but this content feels like it would be much more appropriate for the hard fork coordination process than for the EIP.  The EIP is a technical specification, and it doesn't really care if anyone has implemented this or everyone has implemented this.  Also, when someone is reading this in 5-10 years after all of these clients are no longer maintained and the repositories/organizations have been renamed/deleted then these links will no longer be adding value.

Note: As I mentioned, this is currently a bit of a hot topic so not a blocker for draft at the moment.

EIPS/eip-3076.md
```diff
@@ -0,0 +1,292 @@
...
+The `interchange_format_version` is set to 5 because the specification went through several breaking changes during its design, incorporating feedback from implementers. 
+
+## Backwards Compatibility
+
+This specification is not backwards-compatible with versions 3 and 4 of the draft format that have been circulating amongst client teams the last few months. However, we expect clients will update to `v5` for mainnet launch.
+
+
+## Test Cases
+
+The relevant test cases can be found in [this repository](https://github.com/eth2-clients/slashing-protection-interchange-tests).
```
MicahZoltu, line 258: Any chance we could just inline these into the EIP rather than having them be linked externally?  Can we just copy them over to `../assets/eip-3076/*` and call it a day?  Test cases like this feel pretty critical to this EIP, and having them live outside of the EIP change control process feels like it isn't quite right.

MicahZoltu (changes requested):
I think the only real blocker for merging as draft here is the `discussion-to` link, which I need to get feedback from other EIP editors on.

Some of the feedback here is similar to my feedback on previous ETH2 EIPs, which is a bit of a hot discussion topic.  The core issue is that ETH2 is mostly defined elsewhere, but there are a small number of pieces that are being defined in this repository.  This creates complications when one wants to reference one from the other, since our *general* policy is to not have EIPs link externally.  That being said, we can put off finalizing that debate until a later phase of this EIP (e.g., Last Call/Final) and for now it isn't a blocker like it was for some other ETH2 specs in here.

EIPS/eip-3076.md
sachayves: So I realise this was a strange choice, but I couldn't think of a better one. 

From EIP-1:

>  examples for places to discuss your EIP include Ethereum topics on Gitter, an issue in this repo or in a fork of this repo, Ethereum Magicians (this is suitable for EIPs that may be contentious or have a strong governance aspect), and Reddit r/ethereum.

Since this EIP is neither contentious nor has a strong governance aspect, I didn't think it was worth starting an  Ethereum Magicians thread. 

Since (I assume?) Discord channels perform much the same role as Gitter topics did back then, and the discussion has already naturally occurred there (see https://discord.com/channels/595666850260713488/598292067260825641/745665794435973170), I thought it might be appropriate.

Although, I have to say, I'm not completely happy with it, since the url is less expressive than a Gitter topic url, and one has to join the entire discord group to join the channel.

michaelsproul: We've started an Ethereum Magicians thread nonetheless, as it's nice to have a focussed discussion, whereas the previous `eth2` channel covers a lot of other topics and is likely to be noisy:

https://ethereum-magicians.org/t/eip-3076-validator-client-interchange-format-slashing-protection/

EIPS/eip-3076.md
michaelsproul: The links use a GitHub tag rather than a branch name, so will only go stale if the tag gets deleted (which I hope is unlikely).

Our intent here was to link to the latest version of the spec at time of writing, and then update the tag to `v1.0.0` once it is released (before this EIP is finalised). Granted, that still doesn't stop this EIP from bitrotting if breaking changes are introduced to the Eth2 spec post 1.0, but in that case I think the EIP should be amended, or a new EIP drafted in its stead

EIPS/eip-3076.md
michaelsproul: Yes! I'm happy for them to live within the EIP. We'll push them shortly.

EIPS/eip-3076.md
sachayves: Replaced discord with ethereum magicians link. We can mark this as resolved?

GesusBitcoin (commented):
get_committee_assignment(state, epoch, validator_index) where epoch <= next_epoch

GesusBitcoin:
get_committee_assignment(state, epoch, validator_index) where epoch <= next_epoch

GesusBitcoin (commented):
4ae64291cdb43b9a73ab0cde590e92cfcec258e6
[visual impoortant.txt](https://github.com/ethereum/EIPs/files/10505208/visual.impoortant.txt)

PR #3078, opened 2020-10-31
Implementation => Reference Implementation

MicahZoltu:
In reality, the `Implementation` section almost always involves someone linking out to a personal or company repository when they create an EIP from the template.  External repositories cause problems because:
1. Links move/change over time and eventually the link will be dead.
2. The implementation likely isn't CC0 as the EIP is, and naive users may follow the link not realizing it is GPL and they are screwing themselves legally.
3. People are using EIPs for advertising their project, and they look for any opportunity they can to link out to their project in the body of the EIP.
4. People often link to some monolithic repository (e.g., go-ethereum) as the "implementation" which is less than useful.

Also, the description was written back when EIPs were part of the hard fork coordination process, so that has been rewritten.

This change makes it so it is more clear that this section is optional and that external links are not the preferred solution to completing this section.

### DIFF
--- a/eip-template.md
+++ b/eip-template.md
@@ -38,8 +38,8 @@ All EIPs that introduce backwards incompatibilities must include a section descr
 ## Test Cases
 Test cases for an implementation are mandatory for EIPs that are affecting consensus changes. Other EIPs can choose to include links to test cases if applicable.
 
-## Implementation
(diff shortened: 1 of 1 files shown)


lightclient (approved):
I like this. We should probably also update EIP-1.

lightclient:
@MicahZoltu I added a separate PR to make this update in EIP-1 as well. If you find it acceptable, please feel free to merge both PRs.

PR #3080, opened 2020-10-31
Adds words to abstract definition.

MicahZoltu:
I have to coach almost every EIP author on what the abstract is, integrating into the template will hopefully help with that process.

### DIFF
--- a/eip-template.md
+++ b/eip-template.md
@@ -21,7 +21,7 @@ The title should be 44 characters or less.
 If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


eip-template.md
```diff
@@ -21,7 +21,7 @@ The title should be 44 characters or less.
 If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP.
 
 ## Abstract
-A short (~200 word) description of the technical issue being addressed.
+A short (~200 word) description of the technical issue being addressed. This should be a very terse and human readable version of the specification section.  Someone should be able to read the abstract and by the end get the gist of what this specification does.
```
lightclient, line 24:
```suggestion
A short (~200 word) description of the technical issue being addressed. This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does.
```

lightclient (approved):
This also looks good. Please consider accepting the minor edits.

PR #3085, opened 2020-11-02
EIP-3085: wallet_addEthereumChain

rekmarks:
Rendered file: https://github.com/rekmarks/EIPs/blob/2016/EIPS/eip-3085.md

This PR adds EIP-3085, which introduces the `wallet_addEthereumChain` RPC method. It's basically [EIP-2015](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2015.md), without requiring the network to be switched.

I think 2015 is a great idea, but wallets may desire an RPC method that only adds a chain, without requiring that the network is changed. For example, allowing dapps to request network changes is bound to cause UX issues, especially as we see dapps moving to different layer 2 chains.

2015 and 3085 can proceed independently of each other.

Co-author credits go to @pedrouid

### DIFF
--- a/EIPS/eip-2016.md
+++ b/EIPS/eip-2016.md
@@ -2,7 +2,7 @@
 eip: 2016
 title: Wallet Add Ethereum Chain RPC Method (`wallet_addEthereumChain`)
 author: Erik Mars (@rekmarks), Pedro Gomes (@pedrouid)
-discussions-to: https://github.com/ethereum/EIPs/issues/
+discussions-to: https://github.com/ethereum/EIPs/issues/3086
 status: Draft
 type: Standards Track
 category: Interface
--- a/EIPS/eip-2016.md
+++ b/EIPS/eip-2016.md
@@ -1,7 +1,7 @@
 ---
 eip: 2016
 title: Wallet Add Ethereum Chain RPC Method (`wallet_addEthereumChain`)
-author: Erik Mars (@rekmarks), Pedro Gomes (@pedrouid)
+author: Erik Marks (@rekmarks), Pedro Gomes (@pedrouid)
 discussions-to: https://github.com/ethereum/EIPs/issues/3086
 status: Draft
 type: Standards Track
--- a/EIPS/eip-2016.md
+++ b/EIPS/eip-2016.md
@@ -2,7 +2,7 @@
 eip: 2016
 title: Wallet Add Ethereum Chain RPC Method (`wallet_addEthereumChain`)
 author: Erik Mars (@rekmarks), Pedro Gomes (@pedrouid)
-discussions-to: about:blank
+discussions-to: https://github.com/ethereum/EIPs/issues/
 status: Draft
 type: Standards Track
 category: Interface
--- a/EIPS/eip-2016.md
+++ b/EIPS/eip-2016.md
@@ -71,8 +71,8 @@ Only the `chainId` is required per this specification, but a wallet **MAY** requ
 - `rpcUrl`
   - If provided, **MUST** specify the URL of the RPC endpoint used to communicate with the chain.
 
-All URL string fields **SHOULD** include the protocol component of the URL.
-If the protocol of the URL is HTTP, HTTPS **SHOULD** be used, and the `https://` prefix **SHOULD** be included.
+All URL string fields **MUST** include the protocol component of the URL.
+If the protocol of the URL is HTTP, HTTPS **SHOULD** be used.
 
 #### Returns
 
--- a/EIPS/eip-2016.md
+++ b/EIPS/eip-2016.md
@@ -1,0 +1,237 @@
+---
+eip: 2016
+title: Wallet Add Ethereum Chain RPC Method (`wallet_addEthereumChain`)
+author: Erik Mars (@rekmarks), Pedro Gomes (@pedrouid)
+discussions-to: about:blank
+status: Draft
+type: Standards Track
+category: Interface
+created: 2020-11-01
+requires: 155, 695
+---
+
+## Abstract
+
+The `wallet_addEthereumChain` RPC method allows Ethereum applications ("dapps") to suggest chains to be added to the wallet application.
+The caller must specify a chain ID and some chain metadata.
+The wallet application may arbitrarily refuse or accept the request.
+`true` is returned if the chain was added, and an error otherwise.
+
+Important cautions for implementers of this method are included in the [Security Considerations](#security-considerations) section.
+
+## Motivation
+
+Although Mainnet is by far the most important, a dapp may require the user to interact with other Ethereum chains in order to function.
+Such a chain may or may not be supported by the user's wallet application.
+`wallet_addEthereumChain` enables dapps to request any necessary chains to be added to the user's wallet.
+This provides UX benefits for dapps and wallets that want to provide such functionality.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+### `wallet_addEthereumChain`
+
+The method takes a `chainId` and some chain metadata, returning `true` if the chain was added to the wallet, and an error otherwise.
+
+The wallet **MAY** reject the request for any reason. The wallet **MUST** reject the request if the `chainId` is improperly formatted.
+
+> Note that this method makes **no** statement about whether the wallet should change the user's currently selected chain, if the wallet has a concept thereof.
+
+#### Parameters
+
+```typescript
+interface AddEthereumChainParams {
+  chainId: string;
+  blockExplorerUrl?: string;
+  chainName?: string;
+  nativeCurrency?: {
+    name: string;
+    symbol: string;
+    decimals: number;
+  };
+  rpcUrl?: string;
+}
+```
+
+Only the `chainId` is required per this specification, but a wallet **MAY** require any other fields listed, and impose additional requirements on them.
+
+- `chainId`
+  - **MUST** specify the integer ID of the chain as a hexadecimal string, per the [`eth_chainId`](./eip-695.md) Ethereum RPC method.
+- `blockExplorerUrl`
+  - If provided, **MUST** specify the URL of a block explorer web site for the chain.
+- `chainName`
+  - If provided, **MUST** specify a human-readable name for the chain.
+- `nativeCurrency`
+  - If provided, **MUST** describe the native currency of the chain using the `name`, `symbol`, and `decimals` fields.
+- `rpcUrl`
+  - If provided, **MUST** specify the URL of the RPC endpoint used to communicate with the chain.
+
+All URL string fields **SHOULD** include the protocol component of the URL.
+If the protocol of the URL is HTTP, HTTPS **SHOULD** be used, and the `https://` prefix **SHOULD** be included.
+
+#### Returns
+
+The method **MUST** return the boolean value `true` if the request was successful, and an error otherwise.
+
+If a chain with the provided parameters was already added, the request is considered successful, and the method **MUST** return the boolean value `true`.
+
+The wallet **MAY** define what it means for a chain to be "already added", but care should be taken to protect user security.
+The wallet **SHOULD NOT** allow the same `chainId` to be added multiple times.
+See [Security Considerations](#security-considerations) for more information.
+
+### Examples
+
+These examples use JSON-RPC, but the method could be implemented using other RPC protocols.
+
+To add the Goerli test chain:
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "wallet_addEthereumChain",
+  "params": [
+    {
+      "chainId": "0x5",
+      "chainName": "Goerli",
+      "rpcUrl": "https://goerli.infura.io/v3/406405f9c65348f99d0d5c27104b2213",
+      "nativeCurrency": {
+        "name": "Goerli ETH",
+        "symbol": "gorETH"
+      },
+      "blockExplorerUrl": "https://goerli.etherscan.io"
+    }
+  ]
+}
+```
+
+To add POA Network's xDAI chain:
+
+```json
+{
+  "id":1,
+  "jsonrpc": "2.0",
+  "method": "wallet_addEthereumChain",
+  "params": [
+    {
+      "chainId": "0x64",
+      "chainName": "xDAI Chain",
+      "rpcUrl": "https://dai.poa.network",
+      "nativeCurrency": {
+        "name": "xDAI",
+        "symbol": "xDAI",
+        "decimals": 18,
+      }
+    }
+  ]
+}
+```
+
+A success response:
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": true
+}
+```
+
+A failure response:
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "error": {
+    "code": 4001,
+    "message": "The user rejected the request."
+  }
+}
+```
+
+## Rationale
+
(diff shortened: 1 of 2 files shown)


danfinlay:
It might be nice to include a "request to switch network" method as part of the same proposal? Seems like only with both can we eliminate the situation where dapps have to instruct users how to make network-related selections within their wallet.

rekmarks:
@danfinlay, @MicahZoltu has convinced me that granularity is a virtue when it comes to EIPs, and I think the network switching should be in its own one. 2015 can then extend/bundle the two, as necessary. I noticed that the next free EIP number appears to be 2017 😉

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
+---
+eip: 2016
+title: Wallet Add Ethereum Chain RPC Method (`wallet_addEthereumChain`)
+author: Erik Mars (@rekmarks), Pedro Gomes (@pedrouid)
```
danfinlay, line 4:
```suggestion
author: Erik Marks (@rekmarks), Pedro Gomes (@pedrouid)
```

rekmarks: Haha, thanks!

pedrouid:
I've opened a PR to update EIP-2015 to have closer standardization with this EIP
_____________________________________________________________________________________________________

Something that I'm reconsidering whether or not it's a reasonable parameter is `blockExplorerUrl`. The whole point of these EIPs or more specifically ERCs is to expand interoperability by reaching an agreement in interfaces.

Applications can permissionessly interface with each other given that standards are followed and become interoperable.

However this wouldn't be true for `blockExporerUrl` UNLESS we actually had blockchain explorers agree on an API.

What is the value of providing that url if you are not aware of the API for that website? How is the wallet going to simply link transactions, adresses, tokens, contracts, etc??

### Examples

#### Blocks
Blockscout (chainId=100) -> https://blockscout.com/poa/xdai/blocks/<BLOCK_HASH_OR_HEIGHT>
Etherscan (chainId=1) -> https://etherscan.io/block/<BLOCK_HASH_OR_HEIGHT>
Etherchain (chainId=1) -> https://etherchain.org/block/<BLOCK_HASH_OR_HEIGHT>
Ethplorer (chainId=1) -> unsupported

Etherscan and Etherchain are in consensus but Blockscout uses `/blocks` endpoint instead and Ethplorer doesn't have a page for blocks so returns 404 not found.

#### Transactions
Blockscout (chainId=100) -> https://blockscout.com/poa/xdai/tx/<TX_HASH>
Etherscan (chainId=1) -> https://etherscan.io/tx/<TX_HASH>
Etherchain (chainId=1) -> https://etherchain.org/tx/<TX_HASH>
Ethplorer (chainId=1) -> https://ethplorer.io/tx/<TX_HASH>

All block explorers are in consensus for transactions endpoint.

#### Accounts
Blockscout (chainId=100) -> https://blockscout.com/poa/xdai/address/<ACCOUNT_ADDRESS>
Etherscan (chainId=1) -> https://etherscan.io/address/<ACCOUNT_ADDRESS>
Etherchain (chainId=1) -> https://etherchain.org/account/<ACCOUNT_ADDRESS>
Ethplorer (chainId=1) -> https://ethplorer.io/address/<ACCOUNT_ADDRESS>

All block explorers are in consensus for accounts endpoint except Etherchain which uses `/account` endpoint instead

#### ERC-20 Tokens
Blockscout (chainId=100) -> https://blockscout.com/poa/xdai/tokens/<TOKEN_ADDRESS>
Etherscan (chainId=1) -> https://etherscan.io/token/<TOKEN_ADDRESS>
Etherchain (chainId=1) -> https://etherchain.org/token/<TOKEN_ADDRESS>
Ethplorer (chainId=1) -> https://ethplorer.io/address/<TOKEN_ADDRESS>

Etherscan and Etherchain are in consensus but Blockscout uses `/tokens` endpoint instead and Ethplorer doesn't have a page for tokens so redirects to accounts page (`/address`).

**CONCLUSION:** Some block explorers do share the same API but there are still inconsistencies and in order to make the `blockExplorerUrl` suitable to be included as part of this EIP I would suggest writing an EIP for block explorer endpoints that would be required by this EIP.

pedrouid:
Anyway instead of talking... I've created a draft for this issue which I believe will be very valuable for Wallet development and additionally for this EIP

https://github.com/ethereum/EIPs/pull/3091

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+  - If provided, **MUST** specify the URL of a block explorer web site for the chain.
+- `chainName`
+  - If provided, **MUST** specify a human-readable name for the chain.
+- `nativeCurrency`
+  - If provided, **MUST** describe the native currency of the chain using the `name`, `symbol`, and `decimals` fields.
+- `rpcUrl`
+  - If provided, **MUST** specify the URL of the RPC endpoint used to communicate with the chain.
+
+All URL string fields **MUST** include the protocol component of the URL.
+If the protocol of the URL is HTTP, HTTPS **SHOULD** be used.
```
MicahZoltu, line 75:
```suggestion
HTTPS **SHOULD** be used.
```
I'm not sure if this is what you meant or not, but it definitely seems wrong to say that "if HTTP is explicitly supplied, ignore it and use HTTPS instead".  If that is really what you mean, why not instead say that `HTTPS **MUST** be used if available, even if the URL provided is HTTP` or something?

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "wallet_addEthereumChain",
+  "params": [
+    {
+      "chainId": "0x5",
+      "chainName": "Goerli",
+      "rpcUrl": "https://goerli.infura.io/v3/406405f9c65348f99d0d5c27104b2213",
```
MicahZoltu, line 102: This isn't a valid API key is it?  If so, you should probably roll your keys.  😬

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+      "nativeCurrency": {
+        "name": "Goerli ETH",
+        "symbol": "gorETH"
+      },
```
MicahZoltu, line 106: `decimals` isn't optional in the spec, but it is left out here so at the moment this is invalid.

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+
+The wallet **MAY** reject the request for any reason. The wallet **MUST** reject the request if the `chainId` is improperly formatted.
+
+> Note that this method makes **no** statement about whether the wallet should change the user's currently selected chain, if the wallet has a concept thereof.
+
+#### Parameters
+
+```typescript
+interface AddEthereumChainParams {
+  chainId: string;
```
MicahZoltu, line 49: Why is this a string?  `chainId` is *always* a number I believe, and passing it as a string just makes using this API harder.

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+  - If provided, **MUST** describe the native currency of the chain using the `name`, `symbol`, and `decimals` fields.
+- `rpcUrl`
+  - If provided, **MUST** specify the URL of the RPC endpoint used to communicate with the chain.
+
+All URL string fields **MUST** include the protocol component of the URL.
+If the protocol of the URL is HTTP, HTTPS **SHOULD** be used.
+
+#### Returns
+
+The method **MUST** return the boolean value `true` if the request was successful, and an error otherwise.
```
MicahZoltu, line 79: `Promise<true>` seems a bit silly, why not just `Promise<void>`?

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+For [security reasons](#security-considerations), a wallet should always attempt to validate the chain metadata provided by the dapp, and may choose to fetch the metadata elsewhere entirely.
+Either way, there's no way to know what the wallet will _need_ with respect to the chain metadata.
+Therefore, all parameters except `chainId` are listed as optional, even though a wallet may require them in practice.
+
+This specification does not mandate that the wallet "switches" its active chain, if the result is successful.
+
+For related work, see [EIP-2015](./eip-2015.md).
+
+For requests of this feature in the community, see e.g. [this GitHub issue](https://github.com/MetaMask/metamask-extension/issues/3604).
+
```
MicahZoltu, line 177:
```suggestion
```
This is just history, not of value to a reader 5-10 years in the future reading this specification (target audience).  Also, would be more appropriate in Motivation not Rationale (though not really appropriate anywhere IMO).

EIPS/eip-2016.md
MicahZoltu: Perhaps the issue is that JSON-RPC doesn't have a `void` result type?  In that case, I recommend `null` as a sentinel value rather than `true` as returning `true` implies to the reader that `false` is also an option.

EIPS/eip-2016.md
MicahZoltu: Found the answer, and I disagree with it:
> It should be submitted as a `0x`-prefixed hexadecimal string per [EIP-695](./eip-695.md), and parse to an integer number.

695 had chainId as a hex string because at the time it was effectively a uint256, which cannot be reliably decoded across all languages if encoded in json as a number.  However, I think that since then chainId has been narrowed to a uint32 (I forget where) so we no longer need to encode it as a string on the wire, and we should avoid encoding things as strings where we can.

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+`wallet_addEthereumChain` is a powerful method that exposes the end user to serious risks if implemented incorrectly.
+Most of these risks can be avoided by validating the request data in the wallet, and clearly disambiguating different chains in the wallet UI.
+
+### Chain IDs
+
+Since the chain ID used for signing determines which chain a transaction is valid for, handling the chain ID correctly is of utmost importance.
+The wallet should:
+
+- Ensure that a submitted chain ID is valid.
+  - It should be submitted as a `0x`-prefixed hexadecimal string per [EIP-695](./eip-695.md), and parse to an integer number.
```
MicahZoltu, line 189: This should be in the Rationale, not Security Considerations.

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+- Prevent the same chain ID from being added multiple times, for example with different RPC URLs.
+- Only use the submitted chain ID to sign transactions, **never** a chain ID received from an RPC endpoint.
+  - A malicious or faulty endpoint could return arbitrary chain IDs, and potentially cause the user to sign transactions for unintended chains.
+- Verify that the specified chain ID matches the return value of `eth_chainId` from the endpoint.
+  - This should be done before any other interactions with the endpoint.
```
MicahZoltu, line 194: It feels like this should be done before responding to `wallet_addEthereumChain`.  Consider moving this up to the specification as a hard requirement.

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+### Validating Chain Data
+
+A wallet that implements `wallet_addEthereumChain` should expect to encounter requests for chains completely unknown to the wallet maintainers.
+That said, community resources exist that can be leveraged to verify requests for many Ethereum chains.
+The wallet should maintain a list of known chains, and verify requests add chains against that list, perhaps preferring data from the list over data submitted with a `wallet_addEthereumChain` request.
+
+At present, such community resources include:
+
+- [chainId.network](https://chainid.network)
+
```
MicahZoltu, line 229:
```suggestion
```
The chances that this website will stand the test of time are low, so it really shouldn't be included in the EIP.  Content like this should be made available to users elsewhere on the internet, just not in the specification document.  Keep in mind, the EIP is purely a specification, it is not a best practices, tutorial, help site, FAQ, etc.  The EIP should be kept as simple as possible and focus only on the specification itself and not worry itself with assisting users in actually implementing anything (that is better suited for a blog or forum post).

EIPS/eip-2016.md
```diff
@@ -0,0 +1,241 @@
...
+## References
+
+- [chainId.network](https://chainid.network)
+- Related EIPs
+  - [EIP-155: Simple replay attack protection](./eip-155.md)
+  - [EIP-695: Create `eth_chainId` method for JSON-RPC](./eip-695.md)
+  - [EIP-1102: Opt-in Account Exposure](./eip-1102.md)
+  - [EIP-2015: Wallet Update Chain JSON-RPC Method (`wallet_updateChain`)](./eip-2015.md)
+
```
MicahZoltu, line 238:
```suggestion
```
This is not an official EIP section, and we tend to avoid non-standard sections whenever possible.  I believe all of these are already linked inline, so this section doesn't provide any additional value to the EIP and the content probably should be removed.

MicahZoltu (approved):
I left a bunch of comments, but nothing blocking getting this merged.

I am a little concerned about the cherry picked EIP number.  We have had problems with users picking EIP numbers in the past as people do bad things in order to get vanity numbers so my preference is to disallow any number picking.  That being said, I don't have a *specific* argument against this particular selection and in the past the editors have ruled on the side of being more permissive than I about EIP number selection.

If you want this to be merged by me right away, change the EIP number to 3085 (in the frontmatter and in the file name).  If you really are set on 2016, you can wait until I get more feedback from other editors.

pedrouid:
Just got EIP-3091 merged... I will include it as requirement for EIP-2015... You should also include it for EIP-2016 @rekmarks

EIPS/eip-2016.md
rekmarks: If it is, it's not mine 😅 

I believe I copypasted it from @pedrouid's examples in EIP-2015.

EIPS/eip-2016.md
rekmarks: IMO, that ship has sailed. `eth_chainId` has always returned a hex string, and whatever benefits there are to encoding it as a `number` are overshadowed by the drawbacks of different RPC methods specifying different encodings for the same value. I'd much rather we live with what we have.

EIPS/eip-2016.md
rekmarks: I'll just remove it, since the definition of a valid `chainId` is already specified in the "Parameters" section.

rekmarks:
All feedback addressed, @MicahZoltu and @pedrouid.

If I didn't respond with a comment, I always attempted to fix the problem directly in the EIP.

MicahZoltu (approved):

MicahZoltu:
Hmm, something seems to be broken with GitHub status checks.  Travis is green, yet GitHub says it is pending.  Will check again later and kick it again if it doesn't self resolve.

BelfordZ:
Any chance you could add an OpenRPC definition of the method? I need json schemas for this to add it.

MicahZoltu:
@BelfordZ You should ask over in the `discussions-to` link: https://github.com/ethereum/EIPs/issues/3086

EIPS/eip-3085.md
```diff
@@ -0,0 +1,234 @@
...
+  - The confirmation used in [EIP-1102](./eip-1102.md) may serve as a point of reference.
+- Ensure that any chain metadata, such as `nativeCurrency` and `blockExplorerUrl`, are validated and used to maximum effect in the UI.
+- If the wallet UI has a concept of a "currently selected" or "currently active" chain, ensure that the user understands when a chain added using `wallet_addEthereumChain` becomes selected.
+
+### Validating Chain Data
+
+A wallet that implements `wallet_addEthereumChain` should expect to encounter requests for chains completely unknown to the wallet maintainers.
+That said, community resources exist that can be leveraged to verify requests for many Ethereum chains.
+The wallet should maintain a list of known chains, and verify requests add chains against that list.
+Indeed, unless there are good reasons not to, the wallet should _prefer its own chain metadata_ over anything submitted with a `wallet_addEthereumChain` request.
```
danfinlay, line 230: Possible amendment:

```suggestion
Indeed, unless there are good reasons not to, the wallet should _make clear to the user the existence of any previously configured chain metadata_ over anything submitted with a subsequent `wallet_addEthereumChain` request.
```

This
- Avoids being overly zealous about wallet defaults, and leaves a friendly path to users adding new custom providers.
- Extends the concern to include any pre-loaded chain metadata, not just wallet-developer provided metadata.

PR #3088, opened 2020-11-02
Add chain id to 2930

lightclient:
This change will deprecate EIP-155 behaviour for this transaction type. Many proposed EIP-2718 transactions already separate the chain id from the `v` value, so I believe there is demand for this modification.

### DIFF
--- a/EIPS/eip-2930.md
+++ b/EIPS/eip-2930.md
@@ -16,7 +16,7 @@ Adds a transaction type which contains an access list, a list of addresses and s
 
 ## Abstract
 
-We introduce a new [EIP-2718](./eip-2718.md) transaction type, with the format `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`.
+We introduce a new [EIP-2718](./eip-2718.md) transaction type, with the format `rlp([3, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`.
 
 The `access_list` specifies a list of addresses and storage keys; these addresses and storage keys are added into the `accessed_addresses` and `accessed_storage_keys` global sets (introduced in [EIP-2929](./eip-2929.md)). A gas cost is charged, though at a discount relative to the cost of accessing outside the list.
 
@@ -37,7 +37,7 @@ This EIP serves two functions:
 | `ACCESS_LIST_STORAGE_KEY_COST` | 1900 |
 | `ACCESS_LIST_ADDRESS_COST` | 2400 |
 
-As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list]])`.
+As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([3, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`.
 
 For the transaction to be valid, `access_list` must be of type `[[{20 bytes}, [{32 bytes}...]]...]`, where `...` means "zero or more of the thing to the left". For example, the following is a valid access list (all hex strings would in reality be in byte representation):
 
@@ -90,6 +90,8 @@ def process_access_list(access_list) -> Tuple[List[Set[Address], Set[Pair[Addres
 ```
 
 The access list is NOT charged per-byte fees like tx data is; the per-item costs described above are meant to cover the bandwidth costs of the access list data in addition to the costs of accessing those accounts and storage keys when evaluating the transaction.
+
+Additionally, the `chainId` field must be equal to either the chain id of the current network or `0` if it is unprotected.
 
 ## Rationale
 
--- a/EIPS/eip-2930.md
+++ b/EIPS/eip-2930.md
@@ -37,7 +37,7 @@ This EIP serves two functions:
 | `ACCESS_LIST_STORAGE_KEY_COST` | 1900 |
 | `ACCESS_LIST_ADDRESS_COST` | 2400 |
 
-As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`.
+As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`. `chainId` is defined to be an integer of arbitrary size.
 
 For the transaction to be valid, `access_list` must be of type `[[{20 bytes}, [{32 bytes}...]]...]`, where `...` means "zero or more of the thing to the left". For example, the following is a valid access list (all hex strings would in reality be in byte representation):
 
--- a/EIPS/eip-2930.md
+++ b/EIPS/eip-2930.md
@@ -37,7 +37,7 @@ This EIP serves two functions:
 | `ACCESS_LIST_STORAGE_KEY_COST` | 1900 |
 | `ACCESS_LIST_ADDRESS_COST` | 2400 |
 
-As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([3, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`.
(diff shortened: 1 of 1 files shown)


eip-automerger:

EIPS/eip-2930.md
```diff
@@ -37,7 +37,7 @@ This EIP serves two functions:
 | `ACCESS_LIST_STORAGE_KEY_COST` | 1900 |
 | `ACCESS_LIST_ADDRESS_COST` | 2400 |
 
-As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list]])`.
+As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([3, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`.
```
holiman, line 40:
```suggestion
As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([3, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([3, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`.
```
right?

EIPS/eip-2930.md
```diff
@@ -91,6 +91,8 @@ def process_access_list(access_list) -> Tuple[List[Set[Address], Set[Pair[Addres
 
 The access list is NOT charged per-byte fees like tx data is; the per-item costs described above are meant to cover the bandwidth costs of the access list data in addition to the costs of accessing those accounts and storage keys when evaluating the transaction.
 
+Additionally, the `chainId` field must be equal to either the chain id of the current network or `0` if it is unprotected.
```
holiman, line 94: Is the `chainId` rlp format defined somewhere? It can either be an arbitrarily large integer (in which case the empty value is one byte only), or e.g. a `uint64`, in which case it's fixed size (and quite largre). So I suppose an arbitrary-size integer would make the most sense. 
Also, having `0` be unprotected is a pretty important piece of consensus-vital information, so it should not be mentioned as a mere afterthought. Maybe make a bold "Note" or move it up or something, to make it obvious. A consequence would also be (I guess) that `V` is `27` or `28` (or `0` and `1`?)  

Also, under "testcases", we could add some RLP-encoded blobs of txs

- With `unprotected` transaction, rlp + signing hash 
- With goerli-chainid transaction, rlp + signing hash

That would also clarify the `V` domain. 

@vbuterin thoughts?

holiman: I'm not 100% certain we should use `0` to denote unprotected txs. Do we really want unprotected txs to be "a thing"? It existed earlier for historic reasons, but since only new signers will be able to sign these, we have the option now to get rid of it.

lightclient: I think @MicahZoltu had some reasoning behind why we'd still want to have unprotected transactions. However, I lean on the side of removing the functionality completely.

lightclient: I don't believe `chainId` is defined anywhere, but I agree that an arbitrarily large integer makes the most sense.

EIPS/eip-2930.md
```diff
@@ -16,7 +16,7 @@ Adds a transaction type which contains an access list, a list of addresses and s
 
 ## Abstract
 
-We introduce a new [EIP-2718](./eip-2718.md) transaction type, with the format `rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`.
+We introduce a new [EIP-2718](./eip-2718.md) transaction type, with the format `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`.
```
lightclient, line 19: Note that I've updated the "type" from 3 to 1 since legacy transactions could later occupy the 0 spot and this is the first EIP-2718 transaction type that will be implemented on mainnet.

EIPS/eip-2930.md
MicahZoltu: Unprotected transactions allow for trustless contract deployments across multiple chains which I believe to be of significant value as it makes testing and integration *way* easier when a particular contract (e.g., Uniswap) lives at the same address on every testnet and mainnet.

MicahZoltu:
I started drafting up a larger change to this transaction type (re: how it is serialized) but never got resolution on a couple things in discussion with @vbuterin on Discord, so it never made it into an actual PR.

https://discord.com/channels/595666850260713488/745077610685661265/754676190995939378

**TL;DR**:
* I agree that chainID should be separate.
* I think `v` should be changed to `yParity`: get rid of the `+27` as it is just a vestigial tail from Bitcoin roots.
* I think we should isolate the signed data from the signature.
* We should consider introducing SSZ with this transaction type.  We are going to have to introduce it eventually (ETH2 stuff), and if there isn't a compelling argument against "why not now" then now seems better than later.

holiman:
I agree about this:

> 
>     * I agree that chainID should be separate.
> 
>     * I think `v` should be changed to `yParity`: get rid of the `+27` as it is just a vestigial tail from Bitcoin roots.
> 

I don't quite know what you propose by this, and I'm ok with the way it's specified in the EIP now:
 
>     * I think we should isolate the signed data from the signature.
> 

I don't quite agree about this, I think it makes the introduction of the new transaction type a lot more complex, requiring quite substantial changes to introduce what is otherwise not that big of a change:
 
>     * We should consider introducing SSZ with this transaction type.  We are going to have to introduce it eventually (ETH2 stuff), and if there isn't a compelling argument against "why not now" then now seems better than later.

EIPS/eip-2930.md
vbuterin: > Unprotected transactions allow for trustless contract deployments across multiple chains

But isn't the "correct" way to do this to just use CREATE2? Just use a single old-style transaction to make a standardized contract factory and then use that to make any desired fixed-address contracts.

vbuterin: I'm happy with separating out chainid from the v value; it definitely is cleaner to have each element represent a separate thing than to have an awkward de-facto mash of integers and bitfields.

lightclient: If we're in agreement, I'll remove the caveat that 2930 txs are unprotected when `chainId == 0`.

MicahZoltu: As long as there is a way to deploy a factory without a chain ID then create 2 will work. This does mean that we won't be able to use access lists for everything though, since at least that factory deployment transaction won't be able to use access lists. In that case though, the access list would only have contained the account info for the deployer.

EIPS/eip-2930.md
```diff
@@ -91,6 +91,8 @@ def process_access_list(access_list) -> Tuple[List[Set[Address], Set[Pair[Addres
 
 The access list is NOT charged per-byte fees like tx data is; the per-item costs described above are meant to cover the bandwidth costs of the access list data in addition to the costs of accessing those accounts and storage keys when evaluating the transaction.
 
+Additionally, the transaction is valid only if `chainId` is equal to the chain id of the current network.
```
holiman, line 94:
```suggestion
Clarification: the transaction is valid only if `chainId` is equal to the chain id of the current network. 
```
That sentence is now more of a clarification, it doesn't really _add_ anything anymore

MicahZoltu:
> I think we should isolate the signed data from the signature.
Currently proposed layout:
```
rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])
```
Pseudocode for signing:
```
signed_data = [1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]]
encoded_signed_data = rlp_encode(signed_data)
signature = sign(encoded_signed_data)
transaction = [1, [...signed_data[1], ...signature]
encoded_transaction = rlp_encode(transaction)
```
Pseudocode for validating:
```
decoded_transaction = rlp_decode(transaction)
signature = decoded_transaction[1].slice(8)
signed_data = [1, decoded_transaction[1].slice(0, 8)]
validate_signature(signed_data, signature)
```

If the signature is separate from the signed data such as with a format like:
```
[1, [yParity, r, s], [1, nonce, gasPrice, gasLimit, to, value, data, accessList]]
```
then the pseudocode for singing and validation instead looks like this:
```
signed_data = [1, nonce, gasPrice, gasLimit, to, value, data, accessList]
encoded_signed_data = rlp_encode(signed_data)
signature = sign(encoded_signed_data)
transaction = [1, signature, transaction]
encoded_transaction = rlp_encode(transaction)
```
```
decoded_transaction = rlp_decode(transaction)
signature = decoded_transaction[1]
signed_data = decoded_transaction[2]
validate_signature(signed_data, signature)
```

At the cost of one extra byte, we reduce the complexity of the signing/validating logic which is code that exists in many libraries spread across many languages.  It is important to note that in the pseudocode above I use the spread operator which is not available in all languages and without access to the spread operator things become notably more complicated.

----

If we were to use SSZ, we could avoid the single duplicated byte (transaction type) while still achieving the simplicity provided above since SSZ encodes `byte || ssz(more_bytes)` the same as `ssz(byte || more_bytes)`.

With SSZ and a duplicated type byte we actually can fully validate the signature without decoding the transaction at all, since you can simply pop off the first 66 bytes (transaction type and signature) and validate the remaining bytes without deserializing them at all.  This means transaction validation can occur without even needing to know how to do SSZ decoding, which is a nice to have.

I feel more strongly about having the signature be separate from the signed data than I do about using SSZ encoding.  However, without SSZ encoding we will need to duplicate the transaction type byte in the transaction.

holiman (approved):

holiman:
Sorry, @MicahZoltu I approved it and it got merged before I saw your latest comment. However, it might be better to have that discussion on some other PR, since this PR is mainly about whether to break out `chainId` or not. So maybe you'd like to do another PR proposing your changes? 

Now, I don't really agree with you. 

> At the cost of one redundant byte, we reduce the complexity of the signing/validating logic...

What I see as the big flaw in that scheme, is that you introduce an ambiguity. The same field is present in two different places, and only one of them, the second, are 'signed'. So let's say a client get's a `[1337, [yParity, r, s], [1, nonce, gasPrice, gasLimit, to, value, data, accessList]]` transaction from the network. The client can either discard it, or just 'correct' it into `[1, [yParity, r, s], [1, nonce, gasPrice, gasLimit, to, value, data, accessList]]`. However, if a client chooses to discard it, maybe it will also 'remember' the hash as 'bad'. But should the has be derived from  `[1, [0,0,0], [1, nonce, gasPrice, gasLimit, to, value, data, accessList]]`  (the 'canon' repr) or `[1337, [0,0,0], [1, nonce, gasPrice, gasLimit, to, value, data, accessList]]` (the possibly modified version obtained from the network)? 

In general, I think that ambiguity in a protocol often leads to subtle security flaws at some later part in the process.

MicahZoltu:
No worries about merging this, you are right that it isn't the right place to discuss other changes.  I'll continue the discussion at the discussion-to link.

PR #1889, opened 2019-03-31
Remove Mist from EIP-55 as it is deprecated

ligi:
context: 

Selection_042

https://medium.com/@avsa/sunsetting-mist-da21c8e943d2

PR #3092, opened 2020-11-04
Cleans up EIP-55

MicahZoltu:
This EIP was written in a time when we were using EIPs for hardfork coordination and we weren't worried too much about link rot.  However, times have changed and maintaining EIPs like this is becoming more and more of a burden.  In this case, the specification is what is important, not who implements it.  Also, editors do not want to maintain external links to things that don't add *significant* value.

### DIFF
--- a/EIPS/eip-55.md
+++ b/EIPS/eip-55.md
@@ -2,6 +2,7 @@
 eip: 55
 title: Mixed-case checksum address encoding
 author: Vitalik Buterin <vitalik.buterin@ethereum.org>, Alex Van de Sande <avsa@ethereum.org>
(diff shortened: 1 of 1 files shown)


lightclient (approved):
Any chance we can get an approval from @vbuterin or @alexvandesande?

lightclient:
Just read the discussion in #1889 - as this is simply modernizing the EIP, I will go ahead and merge this.

ligi:
Hm - I think the list of implementations at least has been helpful - not really sure why this needed to go

MicahZoltu:
@ligi Not all things that are useful are appropriate for a standard.  😀  Lists of things in particular are pretty poorly suited for a standards repository.

PR #3095, opened 2020-11-05
Update eip-2535.md

mudgen:


MicahZoltu:
Can you change the Author line for this EIP to either:
```
author: Nick Mudge (@mudgen)
```
or
```
author: Nick Mudge (@mudgen), Nick Mudge <nick@perfectabstractions.com>
```
We have had problems in the past with the bot struggling with email addresses so we have a new policy that there should always be at least one author with a GitHub handle so the bot can auto-merge reliably.

mudgen:
Yes, I can do this.  Thanks @MicahZoltu

PR #3096, opened 2020-11-06
Update EIP statuses

lightclient:
I'm updating EIP statuses according to #2996. The mappings are as follows:

```
Draft => Draft
Last Call => Last Call
Accepted => Final
Final => Final
Superseded => Final
Abandoned => Withdrawn
Rejected => Withdrawn
Active => Living
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
     - language: rust
       cache: cargo
       before_script:
-        - cargo install eipv --version=0.0.6
+        - cargo install eipv --version=0.1.0
       env: TASK='eip-validator'
     - python: 3.3
       env: TASK='codespell'
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -1,7 +1,7 @@
 ---
 eip: 1
 title: EIP Purpose and Guidelines
-status: Active
+status: Living
 type: Meta
 author: Martin Becze <mb@ethereum.org>, Hudson Jameson <hudson@ethereum.org>, et al.
 created: 2015-10-27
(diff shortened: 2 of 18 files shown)


lightclient:
cc: @axic @MicahZoltu @Souptacular

MicahZoltu:
Going to merge this since we have already discussed it at length and I *suspect* (based on history) no one else will comment.  It touches fairly few files, so if we decide we want to rollback later we can pretty easily.

MicahZoltu:
Oops, just after clicking merge I realized that `Accepted => Review` is what it should have been.  I'll get that fixed.

edsonayllon:
Are we not using Stagnant anymore? Because `Abandoned` would be `Stagnant`.

MicahZoltu:
I believe we only had Abandoned previously, and I *believe* that previous to now we have never marked an EIP as Abandoned except when the author requested/supported it.  Would probably be valuable to go through and check the PRs that changed each of these to Abandoned and verify that is the case though.

MicahZoltu:
I just verified all of them were moved to Abandoned by the author, so I believe Withdrawn is correct here.

lightclient:
I guess my understanding if ProgPow's status was that it was fully specified and ready to be included in a hard fork if it were decided by the core developers necessary. In which case, I believe the appropriate status is `Final`.

MicahZoltu:
I'm *really* hesitant to automatically move anyone's EIP into Final without their consent, since they lose the ability to make normative changes at that point.

edsonayllon:
Yeah, I think `Review` is more appropriate for now, unless Greg wants it in `Final`.

PR #3097, opened 2020-11-06
Correcting a mistake in a mass-update.

MicahZoltu:
EIPs have a new set of statuses, Accepted is no longer a status but we accidentally mass updated Accepted => Final, when it should have been Accepted => Review.


MicahZoltu:
Going to merge this as a hotfix, since I merged the incorrect one moments ago.  If other editors disagree with this hotfix I'm fine to roll it back, though I don't believe this fix is contentious.

PR #3101, opened 2020-11-06
EIP-2930: Updates transaction format to match EIP-2718

MicahZoltu:
2718 has the first byte of every transaction be the transaction type, and the remaining bytes are interpreted however this EIP wants. This is the minimal change required to get it in line with 2718 so I think we should merge this ASAP and then continue discussion on whether we think this is the optimal format or not.

### DIFF
--- a/EIPS/eip-2930.md
+++ b/EIPS/eip-2930.md
@@ -37,7 +37,7 @@ This EIP serves two functions:
 | `ACCESS_LIST_STORAGE_KEY_COST` | 1900 |
 | `ACCESS_LIST_ADDRESS_COST` | 2400 |
 
-As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`. `chainId` is defined to be an integer of arbitrary size. Since `chainId` no longer resides in the signature, `v` must be either 0 or 1 depending on recovery id.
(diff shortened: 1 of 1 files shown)


eip-automerger:

EIPS/eip-2930.md
```diff
@@ -16,7 +16,7 @@ Adds a transaction type which contains an access list, a list of addresses and s
 
 ## Abstract
 
-We introduce a new [EIP-2718](./eip-2718.md) transaction type, with the format `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`.
+We introduce a new [EIP-2718](./eip-2718.md) transaction type, with the format `3 || rlp([nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS])`.
```
lightclient, line 19:
```suggestion
We introduce a new [EIP-2718](./eip-2718.md) transaction type, with the format `1 || rlp([nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS])`.
```

lightclient (changes requested):

EIPS/eip-2930.md
holiman, line 19: Any reason to set this back from `1` to `3`? Why `3`?

MicahZoltu: Sorry, was a copy/paste error I think (not intentional).  Though, 2972 tentatively uses ID 0 and 1, so maybe we actually want this to be 2?  I'll switch it to 1 for this PR, but I will create a follow-up to change it to 2.

MicahZoltu: Fixed.  Once this is approved/merged I'll PR a change to `2` so it doesn't conflict with 2972, and I do feel like types 0 & 1 for 2972 is appropriate, since it isn't actually a new transaction type, just wrapper.

lightclient: Agreed.

holiman:
This PR still leaves other portions, e.g. 

> As of `FORK_BLOCK_NUMBER`, we introduce a new transaction type, with the format `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])`. The signing hash would be the hash of the RLP of the entire data structure but popping the last three items representing the signature from the inner list: `rlp([1, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]])`. `chainId` is defined to be an integer of arbitrary size. Since `chainId` no longer resides in the signature, `v` must be either 0 or 1 depending on recovery id.

That needs to be updated aswell, right?

MicahZoltu:
You are totally right, it was totally just me being bad at authoring/editing EIPs that missed that.  😬

MicahZoltu:
@holiman Any chance we can get this approved/merged?  I think it would help in discussions and future PRs to work from a "correct" state.

holiman (approved):
LGTM

PR #3102, opened 2020-11-06
Binary Trie

gballet:


gballet:
Hmmm, looks like there were some connectivity issues during the travis build. I don't have the rights to restart the build.

EIPS/eip-3102.md
```diff
@@ -0,0 +1,190 @@
...
+title: Binary trie structure
+author: Guillaume Ballet (@gballet), Vitalik Buterin (@vbuterin)
+status: Draft
+discussions-to: https://ethresear.ch/t/binary-trie-format/7621
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-09-01
+requires: None
+replaces: None
```
MicahZoltu, line 12:
```suggestion
```
Just leave them out if there are no dependencies.  Otherwise the bot will complain (as it is).

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+If a prefix is present, the length of the path from the root to the prefixed node is further concatenated with the output of the prefix-less rule, and hashed again:
+
+```
+internal_hash_with_prefix = hash(u256(path_length_u256 - 1) || internal_hash)
+leaf_hash_with_prefix = hash(u256(path_length_u256 - 1) || leaf_hash)
+```
+
+## Rationale
+
+### BLAKE2-256
```
axic, line 102: What is BLAKE2-256? Is it BLAKE-256, BLAKE2s, or even BLAKE2b?

gballet: right, it's blake2b. I'll fix it.

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+discussions-to: https://ethresear.ch/t/binary-trie-format/7621
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-09-01
+---
+
+## Simple Summary
+
+A proposal for changing the format of the data-storage structure from hexary to binary, merging the account and storage tries, and using the faster BLAKE2. The goal is to make stateless proofs more wieldy.
```
MicahZoltu, line 15: Simple Summary should be akin to an email subject line, forum topic title, GitHub PR title, etc.  Also, goals should be part of the motivation.
```suggestion
Change storage structure from hexary to binary, merge the account and storage tries, and use BLAKE2.
```

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Abstract
+
+This proposal presents a binary structure and merkelization rule for the account and storage tries, which are merged into a single “state” trie. RLP and most of the MPT’s optimizations are dropped to simplify the design. Keccak256 is replaced with BLAKE2-256.
```
MicahZoltu, line 19: The Abstract should be a very terse human readable version of the specification.  Ideally, someone should be able to read the abstract and get a rough idea of what it would take to implement this specification.  Minutia and edge cases should be left out, but the core should be included.  In this case, this specification is pretty big so it will be hard to cut it down but the ideal abstract is 1 or two paragraphs in mostly plain English.

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+The MPT design is also rife with uncanny optimizations for size, that have a limited effect - at the cost of prohibitive complexity. For example, nesting for children whose RLP is less than 32 bytes saves an estimated 1MB of disk space. A paltry compared to the 300GB required by a fast sync at the time of this writing. These optimizations are a significant source of errors, and therefore a consensus-breaking risk.
+The reliance on RLP has also been criticized for its complexity, while the overhead of a general-purpose encoding scheme isn’t warranted for the rigid structure of a Merkle trie.
+
+The desire to switch the storage model from an hexary trie to a binary trie provides an opportunity to adopt a simpler trie architecture that pushes optimizations from the protocol level down to that of the client implementors.
+
+## Specification
+
+### Conventions
+
+As defined in the yellow paper, $\mathbb{B}$ is the set of all possible byte sequences, and $\mathbb{B}_{32}$ is the set of all byte sequences of length 32. Accordingly, $\mathbb{B}_{l}$ is the set of all byte sequences of length $l$.
```
MicahZoltu, line 39: I don't believe any of the EIP rendering systems support latex.  I recommend using programmer style rather than mathematician style as that will make this specification readable in just about any rendering system (including plain text).

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+leaf_hash_with_prefix = hash(u256(path_length_u256 - 1) || leaf_hash)
+```
+
+## Rationale
+
+### BLAKE2-256
+
+BLAKE2 offers better performance, which is key to compensate for the loss of performance associated to a ~4x increase in the number of nodes.
+
+BLAKE3 offers even better performance. No official golang release exists at the time of the writing of this document. This presents a security risk, and therefore BLAKE2 is considered instead.
```
MicahZoltu, line 106: Why is only golang mentioned here?  If the argument is that language X doesn't have a library for Y but does have a library for Z, then X should actually be a set of all of the major languages used in the ecosystem including Python, C#, Java, JavaScript, Rust, Go.  Is there a BLAKE2 library available for all of them?  Go should not be given special treatment as we are trying to get away from "Geth dominance" and encourage a wider adoption of a variety of different clients.

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+An alternative proposal has been made, which provides optimal witnesses. The trade-off is that extension nodes must be removed.
+
+```
+node_hash = hash(left_child_hash || right_child_hash)
+leaf_hash = hash(0 || leaf_value)
+```
+
+The removal of extension nodes induces 40x higher hashing costs (on the order of 25ms for a trie with only 1k leaves) and as a result they have been kept.
+
+An attempt to keep extension nodes for witness and not the merkelization rule can be found [here](https://notes.ethereum.org/m5VMkX8FRvi0Q_OOR7TF4A).
```
MicahZoltu, line 123: When possible, try not to link externally.  Can the contents of this document be included inline or as an asset in `../assets/eip-3102/`?

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+
+This section is still incomplete and will be fleshed out as clients start implementing the support for binary tries.
+
+An initial implementation has been made for Geth. Unit tests can be found [here](https://github.com/gballet/go-ethereum/blob/binary-trie-m5-full-path/trie/binary_test.go).
+
+## Implementation
+
+ * As of [commit 0db87e187dc0bfb96046a47e3d6768c93a2e3331](https://github.com/gballet/multiproof-rs/commit/6d22b1aef9548581826b3c04b3e00d6cc709388c), [multiproof-rs](https://github.com/gballet/multiproof-rs) implements this merkelization rule in the `hash_m5()` function, found in file `src/binary_tree.rs`.
+ * An implementation of this structure for go-ethereum is available [in this branch](https://github.com/gballet/go-ethereum/tree/binary-trie-m5).
+
```
MicahZoltu, line 178: The implementation section is optional, and I general prefer that if it is included it has an inline reference implementation or something in `../assets/eip-3102/` rather than linking externally.  External links like this, especially to private repositories, tend to not stand the test of time very well.  Keep in mind that EIPs are meant to be pure technical specifications, and things like examples, tutorials, libraries, etc. should be handled elsewhere.

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+## Security Considerations
+
+Issues could arise when performing the transition. In particular, a heavy conversion process could incentivize clients to wait the transition out. This could lead to a lowered network security at the time of the transition. A transition process has been proposed with [EIP-2584](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2584.md).
```
MicahZoltu, line 181:
```suggestion
Issues could arise when performing the transition. In particular, a heavy conversion process could incentivize clients to wait the transition out. This could lead to a lowered network security at the time of the transition. A transition process has been proposed with [EIP-2584](./eip-2584.md).
```

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Acknowledgment
+
+This EIP has benefited from the valuable input of @poemm, @BoltonBailey, @s1na, @ledgerwatch and @holiman.
+
```
MicahZoltu, line 186:
```suggestion
```
EIPs are public domain, which means attribution is not required and we want to avoid EIPs turning into a place where people give out props/kudos/etc.  Social media, blogs, discussion forums, personal/company websites, etc. are a much better place for such things than a technical specification.

EIPS/eip-3102.md
```diff
@@ -0,0 +1,189 @@
...
+## Backwards Compatibility
+
+A hard fork is required in order for blocks to have a trie root using a different structure.
+
+## Test Cases
+
+This section is still incomplete and will be fleshed out as clients start implementing the support for binary tries.
+
+An initial implementation has been made for Geth. Unit tests can be found [here](https://github.com/gballet/go-ethereum/blob/binary-trie-m5-full-path/trie/binary_test.go).
```
MicahZoltu, line 172:
```suggestion
TBD
```
The EIP should be written toward its final state and not contain transient information.  If Test Cases aren't available yet that is totally fine, but it is better to just put TBD in rather than temporary filler material.  As mentioned elsewhere, external links should be avoided and an ideal test cases section would just have a list of test inputs and expected outputs that can easily be ported to any language/testing system.

MicahZoltu (approved):
Structure and layout generally looks good.  There are a number of things I would like to see seen before this moves on to `Review` but nothing that blocks from merging as a draft.

PR #3111, opened 2020-11-10
Decrease the target Collateralisation Ratio for SNX stakers to 500%

nottrunner:
**Simple Summary**
Decrease the target Collateralisation Ratio for SNX stakers to 500%

**Abstract**
The system is much more stable with the first successful set of liquidations executing without a hitch. Its a good time to further lower c-rations to increase the opportunity to mint more debt and increase returns for synth holders and traders

**Motivation**

1. Liquidation process in place and working 
2. Network and Active c-ratio are a very health 783% and 635% respectfully, even with recent drop in SNX price
3. Increased liquidity will support increased use of the exchange(s) and transaction fees
4. Capital Efficiency 
5. Increased demand for synths will drive more integrations with other DeFi platforms
6. It will allow us to further collect data on stability of the peg, would be market absorb the increased supply?
7. We now have liquidations

Copyright
Copyright and related rights waived via CC0.

Author : akinsawyerr


MicahZoltu:
I'm not sure what happened here, but I'm pretty sure this PR isn't the solution to any problem.  Going to close, @mention me if you believe this should be left open.

nottrunner:
Sorry was submitted in error.
thanks

PR #3115, opened 2020-11-13
EIP-2972: Deprecateds legacy transactions in blocks.

MicahZoltu:
Per discussion in ACD call.


MicahZoltu:
Manually merged since bot seems to be down, and I am the author.

PR #3124, opened 2020-11-21
EIP-1559 constant tip strategy

lightclient:
It's clear that managing a mempool of EIP-1559 transactions is non-trivial. As the spec stands today, the value that a transaction may provide the miner is actually a function of the current `base_fee`. This is problematic as it implies that the mempool must constantly reorder its transactions lists as the `base_fee` fluctuates.

The alternative that I propose is to change from using a *dynamic tip* to a *constant tip*. The *constant tip* has several properties that are advantageous, most notably that a mempool can easily maintain an ordered list of the most lucrative transactions, since their payouts do not depend on the `base_fee`. 

I've done some preliminary comparisons between the *dynamic tip* mechanism vs. the *constant tip* mechanism, and shared some thoughts on how a mempool might handle the two: https://hackmd.io/@matt/HJ4YbvNcw

The main caveat is that *dynamic tip* transactions offered a very flexible mechanism for legacy transactions to be ported over. I believe that *constant tip* transactions can be similarly flexible, but require a bit more complexity. For example, I've packed the upper byte in `gas_price` to be used as the `tip` in `gwei`. This leaves 56 bits to specify a maximum `fee_cap` of `72_057_594` gwei. This should be acceptable for the foreseeable future.

### DIFF
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -61,7 +61,7 @@ class Transaction1559:
 	destination: int = 0
 	amount: int = 0
 	payload: bytes = bytes()
-	max_miner_bribe_per_gas: int = 0
+	miner_bribe_per_gas: int = 0
 	fee_cap_per_gas: int = 0
 	v: int = 0
 	r: int = 0
@@ -138,10 +138,8 @@ class World(ABC):
 
 			# ensure that the user was willing to at least pay the base fee
 			assert transaction.fee_cap_per_gas >= block.base_fee
-			# bribe is capped such that base fee is filled first
-			bribe_per_gas = min(transaction.max_miner_bribe_per_gas, transaction.fee_cap_per_gas - block.base_fee)
 			# signer pays both the bribe and the base fee
-			effective_gas_price = bribe_per_gas + block.base_fee
+			effective_gas_price = transaction.miner_bribe_per_gas + block.base_fee
 			signer.balance -= transaction.gas_limit * effective_gas_price
 			assert signer.balance >= 0, 'invalid transaction: signer does not have enough ETH to cover gas'
 			gas_used = self.execute_transaction(transaction, effective_gas_price)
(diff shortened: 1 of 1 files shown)


eip-automerger:

MicahZoltu:
I like the idea of having constant miner bribe, but I dislike the idea that all existing signing tools in the ecosystem will break until updated to handle EIP-1559.  Are there any alternative solutions to the gas price => tip conversion problem?

lightclient:
One alternative is to default to a constant tip (like 1 gwei) in the event that the upper byte isn't packed. That would allow wallets which haven't updated to still have a bit more priority.

I don't think the bit packing is *too* bad, because nearly every wallet supports choosing the gas price. It would be easy to create a site which takes in a `tip` and `fee_cap`, and then outputs the corresponding `gas_price`.

MicahZoltu:
> nearly every wallet supports choosing the gas price

Sadly, this isn't true, with Coinbase wallet being that most popular example of a wallet that doesn't support user configurable gas price.  Also, I don't believe most users would be able to do this on their own.  Trying to teach a user to set a gas price to a number like "`452312848583266388373324160190187140051835877600158453279131187530910662656` plus the gas price you want to use" will result in a lot of users messing things up.

> One alternative is to default to a constant tip (like 1 gwei) in the event that the upper byte isn't packed.

I'm generally concerned with hard coding anything like this.  That being said, I do feel like 10 nanoeth is sufficiently above the marginal cost of pretty much any miner if we believe that the mathing Vitalik did a while back is correct.  I think 1 nanoeth is too low, as it is near the estimated marginal cost, which means it is very possible it is actually less than marginal cost.

lightclient:
The dynamic tip approach is definitely superior in terms of supporting legacy txs. However, constant tip transactions are much easier to reason about for miners and I believe we should push for wallets to support it. I expect most will. Continuing to support legacy transactions should be an act of good will during the transition to 1559. I don't think it should dictate the mechanism, especially if that mechanism is suboptimal long term.

lightclient:
We ended up deciding that i) supporting legacy transactions was a very high priority and ii) sorting dynamic tip transactions aren't as bad as originally thought. Here is a write up from @adietrichs on this topic: https://hackmd.io/UhdsNdF7RTOyELAlB6-MCA?view

PR #3128, opened 2020-11-22
EIP-1559: Add check that gas used is less than gas limit

lrettig:
In the specification logic as it stands, there's no check that the gas consumed by executing a transaction (`gas_used`) is below the max gas specified for the transaction (`transaction.gas_limit`). This means the `gas_refund` could be negative. This adds a simple check that the gas consumed by executing a transaction is less than the limit. (If this is meant to be implicit, i.e., `self.execute_transaction()` respects `transaction.gas_limit`, a comment should be added to this effect.)

### DIFF
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -145,6 +145,7 @@ class World(ABC):
 			signer.balance -= transaction.gas_limit * effective_gas_price
 			assert signer.balance >= 0, 'invalid transaction: signer does not have enough ETH to cover gas'
 			gas_used = self.execute_transaction(transaction, effective_gas_price)
+			assert gas_used <= transaction.gas_limit, 'transaction used more gas than gas_limit'
 			gas_refund = transaction.gas_limit - gas_used
 			cumulative_transaction_gas_used += gas_used
 			# signer gets refunded for unused gas


eip-automerger:

lightclient:
Adhering to the `gas_limit` is implicit. `execute_transaction` must execute following consensus rules. I don't think we should explicitly state that the function adheres to a specific consensus rule, otherwise it might raise a question as to the validity of the functions output in the first place.

MicahZoltu:
@lrettig I'm not directly against adding the check you have suggested, but I'm curious where you would draw the line?  If we add *all* of the checks, we would have to write a full consensus compliant client, which certainly seems out of scope for the 1559 specification.  If we write *none* of the checks, then we have no specification.

The goal with the current set of checks is to focus on checks that are impacted/touched by EIP1559.  Since this EIP doesn't change the behavior around transaction gas limits, it *feels* like the check you have proposed is across that line.  My guess is that in your mental model there is a subtly different line that you are imagining, and I would like to understand that better.

lrettig:
@MicahZoltu This is a very helpful comment, thank you. Actually, I think our mental models here are more or less the same, but you've helped me clarify why I find this addition so helpful. While this EIP doesn't impact per-transaction gas limit, it _does_ of course directly impact per-transaction gas _spending,_ in particular through `effective_gas_price`. In my mental model, as I read this code, I find myself wondering, "Given the new effective gas price math, how much is this transaction spending? How do I know the amount it spends now is still under the gas limit?"

> I'm curious where you would draw the line?

I think the answer is, gas computation.

MicahZoltu:
Perhaps the confusion is that 
```
gas_used = self.execute_transaction(transaction, effective_gas_price)
```
makes it look like `gas_used` is a function of `transaction` and `effective_gas_price`, but in reality it isn't, it is only a function of the transaction.  The reason we pass `effective_gas_price` into `execute_transaction` is only because there is an opcode in the EVM to get the `GASPRICE`, and that opcode needs to use the `effective_gas_price`.  However, unless the contract branches on the value of `GASPRICE`, the `gas_used` will be the same no matter what `effective_gas_price` is.

If the code was:
```
gas_used = self.execute_transaction(transaction)
```
would you still find value in the additional assert after that function, even though in this case it is more clear that `effective_gas_price` isn't a dependent variable?

lrettig:
> ... makes it look like gas_used is a function of transaction and effective_gas_price, but in reality it isn't, it is only a function of the transaction. The reason we pass effective_gas_price into execute_transaction is only because there is an opcode in the EVM to get the GASPRICE, and that opcode needs to use the effective_gas_price. However, unless the contract branches on the value of GASPRICE, the gas_used will be the same no matter what effective_gas_price is.

Well, if it can branch on the value of `GASPRICE`, which uses `effective_gas_price`, then `execute_transaction` _is_ a function of `effective_gas_price` isn't it?

> would you still find value in the additional assert after that function, even though in this case it is more clear that effective_gas_price isn't a dependent variable?

Is this code, or is this a spec/pseudocode? In the former case, I'd look at the method signature/documentation for `self.execute_transaction` and expect to see something there that would explain its return value. In the latter case -- and I thought this was intended as a spec/pseudocode, not as an implementation -- then I think a clarifying comment here would make this more readable and easier to understand.

MicahZoltu:
What about instead of an assert, just a comment like
```py
# the following function executes the transaction as normal, with the exception that the GASPRICE opcode will return `effective_gas_price`
gas_used = self.execute_transaction(transaction, effective_gas_price)
```
That would feel a bit better to me, as we then wouldn't be picking out one very specific part of what that function does and liflting it into this spec.

lrettig:
I'd find the following comment more helpful:

```
# Execute the transaction according to the rules of the protocol. The GASPRICE opcode will return `effective_gas_price` and `gas_used` cannot exceed `transaction.gas_limit.`
```

Issue #3132, opened 2020-11-25
Discussion for EIP-3135: Exclusive Claimable Token


Ungigdu:
---
eip: 3135
title: Exclusive Claimable Token
author: Zhenyu Sun (@Ungigdu)
discussions-to: https://github.com/ethereum/EIPs/issues/3132
status: Draft
type: Standards Track
category: ERC
created: 2020-08-10
requires: 20
---

## Simple Summary

This standard defines a token which can be claimed only by token issuer with payer's signature.

## Abstract

This EIP defines a set of additions to the default token standard such as ERC-20, that allows online/offline service providers establish micropayment channels with any number of users by signing and verifying messages about the consumption of token off chain. Using this mechanism will reduce interactions with blockchain to minimal for both participants, thus saving gas and improve performance.

## Motivation

There are two main purposes of this EIP, one is to reduce interactions with blockchain, the second is to link Ethereum to real-world payment problems.

Many small businesses want to build payment system based on blockchain but find it difficult. There are basically two ways: 

1. Directly pay with token. There are many wallet can receive and transfer token but transactions on Ethereum cost gas and take time to confirm.
2. User lock token on payment smart contract and service provider use payment messages signed by user to release token, establishing a micropayment channel. The advantage is interactions with blockchain is reduced and the signing/verifying process is off-chain. But interact with payment contract needs service provider to build a DApp, which require resources many small businesses do not have. Even if they managed to build DApps, they are all different, not standardized. Also, user should have a wallet with DApp browser and has to learn how to use it.

This EIP helps to standardize the interactions of micropayment system, and make it possible for wallet build a universal UI in the future.

## Specification

```solidity

/// @return Image url of this token or descriptive resources
function iconUrl() external view returns (string memory);

/// @return Issuer of this token. Only issuer can execute claim function
function issuer() external view returns (address);

/**
 *  @notice   Remove consumption from payer's deposite
 *  @dev      Check if msg.sender == issuer
 *  @param    from          Payer's address
 *  @param    consumption   How many token is consumed in this epoch, specified
 *  @param    epoch         Epoch increased by 1 after claim or withdraw, at the beginning of each epoch, consumption goes back to 0
 *  @param    signature     Signature of payment message signed by payer
*/
function claim(address from, uint256 consumption, uint256 epoch, bytes calldata signature) external;

function transferIssuer(address newIssuer) external;

/// @notice   Move amount from payer's token balance to deposite balance to ensure payment is sufficient
[...]

PR #3135, opened 2020-11-25
Exclusive Claimable Token

Ungigdu:
---
eip: 3135
title: Exclusive Claimable Token
author: Zhenyu Sun (@Ungigdu)
discussions-to: https://github.com/ethereum/EIPs/issues/3132
status: Draft
type: Standards Track
category: ERC
created: 2020-08-10
requires: 20
---

## Simple Summary

This standard defines a token which can be claimed only by token issuer with payer's signature.

## Abstract

This EIP defines a set of additions to the default token standard such as ERC-20, that allows online/offline service providers establish micropayment channels with any number of users by signing and verifying messages about the consumption of token off chain. Using this mechanism will reduce interactions with blockchain to minimal for both participants, thus saving gas and improve performance.

## Motivation

There are two main purposes of this EIP, one is to reduce interactions with blockchain, the second is to link Ethereum to real-world payment problems.

Many small businesses want to build payment system based on blockchain but find it difficult. There are basically two ways: 

1. Directly pay with token. There are many wallet can receive and transfer token but transactions on Ethereum cost gas and take time to confirm.
2. User lock token on payment smart contract and service provider use payment messages signed by user to release token, establishing a micropayment channel. The advantage is interactions with blockchain is reduced and the signing/verifying process is off-chain. But interact with payment contract needs service provider to build a DApp, which require resources many small businesses do not have. Even if they managed to build DApps, they are all different, not standardized. Also, user should have a wallet with DApp browser and has to learn how to use it.

This EIP helps to standardize the interactions of micropayment system, and make it possible for wallet build a universal UI in the future.

## Specification

```solidity

/// @return Image url of this token or descriptive resources
function iconUrl() external view returns (string memory);

/// @return Issuer of this token. Only issuer can execute claim function
function issuer() external view returns (address);

/**
 *  @notice   Remove consumption from payer's deposite
 *  @dev      Check if msg.sender == issuer
 *  @param    from          Payer's address
 *  @param    consumption   How many token is consumed in this epoch, specified
 *  @param    epoch         Epoch increased by 1 after claim or withdraw, at the beginning of each epoch, consumption goes back to 0
 *  @param    signature     Signature of payment message signed by payer
*/
function claim(address from, uint256 consumption, uint256 epoch, bytes calldata signature) external;

function transferIssuer(address newIssuer) external;

/// @notice   Move amount from payer's token balance to deposite balance to ensure payment is sufficient
function deposit(uint256 amount) external;

/**
 *  @notice   Give remaining deposite balance back to "to" account, act as "refund" function
 *  @dev      In prepayment module, withdraw is executed from issuer account
 *            In lock-release module, withdraw is executed from user account
 *  @param    to            the account receiving remaining deposite
 *  @param    amount        how many token is returned
*/
function withdraw(address to, uint256 amount) external;

function depositBalanceOf(address user) external view returns(uint256 depositBalance, uint256 epoch);

event Deposit(
    address indexed from,
    uint256 amount
);

event Withdraw(
    address indexed to,
    uint256 amount
);
    
event TransferIssuer(
    address indexed oldIssuer,
    address indexed newIssuer
);

event Claim(
    address indexed from,
    address indexed to,
    uint256 epoch,
    uint256 consumption
);

```

### signature

the pseudo code generating an ECDSA signature:
```
sign(keccak256(abi_encode(
    "\x19Ethereum Signed Message:\n32", 
        keccak256(abi_encode(
            token_address,
            payer_address,
            token_issuer,
            token_consumption,        //calculated by user client
            epoch
        ))
    ))
,private_key)

```

### verification process

the verification contains check about both signature and token_consumption

the pseudo code run by verification server is as follows:

```

serving_loop:

    for {
        /**
         * unpaied_consumption is calculated by provider
         * signed_consumption is claimable amount
         * tolerance allows payer "owes" provider to a certain degree
        */
        //getSignedConsumption returns amount that are already claimable 
        if(unpaied_consumption <  signed_consumption + tolerance){
            informUser("user need charge", unpaied_consumption)
            interruptService() 
        }else{
            isServing() || recoverService()
        }
    }

verification_loop:

    for {
        message = incomingMessage()
        if(recover_signer(message, signature) != payer_address){
            informUser("check signature failed", hash(message))
            continue
        }

        /**
        * optional: when using echo server to sync messages between verification servers
        * more info about this in Security Considerations section
        */
        if(query(message) != message){
            informUser("message outdate", hash(message))
            continue   
        }

        if(epoch != message.epoch || message.consumption > getDepositBalance()){
            informUser("invalid message", epoch, unpaied_consumption)
            continue
        }
       
        signed_consumption = message.consumption
        save(message)
    }
    
claim_process:

    if(claim()){
        unpaied_consumption -= signed_consumption
        signed_consumption = 0
        epoch+=1
    }

```
### About withdraw

The withdraw function is slightly different based on business models

1. prepayment model

In prepayment business model such as using token as recharge card of general store, the user pays (crypto)currency to store in advance for claimable token as recharge card (with bonus or discount). When checking out, the customer signs a message with updated consumption (old consumption + consumption this time) to store and store verifies this message off chain. The shopping process loops without any blockchain involved, until the customer wants to return the card and get money back. Because the store already holds all currency, the withdraw function should be executed by token issuer (store) to return remaining deposit balance after claim. The prepayment model can easily be built into a wallet with QR-code scanning function.

2. lock-release model

If we run a VPN service that accepts token as payment, we can use lock-release model. Unlike prepayment, we charge user X token for every Y bits went through. In this "pay for usage" scenario, the counting of services happens on both client and server side. The client should not trust charge amount given by server in case the it's malfunctioning or malicious. When client decide not to trust server, it stops signing messages, but some of token is taken hostage in deposit balance. To fix this problem, the withdraw function should be executed by payer account with limitation such as epoch didn't change in a month.

## Rationale

This EIP targets on ERC-20 tokens due to its widespread adoption. However, this extension is designed to be compatible with other token standard.

The reason we chose to implement those functions in token contract rather than a separate record contract is as follows:
- Token can transfer is more convenient and more general than interact with DApp
- Token is more standardized and has better UI support
- Token is equal to service, make token economy more prosperous
- Remove the approve process

## Backwards Compatibility

This EIP is fully backwards compatible as its implementation extends the functionality of [ERC-20](./eip-20.md).

## Implementation

```solidity

mapping (address => StampBalance) private _depositBalance;
    
struct StampBalance{
    uint256 balance;
    uint256 epoch;
}
    
function deposit(uint256 value) override external{
    require(value <= _balances[msg.sender]);
    _balances[msg.sender] = _balances[msg.sender].sub(value);
    _depositBalance[msg.sender].balance = _depositBalance[msg.sender].balance.add(value);
    emit Deposit(msg.sender, value);
}

function withdraw(address to, uint256 value) override onlyIssuer external{
    require(value <= _depositBalance[to].balance);
    _depositBalance[to].balance = _depositBalance[to].balance.sub(value);
    _depositBalance[to].epoch += 1;
    _balances[to] = _balances[to].add(value);
    emit Withdraw(to, value);
}
    
function depositBalanceOf(address user) override public view returns(uint256 depositBalance, uint256 epoch){
    return (_depositBalance[user].balance, _depositBalance[user].epoch);
}

// prepayment model
function claim(address from, uint credit, uint epoch, bytes memory signature) override onlyIssuer external{
    require(credit > 0);
    require(_depositBalance[from].epoch + 1 == epoch);
    require(_depositBalance[from].balance >= credit);
    bytes32 message = keccak256(abi.encode(this, from, _issuer, credit, epoch));
    bytes32 msgHash = prefixed(message);
    require(recoverSigner(msgHash, signature) == from);
    _depositBalance[from].balance = _depositBalance[from].balance.sub(credit);
    _balances[_issuer] = _balances[_issuer].add(credit);
    _depositBalance[from].epoch += 1;
    emit Claim(from, msg.sender, credit, epoch);
}

function prefixed(bytes32 hash) internal pure returns (bytes32) {
    return keccak256(abi.encode("\x19Ethereum Signed Message:\n32", hash));
}

function recoverSigner(bytes32 message, bytes memory sig) internal pure  returns (address) {
    (uint8 v, bytes32 r, bytes32 s) = splitSignature(sig);
    return ecrecover(message, v, r, s);
}

function splitSignature(bytes memory sig) internal pure returns (uint8 v, bytes32 r, bytes32 s) {
    require(sig.length == 65);
    assembly {
        r := mload(add(sig, 32))
        s := mload(add(sig, 64))
        v := byte(0, mload(add(sig, 96)))
    }
    return (v, r, s);
}

```

## Security Considerations

By restricting claim function to issuer, there is no race condition on chain layer. However double spending problem may occur when the issuer use multiple verifiers and payer signs many payment messages simultaneously. Some of those messages may get chance to be checked valid though only the message with the largest consumption can be claimed. This problem can be fixed by introducing an echo server which accepts messages from verifiers, returns the message sequentially with largest consumption and biggest epoch number. If a verifier gets an answer different from the message he send, it updates the message from echo server as the last message it receives along with local storage of the status about this payer. Then the verifier asks the payer again for a new message.

## Copyright

### DIFF
--- a/EIPS/eip-3135.md
+++ b/EIPS/eip-3135.md
@@ -20,36 +20,41 @@ This standard defines a token which can be claimed only by token issuer with pay
 
 When monitoring network traffic, counting paid e-mail messages or buy merchandise in general store, small but frequent payments are needed. We can't directly transfer tokens on chain to make payment because the sum gas cost is big and it takes time for blockchain to reach a consensus. 
 
-This standard allows service provider to establish micropayment channels with many users by creating claimable token. A user gets claimable token and can make part of balance "active" as deposit to this channel. When using service, user should sign messages off chain to service provider and provider checks if user is honest with accumulating amount in message and checks if this amount is less than active balance. If check passed, provider continue to serve user, otherwise reject. 
+This standard allows service provider to establish micropayment channels with many users by creating claimable token. A user gets claimable token and can take part of balance as deposit to this channel. When using service, user should sign messages off chain to service provider and provider checks if user is honest with the accumulating credit in message and checks if this amount is less than deposit balance. If check passed, provider continue to serve user, otherwise reject. 
 
-At anytime, the service provider can claim the right amount of token on chain providing the last signed message it accepts and this claim function can only be called by issuer to avoid double spending.
+At anytime, the service provider can claim the right amount of token on chain providing the last signed message it accepts. And this claim function can only be called by issuer to avoid double spending, that's the "exclusive" part of token.
 
 ## Motivation
 <!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
 There are two motivations for this standard, one is to reduce gas costs, the other is to link Ethereum to real-world payment problem: pay for usage.
 
-This token standard is useful in basically two types of businesses: 1. Real-world merchant (whose identity is known by user) can use this token as recharge card; 2. Online service provider (whose identity may not be known by user or is hard to reach) can use this token to build an automatic micropayment channel. In both cases, the token is the representation of provider's service/products, and should only be claimed by provider's "issuer" account.
+This token standard is useful in basically two types of businesses: 1. Real-world merchant (whose identity is known by user) can use this token as recharge card; 2. Online service provider (whose identity may not be known by user or is hard to reach) can use this token to build an automatic micropayment channel. In both cases, the token is the representation of provider's service/products, explained by provider and claimed by provider.
 
 ## Specification
 <!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (go-ethereum, parity, cpp-ethereum, ethereumj, ethereumjs, and [others](https://github.com/ethereum/wiki/wiki/Clients)).-->
 
 ```solidity
+pragma solidity  ^0.6;
+
+import "./IERC20.sol";
+// SPDX-License-Identifier: MIT
+
 interface IStamp is IERC20{
 
     function iconUrl() external view returns (string memory);
     function issuer() external view returns (address);
     function claim(address from, uint256 credit, uint256 epoch, bytes calldata signature) external;
     function transferIssuer(address newIssuer) external;
-    function active(uint256 amount) external;
-    function deactive(address to, uint256 amount) external;
-    function activeBalanceOf(address user) external view returns(uint256 balance, uint256 activedSum, uint256 epoch);
+    function deposit(uint256 amount) external;
+    function withdraw(address to, uint256 amount) external;
+    function depositBalanceOf(address user) external view returns(uint256 depositBalance, uint256 epoch);
 
-    event Active(
+    event Deposit(
         address indexed from,
         uint256 amount
     );
 
-    event Deactive(
+    event Withdraw(
         address indexed to,
         uint256 amount
     );
@@ -86,26 +91,26 @@ Transfer issuer role by changing its address.
 | ---------|-------------|
 | newIssuer | The new issuer address |
 
-#### active
+#### deposit
 
-User can make part of token balance 'active' as deposit to ensure micropayments. Token can only be claimed from active balance. This works likes 'increaseAllowance', except the active amount is removed from user's balance.
-
-| Parameter | Description |
-| ---------|-------------|
-| amount | amount to active |
-
-#### deactive
-
-This function should have different implementation based on business model. For real-world businesses, such as recharge card of general store or membership card of fitness center, deactive should be called by merchant, act as refund function to return active balance to user's balance; for online businesses, user may found it hard to contact with service provider, so the deactive is called by user itself, with limitations such as epoch number didn't change for some period of time. Also gets token in active balance back.
+User can take part of token balance as deposit to ensure micropayments. Token can only be claimed from deposit balance. This works likes 'increaseAllowance', except the deposit is removed from user's balance.
 
 | Parameter | Description |
 | ---------|-------------|
-| to | the account to perform deactive on |
-| amount | amount to active |
+| amount | amount to deposit |
 
-#### activeBalanceOf
+#### withdraw
 
-Returns balance, active balance and epoch of user account. The epoch is an increasing number which is added by one when account get claimed or deactive. Epoch is a part of sign data, verified by both user and provider, prevents double spend and double claim.
+This function should have different implementation based on business model. For real-world businesses, such as recharge card of general store or membership card of fitness center, withdraw should be called by merchant, act as refund function to return remaining deposit balance to user's balance; for online businesses, user may found it hard to contact with service provider, so the deposit is called by user itself, with limitations such as epoch number didn't change for some period of time. Also gets token in deposit back.
+
+| Parameter | Description |
+| ---------|-------------|
+| to | the deposit account |
+| amount | amount to return |
+
+#### depositBalanceOf
+
+Returns deposit balance and epoch of user account. The epoch is an increasing number which is added by one when account get claimed or withdraw. Epoch is a part of sign data, verified by both user and provider, prevents double spend and double claim.
 
 | Parameter | Description |
 | ---------|-------------|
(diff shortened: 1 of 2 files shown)


axic:
Was #3134 and #3133 opened and closed to get a specific EIP number?

Ungigdu:
> Was #3134 and #3133 opened and closed to get a specific EIP number?

Not intentionally. The second check somehow got stuck and I canceled to recheck

MicahZoltu:
In the future, there is no need to create a new PR.  The check gets stuck from time to time but it is something that can be fixed.

EIPS/eip-3135.md
```diff
@@ -0,0 +1,203 @@
...
+#### Claim
+
+Event emits when claim is called
+
+| Parameter | Description |
+| ---------|-------------|
+| from | user account to be claimed |
+| to | msg.sender, the issuer |
+| epoch | which epoch is based on |
+| credit | how many token be claimed |
```
MicahZoltu, line 167: Consider using [NatSpec](https://docs.soliditylang.org/en/v0.7.2/natspec-format.html) for inline comments in the interface, rather than a separate set of markdown definitions.  This would make this specification a bit more tool-friendly.

EIPS/eip-3135.md
```diff
@@ -0,0 +1,203 @@
...
+
+At anytime, the service provider can claim the right amount of token on chain providing the last signed message it accepts. And this claim function can only be called by issuer to avoid double spending, that's the "exclusive" part of token.
+
+## Motivation
+<!--The motivation is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.-->
+There are two motivations for this standard, one is to reduce gas costs, the other is to link Ethereum to real-world payment problem: pay for usage.
+
+This token standard is useful in basically two types of businesses: 1. Real-world merchant (whose identity is known by user) can use this token as recharge card; 2. Online service provider (whose identity may not be known by user or is hard to reach) can use this token to build an automatic micropayment channel. In both cases, the token is the representation of provider's service/products, explained by provider and claimed by provider.
```
MicahZoltu, line 31: The motivation section should explain why this needs to be a Standard rather than just a Good Idea.  It seems to me that every dapp can use a slightly different payment system and that is OK as they'll all be building their own UI anyway.  What interoperability benefits do we gain by having a universal standard for payment channels?

EIPS/eip-3135.md
```diff
@@ -0,0 +1,203 @@
...
+| to | msg.sender, the issuer |
+| epoch | which epoch is based on |
+| credit | how many token be claimed |
+
+## Rationale
+This standards provides a way for user and service provider to establish micropayment channel to make free and repaid payment messages off blockchain. Those payment messages are linked one by one, providing a trace of service consumption and keeps an increasing number of credit, which is how much the payer owned to the service provider.
+
+The provider can stop serving when payer no longer send new valid payment message or the credit is bigger than deposit balance. The user can stop signing new message if provider is not available or is cheating. The provider can claim token on chain at any time using the message contains the largest credit number.
+
+Let's explain the usage of exclusive claimable token by two cases.
```
MicahZoltu, line 174: This seems to be more appropriate for the abstract.  The Rationale section should focus on specific choices this spec made that either do not have obvious reasons or may have alternative options available.  For example, "We chose constant X because ..." or "We went with the function name Y because ...".

EIPS/eip-3135.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Token is equal to service, and can be transferred
+- Token can act as stock of other valuable token
+- Token consumes only when server servers, no trust needed
+- Interaction with blockchain is minimal
+
```
MicahZoltu, line 192: Example use cases are more appropriate for Motivation section rather than Rationale section.

EIPS/eip-3135.md
```diff
@@ -0,0 +1,203 @@
...
+#### claim
+
+Service provider (issuer) recovers unsigned message, then checks the message with it's signature (probably using built-in function 'ecrecover'). If the signature is successfully verified, provider get "credit" amount token from user's deposit balance and epoch is increased by 1. It is worth mentioning that the provider may not need to get token back because this token is basically created by this provider and may holds no value to itself. In that case, the claim function can release some other more valuable token to provider. 
+
+| Parameter | Description |
+| ---------|-------------|
+| from | which user account to be claimed |
+| credit | how many token should be claimed |
+| epoch | current epoch number |
+| signature| micropayment message signature signed by user|
```
MicahZoltu, line 128: This is very underspecified.  The `signature` parameter specifically needs to describe exactly what the signature format is (encoding mechanism, signing mechanism, etc.)

MicahZoltu (changes requested):
The frontmatter changes are required before merging as draft.  The rest of it can wait until after draft, but I strongly recommend addressing it all.

I would very much like to at least hear the argument for why this should be a Standard instead of a Good Idea before merging as draft, but that isn't a hard requirement.

EIPS/eip-3135.md
Ungigdu: Thank you for pointing this out. Well, for online service which needs to sign micropayment messages frequently, client should have a open wallet for signing. And payment systems are slightly different, as you say, it's not necessary to make it a standard. But for real-world businesses, like a coffee shop, payment signing only happens when checking out. A wallet like Imtoken or metamask is capable to do both deposit and signning, and the pay system is just the same. Many small businesses do not have the resourece to develop their own Dapp. So I think it may be useful to make it a standard.

PR #3139, opened 2020-11-27
EIP-2681: move to last call

lightclient:
Moving this to last call since there was no dissent in ACD 101.

### DIFF
--- a/EIPS/eip-2681.md
+++ b/EIPS/eip-2681.md
@@ -27,7 +27,7 @@ Lastly, this facilitates a minor optimisation in clients, because the nonce no l
 Introduce two new restrictions retroactively from genesis:
 
 1. Consider any transaction invalid, where the nonce exceeds `2^64-1`.
-2. The `CREATE` instruction to abort with an exceptional halt, where the account nonce is `2^64-1`.
+2. The `CREATE` and `CREATE2` instructions to abort with an exceptional halt, where the account nonce is `2^64-1`.
 
 ## Rationale
 
--- a/EIPS/eip-2681.md
+++ b/EIPS/eip-2681.md
@@ -4,6 +4,7 @@ title: Limit account nonce to 2^64-1
 author: Alex Beregszaszi (@axic)
 discussions-to: https://ethereum-magicians.org/t/eip-2681-limit-account-nonce-to-2-64-1/4324
 status: Last Call
+review-period-end: 2020-12-18
 type: Standards Track
 category: Core
 created: 2020-04-25
--- a/EIPS/eip-2681.md
+++ b/EIPS/eip-2681.md
@@ -23,7 +23,7 @@ Lastly, this facilitates a minor optimisation in clients, because the nonce no l
 
 ## Specification
 
-If `block.number >= FORK_BLOCK` introduce two new restrictions:
+Introduce two new restrictions retroactively from genesis:
 
 1. Consider any transaction invalid, where the nonce exceeds `2^64-1`.
 2. The `CREATE` instruction to abort with an exceptional halt, where the account nonce is `2^64-1`.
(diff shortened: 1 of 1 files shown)


MicahZoltu:
I would still like to see this changed to `2^52` instead of `2^64-1`.  I can't think of *any* advantages to `2^64-1`, and I can think of at least 2 reasons (JavaScript and JSON) for `2^52`, so to me it seems like an obvious win.

lightclient:
@MicahZoltu what is stopping JS libraries from just interpreting nonces as `2^52` sized integers? Many clients are already not in compliance with spec since they already represent nonces as `u64`s. I feel like JS libraries should be able to do the same as we won't reasonably reach that point in the next 100 years.

axic:
> If `block.number >= FORK_BLOCK` introduce two new restrictions:

If this is enabled from genesis (as discussed), the above line should be changed.

axic:
> I would still like to see this changed to 2^52 instead of 2^64-1. I can't think of any advantages to 2^64-1, and I can think of at least 2 advantages (JavaScript and JSON) to 2^52, so to me it seems like an obvious win.

Should we discuss this on the forum instead? There's a lot to unpack.

EIPS/eip-2681.md
```diff
@@ -23,7 +23,7 @@ Lastly, this facilitates a minor optimisation in clients, because the nonce no l
 
 ## Specification
 
-If `block.number >= FORK_BLOCK` introduce two new restrictions:
+Introduce two new restrictions retroactively from genesis:
 
 1. Consider any transaction invalid, where the nonce exceeds `2^64-1`.
 2. The `CREATE` instruction to abort with an exceptional halt, where the account nonce is `2^64-1`.
```
axic, line 29: Even if `CREAT2` doesn't use the nonce as an input, does it actually increments it? Need to double check this.

lightclient: @axic very good point. `CREATE2` does increment the nonce. I've updated the EIP.

lightclient:
@axic - okay this should be good to move to last call now.

eip-automerger:

 - Trying to change EIP 2681 state from Review to Last Call
 - EIP 2681 requires approval from one of (@axic)

MicahZoltu:
I would still really like to see this get changed to `2^52`.  There is no reason to prefer a standardized limit of 2^64 over 2^52.

lightclient:
My stance on that remains the same. I don't think we should consider the capabilities of javascript when making a core protocol decision. Especially when it will take decades (centuries?) to send enough transactions for this to matter. `2^64` is more standard and lends itself naturally to overflow detection in clients.

MicahZoltu:
> 2^64 is more standard and lends itself naturally to overflow detection in clients.

Can you expand on this?  Is the idea that you can lean on native overflow detection rather than having to do checks like `if (x > x +y) throw`?

lightclient:
While writing code, it is more natural to consider what would should happen in the event of an overflow. Checking `x < 2^52` requires developers to *know* that constraint exists. It's not a strong argument, client devs already have many constraints they must be aware of.

MicahZoltu:
> My stronger argument is that javascript capabilities should not drive protocol design decisions.

I actually think this is the weaker of the two arguments.  JavaScript is probably the language most widely used in the ecosystem and despite all of us wishing it would go away I suspect it will be around for some time.  We have also seen many instances of errors stem from JavaScript double overflows.  Making one less problem that JS developers have to worry about is a net positive IMO, at least on par with giving client developers one less constraint to manually check for.

lightclient:
I don't think I'm going to convince you on this, and I am very certain in my position. We'll have to have @axic make the call since he is the author of this EIP.

MicahZoltu:
In the meantime, I'm moving forward with https://eips.ethereum.org/EIPS/eip-3338 since this has been sitting for ~5 months.  If there is sufficient desire to do 2^64, we can easily "supersede" 3338 with this one.

lightclient:
@axic does this look good to move to last call?

axic (approved):

alita-moore:


	 - EIP 2681 state was changed from review to last call
 @axic

Issue #2967, opened 2020-09-10
EIP-1: Stronger rules for discussion-url


axic:
This is sparked by [EIP-1046](https://eips.ethereum.org/EIPS/eip-1046) and [EIP-1047](https://eips.ethereum.org/EIPS/eip-1047) linking to reddit threads:
- https://www.reddit.com/r/raredigitalart/comments/8hfh1g/erc20_metadata_extension_eip_1046/
- https://www.reddit.com/r/raredigitalart/comments/8hfk2a/token_metadata_json_schema_eip_1047/

Both of those now display this:
![Screenshot](https://user-images.githubusercontent.com/20340/92815625-54c7c180-f3bc-11ea-8d28-58c46d14c0b2.png)

That means their discussion urls are read-only, which defeats the purpose.

---

I do not have a clear proposal here, but only a conversation starter. Likely allowing the following makes the most sense based on past experience:
- `https://github.com/ethereum/EIPs/issues/*`
- `https://ethereum-magicians.org/*`
- `https://ethresear.ch/*`

The other currently used URLs are:
- reddit (as above)
- `t.me` (telegram)
- `https://discuss.ens.domains/t/new-standard-proposal-ens-multicoin-support/1148`
- `https://gitter.im/ethereum/topics/topic/5ac4d974109bb043328911ce/eip-969-discussion`
- `https://gitter.im/ethereum/EIPs`
- `https://gitter.im/ethereum/AllCoreDevs`
- and countless cases of `github.com/<username>/issues/<n>`

We already disallow `https://github.com/ethereum/EIPs/pull/*` (as an archaic decision), we could perhaps just disallow reddit as well?

Probably we need to ask the question here: what is the goal of the discussion url? Is it for ephemeral discussions only? Is it something which is important after something becomes Final? If so, should it be removed after it became Final?

MicahZoltu:
> Probably we need to ask the question here: what is the goal of the discussion url?

I have wondered this myself on many occasions.  I would support a change that makes the discussion URL go away at final.  This would be similar to RFCs where the final RFC doesn't include a link to historical discussions.  If people really wanted to find the historical discussion, they can dig through git history so it isn't like we are purging it from the internet.

This policy would also, importantly, let us eventually close GitHub issue discussions so we don't have an ever growing list of open issues.

lightclient:
I am in favor of this also. However, my preference would be to disallow `ethresear.ch` topics.

BrentAllsop:
[...]

PR #3143, opened 2020-12-01
Proposal to increase block rewards to 5 ETH

Terra854:
This EIP proposes a block reward increase to the original block reward of 5 ETH when Ethereum first started.


EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
+---
+eip: <to be assigned>
```
lightclient, line 2:
```suggestion
eip: 3143
```

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+title: Increase block rewards to 5 ETH
+author: Ben Tinner (@Terra854)
+discussions-to: https://www.reddit.com/r/ethereum/comments/k4252u/thoughts_about_increasing_block_rewards_till_we/
```
lightclient, line 5: Please use a forum like http://ethereum-magicians.com instead of Reddit. It is a more suitable long-term destination for discussion (see #2967).

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+status: Draft
+type: Standards Track
+category Core
```
lightclient, line 8:
```suggestion
category: Core
```

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Specification
+All we need to do is to modify the parameters that define the block reward inside the source code
```
lightclient, line 22: Please use this section to clearly specify what changes should occur in the protocol. I recommend reviewing some other core EIPs to get an idea of formatting: https://eips.ethereum.org/core

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Rationale
+The ultimate end goal for this EIP is to give POW miners more incentive to switch to POS once ETH 2.0 is fully implemented since the transition will take a few years to complete and during that time, they will be incentivised to hold on to the tokens instead of selling it straightaway in order to prepare to be a validator for ETH 2.0, reducing the selling pressure on ETH and increasing it's value in the long run. A side effect of miners staying on Ethereum is that network security will be assured during the transition period.
+
+## Backwards Compatibility
+This EIP is not forward compatible as there will be changes to the block, uncle and nephew reward structure. Therefore, it should be included in a scheduled hardfork at a certain block number.
```
lightclient, line 28: I think you mean backwards compatible? Also, these sorts of things are implied by changes to the core protocol.

lightclient (changes requested):

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+author: Ben Tinner (@Terra854)
+discussions-to: https://www.reddit.com/r/ethereum/comments/k4252u/thoughts_about_increasing_block_rewards_till_we/
+status: Draft
+type: Standards Track
+category Core
+created: 2020-12-01
+---
+
+## Simple Summary
+This EIP will increase the block reward paid to proof-of-work (POW) miners, giving them more opporturnities to acquire the minimum number of ETH needed to run a standalone proof-of-stake (POS) validator node before full implementation of ETH 2.0
```
MicahZoltu, line 13:
```suggestion
Changes the block reward paid to proof-of-work (POW) miners to 5 ETH.
```

MicahZoltu: Simple summary should just state in very few words what this EIP does.  It should not include motivation or reasoning.

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Abstract
+By increasing the block rewards back to the original 5 ETH when the network first started, it will convince POW miners to stay on mining Ethereum without worrying too much on profitability. It will also give smaller POW miners more opporturnities to be able to acquire the minimum amount of ETH needed to become a POS validator before we transition fully to ETH 2.0
```
MicahZoltu, line 16: This should be moved to the Motivation section.  The `Abstract` should be a short, human readable, version of the specification.  The Motivation section is where you want to talk about *why* this EIP is a good idea.

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Specification
+All we need to do is to modify the parameters that define the block reward inside the source code
+
+## Rationale
+The ultimate end goal for this EIP is to give POW miners more incentive to switch to POS once ETH 2.0 is fully implemented since the transition will take a few years to complete and during that time, they will be incentivised to hold on to the tokens instead of selling it straightaway in order to prepare to be a validator for ETH 2.0, reducing the selling pressure on ETH and increasing it's value in the long run. A side effect of miners staying on Ethereum is that network security will be assured during the transition period.
```
MicahZoltu, line 25: The rationale section should describe why certain values were chosen.  For this EIP, this would be a good place to mention why 5 ETH was chosen instead of 4 ETH or 6 ETH (or any other value).  The content currently in here is more appropriate for the Motivation section.

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Backwards Compatibility
+This EIP is not forward compatible as there will be changes to the block, uncle and nephew reward structure. Therefore, it should be included in a scheduled hardfork at a certain block number.
+
+## Security Considerations
+There will be short term economic drawbacks such as a significantly higher inflation rate (estimated 10.4% per annum for the first year) for a few years, but I believe that these concerns will be moot once the transition is complete.
```
MicahZoltu, line 31: We generally don't consider economic concerns to be "security concerns" in the sense of the word that this section is intended for.  Also, increasing miner payout increases network security, at the cost of economic stability, so this particular issue isn't a security concern.

EIPS/Increase block rewards to 5 ETH.md
Terra854: Well, I checked the EIPs that involved adjusting the block rewards (specifically EIP-649 and EIP-1234) and they used the 'forward compatible' language to describe this.

EIPS/Increase block rewards to 5 ETH.md
MicahZoltu: In particular, I recommend looking at past EIPs that have changed the block reward.

MicahZoltu (changes requested):

EIPS/Increase block rewards to 5 ETH.md
MicahZoltu: Bleh.  I don't know why those EIPs say what they do, but you are right that what you have matches them.  For now I would say just delete it and say something like:
```suggestion
There are no known backward compatibility issues with the introduction of this EIP.
```

Maybe one day we can fix 649 and 1234...

EIPS/Increase block rewards to 5 ETH.md
Terra854: So is it fine that I leave out this portion? Because the guidelines require this to be filled up in order to not be outright rejected and i can't think of one. Furthermore, the EIPs that proposed a change to block rewards (specifically EIP-649 and EIP-1234) left this part out.

MicahZoltu: The security considerations **section** must be included, but it can just be a one-liner indicating that there are no known security issues presented by this change.  During review, if someone brings up a security issue it can be added, if not then it can go to final with just that.

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,56 @@
...
+
+Therefore, the ultimate end goal for this EIP is to give POW miners more incentive to switch to POS once ETH 2.0 is fully implemented since the transition will take a few years to complete and during that time, they will be incentivised to hold on to the tokens instead of selling it straightaway in order to prepare to be a validator for ETH 2.0, reducing the selling pressure on ETH and increasing it's value in the long run. A side effect of miners staying on Ethereum is that network security will be assured during the transition period.
+
+## Specification
+#### Adjust Block, Uncle, and Nephew rewards
+To ensure a constant Ether issuance, adjust the block reward to `new_block_reward`, where
```
MicahZoltu, line 25:
```suggestion
Adjust the block reward to `new_block_reward`, where
```
I'm guessing this is just a copy/paste mistake from another EIP, since this EIP specifically aims to change the ETH issuance.  😄

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Backwards Compatibility
+There are no known backward compatibility issues with the introduction of this EIP.
+
+## Test Cases
+Test cases shall be created once the specification is to be accepted by the developers or implemented by the clients.
+
+## Implementation
+The implementation in it's logic does not differ from [EIP-649](./eip-649.md) or [EIP-1234](./eip-1234.md); an implementation shall be created once the specification is to be accepted by the developers or implemented by the clients.
+
```
MicahZoltu, line 54: This is not a required section, so I recommend just leaving it out rather than including it as empty.

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,56 @@
...
+## Rationale
+I have decided to go with a 5 ETH base reward as i believe that it is a good middle ground between wanting to prevent too high of an inflation rate (10.4% per annum for the first year if we go with 5 ETH) and the desires to convert as many POW miners as possible into POS validators by making it easier to amass the required ETH needed through POW mining.
+
+## Backwards Compatibility
+There are no known backward compatibility issues with the introduction of this EIP.
+
+## Test Cases
+Test cases shall be created once the specification is to be accepted by the developers or implemented by the clients.
+
```
MicahZoltu, line 51: This is no longer a required section, so I recommend leaving it out rather than including it and having it be empty.

EIPS/Increase block rewards to 5 ETH.md
```diff
@@ -0,0 +1,56 @@
...
+This is the existing formula for uncle rewards, simply adjusted with `new_block_reward`.
+
+The nephew reward for `block.number >= FORK_BLKNUM` is
+
+    new_nephew_reward = new_block_reward / 32
+
+This is the existing formula for nephew rewards, simply adjusted with `new_block_reward`.
+
+## Rationale
+I have decided to go with a 5 ETH base reward as i believe that it is a good middle ground between wanting to prevent too high of an inflation rate (10.4% per annum for the first year if we go with 5 ETH) and the desires to convert as many POW miners as possible into POS validators by making it easier to amass the required ETH needed through POW mining.
```
MicahZoltu, line 44: Nitpick: Consider rephrasing this from "I have decided" to a statement of fact.  Ideally, EIPs are written for a target audience of future readers after the EIP has been accepted and implemented somewhere.

```suggestion
A 5 ETH base reward was chosen as a middle ground between wanting to prevent too high of an inflation rate (10.4% per annum for the first year at 5 ETH per block) and converting as many POW miners as possible into POS validators by making it easier to amass the required ETH needed through POW mining.
```

MicahZoltu (approved):
Structure generally looks good.  I left a few more points of feedback that I recommend considering prior to moving this to review, but nothing that would hold this up from being merged as a draft.

MicahZoltu:
I'll give 24 hours for @Terra854 to make any last changes based on my recent round of feedback and then merge after that.

EIPS/eip-3143.md
```diff
@@ -0,0 +1,53 @@
...
+
+Therefore, the ultimate end goal for this EIP is to give POW miners more incentive to switch to POS once ETH 2.0 is fully implemented since the transition will take a few years to complete and during that time, they will be incentivised to hold on to the tokens instead of selling it straightaway in order to prepare to be a validator for ETH 2.0, reducing the selling pressure on ETH and increasing it's value in the long run. A side effect of miners staying on Ethereum is that network security will be assured during the transition period.
```
MicahZoltu, line 21:
```suggestion
Therefore, the ultimate end goal for this EIP is to give POW miners more incentive to switch to POS once ETH 2.0 is fully implemented since the transition will take a few years to complete and during that time, they will be incentivised to hold on to the tokens instead of selling it straight away in order to prepare to be a validator for ETH 2.0, reducing the selling pressure on ETH and increasing it's value in the long run. A side effect of miners staying on Ethereum is that network security will be assured during the transition period.
```
CI won't let me merge without this.

shannonlitton:
4 for sure

KaycoinsInc (approved):

KaycoinsInc (approved):

PR #3148, opened 2020-12-03
Add radicle field to 634

ligi:
when we have github we should also support the decentralized alternative radicle

### DIFF
--- a/EIPS/eip-634.md
+++ b/EIPS/eip-634.md
@@ -50,6 +50,7 @@ services must be prefixed with `vnd.`.
 
 - **vnd.github** - a GitHub username
 - **vnd.peepeth** - a peepeth username
+- **vnd.radicle** - a radicle ID
 - **vnd.twitter** - a twitter username
 
 


froid1911:
great idea!

eip-automerger:

 - EIP 634 requires approval from one of (@ricmoo)

ricmoo:
Do you own a .eth or DNS TLD for this? I'll be updating the EIP shortly, instead of using vnd, it will recommend using reverse domain resolution. So, if you for example own radicle.eth or radicle.com, you could instead use the ENS field `eth.radicle.profiles` or `com.radicle.parofiles`, for example.

The big advantage is that this EIP does not need to be updated with each new service, as the owner of a namespace can simply start using it.

ligi:
radicle has radicle.xyz - not sure about radicle.eth.  But AFAIR xyz is also supported by ENS.
paging @abbey-titcomb for this

ricmoo:
Yeah, that's fine too. Basically all that matters is there is a canonical namespace that the project owns. So, `xyz.radicle` (or perhaps `xyz.radicle.profiles`) works well for a key. :)

abbey-titcomb:
We own `radicle.eth` and  `radicle.xyz`, so either could work :)

Thanks for pushing this forward @ligi !

MicahZoltu:
Given the discussion above, can this EIP be closed?

ricmoo:
@MicahZoltu do you mean EIP-634 or the PR 3148? Is the term EIP and PR interchangeable?

I have updated 634, so it should now encompass 3148.

ligi:
A bit disappointed about the change to be honest as it now mentions com.github (centralized) - but not mentions radicle (decentralized) - so I do not think the update to 634 encompasses 3148

MicahZoltu:
Sorry!  Typo on my part, I was asking if we can close this PR.  😬

ricmoo:
@ligi 

Oh, I can add them if they (is it someone you represent?) agree.

The main reason for including GitHub is because it is something that is somewhat necessary today, but they (GitHub) are unlikely to officially announce that want `com.github` associated with their customers.

While a decentralized platform is quite likely to advertise a "please use `xyz.radicle` for your ENS key" somewhere, and that is the direction the EIP is favouring; services announcing their keys rather than some document providing that list.

I would ideally not refer to GitHub either. It is not meant to be an exhaustive list, and is the reason to move to using reverse-dot-notation, so we don't need an exhaustive list. Anyone can just *try* `xyz.radicle` and see it is works.

Does that make sense? If radicle would like to be inlaced among the examples, I can add them, but that is not something I plan to make a habit of in the future, especially once the EIP moves to final.

ricmoo:
(note: the goal of the EIP is not to advertise or promote any service; Twitter, GitHub, etc. are fairly established and well-known, where as prior to this PR I had never heard of radicle)

ligi:
@ricmoo thanks!

> Oh, I can add them if they (is it someone you represent?) agree.

No I do not represent - that's @abbey-titcomb - I just think it is a good idea to add them here.

> (note: the goal of the EIP is not to advertise or promote any service; Twitter, GitHub, etc. are fairly established and well-known, where as prior to this PR I had never heard of radicle)

Yea - radicle just recently reached Beta - so it is not as well known as Github yet. By listing github here you further cement this status quo - but IMHO especially web3 projects should IMHO distribute their code in a decentralized manner and not solely on a centralized service.

ricmoo:
> Yea - radicle just recently reached Beta

That seems like more reason to not include it, then... :s

> By listing github here you further cement this status quo

Keep in mind the goal here is to list **well-established** *social media sites*, for which GitHub certainly qualifies (whether it should be the de facto standard, etc. or used for web3 is orthogonal and unrelated to the reasons it is included).

It is in these services best interest to decide their own key, for example, they may prefer `xyz.radicle.profiles`, and advertise it themselves. For example, radicle should consider approaching ENS and having their key (that they choose) added to the [app.ens.domains](app.ens.domains) tool.

The keys provided are there because it is unlikely that GitHub, Twitter, etc. are to take any official interest in ENS any time soon and those are keys people want to use today.

Does that make sense?

ligi:
Just my 2 cents on this:

I think one of the missions of Ethereum and Decentralisation in a wider sense is resilience. One building block of this is unstoppable apps and I think Ethereum can recover from wide infrastructure outtakes. But another building block is access to the code behind the apps - this also enables keeping it alive and dynamic instead of just having a (maybe suboptimal) running version. Currently if github vanishes this will cause major loss/interruption. By having a decentralized alternative listed in the EIP the chances increase that libraries that support this EIP add support - and this increases chances that projects add a decentralized reference to their code to the ENS page of their project.

ricmoo:
I don't disagree that a broader goal (in general) should be to encourage decentralized solutions.

But this EIP is more about a generic solution that any decentralized service can adopt. The explicit list provided is meant to cover well-established social networks, especially those which are not likely (for now) to participate voluntarily, so that they can be used in the meantime.

Decentralized solutions that wish to be included should approach the ENS tools (such as the [ENS manager](https://github.com/ensdomains/ens-app)) directly, opening issues or [making a PR](https://github.com/ensdomains/ens-app/blob/fe58b02b25e4ed945d2fa685d4cd1f72c647a8b7/src/components/Links/RecordLink.js) to have their key added.

Does that make sense? If so, I think this EIP can be closed?

tjayrush:
In response to @ligi's point, which I think is very well put, and @ricmoo's points (less convincing) I wonder why you choose to explicitly list 'vnd' types. Why not an arbitrarily long list.

Instead of:

```
vnd.github - a GitHub username
vnd.peepeth - a peepeth username
vnd.radicle - a radicle ID
vnd.twitter - a twitter username
```

Which clearly doesn't scale over time and requires the community to change the standard if it changes its mind, perhaps

```
[{ type: github, username: user}, {type: twitter, username: whatsis},{type: anything..., username: whatsis}...]
```

would be more future proof.

Maybe I'm missing something.

ricmoo:
The `vnd.` was removed from the EIP in favour of reverse-dot-notation for that very reason. :)

If you own “radicle.xyz”, the key `xyz.radicle` Is yours implicitly to carve up how you wish. You can just use `xyz.radicle` (or `xyz.radicle.repos` and `xyz.radicle.users` or any sub-domain or combination of sub-domains) without needing to update any central registry or EIP.

My main point was that an EIP should not be an advertising platform and so unestablished services don’t make as much sense to include.

ricmoo:
For the latest incarnation , see https://eips.ethereum.org/EIPS/eip-634. :)

PR #3150, opened 2020-12-03
2930: Adds receipt and non-normative change to improve readability.

MicahZoltu:
We forgot to define the receipt structure for this transaction (required by 2718).  I went with a mirror of legacy transaction receipts, though I am open to some alternative structure.

Some of the stuff we were defining here is already defined in 2718, and I believe redefining it was adding confusion and making iteration on this spec more challenging than it needed to be.


eip-automerger:

holiman (approved):

PR #3156, opened 2020-12-07
ERC3156 - Flash Loans

albertocuestacanada:
Flash loans allow smart contracts to lend an amount of tokens without a requirement for collateral, with the condition that they must be returned within the same transaction.

The aim of this ERC is to provide a standard interfaces and processes for flash lenders and borrowers, allowing for flash loan integration without a need to consider each particular implementation.


EIPS/eip-flash-loans.md
```diff
@@ -0,0 +1,165 @@
+---
+eip: <to be assigned>
```
lightclient, line 2:
```suggestion
eip: 3156
```

EIPS/eip-flash-loans.md
```diff
@@ -0,0 +1,165 @@
...
+title: Flash Loans
+author: Alberto Cuesta Cañada (@albertocuestacanada), Fiona Kobayashi (@fifikobayashi), fubuloubu (@fubuloubu)
+discussions-to: https://ethereum-magicians.org/t/flash-loan-eip-early-draft/4993/2
+status: Draft
+type: Standards Track
+category: ERC
+created: 2020-11-15
+---
+
+<!--This is the suggested template for new EIPs.
```
lightclient, line 12: Please remove these comments.

EIPS/eip-flash-loans.md
```diff
@@ -0,0 +1,165 @@
...
+
+Note that an EIP number will be assigned by an editor. When opening a pull request to submit your EIP, please use an abbreviated title in the filename, `eip-draft_title_abbrev.md`.
+
+The title should be 44 characters or less.-->
+
+## Simple Summary
+
+Flash loans allow smart contracts to lend an amount of tokens without a requirement for collateral, with the condition that they must be returned within the same transaction.
```
lightclient, line 20: The simple summary should generally one sentence with no motivation.

EIPS/eip-flash-loans.md
```diff
@@ -0,0 +1,165 @@
...
+
+The aim of this ERC is to provide a standard interfaces and processes for flash lenders and borrowers, allowing for flash loan integration without a need to consider each particular implementation.
+
+## Abstract
```
lightclient, line 24: The abstract should be a technical overview of the EIP. I don't recommend it for an ERC.

EIPS/eip-flash-loans.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Backwards Compatibility
+
+No backwards compatibility issues identified.
+
+## Test Cases
+
+[WETH10](https://github.com/WETH10/WETH10/pull/81), [MakerDAO](https://github.com/hexonaut/dss-flash/pull/18).
+
+## Implementation
```
lightclient, line 107: It's preferred for a reference implementation to be included inline.

EIPS/eip-flash-loans.md
```diff
@@ -0,0 +1,165 @@
...
+
+[WETH10](https://github.com/WETH10/WETH10/pull/81), [MakerDAO](https://github.com/hexonaut/dss-flash/pull/18).
+
+## Security Considerations
+
+### Flash lending security considerations
+
+
```
lightclient, line 115:
```suggestion
```

EIPS/eip-flash-loans.md
```diff
@@ -0,0 +1,165 @@
...
+The coupling of flash minting with business specific features in the same platform can easily lead to unintended consequences.
+
+#### Example - Treasury draining
+In early implementations of the Yield Protocol flash loaned fyDai could be redeemed for Dai, which could be used to liquidate the Yield Protocol CDP vault in MakerDAO:
+1. Flash mint a very large amount of fyDai.
+2. Redeem for Dai as much fyDai as the Yield Protocol collateral would allow.
+3. Trigger a stability rate increase with a call to `jug.drip` which would make the Yield Protocol uncollateralized.
+4. Liquidate the Yield Protocol CDP vault in MakerDAO.
+    
+
```
lightclient, line 161: Please remove extra spacing.

lightclient (changes requested):

EIPS/eip-flash-loans.md
albertocuestacanada: Fixed

EIPS/eip-flash-loans.md
albertocuestacanada: Fixed

EIPS/eip-flash-loans.md
albertocuestacanada: Please check again, thanks.

EIPS/eip-flash-loans.md
albertocuestacanada: Fixed

EIPS/eip-flash-loans.md
albertocuestacanada: Fixed

EIPS/eip-flash-loans.md
albertocuestacanada: Please check again, thanks.

EIPS/eip-flash-loans.md
albertocuestacanada: Included now. What should I do with regards to tests? Would a link be enough?

lightclient: A link is okay. Can you update the link to where the test cases are, rather than a PR please?

aoaham (approved):

MickdeGraaf:
Standardizing flashloans would be great! One thing this proposal does not take into accounts is contracts which hold multiple assets like PieDAO pies. I propose to add an extra param to the flashloan function to signal which asset the receiver is expecting like this:

```
function flashLoan(address receiver, address asset, uint256 amount, bytes calldata data) external {
  ...
  FlashBorrowerLike(receiver).onFlashLoan(msg.sender, asset, amount, fee, data);
  ...
}
```

Flash minting an asset could be done by passing the address of the contract itself to the ``asset`` param.

I've did a POC implementation a while back and the interface looks pretty much the same:

https://github.com/pie-dao/ExperiPie/blob/feature/flash-loans/contracts/facets/FlashLoan/FlashLoanFacet.sol
https://github.com/pie-dao/ExperiPie/blob/feature/flash-loans/contracts/facets/FlashLoan/FlashLoanMint.sol

fubuloubu:
> Standardizing flashloans would be great! One thing this proposal does not take into accounts is contracts which hold multiple assets like PieDAO pies.

@MickdeGraaf would it be possible to use the `data` parameter for this type of interaction? Why would adding an `asset` field be generally useful, or avoid an extra security assumption?

MickdeGraaf:
> > Standardizing flashloans would be great! One thing this proposal does not take into accounts is contracts which hold multiple assets like PieDAO pies.
> 
> @MickdeGraaf would it be possible to use the `data` parameter for this type of interaction? Why would adding an `asset` field be generally useful, or avoid an extra security assumption?

Not using the ``data`` parameter would avoid some confusion for implementers because they don't have to make an distinction between multi and single asset contracts but I guess it would work.

Additionally it would be nice to have some getters allowing me to get a quote from a contract on how much fee I would have to pay and if there is sufficient liquidity for my flashloan.

fubuloubu:
> Not using the data parameter would avoid some confusion for implementers because they don't have to make an distinction between multi and single asset contracts but I guess it would work.
>
> In most circumstances I would expect the data to only be actually used by flashLoanBorrower and not the lender itself.

So, the intent of the `data` field was the facilitate more complex interactions with flash loan borrowing, and be flexible enough to support innovative use cases like yours where the intention is not generic enough to be useful for others.

---

> Additionally it would be nice to have some getters allowing me to get a quote from a contract on how much fee I would have to pay and if there is sufficient liquidity for my flashloan.

This is a great idea!

MickdeGraaf:
> So, the intent of the `data` field was the facilitate more complex interactions with flash loan borrowing, and be flexible enough to support innovative use cases like yours where the intention is not generic enough to be useful for others.

Makes sense. I do think however that a very large portion of contracts is a token itself which holds one or more other tokens. In these cases the ``asset`` param would be a welcome addition imo.

albertocuestacanada:
Hi @MickdeGraaf, thanks a lot for your suggestion.

I think that for projects that hold and lend multiple assets an architecture like Uniswap would work with proposed standard. In Uniswap you go to a given Pair to exchange two tokens. Using this standard you would go to a given FlashLender (possibly built with a factory) to obtain a flash loan of a given token. You could have multiple receivers, or use the `data` parameter.

I don't dislike your suggestion, but I think there is value in simpler interfaces as long as we are not significantly limited by them.

lightclient (approved):

fubuloubu:
@albertocuestacanada @MickdeGraaf please migrate your conversation to https://ethereum-magicians.org/t/flash-loan-eip-erc-3156-early-draft/4993

PR #3162, opened 2020-12-14
Updates statuses to match EIP-1

MicahZoltu:
We forgot to update the statuses here, which causes Review items to not appear on the EIPs site.

I also re-ordered things in the *hope* that they show up on the page in this order, though I suspect they won't.


lightclient (approved):
I believe this will resolve our issue, but haven't had a chance to confirm locally.

PR #3163, opened 2020-12-14
Changes the max width of EIPs site to be something reasonable.

MicahZoltu:
Uses a highly specific CSS selector to ensure that it overrides the minima selector of `.wrapper` which sets the max-width to 740px.


MicahZoltu:
Merging since no one has said no over the last month.

PR #3170, opened 2020-12-22
Request peer review for ERC-681

nagydani:
ERC 681 has been implemented for a while by several wallet applications: WallETH, MyCrypto, Q-Wallet, MetaMask (partial), Status (partial) and possibly others. I believe that it is a de-facto standard and therefore it is time to finalize it as such.

### DIFF
--- a/EIPS/eip-681.md
+++ b/EIPS/eip-681.md
@@ -1,7 +1,7 @@
 ---
 eip: 681
 title: URL Format for Transaction Requests
-author: Daniel A. Nagy <daniel@ethereum.org>
+author: Daniel A. Nagy (@nagydani)
 type: Standards Track
 category: ERC
 status: Draft
--- a/EIPS/eip-681.md
+++ b/EIPS/eip-681.md
@@ -4,7 +4,7 @@ title: URL Format for Transaction Requests
 author: Daniel A. Nagy <nagydani@epointsystem.org>
 type: Standards Track
 category: ERC
-status: Last Call
+status: Draft
 created: 2017-08-01
 requires: 20, 137, 831
 ---
--- a/EIPS/eip-681.md
+++ b/EIPS/eip-681.md
@@ -80,7 +80,7 @@ In the fairly common case of only indicating the recipient address in a request 
 ## Security Considerations
 
 Since irreversible transactions can be initiated with parameters from such URLs, the integrity and authenticity of these URLs are of great importance.
-In particular, changing either the recipient address or the amount transfered can be a profitable attack. Users should only use URLs received from authenticated sources with adequate integrity protection.
+In particular, changing either the recipient address or the amount transferred can be a profitable attack. Users should only use URLs received from authenticated sources with adequate integrity protection.
 
 To prevent malicious redirection of payments using ENS, hexadecimal interpretation of Ethereum addresses must have precedence over ENS lookups. Client software may alert the user if an ENS address is visually similar to a hexadecimal address or even outright reject such addresses as likely phishing attacks.
 
--- a/EIPS/eip-681.md
+++ b/EIPS/eip-681.md
@@ -6,11 +6,11 @@ type: Standards Track
 category: ERC
 status: Review
 created: 2017-08-01
-requires: 20, 137, 831
+requires: 20, 137
 ---
 
 ## Simple Summary
-A standard way of representing various transactions, especially payment requests in Ethers and ERC #20 tokens as URLs.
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 681 state from Draft to Review

MicahZoltu (changes requested):
I believe `ether` should be used rather than `Ethers`.  I don't *think* pluralizing it is appropriate, nor do  *believe* it is correct for it to be capitalized as a proper noun.  I am curious to hear arguments against this though.

----

`## Backwards Compatibility` and `## Security Considerations` sections are required (after `## Rationale` section).

----

`Compatibility and Versioning` section is not a standard section and should be removed (consider moving its current contents over to the `Rationale` section describing the `erc831_part`).

----

EIP-831 is not currently final, which means this EIP cannot go to final.  You will either need to remove the dependency, or get that EIP get it final, or fork that EIP and get your forked EIP to final and then depend on that.

----

The first reference (at least) to an EIP within this document must be of the form `[EIP-123](./eip-123.md)`.  Also, all EIPs must be referred to as `EIP-XXX` or `ERC-XXX` not `EIP #XXX`.

----

Why does `ethereum_address` start with `0x`.  We already know the encoding from context (reading a URI) so it is just wasted characters that serve no purpose.

----

The UNIT part of this specification is problematic at best.  Consider just removing it.

The spec says that if there is a symbol field on the contract and a decimals field then the value should be multiplied by 10^decimals.  This means that if someone creates a URL to send `10e18 DAI`, a spec compliant implementation will attempt to send `10e36 DAI`, which is almost certainly not the intent.

The spec doesn't say what to do if there is no UNIT provided but contract access is available.  Should it still multiply the provided value by decimals?

It seems that the spec should fail if a non-empty `UNIT` is provided and blockchain access is not available, rather than succeed as the signer has no way of validating that the units provided are correct.  I could create a transfer URL for `1e18 microETH` and the signer would blindly render this, and then proceed to send 1e18 ETH instead.  Given non-verifiable information, tools should prefer to protect the user rather than put them in danger.

Why is this specification dictating what tooling MUST/MUST NOT do with the data (like requiring they fail in certain situations)?  This seems like something that should be left up to the tooling to decide what is most appropriate for their users.  For example, I may have an offline signer that has no blockchain access but **DOES** have a pre-seeded list of well known token details.

----

In general, **SHOULD** and **MAY** statements aren't great for specs since they don't add any value in standardization and are really just style guidelines that should be left for other resources like blog posts, tutorials, etc.

----

What happens if the user supplies a non-integer value like `1.2345e2`?

----

Security Considerations section should go into great detail about all of the potential attack vectors and the trust model required to use these URL.

nagydani:
@MicahZoltu thank you for the detailed review! I believe I have addressed all your concerns.

MicahZoltu (approved):

EIPS/eip-681.md
```diff
@@ -35,50 +35,57 @@ Payment request URLs contain "ethereum" in their schema (protocol) part and are
...
+Future upgrades that are partially or fully incompatible with this proposal must use a prefix other than `pay-` that is separated by a dash (`-`) character from whatever follows it.
+
+## Backwards Compatibility
+
+In the fairly common case of only indicating the recipient address in a request for payment in ether, this specification is compatible with the superseded ERC-67.
+
+## Security Considerations
+
+Since irreversible transactions can be initiated with parameters from such URLs, the integrity and authenticity of these URLs are of great importance.
+In particular, changing either the recipient address or the amount transfered can be a profitable attack. Users should only use URLs received from authenticated sources with adequate integrity protection.
```
MicahZoltu, line 83:
```suggestion
In particular, changing either the recipient address or the amount transferred can be a profitable attack. Users should only use URLs received from authenticated sources with adequate integrity protection.
```

MicahZoltu:
I fixed a typo that was blocking CI.  Once it passes I'll merge.

shannonlitton:
I need help... I need a tudor

DGKSK8LIFE:
@shannonlitton is a troll

MicahZoltu:
@DGKSK8LIFE Please refrain from responding to spam content like this.  If you want to help out, just report the content as spam to GitHub via the `...` menu on the right of the post.

shannonlitton:
I'm spam now...

shannonlitton:
but going this impromptu and blind

DGKSK8LIFE:
ok thanks @MicahZoltu

Issue #102, opened 2016-05-01
Precompiles for ECADD and ECMUL


vbuterin:
### Parameters
- `METROPOLIS_FORK_BLKNUM`: TBA
- `GECADD`: 200
- `GECMUL`: 1000
### Specification

Adds a precompile at address `0x0000....0a` which accepts as input 224 bytes that it interprets as `[x1, y1, x2, y2, P, A, B]` (if less than 224 bytes are provided the remaining bytes are assumed to be zero, if more than 224 bytes are provided they are ignored). Throws if `(x1, y1)` or `(x2, y2)` are not on the curve defined by `(P, A, B)` (ie. `x**3+A*x+B-y**2` is not a multiple of `P`), otherwise returns 64 bytes representing the elliptic curve addition `(x1, y1) + (x2, y2)` on that curve.

Adds a precompile at address `0x0000....0b` which accepts as input192 bytes that it interprets as `[x, y, n, P, A, B]` (if less than 192 bytes are provided the remaining bytes are assumed to be zero, if more than 192 bytes are provided they are ignored). Throws if `(x1, y1)` is not on the curve defined by `(P, A, B)` (ie. `x**3+A*x+B-y**2` is not a multiple of `P`), otherwise returns 64 bytes representing the elliptic curve multiplication `(x1, y1) * n` on that curve.
### Rationale

Allows for the efficient application of ring signatures and other cryptography based on elliptic curves. Currently, an secp256k1-focused implementation of these operations (see https://github.com/ethereum/serpent/blob/develop/examples/ecc/jacobian_arith.se ) takes ~200k gas for a multiply, ~750k gas for an ecrecover and ~900k gas per linkable ring signature participant, which is unacceptably high.

chfast:
This seems reasonable if such functions are needed.
1. I like the costs being constant. Where the numbers come from?
2. Are the functions included in the secp256k1 library?

vbuterin:
Updated to make it more generic (set your own curve parameters).

> 1. I like the costs being constant. Where the numbers come from?

Reverse engineering from the current ECRECOVER cost, ie. an ECRECOVER is 3 multiply ops plus some adds and modexp, so the cost of ECMUL should be close to a third of that of ECRECOVER.

>  2. Are the functions included in the secp256k1 library?

Not sure, especially with the curve generalization.

androlo:
[...]

PR #3171, opened 2020-12-22
Add CREATE3 opcode - no init_code hash

moodysalem:
Adds a CREATE3 opcode that does not include the init code hash, which is difficult to use on-chain to compute addresses in combination with constructor arguments.

The rationale for including the init code hash in the CREATE2 implementation is that it forces the address to contain specific bytecode. However it is not useful for cases where you only care that the contract was deployed by a specific factory account, with certain arguments (encoded in the salt), which is sufficient proof that the contract bytecode is as expected in many use cases. 

I.e., if you want to compute the address of a contract with constructor arguments created via CREATE2 on-chain, you have to include the entire bytecode of the contract for which you are computing the address (instead of just the hash) to get the preimage of the contract bytecode.


EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+
+The preimage for the final hashing round is thus always exactly `53` bytes long (does not include hash of the init code).
+
+### Motivation
+
+The core motivations are the same as [./eip-1014.md](EIP-1014), i.e. deploying a smart contract at a deterministic address.
+This variation removes the hash of the init code from the preimage. The issue with the existing `CREATE2` opcode
+is that computing the hash of the init code on-chain requires the entire bytecode of the target contract to be included
+in the contract computing the first preimage.
+
```
MicahZoltu, line 36: This section should come before the `# Specification` section.

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+eip: x
+title: CREATE3
+author: Moody Salem
+category: Core
+type: Standards Track
+status: Draft
+created: 2020-12-22
+---
+
+### Specification
```
MicahZoltu, line 11:
```suggestion
## Simple Summary
TBD

## Abstract
TBD

## Motivation
<moved up from below>

## Specification
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
+---
+eip: x
```
MicahZoltu, line 2:
```suggestion
eip: 3171
```
Make sure to also rename the file to `eip-3171.md`.

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+The preimage for the final hashing round is thus always exactly `53` bytes long (does not include hash of the init code).
+
+### Motivation
+
+The core motivations are the same as [./eip-1014.md](EIP-1014), i.e. deploying a smart contract at a deterministic address.
+This variation removes the hash of the init code from the preimage. The issue with the existing `CREATE2` opcode
+is that computing the hash of the init code on-chain requires the entire bytecode of the target contract to be included
+in the contract computing the first preimage.
+
+### Rationale
```
MicahZoltu, line 37: Top level sections should have `##` and subsections should have `###`.  This is a problem throughout this document.  See https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md for an example layout.

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+* The addresses created with this scheme cannot collide with addresses created using the `CREATE2`, as `0xfe` will never collide with `0xff`.
+* Ensures that the hash preimage has a fixed size.
+
+#### Gas cost
+
+The concerns for `CREATE2` gas cost in regard to init code preimage length do not apply.
+
+### Clarifications
+
+This EIP makes collisions possible. The behaviour at collisions is specified by [EIP-684](https://github.com/ethereum/EIPs/issues/684):
```
MicahZoltu, line 51: EIP-684 is not an EIP, just an idea.  I don't know what the current on-chain behavior is since 1014 is a valid final EIP and it references 684, but regardless, you should not reference non-final EIPs in an EIP (note: some older EIPs don't follow this rule as it is relatively new).

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+
+> If a contract creation is attempted, due to either a creation transaction or the `CREATE`, `CREATE2`, (or future `CREATE3`) opcode,
+> and the destination address already has either nonzero nonce, or nonempty code, then the creation throws immediately, with exactly 
+> the same behavior as would arise if the first byte in the init code were an invalid opcode. This applies retroactively starting from genesis.
+
+Specifically, if `nonce` or `code` is nonzero, then the create-operation fails.
```
MicahZoltu, line 57: You should verify that this is actually true in Ethereum.  Given that 684 never actually turned into an EIP I'm not confident it was.

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+
+* As with CREATE2, ensures that addresses created with this scheme cannot collide with addresses created using the traditional `keccak256(rlp([sender, nonce]))` formula, as `0xfe` can only be a starting byte for RLP for data many petabytes long.
+* The addresses created with this scheme cannot collide with addresses created using the `CREATE2`, as `0xfe` will never collide with `0xff`.
+* Ensures that the hash preimage has a fixed size.
+
+#### Gas cost
+
+The concerns for `CREATE2` gas cost in regard to init code preimage length do not apply.
+
+### Clarifications
```
MicahZoltu, line 49: `Clarification` isn't a valid EIP section.  Recommend moving this content into the specification or possibly `Backwards Compatibility` or `Security Considerations` sections.

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+> Account creation transactions and the `CREATE` operation SHALL, prior to the execution of the initialisation code,
+> increment the nonce over and above its normal starting value by one
+
+This means that if a contract is created in a transaction, the `nonce` is immediately non-zero, with the side-effect 
+that a collision within the same transaction will always fail -- even if it's carried out from the `init_code` itself.
+
+It should also be noted that `SELFDESTRUCT` (`0xff`) has no immediate effect on `nonce` or `code`, thus a contract cannot 
+be destroyed and recreated within one transaction.
+
+### Examples
```
MicahZoltu, line 70:
```suggestion
## Test Cases
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,113 @@
...
+* salt `0x00000000000000000000000000000000000000000000000000000000cafebabe`
+* gas (assuming no mem expansion): ~`32012`~
+* result: ~`0x1d8bfDC5D46DC4f61D6b6115972536eBE6A8854C`~
+
+Example 6
+* address `0x0000000000000000000000000000000000000000`
+* salt `0x0000000000000000000000000000000000000000000000000000000000000000`
+* gas (assuming no mem expansion): ~`32000`~
+* result: ~`0xE33C0C7F7df4809055C3ebA6c09CFe4BaF1BD9e0`~
+
```
MicahZoltu, line 113:
```suggestion

## Backwards Compatibility
TBD

## Security Considerations
TBD

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```

MicahZoltu (changes requested):

EIPS/eip-3171.md
```diff
@@ -0,0 +1,148 @@
...
+
+- `0xfe` is a single byte, 
+- `address` is always `20` bytes, 
+- `salt` is always `32` bytes (a stack item). 
+
+The preimage for the final hashing round is thus always exactly `53` bytes long (does not include hash of the init code).
+
+## Rationale
+
+### Address formula
```
lightclient, line 65: I'm don't think it is required to use `0xfe` instead of `0xff`. Every element of the `create2` and `create3` formula is fixed, so it's not trivial to create a collision. I recommend maintaining the `0xff` beginning over introducing a new one.

moodysalem: I agree, changed to `0xff` and the collisions are prevented by the difference in preimage length

EIPS/eip-3171.md
```diff
@@ -0,0 +1,148 @@
+---
+eip: 3171
+title: CREATE3
+author: Moody Salem <moody@uniswap.org>
+category: Core
+type: Standards Track
+status: Draft
+created: 2020-12-22
+---
+
```
shemnon, line 10: Where should discussions for this EIP go?

EIPS/eip-x.md
moodysalem: I removed this stuff and added a TODO

Worker36ruS (approved):

EIPS/eip-3171.md
```diff
@@ -0,0 +1,148 @@
...
+## Simple Summary
+
+This EIP is a fork of [EIP-1014](./eip-1014.md) (`CREATE2`) which adds an account creation opcode (`CREATE3`) that does
```
lightclient, line 13: I think this sentence makes more sense in the abstract, and the simple summary can just be sentence below.

EIPS/eip-3171.md
```diff
@@ -0,0 +1,148 @@
...
+not involve the init code hash. It is otherwise identical to `CREATE2` opcode.
+
+This opcode allows users to create accounts at an address independent of the bytecode deployed to the address.
```
lightclient, line 16:
```suggestion
`CREATE3 (0xf6)` allows users to create accounts at an address independent of the bytecode deployed to the address.
```

EIPS/eip-3171.md
```diff
@@ -0,0 +1,148 @@
...
+As in `CREATE2`, `CREATE3` makes collisions possible.
+
+`CREATE3` is not expected to have address collisions with `CREATE2`, as the preimage length differs by the `32` bytes
+of the init code hash.
+
+TODO: are there any existing protections in place that prevent contracts from self destructing and being recreated with `CREATE2`?
+
+## Security Considerations
+
+This new opcode allows the deterministic contract address to be independent of the bytecode. Users may not depend on the deterministic
```
lightclient, line 134: I'm not sure it's fair to say this opcode "allows" this behaviour as it's possible to deploy arbitrary contracts to deterministic addresses with `CREATE2`.

lightclient (changes requested):
I left a few recommendations on the eip. The only things blocking a merge are updating the author field with your github handle and creating a `discussions-to` field.

EIPS/eip-3171.md
```diff
@@ -0,0 +1,148 @@
...
+to save gas in computing the pair address from the Uniswap V2 Router. The gas cost of the `pairFor` method which
+computes the pair address for a given pair of tokens is only ~649 gas, while an external call + an SLOAD from mapping
+may cost upwards of 3k gas. A swap that is routed through multiple pairs requires multiple calls of this method.
+
+This is feasible only because the Uniswap V2 Pair does not have any constructor arguments, which means the init code hash
+can be hard coded in the library. However, the init code hash cannot be hard coded when the deployed smart contract has
+constructor arguments. In order to compute the init code hash, one must append the constructor arguments to the init code
+before computing the init code hash. Thus, in order to compute the address of an account created via `CREATE2`, it is 
+necessary to include the entire init code without the constructor arguments, and hash 
+`ceil(len(init_code_base + constructor_arguments) / 32)` words on-chain.
```
MrChico, line 41: It's not particularly pretty, but I believe another workaround to this problem is to have the init code pull the arguments from the creator:
```
interface F {
  function provideX() external returns (uint) {
  }

contract A {
   uint x;
   constructor() {
      x = F(msg.sender).provideX();
   }
}

contract Factory {
   uint x;
   function provideX() public returns (uint) {
      return x;
   }
   function deployA(uint initArg) public {
       x = initArg;
       new A();
       x = 0;
   }
}

Worker36ruS (approved):

EIPS/eip-3171.md
moodysalem: Thank you, I heard about this in the #allcoredevs channel of Discord and I will add it as a workaround

EIPS/eip-3171.md
lightclient: @moodysalem still missing a `discussions-to` section. Merging is blocked until this is resolved.

lightclient (changes requested):

moodysalem:
Since there is a great workaround I don't see a reason for this EIP any longer. Closing this PR so as not to reserve the name CREATE3 for only an incremental improvement over CREATE2. 

Would prefer a CREATE3 or other opcode that does something to improve the cost of deploying many contracts with identical init code and near identical runtime bytecode (without any runtime overhead like minimal proxy)

lightclient:
@moodysalem the work around being the delegatecall trampoline in the init code?

moodysalem:
> @moodysalem the work around being the delegatecall trampoline in the init code?

There's no delegatecall involved (it just does a staticcall to get the constructor arguments from another address, with 0 real constructor arguments)

lightclient:
@moodysalem ah okay, and that is fine because you're always deploying the same code - just with different constructor arguments?

Worker36ruS (approved):

moodysalem:
I believe there is still a strong reason for removing init code hash from the address generation, which is the controlling the same address on multiple networks but with different code. With layer 2s, different code may need to be deployed at the same address for each network.

MicahZoltu:
> I believe there is still a strong reason for removing init code hash from the address generation, which is the controlling the same address on multiple networks but with different code. With layer 2s, different code may need to be deployed at the same address for each network.

With some effort, you can have different deployed bytecode on each network without a new opcode.  You just need to fetch your deployment bytes (or parameters) from storage as part of the constructor, rather than deploying static bytes.  While the current tooling doesn't support this, I think changing the tooling is superior to changing the EVM when possible.  The gas cost of deploying like this is a bit higher, but usually deployments are cheap.

Separately, this PR still needs the adjustments mentioned above, plus the Simple Summary is no longer a section and it is replaced with a `description: ` frontmatter item.

moodysalem:
> I think changing the tooling is superior to changing the EVM when possible. The gas cost of deploying like this is a bit higher, but usually deployments are cheap.

The reason to add a new opcode is that the original rationale for init code hash being included is incorrect (guarantee that the code at that address is the same) because it can be circumvented as you describe.

I will work on updates to the PR to make it mergeable.

MicahZoltu:
> The reason to add a new opcode is that the original rationale for init code hash being included is incorrect (guarantee that the code at that address is the same) because it can be circumvented as you describe.

This seems like an argument for removing/fixing/altering CREATE2, rather than an argument for adding CREATE3?

MicahZoltu:
I just meant to say that the argument you made is essentially that "CREATE2 doesn't achieve its stated goals", but it wasn't an argument that "CREATE3 is necessary".

k06a:
CREATE3 is very useful for `immutable` variables which usually affect address of CREATE2

0xngmi:
Dealing with this right now and CREATE3 would really make a lot of things simpler, voicing support for it too :)

Kifzy:
Like.

zemse:
This may help: [solmate/CREATE3.sol](https://github.com/Rari-Capital/solmate/blob/eaaccf88ac5290299884437e1aee098a96583d54/src/utils/CREATE3.sol).

SKYBITDev3:
> This pull request was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.

Please reopen this.

xinbenlv:
Hi, thanks I love to follow the discussion about Create3.

Where do I follow the latest discussion? Is this pull request the canonical discussion place?

I saw it was put on AllCoreDevs #102 for a decision but checked the agenda and notes  
https://github.com/ethereum/pm/blob/master/AllCoreDevs-EL-Meetings/Meeting%20103.md but couldn't find a discussion on that meeting.

Lohann:
> CREATE3 is very useful for immutable variables which usually affect address of CREATE2

@k06a I create an solution that allow you to initialize immutables without `CREATE3`, let me know what do you think:

Essentially it is a **Create2Factory** contract, but with some advanced features:

- It automatically detect the EVM version `shangai` or `cancun`.
- For `cancun` it caches the constructor arguments in the [EIP-1153 Transient Storage](https://eips.ethereum.org/EIPS/eip-1153) the evm supports `cancun` (~100 gas per word).
- For `shanghai` it caches the arguments in the storage (~2900 gas per word).
- By retrieving the [Context](https://github.com/Lohann/universal-factory/blob/main/src/UniversalFactory.sol#L45-L60) from the factory, the contract knows who is the `msg.sender`, [example here](https://github.com/Lohann/universal-factory/blob/main/test/examples/Owned.sol#L16-L17).

Source code:.
https://github.com/Lohann/universal-factory/blob/main/src/UniversalFactory.sol

Lohann:
> CREATE3 is very useful for immutable variables which usually affect address of CREATE2

@k06a I create an solution that allow you to initialize immutables without `CREATE3`, let me know what do you think:

Essentially it is a **Create2Factory** contract, but with some advanced features:

- It automatically detect the EVM version `shangai` or `cancun`.
- For `cancun` it caches the constructor arguments in the [EIP-1153 Transient Storage](https://eips.ethereum.org/EIPS/eip-1153) (~100 gas per word).
- For `shanghai` it caches the arguments in the storage (~2900 gas per word).
- By retrieving the [Context](https://github.com/Lohann/universal-factory/blob/main/src/UniversalFactory.sol#L45-L60) from the factory, the contract knows who is the `msg.sender`, [example here](https://github.com/Lohann/universal-factory/blob/main/test/examples/Owned.sol#L16-L17).

Source code:.
https://github.com/Lohann/universal-factory/blob/main/src/UniversalFactory.sol

PR #3195, opened 2021-01-09
Update eip-1559.md

edocodi:
typo

 - The build passes.

### DIFF
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -201,7 +201,7 @@ class World(ABC):
 Legacy Ethereum transactions will still work and be included in blocks, but they will not benefit directly from the new pricing system.  This is due to the fact that upgrading from legacy transactions to new transactions results in the legacy transaction's `gas_price ` entirely being consumed either by the `base_fee` or the `miner_bribe`.
 
 ### GASPRICE
-Previous to this change, `GASPRICE` represented both the ETH paid by the signer per gas for a transaction as well as the ETH received by the miner per gas.  As of this change, `GASPRICE` now only represents the amount of ETH paid by the signerper gas, and the amount a miner was paid for the transaction is no longer accessible directly in the EVM.
+Previous to this change, `GASPRICE` represented both the ETH paid by the signer per gas for a transaction as well as the ETH received by the miner per gas.  As of this change, `GASPRICE` now only represents the amount of ETH paid by the signer per gas, and the amount a miner was paid for the transaction is no longer accessible directly in the EVM.
 
 ## Test Cases
 


eip-automerger:

MicahZoltu:
Exercising executive power to merge this without author approval since it is such a small non-normative change (just a typo).

Rufia75:
Who is going to mine and support the blockchain if you get rid of transaction fees? If everyone turns off their miners and only ASIC farms are left running? Who holds all the power then. 

This is really treating miners as 2nd-rate citizens.

SoftwareZeus:
Mining is important and transaction fees are necessary to allow the small, non-farm miner to exist because residential electricity rates are higher for the home miner. Mining is what gives a currency its value because it costs money to create one ETH. Think about the difference between water and diamonds. What gives diamonds their value? Water is useful and necessary for life yet one gallon of water is cheap because it is cheap to produce. One pound of diamonds is more valuable because it costs a lot more money to produce. Diamonds have very little use other than being pretty to look at and some minor industry use. If diamonds were easy to pick up off the ground for free, it would not have a value. Getting rid of miners means it costs very little in resources and money to create ETH in the same way that diamonds our gold would lose their value if it were easy to create these out of thin air for little to no cost. The long term consequence of getting rid of mining would be to cause ETH to lose its value. This is a bad proposal and should not be launched. We should keep the miners in business and keep mining decentralized by encouraging home users with GPUs to mine at home.

gilmardal:
Besides all the software stuff, I just want to say something. Hardware mining empower the poor people like me. Stake mining just keep a world with a lot of inequality as government does. Think about it!

evwitmer:
What sounds more decentralized and empowering to new Ethereum Users, having them purchase $38,400 worth of Ether or having them turn on the graphics card in their gaming PC to get their first Ether experience? Eth2.0 can still happen, but there needs to be some sort of layer 2 solution that uses POW GPU mining to keep the simple barriers to entry for the new enthusiasts coming into the space IMO. 

Regardless of what ETH2.0 brings, eip-1559 is just a quicker way of helping out big holders of the token. Burning transaction fees only helps people who already hold large amounts of ether and would look more appealing to institutions as they start to look at taking a stake. Remember where Blockchain came from and who has supported it the entirety of its life. It has always been the community and fanboys that scream the loudest and carry tokens through. 

Eth2.0 + eip-1559 = Good for Institutions
while
GPU Mining + Layer 2 = Good for Community and Decentralization

gmzombie:
Heres my take on eip 1559. First off i am a smalltime miner and i wouldnt have even gotten into crypto without learning how to mine for ethereum. As I have been in this for a few years now since 2017 i have seen all the arguements about why this eip needs to happen or the block reward dropping and i can understand alot of the arguements for and against those other eips and i can even understand why eip 1559 has been setup. But here is what i dont get and this simply this. Why crap on the miners who built this network and brought your coin from nothing to second in the world? This eip has some good in it but not enough to justify yet another gut punch to miners. I read that by doing this eip you risk seting up a problem where eth burn precluded the fixed supply. So now we would have a no limit coin? Seems to me that a crypto that can just continue to make coins sounds alot like fiat currency. I for one believe bitcoin is going so high due to its 21million coin only attitude among other things. I believe in ethereum as I always have but this eip is simply put a gut punch to the people who helped you get this coin off of the floor in the first place. I totally get that we will eventually get to POS, but why not allow the miners to get there with you.  My sincere hope is that you devs understand that taking out the tip we miners get for keeping this network running is not logical. Or if it has to be why not push the eth block reward back go 4 or 5 eth as a way of trying to make it right with us. Thanks for reading.

Lynchb7:
Small miner here, and I just want to say I don't think EIP-1559 is a bad thing for the network. However, in my opinion it must include enough of a miner fee to keep it profitable for small timers. Maybe rather than burning the majority of the fee you burn half and give the other half to miners. If it completely kills small timers all that will be left is the large farms and ASICs, this would greatly hurt decentralization and make the network more vulnerable to things like 50% attacks. I enjoy mining and supporting the network, but do not want to do so at a loss. Thank you for reading.

Alpejohn:
Hi, i just got into mining a few months ago and i really enjoying it. Wish i got into this alot sooner.
It makes me sad that this will possible end the profits for mining. 
IF miners where the people that got Ethereum to where it is now, its just fair that you let us continu to mine with profit atleast until ETH 2.0 fully launches. IMHO.

And, its not long until the mining stops completely anyways (POS). So why bother changing this at "the end"? Its been like this for years, if its like this for another year its not a big deal.. :-)

Pleace don't do this.
Thak you for reading.

Sensone1:
Think a little bit, really we who supported the network, and made this possible.. Iam speech less, don't know what to say.. Money changing people from a good idea of decentralised banking, to this and even pos i disagree with when it comes to decentralisation.. GGreasy greedy development.. Makes mi sad

AdamVlcan:
Hi guys, watching situation about ethereum for a while, and it makes me sad. Eth2.0, be my guest(except that I think it's pretty bad idea and I totally disagree with POS, which killing decentralization and looks more like central banking again, also take 32 ethereum to be validator [without pool] is pretty turning away from that great idea, that everyone can participate and support the net) but still... ok. 
But cut the transaction fees from miners? From that miners who "build" it up, because of them it's working? Come one. It's pretty nonsense and very bad treating. 
But whatever, imho... miners will turn back to you and find something else. But it is sad watch this awesome project end up like this :( 
How to become from decentralized idea for everyone to not-so-decentralized, community killing, commerce piece of crypto which was great one day. 
R.I.P and have a nice day.

blacknight2u:
Sadness, don’t donk this up man. That would be whatever is the opposite of pog.

gcoin44:
Decentralized into the hands of a few....

gcoin44:
"By burning the base fee, we can no longer guarantee a fixed token supply. This could result in economic instabality as the long term supply of ETH will no longer be constant over time. While a valid concern, it is difficult to quantify how much of an impact this will have. If more is burned on base fee than is generated in mining rewards then ETH will be deflationary and if more is generated in mining rewards than is burned then ETH will be inflationary. Since we cannot control user demand for block space, we cannot assert at the moment whether ETH will end up inflationary or deflationary, so this change causes the core developers to lose some control over Ethereum's long term monetary policy." https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md

ishcarr:
This is not Decentralized Crypto anymore, probably never has been. Only the ones on top of the PYRAMID get to make the decisions. Please ***STOP EIP-1559*** This will hurt so much people who invested in Mining hardware. Whomever is in agreeing with this is only in it for the money, they do not care about the little MOM and POP shops like me trying to make a few bucks mining ETH. DISGUSTING!!! STOP THIS EIP-1559 PLEASE!!!

MicahZoltu:
Please use the `discussions-to` URL (https://ethereum-magicians.org/t/eip-1559-fee-market-change-for-eth-1-0-chain/2783) to discuss this EIP.  Keep in mind, an EIP is just a technical document, many EIPs never make it into a major client's codebase, and those that do don't always end up being deployed to a network.

PR #3203, opened 2021-01-15
Erc3156 review

albertocuestacanada:
Updated the reference implementations [to actually run](https://github.com/albertocuestacanada/ERC3156), fixing some things along the way:

 - Added natspec to interfaces
 - Used `amount` instead of `value` throughout the code
 - Corrected argument type on `onFlashLoan`
 - Minor descriptive edits
 - Added rationale on choosing `pull` over `push` repayment behaviour
 - Expanded reference implementation of flash borrower
 - Bug fixing on reference implementations
 - Better comments on reference implementations

### DIFF
--- a/EIPS/eip-3156.md
+++ b/EIPS/eip-3156.md
@@ -33,22 +33,46 @@ A flash lending feature integrates two smart contracts using a callback pattern.
 
 A `lender` MUST implement the IERC3156FlashLender interface.
 ```
+pragma solidity ^0.7.0 || ^0.8.0;
+import "./IERC3156FlashBorrower.sol";
+import "./IERC20.sol";
+
+
 interface IERC3156FlashLender {
-    function maxFlashAmount(
-        IERC20 token
-    ) external view returns (uint256);
-    
-    function flashFee(
-        IERC20 token,
-        uint256 amount
-    ) external view returns (uint256);
-    
+
+    /**
+     * @dev Initiate a flash loan.
+     * @param receiver The receiver of the tokens in the loan, and the receiver of the callback.
+     * @param token The loan currency.
+     * @param value The amount of tokens lent.
+     * @param data Arbitrary data structure, intended to contain user-defined parameters.
+     */
     function flashLoan(
         IERC3156FlashBorrower receiver,
         IERC20 token,
-        uint256 amount,
+        uint256 value,
         bytes calldata data
     ) external;
+
+    /**
+     * @dev The fee to be charged for a given loan.
+     * @param token The loan currency.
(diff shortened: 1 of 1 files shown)


EIPS/eip-3156.md
```diff
@@ -33,16 +33,40 @@ A flash lending feature integrates two smart contracts using a callback pattern.
 
 A `lender` MUST implement the IERC3156FlashLender interface.
 ```
+pragma solidity ^0.7.0 || ^0.8.0;
+import "./IERC3156FlashBorrower.sol";
+import "./IERC20.sol";
```
MicahZoltu, line 38: This is a pretty major spec change.  Previously, this worked with *any* token type, the only requirement was that it was transferable and countable and referencable by address.  This change makes it so *only* ERC20 tokens are supported.  Is this intentional?

Also, I wonder if it would be worthwhile to enhance the specification to support stuff like ERC-1155 tokens as well?  Maybe address + extradata (byte array) and that extra data could be EIP-1155 token ID?

Another option would be to rename the functions to be like `flashLoanErc20(...)` and `flashFeeErc20(...)`.  Then another Flash Loan EIP could be created for EIP-1155 tokens that is very similar to this one and they are compatible so something could support either or both.

albertocuestacanada: I didn't think that by specifying IERC20 I would be barring other tokens from being lent, I'll revert that.

Support for ERC1155 tokens can be added as an extension to this ERC, same as we are doing for batch flash loans.

MicahZoltu:
Just wanted to check, did you see this comment in the previous PR?  I didn't see you respond to it nor change it in this EIP:
> Consider adding support for flash lending ETH by making `token == SENTINEL` mean ETH in all functions.

albertocuestacanada:
Forgot to comment about that. I don't see an issue reserving `address(1)` as a sentinel for ETH in this context, but how would you go about it?

I don't think I can add it to the interfaces, and adding it to the reference implementation would complicate things, apart from being just a suggestion.

Should it be just be an statement ("If Ether is supported address(1) MUST be used to denote it")?

albertocuestacanada:
Also, a [bunch of good comments](https://ethereum-magicians.org/t/erc-3156-flash-loans-review-discussion/5077/12) from BoringCrypto in the discussions link that I also need to address.

MicahZoltu:
I don't know why this isn't auto-merging.  If it isn't auto-merged by the bot by tomorrow, feel free to mention me to manually merge it.

MicahZoltu:
One more try to trigger the auto-merge bot before manually merging.

PR #3204, opened 2021-01-17
1559: Changes starting base fee to 1 nanoeth and floors base fee at 0.

MicahZoltu:
Original spec had the starting base fee (at fork block) as 1 nanoeth, but this was unintentionally lost along the way.  I believe current clients have implemented it the original way so this change shouldn't require any client changes.

Also fixed a bug in the spec that allowed the base fee to go negative by just flooring it to 0.


eip-automerger:

MicahZoltu:
Closing in favor of #3212

PR #3209, opened 2021-01-20
erc3156: added return types to avoid vulnerability

albertocuestacanada:
BoringCrypto from SushiSwap revealed a vulnerability in the standard, that we are solving but requiring that `onFlashLoan` returns `true`, and requiring that the `lender` verifies that `true` was returned.

Aave v2 [already does this](https://github.com/aave/protocol-v2/blob/29448c19c1e9b8aa87dccd70b716d15cdf4128f6/contracts/protocol/lendingpool/LendingPool.sol#L507), thankfully.

### DIFF
--- a/EIPS/eip-3156.md
+++ b/EIPS/eip-3156.md
@@ -153,7 +153,7 @@ The `lender` MUST verify that the `onBatchFlashLoan` callback returns `true`.
 
 After the callback, for each `token` in `tokens`, the `batchFlashLoan` function MUST take the `amounts[i] + fees[i]` of `tokens[i]` from the `receiver`, or revert if this is not successful.
 
-If successful, `batchFlashLoan` MUST return `true`.
+If successful, `batchFlashLoan` MUST return the keccak256 hash of "ERC3156FlashBorrower.onFlashLoan".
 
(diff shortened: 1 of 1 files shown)


MicahZoltu:
What is the vulnerability?

MicahZoltu:
I read over the addition in the security section, but I'm not following.  If you are in Ethereum R&D discord, feel free to mention me if you want to discuss in a more synchronous way!

EIPS/eip-3156.md
```diff
@@ -87,9 +87,12 @@ function flashLoan(
     IERC20 token,
     uint256 amount,
     bytes calldata data
-) external {
+) external returns (bool) {
```
MicahZoltu, line 90: All returned values except `0` will map to `true`.  Recommend instead requiring a sentinel `uint256` or `bytes32` value be returned like `keccak256('3156')` or `0xdeadbeef...` to ensure that a fallback function that just returns "success" or any other positive value won't accidentally overlap with the return value we expect to ensure that the loan was *actually* understood/used.

albertocuestacanada: After some consideration of possible sentinel values, I think that using `keccak256("ERC3156FlashBorrower.onFlashLoan")` is good enough.

PR #3213, opened 2021-01-22
Removes Resources section.

MicahZoltu:
This EIP was originally written quite a while ago, this change just brings it in line with modern EIP editing standards.

EIPs are technical documents, not repositories for general information about a particular change.  Generally speaking, people should not be linking to EIPs directly when talking with the general public.  The time you would link a human to an EIP is when you want to discuss the technical aspects of it or implementation details.  In most other cases (such as governance discussions), a higher level document should be linked to and the EIP should just be a footnote in the references, meant only for implementors and people interested in understanding the EIP at a technical level.

@timbeiko has created this Hack MD document (https://hackmd.io/@timbeiko/1559-resources) that contains a much more maintained collection of resources related to EIP-1559 and generally should be what we are linking people to when we want to give something other than a pure technical document.


eip-automerger:

abdelhamidbakhta (approved):
Looks good to me.

PR #3224, opened 2021-01-27
Adding EIP for described data

ricmoo:
Adding initial draft for described data to allocate an EIP so a discussion-to URL can be added in the issues of the EIP repo.

**Abstract**

Human-readable descriptions for machine executable operations,
described in higher level machine readable data, so that wallets
can provide meaningful feedback to the user describing the
action the user is about to perform.


EIPS/eip-3224.md
```diff
@@ -0,0 +1,465 @@
...
+as transaction descriptions can be audited at the same time as
+contract code, allowing auditors and code reviewers to check that
+transaction descriptions are accurate as part of their review.
+
+
+## Specification
+
+The **description string** and **described data** are generated
+simultaneously by evaluating **describer** bytecode, passed
+**describedInput** as the calldata with the encoded method:
```
MicahZoltu, line 57: Most of these names don't align with parameter names in the Solidity below, which makes it really difficult to tell what is being talked about.  Recommend following a consistent naming scheme throughout the document.

ricmoo: You mean I should rename the signature to:

```
function eipXXXDescribe(bytes describer_input) view returns (string description_string, bytes described_data);
```

?

I can make that change now.

EIPS/eip-3224.md
```diff
@@ -0,0 +1,465 @@
...
+
+When executing the bytecode, best efforts should be made to
+ensure `BLOCKHASH`, `NUMBER`, `TIMESTAMP` and `DIFFICULTY`
+match the `"latest"` block. The `COINBASE` should be the zero
+address.
+
+If evaluating the function results in a revert, and the return
+data is of a length that is congruent to `4 mod 32`, and the
+first 4 bytes represent the selector `Error(string)`, the
+reason string should be displayed to the user. If evaluation
```
MicahZoltu, line 81: Displayed where?  Also, in general `should` is a red flag in a standard as anything that is a `should` or `may` cannot be enforced for compatibility.  This is especially true in this case where I *think* you are trying to tell user interfaces what they should do, which is more appropriate for an opinion piece like a blog post, tutorial, best practices document, patterns book, etc.

EIPS/eip-3224.md
```diff
@@ -0,0 +1,465 @@
...
+During evaluation, the `ADDRESS` (i.e. `to`), `CALLER`
+(i.e. `from`), `VALUE`, and `GASPRICE` must be the same as the
+values for the transaction being described, so that the
+code generating the description can rely on them. For signing
+**described messages**, `VALUE` should always be 0.
+
+When executing the bytecode, best efforts should be made to
+ensure `BLOCKHASH`, `NUMBER`, `TIMESTAMP` and `DIFFICULTY`
+match the `"latest"` block. The `COINBASE` should be the zero
+address.
```
MicahZoltu, line 76: Maybe I'm misunderstanding this specification (very possible), but it seems like none of this needs to be part of the standard.  The specification I believe can be simplified to just an ERC that defines an `eipXXXDescribe` function which takes in a byte array and returns a string (and maybe that data parameter, it is not clear to me what exactly that does or is for yet).

EIPS/eip-3224.md
```diff
@@ -0,0 +1,465 @@
...
+Clients which manage private keys should expose additional
+methods for interacting with the related accounts.
+
+If an user interface is not present or expected for any other
+account-based operations, the description strings should be
+ignored and the described data used directly.
+
+These JSON-RPC methods will also be implemented in standard
+Ethereum libraries, so the JSON-RPC description is meant more
+of a canonical way to describe them.
```
MicahZoltu, line 97: Above you are specifying a contract interface (ERC) and here you are specifying new JSON-RPC methods (Interface).  This should be split into two EIPs, where perhaps the JSON-RPC one has a dependency on the contract interface one.

EIPS/eip-3224.md
```diff
@@ -0,0 +1,465 @@
...
+`eipXXXDescription(bytes)`. The `VALUE` during execution must
+be 0.
+
+If the wallet contains a user interface for accepting or
+denying signing a message, it should present the description
+string to the user. Optionally, a wallet may wish to
+additionally provide a way to examine the described data.
+
+If accepted, the computed **described data** is signed
+according to [EIP-191](./eip-191.md), with the *version
```
MicahZoltu, line 123: Note: EIP-191 is still just a draft, so this EIP will get blocked by that unless you can move it forward.

EIPS/eip-3224.md
ricmoo: It requires returning a string and a bytes. The bytes passed in are a higher level description of the bytes being asked to be generated. A good example of this is the ENS commit example.

The input data describes what is wanted, the string is shown to the user (e.g. in metamask) and the output data is the data to use in a transaction.

EIPS/eip-3224.md
```diff
@@ -0,0 +1,465 @@
...
+**describerInput** passed to the ABI encoded call to
+`eipXXXDescription(bytes)`. The `VALUE` during execution must
+be 0.
+
+If the wallet contains a user interface for accepting or
+denying signing a message, it should present the description
+string to the user. Optionally, a wallet may wish to
+additionally provide a way to examine the described data.
+
+If accepted, the computed **described data** is signed
```
MicahZoltu, line 122: Why does the described data have to be signed?  What value does that add?  Who are you trying to prove this to that cannot just run the describer again?

ricmoo: As an example, OpenSea today has you sign a hash, which is submitted to a contract. That hash would not change, and would still be what is required to be signed. What does change is what the UI shows you; instead of "sign this hash: 0xsometing", it would show "Do you agree to place item 1234 for sale at a price of 1.23 ether". Upon accepting this, the wallet would continue signing the data, submitting it to OpenSea's servers, at this point nothing much changes.

EIPS/eip-3224.md
ricmoo: Interesting... @Arachnid, is there any reason not to move this forward?

EIPS/eip-3224.md
ricmoo: I guess the reason for the *should* is that not all UI's are interactive or have a display.

EIPS/eip-3224.md
ricmoo: I've updated this section.

MicahZoltu: Yes, though I still don't know what `describer` is.

ricmoo: The describer is just the contract itself. I can just change the name to contract? Originally this was more abstract, but we decided it made more sense to reduce the scope to a contract doing this itself (as a delegatecall can be used to outsource to another contract rather than returning a separated address to operate as the describer).

EIPS/eip-3224.md
```diff
@@ -0,0 +1,466 @@
...
+(i.e. the **describer**), passing the **describer inputs** to the 
+method:
+
+```solidity
+function eipXXXDescribe(bytes describer_inputs) view returns (string description_string, bytes described_data);
+```
+
+The method MUST be executable in a static context, i.e. any
+side effects (logX, sstore, etc., (including through indirect
+calls) will be ignored.
```
MicahZoltu, line 66: You have two open parenthesis and one close parenthesis.

EIPS/eip-3224.md
```diff
@@ -0,0 +1,466 @@
...
+
+During evaluation, the `ADDRESS` (i.e. `to`), `CALLER`
+(i.e. `from`), `VALUE`, and `GASPRICE` must be the same as the
+values for the transaction being described, so that the
+code generating the description can rely on them. For signing
+**described messages**, `VALUE` should always be 0.
```
MicahZoltu, line 72: Consider splitting out `describeTransaction` and `describeMessage` into two different methods.

If you do that, you can split them into two different EIPs.  This would allow a contract to implement one or the other, which can improve discoverability and make it easier to implement.

EIPS/eip-3224.md
```diff
@@ -0,0 +1,466 @@
...
+
+When executing the bytecode, best efforts should be made to
+ensure `BLOCKHASH`, `NUMBER`, `TIMESTAMP` and `DIFFICULTY`
+match the `"latest"` block. The `COINBASE` should be the zero
```
MicahZoltu, line 76: Consider instead asserting that these **MUST NOT** be referenced in the `describe` function.  Alternatively, you could just say they **SHOULD NOT** be referenced and then mention in the rationale section that while in some obscure cases they may be necessary, care should be taken not to assume that signers will be signing against the most recent block (since that is effectively impossible due to the nature of distributed systems).

EIPS/eip-3224.md
MicahZoltu: This paragraph feels out of scope for an ERC.  UIs can do whatever they want once they have completed the interaction with the contract.  I do agree that what you have described is a Good Idea, but EIPs are for standardization, not Good Ideas.  😄

EIPS/eip-3224.md
```diff
@@ -0,0 +1,466 @@
...
+- While Solidity is not awesome at string manipulation, there is nothing precluding a new compiler which simplifies string operations that compiles to bytecode or more advanced string libraries.
+- For popular legacy contracts, bytecode can be baked into wallets, or deployed on chain with the describer address baked in
+- When signing described data, the `to` address must be entangled in the signature to prevent replaying. In the case of transactions, this is implicitly handled by the `to` in the transaction. In messages, this is handled by including the to address in the EIP-191 *version specific data*. This is similar to the domain in [EIP-712](./eip-712.md), but allows for the dapp developer to define their own replay strategy (if any is desired)
+- Localization is quite important; there should be a way to encode this... This can be a separate EIP
+- Should a simple Markdown be supported? Keep in mind chain data must then be markdown-escaped.
+
+**Desired Features:**
+
+- Simple
+- re-invent as few wheels as possible
```
MicahZoltu, line 466:
```suggestion
```

EIPs should be written toward their final form, and this section isn't a valid EIP section so should be removed.  If you want, you can mark this PR as a review so people don't start reading it, and then you can delete this before submitting to be merged as a draft.  Alternatively, consider moving this content to something like HackMD or the discussions-to link.

MicahZoltu (changes requested):
Only real blocker is the footer, which should be deleted.  If you aren't ready to merge this as a draft you can mark the PR as a draft PR and then editors won't look at it.

The rest of the feedback I encourage reviewing.  We also discussed in Discord a lot of it and I think we *generally* are in agreement for most things, but that is largely unrelated to my editorial duties.

EIPS/eip-3224.md
```diff
@@ -0,0 +1,466 @@
...
+and the matching data can be computed (i.e. `commit(hash(name, owner, secret))`).
+
+### Alternatives
+
+- [NatSpec](https://docs.soliditylang.org/en/latest/natspec-format.html) and company; these languages are usually quite large and complex, requiring entire JavaScript VMs with ample processing power and memory, as well as additional sandboxing to reduce security concerns. One goal of this is to reduce the complexity to something that could execute on hardware wallets and other simple wallets.
+
+- Custom Languages; due to the complexity of Ethereum transactions, any language used would require a lot of expressiveness and re-inventing the wheel. The EVM already exists (it may not be ideal), but it is there and can handle everything necessary. 
+
+- Format Strings (e.g. [Trustless Signing UI Protocol](https://github.com/ethereum/EIPs/issues/719)); format strings can only operate on the class of regular languages, which in many cases is insufficient to describe an Ethereum transaction. This was an issue quite often during [early attempts](https://github.com/ricmoo/distinguished-format-strings) at solving this problem
```
MicahZoltu, line 247: Generally, we should try to avoid linking to external resources as the links will almost certainly break with time and editors do not have the manpower to maintain them.  It is acceptable to link to published (non-draft) EIPs, but for this section I recommend leaving out the links and specific products and instead just describing the alternative solutions briefly.

MicahZoltu: (same goes for the NatSpec link above)

ricmoo:
Thanks @MicahZoltu! I’ll get these updates made today. :)

EIPS/eip-3224.md
ricmoo: I've removed this and simple mentioned that the method may revert and that if so, the signing must be aborted.

EIPS/eip-3224.md
ricmoo: I will bring up splitting this EIP up on the discussion-to link, to help figure out exactly where the lines should be drawn. :)

EIPS/eip-3224.md
ricmoo: And I will include a reference to EIP-165 for discovery.

EIPS/eip-3224.md
ricmoo: I think there are common (may not common, but not uncommon) case where they are needed, which I'll flesh out in the rationale. Even something as simple as CryptoKitties could use it for the birth alarm clock, although it may make sense to discourage alarm clocks...

ricmoo:
It took longer to get to than expected. But I've addressed the reviews above.

I'll also start two discussions on the discussion-to link; localization and splitting the EIP up. @Arachnid and I have been having some conversations regarding this, so it would be good to get them out for others to comment on.

EIPS/eip-3224.md
ricmoo: I think for now this is fine, but will need to be addressed later. But we certainly don't want to define some new standard that's not already in the works for something like this. EIP-191 is already widely used.

ricmoo:
I've resolved all the conversations. Let me know if there is something else that is needed.

Thanks! :)

EIPS/eip-3224.md
```diff
@@ -0,0 +1,447 @@
...
+
+- HTML could be enbedded to attempt to trick web-based wallets into [executing code](https://en.wikipedia.org/wiki/Code_injection) using the script tag (possibly uploading any private keys to a server)
+- In general, extreme care must be used when rendering HTML; consider the ENS names `<span style="display:none">not-</span>ricmoo.eth` or `&thinsp;ricmoo.eth`, which if rendered without care would appear as `ricmoo.eth`, which it is not
+- Other marks which require escaping could be included, such as quotes (`"`), formatting (`\n` (new line), `\f` (form feed), `\t` (tab), any of many [non-standard whitespaces](https://en.wikipedia.org/wiki/Whitespace_character#Spaces_in_Unicode)), back-slassh (`\`)
+- UTF-8 has had bugs in the past which could allow arbitrary code execution and [crashing renderers](https://osxdaily.com/2015/05/27/bug-crashes-messages-app-ios-workaround/); consider using the UTF-8 replacement character (or *something*) for code-points outside common planes or common sub-sets within planes
+- [Homoglyphs attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack)
+- [Right-to-left](https://en.wikipedia.org/wiki/Right-to-left_mark) mark may affect rendering
```
MicahZoltu, line 410: External links should be avoided whenever possible, as they tend to 404 over time and we don't have the manpower to maintain dead links.  I think most of these are easily searchable, so no need to link to them (naming them is fine).

MicahZoltu (approved):

Chucky-ux:
I know nothing how it works the coupling coins back to single coins.
Thanks for helping  me with all this issues!

Gesendet von Yahoo Mail auf Android 
 
  Am Mi., Feb. 17, 2021 at 11:34 schrieb Micah Zoltu<notifications@github.com>:   


Merged #3224 into master.

PR #3228, opened 2021-01-28
EIP-2481: Testcases + proposed change to `GetBlockHeaders`

holiman:
This PR adds testcase for each of the messages that were redefined as part of `eth/66`. 
Since at the time of creating this PR, there are no clients that have implemented these messages, it would be good if the rlp-encoding of the mssages could be properly defined in the specification, and possibly also that others can verify these rlp blobs. 

~~This PR also proposes a change to the RLP-format of `GetBlockHeaders`.~~

~~Before this change:~~ 
EDIT: This is now the 'correct' way, unchanged from the original spec: 
```
rlpdump -hex 0xca820457c682270f050580

[
  0457,
  [
    270f,
    05,
    05,
    "",
  ],
]

```

~~After this change:~~
EDIT: The example below is moot: 

```
rlpdump -hex c982045782270f050580
[
  0457,
  270f,
  05,
  05,
  "",
]

```


eip-automerger:

 - EIP 2481 requires approval from one of (christoph@ethereum.org)

holiman:
- I removed the reference to `les`, 
- Testcases regenerated
- Removed the example from the abstract (the actual spec is just a little bit further down, seems unnecessary to duplicate things into the abstract)

cburgdorf (approved):
Thanks for making these changes. LGTM :+1:

PR #3234, opened 2021-01-29
Initial draft for ERC3234 batch flash loans

albertocuestacanada:
I've segregated the batch flash loan functionality from ERC3156, to allow lenders to support single-asset, multiple asset, or both kinds of flash loans.

I've taken the liberty to assign the 3157 number to the Batch Flash Loan ERC, as I think everyone will find it easier to deal with two very similar ERCs if they have consecutive numbers. Apologies if this can't be done.

### DIFF
--- a/EIPS/eip-3157.md
+++ b/EIPS/eip-3157.md
@@ -1,0 +1,265 @@
+---
+eip: 3157
+title: Batch Flash Loans
+author: Alberto Cuesta Cañada (@albertocuestacanada), Fiona Kobayashi (@fifikobayashi), fubuloubu (@fubuloubu), Austin Williams (@onewayfunction)
+discussions-to: https://ethereum-magicians.org/t/erc-3157-candidate-batch-flash-loans/5260
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-01-31
+---
+
+## Simple Summary
+
+This ERC provides standard interfaces and processes for flash lenders and borrowers, allowing for flash loan integration without a need to consider each particular implementation. The flash loans covered by this ERC can be composed of multiple assets at once, as opposed to ERC3156 where a single asset is included in each loan.
+
+## Motivation
+
+Flash loans of multiple assets, or batch flash loans, are a common offering of flash lenders, and have a strong use case in the simultaneous refinance of several positions between platforms. At the same time, batch flash loans are more complicated to use than single asset flash loans (ER3156). This divergence of use cases and user profiles calls for independent, but consistent, standards for single asset flash loans and batch flash loans.
+
+
+## Specification
+
+A batch flash lending feature integrates two smart contracts using a callback pattern. These are called the LENDER and the RECEIVER in this EIP.
+
+### Lender Specification
+
+A `lender` MUST implement the IERC3157BatchFlashLender interface.
+```
+pragma solidity ^0.7.0 || ^0.8.0;
+import "./IERC3157BatchFlashBorrower.sol";
+
+
+interface IERC3157BatchFlashLender {
+
+    /**
+     * @dev The amount of currency available to be lended.
+     * @param tokens The currency for each loan in the batch.
+     * @return The maximum amount that can be borrowed for each loan in the batch.
+     */
+    function maxFlashAmount(
+        address[] calldata tokens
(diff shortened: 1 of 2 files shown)


EIPS/eip-3157.md
```diff
@@ -0,0 +1,265 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-3157-candidate-batch-flash-loans/5260
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-01-31
+---
+
+## Simple Summary
+
+This ERC provides standard interfaces and processes for flash lenders and borrowers, allowing for flash loan integration without a need to consider each particular implementation. The flash loans covered by this ERC can be composed of multiple assets at once, as opposed to ERC3156 where a single asset is included in each loan.
```
MicahZoltu, line 14: This is a bit long for a simple summary.  Think of a GitHub PR title, email subject line, forum thread title, etc.  Consider moving some or all of this to the abstract and having the simple summary just be a nice one-line statement.

EIPS/eip-3157.md
```diff
@@ -0,0 +1,265 @@
+---
+eip: 3157
```
MicahZoltu, line 2: I'm generally pretty against cherry picking EIP numbers.  The reason for this is that it is a slippery slope toward people doing things like spamming the repository to get the issue number they want (a problem we have had in the past) and it is easier to just say that you should always just use whatever number the editors assign (which is usually the PR number).

That being said, we have allowed some cherry picking in the past and other editors have expressed less concern than me.

EIPS/eip-3157.md
```diff
@@ -0,0 +1,265 @@
...
+     * @dev Receive a flash loan.
+     * @param sender The initiator of the loan.
+     * @param token The loan currency.
+     * @param amount The amount of tokens lent.
+     * @param fee The additional amount of tokens to repay.
+     * @param data Arbitrary data structure, intended to contain user-defined parameters.
+     * @return The keccak256 hash of "ERC3157BatchFlashBorrower.onFlashLoan"
+     */
+    function onFlashLoan(
+        address sender,
```
MicahZoltu, line 129: Consider renaming this parameter to `initiator` to add a bit more clarity.  This address doesn't actually "send" anything.

EIPS/eip-3157.md
```diff
@@ -0,0 +1,265 @@
...
+TBA
+
+### Flash Mint Reference Implementation
+
+TBA - Flash mints make less sense as part of a batch flash loan
+
+### Flash Loan Reference Implementation
+
+TBA
+
```
MicahZoltu, line 206: This section isn't required, so you can just leave it out for now and add it later if you end up writing a reference implementation.

EIPS/eip-3157.md
```diff
@@ -0,0 +1,265 @@
...
+## Security Considerations
+
+_Copy from ERC3156, or reference it?_
```
MicahZoltu, line 209: Recommend copying.  Each EIP should stand on its own as much as possible unless it has a hard dependency.

MicahZoltu (approved):
Tentatively approving this as my only potentially blocking feedback is the EIP number, since I am generally against EIP number cherry picking.  I'll wait for other editors to chime in, or the author to change to EIP-3234 before merging though.

EIPS/eip-3157.md
lightclient: I would prefer the number be updated to 3234.

EIPS/eip-3157.md
albertocuestacanada: Fixed, will do the same for ERC3156.

EIPS/eip-3157.md
albertocuestacanada: Fixed

EIPS/eip-3157.md
albertocuestacanada: Fixed, will do the same for ERC3156.

EIPS/eip-3157.md
albertocuestacanada: Perfect.

EIPS/eip-3157.md
albertocuestacanada: Copied.

Issue #3086, opened 2020-11-02
EIP-3085 Discussion: wallet_addEthereumChain


rekmarks:
Discussion for EIP-3085: `wallet_addEthereumChain`

The EIP was briefly known as EIP-2016.

Initial PR: https://github.com/ethereum/EIPs/pull/3085

wighawag:
Great to see some EIP materializing to help apps make it easy for user to use other chains that what wallet provider offer by default.

I got few comments:

A general one: in my opinion we should avoid as much as possible the use of `should` or `may` and make the standard more strict

Regarding `The wallet application may arbitrarily refuse or accept the request.` 
This should be accompanied with a an error code, so it should not be completely arbitrary in so far as a error code should be provided, to ensure the apps can explain its users the reason. As such the EIP should define the set of error code allowed for rejection. 
We obviously cannot prevent the possibility of a unexpected error, but my point is that if we expect wallet to reject for different reason, the standard should aims to define these.

Regarding `all parameters except `chainId` are listed as optional, even though a wallet may require them in practice.`
I am not sure it is a great idea to have such ambiguous standard. Why not explicitly define which parameter are optional and which one are mandatory
If we think this is not possible, we should at least define a set of error code to explain to the apps the reason of rejection.


Regarding 
```
If a chain with the provided parameters was already added, the request is considered successful, and the method **MUST** return the boolean value `true`.`

The wallet **MAY** define what it means for a chain to be "already added", but care should be taken to protect user security.
The wallet **SHOULD NOT** allow the same `chainId` to be added multiple times.
```

As mentioned as general comment, we should not use MAY or SHOULD

Also regarding `The wallet **SHOULD NOT** allow the same `chainId` to be added multiple times.`

what happen in the following scenario

- App A request to add chain 6 with their own url endpoint
- time passes and App A url endpoint is shutdown 
- user go to App B that also want to connect chain 6, it request to add chain 6 but this time with a different working endpoint url

what about chain 1 (that is expected to be supported by most wallet)?

Some more consideration: 

Could the standard not add a way for the app to check if the wallet already support a chain ?
Or is it a privacy concern ?


Regarding security, I don't think this standard by being just a "request to add" is avoiding the security consideration mentioned in EIP-2015 (see https://ethereum-magicians.org/t/eip-2015-wallet-update-chain-json-rpc-method-wallet-updatechain/3274/13)

[...]

PR #3237, opened 2021-02-01
3085: Add "Chain Already Added" Error

rekmarks:
Per recent discussion in #3086, this PR adds a "Chain Already Added" error to EIP-3085. This error is to be returned by the method if the implementer (i.e. wallet) rejects the request because a chain with the provided chain ID was already added. _This is a breaking change in behavior_, since previously the method would return `null` in cases where the chain was already added.

Note that the error follows the precedent set by EIP-1193 / EIP-2696. Therefore, EIP-2696 (the relevant subset of 1193) is added as a dependency.

The PR also includes some minor cleanup / editing.

### DIFF
--- a/EIPS/eip-3085.md
+++ b/EIPS/eip-3085.md
@@ -7,7 +7,7 @@ status: Draft
 type: Standards Track
 category: Interface
 created: 2020-11-01
-requires: 155, 695, 3091
+requires: 155, 695, 2696, 3091
 ---
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


EIPS/eip-3085.md
```diff
@@ -7,7 +7,7 @@ status: Draft
 type: Standards Track
 category: Interface
 created: 2020-11-01
-requires: 155, 695, 3091
+requires: 155, 695, 2696, 3091
```
MicahZoltu, line 10: While not changed in this PR, be aware that 3091 is still draft and thus will prevent this EIP from moving forward.  You may want to reach out to the author of that EIP and see what their plans are for getting that moved to review and eventually final.

EIPS/eip-3085.md
```diff
@@ -37,13 +37,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 ### `wallet_addEthereumChain`
 
 The method accepts a single object parameter, with a `chainId` and some chain metadata.
-The method returns `null` if the chain was added to the wallet, and an error otherwise.
+The method returns `null` if the requested chain was added to the wallet as a result of the method call, and an error otherwise.
```
MicahZoltu, line 40: I feel like this should be a success if the chain was previously added with the same parameters.  As a caller, what I care about is that the chain is present.  I don't really care whether this was the first time it was added or if it was added in a previous call.

There are some security issues with a malicious dapp adding chain 123 and then an honest dapp adding chain 123 but with otherwise different configuration, so there may be value in returning an error if the two chains are sufficiently different.  On the other hand, the user may be using their own custom block explorer, or their own JSON-RPC endpoint which would cause the chain added to not exact match.

In other words, this is all complicated, but I feel like an error is not the correct result in the case of a duplicate chain.

MicahZoltu: To put it another way: This error is almost certainly going to be expressed as an exception in the vast majority of libraries, and exceptions should not be used for flow control, and this "error" will occur every time after the first time a user interacts with a particular chain.

MicahZoltu: One potential solution would be to provide a mechanism for first asking if a particular chain is available.  This would allow the app developer to make this request only when it is actually needed, rather than requiring the developer to use exception handling for flow control.

ytbigvinny27 (commented):
~

PR #3240, opened 2021-02-01
Erc3156: Rename `maxFlashAmount` as `maxFlashLoan`

albertocuestacanada:
Shorter, and easier to remember.


albertocuestacanada:
Will branch from the right spot.

PR #3243, opened 2021-02-01
ERC-3156: Move to Last Call

albertocuestacanada:
After two months of intense developement and consultation, I'd like to move ERC-3156 to Last Call.

In its development we have discussed it with members of Aave, Uniswap, Sushiswap, Yearn, Yield, DefiSaver, OpenZeppelin, and StakeDAO. It's been discussed on Ethereum Magicians, Twitter, ETHSecurity and FlashF*ck Researchers. The feedback has been very positive.

### DIFF
--- a/EIPS/eip-3156.md
+++ b/EIPS/eip-3156.md
@@ -3,7 +3,8 @@ eip: 3156
 title: Flash Loans
 author: Alberto Cuesta Cañada (@albertocuestacanada), Fiona Kobayashi (@fifikobayashi), fubuloubu (@fubuloubu), Austin Williams (@onewayfunction)
 discussions-to: https://ethereum-magicians.org/t/erc-3156-flash-loans-review-discussion/5077
-status: Review
+status: Last Call
+review-period-end: 2021-02-16
 type: Standards Track
 category: ERC
 created: 2020-11-15
--- a/EIPS/eip-3156.md
+++ b/EIPS/eip-3156.md
@@ -18,13 +18,13 @@ This ERC provides standard interfaces and processes for single-asset flash loans
 
 Flash loans allow smart contracts to lend an amount of tokens without a requirement for collateral, with the condition that they must be returned within the same transaction.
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 3156 state from Review to Last Call

MicahZoltu (changes requested):
Well formatted and mostly ready for merge, but it has a lot of references to external projects.  EIPs should be technical documents that stand on their own 5-10 years from now when no one remembers any of the projects that are around on Ethereum today.  Try to describe everything from the point of view of someone who isn't aware of previous/existing implementations of loans and instead prefer language that is agnostic to any specific project.

----

Remove external links and references to specific projects in the Motivation section.  In 5-10 years when someone is reading this EIP, most of those links won't work and individual products mentioned will probably be mostly irrelevant.  Recommend just rewording things so that you talk about the different strategies used by those dapps, but don't bring up the dapps themselves.

In Specification section, `lended` should be `loaned` I believe?  Or maybe `lent` (I don't think so)?  I don't think `lended` is a word in either case.

In Specification section, `lent` should be `loaned` I believe.  `loaned` is the past tense for `loan` and `lent` is the past tense for `lend`.

In Rationale section, remove references to specific projects.  Target audience of this EIP should be developers in 5-10 years who may not care at all about dapps that are around today.

In the Security Considerations section, same as above -- don't reference specific projects.  Stick to patterns, facts, assertions, etc. and avoid mentioning specific projects in examples whenever possible.  EIPs are technical documents and they don't need to include histories of things.  Try to describe the various security issue clearly without requiring the reader have any knowledge/understanding of existing projects.  Even references to DAI I would replace with references to an arbitrary token (e.g., STAB if you want to refer to an arbitrary stable coin for some reason, and just define it as "stable against some external asset")

albertocuestacanada:
I had a look into `loan`, `lend`, their uses and conjugations. In short, the verb is `lend` and the noun is `loan`. The past and participle of `to lend` is `lent`. You can choose not to follow the rules sometimes, which makes it confusing.

PR #3248, opened 2021-02-04
Move EIP-3076 to Review status

michaelsproul:
With EIP-3076 (slashing protection interchange) now implemented by the 4 major Eth2 clients (Lighthouse, Nimbus, Prysm, Teku), I would like to nominate it for review.

There are a few tweaks to the security recommendations that I would like to make before moving to Final Call, but the substance of the EIP has been stable for several months now and is unlikely to change.

### DIFF
--- a/EIPS/eip-3076.md
+++ b/EIPS/eip-3076.md
@@ -1,50 +1,38 @@
 ---
 eip: 3076
-title: Validator client interchange format (slashing protection)
+title: Slashing Protection Interchange Format
 author: Michael Sproul (@michaelsproul), Sacha Saint-Leger (@sachayves), Danny Ryan (@djrtwo)
 discussions-to: https://ethereum-magicians.org/t/eip-3076-validator-client-interchange-format-slashing-protection/
 status: Review
 type: Standards Track
 category: Core
 created: 2020-10-27
+updated: 2021-02-08
 ---
 
 ## Simple Summary
 
-A JSON interchange format for Serenity (eth2) that contains the necessary slashing protection information required to safely migrate keys between clients.
+A JSON interchange format for proof of stake validators to migrate slashing protection data between clients.
 
 ## Abstract
 
- A standard format for transferring a key's signing history allows validators to easily switch between clients without the risk of signing conflicting messages. While a [common keystore format](https://eips.ethereum.org/EIPS/eip-2335) provides part of the solution, it does not contain any information about a key's signing history. For a validator moving their keys from client A to client B, this can lead to scenarios in which client B inadvertently signs a message that conflicts with an earlier message signed with client A.  
- 
- We propose a database interchange format that helps solve this problem. The format contains a record of all blocks and attestations signed by a set of validators. 
+A standard format for transferring a key's signing history allows validators to easily switch between clients without the risk of signing conflicting messages. While a [common keystore format](https://eips.ethereum.org/EIPS/eip-2335) provides part of the solution, it does not contain any information about a key's signing history. For a validator moving their keys from client A to client B, this could lead to scenarios in which client B inadvertently signs a message that conflicts with an earlier message signed with client A. The interchange format described here provides a solution to this problem.
 
 ## Motivation
 
-Eth2 penalises validators for voting in ways that could result in two different versions of the chain being finalised. These types of penalties are called slashings.
+The proof of stake (PoS) protocol penalises validators for voting in ways that could result in two different versions of the chain being finalised. These types of penalties are called slashings.
 
-> The two main ways a validator can be slashed are 
->
-> 1. **Double voting**: voting for two different blocks during the same epoch.
-> 2. **Surround voting**: voting for one version of reality, and later voting for another version in a way that doesn't make it clear that they no longer believe in the first.
+For a validator following the protocol correctly, there is, in principle, no risk of being slashed. However, changing clients (from client A to client B, say) can result in a slashing risk if client B is unaware of the blocks and attestations that were signed with client A.
 
-For a validator following the protocol correctly, there is, in principle, no risk of being slashed. However, changing clients (from client A to client B, say) can result in a slashing risk if client B is unaware of the blocks and attestations that were signed with client A. 
+This can can occur if client A and client B do not agree on what the present time is. For example, say client A's time is accidentally set to a day in the future (225 epochs), and a validator switches from client A to client B without giving B a record of the blocks and attestations signed with A. The validator in question now runs the risk of attesting to two different blocks in the same epoch (a slashable offence) for the next 225 epochs (since they've already voted on these epochs with client A, and now stand to vote on them again with client B). Such time-skew bugs have been observed in the wild.
 
-Specifically, problems can occur if client A and client B do not agree on what the present time is (i.e if their clocks are not synchronised within a certain bound).
-
-For example, say client A's time is accidentally set to a day in the future (225 epochs), and a validator switches from client A to client B without giving B a record of the blocks and attestations signed with A.
-
-The validator in question now runs the risk of voting for two different blocks in the same epoch (a slashable offence) for the next 225 epochs (since they've already voted on these epochs with client A, and now stand to vote on them again with client B).
-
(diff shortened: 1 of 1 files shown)


kaycoinsUSA (approved):

MicahZoltu (changes requested):
Overall looks good, but there are some editorial changes that I would like to see made before moving to Review status to bring this EIP in line with the latest standards we are trying to encourage/enforce.

----

"Serenity" and "ETH2" are transient names.  EIPs should be written with a future reader (5-10 years from now) in mind, who may not be familiar with ancient internal code names or weird versioning schemes.  Can we reword the simple summary to not mention either?  Maybe something like (I also shortened it a bit):
> A JSON interchange format for proof of stake validators for key migration between clients with slashing protection.

Title probably should be a bit shorter if possible:
> PoS Validator Data Interchange Format

----

> We propose a database interchange format that helps solve this problem. The format contains a record of all blocks and attestations signed by a set of validators.

EIPs don't propose, they assert.  Recommend just removing this first sentence of this paragraph entirely as it just seems to reiterate what the previous paragraph says, and shorter is better when possible.

----

> The above situation may sound far-fetched, but it’s actually very similar to the incident that occurred on the eth2 Medalla testnet on 2020-08-14, where a time skew that affected all Prysm clients spiraled into a series of cascading failures. It resulted in the testnet experiencing several days without finality – an outcome that could have resulted in tens of millions of dollars in collective penalties had it occurred on mainnet.

Consider just stating that this situation was seen in the wild due to time skew rathere than doing a mini-retrospective.  EIPs should try to remain as technical as possible and avoid including history, opinions, too much discussion, etc.  Motivation is meant to help provide a high level understanding of "why" this EIP is valuable, but it doesn't need to be exhaustive or even convincing.

----

> An earlier design encompassed two separate formats to achieve the same goal: complete and minimal (for the complete changelog see here).

EIPs are not the appropriate place to talk about history.  Write them as a moment in time, with all necessary information for implementation contained within the EIP.  History type stuff is best recorded in other sources.

If there is something specific that was tried and failed that you think is worth mentioning, explicitly call it out in the Rationale section as "X was considered as an alternative, but we decided not do do that because of Y".

Note: This feedback applies to most of the rationale section.  Try to clean up up the whole thing and have it self-standing rather than being a sort of changelog from vPrev.

----

Consider splitting up the Rationale section into `### <discussion item>` subsections to help make it a bit easier to read, with each subsection covering one specific decision (again, strive for each subsection being totally self contained when possible).

----

Test Cases should either be inlined, added as an `../assets/eip-3076/<file>`, or the whole section can be removed (it is optional).  External links tend to die over time and we do not have the manpower to maintain expiring links.

----

Implementation section should be removed (it is optional) or a pseudo-code implementation should be inlined, or an implementation should be included in `../assets/eip-3076/<files>`.  External links tend to break, and a list of implementations will almost certainly end up out of date and we do not have the manpower to maintain such lists.

michaelsproul:
>  Can we reword the simple summary to not mention either?

Yeah, very happy with this change, as well as the of shortening of the title and other style changes. I'll make some changes and push them soon(ish).

> Test Cases should either be inlined, added as an ../assets/eip-3076/<file>, or the whole section can be removed (it is optional). External links tend to die over time and we do not have the manpower to maintain expiring links.

The tests are too big to inline, and I was hesitant to add them to the repo because I would like the option to add to them over time (we don't yet have any tests for files that should parse as invalid for example).

On a legalistic reading of EIP-1, it implies links to tests are OK, but I'm happy to accept your judegement as editor. Maybe I'll just remove the Test Cases section entirely.

> (EIP-1) Test Cases - Test cases for an implementation are mandatory for EIPs that are affecting consensus changes. Other EIPs can choose to include links to test cases if applicable.

----

> Implementation section should be removed (it is optional) or a pseudo-code implementation should be inlined

I'll remove it

MicahZoltu:
I'll work on getting EIP-1 fixed, as that doesn't align with our actual current policy.  😬

EIPS/eip-3076.md
```diff
@@ -1,50 +1,38 @@
...
+This can can occur if client A and client B do not agree on what the present time is. For example, say client A's time is accidentally set to a day in the future (225 epochs), and a validator switches from client A to client B without giving B a record of the blocks and attestations signed with A. The validator in question now runs the risk of attesting to two different blocks in the same epoch (a slashable offence) for the next 225 epochs (since they've already voted on these epochs with client A, and now stand to vote on them again with client B). Such time-skew bugs have been observed in the wild.
 
+Another situation in which slashing protection is critical is in the case of re-orgs. During a re-org it is possible for a validator to be assigned new attestation duties for an epoch in which it has already signed an attestation. In this case it is essential that the record of the previous attestation is available, even if the validator just moved from one client to another in the space of a single epoch.
```
michaelsproul, line 29: I added this paragraph about re-orgs to the motivation as I think it's just as compelling as the time-skew argument.

eip-automerger:

 - Trying to change EIP 3076 state from Draft to Review

EIPS/eip-3076.md
```diff
@@ -210,72 +197,60 @@ target_epoch <=
...
+* If your database records the signing roots of messages in addition to their slot/epochs, you should ensure that imported messages without signing roots are assigned a suitable dummy signing root internally. We suggest using a special "null" value which is distinct from all other signing roots, although a value like `0x0` may be used instead (as it is extremely unlikely to collide with any real signing root).
+* Care must be taken to avoid signing messages within a gap in the database (an area of unknown signing activity). This could occur if two interchanges were imported with a large gap between the last entry of the first and the first entry of the second. Signing in this gap is not safe, and would violate conditions (2), (4) and (5). It can be avoided by storing an explicit low watermark in addition to the actual messages of the slashing protection database, or by pruning on import so that the oldest messages from the interchange become the oldest messages in the database.
```
michaelsproul, line 251: I also added this security recommendation based on my own experience with Lighthouse. I had implemented slashing protection oblivious to this case until it was pointed out to me by @ajsutton (thanks again :heart:)

michaelsproul (commented):
Thanks @MicahZoltu for the review. I've attempted to address all of your concerns, and I think we're in a better shape to move EIP-3076 to Review. Please let me know if there are any spots you think could still be improved!

EIPS/eip-3076.md
```diff
@@ -210,72 +197,60 @@ target_epoch <=
...
 
-[pps]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#proposer-slashings
-[isad]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#is_slashable_attestation_data
-[csr]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#compute_signing_root
-[cd]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0-rc.0/specs/phase0/beacon-chain.md#compute_domain
+[pps]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0/specs/phase0/beacon-chain.md#proposer-slashings
+[isad]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0/specs/phase0/beacon-chain.md#is_slashable_attestation_data
+[csr]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0/specs/phase0/beacon-chain.md#compute_signing_root
+[cd]: https://github.com/ethereum/eth2.0-specs/blob/v1.0.0/specs/phase0/beacon-chain.md#compute_domain
```
MicahZoltu, line 209: If possible, I would like to see these external links removed.  I'm a bit more amenable to linking to a tag of an Ethereum specification repository (as you have done here) than my general stance on external links, but after watching how many sources of truth Ethereum 1 specification has gone through over time, my confidence that these links will survive the test of time is incredibly low.

Could we perhaps store a copy of `beacon-chain.md` from `eth2.0-specs` in `../assets/eip-3076/beacon-chain.md` and link to that instead?

MicahZoltu (approved):
One comment about external links to eth2.0 specs, but good enough to merge as Review.

PR #3250, opened 2021-02-05
Clarifies that test cases shouldn't be external links.

MicahZoltu:
We have been enforcing this for a while, I didn't realize EIP-1 said something different.


MicahZoltu:
Been open for a month, interpreting silence to mean tacit agreement.

PR #3252, opened 2021-02-05
Make clear that Berlin is tracked in eth1-specs

timbeiko:
We now track Berlin (and future upgrades)'s spec in the eth1 specs repo, not Meta EIPs. This makes it clear in the Berlin Meta EIP, given this is where people go by default to learn about the upgrade.

### DIFF
--- a/EIPS/eip-2070.md
+++ b/EIPS/eip-2070.md
@@ -4,36 +4,25 @@ title: "Hardfork Meta: Berlin"
 author: Alex Beregszaszi (@axic), Tim Beiko (@timbeiko)
 discussions-to: https://ethereum-magicians.org/t/hardfork-meta-eip-2070-berlin-discussion/3561
 type: Meta
-status: Draft
+status: Withdrawn
 created: 2019-05-20
 requires: 1679
 ---
 
 ## Abstract
 
-This meta-EIP specified the changes included in the Ethereum hardfork named Berlin. These changes are now tracked as part of the [eth1.0-specs repositoruy](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades). The list of EIPs included in Berlin and current implementation status is specified in [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md).
+This meta-EIP specified the changes included in the Ethereum hardfork named Berlin. These changes are now tracked as part of the [eth1.0-specs repository](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades). The list of EIPs included in Berlin and current implementation status is available at [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md).
 
-## Specification
+## Specification & Timeline
 
-See [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md) in the [eth1.0-specs repositoruy](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades).
-
-### Included EIPs
-
-See [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md) in the [eth1.0-specs repositoruy](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades).
-
-### Accepted EIPs
-
-See [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md) in the [eth1.0-specs repositoruy](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades).
-
-## Timeline
-
-See [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md) in the [eth1.0-specs repositoruy](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades).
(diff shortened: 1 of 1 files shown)


EIPS/eip-2070.md
```diff
@@ -11,37 +11,29 @@ requires: 1679
 
 ## Abstract
 
-This meta-EIP specifies the changes included in the Ethereum hardfork named Berlin.
+This meta-EIP specified the changes included in the Ethereum hardfork named Berlin. These changes are now tracked as part of the [eth1.0-specs repositoruy](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades). The list of EIPs included in Berlin and current implementation status is specified in [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md).
```
poemm, line 14: Typo "repositoruy". This typo also appears below several times.

eip-automerger:

 - EIP 2070 requires approval from one of (@axic)

MicahZoltu (changes requested):
My preference (mildly strong, will put up a fight, won't fight to the death) would be to just delete this whole file.

My second preference would be to move it to Withdrawn.

My third preference would be to keep the header data (so builds don't break) and replace the *entire* body with a single line saying that hard fork coordination occurs over at eth1.0-specs repository.

EIPS/eip-2070.md
timbeiko: Thanks @poemm!

timbeiko:
@MicahZoltu

> My preference (mildly strong, will put up a fight, won't fight to the death) would be to just delete this whole file.

Strong preference for me, keep the file because that's where literally everyone goes to find out about network upgrades. Totally OK not doing one for the next upgrade, though. WDYT? 

> My second preference would be to move it to Withdrawn.

Happy to do that as part of that change. LMK if that works! 

> My third preference would be to keep the header data (so builds don't break) and replace the entire body with a single line saying that hard fork coordination occurs over at eth1.0-specs repository.

I can remove entire sections, sure. @axic any preferences on that?

EIPS/eip-2070.md
timbeiko: Actually, re-reading that, I think "specified" is right: the EIP specified (in the past) the Berlin spec, it no longer does. I'll change the phrasing of the second sentence for clarity :+1:

MicahZoltu:
How would you feel about reducing the EIP to just:
```md
---
eip: 2070
title: "Hardfork Meta: Berlin"
author: Alex Beregszaszi (@axic), Tim Beiko (@timbeiko)
discussions-to: https://ethereum-magicians.org/t/hardfork-meta-eip-2070-berlin-discussion/3561
type: Meta
status: Withdrawn
created: 2019-05-20
requires: 1679
---

This meta-EIP specified the changes included in the Ethereum hardfork named Berlin.
These changes are now tracked as part of the [eth1.0-specs repository](https://github.com/ethereum/eth1.0-specs/tree/master/network-upgrades).
The list of EIPs included in Berlin and current implementation status is available at [`berlin.md`](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md).

## Copyright

```

timbeiko:
@MicahZoltu done :+1: @axic could you approve?

MicahZoltu (approved):

timbeiko:
@axic friendly ping :-)

axic:
Sorry, my notification queue on Github is beyond wrecked.

The initial goal of these hard fork meta EIPs was to document a hard fork **after** it went live. There was no such documentation prior to these meta EIPs and given many clients in their source code refer to them, I think it has served that purpose very well.

After their successful use in this manner we tried to utilise them during the planning phase. It seems that has been superseded with the new eth1.0-specs repository.

I would argue that these meta EIPs **should stay post hard fork** as a documentation means, and in order to reduce confusion, it should point to that resource during the planning phase. However after a fork went live, I think they should be updated with the final details.

What do you all think?

timbeiko:
I'll defer to @MicahZoltu :+1:

MicahZoltu:
@axic My thinking remains the same, that the EIPs repository is just a specifications/standardization repository and it is not a hardfork coordination repository.  That being said, one can argue that the set of EIPs that are included in a particular hardfork and what block that hard fork occurs on *is* a standardization concern.  However, I would personally rather keep the EIPs repository somewhat generalized so that it can apply to any Ethereum-like chain.  We already have EIPs in this repository that only apply to other blockchains and I think that is good for continuing to encourage cross-compatability as much as possible.  Hard fork block and EIP inclusion set feels like a *very* Ethereum specific thing, which is why I'm a fan of it going into the eth1.0-specs repository.

Lastly, having two sources of truth for information is generally a bad idea, and since the eth1.0-specs repository has this information I think it makes sense to not replicate it in this repository.

edsonayllon:
Hmm, keeping these kinds of EIPs while we have the specs repo would put us in this weird limbo where we have the similar or same information in both places.

This: 
- Increases the workload
- Causes confusion

If there is a transition to the specs repo, which I think should happen, in this case, it may be better to have a clean cut. And start getting everyone accustomed to referencing the new specs repo (press, client source code, etc.).

edsonayllon:
In design, a common practice to engineer innovation is separating ideation from judgment or appraisal. I like EIPs just being potential ingredients or puzzle pieces that can be used, while the specs repo is the place where the decisions are made on those pieces, and where people can see what the decisions made were.

timbeiko:
Closed in favor of https://github.com/ethereum/EIPs/pull/3305

PR #3253, opened 2021-02-06
Makes signature have type as the first byte.

MicahZoltu:
This is a security requirement of EIP-2718.

### DIFF
--- a/EIPS/eip-2930.md
+++ b/EIPS/eip-2930.md
@@ -50,7 +50,7 @@ As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introd
 
 The [EIP-2718](./eip-2718.md) `TransactionPayload` for this transaction is `rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, yParity, senderR, senderS])`.
 
-The `yParity, senderR, senderS` elements of this transaction represent a secp256k1 signature over `keccak256(rlp([1, chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]))`.
+The `yParity, senderR, senderS` elements of this transaction represent a secp256k1 signature over `keccak256(1 || rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]))`.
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

holiman (approved):

EIPS/eip-2930.md
```diff
@@ -50,7 +50,7 @@ As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introd
 
 The [EIP-2718](./eip-2718.md) `TransactionPayload` for this transaction is `rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, yParity, senderR, senderS])`.
 
-The `yParity, senderR, senderS` elements of this transaction represent a secp256k1 signature over `keccak256(rlp([1, chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]))`.
+The `yParity, senderR, senderS` elements of this transaction represent a secp256k1 signature over `keccak256(1 || rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, access_list]))`.
```
pipermerriam, line 53: This might be clearer if it was `0x01` instead of `1` which leaves some ambiguity over how it should be encoded (albeit a very minor amount of ambiguity)

holiman: cc @winsvega this will require `evm t8n` (and retesteth?)  changes

winsvega: yes. 
and I would have put the transaction hash as   `hash (rlp ( 1 || rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, access_list])))`

so that would match the hash of a legacy transaction which is calculated from    `hash( rlp( tx fields) )`

MicahZoltu: @winsvega See the security considerations of [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718#security-considerations) for the reason why we need to not have an extra layer of RLP wrapping when signing.  I believe most find it preferable for the unique identifier for the transaction to match the thing that is signed.

PR #3256, opened 2021-02-08
EIP-681: move to Final

nagydani:
Last call review period ended without changes being requested. ERC 681 is ready to become final.

### DIFF
--- a/EIPS/eip-681.md
+++ b/EIPS/eip-681.md
@@ -4,8 +4,7 @@ title: URL Format for Transaction Requests
 author: Daniel A. Nagy (@nagydani)
 type: Standards Track
 category: ERC
-status: Last Call
-review-period-end: 2021-02-07
+status: Final
 created: 2017-08-01
 requires: 20, 137
 ---


MicahZoltu (changes requested):
Discussions-to URL is required in the frontmatter (not sure how this was missed and why the bot isn't complaining):
```
discussions-to: <URL>
```

----

I believe `Ethers` in the simple summary should be `Ether`.

eip-automerger:

 - Trying to change EIP 681 state from Last Call to Final

nagydani:
> Discussions-to URL is required in the frontmatter (not sure how this was missed and why the bot isn't complaining):
> 
> ```
> discussions-to: <URL>
> ```

Should I create a fresh tread at Ethereum Magicians for discussing ERC-681?

> I believe `Ethers` in the simple summary should be `Ether`. "I have 15 Ether" not "I have 15 Ethers".

I can accept that.

When the above changes are made, the status must also go back to Review, right?

MicahZoltu:
Yeah, it should go back to Review (and merged as that) but you can then move it immediately back to Last Call (with a new 2 week timer) if you like.  That will trigger everyone being notified about its last call status starting again.  For non-normative changes like this I would normally say there is no need to do this, but since there was not any actual place for people to provide feedback I think a full 2 weeks in last call is necessary.  Sorry for not noticing it previously!

MicahZoltu:
@lightclient How hard would it be to add a check to make sure `discussions-to` is present?

MicahZoltu:
Actually, I think `ether` may be appropriate rather than `Ether` (saw this discussed elsewhere yesterday and I think `ether` won out).

PR #3260, opened 2021-02-09
2930: Changes 1 to 0x01 for clarity and moved to Review.

MicahZoltu:
Also removed empty optional sections while here.  They can be added back if/when tests/implementations are added, but they are otherwise optional.

I also moved this to review, since that is a much more appropriate state for an EIP that is going into a soon-to-be-scheduled hard fork, and there are no major changes planned to this EIP by the authors that I know of.


eip-automerger:

 - Trying to change EIP 2930 state from Draft to Review

holiman (approved):

PR #3265, opened 2021-02-11
EIP-2930: Make RLP encoding of access_list explicit

ryanschneider:
When reviewing EIP-2930 today I briefly misinterpreted the EIP to imply that the access_list should be a binary concatenated list of the values rather than RLP encoded; this is obviously wrong since it's open to ambiguity of inputs, so I figured making the RLP encoding of the access_list field explicit might help others from going down the same wrong path.


eip-automerger:

ryanschneider:
Actually the wording I used might instead make someone believe that the access_list needs to be "double encoded" so in hindsight I think it's actually clearer as is.

PR #3274, opened 2021-02-17
Future salaries - links to the Solidity sources

vporton:
Links to the Solidity sources were corrected.


EIPS/eip-future_salaries.md
```diff
@@ -1,3 +1,4 @@
+
```
MicahZoltu, line 1:
```suggestion
```
Frontmatter needs to be the first thing in the document.

PR #3275, opened 2021-02-18
ERC-681 back to Review

nagydani:
Fixed the spelling of ether as a unit and added a required header linking to the discussion of ERC-681 at FEM's forum.


eip-automerger:

 - Trying to change EIP 681 state from Last Call to Review

PR #3276, opened 2021-02-18
erc3156: Final Status

albertocuestacanada:
After extensive consultation and collaboration, and after receiving no further feedback for the two weeks this EIP has been in Last Call status, I would like to propose to make it Final.

### DIFF
--- a/EIPS/eip-3156.md
+++ b/EIPS/eip-3156.md
@@ -13,6 +13,10 @@ created: 2020-11-15
 
 This ERC provides standard interfaces and processes for single-asset flash loans.
 
+## Abstract
+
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 3156 state from Last Call to Final

MicahZoltu (changes requested):
This appears to be missing an Abstract:
```
## Abstract
A short (~200 word) description of the technical issue being addressed. This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does.
```
Was this section intentionally left out, or just an oversight?

albertocuestacanada:
Must have been an oversight, I'll write one. :+1:

PR #3290, opened 2021-02-23
EIP time based switch proposal

ivorrr:

 - The build passes.

### DIFF
--- a/EIPS/eip-3290.md
+++ b/EIPS/eip-3290.md
@@ -1,5 +1,5 @@
 ---
-eip: <to be assigned>
+eip: 3290
 title: Time Based Switch
 author: Branislav Djalic (@Omodaka9375), Malisa Pusonja (@horohronos), Ivor Jugo (@ivorrr)
 discussions-to: https://github.com/ethereum/EIPs/issues/3286
--- a/EIPS/eip-3290.md
+++ b/EIPS/eip-3290.md
@@ -10,7 +10,7 @@ created: 2021-02-23
 ---
 
 ## Simple Summary
-This EIP provides standard interfaces and processes for safe, conditional transactions based on user predefined timestamps. It would enable users to secure their accounts from lockout after account activity ceased for a predefining fallback address to transfer the funds.
+Processes for safe, conditional transactions based on user predefined timestamps.
 
 ## Abstract
 This EIP standard addresses the user's need to enable the transfer of his fund from an account that he can not access anymore, whatever the reason is. 
--- a/EIPS/eip-3290.md
+++ b/EIPS/eip-3290.md
@@ -5,7 +5,7 @@ author: Branislav Djalic (@Omodaka9375), Malisa Pusonja (@horohronos), Ivor Jugo
 discussions-to: https://github.com/ethereum/EIPs/issues/3286
 status: Draft
 type: Standards Track
-category (*only required for Standards Track): ERC
+category: ERC
 created: 2021-02-23
 ---
 
--- a/EIPS/eip-3290.md
+++ b/EIPS/eip-3290.md
@@ -62,13 +62,7 @@ TBS (Time Based Switch) feature integrates smart contracts and uses events/callb
 The user creates his TBS by calling the createSwitch function together with four parameters. This returns a signed transaction and puts it in a queue for due execution based on block height or a timestamp.
 
 ```
-function createSwitch(uint _time, uint _amount, address_executor, address payable _benefitor)
-    doesntExist
-    checkAmount(_amount)
-    checkTime(_time)
-    public
-    Payable 
-{...}
+function createSwitch(uint _time, uint _amount, address_executor, address payable _benefitor);
 ```
 
 TBS creation will produce a signed transaction that is put on hold until/if the trigger hits.
--- a/EIPS/eip-3290.md
+++ b/EIPS/eip-3290.md
@@ -70,13 +70,8 @@ All signed transactions are stored in the TBS queue object in the contract.
 
 Only the executor address is able to trigger a funds transfer when the defined timestamp has expired. The triggered event can be defined as any type of transaction or a specific transaction or blockchain information. Besides that, the creator will determine a percentage of balance dispersed per fallback address.
 
-``` 
-function tryExecuteSwitch(address account)
-    onlyValid(account)
-    onlyExecutorsOrOwner(account)
-    public
-    Payable
-{...}
+```
+function tryExecuteSwitch(address account);
 ```
 
 Required event listener methods for tokens transfer are:
--- a/EIPS/eip-time_based_switch.md
+++ b/EIPS/eip-time_based_switch.md
@@ -1,0 +1,172 @@
+---
+eip: <to be assigned>
+title: Time Based Switch
+author: Branislav Djalic (@Omodaka9375), Malisa Pusonja (@horohronos), Ivor Jugo (@ivorrr)
+discussions-to: https://github.com/ethereum/EIPs/issues/3286
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): ERC
+created: 2021-02-23
+---
+
+## Simple Summary
+This EIP provides standard interfaces and processes for safe, conditional transactions based on user predefined timestamps. It would enable users to secure their accounts from lockout after account activity ceased for a predefining fallback address to transfer the funds.
+
+## Abstract
+This EIP standard addresses the user's need to enable the transfer of his fund from an account that he can not access anymore, whatever the reason is. 
+
+We are aware of the complexity and technical knowledge that a user must have to operate with his crypto accounts securely, when we add that the users' costs of time and liability only on themselves for storing their funds, it could be one of the reasons why mass adoption still didn't come.
+
+By providing a trusted decentralized way to trigger this kind of scheduled transaction in the future, this EIP will help to avoid diversification of similar products and establish a universal solution to all future implementations. 
+
+The creator/user is required to fulfill a set of predefined rules when creating a time switch contract (Will go through more details in the specification section):
+
+* Time represents the timestamp between the current moment and the scheduled one. 
+* Assets and amount of the token that he is willing to transfer.
+* Recipients address and the percentage of funds that each one of them will get.
+* Executor address, which can trigger the transactions.
+
+## Motivation
+At this point, there is no standardized, inherent way for a user to have some secure switch that acts as a last line of defense against losing long-time access to an account.
+
+By adopting and standardizing the primary mechanism behind a time based switch, we eliminate the need to consider each particular implementation. This EIP is meant to address wide use cases and setup scenarios on top of it while keeping the core trustable and functional.
+
+Let's list the possible application layer use cases for such a standard: 
+
+1. Lost access to an account
+Users can set up an account and create a switch for it in such a way that, even if he loses his private keys in the future, the funds will be transferred to a predefined secondary account after a period of time.
+
+2. Cryptographic "Testament" or "Last Will."
+Imagine you would like to store your cryptocurrency savings in a contract that will be accessible by members of the family, but only after something has happened to us. Let's say we should "check-in" by evoking some contract function calls every so often.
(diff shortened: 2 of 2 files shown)


EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
+---
+eip: 3290
+title: Time Based Switch
+author: Branislav Djalic (@Omodaka9375), Malisa Pusonja (@horohronos), Ivor Jugo (@ivorrr)
+discussions-to: https://github.com/ethereum/EIPs/issues/3286
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): ERC
```
MicahZoltu, line 8:
```suggestion
category: ERC
```

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+created: 2021-02-23
+---
+
+## Simple Summary
+This EIP provides standard interfaces and processes for safe, conditional transactions based on user predefined timestamps. It would enable users to secure their accounts from lockout after account activity ceased for a predefining fallback address to transfer the funds.
```
MicahZoltu, line 13: Simple summary should be a similar length to an email subject line, pull request title line, etc.
```suggestion
Processes for safe, conditional transactions based on user predefined timestamps.
```

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+We are aware of the complexity and technical knowledge that a user must have to operate with his crypto accounts securely, when we add that the users' costs of time and liability only on themselves for storing their funds, it could be one of the reasons why mass adoption still didn't come.
+
+By providing a trusted decentralized way to trigger this kind of scheduled transaction in the future, this EIP will help to avoid diversification of similar products and establish a universal solution to all future implementations. 
+
+The creator/user is required to fulfill a set of predefined rules when creating a time switch contract (Will go through more details in the specification section):
+
+* Time represents the timestamp between the current moment and the scheduled one. 
+* Assets and amount of the token that he is willing to transfer.
+* Recipients address and the percentage of funds that each one of them will get.
+* Executor address, which can trigger the transactions.
```
MicahZoltu, line 27: Abstract should be a human readable summary of the technical changes.  Most of this content looks like stuff that would be more appropriate for the `Motivation` section as it describes *why* someone would want to implement this EIP.  Something like "Defines an interface for functions X and Y along with event Z".

Should be around a paragraph or two ideally, and be written in plain English (though referring to technical terms and functions is acceptable).  Don't go too detailed, and you don't have to describe the entire system, just the high level and most important bits.

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+The most common reason at the moment to lock funds is called "vesting." Imagine that you have just raised a successful ICO and your company still holds a majority of tokens distributed between your team members.
+It would be beneficial to all parties involved to ensure that people's tokens cannot be traded straightaway. If there are no controls in place, anyone might take action by selling all their tokens, cashing out, and quitting the company. This would negatively affect the market price and make all remaining contributors to the project unhappy.
+
+7. DEFI Yielding
+Protect the deposited liquidity or the pool itself.
+
+## Specification
+TBS (Time Based Switch) feature integrates smart contracts and uses events/callbacks to inform the user about the TBS state.
+
+The user creates his TBS by calling the createSwitch function together with four parameters. This returns a signed transaction and puts it in a queue for due execution based on block height or a timestamp.
```
MicahZoltu, line 62: This type of content is more appropriate for a tutorial or docs site, separate from the EIP.  The EIP should focus exclusively on defining the requirements of someone wanting to implement the interface, not walking people through how to use it.

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+
+```
+function createSwitch(uint _time, uint _amount, address_executor, address payable _benefitor)
+    doesntExist
+    checkAmount(_amount)
+    checkTime(_time)
+    public
+    Payable 
+{...}
+```
```
MicahZoltu, line 72: Interfaces cannot have modifiers or implementation:
````suggestion
```
function createSwitch(uint _time, uint _amount, address_executor, address payable _benefitor);
```
````

Comment (MicahZoltu) [reply EIPS/eip-3290.md:72]: The statement above says this should return something, but the function definition doesn't contain a return value.

File: EIPS/eip-3290.md [hunk 6]
```diff
@@ -0,0 +1,172 @@
...
+Only the executor address is able to trigger a funds transfer when the defined timestamp has expired. The triggered event can be defined as any type of transaction or a specific transaction or blockchain information. Besides that, the creator will determine a percentage of balance dispersed per fallback address.
+
+``` 
+function tryExecuteSwitch(address account)
+    onlyValid(account)
+    onlyExecutorsOrOwner(account)
+    public
+    Payable
+{...}
+```
```
MicahZoltu, line 86:
````suggestion
```
function tryExecuteSwitch(address account);
```
````

File: EIPS/eip-3290.md [hunk 7]
```diff
@@ -0,0 +1,172 @@
...
+
+Required event listener methods for tokens transfer are:
+```approveToken, withdrawToken```
```
MicahZoltu, line 89: If these are Ethereum events, they should contain a full definition.

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+
+TBS will keep track/store all created switches to be later updated by the owners at will.
+The changeable variables would be:
+
+* Change function fallback addresses as owner
+* Change function condition variables as owner
+* Cancel TBS as an owner
```
MicahZoltu, line 96: This looks like an implementation detail, which isn't appropriate for an EIP.

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+TBS smart contracts should be implemented/imported as a library with only a couple of public functions. Beneficiary logic, notifications, and any other implementation (see motivation) can be abstracted from this standard and implemented on top of it. 
+
+Here are some of the different approaches may include this standard:
+
+* Custom wallet integration for TBS setup
+* (De)centralized notifications/reminders before an action occur.
+* Custom fallback consensus and validation logic
+* Multi TBS setup
+* NFT transfers 
+
```
MicahZoltu, line 110:
```suggestion
```
Rationale should focus on specific design decisions and their reasoning.  For example, "We chose to return have function X return a value rather than not because Y."

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+We also considered a solution that triggers the tokens transfer. Still, after long consideration, we decided that it should be implemented on the application layer because we wanted to leave a more extensive range of integration with this standard. 
+
+## Backwards Compatibility
+This standard is compatible with any fungible and non-fungible token, such as ERC-20 or ERC-721.
+
+## Reference Implementation
+https://github.com/Omodaka9375/Time-Based-Switch-on-Ethereum/blob/main/contracts/TimeBasedSwitch.sol
```
MicahZoltu, line 117: This code should be inlined or included in `../assets/3290/TimeBasedSWitch.sol` and then linked from this document via relative link.

EIPS/eip-3290.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Miner Vulnerabilities
+TBS uses 'block.timestamp' to track and schedule transactions in the future only. The switch lifetime is defined as a minimum of 1 day. There are no current security considerations regarding outside dependencies for this core feature.
+
+### Gas Limit
+Avoiding gas limit issues:
+* Not using loops in TBS
+* No strings, only byte32
+* No SafeMath to reduce contract size and dependencies
+* All searches are in constant time or O(1), with the exception of ERC20 tokens, where performance is in proportion to the number of tokens added to the switch or O(N).
```
MicahZoltu, line 169: This section should describe gotchas and known risks that implementers need to watch out for.  Currently, this looks more like the result of a business risk analysis or an implementation audit report, which isn't appropriate for an EIP.

MicahZoltu (changes requested):
This feels very much like a Good Idea rather than a Standard.  The EIPs repository is for standardization, not a repository of good ideas.

I would really like to see a strong argument as to why this needs to be standardized.  Standards are good when you have a many-to-many relationship, like websites (many) to browsers (many).  They don't make sense when you have a one-to-many relationship like a Twitter (one) and integrators (many).

Standards also aren't necessary when there is little benefit to interoperability.  For example, if I build a dapp with some feature that dapp likely has a UI for it.  Someone else can build a dapp with a similar feature and their dapp will have a UI and there is no need for interoperability, so we don't need to share public surface area.  In fact, in situations like this standardization can lead to less innovation as people try to make their contracts look like each other without meaningful benefit.

Omodaka9375:
> 
> 
> This feels very much like a Good Idea rather than a Standard. The EIPs repository is for standardization, not a repository of good ideas.
> 
> I would really like to see a strong argument as to why this needs to be standardized. Standards are good when you have a many-to-many relationship, like websites (many) to browsers (many). They don't make sense when you have a one-to-many relationship like a Twitter (one) and integrators (many).
> 
> Standards also aren't necessary when there is little benefit to interoperability. For example, if I build a dapp with some feature that dapp likely has a UI for it. Someone else can build a dapp with a similar feature and their dapp will have a UI and there is no need for interoperability, so we don't need to share public surface area. In fact, in situations like this standardization can lead to less innovation as people try to make their contracts look like each other without meaningful benefit.

Hi @MicahZoltu , 

Thank you for taking the time to respond in such detail. 

The benefit of this EIP is not so obvious at the first look so allow me to explain.

The aim is to standardize time-based actions on the chain and allow for much more fun stuff to be built on top by any other entity, be it a company or a single developer. Rather than just being a good idea, by extracting and generalizing a simple proven mechanism, it enables others to innovate in an important niche where the business model is based on time tracking (eg. see use-cases).

Why do we believe this is necessary? 

Since delayed transactions on the chain can't be done without external dependencies, every approach we researched so far is done in isolation or with a specific implementation in mind. Therefore it is being coupled or hardcoded to a third-party oracle that is used to execute those actions after a certain amount of time. There are also different security considerations with different technical approaches and choices. 

So rather than making just another dApp, we believe it is important to have a well-known, secure way to bootstrap these types of applications, and standardization of this process can benefit the entire ecosystem. 

The proposed contract is agnostic towards the executors of the actions and is written in a way to enable future dApps to be on a many-to-many level. It allows the implementation of protocolary executors as well as manual ones. All complex business logic can be abstracted in another interfacing contract thus making it an independent layer of trust.

MicahZoltu:
Ethereum has timestamps built in, why do you need oracles or executors?

Omodaka9375:
Many business cases are in need of a decentralized version of a "cron job" or automated delayed transactions that are executed in a decentralized manner. Executors are the ones that users trust to trigger the switch after the designated time has passed. The executor can be a switch creator himself or any other address or a smart contract that will provide protocolar executors, for example Ethereum Alarm Clock etc...

MicahZoltu:
Ah, I can see value in standardizing an incentivization mechanism and interface for alarm clocks.  I think your example use cases are not a good use of such things though.  😄 

>  This EIP standard addresses the user's need to enable the transfer of his fund from an account that he can not access anymore, whatever the reason is. 

For this you are much better off with a simple contract wallet that just looks at the block time to determine whether it should release funds or not.

Omodaka9375:
Oh 😄 yup, that could be removed from there. I see why it misled you. It is actually just one of the edge use cases, and should not be there as anything else but an example for the implementation. Will will correct the text.

EIPS/eip-3290.md
ivorrr: This has now been addressed in the referenced implementation.

EIPS/eip-3290.md
ivorrr: Could you please elaborate on what do you mean by full definition?

MicahZoltu:
```solidity
event approveToken(address token, uint256 amount);
```

PR #3293, opened 2021-02-24
Update eip-1559.md

Mister-Meeseeks:
The motivation section contains examples of peak gas prices. Since the document was drafted, those records have been shattered. For better accuracy, I updated with more current information.

 - The build passes.


eip-automerger:

EIPS/eip-1559.md
```diff
@@ -26,7 +26,7 @@ The transaction will always pay the base fee per gas of the block it was include
...
+* **Mismatch between volatility of transaction fee levels and social cost of transactions**: bids to include transactions on mature public blockchains, that have enough usage so that blocks are full, tend to be extremely volatile. On Ethereum, minimum bids range between 1 and 200 nanoeth (10^9 nanoeth = 1 ETH), but sometimes go over 500 nanoeth and have reached over 2,000 nanoeth. This clearly creates many inefficiencies, because it's absurd to suggest that the cost incurred by the network from accepting one more transaction into a block actually is 2000x more when gas prices are 2000 nanoeth than when they are 1 nanoeth; in both cases, it's a difference between 8 million gas and 8.02 million gas.
```
MicahZoltu, line 29: I would much rather just change this paragraph to be much shorter and focus purely on the economic argument rather than using real-world example numbers.  EIPs should be written targeting a future audience, and current gas prices on Ethereum are not relevant to such a reader.
```suggestion
* **Mismatch between volatility of transaction fee levels and social cost of transactions**: bids to include transactions on mature public blockchains, that have enough usage so that blocks are full, tend to be extremely volatile. It's absurd to suggest that the cost incurred by the network from accepting one more transaction into a block actually is 10x more when the cost per gas is 10 nanoeth compared to when the cost per gas is 1 nanoeth; in both cases, it's a difference between 8 million gas and 8.02 million gas.
```

abdelhamidbakhta (approved):
Looks good to me.

PR #3295, opened 2021-02-26
Change EIP-2876 status from "Draft" to "Review"

junderw:

 - The build passes.

### DIFF
--- a/EIPS/eip-2876.md
+++ b/EIPS/eip-2876.md
@@ -3,7 +3,7 @@ eip: 2876
 title: Deposit contract and address standard
 author: Jonathan Underwood (@junderw)
 discussions-to: https://github.com/junderw/deposit-contract-poc/issues/1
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 2876 state from Draft to Review

MicahZoltu (changes requested):
Recommend tightening up simple summary:
> A contract interface for managing deposits by defining a specialized address format.

----

## Additional thoughts

* Payment channels and zkRollups feel like a significantly better solution to this class of problems.  They allow both ETH and Token transfers to exchanges/merchants at notably lower gas costs than today, and they come with a number of other benefits over a solution like this.
* What advantage is there for standardizing on this?  Can't exchanges and merchants do the exact same thing that every dapp does when they want a user to call a contract (put a button on the page and the user interacts via their wallet)?

PR #3296, opened 2021-02-26
Fixes some misspelled words, grammar and comprehension

varunkashyap:
- Replaces hyphen with colons where applicable
- Introduces em-dashes wherever necessary
- Fixes typos
- A better sentence structure for a few long sentences


varunkashyap:
This is my first PR. I found some issues with `EIP-1`, which I tried to fix in the PR. Do I need to sign a CoC document or agree to a license?

EIPS/eip-1.md
```diff
@@ -46,9 +48,9 @@ REVERT (0xfe)
...
+Once the idea has been vetted, your next responsibility will be to present(by means of an EIP) the idea to the reviewers and all interested parties, invite editors, developers, and the community to give feedback on the aforementioned channels. You should try and gauge whether the interest in your EIP is commensurate with both the work involved in implementing it and how many parties will have to conform to it. For example, the work required for implementing a Core EIP will be much greater than for an ERC and the EIP will need sufficient interest from the Ethereum client teams. Negative community feedback will be taken into consideration and may prevent your EIP from moving past the Draft stage.
```
MicahZoltu, line 53:
```suggestion
Once the idea has been vetted, your next responsibility will be to present (by means of an EIP) the idea to the reviewers and all interested parties, invite editors, developers, and the community to give feedback on the aforementioned channels. You should try and gauge whether the interest in your EIP is commensurate with both the work involved in implementing it and how many parties will have to conform to it. For example, the work required for implementing a Core EIP will be much greater than for an ERC and the EIP will need sufficient interest from the Ethereum client teams. Negative community feedback will be taken into consideration and may prevent your EIP from moving past the Draft stage.
```

EIPS/eip-1.md
```diff
@@ -96,13 +98,13 @@ Each EIP should have the following parts:
...
-- Rationale - The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.
+- Rationale - The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussions.
```
MicahZoltu, line 103: I think `discussion` is correct here.  While technically there may be multiple discussions, I believe `discussion` is one of those words that can be used in a sort-of-plurality sort of way.

EIPS/eip-1.md
```diff
@@ -96,13 +98,13 @@ Each EIP should have the following parts:
...
 - Backwards Compatibility - All EIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The EIP must explain how the author proposes to deal with these incompatibilities. EIP submissions without a sufficient backwards compatibility treatise may be rejected outright.
-- Test Cases - Test cases for an implementation are mandatory for EIPs that are affecting consensus changes. Other EIPs can choose to include links to test cases if applicable.
+- Test Cases - Test cases for the implementation are mandatory for EIPs that are affecting consensus changes. Other EIPs can choose to include links to test cases if applicable.
```
MicahZoltu, line 105: `an` is correct here.  There will be many implementations of an EIP, and the tests cases need to work with all of them.

EIPS/eip-1.md
```diff
@@ -213,7 +215,7 @@ Images, diagrams and auxiliary files should be included in a subdirectory of the
...
 
-If you are interested in assuming ownership of an EIP, send a message asking to take over, addressed to both the original author and the EIP editor. If the original author doesn't respond to email in a timely manner, the EIP editor will make a unilateral decision (it's not like such decisions can't be reversed :)).
+If you are interested in assuming ownership of an EIP, send a message asking to take over, addressed to both the original author and the EIP editor. If the original author doesn't respond to the email in time, the EIP editor will make a unilateral decision (it's not like such decisions can't be reversed :)).
```
MicahZoltu, line 218: I believe `in a timely manner` is correct here.  There is no fixed time limit, and how long is acceptable is quite subjective.  "a timely manner" is a common way of expressing such a concept, while `in time` implies a hard deadline.

MicahZoltu (changes requested):

EIPS/eip-1.md
varunkashyap: We use `discussions` in this section multiple times in the next few lines. So although both can work, I changed to `discussions` for consistency.

EIPS/eip-1.md
varunkashyap: Reverted. Thanks!

EIPS/eip-1.md
varunkashyap: I agree, `in a timely manner` is less formal and implies a softer deadline, albeit still seems a bit redundant. I do feel having a generous upper time limit will add clarity to the text as well as to the process. I am not sure there is precedent for such a proposed time limit though.

So in the meanwhile, I have reverted the text back to `in a timely manner`. I will keep this in mind as I get more familiar with the process and see if we need more details here.

EIPS/eip-1.md
MicahZoltu: There appears to be a bit of a mix `discussion` and `discussions` throughout this document.  I just went through them and feel like they are all used correctly currently.

Thinking and researching more on why I think this should be `discussion`, I believe it is because its usage here is as an uncountable noun, while `discussions` is the plural form of a countable noun.  `discussion` can be either a countable noun or an uncountable noun, and which one is appropriate depends on the context.  In this particular context, I am reasonably confident that the uncountable noun version is appropriate.

varunkashyap: Reverted to discussion and added a missing article

EIPS/eip-1.md
```diff
@@ -213,7 +215,7 @@ Images, diagrams and auxiliary files should be included in a subdirectory of the
...
 
-If you are interested in assuming ownership of an EIP, send a message asking to take over, addressed to both the original author and the EIP editor. If the original author doesn't respond to email in a timely manner, the EIP editor will make a unilateral decision (it's not like such decisions can't be reversed :)).
+If you are interested in assuming ownership of an EIP, send a message asking to take over, addressed to both the original author and the EIP editor. If the original author doesn't respond to the email in a timely manner, the EIP editor will make a unilateral decision (it's not like such decisions can't be reversed :)).
```
MicahZoltu, line 218: Ugh, I didn't even know this section was in here... I should probably go through and clean this up to align better with reality after we merge your changes...

MicahZoltu (approved):

PR #3298, opened 2021-02-26
Removal of refunds

vbuterin:
Remove gas refunds for SSTORE and SELFDESTRUCT.

### DIFF
--- a/EIPS/eip-3297.md
+++ b/EIPS/eip-3297.md
@@ -2,7 +2,7 @@
 eip: 3297
 title: Removal of refunds
 author: Vitalik Buterin (@vbuterin), Martin Swende (@holiman)
-discussions-to: https://to_be_determined.com/example.link
+discussions-to: https://ethereum-magicians.org/t/eip-3298-removal-of-refunds/5430
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-3297.md
+++ b/EIPS/eip-3297.md
@@ -1,0 +1,95 @@
+---
+eip: 3297
+title: Removal of refunds
+author: Vitalik Buterin (@vbuterin), Martin Swende (@holiman)
+discussions-to: https://to_be_determined.com/example.link
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-02-26
+---
+
+## Simple Summary
+
+Remove gas refunds for SSTORE and SELFDESTRUCT.
+
+## Motivation
+
+Gas refunds for SSTORE and SELFDESTRUCT were originally introduced to motivate application developers to write applications that practice "good state hygiene", clearing storage slots and contracts that are no longer needed. However, they are not widely used for this, and poor state hygiene continues to be the norm. It is now widely accepted that the only solution to state growth is some form of [statelessness or state expiry](https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/state_size_management), and if such a solution is implemented, then disused storage slots and contracts would start to be ignored automatically.
+
+Gas refunds additionally have multiple harmful consequences:
+
+* Refunds give rise to [GasToken](https://gastoken.io/). GasToken has benefits in moving gas space from low-fee periods to high-fee periods, but it also has downsides to the network, particularly in exacerbating state size (as state slots are effectively used as a "battery" to save up gas) and inefficiently clogging blockchain gas usage
+* Refunds increase block size variance. The theoretical maximum amount of actual gas consumed in a block is nearly twice the on-paper gas limit (as refunds add gas space for subsequent transactions in a block, though refunds are capped at 50% of a transaction's gas used). This is [not fatal](https://notes.ethereum.org/@vbuterin/eip_1559_spikes), but is still undesirable, especially given that refunds can be used to maintain 2x usage spikes for far longer than EIP 1559 can.
+
+## Specification
+
+### Parameters
+
+| Constant | Value |
+| - | - |
+| `FORK_BLOCK` | TBD |
+
+For blocks where `block.number >= FORK_BLOCK`, the following changes apply.
+
+Do not apply the `refund`. 
+
+The description above is sufficient to describe the change, but for the sake of clarity we enumerate all places where gas refunds are currently used and which should/could be removed within a node implementation. 
+
+1. Remove all use of the "refund counter" in SSTORE gas accounting, as defined in [EIP 2200](https://eips.ethereum.org/EIPS/eip-2200). Particularly:
+
+    * If a storage slot is changed and the _current value_ equals the _original value_, but does not equal the _new value_, `SSTORE_RESET_GAS` is deducted (plus `COLD_SLOAD_COST` if [prescribed by EIP 2929 rules](https://eips.ethereum.org/EIPS/eip-2929#sstore-changes)), but no modifications to the refund counter are made.
+    * If a storage slot is changed and the _current value_ equals neither the _new value_ nor the _original value_ (regardless of whether or not the latter two are equal), `SLOAD_GAS` is deducted (plus `COLD_SLOAD_COST` if [prescribed by EIP 2929 rules](https://eips.ethereum.org/EIPS/eip-2929#sstore-changes)), but no modifications to the refund counter are made.
+
+2. Remove the `SELFDESTRUCT` refund.
+
+## Rationale
+
+A full removal of refunds is the simplest way to solve the issues with refunds; any gains from partial retention of the refund mechanism are not worth the complexity that that would leave remaining in the Ethereum protocol.
+
+## Backwards Compatibility
+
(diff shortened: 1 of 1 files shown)


lepidotteri:
NO

lightclient (approved):

The-3D:
I'm curious to know if any study has being made on what will the impact of removing refunding after cleaning up the state. Is the state bloat due to gastoken minting really worth removing a gas saving feature and an incentive to keep the state in check for smart contract developers?
I went through hell sometimes in the past trying to clean up the state, pretty damn sure i won't do the same if no refunds are being made. Let's see if it's actually worth removing state management incentives or not.

lightclient:
@The-3D please see the motivation in the EIP, it says this EIP is more motivated by reducing the variability of the block size.

vorot93:
@The-3D

"GasTokens" are not a gas saving "feature" - it's an exploit in the gas calculation for the more cunning, at the expense of everyone else _and_ the nodes.

State should _not_ be naively "cleaned up", as it makes client implementations a lot more complex. We should explore alternative solutions like reGenesis.

The-3D:
> @The-3D please see the motivation in the EIP, it says this EIP is more motivated by reducing the variability of the block size.

It says both, and to be honest far more relevance is given to state bloat due to the gasToken (it even mentions block size variance as undesiderable, but not fatal). I can understand the block size variability of course, what i'm asking is if there is any research/simulation on estimated impact on the network should this EIP go through.

I also don't agree with this 
```However, they are not widely used for this, and poor state hygiene continues to be the norm.```

This might be true for many contracts (maybe the majority) but for sure it's not true for the most used ones (the ones that actually generate the majority of the activity on the network). Most of the best protocols out there, include top Defi protocols, have excellent state management.

**Comment by The-3D** [comment]
> @The-3D
> 
> "GasTokens" are not a gas saving "feature" - it's an exploit in the gas calculation for the more cunning, at the expense of everyone else _and_ the nodes.
> 
> State should _not_ be naively "cleaned up", as it makes client implementations a lot more complex. We should explore alternative solutions like reGenesis.

I never said anything against this. In fact, i agree with both. I'm all for finding an alternative way for state management, manual garbage collection is painful and error prone.

**Comment by Alpha-Guy** [comment]
LGTM!!

**Comment by greenlucid** [comment]
Isn't this a bit drastic?

**Comment by vbuterin** [comment]
> Isn't this a bit drastic?

See this doc for my more generalized case in favor of doing drastic things between now and the merge: https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/evm_feature_removing

> Most of the best protocols out there that i checked, including top Defi protocols, have excellent state management.

I think the most important category that even the best state management practices can't avoid is users forgetting about some application. Has anyone who stopped using MakerDAO self-destructed their CDP? Do orders on an on-chain order book that just never get filled because the price moves in one direction soon after they're made ever get deleted? Do airdrop tokens worth $0.001 ever go away? There's this large unavoidable category of state growth that comes from users ceasing to interact with some application (or ethereum entirely) and not bothering to clean up their state, and only some form of statelessness or state expiry (including regenesis) can fix that.

**Comment by k06a** [comment]
Maybe to keep this refunds for storage dirtied in the same tx? Else this would explode gas costs for thing like approve+transferFrom :(

**Comment by vbuterin** [comment]
Increasing remaining gas by 15000 when storage goes from (orig=zero, current=nonzero) to (orig=zero, current=zero) seems reasonable, though it would violate an invariant in that a child call with N gas would be able to execute more than N gas. Seems worth discussing.

**Comment by GNSPS** [comment]
Seeing this purely from an attack surface mitigation position I fully support this EIP.
The rationale of rolling back a decision that did not add to a better state utilization is also sound.

Fully backing this up as I hope ConsenSys Diligence will. 👍

**Comment by jomarip** [comment]
Honestly, I did not know I could self-destruct contracts and assets. What is the process for doing this so I can clean up while I wait for this to pass

**Comment by Dadybayo** [comment]
It is important to think about the security of the Ethereum blockchain rather than the benefits of the protocols. Yes, Gas token hacked the rules of the Ethereum game, but it did not bring anything developing to the market

Now it is important to focus on Layer 2 solutions and only together we can build a new world!

**Comment by greenlucid** [comment]
> > Isn't this a bit drastic?
> 
> See this doc for my more generalized case in favor of doing drastic things between now and the merge: https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/evm_feature_removing

Thanks, I know understand the rationale behind this.
A bit related and drastic: should gas costs of EVM opcodes stay static? Should they be voted, like miners vote for gas limit? Wondering about possible perverse incentives that allowing for fluid costs of EVM opcodes could make. The advantage I see is that maybe processing-related opcodes could be agreed by miners to stay cheaper, and memory-related opcodes could be considered more expensive by miners.
SUM, a very cheap computation instruction costs 3 gas, meanwhile a permanent store opcode SSTORE costs 20k. Computers can compute millions of SUM operations in a second, but a SSTORE will make the state increasingly heavier for whoever comes next. This is just my perception on why I think gas costs look a bit rigid and arbitrary to me.

**Comment by wjmelements** [comment]
Please defer discussion [here](https://ethereum-magicians.org/t/eip-3298-removal-of-refunds/5430)

File: EIPS/eip-3297.md [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+
+Gas refunds additionally have multiple harmful consequences:
+
+* Refunds give rise to [GasToken](https://gastoken.io/). GasToken has benefits in moving gas space from low-fee periods to high-fee periods, but it also has downsides to the network, particularly in exacerbating state size (as state slots are effectively used as a "battery" to save up gas) and inefficiently clogging blockchain gas usage
+* Refunds increase block size variance. The theoretical maximum amount of actual gas consumed in a block is nearly twice the on-paper gas limit (as refunds add gas space for subsequent transactions in a block, though refunds are capped at 50% of a transaction's gas used). This is [not fatal](https://notes.ethereum.org/@vbuterin/eip_1559_spikes), but is still undesirable, especially given that refunds can be used to maintain 2x usage spikes for far longer than EIP 1559 can.
+
```
jochem-brouwer, line 24: Another side-effect worth mentioning is that `eth_estimateGas` returns the amount of gas paid for. However, due to refunds, the gasLimit one should use could be higher (up to 2x the reported estimated gas).

wjmelements:
No, eth_estimateGas reflects the required gas limit.

EIPS/eip-3297.md
```diff
@@ -0,0 +1,95 @@
...
+2. Remove the `SELFDESTRUCT` refund.
+
+## Rationale
+
+A full removal of refunds is the simplest way to solve the issues with refunds; any gains from partial retention of the refund mechanism are not worth the complexity that that would leave remaining in the Ethereum protocol.
+
+## Backwards Compatibility
+
+Refunds are currently only applied _after_ transaction execution, so they cannot affect how much gas is available to any particular call frame during execution. Hence, removing them will not break the ability of any code to execute, though it will render some applications economically nonviable.
```
jochem-brouwer, line 52: It should be noted that currently we can create blocks which execute almost 2x the gas limit of the block in gas (if each transaction gets the max 50% refund). So this EIP effectively halves the maximum gas executed in the worst case. This will definitely impact how much transactions fit in a block, since now any transaction which used a refund, is now more expensive.

jochem-brouwer:
I just moved these points to magicians.

@wjmelements if you have a very simple transaction, where a storage slot of a contract has a nonzero value, and the code only SSTOREs 0 to that slot, then estimateGas will (IIRC) report that you need something like 13000 gas for the entire transaction (since this is what you are charged), but is obviously incorrect. So it does not reflect the right gas limit (unless I'm missing something here).

SuperXingXing:
If no refund, will SSTORE be cheaper?

KaycoinsInc (approved):

KaycoinsInc:
KaBoom!

KaycoinsInc (approved):

elvisbegovic:
Just Wow

alisscco (commented):
This will work ?

sudoFerraz:
I am intrigued in how Coinbase Commerce for example would be viable if these changes hit production.
The forwarder contract factory strategy they use take advantage of the refunds to be able to sweep the funds from the forwarder contracts to their cold wallet. Without the refunds, the gas costs on this sweep phase would be basically unpractical depending on the amount of the transaction done between their merchant and the customer.

This EIP also affects multiple Crypto payment providers that eat up fees for their customers in order to provide a better UX (generating unique addresses from forwarder contracts created from a factory), in which the refunds help with the operational costs.

Am I wrong when I suppose that these payment providers will become unviable using the same architecture and they will need to provide shared addresses for customers to pay (Consequently having to do KYC with wallet signing to prove their address and creating a way to match these users to these transactions sent to the same address)

If this would not impact on the sweep phase, then I am all in for it, but if it gets into the way of user adoption in the sense that people will need to do KYC and understand how to use actual wallets to sign or call certain operations in contracts instead of just sending a transaction to buy something, I am not sure if I would support it.

@jochem-brouwer @wjmelements @vbuterin

PR #3301, opened 2021-02-27
Change Ether currency symbol from Ξ to ⧫

pggallagher:
Re-opening from virgil's post in ethresear.ch Oct '17

https://ethresear.ch/t/proposal-change-ether-currency-symbol-from-to/120

### DIFF
--- a/EIPS/eip-3297.md
+++ b/EIPS/eip-3297.md
@@ -1,0 +1,95 @@
+---
+eip: 3297
+title: Removal of refunds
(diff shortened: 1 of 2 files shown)


MicahZoltu:
I think you may have submitted a PR for the wrong thing, as this looks like an accidental copy of a different PR.  Going to close for now.  Also, I recommend picking up the discussion on ethresear.ch first to see if people are still interested these days.

PR #3302, opened 2021-02-28
EIP-2228: Promote to Review

fulldecent:
Promoting to Last Call

### DIFF
--- a/EIPS/eip-2228.md
+++ b/EIPS/eip-2228.md
@@ -36,6 +36,14 @@ The network name shall be Ethereum Mainnet, or just Mainnet if the context is kn
 
 Any name or word styling (i.e. capitalization of the letters) of the network which is inconsistent with the test cases cited below shall NOT be used.
 
+### Trademark note
+
+Note: Ethereum [is a registered trademark](https://tsdr.uspto.gov/#caseNumber=86634529&caseType=SERIAL_NO&searchType=statusSearch) of Ethereum Foundation and [the Foundation asserts](https://ethereum.org/en/terms-of-use/) "you must not use [this mark] without the prior written permission of the Foundation".
+
+Your application MAY reference "Ethereum&reg; Mainnet" and inculde a note "Ethereum is a registered trademark of Ethereum Foundation" elsewhere.
+
+Trademark status, fair use of trademarks, and any written permissions you have with the Ethereum Foundation are outside the scope of this EIP.
+
 ## Rationale
 
 Choosing common word use promotes interoperability of implementations and increases customer awareness. Also, it adds a sense of professionalism when customers see the same word and word styling (i.e. capitalization of letters) across different implementations.
--- a/EIPS/eip-2228.md
+++ b/EIPS/eip-2228.md
@@ -40,7 +40,7 @@ Any name or word styling (i.e. capitalization of the letters) of the network whi
 
 Note: Ethereum [is a registered trademark](https://tsdr.uspto.gov/#caseNumber=86634529&caseType=SERIAL_NO&searchType=statusSearch) of Ethereum Foundation and [the Foundation asserts](https://ethereum.org/en/terms-of-use/) "you must not use [this mark] without the prior written permission of the Foundation".
 
-Your application MAY reference "Ethereum&reg; Mainnet" and inculde a note "Ethereum is a registered trademark of Ethereum Foundation" elsewhere.
+Your application MAY reference "Ethereum&reg; Mainnet" and include a note "Ethereum is a registered trademark of Ethereum Foundation" elsewhere.
 
 Trademark status, fair use of trademarks, and any written permissions you have with the Ethereum Foundation are outside the scope of this EIP.
 
--- a/EIPS/eip-2228.md
+++ b/EIPS/eip-2228.md
@@ -22,13 +22,13 @@ This standard specifies the name for this network and provides reference example
 
 The Ethereum network with chainid 1 is referenced from authoritative sources using several names that are conflicting:
 
-* Vitalik has called it "main net" at http://eips.ethereum.org/EIPS/eip-155 http://eips.ethereum.org/EIPS/eip-170
-* Vitalik has called it "mainnet" at http://eips.ethereum.org/EIPS/eip-86 
-* Vitalik has called it "Main net" at http://eips.ethereum.org/EIPS/eip-2* http://eips.ethereum.org/EIPS/eip-150
+* "main net" at [EIP-155](./eip-155.md), [EIP-170](./eip-170.md)
+* "mainnet" at [EIP-86 [DRAFT]](./eip-86.md), [client documentation](https://geth.ethereum.org/support/FAQ)
+* "Main net" at [EIP-2](./eip-2.md), [EIP-150](./eip-150.md)
    * \* In EIP-2, it is unclear if the first letter is capital because it is a proper noun or if it is the the first word in a text column presented in sentence case
-* Vitalik has called it "Mainnet" at http://eips.ethereum.org/EIPS/eip-7
-* Geth documentation calls its "mainnet" at https://geth.ethereum.org/support/FAQ
-* A final EIP here has referred to it as "Mainnet" at http://eips.ethereum.org/EIPS/eip-779
+* "Mainnet" at [EIP-7](./eip-7.md), [EIP-779](./eip-779.md)
+
+In several examples above, even papers written by the same author use inconsistent names to refer to the Ethereum network with chainid 1.
 
 ## Specification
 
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - Trying to change EIP 2228 state from Draft to Last Call

MicahZoltu (changes requested):
Recommend removing Vitalik's name from the motivation.  I think it is better to just refer to it being called XXX in EIP-YYY rather than appealing to authority and saying who the speaker was in each case.

----

EIP links must be in the form `[EIP-XXX](./eip-XXX.md)`

----

Remove the external link to Geth.  We can instead just say that some client documentation calls it `mainnet`.

----

Consider just mentioning in the motivation that it is called X, Y, and Z in various EIPs and around the internet.  Presumably, once we finalize this EIP someone will go through and clean up the EIPs with the incorrect name at which point all of the links to "evidence" will no longer make sense.  If the motivation of this EIP just says "there are places where it is called these other things" then we don't have to worry about their mutability.

----

Remove `## References` section, it isn't a standard EIP section.

----

Remove `## Implementation` section, it isn't a required section and since it contains no reference implementation it can be removed.

poojaranjan:
I think there is an intermediary status "Review" introduced with the new [EIP Standardization process](https://medium.com/ethereum-cat-herders/the-new-ethereum-improvement-process-928c628b306e). 

Instead of moving it to the "Last Call", @fulldecent consider moving to the "Review" status.

MicahZoltu:
Oh, good catch @poojaranjan.  This should definitely be moving to Review rather than Last Call.

fulldecent:
@MicahZoltu all items are addressed in these commits. Requesting, please your, promotion to Review status.


5c94ace02d5efdc2a68f029d6fd245c0cfe55b80

135a1da21be1aba1aeeb541b4a7dd4dd8ee56007

14be2c7e0d77f862b2d76bfda5a72194e5c58cb8

602622ba39121f7ba3cc24a229d82c99befca155

EIPS/eip-2228.md
```diff
@@ -20,22 +20,30 @@ This standard specifies the name for this network and provides reference example
...
 
 The network name shall be Ethereum Mainnet, or just Mainnet if the context is known to be discussing Ethereum networks. This IS a proper noun. Several examples are given below which differentiate between usage of the name of the network versus a descriptive reference to the network.
 
 Any name or word styling (i.e. capitalization of the letters) of the network which is inconsistent with the test cases cited below shall NOT be used.
 
+### Trademark note
+
+Note: Ethereum [is a registered trademark](https://tsdr.uspto.gov/#caseNumber=86634529&caseType=SERIAL_NO&searchType=statusSearch) of Ethereum Foundation and [the Foundation asserts](https://ethereum.org/en/terms-of-use/) "you must not use [this mark] without the prior written permission of the Foundation".
+
+Your application MAY reference "Ethereum&reg; Mainnet" and inculde a note "Ethereum is a registered trademark of Ethereum Foundation" elsewhere.
```
MicahZoltu, line 43:
```suggestion
Your application MAY reference "Ethereum&reg; Mainnet" and include a note "Ethereum is a registered trademark of Ethereum Foundation" elsewhere.
```

EIPS/eip-2228.md
```diff
@@ -20,22 +20,30 @@ This standard specifies the name for this network and provides reference example
...
 Any name or word styling (i.e. capitalization of the letters) of the network which is inconsistent with the test cases cited below shall NOT be used.
 
+### Trademark note
+
+Note: Ethereum [is a registered trademark](https://tsdr.uspto.gov/#caseNumber=86634529&caseType=SERIAL_NO&searchType=statusSearch) of Ethereum Foundation and [the Foundation asserts](https://ethereum.org/en/terms-of-use/) "you must not use [this mark] without the prior written permission of the Foundation".
+
+Your application MAY reference "Ethereum&reg; Mainnet" and include a note "Ethereum is a registered trademark of Ethereum Foundation" elsewhere.
+
+Trademark status, fair use of trademarks, and any written permissions you have with the Ethereum Foundation are outside the scope of this EIP.
+
```
MicahZoltu, line 46: The general EIP policy is to not include any external links.  Also, I feel like an EIP is not the appropriate place to mention trademarks or other legal statuses.  If people want to read the ToU they can read the ToU, no need to restate it here.  Your last line of this sums up my position pretty cleanly, but feels unnecessary to include in the EIP.
```suggestion
```

EIPS/eip-2228.md
```diff
@@ -20,22 +20,30 @@ This standard specifies the name for this network and provides reference example
...
+* "main net" at [EIP-155](./eip-155.md), [EIP-170](./eip-170.md)
+* "mainnet" at [EIP-86 [DRAFT]](./eip-86.md), [client documentation](https://geth.ethereum.org/support/FAQ)
+* "Main net" at [EIP-2](./eip-2.md), [EIP-150](./eip-150.md)
    * \* In EIP-2, it is unclear if the first letter is capital because it is a proper noun or if it is the the first word in a text column presented in sentence case
-* Vitalik has called it "Mainnet" at http://eips.ethereum.org/EIPS/eip-7
-* Geth documentation calls its "mainnet" at https://geth.ethereum.org/support/FAQ
-* A final EIP here has referred to it as "Mainnet" at http://eips.ethereum.org/EIPS/eip-779
+* "Mainnet" at [EIP-7](./eip-7.md), [EIP-779](./eip-779.md)
+
+In several examples above, even papers written by the same author use (as published at the time EIP 2228 was written) inconsistent names to refer to the Ethereum network with chainid 1.
```
MicahZoltu, line 31:
```suggestion
The Ethereum network with chainid 1 is referenced using several conflicting names across EIPs, client implementations, and information published on the internet at large.  In several locations, even documents written by the same author use inconsistent names to refer to the Ethereum network with chainid 1.  Names in use at the time of this writing include:
* `main net`
* `mainnet`
* `Main net`
* `Mainnet`
```

At the least, you'll need to remove draft EIPs from the list and the external link to `https://geth.ethereum.org/support/FAQ`.  I don't believe that linking to these documents provides any future value to this EIP so I recommend just getting rid of all of the links and instead just providing a list of options you have seen used around the internet.

Also, I recommend adding something to the motivation section that explains *why* there is value in consistency here, as I think that is more important than showing that there is a historic lack of consistency.

EIPS/eip-2228.md
MicahZoltu, line 46: The general EIP policy is to not include any external links.  Also, I feel like an EIP is not the appropriate place to mention trademarks or other legal statuses.  If people want to read the ToU they can read the ToU, no need to restate it here.  Your last line of this sums up my position pretty cleanly, but feels unnecessary to include in the EIP.
```suggestion
```

MicahZoltu (changes requested):

fulldecent:
@MicahZoltu has raised a few points here:

- Whether it is appropriate to mention Geth in a document surveying usage of "main net" wording
- Whether EIPs may link to things
- Whether a document published under the (delegated) authority of Ethereum Foundation and discussing the naming of a product created by the Ethereum Foundation is appropriate to reference a written policy Ethereum Foundation already has on usage of that name

I would like a chance to make points on these topics.

At the moment, this PR is requesting to promote EIP-2228 into "Review" status which represents "An EIP Author marks an EIP as ready for and requesting Peer Review". I think "Review" is an appropriate status given the level of discussions we are having and I request that those above questions are closed without prejudice here so they can be more appropriately considered at the EIP Discussions-To address.

MicahZoltu:
Ah, I forgot that this was Review and not Last Call (just updated PR title to reflect that).

That being said, the bar to get to Review is that it follows all of the current editorial rules (no external links, no links to draft EIPs).

While I agree that my commentary on the terms of use reference is perhaps a discussion that could happen elsewhere and could a change that isn't necessary for moving to Review, the link to it goes against our *current* policy of no external links.

We can move the discussion of whether ToU is appropriate over to the discussions-to link if you like, but if you want to propose a change the current policy of no external links that probably should happen elsewhere (unrelated to this EIP).

EIPS/eip-2228.md
fulldecent: This is adopted substantially at 199121d174f28f35264980b19cfc96ec8fd8e8fd

EIPS/eip-2228.md
fulldecent: The normative part of this comment is addressed in b5ccfed34f56831fa7d9a1a1becbb87a24a2739a. I refer further discussion to the discussions-to page.

fulldecent:
The requirement "no external links" is not documented. Here is a PR that adds that documentation: https://github.com/ethereum/EIPs/pull/3358

---

The requirement "no links to draft EIPs" is not documented. Here is a PR that adds that documentation: https://github.com/ethereum/EIPs/pull/3357

Well noted, thank you

---

I have made commits to address all normative feedback here. Requesting please a ⬇️ CREATE A MERGE COMMIT, and for discussion to continue at the discussion-to address.

PR #3308, opened 2021-03-02
mark 2612 for last call

MrChico:
This spec has been solid for quite a while and is more than ready for last call

### DIFF
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -3,7 +3,7 @@ eip: 2612
 title: permit – 712-signed approvals
 author: Martin Lundfall (@Mrchico)
 discussions-to: https://github.com/ethereum/EIPs/issues/2613
-status: Draft
+status: Last call
 type: Standards Track
 category: ERC
 created: 2020-04-13


poojaranjan:
There is an intermediary status "Review" introduced with the new [EIP Standardization process](https://medium.com/ethereum-cat-herders/the-new-ethereum-improvement-process-928c628b306e). 

Instead of moving the proposal to the "Last Call", @MrChico consider moving to the "Review" status.

EIPS/eip-2612.md
```diff
@@ -3,7 +3,7 @@ eip: 2612
 title: permit – 712-signed approvals
 author: Martin Lundfall (@Mrchico)
 discussions-to: https://github.com/ethereum/EIPs/issues/2613
-status: Draft
+status: Last call
```
MrChico, line 6:
```suggestion
status: Review
```

MicahZoltu (changes requested):
This EIP depends on EIP-712 but EIP-712 is still in draft.  Either convince the authors to move that EIP forward, or create your own EIP that this can depend on.  This likely will block forward movement/progress of this EIP so I recommend addressing this ASAP.

----

Simple Summary should be much shorter.  Think email subject line, GitHub PR title, forum thread title, etc.  Most of the content currently in the simple summary would be better suited in the motivation section.

----

The abstract should be a summary of the Specification section in a human readable form.  It doesn't need to cover all of the edge cases or minutia, but it should give the reader a high level overview of the specification.  All of the current content in the Abstract should be moved to the Motivation section.

----

Avoid linking to non-final EIPs anywhere in the EIP.

When you link EIPs, it should also be to `./eip-###.md` (never to GitHub).

----

In the motivation section, try to avoid spending too much time talking about history.  Instead, talk about use cases that are fulfilled by this EIP (consider just deleting the whole first three paragraphs of the Motivation section).

Keep in mind, EIPs are technical documents and not argumentative essays, repositories of history, etc.  The purpose of the motivation section is to simply give readers a high level overview of use cases for the EIP and why someone would want to use it, not to convince a naysayer of its value.

----

EIPs should always be in the format `EIP-###` when referenced.  `ERC20` is incorrect.

----

The Rationale should indicate why certain design choices were made over other options.  The current contents of the Rationale would be more appropriate in the Motivation section (which is for describing use cases for the EIP in general).  An example design decision might be why a particular constant was chosen over another, why a particular name was chosen, other patterns, etc.

----

Backward compatibility section should be higher level.  Rather than pointing at specific instances of alternate implementations, just mention that there are incompatible implementations in the wild and potentially how incompatibility with them could be addressed by an application author.  Definitely don't link to external sources.

----

Either inline the test cases, add them to `../assets/eip-2612/*` or remove the section (it is optional).

----

Remove the Implementation section or add a reference implementation inline or add a reference implementation to `../assets/eip-2612/*`.

----

Don't externally link to documents tin the Security Considerations section.  I wouldn't even bother mentioning the ERC-20 approval issue, as it is a "problem" with ERC-20 and not with this specification.

----

The comment about domain separator feels like it should be in the Security Considerations section of EIP-712, not this EIP.

----

Remove everything after the license, that should be the last section.  I'm not sure what that is supposed to be referencing, but inlining would be preferred.

PR #3309, opened 2021-03-02
Change EIP-2315 status from "Draft" to "Review"

poojaranjan:

 - The build passes.

### DIFF
--- a/EIPS/eip-2315.md
+++ b/EIPS/eip-2315.md
@@ -1,7 +1,7 @@
 ---
 eip: 2315
 title: Simple Subroutines for the EVM
-status: Draft
+status: Review
 type: Standards Track
 category: Core
 author: Greg Colvin <greg@colvin.org>, Martin Holst Swende (@holiman)


eip-automerger:

 - Trying to change EIP 2315 state from Draft to Review
 - EIP 2315 requires approval from one of (@gcolvin, @holiman)

MicahZoltu (changes requested):
The first reference to each new opcode in the Specification section must be in the form `OPCODE (0xNN)`.  e.g., `BEGINSUB (0x5c)`

----

Implementations section should be removed (or reworked).  An inline reference implementation would be acceptable, or something in `../assets/eip-2315/*`.  External links to implementations tend to result in 404s over time and implementations that people care about in 5 years are unlikely to be the same implementations that people care about today.

----

The costs and opcodes need to be moved to the `Specification` section.  They also should not "suggest", but instead they should assert.

----

`Appendix: Comparative costs.` isn't a valid EIP section.  Recommend just removing it as it doesn't add value to future implementers and instead is just an argument for why to implement this (which is appropriate for the Motivation section).

PR #3310, opened 2021-03-02
Change EIP-2537 status from "Draft" to "Review"

poojaranjan:

 - The build passes.

### DIFF
--- a/EIPS/eip-2537.md
+++ b/EIPS/eip-2537.md
@@ -3,7 +3,7 @@ eip: 2537
 title: Precompile for BLS12-381 curve operations
 author: Alex Vlasov (@shamatar)
 discussions-to: https://ethereum-magicians.org/t/eip2537-bls12-precompile-discussion-thread/4187
-status: Draft
+status: Review
 type: Standards Track
 category: Core
 created: 2020-02-21


eip-automerger:

 - Trying to change EIP 2537 state from Draft to Review
 - EIP 2537 requires approval from one of (@shamatar)

MicahZoltu (changes requested):
`Important notes` isn't an EIP section.  Its contents should be moved to `Specifictaion`, `Security Considerations`, or `Rationale` section.  Consider creating a sub-section in one of those.

----

Don't skip nesting levels (`##` to `####` without going through `###` first) as this breaks table of contents and potentially proper formatting in some renderers.

----

Consider adding gas cost per precompile to the abstract table (at least the high level bits, doesn't have to be exhaustive and a gas cost range is fine for the abstract).

----

Test cases need to be moved to the EIPs repository.  `../assets/eip-2537/*` is fine if there are too many to reasonably inline.

----

The implementation section should either be removed or contain a reference implementation inline, or a link to a reference implementation in `../assets/eip-2537/*`.  I recommend just removing it as it is an optional section and I'm guessing a reference implementation would not be trivial to pack up into a few files in the assets directory.  If you do retain this section, rename it to `Reference Implementation` to align with https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md.

----

Consider breaking this EIP up into one EIP per precompile.  Generally, smaller EIPs are much easier to get included.  While I can appreciate that describing BLS-381 in one EIP is easier, you could describe BLS in one place and then have an EIP (that depends on that generalized BLS description) for each precompile so you don't have to repeat yourself.

shamatar:
@poojaranjan do you mind to close this one, and I’ll address the requested changes in my own PR?

@MicahZoltu breaking it into smaller ones will not make it simpler in any sense because all the precompiles are just interfaces to the same arithmetic engine. And the engine is the main complexity point here

poojaranjan:
Closing on the suggestion of @shamatar. A new pull request will be created to address changes.

MicahZoltu:
> breaking it into smaller ones will not make it simpler in any sense because all the precompiles are just interfaces to the same arithmetic engine. And the engine is the main complexity point here

One of the big advantages of breaking apart EIPs is that smaller EIPs tend to go through the process more smoothly.  It makes it much less likely that someone will block the EIP because they disagree with one little piece of it and it also forces conversation to be centralized around the specific issue people have.  In this particular case, breaking it apart would allow us to pursue the recommendation at https://notes.ethereum.org/@GW1ZUbNKR5iRjjKYx6_dJQ/Skxf3tNcg_.  If it isn't broken up, a new EIP will need to be drafted which is likely to slow down the process and potentially cause this to miss London just because it was a monolithic EIP.

shamatar:
I can not access an internal link. But from all practical standpoints the main objection (as we have observed) would be that it's too scary and complex, but all this "complex" staff is in the arithmetic engine that would be shared between any eips if I make a lot of them with different functions. So it would multiply hard discussion about arithmetics + some whining about the interface into much larger N * arithmetics :)

MicahZoltu:
@shamatar Sorry, looks like my previous link somehow was bad:
https://notes.ethereum.org/B1D789itTd-m10W3aXq3CA

> the main objection (as we have observed) would be that it's too scary and complex

I recommend reading the above link, I don't think that is the *real* objection (which I discuss in more detail there).

PR #3314, opened 2021-03-03
3074: Removes change to ORIGIN.

MicahZoltu:
As much as I *hate* ORIGIN with a burning passion and would love to have it broken/removed, I don't think sneaking it in with this EIP is the right way to go.  ORIGIN references the transaction that initiated the transaction, and that doesn't change by adding a new `CALL` opcode.  I would 100% support a separate EIP that removes `ORIGIN` or otherwise breaks it, just not embedded in a new opcode.

### DIFF
--- a/EIPS/eip-3074.md
+++ b/EIPS/eip-3074.md
@@ -11,11 +11,11 @@ created: 2020-10-15
 
 ## Simple Summary
 
-Creates a new EVM instruction, analogous to `CALL` (`0xF1`), that sets `CALLER` (`0x33`) and `ORIGIN` (`0x32`) based on an ECDSA signature.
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - EIP 3074 requires approval from one of (@samwilsn)

SamWilsn:
If we aren't setting origin, we should probably add this precondition:

> If setting `ORIGIN` is unacceptable, an alternative is to not set `ORIGIN` and for `TXCALL` to fail if `sponsor == sponsee`.

That fixes the reentrancy properties of `require(tx.origin == msg.sender)`.

SamWilsn (approved):

PR #3332, opened 2021-03-05
MEDGASPRICE opcode

JChanceHud:
Waiting for feedback and an EIP number to be assigned.

### DIFF
--- a/EIPS/eip-medgasprice_opcode.md
+++ b/EIPS/eip-medgasprice_opcode.md
@@ -1,0 +1,93 @@
+---
+eip: <to be assigned>
+title: OPCODE 0x46 MEDGASPRICE
+author: Justice Hudson (@jchancehud)
+discussions-to:
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-05-03
+---
+
+## Simple Summary
+
+An opcode allowing smart contracts to access information about recent gas prices.
+
+## Abstract
+
+There is an existing opcode `0x3a GASPRICE` that allows contracts to access the gas price of the current transaction. This value can be used to impose a maximum gas price on contract calls by setting a constant value, but cannot be used as a long term solution given the fluctuation of gas prices over time. The proposed opcode `0x46 MEDGASPRICE` would return the median gas price for the parent block. This would allow the implementation of upper bounds on transaction gas prices as a function of the last block median gas price.
+
+## Motivation
+
+With the emergence of rollups as core mechanisms in scaling Ethereum there are a number of common transactions that can be front-run. Optimistic rollups rely on the submission of fraud proofs to maintain the integrity of their systems. As a result actors submitting fraud proofs typically receive a financial reward for doing so. This opens a trivial front-running strategy of watching the mempool for fraud proof submissions and copying such transactions with a much higher gas price to reap the reward. Such front-runners do not perform validation independently and de-incentivize others from performing validation. Adding a mechanism enforcing an upper bound on gas prices for a transaction could be an effective defense against such front-running attacks.
+
+## Specification
+
+If `block.number >= TBD`, add a new opcode:
+
+MEDGASPRICE: 0x46
+
+Pushes the median gas price of the parent block onto the stack.<br />
+Gas costs: 8
+
+The name `MEDGASPRICE` was chosen because the median gas price of the network can only be calculated from the latest complete block. Thus transactions being executed should expect the median gas price to be calculated from the previous block.
+
+## Rationale
+
+Consider a smart contract that wants to implement a first come first serve mechanism. Such a mechanism must defeat the inherently pay-to-win nature of the gas price market. Enforcing a maximum gas price for a transaction relies on the fact that transactions of the same gas price are generally processed in a first in first out way by Ethereum miners. A contract currently has few options for setting a max gas price:
+
+- Set a constant value at a reasonable rate given the current gas prices
+- Allow an individual or group of individuals to adjust a max gas price over time
+
+More elaborate schemes could likely be constructed but all would involve storing gas price information on chain increasing the number of transactions and costing Ether.
+
+Given a median gas price opcode a contract can set a maximum gas price as a function of the last blocks gas price. This can easily be implemented using a strategy such as the following:
+
+```
+// Assume that block.medgasprice is bound to the MEDGASPRICE opcode
+
+function submitFraudProof(bytes calldata proof) public {
+  require(tx.gasprice <= maxGasPrice());
+  // process the fraud proof and provide a reward (if valid)
+}
+
+function maxGasPrice() public view returns (uint) {
+  return 3 * block.medgasprice;
+}
+```
+
+Given the contract implementation above a client would simply call `maxGasPrice` to determine the gas price to use when submitting a fraud proof. This particular implementation allows up to 3x the median gas price of the last block to be used.
+
+## Backwards Compatibility
+
+This proposal introduces a single new opcode. No backwards incompatibilities should be introduced.
+
+## Forwards Compatibility
+
+EIP 1557 plans to change the fee market in a number of ways. Most notably is the creation of a base fee that is burned. In this context an "inclusion fee" still exists as a part of the total fee. Consider the following two cases:
+
+##### Block sizes are increasing (all available gas is being consumed)
+
+In this case there will be bidding contention in the inclusion fee to incentivize miners to include transactions. A `MEDGASPRICE` operator would still be helpful as an attacker could supply a high inclusion fee to bump honest transactions.
+
+##### Block sizes are decreasing (excess gas is available)
+
+In this case an attacker could specify a high inclusion fee to incentivize miners to include their transaction early in the block. Miners are incentivized to do so as including expensive transactions first reduces the risk of a revert (and partial refund) occurring.
+
+Given these two cases this EIP seems relevant in the context of EIP 1557.
+
+Post EIP 1557 the `MEDGASPRICE` operator should return the median `effective_gas_price` of the previous block.
+
+An additional opcode `BASEGASPRICE` (0x47) should be added so contracts can determine the inclusion fee at runtime (by subtracting from `GASPRICE`). Without such an opcode the anti front-running strategies described in this document do not work.
+
+## Security Considerations
+
+The strategy described for preventing front-running by setting an upper bound on the gas price of transactions has a few caveats:
+
+1. It relies on miners being impartial. Reordering transactions with the same gas price is a trivial means of defeating this strategy.
+2. It relies on the implicit behavior of miners ordering transactions of the same value in the order received. This is not standard behavior but seems to be general practice.
+3. The value returned by `MEDGASPRICE` may fluctuate rapidly between blocks. If a transaction is not included immediately it may either fail (if the gas price drops) or become vulnerable to front-running (if the gas price increases).
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
+---
+eip: <to be assigned>
```
lightclient, line 2: 3332

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+title: OPCODE 0x46 MEDGASPRICE
```
lightclient, line 3:
```suggestion
title: MEDGASPRICE Opcode
```

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+discussions-to: https://ethereum-magicians.org/t/medgasprice-opcode-eip/5480
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-03-05
+---
+
+## Simple Summary
+
+An opcode allowing smart contracts to access information about recent gas prices.
```
lightclient, line 14: I recommend referring to the opcode here using the format `MEDGASPRICE (0x46)` instead of saying "an opcode".

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Abstract
+
+There is an existing opcode `0x3a GASPRICE` that allows contracts to access the gas price of the current transaction. This value can be used to impose a maximum gas price on contract calls by setting a constant value, but cannot be used as a long term solution given the fluctuation of gas prices over time. The proposed opcode `0x46 MEDGASPRICE` would return the median gas price for the parent block. This would allow the implementation of upper bounds on transaction gas prices as a function of the last block median gas price.
```
lightclient, line 18:
```suggestion
There is an existing opcode `GASPRICE (0x3A)` that allows contracts to access the gas price of the current transaction. This value can be used to impose a maximum gas price on contract calls by setting a constant value, but cannot be used as a long term solution given the fluctuation of gas prices over time. The proposed opcode `MEDGASPRICE (0x46)` would return the median gas price for the parent block. This would allow the implementation of upper bounds on transaction gas prices as a function of the last block median gas price.
```

lightclient: This is the EIP-1 defined way to refer to opcodes.

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+## Specification
+
+If `block.number >= TBD`, add a new opcode:
+
+MEDGASPRICE: 0x46
+
+Pushes the median gas price of the parent block onto the stack.<br />
+Gas costs: 8
+
+The name `MEDGASPRICE` was chosen because the median gas price of the network can only be calculated from the latest complete block. Thus transactions being executed should expect the median gas price to be calculated from the previous block.
```
lightclient, line 33: This should be under rationale if included, but seems rather self-explanatory.

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+
+Given the contract implementation above a client would simply call `maxGasPrice` to determine the gas price to use when submitting a fraud proof. This particular implementation allows up to 3x the median gas price of the last block to be used.
+
+## Backwards Compatibility
+
+This proposal introduces a single new opcode. No backwards incompatibilities should be introduced.
+
+## Forwards Compatibility
+
+[EIP 1559](https://eips.ethereum.org/EIPS/eip-1559) plans to change the fee market in a number of ways. Most notably is the creation of a base fee that is burned. In this context an "inclusion fee" still exists as a part of the total fee. Consider the following two cases:
```
lightclient, line 67:
```suggestion
[EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) plans to change the fee market in a number of ways. Most notably is the creation of a base fee that is burned. In this context an "inclusion fee" still exists as a part of the total fee. Consider the following two cases:
```

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+
+##### Block sizes are increasing (all available gas is being consumed)
+
+In this case there will be bidding contention in the inclusion fee to incentivize miners to include transactions. A `MEDGASPRICE` operator would still be helpful as an attacker could supply a high inclusion fee to bump honest transactions.
+
+##### Block sizes are decreasing (excess gas is available)
+
+In this case an attacker could specify a high inclusion fee to incentivize miners to include their transaction early in the block. Miners are incentivized to do so as including expensive transactions first reduces the risk of a revert (and partial refund) occurring.
+
+Given these two cases this EIP seems relevant in the context of EIP 1559.
```
lightclient, line 77:
```suggestion
Given these two cases this EIP seems relevant in the context of EIP-1559.
```

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+
+Post EIP 1559 the `MEDGASPRICE` operator should return the median `effective_gas_price` of the previous block.
```
lightclient, line 79:
```suggestion
Post EIP-1559 the `MEDGASPRICE` operator should return the median `effective_gas_price` of the previous block.
```

EIPS/eip-medgasprice_opcode.md
```diff
@@ -0,0 +1,93 @@
...
+
+An additional opcode `BASEGASPRICE` (0x47) should be added so contracts can determine the inclusion fee at runtime (by subtracting from `GASPRICE`). Without such an opcode the anti front-running strategies described in this document do not work.
```
lightclient, line 81: This is the `BASEFEE` opcode, EIP-3198.

lightclient (changes requested):

EIPS/eip-medgasprice_opcode.md
JChanceHud: Thanks for the heads up!

EIPS/eip-medgasprice_opcode.md
JChanceHud: I think it's somewhat counterintuitive for the code `block.medgasprice` to refer to the previous block gas price. I kept it in but moved it to the rationale section.

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+created: 2021-03-05
+---
+
+## Simple Summary
+
+Add `MEDGASPRICE (0x46)` that allows smart contracts to access information about current gas prices.
+
+## Abstract
+
+There is an existing opcode `GASPRICE (0x3a)` that allows contracts to access the gas price of the current transaction. This value can be used to impose a maximum gas price on contract calls by setting a constant value, but cannot be used as a long term solution given the fluctuation of gas prices over time. The proposed opcode `MEDGASPRICE (0x46)` would return the median gas price for the parent block. This would allow the implementation of upper bounds on transaction gas prices as a function of the last block median gas price.
```
MicahZoltu, line 18:
```suggestion
Adds `MEDGASPRICE (0x46) opcode that returns the median gas price for the parent block.
```
The abstract should be a human readable version of the specification.  In this case, the specification is very simple so the abstract ends up being quite simple/short.  I also suggest altering the simple summary to be *less* specific (even more generalized).  What is in the abstract currently is mostly motivation, which should be in the Motivation section.

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+discussions-to: https://ethereum-magicians.org/t/medgasprice-opcode-eip/5480
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-03-05
+---
+
+## Simple Summary
+
+Add `MEDGASPRICE (0x46)` that allows smart contracts to access information about current gas prices.
```
MicahZoltu, line 14:
```suggestion
An opcode for getting the median gas price of the parent block.
```

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Motivation
+
+With the emergence of rollups as core mechanisms in scaling Ethereum there are a number of common transactions that can be front-run. Optimistic rollups rely on the submission of fraud proofs to maintain the integrity of their systems. As a result actors submitting fraud proofs typically receive a financial reward for doing so. This opens a trivial front-running strategy of watching the mempool for fraud proof submissions and copying such transactions with a much higher gas price to reap the reward. Such front-runners do not perform validation independently and de-incentivize others from performing validation. Adding a mechanism enforcing an upper bound on gas prices for a transaction could be an effective defense against such front-running attacks.
```
MicahZoltu, line 22: This wouldn't prevent rollup fraud proof front-running, it would just just make it so miner collusion is required (which is not hard to get these days).  A more appropriate solution would be a commit-reveal scheme for fraud proofs.

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Specification
+
+If `block.number >= TBD`, add a new opcode:
+
+MEDGASPRICE: 0x46
+
+Pushes the median gas price of the parent block onto the stack.<br />
+Gas costs: 8
```
MicahZoltu, line 31: This should specify how many stack inputs it takes and how many stack outputs it yields.

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+
+With the emergence of rollups as core mechanisms in scaling Ethereum there are a number of common transactions that can be front-run. Optimistic rollups rely on the submission of fraud proofs to maintain the integrity of their systems. As a result actors submitting fraud proofs typically receive a financial reward for doing so. This opens a trivial front-running strategy of watching the mempool for fraud proof submissions and copying such transactions with a much higher gas price to reap the reward. Such front-runners do not perform validation independently and de-incentivize others from performing validation. Adding a mechanism enforcing an upper bound on gas prices for a transaction could be an effective defense against such front-running attacks.
+
+## Specification
+
+If `block.number >= TBD`, add a new opcode:
+
+MEDGASPRICE: 0x46
+
+Pushes the median gas price of the parent block onto the stack.<br />
```
MicahZoltu, line 30:
```suggestion
Pushes the median gas price of the parent block onto the stack.

```
Try to avoid using HTML in markdown files as it may not be supported in all rendering engines and doesn't read well when read as a raw file.

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+  require(tx.gasprice <= maxGasPrice());
+  // process the fraud proof and provide a reward (if valid)
+}
+
+function maxGasPrice() public view returns (uint) {
+  return 3 * block.medgasprice;
+}
+```
+
+Given the contract implementation above a client would simply call `maxGasPrice` to determine the gas price to use when submitting a fraud proof. This particular implementation allows up to 3x the median gas price of the last block to be used.
```
MicahZoltu, line 57: This is a use case, which is appropriate for `Motivation` section, not `Rationale` section.
```suggestion
```

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Naming note
```
MicahZoltu, line 59:
```suggestion
### Naming note
```
Try to not skip headers along the way as that breaks tables of contents.

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+
+The name `MEDGASPRICE` was chosen because the median gas price of the network can only be calculated from the latest complete block. Thus transactions being executed should expect the median gas price to be calculated from the previous block.
+
+## Backwards Compatibility
+
+This proposal introduces a single new opcode. No backwards incompatibilities should be introduced.
```
MicahZoltu, line 65:
```suggestion
There are no known backwards incompatibility issues.
```
The first sentence doesn't belong in the Backward Compatibility section (it just repeats what already exists in other sections).

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+##### Block sizes are decreasing (excess gas is available)
+
+In this case an attacker could specify a high inclusion fee to incentivize miners to include their transaction early in the block. Miners are incentivized to do so as including expensive transactions first reduces the risk of a revert (and partial refund) occurring.
+
+Given these two cases this EIP seems relevant in the context of EIP-1559.
+
+Post EIP-1559 `MEDGASPRICE (0x46)` should return the median `effective_gas_price` of the previous block.
+
+[EIP-3198](https://eips.ethereum.org/EIPS/eip-3198) is required for the above strategies to be implemented. With the inclusion of `BASEFEE (0x48)` a contract can subtract the `base_fee_per_gas` from the `effective_gas_price` to determine the inclusion fee per gas being paid for the transaction and thus implement an upper bound.
+
```
MicahZoltu, line 84: This isn't a valid EIP section.  Consider moving this content to the `Motivation` section.
```suggestion
```

EIPS/eip-3332.md
```diff
@@ -0,0 +1,95 @@
...
+## Security Considerations
+
+The strategy described for preventing front-running by setting an upper bound on the gas price of transactions has a few caveats:
+
+1. It relies on miners being impartial. Reordering transactions with the same gas price is a trivial means of defeating this strategy.
+2. It relies on the implicit behavior of miners ordering transactions of the same value in the order received. This is not standard behavior but seems to be general practice.
```
MicahZoltu, line 90: This isn't true.  Miners "often" (from a security standpoint) sort transactions against gas price.  There are also established mechanisms for paying miners out of band.

MicahZoltu (changes requested):
Most of the feedback can wait until a later stage of the process, though I encourage you to address it all.  The Formatting and incorrect sections stuff should be addressed before draft though.

On a non-editorial note, I don't believe this actually solves the problem you want it to solve because miners can still order transactions however they want and there already exist out-of-band miner payment systems so you aren't even going to protect users if this were released today.

The best solution to front-running like this is a commit-reveal scheme where the fraud provider would commit a hash of the fraud proof and then one or more blocks later reveal the preimage for the fraud proof.  This adds a bit of complexity, but since fraud proofs are likely being executed as bots that can largely be ignored and the increased gas costs should be relatively low compared to the actual fraud proof execution.

EIPS/eip-3332.md
JChanceHud: I added simple reasoning to the rationale section but I'm concerned it's still more motivation than rationale. Let me know if I'm missing it.

PR #3336, opened 2021-03-06
Add EIP draft for paged memory

Arachnid:


EIPS/eip-3336.md
```diff
@@ -0,0 +1,71 @@
...
+
+EVM implementations are encouraged to store the pagetable as an associative array (eg, hashtable or dict) mapping from page number to an array of bytes for the page.
+
+### Changes to memory expansion gas cost
+Presently, the total cost to extend the memory to `a` words long is `Cmem(a) = 3 * a + floor(a ** 2 / 512)`. If the memory is already `b` words long, the incremental cost is `Cmem(a) - Cmem(b)`. `a` is the number of words required to cover the range from memory address 0 to the last word that has been read or written by the EVM.
+
+Under this EIP, we define a new memory cost function, based on the number of allocated pages. This function is `Cmem'(p) = max(PAGE_BASE_COST * (p - 1) + floor(2 * (p - 1) ** 2), 0)`. As above, if the memory already contains `q` pages, the incremental cost is `Cmem'(p) - Cmem'(q)`.
+
```
MicahZoltu, line 41: This should mostly move to the Rationale section.  The specification should just say what the gas cost is, not history or *why* it is what it is.

EIPS/eip-3336.md
```diff
@@ -0,0 +1,71 @@
...
+type: Standards Track
+category: Core
+created: 2021-03-06
+---
+
+## Simple Summary
+Changes the memory model for the EVM to use pagination, which permits more flexible use of memory by compilers without unduly complicating implementations.
+
+## Abstract
+Presently, the EVM charges for memory as a linear array starting at address 0 and extending to the highest address that has been read from or written to. This suffices for simple uses, but means that compilers have to generate programs that use memory compactly, which leads to wasted gas with reallocation of memory elements, and makes some memory models such as separate heap and stack areas impractical. This EIP proposes changing to a page-based billing model, which adds minimal complexity to implementations, while providing for much more versatility in EVM programs.
```
MicahZoltu, line 16: The abstract should be a terse human readable version of the specification.  Currently, this discusses history which is more appropriate for the Motivation section.

(normally I would try to provide an alternative abstract, but I don't think I grok the change well enough to do so... sorry!)

EIPS/eip-3336.md
```diff
@@ -0,0 +1,71 @@
...
+author: Nick Johnson (@arachnid)
+discussions-to: https://ethereum-magicians.org/t/eips-3336-and-3337-improving-the-evms-memory-model/5482
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-03-06
+---
+
+## Simple Summary
+Changes the memory model for the EVM to use pagination, which permits more flexible use of memory by compilers without unduly complicating implementations.
```
MicahZoltu, line 13:
```suggestion
Changes the memory model for the EVM to use pagination.
```
The rest of the contents here are better suited for motivation (more flexibility) and rationale (less complicated than alternative proposals).

MicahZoltu (approved):
Recommend reviewing and applying the feedback, but none of it is a blocker for DRAFT (maybe a blocker for Review).

EIPS/eip-3336.md
```diff
@@ -0,0 +1,71 @@
...
+
+| Constant | Value |
+| - | - |
+| `FORK_BLOCK` | TBD |
+| `PAGE_BITS` | 10 |
+| `PAGE_BASE_COST` | 96 |
+
+For blocks where `block.number >= FORK_BLOCK`, the following changes apply.
+
+### Changes to memory alocation in EVM implementations
```
MicahZoltu, line 32:
```suggestion
### Changes to memory allocation in EVM implementations
```

MicahZoltu (changes requested):
I lied, there is one required change.

EIPS/eip-3336.md
Arachnid:
```suggestion
This EIP proposes changing EVM memory to a page-based allocation and gas billing model by dividing the memory address space into 1 kilobyte (32 word) pages, and charging gas on the basis of the number of pages allocated, rather than the address of the highest word referenced in memory. This facilitates more flexible use of memory than is currently possible with a linear model.
```

How about this?

EIPS/eip-3336.md
Arachnid: I think it's important for anyone trying to understand the new cost function to know what the old one is. Removing this would make the standard harder to understand, IMO.

Arachnid:
@MicahZoltu PTAL

EIPS/eip-3336.md
MicahZoltu: EIPs don't propose, they assert.  Also recommend removing the Rationale at the end and keeping focused on the specification.
```suggestion
Makes EVM memory a page-based allocation system and gas billing model by dividing the memory address space into 1 kilobyte (32 word) pages, and charging gas on the basis of the number of pages allocated, rather than the address of the highest word referenced in memory.
```

EIPS/eip-3336.md
MicahZoltu: I agree that it is very valuable information that should be present in the EIP.  I just disagree that it is appropriate for the Specification section.  The specification should focus purely on what one needs to implement in order to be compliant with this standard.  Explaining what a previous iteration defined doesn't help them be compliant with this standard, it helps them be compliant with a previous standard (not the reader's goal).

PR #3337, opened 2021-03-06
Create EIP for frame-pointer-relative memory addressing

Arachnid:


EIPS/eip-3337.md
```diff
@@ -0,0 +1,102 @@
...
+A new opcode `MSTOREFP` is introduced with value `0x5f`. This opcode acts in all ways identical to `MSTORE`, except that the value of the frame pointer is added to the address before storing data to memory. An attempt to store data to a negative address should be treated identically to an invalid opcode, consuming all gas and reverting the current execution context.
+
+## Rationale
+### Cost of new opcodes
+The cost of the new opcodes `MLOADFP` and `MSTOREFP` reflects the cost of `MLOAD` and `MSTORE`. They are generally equivalent in cost with the exception of an extra addition operation, which imposes negligible cost.
+
+The cost of the new opcodes `SETFP` and `GETFP` is based on other common low-cost opcodes such as `PUSH` and `POP`.
+
+### Absence of `MSTORE8FP`
+Whether or not to include an `MSTORE8FP` opcode depends on the frequency with which this will be used, and the desire to minimise the size of the instruction set and to conserve opcodes for future use. If the community believes this will be a substantial asset, the EIP can be modified to include it.
```
MicahZoltu, line 90: Write EIPs targeting a future reader, which means not including commentary on potential changes to the EIP.
```suggestion
`MSTORE8FP` opcode was not included because it is expected that it would be used infrequently, and there is a desire to minimise the size of the instruction set and to conserve opcodes for future use.
```

EIPS/eip-3337.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Backwards Compatibility
+This EIP exclusively introduces new opcodes, and as a result should not impact any existing programs unless they operate under the assumption that these opcodes are undefined, which we believe will not be the case.
+
+## Test Cases
+TBD
+
```
MicahZoltu, line 97:
```suggestion
```
Not a required section, so I recommend just leaving it out until such time in the future as you fill it in.  If you want to leave this in as a reminder to yourself, that is also acceptable.

MicahZoltu (changes requested):
Only the frontmatter change is required to merge as draft.

Arachnid:
@MicahZoltu PTAL

EIPS/eip-3337.md
```diff
@@ -0,0 +1,99 @@
...
+### Parameters
+
+| Constant | Value |
+| - | - |
+| `FORK_BLOCK` | TBD |
+
+For blocks where `block.number >= FORK_BLOCK`, the following changes apply.
+
+### Frame pointer
+A new EVM internal state variable called the "frame pointer" is introduced. This is a signed integer that starts at 0.
```
gcolvin, line 69: Probably needs to be specified as 256-bit twos-complement for the sake of the GETFP and SETFP operations.

MicahZoltu: You may want to bring this up in the discussions-to to make sure it gets seen.  I agree that this needs more specificity!

Side note: We probably don't need 256-bits, likely 32-bits would be way more than enough?

PR #3338, opened 2021-03-06
Creates EIP 3338

MicahZoltu:


MicahZoltu:
@lightclient alternative proposal since getting @axic's eyes on PRs is hard.

PR #3351, opened 2021-03-09
EIP Linting Bot

alita-moore:
Github actions eip linting bot

In this process I converted the [other](https://github.com/eip-automerger/automerger) python bot to typescript, added documentation, and I (mostly) untangled the originally nested logic.

next steps:
1) fix ERRORS as global, keep methods isolated, and make more robust against side-effect async errors
2) Build lots of tests (this repo needs them)

For now, though, the code achieves the goal of linting PRs and is able to be expanded much more easily


lightclient:
I can review this more in-depth, but I don't think the actual code should live in this repo. I think you should create a repository to house this code and only install it in the CI here to use.

alita-moore:
Yeah was thinking the same thing but @MicahZoltu I recommended letting it live here just for the sake of simplicity

MicahZoltu:
`.DSStore` are OSX files I believe, they shouldn't be committed.

MicahZoltu:
🤔 actually, to @lightclient's point, since this is TypeScript and GitHub actions only execute typescript, I think we *still* need a build process/CI to compile TS to JS?  Is it possible to run TS directly in GitHub actions?

.github/bot/src/lib/CheckApprovals.ts
```diff
@@ -0,0 +1,72 @@
...
+export const removeApproved = async (file: FileDiff) => {
+  const approvals = await getApprovals();
+  const authors = assertAuthors(file);
+
+  // remove approvals from authors present on base commit
+  const hasAuthorApproval = !!approvals.find((approver) =>
+    authors.includes(approver)
+  );
+
+  console.log(hasAuthorApproval);
```
alita-moore, line 15: remove

.github/bot/src/lib/Assertions.ts
```diff
@@ -0,0 +1,56 @@
...
+    pull_number: prNum
+  });
+
+  if (pr.merged) {
+    throw `PR ${prNum} is already merged; quitting`;
+  }
+
+  // if (pr.mergeable_state != "clean") {
+  //   throw `PR ${prNum} mergeable state is ${pr.mergeable_state}; quitting`;
+  // }
```
alita-moore, line 41: was stuck here, it's always dirty when submitting a PR but then for some reason is clean after. Any idea why?

.github/workflows/auto-merge-bot.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: EIP Auto-Merge Bot
+    steps:
+      # To use this repository's private action,
+      # you must check out the repository
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Node.js Enviornment
+        uses: actions/setup-node@v2
+        with:
+          node-version: '12'
```
alita-moore, line 15: node 12 is required (only supported option)

alita-moore:
> 🤔 actually, to @lightclient's point, since this is TypeScript and GitHub actions only execute typescript, I think we _still_ need a build process/CI to compile TS to JS? Is it possible to run TS directly in GitHub actions?

ncc takes care of this dist/index.js is a pure js compile version that runs in actions

MicahZoltu:
> ncc takes care of this; dist/index.js is a pure js compile version that runs in actions

`dist/index.js` is derived from the `.ts` files I assume though, and that step should really be handled by a CI system and not manual by people making changes to the repository.  That is, in general we should not be committing derived files.

I bet we can use `ts-node` to have the action actually execute as TypeScript (in-memory transpilation).  Check out https://github.com/TypeStrong/ts-node.  If you can pass CLI arguments to `node` then it should just be a matter of passing `-r ts-node/register` and then reference a `.ts` file as your entrypoint instead of `dist/index.js`.  If you cannot pass CLI arguments to `node` as part of the action then we would need to create a 2-line `index.js` file (not in `dist`, just in the source root) that does something like `require('ts-node').register({})` and then calls into a TS file (I'm not sure how to do that, ts-node docs should have more info on that).

alita-moore:
@MicahZoltu will do

alita-moore:
@MicahZoltu Okay, I can do that. But another option could just be to do a simple action `npm install && npm run build` which would just compile on action run time. Is there any particular benefit to compiling typescript directly? As I understand it (and I could be wrong) a big benefit of ncc is that it avoids issues with packages. Nevertheless, either way works for me, but it's another option I think.

.github/workflows/auto-merge-bot.yml
MicahZoltu: Hmm, node 14 isn't supported?  14 is current active version of NodeJS, so I'm surprised that only 12 is supported.  😢 
https://nodejs.org/en/about/releases/

.github/bot/README.md
```diff
@@ -0,0 +1,139 @@
...
+```
+npm run watch // re-builds after each save
+npm run build // re-builds
+npm run it    // runs the action using a basic development
+```
+### Troubleshooting
+- <i>I make changes but nothing changes when I `npm run it`</i>
+  - This repo requires that the changes be compiled in a pure js distribution `dist/index.js`
+  - Try running `npm run build` and see if that helps
+  - If it does, make sure to keep `npm run watch` running in the background while you make changes, it'll re-compile whenever changes are made and make life easier
```
MicahZoltu, line 61: Per recent conversation, can we get rid of this troubleshooting step now?

.github/bot/src/utils/Types.ts
```diff
@@ -0,0 +1,61 @@
...
+export type Github = ReturnType<typeof getOctokit>;
+const Github = getOctokit("fake");
+
+const CompareCommits = () => Github.repos.compareCommits().then(res => res.data);
+export type CompareCommits = UnPromisify<
+  ReturnType<typeof CompareCommits>
+>;
+
+type UnPromisify<T> = T extends Promise<infer U> ? U : T;
+const PR =  () => Github.pulls.get().then(res => res.data);
```
MicahZoltu, line 14: Nitpick: there is an extraneous space after `=` here.

.github/bot/README.md
```diff
@@ -0,0 +1,139 @@
...
+HEAD_SHA = <head sha of the PR>
+REPO_OWNER_NAME = <your login>
+REPO_NAME = EIPs
+GITHUB_REPOSITORY = <your login>/EIPs
+```
+5) `npm run it`
+
+### npm scripts
+
+This repo uses ncc; ncc compiles the code into a single pure js file, and that makes it easier for github to execute. But you have to build and commit your changes.
```
MicahZoltu, line 50: Can this be removed now that the compilation is part of the build script?

.github/bot/README.md
```diff
@@ -0,0 +1,139 @@
...
+          GITHUB_TOKEN: ${{ secrets.TOKEN }}
+```
+
+# Contributing
+## Development Enviornment Setup
+### Requirements
+1) node package manager (npm)
+2) Github Token 
+3) Forked Repo
+4) node version manager (nvm)
```
MicahZoltu, line 27: I don't see any references to `nvm` in here, can this requirement be removed (perhaps instead require `nodejs`)?

.github/workflows/auto-merge-bot.yml
alita-moore: Unfortunately yes, if you try anything but node 12 it tells you to use docker, and I doubt that's worth the hassle.

Although idk if you setup the node as 14 it may just use node 14.

Either way typescript us using es6 so should be too much of an issue

https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#creating-an-action-metadata-file

.github/bot/action.yml
```diff
@@ -0,0 +1,9 @@
+name: "Auto Merge EIP"
+description: "A bot that lints EIP edits, finds common errors, and can auto-merge"
+
+runs:
+  using: "node12"
+  main: "dist/index.js"
+branding:
+  icon: 'code'
```
MicahZoltu, line 8:
```suggestion
  icon: 'code'
```

.github/bot/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "dotenv": "^8.2.0",
+    "eslint": "^7.21.0",
+    "front-matter": "^4.0.2",
+    "node-fetch": "^2.6.1",
+    "typescript": "^4.1.5"
+  },
+  "devDependencies": {
+    "@types/node-fetch": "^2.5.8",
+    "husky": "^5.1.3",
+    "pretty-quick": "^3.1.0"
```
MicahZoltu, line 30: Use explicit versions (remove the `^`).

.github/bot/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "auto-merge-eip",
+  "version": "1.0.0",
+  "description": "auto merge eip",
+  "main": "index.js",
+  "scripts": {
+    "test": "NODE_ENV=testing jest __tests__",
+    "build": "ncc build ./src/index.ts",
+    "watch": "ncc build ./src/index.ts --watch",
```
MicahZoltu, line 9: What is `ncc` and why is it being used here instead of `npx tsc`?

.github/workflows/auto-merge-bot.yml
```diff
@@ -0,0 +1,22 @@
...
+          check-latest: false
+      - run: cd .github/bot && npm install && npm run build
```
MicahZoltu, line 17:
```suggestion
      - run: cd .github/bot && npm ci && npm run build
```
`ci` will install exactly what `package-lock.json` says rather than whatever npm happens to return for each package.

.github/bot/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "isolatedModules": true,
+    "moduleResolution": "node",
+    "noEmit": true,
+    "strict": true,
+    "baseUrl": "./",
+    "target": "es2015",
+    "paths": {
+    "src/*": ["src/*"]
```
MicahZoltu, line 12:
```suggestion
      "src/*": ["src/*"]
```

.github/workflows/auto-merge-bot.yml
MicahZoltu: Is it possible to move this install/build step into the action itself, rather than outside the action?

.github/bot/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "it": "node -r dotenv/config dist/index.js",
+    "prettier-changed": "prettier --write"
+  },
+  "author": "",
+  "license": "ISC",
```
MicahZoltu, line 14:
```suggestion
  "license": "CC0-1.0",
```
Everything in this repository should be CC0 licensed.

.github/bot/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@actions/core": "^1.2.6",
+    "@actions/github": "^4.0.0",
+    "@octokit/types": "^6.12.1",
+    "@typescript-eslint/parser": "^4.16.1",
+    "@zeit/ncc": "^0.22.3",
+    "dotenv": "^8.2.0",
+    "eslint": "^7.21.0",
+    "front-matter": "^4.0.2",
+    "node-fetch": "^2.6.1",
+    "typescript": "^4.1.5"
```
MicahZoltu, line 25: Several of these should be moved to `devDependencies`.  TypeScript, linters, etc. at the least.

.github/bot/package.json
MicahZoltu: I dug into it a bit, and I don't think it is necessary.  We can simply do `npx tsc` for the build/watch scripts, and we don't actually need a single file output I don't believe.  If we *do* need a single file output for some reason, TypeScript has the ability by using `outFile` instead of `outDir`.

.github/workflows/auto-merge-bot.yml
alita-moore: I'll try to removing ncc

MicahZoltu: Check out the PR I made against your repository, I *think* I may have already removed it (effectively), so you can just remove the dependency.

alita-moore: I think it'd be best to just do it in that PR (speaking of which, not sure how to make changes to your branch)

.github/bot/package.json
MicahZoltu: You would be surprised how often minor version bumps in NPM packages horribly break things.  😀

.github/workflows/auto-merge-bot.yml
MicahZoltu: My PR is against your fork, if you are generally OK with the changes, I recommend just merging it and then working on your fork like you have been (or undoing things you don't like from the PR).

.github/bot/package.json
alita-moore: cool that works!

lightclient:
It looks like this is a pretty straightforward way to use an action in a different repo without publishing it to the marketplace: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-public-action-in-a-subdirectory

MicahZoltu:
> 
> 
> It looks like this is a pretty straightforward way to use an action in a different repo without publishing it to the marketplace: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-public-action-in-a-subdirectory

Ooh, that is pretty compelling.  Maybe @MadeofTin can create a new repository in the `ethereum` GitHub organization that we can put it in?  I didn't realize you could just run from GitHub (though I shouldn't be surprised).

alita-moore:
@MicahZoltu i can also do that?

alita-moore:
Also just to be clear you want tests right? Will take me at least another week to build out fully is that okay @MicahZoltu?

Souptacular:
> > It looks like this is a pretty straightforward way to use an action in a different repo without publishing it to the marketplace: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-public-action-in-a-subdirectory
> 
> Ooh, that is pretty compelling. Maybe @MadeofTin can create a new repository in the `ethereum` GitHub organization that we can put it in? I didn't realize you could just run from GitHub (though I shouldn't be surprised).

I can create a new `ethereum` organization repo. Just DM me if needed.

alita-moore:
I'm going to set this up in a private repo, submit a simple connection PR to this repo, and then contact you @Souptacular for transferring ownership once that's all done.

alita-moore:
I moved code to its own repo so this is just instantiating that, it's the same code so `maybe` don't go live, but if you don't set the secret it'll just fail every time and probably worth merging now @MicahZoltu

MicahZoltu:
> but if you don't set the secret it'll just fail every time and probably worth merging now

Sadly, if it hard fails it'll block CI.  Fun fact, I actually cannot merge this PR *because* the bot is failing against this branch.  😬

.github/workflows/auto-merge-bot.yml
```diff
@@ -0,0 +1,19 @@
+on: [pull_request]
+
+jobs:
+  auto_merge_bot:
+    runs-on: ubuntu-latest
+    name: EIP Auto-Merge Bot
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Node.js Enviornment
```
MicahZoltu, line 10:
```suggestion
      - name: Setup Node.js Environment
```

alita-moore:
@MicahZoltu I have defaulted the action to pass on error, so despite it failing it won't cause problems. the only behavior that will be prevented from failing behavior is the merging of a given PR. I haven't been able to implement a solid work around yet, but I'm certain that some exist. One example would be to just remove the secret whenever you want it to fail / be manual. Then when you want to run it, add a secret, click on the task on a given PR, and then re-run.

MicahZoltu:
I believe only repository owners have the ability to change secrets, and the repository owners are not the editors.  😢

Does this help us at all?  It sounds like exactly what we want, but maybe I'm missing something.  https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

Also, I think we can make it so this particular GitHub action doesn't block merge on failure by adjusting the repository settings.  I believe only @MadeofTin can do that though.  Project page > Settings > Branches > Add Rule >  Require status checks to pass before merging > deselect this github action.

.github/workflows/auto-merge-bot.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: EIP Auto-Merge Bot
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Node.js Environment
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: auto-merge-bot
+        uses: alita-moore/EIP-Bot@master # Uses an action in the root directory
```
MicahZoltu, line 15: Can we have this point at a particular commit sha?  From a security standpoint, that would let me sleep better at night knowing that the code executing against this repository can't change without a PR to this repository.

MicahZoltu (changes requested):

.github/workflows/auto-merge-bot.yml
alita-moore: Yeah that's fine. I'll point it toward a versioning system!

alita-moore:
@MicahZoltu sounds good. I'll work on getting that all setup. Also is feasible to just have two workflows / secrets where one has the ability to merge and the other does not.

alita-moore:
@MicahZoltu

alita-moore:
version 1.1.0 of the EIP bot has the ability to merge commented out

alita-moore:
v1.1.1 also throws and error if anything except a merge occurs

.github/workflows/auto-merge-bot.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: EIP Auto-Merge Bot
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Node.js Environment
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: auto-merge-bot
+        uses: alita-moore/EIP-Bot@1.1.1 # v1.1.1 does not have the ability to merge and throws and error on non-merge like posting comment
```
MicahZoltu, line 15:
```suggestion
        uses: alita-moore/EIP-Bot@83def55233002790da6746afabe4724b8ee4f940 # tag 1.1.1
```
Does this work?  From a security standpoint, it protects us from a number of attack vectors including compromised keys, etc.  Not a critical thing, but if it works I have a slight preference for it.

.github/workflows/auto-merge-bot.yml
```diff
@@ -0,0 +1,19 @@
...
+        continue-on-error: true
+        id: auto-merge-bot
+        with:
+          GITHUB-TOKEN: ${{ secrets.TOKEN }}
```
MicahZoltu, line 19: If this isn't set, what will happen?  Will it fail and flag the build as failing (thus preventing merge), or will it fail and turn the build yellow but not block merge?  Perhaps this depends on settings on this repository (whether an action is required or not)?

MicahZoltu (commented):
Sorry for taking so long on reviewing this, been super busy lately.

@Souptacular I suspect we will need some changes to this repository when this is merged such as changing whether this status check is a required one or not:
image

Specifically, we would like to have this *not* block merging into `master` initially so we (editors) can monitor its behavior/output for a bit to make sure it all looks reasonable.

We may want to coordinate merging this to a time when all 3 of us are online so we can make sure we don't have downtime for the repository if it goes south.  Alternatively, you can be replaced by anyone else with admin access to this repository.

.github/workflows/auto-merge-bot.yml
alita-moore: Oh yeah this would work! Cool standpoint though about security. The more you know 😀

.github/workflows/auto-merge-bot.yml
alita-moore: It will fail but the ability to merge is independent of that / the auto-merge bot status can be made optional

Souptacular:
Happy to coordinate. Best way to do this is to make a private group (maybe Discord?) with parties that are necessary and who will be monitoring. I can make you admin temporarily as well, but will try to be available as much as possible to help.

youtacrsbyvars (commented):
###

PR #3357, opened 2021-03-10
EIP-1: Remove duplicative EIP reference style note, harmonize to preferred citation format

fulldecent:


EIPS/eip-1.md
```diff
@@ -207,6 +207,8 @@ EIPs may also have a `superseded-by` header indicating that an EIP has been rend
 
 References to other EIPs should follow the format `EIP-N` where `N` is the EIP number you are referring to.  Each EIP that is referenced in an EIP **MUST** be accompanied by a relative markdown link the first time it is referenced, and **MAY** be accompanied by a link on subsequent references.  The link **MUST** always be done via relative paths so that the links work in this GitHub repository, forks of this repository, the main EIPs site, mirrors of the main EIP site, etc.  For example, you would link to this EIP with `[EIP-1](./eip-1.md)`.
 
+An EIP **MUST NOT** make reference to another draft EIP. This is because something that is said about a draft EIP may no longer be accurate when it is finalized.
```
MicahZoltu, line 210: The rule we have been enforcing is that EIPs cannot reference EIPs with a state that is lower than them.  So a Review EIP cannot reference a Draft, but it could reference a Review/Last Call/Final EIP.  Functionally, this means that by the time something gets to final all of its dependencies are final, but it allows people to develop multiple dependent EIPs at the same time.

Any thoughts on how we could word this to make that a bit more clear?

PR #3360, opened 2021-03-10
EIP-2315: Simple Subroutines for the EVM

gcolvin:
This version is updated to specify rules for safe execution and amenability to static analysis.

### DIFF
--- a/EIPS/eip-2315.md
+++ b/EIPS/eip-2315.md
@@ -13,17 +13,17 @@ created: 2019-10-17
 
 This proposal introduces three opcodes to support subroutines: `BEGINSUB`, `JUMPSUB` and `RETURNSUB`.
 
-Safety and amenability to static analysis equivalent to EIP-615 can be provided by following a few simple rules, and validated with the provided algorithm.
+Safety and amenability to static analysis equivalent to EIP-615 can be ensured by following a few simple rules, and validated with the provided algorithm.
 
 ## Motivation
 
 The EVM does not provide subroutines as a primitive.  Instead, calls can be synthesized by fetching and pushing the current program counter on the data stack and jumping to the subroutine address; returns can be synthesized by contriving to get the return address back to the top of stack and jumping back to it.  Over the course of 30 years the computer industry struggled with this complexity and cost and settled in on opcodes to directly support subroutines.  These are provided in some form by most all physical and virtual machines going back at least 50 years.
 
-Seeking a simple, proven mechanism, we modeled our design on the [archetypal](http://www.ultratechnology.com/4th_1970.pdf) two-stack [Forth virtual machine](http://www.ultratechnology.com/4th_1970.pdf) of 1970.  The data stack is supplemented with a return stack to provide support for subroutines, as specified below.  The separate return stack ensures that the return address cannot be overwritten, and obviates any need to swap the return address past the arguments on the stack.  It also obviates the need for a dynamic jump to implement subroutine returns, allowing for deprecation of dynamic uses of JUMP and JUMPI.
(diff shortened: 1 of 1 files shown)


MicahZoltu:
Kicking the bot since this should have auto-merged.

EIPS/eip-2315.md
```diff
@@ -150,55 +307,11 @@ Bytecode: `0x6005565c5d5b60035e` (`PUSH1 0x05, JUMP, BEGINSUB, RETURNSUB, JUMPDE
...
-## Security Considerations
-
-These changes do introduce new flow control instructions, so any software which does static/dynamic analysis of evm-code
-needs to be modified accordingly. The `JUMPSUB` semantics are similar to `JUMP` (but jumping to a `BEGINSUB`), whereas the `RETURNSUB` instruction is different, since it can 'land' on any opcode (but the possible destinations can be statically inferred).
+## References
+Charles H.Moore, Geoffrey C. Leach. [FORTH - A Language for Interactive Computing](http://www.ultratechnology.com/4th_1970.pdf). 1970
+Philip J. Koopman, Jr. [Stack Computers: the new wave](https://users.ece.cmu.edu/~koopman/stack_computers/index.html). 1989
+Gavin Wood. [Ethereum:  A  Secure  Decentralized Generalized  Transaction  Ledger](https://ethereum.github.io/yellowpaper/paper.pdf). 2015-2021
+Martin Lundfall.  [EIP-2327: BEGINDATA Opcode](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2327.md). 2019
+Nick Johnson. [EIP-3337: Frame pointer support for memory load and store operations](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3337.md). 2021
```
MicahZoltu, line 315:
```suggestion
```

References aren't a valid EIP section, and external links should not be included in EIPs (think RFCs, which are technical documents that only contain internal links/references to other RFCs).

EIPS/eip-2315.md
```diff
@@ -150,55 +307,11 @@ Bytecode: `0x6005565c5d5b60035e` (`PUSH1 0x05, JUMP, BEGINSUB, RETURNSUB, JUMPDE
...
 
-**Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).**
+**Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).**
```
MicahZoltu, line 317:
```suggestion
## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```

MicahZoltu:
*grumble* no idea why the bot isn't picking this up.  Probably the same race condition we have had before...

holiman:
This doc is afaict identical to the previous iteration of EIP-2315, specification-wise. The difference is that it has an extra clause about Validity within the Security section. 
The EIP does not specify whether an EVM should execute the validity section, so I assume the evm should be oblivious to whether code is 'valid' or not? 
So I don't understand who the intended validator is. Who should check if code is 'valid' and for what purpose?

EIPS/eip-2315.md
gcolvin: Hmm.  I went through hell last week in part because other devs were unaware of the content of the first two references, and wouldn't take my word for it.   The last tree references are to other EIPs.  (Well, Yellow Paper amounts to EIP-0.)

gcolvin:
@holiman  Yes, little has changed in the spec.  The main substantive complaint I heard last week was that we did not forbid "jumping into subroutines", and in general did not provide the sort of guarantees provided by EIP-615.

So I give here the rules for ensuring that level of safety and show how to enforce them - without imposing syntactic constraints, only semantic ones.

I think whether and when to enforce these rules requires serious discussion.  Probably no code should be deployed to the chain that fails to validate.  Note that by being completely backwards compatible no versioning is needed.

EIPS/eip-2315.md
gcolvin: For some reason I can't just commit the suggestion, but have fixed my working copy.  Thanks.

gcolvin:
@holiman Just talked with @AlexeyAkhunov.  One problem with moving directly to validated deployment is that it would immediately break Solidity and other compilers that are not prepared for it.  In particular, Solidity uses dynamic jumps to return from subroutines.  He suggested that in the first upgrade to include this we (at most) enforce static jumps only for JUMPSUB, giving time for compilers and tools to prepare for a future upgrade that actually enforces all of the rules.

gcolvin:
I've merged changes in response to review so far.

EIPS/eip-2315.md
gcolvin: I've removed the two external references, but retained links in the EIP text.

PR #3382, opened 2021-03-14
Hardcoded Gas Limit

PhABC:
.

### DIFF
--- a/EIPS/eip-3382.md
+++ b/EIPS/eip-3382.md
@@ -2,7 +2,7 @@
 eip: 3382
 title: Hardcoded Block Gas Limit
 author: Philippe Castonguay (@PhABC)
-discussions-to: xxxx
+discussions-to: https://ethereum-magicians.org/t/eip-3382-hardcoded-gas-limit
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-hard-gas-target.md
+++ b/EIPS/eip-hard-gas-target.md
@@ -24,19 +24,15 @@ Both Ethereum's Proof of Work and Proof of Stake designs assume that block produ
 ## Specification
 Refer to `gasLimit` as `gasTarget` post EIP-1559.
 
-#### Consensus
+#### Added Consensus Constraint
 
-As of `FORK_BLOCK_NUMBER`, change the headers gas limit validity check such that `header.gasUsed` **MUST** be **smaller** than `BLOCK_GAS_TARGET * ELASTICITY_MULTIPLIER` where `BLOCK_GAS_LIMIT` is a hard-coded constant set to `12,500,000` and `ELASTICITY_MULTIPLIER` is `2` as per [EIP-1559 specifications](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md#specification). In addition, the `header.gasLimit` **MUST** be equal to `BLOCK_GAS_LIMIT`.
-
-#### EVM
-
-The `GASLIMIT` opcode (0x45) should return the constant `12,500,000` as of `FORK_BLOCK_NUMBER`.
+As of `FORK_BLOCK_NUMBER`, the `header.gasLimit` **MUST** be equal to `BLOCK_GAS_LIMIT`, where `BLOCK_GAS_LIMIT` is a hardcoded constant set to `12,500,000`.
 
 ## Rationale
 
-#### Validating GasLimit in Block Headers
+#### Keeping gasLimit in Block Headers
 
-While it would be possible to remove the `gasLimit` field from block headers, it would change the data structure to be hashed, which could lead to unintended consequences. It is therefore easier to leave the gasLimit as part of block headers. However, since this in-header value is set by the block producer, it should be validated against `BLOCK_GAS_LIMIT` in case it is used by off-chain tools or contracts.
+While it would be possible to remove the `gasLimit` field from block headers, it would change the data structure to be hashed, which could lead to unintended consequences. It is therefore easier to leave the gasLimit as part of block headers. 
 
 #### Chosen Gas Limit
 
@@ -47,7 +43,7 @@ The `12,500,000` value is being proposed as it's the current block gas limit as 
 This EIP is backward compatible.
 
 ## Security Considerations
-Rapid changes to the gas limit will likely be more difficult to execute, which could be problematic if an urgent situation arise that required changing the gas limitt.
+Rapid changes to the gas limit will likely be more difficult to execute, which could be problematic if an urgent situation arise that required changing the gas limit.
 
 
 ## Copyright
--- a/EIPS/eip-hard-gas-target.md
+++ b/EIPS/eip-hard-gas-target.md
@@ -7,7 +7,6 @@ status: Draft
 type: Standards Track
 category: Core
 created: 2020-03-13
-
 ---
 
 ## Simple Summary
--- a/EIPS/eip-hard-gas-target.md
+++ b/EIPS/eip-hard-gas-target.md
@@ -45,7 +45,6 @@ This EIP is backward compatible.
 ## Security Considerations
 Rapid changes to the gas limit will likely be more difficult to execute, which could be problematic if an urgent situation arise that required changing the gas limit.
 
-
 ## Copyright
 
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/EIPS/eip-hard-gas-target.md
+++ b/EIPS/eip-hard-gas-target.md
@@ -24,13 +24,9 @@ Both Ethereum's Proof of Work and Proof of Stake designs assume that block produ
 ## Specification
 Refer to `gasLimit` as `gasTarget` post EIP-1559.
 
-#### Block Header
-
-As of `FORK_BLOCK_NUMBER`, remove `gasLimit` field from block headers.
-
 #### Consensus
 
-As of `FORK_BLOCK_NUMBER`, change the headers gas limit validity check such that `header.gasUsed` **MUST** be smaller than `BLOCK_GAS_TARGET * ELASTICITY_MULTIPLIER` where `BLOCK_GAS_LIMIT` is a hard-coded constant set to `12,500,000`, not the current gas limit included in block headers and `ELASTICITY_MULTIPLIER` is `2` as per [EIP-1559 specifications](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md#specification).
+As of `FORK_BLOCK_NUMBER`, change the headers gas limit validity check such that `header.gasUsed` **MUST** be **smaller** than `BLOCK_GAS_TARGET * ELASTICITY_MULTIPLIER` where `BLOCK_GAS_LIMIT` is a hard-coded constant set to `12,500,000` and `ELASTICITY_MULTIPLIER` is `2` as per [EIP-1559 specifications](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md#specification). In addition, the `header.gasLimit` **MUST** be equal to `BLOCK_GAS_LIMIT`.
 
 #### EVM
 
@@ -38,9 +34,9 @@ The `GASLIMIT` opcode (0x45) should return the constant `12,500,000` as of `FORK
 
 ## Rationale
 
-#### Removing the gasLimit from the header
+#### Validating GasLimit in Block Headers
 
-The gas limit was only present in block headers as it was defined by the block producers. Since this value is now a constant accessible to all nodes, there is little benefit to having the gas limit in the block headers. This would cause a change to the data structure to be hashed for block headers, but EIP-1559 already [imposes a change](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md#block-hash-changing) to the data structure.
+While it would be possible to remove the `gasLimit` field from block headers, it would change the data structure to be hashed, which could lead to unintended consequences. It is therefore easier to leave the gasLimit as part of block headers. However, since this in-header value is set by the block producer, it should be validated against `BLOCK_GAS_LIMIT` in case it is used by off-chain tools or contracts.
 
 #### Chosen Gas Limit
 
@@ -48,7 +44,7 @@ The `12,500,000` value is being proposed as it's the current block gas limit as 
 
 ## Backwards Compatibility
 
-Removing the gas limit value from block headers may break some analytic tools that rely on this field and contracts that do on-chain block verification. EIP-1559 likely already breaks the latter. 
+This EIP is backward compatible.
 
 ## Security Considerations
 Rapid changes to the gas limit will likely be more difficult to execute, which could be problematic if an urgent situation arise that required changing the gas limitt.
--- a/EIPS/eip-hard-gas-target.md
+++ b/EIPS/eip-hard-gas-target.md
@@ -12,11 +12,11 @@ created: 2020-03-13
 
 ## Simple Summary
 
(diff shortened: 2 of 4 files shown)


EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
+---
+eip: xxxx
```
lightclient, line 2:
```suggestion
eip: 3382
```

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+title: Hardcoded Block Gas Target
+author: Philippe Castonguay (@PhABC)
+discussions-to: xxxx
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-03-13
+
```
lightclient, line 10:
```suggestion
```

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+---
+
+## Simple Summary
+
+Hardcode the block gas target (previously known as *block gas limit* before [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md)) to `12,500,000` gas per block.
```
lightclient, line 15: Since this EIP does not require 1559, I recommend using non-1559 terminology "gas limit".

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Abstract
+
+Ethereum's block gas target is currently dictated by block producers and is not enforced when validating blocks in clients. This EIP proposes to hardcode the block gas target to `12,500,000`.
+
+## Motivation
+
+Both Ethereum's Proof of Work and Proof of Stake designs assume that block producers are financially rational, but does not assume block producers to be benevolent. There is one exception however, and it is when block producers choose the gas targe of a block (previously known as block gas limit), where it is assumed that block producers care about the long term health and decentralisation of the chain. Indeed, the block gas target is one of the only parameters in Ethereum that is not dictated by node consensus, but instead is chosen by block producers. This decision was initially made to allow urgent changes in the block gas target if necessary. Both drastically increasing or decreasing this parameter could have serious consequences that may not be desired. It is therefore a critical parameter that should require node consensus to avoid any sudden harmful change imposed by a small number of actors on the rest of the network.
```
lightclient, line 23: I'll need to check later if the standard easy of referring to POW and POS is hyphenated or not.

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Specification
+This EIP assumes that EIP-1559 is implemented and blocks' `gasLimit` was replaced with `gasTarget`.
```
lightclient, line 26: If so, I recommend removing this and requiring 1559 in the header.

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### Block Header
+
+As of `FORK_BLOCK_NUMBER`, remove `gas_target` field from block headers.
```
lightclient, line 30: A nit, it is probably easier to keep the gas limit in the block and ignore than remove. But since we're adding BASEFEE already, the point may be moot.

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### Consensus
```
lightclient, line 32: This is already implied by 1559.

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+
+As of `FORK_BLOCK_NUMBER`, change the headers gas target validity check such that `header.gasUsed` **MUST** be smaller than `BLOCK_GAS_TARGET * ELASTICITY_MULTIPLIER` where `BLOCK_GAS_LIMIT` is a hard-coded constant set to `12,500,000`, not the current gas target included in block headers and `ELASTICITY_MULTIPLIER` is `2` as per [EIP-1559 specifications](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md#specification).
+
+#### EVM
```
lightclient, line 36: I don't think this needs to be specified as it is implied.

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,60 @@
...
+
+The gas target was only present in block headers as it was defined by the block producers. Since this value is now a constant accessible to all nodes, there is little benefit to having the gas target in the block headers. 
+
+#### Gas Target Selected
+
+The `12,500,000` value is being proposed as it's the current block gas limit as of time of writing this EIP. The actual amount could be altered with a subsequent EIP to avoid deviating from the core intent of this EIP.
+
+## Backwards Compatibility
+
+Removing the gas target value from block headers may break some analytic tools that rely on this field.
```
lightclient, line 52: Will also break on-chains contracts that verify headers.

lightclient (changes requested):

EIPS/eip-hard-gas-target.md
PhABC: Good point. I'll leave a note about EIP-1559 renaming

EIPS/eip-hard-gas-target.md
PhABC: Fair point, though wouldn't any change to the block structure result to this?

EIPS/eip-hard-gas-target.md
PhABC: Is it? I believe gas_target is still part of block headers with EIP-1559.

EIPS/eip-hard-gas-target.md
PhABC: Indeed, I believe EIP-1559 already forces a change to the data structure to be hashed for block headers.

EIPS/eip-hard-gas-target.md
PhABC: I'll leave it to be explicit about the `<` instead of `<=`, which could cause a fork.

EIPS/eip-hard-gas-target.md
MicahZoltu: 1559 adds an item on the end, so if you have code that hashes the block and/or plucks items out of it by offset then 1559 won't break anything.  However, if we pull gas limit out (rather than just ignore it) any code that plucks data *after* the gas limit will break.

EIPS/eip-hard-gas-target.md
MicahZoltu: No, see comment above.

EIPS/eip-hard-gas-target.md
PhABC: Fair, then for simplicity I'll remove this aspect from the EIP. Even though it would become mainly a cosmetic field, we probably need to add a check in the consensus that the gasLimit provided in the header is equal to the hardcoded gas limit, in case people use the former. Would you both agree?

KaycoinsInc (approved):

EIPS/eip-hard-gas-target.md
lightclient: I'm referring to the consensus section here. If the gas limit is required to be 12.5mm per consensus then the gas used will be appropriately constrained pre and post-1559.

PhABC: Ah, I thought your original comment was referring to `As of `FORK_BLOCK_NUMBER`, remove `gas_target` field from block headers.`

However, EIP-1559 does not imply the hardcoded gasLimit is part of the consensus. Current EIP-1559 spec uses the gas_target from a block's header, not this proposed hardcoded constant. As per previous comments you and Micah made, the gas_target provided in headers will remain, so it's important that the consensus algorithm implementations are changed to using the hard-coded gasLimit, not the header's gas limit.

lightclient: Ah yes, if the limit was removed from the header it would need to be specified. Since the limit will stay in the header, why not a consensus rule that requires `block.gas_limit == 12_500_000`? The rules for EIP-1559 should stay the same in that case?

EIPS/eip-hard-gas-target.md
lightclient: I was actually referring to the EVM section here. I realize now that if the header *didn't* have `gas_limit` specified, this would be needed. However, if the header does include the `gas_limit` the result of `GASLIMIT` should be implied.

lightclient: > I'll leave it to be explicit about the < instead of <=, which could cause a fork if not implemented the same.

What is the rationale for `<` vs. `<=`? I believe today `gas_used` must be less than or equal to `gas_limit`, seems like it is better to maintain that unless there is a particular reason to change?

EIPS/eip-hard-gas-target.md
PhABC: Good point, i'll remove the redundant stuff.

EIPS/eip-hard-gas-target.md
PhABC: Searched and found a few EIPs that use these terms the same way, e.g. https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1011.md and https://eips.ethereum.org/EIPS/eip-1057#motivation while others do used an hyphenated version like https://eips.ethereum.org/EIPS/eip-3368 and  https://eips.ethereum.org/EIPS/eip-1057's title.

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,51 @@
+---
+eip: 3382
+title: Hardcoded Block Gas Limit
+author: Philippe Castonguay (@PhABC)
+discussions-to: xxxx
```
lightclient, line 5: Please add a discussion link when you're ready for this to be merged.

EIPS/eip-hard-gas-target.md
```diff
@@ -0,0 +1,51 @@
...
+The `12,500,000` value is being proposed as it's the current block gas limit as of time of writing this EIP. The actual amount could be altered with a subsequent EIP to avoid deviating from the core intent of this EIP.
+
+## Backwards Compatibility
+
+This EIP is backward compatible.
+
+## Security Considerations
+Rapid changes to the gas limit will likely be more difficult to execute, which could be problematic if an urgent situation arise that required changing the gas limit.
+
+
```
lightclient, line 48:
```suggestion
```

lightclient (approved):
EIP is ready to be merged once the discussion link is filled in.

EIPS/eip-hard-gas-target.md
lightclient: Doesn't appear to be a strong consensus on one version or the other. Please use your preferred. If certain style is decided on the future, all EIPs will be updated accordingly.

```console
$ grep -ril "proof of work" EIPS | wc -l
9
$ grep -ril "proof-of-work" EIPS | wc -l
13
$ grep -ril "proof of stake" EIPS | wc -l
5
$ grep -ril "proof-of-stake" EIPS | wc -l
8
```

EIPS/eip-hard-gas-target.md
PhABC: Added

EIPS/eip-hard-gas-target.md
PhABC: Thanks for doing the search, looks pretty evenly distributed!

EIPS/eip-3382.md
```diff
@@ -0,0 +1,50 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-3382-hardcoded-gas-limit
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-03-13
+---
+
+## Simple Summary
+
+Hardcode the block gas limit (known as *block gas target* under [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md)) to `12,500,000` gas per block.
```
MicahZoltu, line 14: Links to external EIPs should be relative (`[EIP-1559](./eip-1559.md)`).  That being said, this is not really a detail relevant to the simple summary.  The simple summary should be just a short one liner that says what the EIP is.
```suggestion
Hardcode the block gas target to `12,500,000` gas per block.
```

EIPS/eip-3382.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Abstract
+
+Ethereum's block gas limit is currently dictated by block producers and is not enforced when validating blocks in clients. This EIP proposes to hardcode the block gas limit to `12,500,000`.
```
MicahZoltu, line 18: Abstract should be a human readable and terse version of the specification.  The first sentence you have talks about history, which isn't appropriate for the specification/abstract.  The second sentence is a tad less clear than I think would be ideal.

```suggestion
Updates the block validation rules such that a block is invalid if the `gas_target` header field is not equal to 12,500,000.
```

EIPS/eip-3382.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Motivation
+
+Both Ethereum's Proof of Work and Proof of Stake designs assume that block producers are financially rational, but does not assume block producers to be benevolent. There is one exception however, and it is when block producers choose the gas limit of a block where it is assumed that block producers care about the long term health and decentralisation of the chain. Indeed, the block gas limit is one of the only parameters in Ethereum that is not dictated by node consensus, but instead is chosen by block producers. This decision was initially made to allow urgent changes in the block gas limit if necessary. Both drastically increasing or decreasing this parameter could have serious consequences that may not be desired. It is therefore a critical parameter that should require node consensus to avoid any sudden harmful change imposed by a small number of actors on the rest of the network.
+
+## Specification
+Refer to `gasLimit` as `gasTarget` post EIP-1559.
+
```
MicahZoltu, line 26: Write this either under the assumption of EIP-1559 or not.  We can always change the wording one way or the other at the last minute should plans change since that would be a non-normative change.  I recommend writing this under the assumption of EIP-1559, since that seems very likely to land in London.
```suggestion
```

EIPS/eip-3382.md
```diff
@@ -0,0 +1,50 @@
...
+#### Added Consensus Constraint
+
+As of `FORK_BLOCK_NUMBER`, the `header.gasLimit` **MUST** be equal to `BLOCK_GAS_LIMIT`, where `BLOCK_GAS_LIMIT` is a hardcoded constant set to `12,500,000`.
```
MicahZoltu, line 29: Instead of `header.gasLimit`, I recommend referring to it by its position in the header tuple.  `gasLimit` is what most people call it, but from a consensus point of view it is the `n`th (I forget the number) item in the block header array.

MicahZoltu: See note above about using gas **target** throughout this document, including here.

MicahZoltu (approved):
Some changes requested, but nothing blocking merge as draft.

EIPS/eip-3382.md
PhABC: I initially used gasTarget across the entire EIP, however as @lightclient pointed out, these two EIPs are independant so it might be clearer if we use the current term.

EIPS/eip-3382.md
PhABC: See my reply above, but will add tuple number

EIPS/eip-3382.md
MicahZoltu: I agree with @lightclient that this EIP is independent of 1559 and should be written as such.  However, you should write a single EIP as an assertion of a change, without conditionals, which means you have to choose one wording or the other.  In this case, I think EIP-1559 is more likely to be included in London than not, so the "safer" bet is to use "gas target" wording as that minimizes the chances that you will need to update the EIP later should the environment change.

PR #3397, opened 2021-03-15
add impl for eip3074

lightclient:
^

### DIFF
--- a/EIPS/eip-3074.md
+++ b/EIPS/eip-3074.md
@@ -177,7 +177,7 @@ TODO
 
 ## Implementation
 
-TODO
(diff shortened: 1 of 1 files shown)


EIPS/eip-3074.md
```diff
@@ -177,7 +177,7 @@ TODO
 
 ## Implementation
 
-TODO
+https://github.com/quilt/go-ethereum/tree/eip-3074
```
MicahZoltu, line 181: `Implementation` section is no longer an EIP section.  Also, external links to third party repositories should not be used.  Especially transient branches.

Remember: EIPs are technical specifications and not an appropriate place for coordination.

PR #3405, opened 2021-03-16
Update eip-3368.md

bitsbetrippin:
Added as discussed OPTIONAL clause in Rational. This EIP has been in response of existential risk of growing GPU Mining Scope and related to GPU Elastic network movement during specific economic scenarios.

 - The build passes.

### DIFF
--- a/EIPS/eip-3368.md
+++ b/EIPS/eip-3368.md
@@ -38,7 +38,7 @@ elif block.number > TRANSITION_START_BLOCK_NUMBER:
 ```
 
 ## Rationale
-2 years was chosen because it gives miners sufficient time to find alternative uses for their hardware and/or get their hardware back out onto the open market (e.g., in the form of gaming GPUs) without flooding the market suddenly.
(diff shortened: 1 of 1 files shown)


EIPS/eip-3368.md
```diff
@@ -38,7 +38,7 @@ elif block.number > TRANSITION_START_BLOCK_NUMBER:
 ```
 
 ## Rationale
-2 years was chosen because it gives miners sufficient time to find alternative uses for their hardware and/or get their hardware back out onto the open market (e.g., in the form of gaming GPUs) without flooding the market suddenly.
+2 years was chosen because it gives miners sufficient time to find alternative uses for their hardware and/or get their hardware back out onto the open market (e.g., in the form of gaming GPUs) without flooding the market suddenly. This proposal should ONLY be considered as a last resort as something we have in our pocket should the "network need to attract hashing power quickly and then bleed it off over time" rather than "something that is scheduled to be included in X hard fork" ; Recomendation to have in a fast track state, but NOT deployed to mainnet unless needed.
```
MicahZoltu, line 41: This isn't appropriate content for an EIP, it should live in the discussions-to link or some other document.  EIPs should be technical specifications and have nothing to do with the hardfork coordination process or inclusion process.

PR #3407, opened 2021-03-17
EIP 2981 improvements

seibelj:
@VexyCats 

- Fixed issues in precision of the percentages. In some places 4 decimals were used, but we want 5 (0.00001) was the intention.
- tokenId was added to functions but inconsistently
- You can't emit an event on the NFT contract directly, you must wrap it in a function. Having this function requires adding it to the interface
- Interface ERC165 calculation was incorrect (didn't update it for the uint256 tokenId when it was added) plus I had to add in the new function for emitting the event
- Added `_tokenPaid` field to the `RoyaltiesReceived` event to support ERC20 payments
- Removed the 4th index in the event - events can only have 3 indexed parameters. ERC20 and ERC721 both don't index the amount, so not indexing the `_tokenPaid` and the `_amount` is consistent with the standards
- Updated and cleaned up the copy a bit

### DIFF
--- a/EIPS/eip-2981.md
+++ b/EIPS/eip-2981.md
@@ -68,13 +68,18 @@ interface IERC2981 is ERC165 {
     /**
 
     /**
-     *      @notice Called to return both the creator's address and the royalty percentage - this would be the main function called by marketplaces unless they specifically        *       need just the royaltyAmount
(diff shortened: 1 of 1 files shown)


eip-automerger:

 - EIP 2981 requires approval from one of (@jamesmorgan, @blmalone, @vexycats)

VexyCats (approved):
Great changes and finds, thanks!

PR #3416, opened 2021-03-18
EIP-3416: Median gas premium

hexzorro:
This proposal aims to reduce the gas price volatility using a `weighted median gas premium` for gas pricing and to maintain a stable structure of incentives for miners.


jokerksa12:
You can help me?🥵🚀

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
+---
+eip: 3416
+title: Median gas premium and miner fees
+author: HexZorro (@hexzorro), strongly based on EIP-1559 circa Aug-2020.
```
MicahZoltu, line 4:
```suggestion
author: HexZorro (@hexzorro)
```

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Simple Summary
+
+A transaction pricing mechanism that includes a fixed-per-block network fee that is *not*-burned plus a *median* premium fee per transaction. Also, dynamically expands/contracts block sizes to deal with transient congestion. We target the following goals:
+
+* Gas prices spikes are mathematically smoothed out.
+* Maintain gas price preference, i.e. transaction senders willing to pay extra in fees will be rewarded with early preferential inclusion in the blocks, because the miners want to maximize their profits and include transactions with higher fee caps first to maximize the median.
+* Final gas price paid by the sender is, most of the time, smaller than the maximum gas price specified by sender. 
+* Gas pricing is more robust to sender manipulation or miner manipulation.
+* All fees go to miners.
```
MicahZoltu, line 20: Simple summary should be about a sentence long.  In terms of length/content, imagine an email subject line, GitHub pull request title, or forum topic.

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Abstract
+
+There is a base fee per gas in protocol, which can move up or down by a maximum of 1/8 in each block. The base fee per gas is adjusted by the protocol to target an average gas usage per block instead of an absolute gas usage per block.  The base fee is increased when blocks are over the gas limit target and decreases when blocks are under the gas limit target. The base fee per gas is *not burned*. Transaction senders specify their fees by providing *only one value*:
```
MicahZoltu, line 24:
```suggestion
There is a base fee per gas in protocol, which can move up or down by a maximum of 1/8 in each block. The base fee per gas is adjusted by the protocol to target an average gas usage per block instead of an absolute gas usage per block.  The base fee is increased when blocks are over the gas limit target and decreases when blocks are under the gas limit target. Transaction senders specify their fees by providing *only one value*:
```
Generally speaking, EIPs should focus on what they do rather than what they don't do since the former set is small and the latter set is infinite.

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+* [Go-ethereum implementation](https://github.com/vulcanize/go-ethereum-EIP3416)
+* [Implementation-specific Magicians thread](https://ethereum-magicians.org/t/eip-1559-go-etheruem-implementation/3918)
+* [First and second-price auctions and improved transaction-fee markets](https://ethresear.ch/t/first-and-second-price-auctions-and-improved-transaction-fee-markets/2410)
+* [The Challenges of Bitcoin Transaction Fee Estimation](https://blog.bitgo.com/the-challenges-of-bitcoin-transaction-fee-estimation-e47a64a61c72)
+* [Gas Now](https://www.gasnow.org/)
+* [Five arguments against fee burning in EIP-1559](https://ethereum-magicians.org/t/five-arguments-against-fee-burning-in-eip-1559/5184)
+* [Robust statistics - Median](https://en.wikipedia.org/wiki/Robust_statistics#Examples)
+* [Evidence of Mempool Manipulation on Black Thursday: Hammerbots, Mempool Compression, and Spontaneous Stuck Transactions](https://www.blocknative.com/blog/mempool-forensics)
+* [Ethereum Network Utilization](https://etherscan.io/chart/networkutilization)
+
```
MicahZoltu, line 111:
```suggestion
```
Resources is not a valid EIP section.  Also, external links should not be used in EIPs.  Remember, EIPs are technical specifications and not the right place to make arguments over a particular EIP, compare it with other EIPs, etc.  The EIP should be incredibly tightly focused on the technical specification.

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+## Backwards Compatibility
+
+The backward compatibility is very straightforward because there are no new fields added to the transactions. Pricing of the gas per block is still fast to compute but a little more complex. We split the EIP3416 upgrade into two phases with a transition period during which both legacy and EIP3416 transactions can be accepted so that compatibility with wallets and other ETH-adjacent software is maintained while their maintainers have time to upgrade to using the new transaction type. During this transition period legacy transactions are accepted and processed identically to the current implementation, with the only difference being that the amount of gas (gas limit) dedicated to processing legacy transactions is calculated as above and decreases over this period.
+
+## Test Cases
+
```
MicahZoltu, line 90:
```suggestion

TBD

```

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+## Security Considerations
+
+* Senders cannot manipulate the minimum fee because the minimum `BASE_FEE` is controlled by the miners with small increments or decrements on each new block proposed.
+* Above the `BASE_FEE` the senders have a very limited ability to manipulate and lower the final gas price they pay because they have to move the weighted median close to `BASE_FEE` and, as we know, this is a very robust statistic.
+* Miners have a very limited ability to manipulate and raise the final gas price paid by senders above `BASE_FEE` because to influence the final gas price they have to stuff fake transactions beyond the top 5% of the blocks. In average and currently, only the top 3% of the block is empty, so to fill-up 5% of the block they need to start dropping profitable transactions to reach 5%. Only beyond 5% of the top block gas they can start moving the median a little and the median is still a very robust statistic, not liable to being easily manipulated.
```
MicahZoltu, line 95: Miners can drop the bottom 10% of lowest paying transactions and replace them with transactions that that pay an exorbitant fee to raise the median.  This has potential to fully offset the losses from those lowest 10% of transactions.

MicahZoltu, line 95: Since miners receive the `base_fee`, they have very little reason to *not* fill the block with their own transactions (past the 50% full target).

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Backwards Compatibility
+
+The backward compatibility is very straightforward because there are no new fields added to the transactions. Pricing of the gas per block is still fast to compute but a little more complex. We split the EIP3416 upgrade into two phases with a transition period during which both legacy and EIP3416 transactions can be accepted so that compatibility with wallets and other ETH-adjacent software is maintained while their maintainers have time to upgrade to using the new transaction type. During this transition period legacy transactions are accepted and processed identically to the current implementation, with the only difference being that the amount of gas (gas limit) dedicated to processing legacy transactions is calculated as above and decreases over this period.
```
MicahZoltu, line 87: EIPs should be referenced as `[EIP-3416](./eip-3416.md)`.  When referring to itself, instead say "We split this EIP into two phases.."

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+  * Let `GASUSED` be the gas used during the transaction execution/state transition.
+  * The `tx.origin` initially pays `GASPRICE * tx.gasLimit`, and gets refunded `GASPRICE * (tx.gasLimit - GASUSED)`
+  * The `block.coinbase` gains `GASPRICE * GASUSED`.
+
+The miners can still use a `greedy` strategy to include new transactions in the proposed blocks by adding the transactions ordered by larger `FEE_CAP` first. This is similar to how current blocks are filled, and is a consequence of `FEE_CAP` and `GAS_PREMIUM` being a positive linear function of each other.
+
+## Rationale 
+
+The rationale behind the premium being 50% of (fee cap - base fee) is that at any given point the average network sender has an average fee cap, and we assume that between base fee and fee cap the sender has no specific preference, as long as the transaction is included in some block. Then, the sender is happy with a median premium among this uniform range. Another justification can be that the user also knows that this new version of the pricing protocol for the complete block uses a median, then is fair for him to apply a median within his preferential range, assuming an uniform sampling there.
```
MicahZoltu, line 83: Rationale should be included for why 5% instead of any other number.  Similarly, it should mention why `8` as the base fee max change denominator.

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+  * If `FEE_CAP < BASE_FEE` then the transaction is considered invalid and cannot be included in the current block, but might be included in future blocks.
+* During transaction execution, for EIP3416 transactions we calculate the cost to the `tx.origin` and the gain to the `block.coinbase` as follows:
+  * Set `GASPRICE = BASE_FEE + median((tx_i.gasPrice - BASE_FEE) / 2)` among all transactions `tx_i` included in the same block, *weighted by gas consumed* and not including the top 5% of outlier gas price in calculation. By weighted median without 5% of the upper-side outliers, we mean that each gas unit spent is ordered according to the corresponding transaction by `BASE_FEE + tx.gasPrice / 2` and then the value chosen will be the one separating the lower 95% in two parts. The 5%, or similar, removal is to give extra robustness against miner manipulation, because as current network utilization has been around 97% for the last 6 months the miners can include their own transactions on the empty 3% to try to manipulate and increase the median price (even this manipulation effect will be very small on the final price).
```
MicahZoltu, line 74: The latter part of this should be moved to the rationale section, not the specification.

EIPS/eip-3416.md
MicahZoltu, line 87: This says the EIP is split into two phases, but the specification doesn't appear to have any mention of the phases.

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+* `BASE_FEE_CHANGE`: `CURRENT_BLOCK.parent.BASE_FEE`. The amount that the BASE_FEE will increase in `CURRENT_BLOCK` relative to `CURRENT_BLOCK.parent`.
+
+### Process
+
+* At `block.number == INITIAL_FORK_BLOCK_NUMBER` we set `BASE_FEE = INITIAL_BASE_FEE`
+* `BASE_FEE` is set as follows
+  * Let `delta = block.gas_used - TARGET_GASUSED` (possibly negative).
+  * Set `BASE_FEE = PARENT_BASE_FEE + PARENT_BASE_FEE * delta // TARGET_GASUSED // BASE_FEE_MAX_CHANGE_DENOMINATOR`
+  * Clamp the resulting `BASE_FEE` inside of the allowable bounds if needed, where a valid `BASE_FEE` is one such that `abs(BASE_FEE - PARENT_BASE_FEE) <= max(1, PARENT_BASE_FEE // BASE_FEE_MAX_CHANGE_DENOMINATOR)`
+* EIP-3416 transactions are encoded the same as the current ones `rlp([nonce, gasPrice, gasLimit, to, value, data, v, r, s])` where `v,r,s` is a signature of `rlp([nonce, gasPrice, gasLimit, to, value, data])` and `gasPrice` is the `FEE_CAP` specified by the sender according to this proposal.
```
MicahZoltu, line 68: Either there is no change in transactions at all, or there is a change in which case you'll need to specify a new transaction (e.g., an EIP-2718 transaction).  It sounds like you just want no new transaction type, in which case I recommend just leaving this sentence out and rewording some other stuff to make it clear that existing transactions are just being reinterpreted.

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+The transition to this gas price system will occur in two phases, in the first phase both legacy and EIP3416 transactions will be accepted by the protocol. Over the course of this first phase the amount of gas available for processing legacy transactions will decrease while the amount of gas available for processing EIP3416 transactions will increase, moving gas from the legacy pool into the EIP3416 pool until the legacy pool is depleted and the EIP3416 pool contains the entire gas maximum. After all of the gas has transitioned to the EIP3416 pool, the second, finalized, phase is entered and legacy transactions will no longer be accepted on the network.
+
+## Specification
+
+### Definitions
+
+This is a classic fork without a long migration time.
+
+* `FORK_BLOCK_NUMBER`: TBD.  Block number at or after which EIP-3416 transactions are valid.
+* `BLOCK_GAS_TARGET`: 10th item in the block header: The gas limit field, controlled by miner voting.  Previously referred to colloquially as `gas limit`, now referred to as `gas target`.  Controlled by miners in the same way as before where each miner can increase or decrease it by a very small amount relative to the parent block.
```
MicahZoltu, line 52: This constant isn't used elsewhere in the specification.  Is there no block elasticity?  If not, remove this constant (and you'll also need to adjust some other parts of this EIP which imply elasticity).

EIPS/eip-3416.md
```diff
@@ -0,0 +1,114 @@
...
+* `CURRENT_BLOCK`: The current block that is being worked with (either being validated, or being produced).
+* `BASE_FEE`: State maintained by the miners. Represents the base minimum amount of nanoeth spent for every unit of gas a transaction uses. Not included as transaction data.
+* `BASE_FEE_MAX_CHANGE_DENOMINATOR`: `8`
+* `BASE_FEE_MAX_CHANGE`: `CURRENT_BLOCK.parent.BASE_FEE / BASE_FEE_MAX_CHANGE_DENOMINATOR`. The maximum amount the `BASE_FEE` can change between blocks (either up or down).
+* `LEGACY_GAS_PRICE`: Median gas price in `FORK_BLOCK_NUMBER - 1`.
+* `INITIAL_BASE_FEE` : `LEGACY_GAS_PRICE`.
+* `BASE_FEE_CHANGE`: `CURRENT_BLOCK.parent.BASE_FEE`. The amount that the BASE_FEE will increase in `CURRENT_BLOCK` relative to `CURRENT_BLOCK.parent`.
```
MicahZoltu, line 59: Unused constant, remove or use.

MicahZoltu (changes requested):
Not a full review, but hopefully gets this EIP into a better place so it can be reviewed more thoroughly.  At the moment, the specification section doesn't appear to contain enough information to evaluate it properly and the sections don't appear to all agree on what the behavior actually is.

EIPS/eip-3416.md
hexzorro: thanks

EIPS/eip-3416.md
hexzorro: thx

EIPS/eip-3416.md
hexzorro: I dont want any new Tx type as is not needed in this setting, and it also leave wallets without any need to upgrade. Will improve the redaction.

EIPS/eip-3416.md
hexzorro: Wont have any phases, unless is needed as a chain upgrade standard procedure, that is out of my scope. Changes should not affect the wallets, also.

EIPS/eip-3416.md
hexzorro: I want to include *block elasticity*, will improve the specs. Thanks a lot.

EIPS/eip-3416.md
hexzorro: Will improve the elasticity feature specs, got mixed up with the median changes.

EIPS/eip-3416.md
hexzorro: The median is robust enough to handle a large percentage of Txs (<50%) replaced by miners Txs. And they have a penalty if they remove Txs beyond the average empty block utilization (currently around 3%). You mention below miners filling-up 50% of the block, that is a different scenario but might be not profitable.

hexzorro: The miners cannot profit from exorbitant fake fees because the maximum user TXs are willing to pay is `fee_cap - base_fee` with `(fee_cap - base_fee)/2` used to compute the median premium. If the median surpases `fee_cap - base_fee` for a given user Tx then this Tx cannot be included in the block.

hexzorro: I think the `base_fee` being burn or not is an orthogonal problem to the miners attacking the median, because in some scenarios the `base_fee` can be very small and the premium very large. 

In general, my limited micro-economic intuition says that if you are earning a large `base_fee`, lets say premium is around 10% of the `base_fee` then you have *much more to loose* when dropping real TXs, because you have already secured 90% of the fees. In the current proposal, with a Miner filling with fake TX the premium can be pushed a little bit over 2x of the original non-cheating premium (litle above20% of the base_fee in the current paragraph example). But let's assume `base_fees` are small or that we want to burn them...

I am adding an extra 1% mandatory burn of the premium gas fees to make this simplified analysis more compelling.

In my analysis, with a mandatory small burn of around 1% of the premium fess, if a miner fill up 50% of the block to attack the median, then it loose more than 50% of the premium fees, so is not profitable pass this point too. Let see an example assuming base_fees are burned:

Without malicious miner:

* We have baseFee in 100, we have 100 Txs spending 1 Gas unit each with feeCap (same as current gasPrice) uniformly in [102, 104, 106,..., 200] Gwei. Then gasPremium are uniformly [1, 2, ..., 100] Gwei (is (feeCap - baseFee)/2)
* Then the median for gasPremium is 50.5 Gwei, so Miner earns 50.5 Gwei * 100  = 5050 Gwei 

Malicious miner *M* tries to attack:

*  *M* wants to fully control the Median so it drops the lowest 51 Txs with premium [1, .., 50] to include 51 fake Txs with some higher premium. He will be profitable with the attack if he can push the remaining 49 real Tx to a median premium beyond X * 49 > 5050, then X > 5050/49 = 103 
* *M* choose as the fake premiums 104 Gwei because if it goes beyond 104, lets say to 106 (less than 2% increase in profit), it cannot include the real Tx transaction with premium 52 Gwei (willing to pay not more than 104), loosing 1/49 = 2.04% > 2% of the profit. If he goes beyond 106 Gwei as fake premium (going to 108 increase premium less than 4%) it looses Tx with premium [52, 53], loosing 2/49 = 4.08% > 4% of the profit. And so on...if it goes beyond 104 + 2*n in premium it then loses > 2*n% in Txs included.
* *M* cannot be more than 104/103 - 1 = 0.97% profitable when cheating. If a mandatory 1% burn of premium fees is enforced then it cannot profitable (in this type of simplified scenarios).

EIPS/eip-3416.md
hexzorro: ok, thanks.

MicahZoltu (changes requested):
Simple summary is still too long, but I think we can merge as draft and fix that up in a follow-up PR.

PR #3431, opened 2021-03-25
1559: Removes Implementation Sectoin

MicahZoltu:
`## Implementation` isn't a valid EIP section.  Reference implementation is, but that should be inline or in assets and is an optional section.


eip-automerger:

PR #3455, opened 2021-04-01
Add EIP: SUDO Opcode

wjmelements:
#### Changes
Introduces a new opcode that would unilaterally solve all of Ethereum's  user experience issues.
Reviewers @MicahZoltu


lightclient (approved):
Thank you for this submission, it is desperately needed. I look forward to the different types applications this will finally allow!

I know it's pretty late in the cycle, but @timbeiko - do you think we can squeeze this into London?

karalabe:
@lightclient I think there's still time for Berlin.

LefterisJP:
I give my seal of approval to this much needed EIP.

It would really help us in rotki. We don't handle user keys yet. But would love to be able to eventually offer trading capabilities to our users. With this EIP we will finally be able to achieve the goal of trading for them, without having to touch their keys.

Yay science!

boolafish (dismissed):

zashton (approved):

EIPS/eip-3455.md
```diff
@@ -0,0 +1,73 @@
...
+`SUDO` fixes this.
+
+Finally, there are many tokens and ether sitting in smart contracts locked due to a bug.
+We could finally close [#156](https://github.com/ethereum/EIPs/issues/156).
+We cannot currently reclaim ether from stuck accounts.
+`SUDO` fixes this.
+
+## Specification
+Adds a new opcode (`SUDO`) at `0xf8`.
+`SUDO` pops 8 parameters from the stack.
```
vrde, line 48: I like this proposal but I'm worried about the security implications of this. If it gets included as it is, everyone would be able to move funds around!

It would be much more secure if the `SUDO` opcode pops 9 parameters from the stack, where the first one is `password`. My bank doesn't allow passwords longer than 8 bytes and as we know banks are really secure, so it's safe to assume we need only 64 bits for the `password` field.

wjmelements: Don't forget about the PIN! We could use the upper 96 bits of the Sender address field for the password, with 64 bits for the password and the other 32 for a 4-digit PIN. This would address any security concerns.

vrde: It fits perfectly actually, we can squeeze everything in 256 bits. To recap:

- `64 bits`: 8 characters password
- `32 bits`: 4-digit pin
- `160 bits`: address

dimiandre: What about 2FA? Not sure if it's really safe without.

wjmelements: My bank doesn't use 2fa so I don't think that would help much. Besides, the PIN and the password are already two factors.

EIPS/eip-3455.md
```diff
@@ -0,0 +1,73 @@
...
+type: Standards Track
+category: Core
+created: 2021-04-01
+---
+
+## Simple Summary
+A new opcode is introduced to allow calling from an arbitrary sender address.
+
+## Abstract
+A new opcode, `SUDO`, is introduced with the same parameters as `CALL`, plus another parameter to specify the sender address.
```
nevillegrech, line 16: I really like this proposal and I'm sure there are no security issues. Wouldn't it be even more flexible to allow the `SUDO` opcode to perform `DELEGATECALL` on any sender's behalf though?

EIPS/eip-3455.md
parithosh: I think it is good security hygene to use 2FA though. How about we set a field called `2FA` and hardcode it to `passed`? That way everyone is happy.

timbeiko:
@lightclient @karalabe I forgot which day we were and my heart skipped a beat seeing [this thread](https://twitter.com/AndreCronjeTech/status/1377550692404113408) 😂

EIPS/eip-3455.md
arturgontijo: We'll be using FIFO, right?

chriseth:
I think we should do an analysis on MEV before this goes in.

EIPS/eip-3455.md
```diff
@@ -0,0 +1,73 @@
...
+If the call ended with `STOP`, `RETURN`, or `SELFDESTRUCT`, `1` is pushed.
+If the call ended with `REVERT`, `INVALID`, or an EVM assertion, `0` is pushed.
+
+## Rationale
+The `GAS` parameter is first so that callers can tediously compute how much of their remaining gas to send at the last possible moment.
+The remaining parameters inherited from `CALL` are in the same order, with sender inserted between.
+
+
+## Security Considerations
+It will be fine.
```
prestonvanloon, line 70: It might be worth adding that nothing bad has ever happened to a blockchain with SUDO opcode. That alone is evidence that this must be a good idea

brianmcmichael:
This removes the need for many pointless and extraneous `approve()` or `permit()` transactions, which should dramatically cut down on ERC20 state bloat. EIP-3455 is a green solution.

bneiluj:
This is amazing. Love it  🎣

androolloyd:
lgtm

jpitts:
Really great to see this happening. SUDO actually will require the protocol client to make use of the "Master Node" API. Not many people understand the nature of this resource or what its role in the network is. With SUDO, it is at last time to begin to open this early mystery up for all to see.

https://github.com/EthDevOps/master-node/blob/main/README.md

EIPS/eip-3455.md
Kixunil: @vrde this is unnecessary complication. Ethereum does not have security as a primary value, let others worry about it. It's much more important for ETH to focus on being world computer and a computer must first and foremost work. Adding security features increases code complexity and unreliability.

`sudo mine_this_transaction` would be much more useful than worrying about security since filthy capitalist miners are ripping off poor ETH users on fees and cause unreliability. Working class can't use the computer today, only the rich can.

Aldekein:
Also it would be cool to have `GOTO` implemented in a similar way! 👍

GNSPS:
Hmmm, I think when I speak by the whole of Diligence when I say this sounds dangerous... 👀

dtribble:
Because such transactions are clearly priority, they should execute with no gas fees as well.

RobertMCForster:
Finally all you crypto idealists understand that to become mainstream we need to prioritize user-friendliness over things no one cares about like security. Holding onto outdated decentralization ideals is exactly why Ethereum is being dominated by chains like EOS and this EIP is exactly what Ethereum needs to succeed.

EIPS/eip-3455.md
```diff
@@ -0,0 +1,73 @@
...
+A new opcode is introduced to allow calling from an arbitrary sender address.
+
+## Abstract
+A new opcode, `SUDO`, is introduced with the same parameters as `CALL`, plus another parameter to specify the sender address.
+
+## Motivation
+There are many use cases for being able to set the sender.
+
+Many tokens are stuck irretrievably because nobody has the key for the owner address.
+In particular, at address zero there is approximately 17 billion USD in tokens and ether according to [etherscan](https://etherscan.io/address/0x0000000000000000000000000000000000000000).
```
ckknight, line 22: Hmm, it looks like there is only 16.5 million USD, which unfortunately invalidates the next sentence.

wjmelements: It's up to 17.5 *b*illion today.

wjmelements: It used to show a number with over 21 zeroes which would have been a better story but I dunno what changed since then

NicolasMassart (approved):

wjmelements:
Merge?

Pandapip1 (dismissed):
We have decided that there are no issues with this EIP.

CC @SamWilsn for manual merge if need be

Pandapip1 (approved):

n012judge420:
> All Reviewers Have Approved; Performing Automatic Merge...Yaaaaaa

wjmelements:
Thanks @Pandapip1. We march towards a brighter future

PR #3457, opened 2021-04-01
eip2929 and eip2930: move to last call

holiman:
Moving 2929 and 2090 to last call


eip-automerger:

 - Trying to change EIP 2929 state from Review to Last Call
 - Trying to change EIP 2930 state from Review to Last Call

MicahZoltu:
These shouldn't have been merged without editor eyeballs on them.  Part of the point of the bot not auto-merging is to make sure that we get at least one extra set of eyes.

For example, they are both missing the `review-period-end` field, and an editor should go over the actual content as well to make sure everything is in order and follows the guidelines.

holiman:
Apologies!

PR #3463, opened 2021-04-03
Remove effective gas price from EIP-1559 receipt

lightclient:
I don't believe this extra field is useful in a consensus structure. Each field in the receipt can currently *only* be determined by reexecuting the transaction. `effectiveGasPrice` does not require reexecution, only the transaction's `feeCap` and `tip` in addition to the block's `baseFee`. This generally requires 1 additional RPC call to get the block `baseFee`, on top of what is done today (just `eth.getTransaction` today). Client could optimize this further by indexing transaction in relation to the block `baseFee`.


eip-automerger:

AFDudley:
This makes sense. Is there a way we can make sure clients add the new index?

lightclient:
@AFDudley we could modify the RPC spec to return `effectiveGasPrice` as part of `eth.getTransaction` and let clients decide how to implement. In the beginning, they may read the tx from disk and then the block to get the `baseFee` and construct the response, but in the future they could also add a `txhash => effectiveGasPrice` index if desired. The RPC spec is out of my domain of expertise though, so I would defer to those who most regularly interact with it.

abdelhamidbakhta (approved):
Lgtm

PR #3476, opened 2021-04-06
Create eip-antitoken.md

hazae41:
Representing debt in a token. Like an ERC-20 but with `receive` instead of `transfer`. Thus, the token has a negative value: the less you have, the better.

Discussion to: 
- https://github.com/ethereum/EIPs/issues/3477
- https://ethereum-magicians.org/t/erc-antitoken-representing-debt-in-a-token/5957
- https://ethresear.ch/t/erc-antitoken-representing-debt-in-a-token/9144

### DIFF
--- a/EIPS/eip-antitoken.md
+++ b/EIPS/eip-antitoken.md
@@ -1,11 +1,11 @@
 ---
 eip: <to be assigned>
 title: Antitoken
-author: hazae41 (@hazae41) <hazae41@gmail.com>
-discussions-to: 
+author: hazae41 (@hazae41)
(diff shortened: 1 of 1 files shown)


EIPS/eip-antitoken.md
```diff
@@ -0,0 +1,58 @@
+---
+eip: <to be assigned>
```
lightclient, line 2:
```suggestion
eip: 3476
```

lightclient (changes requested):
This EIP is written too informally. Please avoid phrases like "Like an ERC-20", "In a nutshell", and "roughly speaking". Additionally, it is generally better to not use second person pronouns in a specification.

PR #3484, opened 2021-04-09
2929: Fixes typo in max contract size

nventuro:
As per [EIP170](https://eips.ethereum.org/EIPS/eip-170), the maximum contract size is 24576, not 24000. The confusion likely comes from the fact that 24576 bytes are 24kB (as noted later in the same paragraph).


holiman (approved):
Thanks!

ytbigvinny27 (commented):
~

PR #3501, opened 2021-04-15
EIP 2929 -> final

vbuterin:
^^

### DIFF
--- a/EIPS/eip-2929.md
+++ b/EIPS/eip-2929.md
@@ -3,7 +3,7 @@ eip: 2929
 title: Gas cost increases for state access opcodes
 author: Vitalik Buterin (@vbuterin), Martin Swende (@holiman)
 discussions-to: https://ethereum-magicians.org/t/eip-2929-gas-cost-increases-for-state-access-opcodes/4558
-status: Last Call
+status: Final
 type: Standards Track
 category: Core
 created: 2020-09-01


alita-moore:


	 - Trying to change EIP 2929 state from last call to final

timbeiko:
@alita-moore why is this failing?

alita-moore:
@timbeiko this seems like an edge case where the author of the PR is an author on the PR so it fails because the change would have triggered a requirement for review but that edge case wasn't handled yet. I'll look into it.

MicahZoltu:
This failed correctly.  Editor approval is required to change the status to anything other than Withdrawn from Draft or Withdrawn from Review or Withdrawn from Last Call.

timbeiko:
@MicahZoltu so @vbuterin should have approved his on PR, basically?

MicahZoltu:
No, this one was submitted by an author, thus it just needs *Editor* approval (me, lightclient, axic, etc.).  In this case, I "approved" it by merging (since we are going through some bot transition growing pains).

timbeiko:
Got it, thanks! I assumed that an author's PR to their own EIP would be auto-merged, that's why.

MicahZoltu:
> Got it, thanks! I assumed that an author's PR to their own EIP would be auto-merged, that's why.

Normally this is true, but when changing the status (Draft > Review > Last Call > Final) you need an editor to also review.

PR #3508, opened 2021-04-16
Transaction Data Opcodes EIP Inclusion

alex-ppg:
Hello,

This PR is meant to introduce a new EIP for accessing a transaction's original data as well as original recipient to allow introspection and alleviate the issue introduced by [EIP-3074](https://eips.ethereum.org/EIPS/eip-3074).

Feedback is welcome on all points.

### DIFF
--- a/EIPS/eip-draft_transaction_data_opcodes.md
+++ b/EIPS/eip-draft_transaction_data_opcodes.md
@@ -2,7 +2,7 @@
 eip: <to be assigned>
 title: Transaction Data Opcodes
 author: Alex Papageorgiou (@alex-ppg)
-discussions-to: <ADD_MAGICIANS_HERE>
+discussions-to: https://ethereum-magicians.org/t/eip-draft-transaction-data-opcodes/6017
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-draft_transaction_data_opcodes.md
+++ b/EIPS/eip-draft_transaction_data_opcodes.md
@@ -1,0 +1,106 @@
+---
+eip: <to be assigned>
+title: Transaction Data Opcodes
+author: Alex Papageorgiou (@alex-ppg)
+discussions-to: <ADD_MAGICIANS_HERE>
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-04-16
+---
+
+## Simple Summary
+
+Provide access to original transaction data.
+
+## Abstract
+
+This EIP introduces the following four EVM instructions: `ORIGINDATALOAD`, `ORIGINDATASIZE`, `ORIGINDATACOPY` and `ENTRYPOINT`.
+
+The first three instructions are meant to provide access to the original transaction's `data` payload whereas the last instruction is meant to provide access to the original recipient of the transaction, the `to` address.
+
+## Motivation
+
+It is undeniable that smart contracts are becoming more interconnected than ever. Up until this point, smart contracts have entirely relied on compliant interfaces and introspection to introduce a new step in the call chain of a complex multi-contract interaction. However, this presents a forwards-only approach which limits the types of interactions that can manifest.
+
+The purpose of this EIP is to provide a way via which a contract is able to identify the entry-point of a transaction on the blockchain and deduce what was the original intention of the transaction by applying introspection on the original transaction data itself.
+
+This EIP enables the development of new types of smart contracts as it can open new pathways for [EIP-721](https://eips.ethereum.org/EIPS/eip-721) NFTs and [EIP-20](https://eips.ethereum.org/EIPS/eip-20) tokens to detect which action their transaction is part of, such as detecting a liquidity provision to a decentralized exchange or a loan within a collateralized lending protocol.
+
+As a side-effect, it acts as an enhanced "substitute" to the `ORIGIN` check that the [EIP-3074](https://eips.ethereum.org/EIPS/eip-3074) breaks, enabling a greater level of granularity to be applied to the introspection of a transaction f.e. by validating the `ENTRYPOINT` being equal to the currently executed contract.
+
+## Specification
+
+### ORIGINDATALOAD (`0x47`), ORIGINDATASIZE (`0x48`) and ORIGINDATACOPY (`0x49`)
+
+These instructions are meant to operate identically to their `CALL`-prefixed counterparts with the exception that they instead operate on the original `data` of a transaction instead of the current call's data. As the data is retrieved once again from the execution environment, the costs for the three instructions will be `G_verylow`, `G_base` and `G_base + G_verylow * (number of words copied, rounded up)` respectively.
+
+### ENTRYPOINT (`0x4a`)
+
+The `ENTRYPOINT` instruction uses 0 stack arguments and pushes the original `to` member of the transaction onto the stack. The address yielded by the instruction is a 160-bit value padded to 256-bits. The operation costs `G_base` to execute, similarly to `ORIGIN` (`0x32`).
+
+### Node Specification
+
+All instructions are meant to provide access to previously inaccessible data that did not exist in the execution context of the EVM. As a result, the data needs to be provided to the execution environments of all contract calls, similarly to how `ORIGIN` data is available as such. A reference implementation will be provided for `go-ethereum` (Golang) that implements the EIP at both the EVM level as well as the node level.
+
+### Compiler Specification
+
+High-level programming languages that compile to EVM bytecode should provide support for these new instructions from a high-level perspective by introducing new members to the transactional context they already support for the `ORIGIN` instruction.
+
+## Rationale
+
+### Naming Conventions
+
+The `ORIGIN`-prefixed instructions attempted to conform to the existing naming convention of `CALL`-prefixed instructions given the existence of the `ORIGIN` instruction which is equivalent to the `CALLER` (`0x33`) instruction but on the original transaction's context.
+
+The `ENTRYPOINT` instruction came to be by defining a sensible name that immediately and clearly depicts what it is meant to achieve by signaling the first interaction with the blockchain, i.e. the entry-point.
+
+### Instruction Address Space
+
+The instruction address space of the `0x30-0x3f` has been exhausted by calls that already provide information about the execution context of a call so a new range had to be identified that is suitable for the purposes of the EIP.
+
+Given that the [EIP-1344](https://eips.ethereum.org/EIPS/eip-1344) `CHAINID` opcode was included at `0x46`, it made sense to include additional transaction-related data beyond it since the Chain ID is also included in transaction payloads apart from the blocks themselves, rendering the `0x46-0x4f` address space reserved for more transaction-related data that may be necessary in the future, such as the EOA's nonce.
+
+### Gas Costs
+
+The gas costs necessary for a particular instruction must weight the computational expense it brings to the system to assess a particular number. In the case of the instructions proposed by this EIP, this assessment was relatively straightforward given that identical instructions have already been introduced to the EVM and the instructions themselves would operate in an equivalent manner.
+
+### Instruction Space Pollution
+
+One can argue that multiple new EVM instructions pollute the EVM instruction address space and could cause issues in assigning sensible instruction codes to future instructions. This particular issue was assessed and a methodology via which the raw RLP encoded transaction may be accessible to the EVM was ideated. This would *future-proof* the new instruction set as it would be usable for other members of the transaction that may be desired to be accessible on-chain in the future, however, it would also cause a redundancy in the `ORIGIN` opcode.
+
+## Backwards Compatibility
+
+The EIP does not alter or adjust existing functionality provided by the EVM and as such, no known issues exist.
+
+## Test Cases
+
+TODO.
+
+## Reference Implementation
+
+A reference implementation will be provided for `go-ethereum` (Golang) which other implementations can be stemmed from should the EIP be moved to **`Accepted`** status.
+
+## Security Considerations
+
+### Introspective Contracts
+
+Atomically, the `ORIGINDATALOAD` and `ORIGINDATACOPY` values should be considered insecure as they can easily be spoofed by creating an entry smart contract with the appropriate function signature and arguments that consequently invokes other contracts within the call chain. To this end, the EIP has introduced the `ENTRYPOINT` instruction that should be validated along with the original data payload.
+
+Additionally, this type of introspection should solely be applied on pre-approved contracts rather than user-defined ones as the value stemming from this type of introspection entirely relies on a contract's code immutability and proper function, both of which a user supplied contract can easily bypass.
+
+### Denial-of-Service Attack
+
+An initial concern that may arise from this EIP is the additional contextual data that must be provided at the software level of nodes to the EVM in order for it to be able to access the necessary data via the `ORIGINDATALOAD` and `ORIGINDATACOPY` instructions.
+
+This would lead to an increase in memory consumption, however, this increase should be negligible if at all existent given that the data of a transaction should already exist in memory as part of its execution process; a step in the overall inclusion of a transaction within a block.
+
+### Multi-Contract System Gas Reduction
+
+Given that most complex smart contract systems deployed on Ethereum today rely on cross-contract interactions whereby values are passed from one contract to another via function calls, the `ORIGIN`-prefixed instruction set would enable a way for smart contract systems to acquire access to the original `ENTRYPOINT` function arguments at any given step in the call chain execution which could result in cross-contract calls ultimately consuming less gas if the data passed between them is reduced as a side-effect of this change.
+
+The gas reduction, however, would be an implementation-based optimization that would also be solely applicable for rudimentary memory arguments rather than storage-based data, the latter of which is most commonly utilized in these types of calls. As a result, the overall gas reduction observed by this change will be negligible.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


alita-moore:


	 - File with name EIPS/eip-draft_transaction_data_opcodes.md is new and new files must be reviewed
	 - Filename EIPS/eip-draft_transaction_data_opcodes.md is not in EIP format 'EIPS/eip-####.md'

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+The `ENTRYPOINT` instruction uses 0 stack arguments and pushes the original `to` member of the transaction onto the stack. The address yielded by the instruction is a 160-bit value padded to 256-bits. The operation costs `G_base` to execute, similarly to `ORIGIN` (`0x32`).
+
+### Node Specification
+
+All instructions are meant to provide access to previously inaccessible data that did not exist in the execution context of the EVM. As a result, the data needs to be provided to the execution environments of all contract calls, similarly to how `ORIGIN` data is available as such. A reference implementation will be provided for `go-ethereum` (Golang) that implements the EIP at both the EVM level as well as the node level.
+
+### Compiler Specification
+
+High-level programming languages that compile to EVM bytecode should provide support for these new instructions from a high-level perspective by introducing new members to the transactional context they already support for the `ORIGIN` instruction.
+
```
MicahZoltu, line 49:
```suggestion
```
Also unnecessary for the specification.  This specification is explicitly for the new EVM opcode(s), and it is up to HLL compiler teams to decide whether they want to add support and how to do so.

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+The `ORIGIN`-prefixed instructions attempted to conform to the existing naming convention of `CALL`-prefixed instructions given the existence of the `ORIGIN` instruction which is equivalent to the `CALLER` (`0x33`) instruction but on the original transaction's context.
+
+The `ENTRYPOINT` instruction came to be by defining a sensible name that immediately and clearly depicts what it is meant to achieve by signaling the first interaction with the blockchain, i.e. the entry-point.
+
+### Instruction Address Space
+
+The instruction address space of the `0x30-0x3f` has been exhausted by calls that already provide information about the execution context of a call so a new range had to be identified that is suitable for the purposes of the EIP.
+
+Given that the [EIP-1344](https://eips.ethereum.org/EIPS/eip-1344) `CHAINID` opcode was included at `0x46`, it made sense to include additional transaction-related data beyond it since the Chain ID is also included in transaction payloads apart from the blocks themselves, rendering the `0x46-0x4f` address space reserved for more transaction-related data that may be necessary in the future, such as the EOA's nonce.
```
MicahZoltu, line 62:
```suggestion
Given that the [EIP-1344](./eip-1344.md) `CHAINID` opcode was included at `0x46`, it made sense to include additional transaction-related data beyond it since the Chain ID is also included in transaction payloads apart from the blocks themselves, rendering the `0x46-0x4f` address space reserved for more transaction-related data that may be necessary in the future, such as the EOA's nonce.
```
All internal links should be relative so the links work regardless of how/where you are viewing this file.

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Gas Costs
+
+The gas costs necessary for a particular instruction must weight the computational expense it brings to the system to assess a particular number. In the case of the instructions proposed by this EIP, this assessment was relatively straightforward given that identical instructions have already been introduced to the EVM and the instructions themselves would operate in an equivalent manner.
```
MicahZoltu, line 66: Recommend simplifying this to just (fill in X, Y, Z):
```suggestion
These opcodes do essentially the same thing as opcodes X, Y, Z and thus share cast costs with those opcodes.
```

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+
+As a side-effect, it acts as an enhanced "substitute" to the `ORIGIN` check that the [EIP-3074](https://eips.ethereum.org/EIPS/eip-3074) breaks, enabling a greater level of granularity to be applied to the introspection of a transaction f.e. by validating the `ENTRYPOINT` being equal to the currently executed contract.
+
+## Specification
+
+### ORIGINDATALOAD (`0x47`), ORIGINDATASIZE (`0x48`) and ORIGINDATACOPY (`0x49`)
+
+These instructions are meant to operate identically to their `CALL`-prefixed counterparts with the exception that they instead operate on the original `data` of a transaction instead of the current call's data. As the data is retrieved once again from the execution environment, the costs for the three instructions will be `G_verylow`, `G_base` and `G_base + G_verylow * (number of words copied, rounded up)` respectively.
```
MicahZoltu, line 36: When referring to existing opcodes, use their full name and opcode value like `ORIGINDATALOAD (0x47)`.  In this case, please list the CALL opcodes you are referring to along with their values.

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Motivation
+
+It is undeniable that smart contracts are becoming more interconnected than ever. Up until this point, smart contracts have entirely relied on compliant interfaces and introspection to introduce a new step in the call chain of a complex multi-contract interaction. However, this presents a forwards-only approach which limits the types of interactions that can manifest.
+
+The purpose of this EIP is to provide a way via which a contract is able to identify the entry-point of a transaction on the blockchain and deduce what was the original intention of the transaction by applying introspection on the original transaction data itself.
+
+This EIP enables the development of new types of smart contracts as it can open new pathways for [EIP-721](https://eips.ethereum.org/EIPS/eip-721) NFTs and [EIP-20](https://eips.ethereum.org/EIPS/eip-20) tokens to detect which action their transaction is part of, such as detecting a liquidity provision to a decentralized exchange or a loan within a collateralized lending protocol.
```
MicahZoltu, line 28:
```suggestion
This EIP enables the development of new types of smart contracts as it can open new pathways for [EIP-721](./eip-721.md) NFTs and [EIP-20](./eip-20.md) tokens to detect which action their transaction is part of, such as detecting a liquidity provision to a decentralized exchange or a loan within a collateralized lending protocol.
```

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+The EIP does not alter or adjust existing functionality provided by the EVM and as such, no known issues exist.
+
+## Test Cases
+
+TODO.
+
+## Reference Implementation
+
+A reference implementation will be provided for `go-ethereum` (Golang) which other implementations can be stemmed from should the EIP be moved to **`Accepted`** status.
+
```
MicahZoltu, line 83:
```suggestion
```
If you aren't planning on providing a reference implementation inline, just leave this section out.  Links to external repositories are strongly discouraged and usually not allowed, and in this case wouldn't be allowed.

MicahZoltu (changes requested):
A handful of minor editorial comments that need to be addressed and a number needs to be assigned.

The big change that I would recommend doing **before** merging as draft though is splitting this into two EIPs.

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+created: 2021-04-16
+---
+
+## Simple Summary
+
+Provide access to original transaction data.
+
+## Abstract
+
+This EIP introduces the following four EVM instructions: `ORIGINDATALOAD`, `ORIGINDATASIZE`, `ORIGINDATACOPY` and `ENTRYPOINT`.
```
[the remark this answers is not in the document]
alex-ppg: I believe there is a slight disconnect here about why the instructions were grouped together. The instructions `ORIGINDATA[LOAD|SIZE|COPY]` alone are **unusable** without the `ENTRYPOINT` opcode. While I can split the EIP into two different ones where `ORIGINDATA[LOAD|SIZE|COPY]` relies on the EIP of `ENTRYPOINT`, I believe the latter is the one with most controversy. 

Please do let me know if you still believe it is worthwhile to split this into two EIPs.

EIPS/eip-draft_transaction_data_opcodes.md
```diff
@@ -0,0 +1,106 @@
...
+These instructions are meant to operate identically to their `CALL`-prefixed counterparts with the exception that they instead operate on the original `data` of a transaction instead of the current call's data. As the data is retrieved once again from the execution environment, the costs for the three instructions will be `G_verylow`, `G_base` and `G_base + G_verylow * (number of words copied, rounded up)` respectively.
+
+### ENTRYPOINT (`0x4a`)
+
+The `ENTRYPOINT` instruction uses 0 stack arguments and pushes the original `to` member of the transaction onto the stack. The address yielded by the instruction is a 160-bit value padded to 256-bits. The operation costs `G_base` to execute, similarly to `ORIGIN` (`0x32`).
+
+### Node Specification
+
+All instructions are meant to provide access to previously inaccessible data that did not exist in the execution context of the EVM. As a result, the data needs to be provided to the execution environments of all contract calls, similarly to how `ORIGIN` data is available as such. A reference implementation will be provided for `go-ethereum` (Golang) that implements the EIP at both the EVM level as well as the node level.
+
```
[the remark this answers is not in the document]
alex-ppg: Thanks for the feedback, assimilated.

EIPS/eip-draft_transaction_data_opcodes.md
alex-ppg: Again thanks for the feedback, assimilated.

EIPS/eip-draft_transaction_data_opcodes.md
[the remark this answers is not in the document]
MicahZoltu: Regardless of your personal intent for including all of these opcodes, they can exist separately and one can come up with reasonable use cases for them independently.  For example, the origin data opcodes can be used to reduce the gas cost of passing large data payloads between call frames (a problem I have personally run into in the past).

EIPS/eip-draft_transaction_data_opcodes.md
alex-ppg: Should I also remove the Test Cases section in this case?

EIPS/eip-draft_transaction_data_opcodes.md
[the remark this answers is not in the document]
MicahZoltu: If you believe that `ENTRYPOINT` is usless without the ORIGINDATA opcodes, then you could create it as a separate EIP with a dependency on the ORIGINDATA opcodes (which have utility without `ENTRYPOINT`).

EIPS/eip-draft_transaction_data_opcodes.md
MicahZoltu: If you don't think it is possible to have a set of reference test vectors, yes.  However, most opcode EIPs include a set of test vectors in the form of a table or JSON that are platform agnostic.  In this case, maybe a tiny little program (a dozen opcodes long) along with transaction input data.  Check out some of the other opcode EIPs to see examples of how people have done this in the past.

EIPS/eip-draft_transaction_data_opcodes.md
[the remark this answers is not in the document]
alex-ppg: This is a security-related concern, it is not personal intent. The `ORIGINDATA`-based instructions **cannot safely be relied** on even between call frames as they can easily be spoofed if I initiate my transaction on my own contract. Hence, the dependency is the opposite of what you described whereby the `ORIGINDATA`-based instructions have a strict dependency to `ENTRYPOINT`.

EIPS/eip-draft_transaction_data_opcodes.md
alex-ppg: I see, thank you. I will leave the Test Cases as `TODO` for now.

EIPS/eip-draft_transaction_data_opcodes.md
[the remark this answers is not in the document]
MicahZoltu: I believe it is only a security related concern if you are trying to use this for origin discrimination.  If you are just using `ORIGINDATA` for gas cost savings when you have a very large `calldata` payload that has to be passed around, then there is no security concern.  Since there exists a valid use case for `ORIGINDATA` absent `ENTRYPOINT`, I recommend creating an EIP solely for `ORIGINDATA`.  You can then create a second EIP for `ENTRYPOINT` that adds additional functionality (origin discrimination) to the `ORIGINDATA` opcodes.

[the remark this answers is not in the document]
alex-ppg: I strongly believe it is a security related concern that will ultimately cause more hassle than utility. I will break a scenario of the optimization you just mentioned.

The `calldata` trait is applicable between contract calls as well so let's consider the following:

- Malicious Contract A 
- Contract B (Staking)
- Contract C (Token)

I invoke my contract A with the following argument: `1`

I then invoke contract B via my contract A with a "deposit" of `1000000`.

Contract B invokes a transfer from contract C which tries to apply its optimization and reads `1` as the original data for the "deposit" because it **incorrectly assumed** the caller is honest and that whatever was supplied to the "deposit" was the original data.

The usefulness of accessing the original data exists solely if it is possible to tie it with an address that accepted the data which we need to validate as "trusted".

EIPS/eip-draft_transaction_data_opcodes.md
alex-ppg: Thanks, I adopted this change in the latest commit. Kindly let me know if it looks good to you.

EIPS/eip-draft_transaction_data_opcodes.md
alex-ppg: Done, kindly review.

EIPS/eip-draft_transaction_data_opcodes.md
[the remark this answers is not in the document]
MicahZoltu:
```ts
contract MerkleProofValidator {
	function validateMerkleProof(uint256 offset, uint256 length) {
		// read calladata from provided offset and length into memory for processing
		ORIGINDATACOPY(offset, length);
		...
	}
}

contract SomeOtherContract {
	function entrypoint(uint256 something, bytes merkleProof) {
		bool isValid = merkleProofValidator.validateMerkleProof(96, merkleProof.length);
		if (!isValid) throw;
	}
}
```

In this example, `MerkleProofValidator` is some contract deployed to the chain that doesn't need to have any established trust relationship with its caller.  All it is doing is using `ORIGINDATA*` to avoid having to copy the entire merkle proof (which are usually kilobytes large) into memory every stack frame (a very expensive operation).  The caller is simply calling into a contract whose behavior it knows and passing along 2 `uint256`s (which could actually be packed into a single value) which is *way* lower cost than copying around the whole merkle proof between call frames.

I agree with you that there exist usages for this that need both, I'm merely arguing that `CALLDATA*` is useful for *different* reasons without `ENTRYPOINT`.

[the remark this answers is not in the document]
alex-ppg: In the example you provided, `merkleProof` could be completely filled with `0` if I called it via my malicious contract which accepted a valid `merkleProof` and then the function `entrypoint` could be computing the hash of the proof to mark it as "consumed" etc.

As the state transition of `entrypoint` can solely rely on its input arguments and has no way to guarantee that `validateMerkleProof` will acquire the same origin arguments, then relying on the output of `validateMerkleProof` as validation is insecure. An EIP with such an inherent security flaw should not be independent. The usefulness of data is governed by its validity, which in the case of no validation of the `ENTRYPOINT`, is none.

I would highly appreciate it if you can highlight where the above paragraph is logically incorrect.

I am all for splitting this into two EIPs, however, the EIP of the origin-prefixed opcodes must rely on the `ENTRYPOINT` EIP unless a use case is presented whereby the origin prefixed opcodes can be used as is **safely**.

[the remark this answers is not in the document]
MicahZoltu: > In the example you provided, merkleProof could be completely filled with 0 if I called it via my malicious contract which accepted a valid merkleProof and then the function entrypoint could be computing the hash of the proof to mark it as "consumed" etc.

That is fine, the merkle proof wouldn't validate in that case.  It doesn't matter where the merkle proof comes from in this scenario, we just need the data passed down the callstack to some deeply nested merkle validator function.  Note: Without `ORIGINDATA`, code that does the same thing would look like this:

```ts
function apple(bytes someData) { ... }
function banana(bytes someData) {
	var result = apple(someData);
}
function cherry(bytes someData) {
	var result = banana(someData);
}
```
Note that each function just receives a `bytes` parameter from its caller, and it doesn't need to trust or understand the provenance of that data, just like a `token.transfer(destination, amount)` doesn't care about the provenance of the destination or amount in the vast majority of cases, it just needs them passed in.

The reason we don't want to just pass a `bytes` down each frame of the callstack is because when that byte array is 5kb big it is prohibitively expensive to copy it between call frames.  This is *purely* a cost savings (and a massive one) thing.  A storage proof costs around 600,000 gas, and each call frame you pass it through increases that by a a large portion of that cost.  With `ORIGINDATA` opcodes, passing it down the stack would cost 32 bytes of `calldata` per `CALL` (which is relatively cheap).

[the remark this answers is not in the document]
alex-ppg: I fully comprehend the gas savings that the instruction brings as I noted them in the original EIP draft, however, I think there still is a disconnect here. I will utilize the example you have produced to illustrate how the inclusion of the instruction alone can cause seemingly correct code to operate incorrectly:

```js
function apple(bytes someData) { ... }
function banana(bytes someData) {
	var result = apple(someData);
}
function cherry(bytes someData) {
	var result = banana(someData);
}
```

In the above, we have the logical guarantee that the `someData` passed to `apple` will be equivalent to the data that was passed in to `cherry`. In the following:

```js
function apple() {
   // Load data using ORIGINDATALOAD
}
function banana() {
	var result = apple();
}
function cherry(bytes someData) {
	var result = banana();
}
```

This logical guarantee is broken because I could very well call `cherry` from my own contract which accepts an equivalent argument and supplies `cherry` with some other data. As a result, the `cherry` function cannot conduct a safe state transition as it does not know whether `apple` and `banana` indeed acquired the same argument. This can easily be exploited by smart contracts to provide malicious inputs and produce unwanted state transitions.

So the example you provided **does not do the same thing**, however, it could do so by validating that indeed what `ORIGINDATALOAD` returns is from the `cherry` function by validating that `ENTRYPOINT == cherryContract`. Without this association, the results of origin-prefixed opcodes must not be relied on.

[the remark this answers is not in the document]
MicahZoltu: Origin data is immutable, so some malicious function anywhere in the call stack cannot change it.  Essentially the use case I'm describing is pass by reference, rather than pass by value (which is how things work at the moment).  It is the difference between `cherry` calling `banana` and saying, "hey, I put some bytes on the stack for you to read" and it saying "hey, there are some bytes in origin data at offset 6 for you to read".  In both cases, `cherry` has 100% control of what data it passes to `banana`.  Since origin data is immutable, if `banana` then calls into `apple` and says "hey I have some data on the stack for you to read" it is the same trust model as them saying, "hey, I have some data from the origin data at offset 6 for you to read".

We aren't trying to prove provenance here, we are *solely* trying to pass data from function A to function B without having to copy it first.  Since origin data is immutable, we can simply *not* copy it and instead *reference* it.

I *think* you may be trying to apply your provenance detection system to my example, which has nothing to do with that at all.  I'm saying that the `ORIGINDATA*` opcodes are useful for things *besides* establishing data provenance, and that is in improved efficiency of passing large amounts of data down the call stack.  We aren't trying to introduce any caller/callee trust requirements with this usage, so the trust model is exactly the same as passing data via the stack/memory.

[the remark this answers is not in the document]
alex-ppg: Again, I do not think you fully comprehend the scenario I am laying out. Although origin data is immutable, it can be **malicious from the start**. What I mean by this is, my first interaction with the blockchain is my custom-built smart contract that accepts the malicious arguments and sends the correct arguments to `cherry`. In such a case, `cherry` **would not have 100% control of what data it passes to `banana`** because in your example **the pointer would be different**. 

As a result, the **trust model is not the same**. I will lay out two examples once again a bit more fleshed out:

```js
contract A {
    function banana() {
        // Using ORIGINDATALOAD here
    }
}

contract B {
    A private logic;
    
    function cherry(bytes memory someData) {
            logic.banana();
    }
}

contract MaliciousContract {
    bytes private someData;
    B private target;
    
    function cherry(bytes memory maliciousData) {
        target.cherry(someData);
    }
}
```

If the `cherry` function of `B` is called directly, then the `banana` function would indeed read the `someData` pointer. If I call the `cherry` function of `MaliciousContract` however, the `cherry` function of `B` would read the correct `someData` argument while the `banana` function of `A` would read the `maliciousData` argument. So the immutability of origin data does not really solve the issue of a malicious actor altering the data as they see fit since they can simply supply it incorrectly from the start.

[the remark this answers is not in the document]
MicahZoltu: You are assuming there is a *desire* by contract A or contract B to *care* about data provenance.  Imagine in that scenario if contract A literally doesn't care where the data it gets comes from (which is the case for nearly 100% of contracts on chain today, they only care that they have some data that is shaped a certain way).

An ERC20 token doesn't care where the `destination` or `amount` parameters are sourced from, it just uses them to execute some operations.

```ts
contract A {
	function swap(address sourceToken, address depositToken, uint256 sourceTokenAmount) {
		b.swapHelper(sourceToken, depositToken, sourceTokenAmount);
		// do some more stuff after the swap is complete now that we have some depositTokens on hand
	}
}
contract B {
	function swapHelper(address sourceToken, address depositToken, uint256 sourceTokenAmount) {
		uniswapRouter.swapExactTokenForToken(sourceToken, depositToken, sourceTokenAmount);
		depositToken.transfer(msg.sender, depositToken.balanceOf(this));
	}
}
```
In this scenario, the `swapHelper` doesn't look at the contents of the parameters, it just blindly passes them along.  `swap` similarly doesn't actually look at the values, it just passes them along.  We are copying those 3 parameters from calldata to memory, each stack frame, but no one cares where they came from or even what they contain (except Uniswap at the end).

We can save gas by instead doing this:
```ts
contract A {
	function swap(address sourceToken, address depositToken, uint256 sourceTokenAmount) {
		b.swapHelper();
		// do some more stuff after the swap is complete now that we have some depositTokens on hand
	}
}
contract B {
	function swapHelper() {
		sourceToken = tx.calldataread(4, 36);
		depositToken = tx.calldataread(36, 68);
		sourceTokenAmount = tx.calldataread(68, 100);
		uniswapRouter.swapExactTokenForToken(sourceToken, depositToken, sourceTokenAmount);
		depositToken.transfer(msg.sender, depositToken.balanceOf(this));
	}
}
```
Note: this example is obviously contrived as the gas savings is pretty insignificant (possibly negative in the end), but imagine if there was a 5kb parameter being passed along.

In this second example, `A.swap` calls into `B.swapHelper` knowing full well that it is going to read the transaction calldata and exactly what `B.swapHelper` will do (`B.swapHelper` is well known to `A` at deploy time).  `A.swap` could have used the overload from the first example to achieve the exact same goals, but it chose instead to save some gas by avoiding a calldata copy.

lightclient:
@alex-ppg thanks for this EIP, it is well written. I agree with @MicahZoltu here though regarding splitting the EIP into two. I understand that for your use case, it is paramount to also include `ENTRYPOINT`. However, there are practical use cases to include one of the other. Optimism is a good example:

The `optimism-geth` clients have a neat little trick (paraphrasing) where they watch chain events for a txs that update the latest root. When that happens, they get the tx hash that updated the root and query some 3rd party service (e.g. Infura) for that tx's calldata. They then use that txs calldata as the input to their state transition function. They can rely on this because their contract checks `msg.sender == tx.origin`, so `tx.data` will always be the exact input given to the optimism contract. EIP-3074 breaks this, so now `tx.data` may have additionally outer layers that `optimism-geth` wouldn't understand. They could fix this issue by using *either* `ORIGINCOPY` or `ENTRYPOINT`.

alex-ppg:
Thanks @MicahZoltu and @lightclient for the very good feedback, I have assimilated your thoughts into the EIP and separated it from the `ENTRYPOINT` EIP which I plan on opening a PR soon. Kindly let me know your thoughts on the latest version!

EIPS/eip-3508.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Specification
+
+### ORIGINDATALOAD (`0x47`), ORIGINDATASIZE (`0x48`) and ORIGINDATACOPY (`0x49`)
+
+These instructions are meant to operate identically to their call-prefixed counterparts with the exception that they instead operate on the original `data` of a transaction instead of the current call's data. In detail:
+
+- ORIGINDATALOAD (`0x47`) performs identically to CALLDATALOAD (`0x35`)
+- ORIGINDATASIZE (`0x48`) performs identically to CALLDATASIZE (`0x36`)
+- ORIGINDATACOPY (`0x49`) performs identically to CALLDATACOPY (`0x37`)
```
MicahZoltu, line 40: Change `identically` to `similarly` since these aren't strictly identical opcodes.

EIPS/eip-3508.md
```diff
@@ -0,0 +1,100 @@
...
+
+As the data is retrieved once again from the execution environment, the costs for the three instructions will be `G_verylow`, `G_base` and `G_base + G_verylow * (number of words copied, rounded up)` respectively.
+
+### `AUTHCALL` (`0xf7`) Interaction
+
+The [EIP-3074](./eip-3074) that will be part of the London fork has introduced a new call instruction called `AUTHCALL` (`0xf7`) that will replace a transaction's `ORIGIN` (`0x32`) with the context variable `authorized`. The intention of `AUTHCALL` is to prevent discrimination between smart contracts and EOAs which `ORIGIN` initially facilitated and as a result, it is sensible also replace the values retrieved by the `ORIGINDATA*` opcodes to the ones used in the `AUTHCALL`.
```
MicahZoltu, line 46:
```suggestion
The [EIP-3074](./eip-3074) that will be part of the London fork has introduced a new call instruction called `AUTHCALL` (`0xf7`) that will replace a transaction's `ORIGIN` (`0x32`) with the context variable `authorized`. One of the reasons `AUTHCALL` changes `ORIGIN`'s meaning is to prevent discrimination between smart contracts and EOAs which `ORIGIN` initially facilitated and as a result, it is sensible also replace the values retrieved by the `ORIGINDATA*` opcodes to the ones used in the `AUTHCALL`.
```

MicahZoltu: This should be moved to Rationale section.  The Specification section should only describe the technical changes, not the reasoning for them.

EIPS/eip-3508.md
```diff
@@ -0,0 +1,100 @@
...
+
+In detail, the transaction data the `ORIGINDATA*` opcodes operate on will be equivalent to the `calldata` specified in the `commit` of the `AUTH` message format. If a single-call `AUTH` message is defined, the opcodes will retrieve the data of `calldata`. Otherwise, the opcodes will retrieve the `calldata` of each sub-call (i.e. `call 1`, `call 2` ... `call X`) depending on the call execution context.
```
MicahZoltu, line 48:
```suggestion
The transaction data the `ORIGINDATA*` opcodes operate on will be equivalent to the `args*` parameter to the nearest `AUTHCALL` up the stack.  If there is no `AUTHCALL` in the stack then `ORIGINDATA*` will operate on the transaction's data field.
```
I'm pretty sure this is what you mean here.  Using the `AUTH` commit would be very bizarre.

EIPS/eip-3508.md
```diff
@@ -0,0 +1,100 @@
...
+
+This interaction ensures full compatibility with [EIP-3074](./eip-3074) and ensures that no form of discrimination is introduced back into the system by this EIP f.e. by contracts entirely relying on `ORIGINDATA*` and thus allowing only EOAs to supply data to them.
```
MicahZoltu, line 50: This interaction ensures full compatibility with [EIP-3074](./eip-3074) and ensures that no form of discrimination is introduced back into the system by this EIP e.g. by contracts entirely relying on `ORIGINDATA*` and thus allowing only EOAs to supply data to them.

MicahZoltu (approved):
No blockers for merging as draft.  I left some feedback that I encourage reviewing/incorporating though.

PR #3523, opened 2021-04-22
Add events to EIP1967

Amxx:


eip-automerger:

 - EIP 1967 requires approval from one of (@spalladino)

spalladino (approved):

wighawag:
Hey, great to see events being added.
Just one minor point of discussion that I do not feel too strongly about but I think it is worth mentioning :

the wording `Upgraded` has too strong connotation in my opinion, since updating an implementation could also be used to rollback to a previous version.

Following the wording of `AdminChanged`  we could instead use `ImplementationChanged`

frangio:
@wighawag Thanks for the feedback! We kind of agree that `ImplementationChanged` would be better, but we feel that it's already too late to change this, since OpenZeppelin proxies have been deployed many times for a long time now, and `Upgraded` is the event that they use.

PR #3538, opened 2021-04-28
EIP-3529: Rename NEW_MAX_REFUND_QUOTIENT to MAX_REFUND_QUOTIENT

chfast:
Rename `NEW_MAX_REFUND_QUOTIENT` to `MAX_REFUND_QUOTIENT` and explain what this constant refers to in Yellow Paper.

### DIFF
--- a/EIPS/eip-3529.md
+++ b/EIPS/eip-3529.md
@@ -28,13 +28,16 @@ Gas refunds for SSTORE and SELFDESTRUCT were originally introduced to motivate a
 | Constant | Value |
 | - | - |
 | `FORK_BLOCK` | TBD |
-| `NEW_MAX_REFUND_QUOTIENT` | 5 |
+| `MAX_REFUND_QUOTIENT` | 5 |
 
 For blocks where `block.number >= FORK_BLOCK`, the following changes apply.
 
 1. Remove the `SELFDESTRUCT` refund.
 2. Replace `SSTORE_CLEARS_SCHEDULE` (as defined in [EIP-2200](./eip-2200.md)) with `SSTORE_RESET_GAS + ACCESS_LIST_STORAGE_KEY_COST` (4,800 gas as of [EIP-2929](./eip-2929.md) + [EIP-2930](./eip-2930.md))
-3. Reduce the max gas refunded after a transaction to `gas_used // NEW_MAX_REFUND_QUOTIENT`
+3. Reduce the max gas refunded after a transaction to `gas_used // MAX_REFUND_QUOTIENT`
+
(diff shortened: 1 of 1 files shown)


alita-moore:


	 - eip-3529.md requires approval from one of (vbuterin, holiman)
 @vbuterin @holiman

EIPS/eip-3529.md
```diff
@@ -28,13 +28,16 @@ Gas refunds for SSTORE and SELFDESTRUCT were originally introduced to motivate a
...
+| `MAX_REFUND_QUOTIENT` | 5 |
 
 For blocks where `block.number >= FORK_BLOCK`, the following changes apply.
 
 1. Remove the `SELFDESTRUCT` refund.
 2. Replace `SSTORE_CLEARS_SCHEDULE` (as defined in [EIP-2200](./eip-2200.md)) with `SSTORE_RESET_GAS + ACCESS_LIST_STORAGE_KEY_COST` (4,800 gas as of [EIP-2929](./eip-2929.md) + [EIP-2930](./eip-2930.md))
-3. Reduce the max gas refunded after a transaction to `gas_used // NEW_MAX_REFUND_QUOTIENT`
+3. Reduce the max gas refunded after a transaction to `gas_used // MAX_REFUND_QUOTIENT`
+
+Remark: Ethereum Yellow Paper introduced the _max gas refunded_ in [(65)](https://github.com/ethereum/yellowpaper/blob/constantinople/Paper.tex#L648) as `gas_used // 2`. Here we
```
MicahZoltu, line 39: Don't link to external GitHub repositories.  We *may* be able to make an exception for the yellow-paper, but it would be preferable if we just avoided it all together.

In this case, is this remark necessary?  It feels like it isn't needed for the spec?

chfast: So where should I put it?

MicahZoltu: My suggestion would just be to leave the remark out entirely.  I'm not sure that it adds value.

Another option would be to retain the remark, but remove the reference to the yellow paper like:
```
Remark: Max gas refund was previously defined as `gas_used // 2`.  Here we...
```

A third option would be to include an excerpt from the yellow paper inline.  Part of the propblem is that the yellowpaper is mutable, and it may not always live in that particular location in that particular GitHub repository.

holiman (approved):
LGTM

axic:
Comment: I think this EIP should have a `requires: 2200` because it clearly depends on it and in fact just redefines one parameter introduced by 2200, so it can not be used/understood in isolation.

PR #3540, opened 2021-04-28
EIP-3540: EVM Object Format (EOF) v1

axic:


alita-moore:


	 - File with name EIPS/eip-3540.md is new and new files must be reviewed

chriseth:
Obsoletes EIP-2327 - not sure if we have a standardized field for that.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+Validating code during the contract creation process allows code versioning without an additional version field in the account. Versioning is a useful tool for introducing or deprecating features, especially for larger changes (such as significant changes to control flow, or features like account abstraction).
+
+The format described in this EIP introduces a simple and extensible container with a minimal set of changes required to both clients and languages, and introduces validation.
+
+The first tangible feature it provides is separation of code and data. This separation is especially beneficial for on-chain code validators (like those utilised by layer-2 scaling tools, such as Optimism), because they can more efficiently validate contracts (for example by skipping the data section). This can result in significant gas savings. Additionally, various (static) analysis tools can also benefit, though off-chain tools can already deal with existing code, so the impact is smaller.
```
chriseth, line 24: I would say the main benefit is not the efficiency of the validation but the fact that such systems do not need to manually deactivate the "store data in code" feature of a compiler and especially that they can use contracts that have constructors with parameters (constructor arguments have to be stored in the init code). If contracts contain data and the validation routine cannot distinguish code and data, there is always the risk of data failing the validation.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+A non-exhaustive list of future improvements which could benefit from this format:
```
chriseth, line 26: One important motivation is also the inclusion of a version number that allows changes that would otherwise be "very breaking" to be non-breaking.

chriseth: Maybe also clarify: These are not part of this EIP and whether or not these changes are actual improvements have not been verified by the authors.

Just to be safe here :)

MicahZoltu:
> Obsoletes EIP-2327 - not sure if we have a standardized field for that.

2327 is a draft, not a final EIP, so it should **NOT** be referenced in a new EIP.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+created: 2021-03-16
+---
+
+## Abstract
+
+We introduce an extensible and versioned container format for the EVM with a once-off validation at deploy time. The version described here brings the tangible benefit of code and data separation, and allows for easy introduction of a variety of changes in the future. It is introduced via two hard forks to avoid breaking any existing executable contracts.
+
+## Motivation
+
+On-chain deployed EVM bytecode contains no pre-defined structure today. Code is typically validated in clients to the extent of `JUMPDEST` analysis at runtime, every single time prior to execution. This poses not only an overhead, but also a challenge for introducing new or deprecating old features. [This initial proposal](https://notes.ethereum.org/@axic/evm-object-format) explains some of these challenges.
```
MicahZoltu, line 18: External links are strongly discouraged.  Please inline the relevant content here rather than linking externally to it.

Alternatively, consider including it as an asset in `../assets/eip-3540/evm-object-format.pdf` or something.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+The first tangible feature it provides is separation of code and data. This separation is especially beneficial for on-chain code validators (like those utilised by layer-2 scaling tools, such as Optimism), because they can more efficiently validate contracts (for example by skipping the data section). This can result in significant gas savings. Additionally, various (static) analysis tools can also benefit, though off-chain tools can already deal with existing code, so the impact is smaller.
+
+A non-exhaustive list of future improvements which could benefit from this format:
+- Including a `JUMPDEST`-table (to avoid analysis at execution time) or removing `JUMPDEST`s entirely.
+- Introducing static jumps (with relative addresses) and jump tables, and disallowing dynamic jumps at the same time.
+- Requiring code section(s) to be terminated by `STOP`. (Assumptions like this can provide significant speed improvements in interpreters, such as a speed up of ~7% seen in [evmone](https://github.com/ethereum/evmone/pull/295).)
```
MicahZoltu, line 29: Remove external link.  Recommend just removing the entire second sentence as a specific example like this isn't necessary for an EIP and is better suited for external documentation/blog posts/discussions.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+- Multi-byte opcodes without any workarounds.
+- Representing functions as individual code sections instead of subroutines.
+- Introducing a specific section for the [EIP-2938 Account Abstraction](./eip-2938.md) "top-level AA execution frame", simplifying the proposal.
```
MicahZoltu, line 32: Recommend removing the link to 2938 as it will block this EIP from becoming final until 2938 is final.  Since this is just the motivation section, no link is necessary and you can just say:
```suggestion
- Introducing a specific section for Account Abstraction "top-level AA execution frame".
```

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Specification
+
+*We use [RFC2119](https://tools.ietf.org/html/rfc2119) keywords in this section.*
+
+In order to guarantee that every EOF-formatted contract in the state is valid, we need to prevent already deployed (and not validated) contracts from being recognized as such format. This will be achieved by choosing a byte sequence for the *magic* that doesn't exist in any of the already deployed contracts. To prevent the growth of the search space and to limit the analysis to the contracts existing before these changes, we split changes into two hard forks, and disallow the starting byte of the format (the first byte of the magic) in the first hard fork.
```
MicahZoltu, line 38: This is motivational, and should be moved to that section of the EIP rather than the specification section.  That being said, this feels a bit redundant to what is already in the motivation section so consider just dropping it.

MicahZoltu: Actually, this looks like it is saying that this EIP has a hard dependency on EIP-3541.  Recommend removing the backlink from 3541 to this EIP, and making this EIP depend on 3541.  Then you can just delete this whole paragraph.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+The *initcode* is the code executed in the context of the *create* transaction, `CREATE`, or `CREATE2` instructions. The *initcode* returns *code* (via the `RETURN` instruction), which is inserted into the account. See section 7 ("Contract Creation") in the Yellow Paper for more information.
+
+The opcode `0xEF` is currently an unimplemented instruction, therefore: *It pops no stack items and pushes no stack items, and it causes an exceptional abort when executed.* This means *initcode* or already deployed *code* starting with this instruction will continue to abort execution.
+
+### Between the two hard forks
+
+Next step to be conducted manually: Once the first hard fork went live, all existing contracts at `block.number == HF1_BLOCK` having their first byte as the `FORMAT` are examined off-chain. The goal is to find the shortest sequence after the `FORMAT` not matching any existing contract. We expect a 1- or 2-byte sequence would suffice. We call this byte sequence the *magic*, which will be used in the second hard fork.
+
+### Second hard fork
+
```
MicahZoltu, line 57: EIPs do not concern themselves with the hardfork coordination process.  This EIP should just have a dependency on 3541 and otherwise not bring up the "first hardfork" at all.  You may want to put a note in the backward compatibility section mentioning that there is value in having a delay between the introduction of 3541 and this EIP, but extended discussion about the migration process and timing should be left out of the EIP and put in the discussions-to link or some other external source.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+In this fork we introduce _code validation_ for new contract creation. To achieve this, we define a format called EVM Object Format (EOF), containing a version indicator, and a ruleset of validity tied to a given version.
```
MicahZoltu, line 58:
```suggestion
We define a format called EVM Object Format (EOF), containing a version indicator, and a ruleset of validity tied to a given version.
```

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+At `block.number == HF2_BLOCK` new contract creation is modified:
```
MicahZoltu, line 60:
```suggestion
As of FORK_BLOCK_NUMBER, new contract creation behaves as follows:
```

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+| 1            | code       |
+| 2            | data       |
+
+If the terminator is encountered, section size MUST NOT follow.
+
+The section contents follow after the header, in the order and size they are defined, without any padding bytes.
+
+To summarise, the bytecode has the following format:
+```
+format, magic, version, (section_kind, section_size)+, 0, <section contents>
```
MicahZoltu, line 98: This would be a great summary to include in the ## Abstract.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+#### Changes to execution semantics
+
+For clarity, the *container* refers to the complete account code, while *code* refers to the contents of the code section only.
+
+1. Jumpdest analysis is only run on the *code*.
+2. Execution starts at the first byte of the *code*, and `PC` is set to this position within the container format (e.g. `PC=10` for a *container* with a code and data section).
+3. If `PC` goes outside of the code section bounds, execution  aborts with failure.
+4. `PC` returns the current position within the *container*.
+5. `JUMP`/`JUMPI` uses an absolute offset within the *container*.
```
MicahZoltu, line 123: Why is PC and JUMP container absolute rather than code relative?  Given that neither can point outside of the code, it seems like this will introduce more complexity because now you have to know the offset of the code within the container in order to properly construct the JUMPs and during execution as well.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+#### Changes to contract creation semantics
+
+For clarity, the *EOF prefix* together with a version number $v$ is denoted as the *EOF$v$ prefix*, e.g. *EOF1 prefix*. 
+
+1. If _initcode's container_ has EOF1 prefix it must be valid EOF1 code.
+2. If _code's container_ has EOF1 prefix it must be valid EOF1 code.
+
+## Rationale
+
+EVM and/or account versioning has been discussed numerous times over the past years. This proposal aims to learn from them. A good starting point collecting the previous proposals is [this thread](https://ethereum-magicians.org/t/ethereum-account-versioning/3508).
```
MicahZoltu, line 136: Don't include external links in EIPs.  Remember, EIPs are technical specifications.  For complex topics (like this one), you **SHOULD** have a centralized external document for linking to discussions, further reading, etc. (similar to EIP-1559).

Also, this statement doesn't really belong in the rationale section as it doesn't actually explain any particular decisions.  Consider moving the sentiment to the motivation section, or just removing it (shorter EIPs tend to be better).
```suggestion
```

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Execution vs. creation time validation
+
+This specification introduces creation time validation, which means:
+- All created contracts with EOF$n$ prefix are valid according to version $n$ rules. This is very strong and useful property. The client can trust that the deployed code is well-formed.
+- In future, this allows serializing `JUMPDEST` map in the EOF container and eliminate the need of implicit `JUMPDEST` analysis required before execution.
+- Or completely remove the need for `JUMPDEST` instructions.
+- This helps with deprecating EVM instructions and/or features.
+- The biggest disadvantage is that deploy-time validation of EOF code must be enabled in two hard-forks.
```
MicahZoltu, line 145: Remove the multi-hard fork mentions, as they are out of scope for this specification and in scope for hard fork coordination discussion.
```suggestion
```

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+
+The alternative is to have execution time validation for EOF. This is performed every single time a contract is executed, however clients may be able to cache validation results. This approach has the following properties:
+- Because the validation is consensus-level execution step, it means the execution always requires the entire code. This makes code merkleization impractical.
+- The main advantage is it can be enabled via a single hard-fork.
```
MicahZoltu, line 149:
```suggestion
```

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+- A second advantage is better backwards compatibility: data contracts starting with the `0xEF` byte or the *EOF prefix* can be deployed. This is a dubious benefit however.
+
+### Initcode vs code
+
+After HF1 and before HF2 the `FORMAT` first byte check only applies to _code_. Applying the rule also to _initcode_ would have the same effects because is not distinguishable from executing _initcode_: if it starts with `FORMAT` execution would exceptionally abort. Yet, we decided against introducing an explicit check for _initcode_ for the simplicity of the HF1 spec.
```
MicahZoltu, line 154: Again, this EIP should be written standalone and without regard to the hardfork coordination process.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+To limit the number of exotic bytecode version combinations, additional restrictions are considered, but current not part of the specification:
+
+1. The EOF version of _initcode_ must much the version of _code_.
+2. An EOF1 contract must not create legacy contracts.
+
+Finally, create transaction must be allowed to contain legacy _initcode_ and deploy legacy _code_ because otherwise there is no transition period allowing upgrading transaction signing tools. Deprecating such transactions may be considered in future.
+
+### The FORMAT byte
+
+The `0xEF` byte was chosen because it resembles **E**xecutable **F**ormat. It has a long history of being proposed for this use case, starting with [this](https://github.com/ethereum/EIPs/issues/154).
```
MicahZoltu, line 176:
```suggestion
 The `0xEF` byte was chosen because it resembles **E**xecutable **F**ormat.
```
Don't include external links.  Also, having a long history of being proposed for a case doesn't really mean anything so I recommend just leaving that bit out entirely.

EIPS/eip-3540.md
```diff
@@ -0,0 +1,218 @@
...
+- Whether to have a header terminator or to encode `number_of_sections` or `total_size_of_headers`. Both of the latter raise the question how large of a value these fields should be able to hold. While today there will be only two sections, in case each "EVM function" would become their section, a fixed 8-bit field may not be big enough. A terminator byte seems to avoid these problems.
+- Whether to encode `section_size` as a fixed 16-bit value or some kind of variable length field (e.g. [LEB128](https://en.wikipedia.org/wiki/LEB128)). We have opted for fixed size, because it simplifies client implementations, and 16-bit seems enough, because of the currently exposed code size limit of 24576 bytes (see [EIP-170](./eip-170.md) and [EIP-2677](./eip-2677.md)). Should this be limiting in the future, a new EOF version could change the format.
```
MicahZoltu, line 183:
```suggestion
- Whether to encode `section_size` as a fixed 16-bit value or some kind of variable length field: We have opted for fixed size, because it simplifies client implementations, and 16-bit seems enough, because of the currently exposed code size limit of 24576 bytes (see [EIP-170](./eip-170.md) and [EIP-2677](./eip-2677.md)). Should this be limiting in the future, a new EOF version could change the format.
```

Also note: This creates a dependency on EIP-2677.  This may be fine, but consider removing the reference just to be safe.

MicahZoltu (changes requested):
I left a bunch of comments related to the HF1/HF2 situation.  I'm a huge fan of breaking this change up into multiple EIPs, and you may even want to consider breaking it up further into having a generalized contract versioning EIP and another for this first new contract version.  However, this EIP should just reference its dependency but otherwise should ignore the hard fork coordination process in this EIP outside of maybe a one-line mention in backward compatibility section.

EIPS/eip-3540.md
chriseth: This is indeed a debatable point. The argument for PC/JUMP/JUMPI being relative to container start is that this keeps it consistent with the other opcodes (`codecopy` etc.). The downside is that an assembler needs to determine the header size before filling in the final jump targets. Another argument for keeping it as it is here is that it might be a smaller change to interpreter implementations.

EIPS/eip-3540.md
axic: I disagree with this.

EIPS/eip-3540.md
axic: The reason it is kept this way is to reduce the impact on client implementations. The only change really is what section to run jumpdest analysis on, what is the starting PC and what is the "end of code" (i.e. the first 3 rules). 

Arguably it is easier to correctly change 2-3 assemblers, than changing all the clients correctly.

I personally was in favour of changing it already to use the code section only, but means we need to introduce `DATACOPY` in this version, which is a lot of more added complexity for little benefit.

axic:
@MicahZoltu thank you for the comprehensive review. We have debated this even before publishing this draft and out conclusion is that we would very much like to merge it in its current form with the long explainer and two hard forks. We expect to substantially simplify this EIP once 3541 is agreed on, but for now we feel this explainer is very much needed to explain the motivation and process.

We did consider that the better approach would have been writing a long explainer and two shorter EIPs, but due to the time pressure on London, I think this is a good solution for the moment.

EIPS/eip-3540.md
axic: I'm unsure why rationale/motivation would create any dependency when linking to other EIPs. The only dependency on other EIPs happens when referral is done in the specification.

This has been the case for many EIPs so far and I don't see a good reason why gutting everything is a good approach. I appreciate that goal of slimming specifications, but I don't think in practice it has worked out to find all relevant sources of information if external links are not permitted. Kind of catch 22.

EIPS/eip-3540.md
gumb0: Hard fork coordination complexities can affect the design desicions of the spec, can't they?

MicahZoltu: I suppose since this is the rationale section this line is probably fine to leave.  I still standby that the HF coordination stuff should be removed from the specification, and it should be limited in the rationale.  The key thing to remember is that EIPs are technical specifications designed to be read by future developers and they shouldn't overly focus on the discussions of *today*, but instead around how to make the specification understandable to developers of the future.

MicahZoltu:
> We expect to substantially simplify this EIP once 3541 is agreed on, but for now we feel this explainer is very much needed to explain the motivation and process.

@axic It sounds like what you want is a HackMD where you can have more free form iteration on this *idea* and a place to present future plans while pushing through EIP-3541.  Account Abstraction did this, and I feel it worked out quite well as they were able to lay out their roadmap while only the first step in the series was an actual EIP.

EIPS/eip-3540.md
axic: Addressed, please check.

EIPS/eip-3540.md
axic: Done.

MicahZoltu:
I still stand by all of my comments and consider this EIP as currently written to not be of appropriate quality to merge as a draft, mostly due to all of the external links/references.  However, there is disagreement as to whether the EIP process should be a change control process or a standardization process, and it seems a number of core developers want it to be a change control process.  Given that, I'm going to start merging EIPs with far less editorial review so they can be used more generally as a central document to record things related to proposed changes, rather than as a place to define standards.

EIPS/eip-3540.md
axic: How come a lot of EIPs can have external links in the abstract and rationale? Two recent examples are: https://eips.ethereum.org/EIPS/eip-2938 (starts with a very similar FEM link in the abstract) and https://eips.ethereum.org/EIPS/eip-3529 (has links to various draft proposals).

axic:
I really appreciate your review @MicahZoltu and hopefully you will engage when the time comes (i.e. getting closer to the Review state). However I am still super puzzled what you mean when you say this is not of the appropriate quality or that it is violating rules. I have thoroughly reviewed https://eips.ethereum.org/EIPS/eip-1#what-belongs-in-a-successful-eip again and it seems every section matches the description of that.

Can you please show me what rules you are referring to? Where are they described and what are they?

maurelian:
@axic, afaict this is the main issue @MicahZoltu  has:

> mostly due to all of the external links/references

(Only jumping in here because I would love to see this go ahead).

EIPS/eip-3540.md
mohamm0111294-afk: 1358

mohamm0111294-afk: 1358

PR #2186, opened 2019-07-10
EIP-1: remove requirement for simple summary

axic:

PR #3541, opened 2021-04-28
Add first draft of EOF1 HF1

axic:


alita-moore:


	 - File with name EIPS/eip-3541.md is new and new files must be reviewed

EIPS/eip-3541.md
```diff
@@ -0,0 +1,50 @@
...
+created: 2021-03-16
+---
+
+## Abstract
+
+Disallow new code starting with the `0xEF` byte to be deployed. Code already existing in the account trie starting with `0xEF` byte is not affected semantically by this change.
+
+## Motivation
+
+Contracts conforming to the EVM Object Format (EOF), described by [EIP-3540](./eip-3540.md), are going to be validated at deploy time. In order to guarantee that every EOF-formatted contract in the state is valid, we need to prevent already deployed (and not validated) contracts from being recognized as such format. This will be achieved by choosing a byte sequence for the *magic* that doesn't exist in any of the already deployed contracts. To prevent the growth of the search space and to limit the analysis to the contracts existing before this fork, we disallow the starting byte of the format (the first byte of the magic).
```
axic, line 18: Because of this link need to get #3540 merged first.

EIPS/eip-3541.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Specification
+
+After `block.number == HF_BLOCK` new contract creation (via create transaction, `CREATE` or `CREATE2` instructions) results in an exceptional abort if the _code_'s first byte is `0xEF`.
```
chriseth, line 22:
```suggestion
After `block.number == HF_BLOCK` the code deployment via create transaction, `CREATE` or `CREATE2` instructions results in an exceptional abort if the _deployed code_'s first byte is `0xEF`. 
```

This clarifies that the abort happens after the init code has been executed.

EIPS/eip-3541.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Remarks
+
+The *initcode* is the code executed in the context of the *create* transaction, `CREATE`, or `CREATE2` instructions. The *initcode* returns *code* (via the `RETURN` instruction), which is inserted into the account. See section 7 ("Contract Creation") in the Yellow Paper for more information.
+
+The opcode `0xEF` is currently an undefined instruction, therefore: *It pops no stack items and pushes no stack items, and it causes an exceptional abort when executed.* This means *initcode* or already deployed *code* starting with this instruction will continue to abort execution.
+
+## Rationale
+
+The `0xEF` byte was chosen because it resembles **E**xecutable **F**ormat.
```
MicahZoltu, line 32: Why not use the `REVERT (0xFD)`, `INVALID (0xFE)`, or `SELFDESTRUCT (0xFF)` instead, since we know that any such contracts that exist would do nothing?  Do we specifically want to "consumed" a new instruction for this?

EIPS/eip-3541.md
MicahZoltu: Alternatively, you could potentially reword this so it isn't dependent on 3540 specifically.  Maybe something like "having a reserved first byte will allow us to introduce EVM versioning in a future EIP."

EIPS/eip-3541.md
chriseth: This way it is "less breaking": It is generally agreed in the Ethereum community that contracts should not use unassigned opcodes, because hard forks could change the meaning of these opcodes. This means that contracts that conform to this agreement are unaffected by this EIP. If we change the EIP to use one of the opcodes you propose, then a contract that deploys code that starts with such an opcode will change it semantics: The deployment will fail after the hard fork. Note that a failure to deploy and deploying code that always reverts are two completely different things.

EIPS/eip-3541.md
chfast: This hits the naming issue we are currently struggle with. Yellow Paper defined _initcode_ and _code_ where the later means the deployed code. This is very confusing already in this case and only gets worse when EOF1 introduces code section.

I was carefully avoiding "deploy" name in the spec because this is not properly defined. But maybe we should introduce more precise names instead.

EIPS/eip-3541.md
axic: The EIP is not dependent on 3540 specification wise, but only in the motivation section. I think this is extremely important for understanding the motivation, so would prefer to keep it this way.

If/once 3540 goes live, we expect to simplify 3541.

EIPS/eip-3541.md
axic: Also during the initial discussions 0xEF was described as a new instruction called FORMAT, which if encountered anywhere results in abort. In the end we simplified the description to be a byte check in the deployed code, and given it is an invalid opcode, it results in abort in the initcode.

The risk we imagined is if people deploy contracts purposefully with 0xFE so that they are not executable, that means a lot of contracts could be already using it. This results in a longer prefix (magic).

If I'm not mistaken @holiman did a check on existing contracts on a recent block and it seems none of them are starting with 0xEF, 0xFD, 0xFE, or 0xFF.

However that does not mean there aren't any cases where code is already valid/expected (i.e. via a precalculated create2 hash in some contract), but not yet deployed. This is one of the problems @chriseth is referring to.

I think using 0xFE could be considered if we really want to save the 0xEF opcode for something else, but would refrain from 0xFD or 0xFF.

holiman: The check is still running. Right now, the numbers are, with `4646239` contracts checked `0` start with `0xef`, `1` starts with `fe`, `1` with `fd`, and `4` with `ff`. 
Still some time to go

axic: I'll update the rationale with some condensed form we discussed in this thread, but will wait for @holiman's final numbers to be included it in.

EIPS/eip-3541.md
MicahZoltu: Having a reference to 3540 in the motivation will mean that this EIP cannot advance through the EIP process faster than EIP-3541.  I strongly recommend removing the reference so this EIP doesn't get blocked by that.  This EIP is definitely reasonable to stand on its own IMO, as having a mechanism for contract versioning is important no matter what future versioning EIP we go with.

EIPS/eip-3541.md
holiman: Final update btw, `ef` was unused: https://gist.github.com/holiman/466cf6598e4a967620b9833b5ce49a23

EIPS/eip-3541.md
holiman: >  I strongly recommend removing the reference so this EIP doesn't get blocked by that. This EIP is definitely reasonable to stand on its own IMO, as having a mechanism for contract versioning is important no matter what future versioning EIP we go with.

Is that the current blocker to this eip getting merged?

MicahZoltu: Historically I haven't blocked EIPs merging as draft for that, though I do strongly prefer people try to have their EIPs as high of quality as possible and as alignment with the repository patterns/practices before merging as draft.

The current blocker is that this PR is marked as a draft PR, which is why I haven't given it an official review yet.  😄

gumb0: @MicahZoltu What is the origin of the rule that every linked EIP is a dependency of the current one?
In EIP-1 I see only the requirement that all dependencies should be explicitly listed in `requires`, but no restrictions on what can be linked.

holiman:
I didn't know that I could do that, but I un-drafted it and clicked "ready for review" @axic

EIPS/eip-3541.md
MicahZoltu: > What is the origin of the rule that every linked EIP is a dependency of the current one?
> In EIP-1 I see only the requirement that all dependencies should be explicitly listed in `requires`, but no restrictions on what can be linked.

@gumb0 EIP authors linking to EIPs that never make it past draft and eventually get moved to withdrawn or stagnant.  Then 3 years later you have a final EIP that references completely irrelevant content.  EIPs should be written for *future* readers, not *current* core developers.

I think the target audience of EIPs is the major point of contention here.  Many viewthe EIP process as a change control process for Ethereum, rather than a standardization process.  If it were a change control process, then the utility of an EIP would end when it reached `Final`.  As a standardization process, the utility of an EIP is considered to *begin* when it reaches `Final`.  We have been talking a lot lately about revamping the Ethereum specification/standardization/change control process to help address this conflict, but at the moment EIPs are standards, not change control documents.

chfast: I think it is exactly the opposite. The EIP **draft** is your best chance to properly explain and motivate a proposed EIP. This is because all other channels (Discord chat and ACD calls) are temporary and have very limited capacity. 
- An EIP draft should be written for _current_ core developers.
- Withdrawn or stagnant EIPs are relevant.

MicahZoltu: The purpose of the `discussions-to` link in an EIP is to fulfill that need.  The EIP should be written as it would be when final, purely as a standard/specification.  The information included in the motivation/rationale section should be things that will be relevant to people in the future.

chfast: The `discusssions-to` absolutely does not work for this. Nobody will ready 20+ posts there. Unless you are suggesting everyone should ask "What is the purpose of this EIP?" all over again. Moreover, the `FAQ` section is not allowed in an EIP. So we are gradually collecting answers for common questions and put then in Rationale. 

We have clear disagreement here. How do you plan to resolve it?

holiman: IMO: the discussions-to is a place to discuss and nitpick over details in the eip. However, I do think that the EIP itself should contain a bit of history for "why do we want this eip". Even for future developers, it's highly relevant to understand the context in which an eip was created. and what preceded it.
The discussions-to can be used to discuss "hey that reference is not useful", or "can't we change param X to `1337` instead of `5`". When all is said and done, however, the EIP should stand alone, and any _useful_ parts of previous discussions may well be _inside_ the eip (in some abbreviated form)

EIPS/eip-3541.md
```diff
@@ -0,0 +1,50 @@
...
+title: Reject new contracts starting with the 0xEF byte
+author: Alex Beregszaszi (@axic), Paweł Bylica (@chfast), Andrei Maiboroda (@gumb0)
+discussions-to: https://ethereum-magicians.org/t/evm-object-format-eof/5727/4
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-03-16
+---
+
+## Abstract
```
MicahZoltu, line 12:
```suggestion
## Simple Summary
Rejects new contracts starting with the 0xEF byte.

## Abstract
```

EIPS/eip-3541.md
```diff
@@ -0,0 +1,50 @@
+---
+eip: 3541
+title: Reject new contracts starting with the 0xEF byte
```
MicahZoltu, line 3:
```suggestion
title: Contract Leading Byte Reservation
```

MicahZoltu (changes requested):
Missing Simple Summary section.  The title also felt a bit wordy, so I provided a suggestion on tightening up the title and using the old title as the simple summary.

MicahZoltu:
@axic Are you OK with this being merged once the Simple Summary is added and CI is passing?  It was pulled out of draft so I gave it a formal review, but it was pulled out of draft by someone else so maybe you aren't actually ready?

EIPS/eip-3541.md
axic: The "Simple Summary" is not part of EIP-1. It has been removed over a year ago, see #2186.

EIPS/eip-3541.md
axic: Added explanation based on @chriseth's comment and explained the findings by @holiman.

axic:
Even though I updated this with master, it seems the EIP bot is still broken:
```
This simple branched action awaits for the CI to complete before either failing or succeeding. It checks the status every 30 seconds.
Error: Error: Parameter token or opts.auth is required
```

EIPS/eip-3541.md
MicahZoltu: Hmm, the template still references it and doesn't mention that it is optional.  We should probably resolve that.  I will bring it up with the other editors.

https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md

axic: This has been brought up months ago either on the channel and/or some other PR, and the agreement was to just fix the template, but apparently nobody did it 😉

MicahZoltu:
This cannot be merged (CI blocks it, and if it didn't I would block it) until EIP-3540 is merged.  I believe EIP-3540 cannot be merged (also blocked by CI) until this is merged.

You currently have a circular dependency that can trivially be removed by just having this EIP stand on its own as "reserving first byte of contracts for future usage".

MicahZoltu:
> Did you actually check the PR before claiming circular dependency? 2c6fd94 removed the link. The CI is broken because somehow the CI is broken: #3541 (comment)

Ah, I checked it a couple hours ago and it was still there, I missed that an updated happened since then.  It appears the current blocker is a spelling error.

EIPS/eip-3541.md
```diff
@@ -0,0 +1,56 @@
...
+
+The opcode `0xEF` is currently an undefined instruction, therefore: *It pops no stack items and pushes no stack items, and it causes an exceptional abort when executed.* This means *initcode* or already deployed *code* starting with this instruction will continue to abort execution.
+
+The exceptional abort due to *code* starting with `0xEF` behaves exactly the same as any other exceptional abort that can occur during *initcode* execution, i.e. in case of abort all gas provided to a `CREATE*` or create transaction is consumed.
+
+## Rationale
+
+The `0xEF` byte was chosen because it resembles **E**xecutable **F**ormat.
+
+Contracts using unassigned opcodes are generally understood to be at risk of changing semantics. Hence using the unassinged `0xEF` should have lesser effects, than chosing an assigned opcode, such as `0xFD` (`REVERT`), `0xFE` (`INVALID)`, or `0xFF` (`SELFDESTRUCT`). Arguably while such contracts may not be very useful, they are still using valid opcodes.
```
MicahZoltu, line 36:
```suggestion
Contracts using unassigned opcodes are generally understood to be at risk of changing semantics. Hence using the unassigned `0xEF` should have lesser effects, than choosing an assigned opcode, such as `0xFD` (`REVERT`), `0xFE` (`INVALID)`, or `0xFF` (`SELFDESTRUCT`). Arguably while such contracts may not be very useful, they are still using valid opcodes.
```

axic:
```
Error: Error: Parameter token or opts.auth is required
```

How is this a spelling error?

MicahZoltu:
> How is this a spelling error?

We are in the midst of a CI transition so the checks are a bit confusing at the moment.  You need to look at the Travis CI run, not the Await Travis CI run.

EIPS/eip-3541.md
```diff
@@ -0,0 +1,56 @@
...
+
+Analysis at block 18084433 showed that there are 0 existing contracts starting with the `0xEF` byte, as opposed to 1, 4, and 12 starting with `0xFD`, `0xFE`, and `0xFF`, respectively.
+
+## Test Cases
+
+Tests can be found at https://github.com/ethereum/tests/pull/835.
```
MicahZoltu, line 42: No external links.  Please inline the tests here (it is common to include input and output or input and expectations in a language neutral format), or add them as an asset in `../assets/eip-3541/*`.

EIPS/eip-3541.md
axic: I'm open to changes, but would like to get feedback from @chfast @gumb0 @AlexeyAkhunov @holiman @chriseth which one they prefer. However I would not make this a blocker for merging, as we can update it afterwards, i.e. in the one marking it review?

axic: My personal view is that perhaps the current one is better.

EIPS/eip-3541.md
axic: I find this no external links policy for drafts really not that useful. I will just remove this content for now, that's the least resistance I can do and is the result of these policies 😉

MicahZoltu: FWIW, I'm actively *trying* to get people's feedback on the direction we want to take EIPs (change control process vs standardization process) so we (me vs everyone) can stop clashing on this stuff.  The problem is that it is *really* hard to get anyone to engage on process changes, so it keeps defaulting back to "what Micah thinks is best because they are the only one actually engaging in the conversation outside of expressing annoyance in PR reviews".  This may not be what is *actually* best for the process (though I currently believe it is), but it is hard to find "best" when there is only one person actively engaging in long-term discussion.

EIPS/eip-3541.md
MicahZoltu: I'm willing to go to Draft without the Simple Summary, but this EIP should really be moving into Review almost immediately after draft since you want it included in London and I *believe* the current "policy" (desired operating procedure) is that things shouldn't even be considered/eligible for inclusion until they are in Review.  The simple solution if you are in a hurry (which I think you are) would be to just add the Simple Summary since we have been making everyone include a Simple Summary for a while until we can resolve the inconsistency between EIP-1 and the Template.

Note: The reason I'm such a stickler on these things is because we are hugely understaffed on the EIP editors front, and having simple hard and fast rules that *everyone* has to follow is *way* less time consuming than having fuzzy rules that we make exceptions for and have to defend those exceptions to authors of other EIPs.  The "rules" I have been following are that "EIPs should be formatted like the template outlines" in this case.

MicahZoltu (approved):

EIPS/eip-3541.md
axic: I think it is fair to say I did my part in engaging for years in the process :wink:

I do not fully understand what you mean when you say "change control" vs "standardization" in the context of EIPs, but drafts are very much a work-in-progress state where having authoritative external links is very useful. For example tests or code may be not final, but is essential to making use of an EIP efficiently, and hence during the draft state external links can accomplish that goal. I agree that the final state must have stricter rules, but it seems to be like putting a giant burden on people when asking them to keep the readyness/completeness of a draft at the equal level needed for a final EIP.

Why have drafts (and the draft status) in that case at all? Why not just have review/final statuses?

The answer to that (so far) was: the point of drafts was to collect efforts into this single directory (the EIP repository) as opposed to people trying to find various outlets to make and share their drafts.

MicahZoltu: Lets move this discussion to Discord, or if you prefer we can create a thread on Ethereum-Magicians, or we can discuss at the next EIPIP meeting.

I have been bringing this issue up for a while now and trying to get people to engage *outside of PRs* but so far I haven't had any luck.  Usually what happens is the person capitulates to the current process but complains about the process in the PR, and then once the PR is merged they stop caring enough to engage further and the issue gets dropped.

EIPS/eip-3541.md
axic: Here is my contribution to resolving the problem: #3564

Of course if you think the consensus of editors from >1.5 years ago is not valid or if you think the sentiment has changed, then it can't be easily merged.

EIPS/eip-3541.md
chriseth: I also prefer a more explicit title.

Issue #3239, opened 2021-02-01
EIP-3238: Difficulty Bomb Delay to Summer 2022


q9f:
Delays the difficulty bomb so 30 second blocks won't happen until around summer 2022.

PR: https://github.com/ethereum/EIPs/pull/3238

q9f:
:O

PR #3554, opened 2021-05-06
Create EIP-3554 for delaying the difficulty bomb until Shanghai and/or Merge.

MadeofTin:
Targeting 0.1 second effect on the first week of December 2021 with a delay of 9.5 million blocks.

### DIFF
--- a/EIPS/eip-3554.md
+++ b/EIPS/eip-3554.md
@@ -2,7 +2,7 @@
 eip: 3554
 title: Difficulty Bomb Delay to December 1st 2021
 author: James Hancock (@madeoftin)
-discussions-to: https://github.com/ethereum/EIPs/issues/3554
+discussions-to: https://ethereum-magicians.org/t/eip-3554-ice-age-delay-targeting-december-2021/6188
 type: Standards Track
 category: Core
 status: Draft
@@ -21,14 +21,12 @@ Targeting for the Shanghai upgrde and/or the Merge to occur before Decemember 20
 ## Specification
 #### Relax Difficulty with Fake Block Number
 For the purposes of `calc_difficulty`, simply replace the use of `block.number`, as used in the exponential ice age component, with the formula:
-
+```py
     fake_block_number = max(0, block.number - 9_500_000) if block.number >= FORK_BLOCK_NUMBER else block.number
-
+```
 ## Rationale
 
-I choose this amount using the following script; credit @vbuterin with slight modifications by me to make it easier to use. The following predicts a 0.2 second delay to blocktime the first week of december and a 1 second delay by the end of the month. This gives reason to address because the effect will be seen, but not so much urgency we don't have space to work around if needed.
-
-The adjustment is less the usual because the difficulty is so much higher today then previous adjustments.
+The following script predicts a 0.2 second delay to blocktime the first week of december and a 1 second delay by the end of the month. This gives reason to address because the effect will be seen, but not so much urgency we don't have space to work around if needed.
 
 ```python
 def predict_diff_bomb_effect(current_blknum, current_difficulty, block_adjustment, months):
@@ -51,9 +49,10 @@ diff_adjust_coeff = predict_diff_bomb_effect(12382958,7393633000000000,9500000,5
 ```
 
 ## Backwards Compatibility
-This EIP is not forward compatible and introduces backwards incompatibilities in the difficulty calculation.
+No known backward compatibility issues.
 
 ## Security Considerations
 Misjudging the effects of the difficulty can mean longer blocktimes than anticipated until hardfork is released. Wild shifts in difficulty can effect this number greatly. Also gradual changes in blocktimes due to longer term adjustments in difficulty can effect the timing of difficulty bomb epochs. This affects usability of the network, but unlikely to have security ramifications.
 
 ## Copyright
+Copyright and related rights waived via [CC0]
--- a/EIPS/eip-3554.md
+++ b/EIPS/eip-3554.md
@@ -55,4 +55,4 @@ No known backward compatibility issues.
 Misjudging the effects of the difficulty can mean longer blocktimes than anticipated until a hardfork is released. Wild shifts in difficulty can affect this number severely. Also, gradual changes in blocktimes due to longer-term adjustments in difficulty can affect the timing of difficulty bomb epochs. This affects the usability of the network but unlikely to have security ramifications.
 
 ## Copyright
-Copyright and related rights waived via [CC0]
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/EIPS/eip-3554.md
+++ b/EIPS/eip-3554.md
@@ -52,7 +52,7 @@ diff_adjust_coeff = predict_diff_bomb_effect(12382958,7393633000000000,9500000,5
 No known backward compatibility issues.
 
 ## Security Considerations
-Misjudging the effects of the difficulty can mean longer blocktimes than anticipated until hardfork is released. Wild shifts in difficulty can effect this number greatly. Also gradual changes in blocktimes due to longer term adjustments in difficulty can effect the timing of difficulty bomb epochs. This affects usability of the network, but unlikely to have security ramifications.
+Misjudging the effects of the difficulty can mean longer blocktimes than anticipated until a hardfork is released. Wild shifts in difficulty can affect this number severely. Also, gradual changes in blocktimes due to longer-term adjustments in difficulty can affect the timing of difficulty bomb epochs. This affects the usability of the network but unlikely to have security ramifications.
 
 ## Copyright
 Copyright and related rights waived via [CC0]
--- a/EIPS/eip-3554.md
+++ b/EIPS/eip-3554.md
@@ -33,7 +33,7 @@ The adjustment is less the usual because the difficulty is so much higher today 
 ```
 def predict_diff_bomb_effect(current_blknum, current_difficulty, block_adjustment, months):
     '''
-    Predicts the effect on block time (as a ratio) in a specificed amount of months in the future.
+    Predicts the effect on block time (as a ratio) in a specified amount of months in the future.
     Vars used in last prediction:
     current_blknum = 12382958
     current_difficulty = 7393633000000000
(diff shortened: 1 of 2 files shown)


EIPS/eip-3554.md
```diff
@@ -0,0 +1,59 @@
+---
+eip: 3554
+title: Difficulty Bomb Delay to December 1st 2021
+author: James Hancock (@madeoftin)
+discussions-to: https://github.com/ethereum/EIPs/issues/3554
```
MicahZoltu, line 5: This has to be a link to somewhere other than the PR itself.  https://ethereum-magicians.org/ thread is preferred, a GitHub issue in this repository can be used if you are unable/unwilling to create a magicians thread.

EIPS/eip-3554.md
```diff
@@ -0,0 +1,59 @@
...
+Starting with `FORK_BLOCK_NUMBER` the client will calculate the difficulty based on a fake block number suggesting to the client that the difficulty bomb is adjusting 9,500,000 blocks later than the actual block number.
+
+## Motivation
+Targeting for the Shanghai upgrde and/or the Merge to occur before Decemember 2021. Either the bomb can be readjusted at that time, or removed all together.
+
+## Specification
+#### Relax Difficulty with Fake Block Number
+For the purposes of `calc_difficulty`, simply replace the use of `block.number`, as used in the exponential ice age component, with the formula:
+
+    fake_block_number = max(0, block.number - 9_500_000) if block.number >= FORK_BLOCK_NUMBER else block.number
```
MicahZoltu, line 25:
````suggestion
```py
fake_block_number = max(0, block.number - 9_500_000) if block.number >= FORK_BLOCK_NUMBER else block.number
```
````

File: EIPS/eip-3554.md [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+
+## Rationale
+
+I choose this amount using the following script; credit @vbuterin with slight modifications by me to make it easier to use. The following predicts a 0.2 second delay to blocktime the first week of december and a 1 second delay by the end of the month. This gives reason to address because the effect will be seen, but not so much urgency we don't have space to work around if needed.
```
MicahZoltu, line 29: Recommend removing casual discussion and fluff and focus purely on the standard.
```suggestion
The following script predicts a 0.2 second delay to blocktime the first week of december and a 1 second delay by the end of the month. This gives reason to address because the effect will be seen, but not so much urgency we don't have space to work around if needed.
```

EIPS/eip-3554.md
```diff
@@ -0,0 +1,59 @@
...
+
+The adjustment is less the usual because the difficulty is so much higher today then previous adjustments.
+
```
MicahZoltu, line 32:
```suggestion
```
This doesn't really add value to the standard as it is just a commentary on history.

EIPS/eip-3554.md
```diff
@@ -0,0 +1,59 @@
...
+    diff_adjustment = 2 ** ((future_blknum - block_adjustment) // 100000 - 2)
+    diff_adjust_coeff = diff_adjustment / current_difficulty * 2048
+    return diff_adjust_coeff 
+
+
+diff_adjust_coeff = predict_diff_bomb_effect(12382958,7393633000000000,9500000,5)
+```
+
+## Backwards Compatibility
+This EIP is not forward compatible and introduces backwards incompatibilities in the difficulty calculation.
```
MicahZoltu, line 54: All core EIPs that require a hard fork are assumed to have this, so it doesn't need to be reiterated.
```suggestion
No known backward compatibility issues.
```

EIPS/eip-3554.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Security Considerations
+Misjudging the effects of the difficulty can mean longer blocktimes than anticipated until hardfork is released. Wild shifts in difficulty can effect this number greatly. Also gradual changes in blocktimes due to longer term adjustments in difficulty can effect the timing of difficulty bomb epochs. This affects usability of the network, but unlikely to have security ramifications.
+
+## Copyright
```
MicahZoltu, line 59:
```suggestion
## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```

MicahZoltu (changes requested):

alita-moore:


	 - File with name EIPS/eip-3554.md is new and new files must be reviewed

EIPS/eip-3554.md
MicahZoltu: We have been using "No known backward compatibility issues" for Core EIPs for a while now (I think since before I started doing EIP editing).  Things that end up in this section are stuff where the reader of the EIP will need to deal with something out of the ordinary, and "requires a hard fork" is pretty ordinary given that *all* Core EIPs require a hard fork.

A quick glance through final Core EIPs shows that there does appear to have been a period in time where people specified "requires a hard fork" in the backward compatibility section, but a substantial (over half based on my quick spot check) don't mention that.

I'm generally a fan of shorter EIPs, as they are more likely to be read and paid attention to, which is why I advocate for removing redundancy like this.

holiman: Ah.. Actually, that deeply offends me. There _are_ core eips that do not modify consensus, like the change to consider gas being bound to `uint64`. 

Saying something "has no known backward compatibility issues" when it requires a hard-fork to is ... just... (ノ°Д°）ノ︵ ┻━┻

tjayrush:
Quick back of the envelope calc:

|                                                  |     |            |
| ------------------------------------------------ | --- | ---------- |
| A. Fork block (assumes today)                    |     | 12,388,007 |
| B. Setback                                       |     | 9,500,000  |
| C. New Fake Block (A - B)                        |     | 2,888,007  |
| D. New Fake Period (C // 100,000)                |     | 28         |
| E. Expected Period Bomb First Becomes Noticable* |     | 42         |
| F. Periods before Noticable Bomb (E - D)         |     | 14         |
| G. Blocks before Noticable Bomb (F * 100,000)    |     | 1,400,000  |
| H. Seconds in that many blocks (G * 14)          |     | 19,600,000 |
| I. Days in that many seconds (H / 60 / 60 / 24)  |     | ~225       |
| J. Months in that many days (I / 30)             |     | 7.5        |
 
*\* - based on previous bombs.*

EIPS/eip-3554.md
timbeiko: Given this supersedes 3238, can we use the same discussions-to link, i.e. https://github.com/ethereum/EIPs/issues/3239?

EIPS/eip-3554.md
timbeiko: But in this case, this actually _is_ a breaking change, no? For example, EIP-3198 doesn't state that it has backward compatibility issues because (I assume) it introduces something new vs. changing existing behavior. This EIP changes existing behavior, so shouldn't we mention that?

EIPS/eip-3554.md
timbeiko: @lightclient weird - that issue redirects to this page.

timbeiko:
@tjayrush am I reading it correctly in thinking that the bomb would only be _noticeable_ 225 days after the fork block, i.e. late-Feb/early March 2022? If so, we likely want to shorten this period more so that we are closer to December.

EIPS/eip-3554.md
lightclient: Ohh I see, that is the PR number and so github just redirects it to PR even if it says its an issue. In that case, we haven't allowed discussion links to be PRs for a while.

MicahZoltu: `/issues/` and `/pulls/` are basically aliases of each other on GitHub.  All that matters is the number.

MicahZoltu: I would *prefer* a new issue just to reduce confusion.  Otherwise you can end up with two people talking about different things in the discussion.

tjayrush:
> @tjayrush am I reading it correctly in thinking that the bomb would only be _noticeable_ 225 days after the fork block, i.e. late-Feb/early March 2022? If so, we likely want to shorten this period more so that we are closer to December.

I'm not sure you can be that precise, so I think James' calculation is probably good.

If the actual fork block is after 12,400,000 then it would be one more period closer to 42. I started with yesterday's block (12,388,007). Seven and a half months from now is around December.

timbeiko:
Sorry, that's my bad! Got confused because 12833000 is a possible fork block (see https://github.com/ethereum/pm/issues/245#issuecomment-832122309) and yours starts with 12*38*. Misread and though they were the same 😅

7.5 months from now sounds good 👍

EIPS/eip-3554.md
MadeofTin: Good Idea.

EIPS/eip-3554.md
MadeofTin: removed.

MadeofTin: or moved to the discussion thread.

EIPS/eip-3554.md
MadeofTin: Not sure how that slipped out. Good catch.

EIPS/eip-3554.md
MadeofTin: Can I say something about it being a core EIP?
I agree the core EIP assumption makes sense. It might be nice to note it for those who don't have the "cultural" knowledge.

MadeofTin: Changed to the simple version. We can talk about if anything for core-EIPs makes sense separately.

EIPS/eip-3554.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+## Backwards Compatibility
+No known backward compatibility issues.
+
+## Security Considerations
+Misjudging the effects of the difficulty can mean longer blocktimes than anticipated until a hardfork is released. Wild shifts in difficulty can affect this number severely. Also, gradual changes in blocktimes due to longer-term adjustments in difficulty can affect the timing of difficulty bomb epochs. This affects the usability of the network but unlikely to have security ramifications.
+
+## Copyright
+Copyright and related rights waived via [CC0]
```
MicahZoltu, line 58:
```suggestion
## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```
Looks like this got messed up again.

MadeofTin: Got the text, but lost the link somehow. Should be fixed now.

PR #3556, opened 2021-05-07
Start with default base fee on fork block

lightclient:
As discussed on the infrastructure call, we are going to update the spec to align with client implementations. Previously, the spec defined the base fee for all blocks *before* the fork block to be `1_000_000_000`. Clients ended up implementing this as *the fork block* base fee is set to `1_000_000_000`. There is very little difference in the two, and since clients have already implemented the latter, we're just going to make it official via the spec.

It's important to note that while the base fee calculation doesn't require a different interpretation of the parent blocks gas limit, the gas target adjustment does. So before calculating the allowed range in which the gas target can be adjusted, clients should not divide out the `ELASTICITY_PARAMETER`.

### DIFF
--- a/EIPS/eip-1559.md
+++ b/EIPS/eip-1559.md
@@ -139,7 +139,7 @@ class Block:
 	extra_data: bytes = bytes()
 	proof_of_work: int = 0
 	nonce: int = 0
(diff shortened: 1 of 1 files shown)


alita-moore:


vbuterin (approved):

PR #3564, opened 2021-05-10
Remove the simple summary from the template

axic:
This has been removed from EIP-1 on 15-09-2019 in the PR https://github.com/ethereum/EIPs/pull/2186.

### DIFF
--- a/eip-template.md
+++ b/eip-template.md
@@ -17,9 +17,6 @@ Note that an EIP number will be assigned by an editor. When opening a pull reque
 
 The title should be 44 characters or less.
 
-## Simple Summary
-"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP. Imagine an email subject line, GitHub PR title, or forum post title.
-
 ## Abstract
 A short (~200 word) description of the technical issue being addressed. This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does.
 


MicahZoltu:
I *believe* that editor sentiment has changed since #2186, and we have come up with much better ways of describing the abstract vs title vs simple summary since then.

Though I originally found them confusing when I first got into EIP authoring, I have since gained a deeper understanding of their purpose and think they are valuable.  I have also gotten much better at explaining the difference to users, which I think helps resolve the confusion that #2186 was trying to resolve.

axic:
Outside of sentiment, are you stating that the template has stronger authority than EIP-1? If so this PR is a material, otherwise it is just a bugfix 😉

MicahZoltu:
I think the effective operating procedure has precedent, and we should be updating the documentation to match that.  Operating procedure for the last 9 months or so has been "Simple Summary is required", so the "bugfix" is EIP-1.

That being said, I recognize that it is very much non-ideal to have a different operating procedure from the documentation and it is certainly unfortunate that we got into this situation.

gcolvin:
I once opposed having a simple summary, but I with @MicahZoltu on this now.

lightclient:
I'm in favor of having an optional simple summary. I've opened a PR to update EIP-1 here: https://github.com/ethereum/EIPs/pull/3565

axic:
> There are currently around 270 EIPs using it.

I think a large number of them had it before the #2186 change, and we did not change them obviously. Since then apparently @MicahZoltu was enforcing having them, hence the number.

The template currently says this:
> "If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the EIP. Imagine an email subject line, GitHub PR title, or forum post title.

I want to emphasize `Imagine an email subject line, GitHub PR title, or forum post title`. If simple summary is a title, what is the `title` field we have? The same question was discussed in #2186.

If I can suggest one thing that would be rephrasing the explanation of abstract and simple summary so that it actually makes more sense than it does currently. (For example in the change #3541 which triggered this conversation, I would just duplicate the abstract as-is if this section is required.)

lightclient:
@axic: yeah I think the template (and the old EIP-1 description) aren't the best ways of explaining the purpose of the "Simple Summary". I think large EIP greatly benefit from both sections. For example, [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) and [EIP-2938](https://eips.ethereum.org/EIPS/eip-2938) seem to do a good job of utilizing both. In the case of EIP-3541, I don't see a need for a "Simple Summary". The abstract is enough of a description. This is why I believe the "Simple Summary" section should be *optional*.

axic:
I wonder if actually the abstract should have the first paragraph as the "simple summary". A very large number of the 270 EIPs containing the simple summary have just a duplicated section.

Another idea which occurred to me is why not including a `description` field in the header serving as the summary. If the goal is to use this in "social media preview" renderings and the like, that seems to be a more relevant place.

lightclient:
> A very large number of the 270 EIPs containing the simple summary have just a duplicated section.
I definitely think this should be avoided.

> Another idea which occurred to me is why not include a `description` field in the header serving as the summary. If the goal is to use this in "social media preview" renderings and the like, that seems to be a more relevant place.

That's a good idea.

> Oh and I did a quick browse in the afternoon and there are plenty of (even recent) EIPs which do not have an abstract, only simple summary. Why weren't having abstracts enforced?

It appears even I'm [guilty](https://eips.ethereum.org/EIPS/eip-3521)! I would say it's due to the lack of automated tools enforcing it.

MicahZoltu:
Title: The thing people call it in a sentence.  "What do you think about the **EVM Versioning** EIP?"

Simple Summary: The thing that shows up as the short description in lists, and is used to give a very brief overview to people who need a refresher.  "Which one was the EVM Versioning EIP?"  "It is the one that **introduces a leading reserved byte for contracts**."

Abstract: A human readable and terse description of the specification.  This can leave out edge cases and gotchas, but by the time someone finishes reading the abstract they should have a very rough idea of what this EIP is proposing on a technical level.  For very simple EIPs, it is acceptable to me that this mirrors the Simple Summary.  The two may show up in different locations/contexts, and if the EIP is so simple that a one-liner can give a reader a technical understanding of it then great, use that sentence for both.

axic:
This was replaced by #3706.

PR #3565, opened 2021-05-10
Add simple summary to eip-1

lightclient:
For quite a while, simple summary has been an optional section in EIPs. This formalizes that in EIP-1. This would close #3564.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -97,6 +97,7 @@ If this period results in necessary normative changes it will revert the EIP to 
 Each EIP should have the following parts:
 
 - Preamble - RFC 822 style headers containing metadata about the EIP, including the EIP number, a short descriptive title (limited to a maximum of 44 characters), and the author details. See [below](./eip-1.md#eip-header-preamble) for details.
+- Simple Summary (*optional) - A short (~1 sentence) non-technical summary of the EIP.
 - Abstract - A short (~200 word) description of the technical issue being addressed.
 - Motivation (*optional) - A motivation section is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. EIP submissions without sufficient motivation may be rejected outright.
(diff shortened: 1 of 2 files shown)


alita-moore:


	 - eip-1.md is in state living, not draft or last call or review
	 - eip-1.md requires approval from one of (mb@ethereum.org, hudson@ethereum.org)

EIPS/eip-1.md
```diff
@@ -97,6 +97,7 @@ If this period results in necessary normative changes it will revert the EIP to
 Each EIP should have the following parts:
 
 - Preamble - RFC 822 style headers containing metadata about the EIP, including the EIP number, a short descriptive title (limited to a maximum of 44 characters), and the author details. See [below](./eip-1.md#eip-header-preamble) for details.
+- Simple Summary (*optional) - A short (~1 sentence) non-technical summary of the EIP.
```
MicahZoltu, line 100: We have been enforcing this as non-optional.  What is the argument for optionality?  Having it required means that someone building a table of EIPs can rely on its presence for a sub-title/short description.

lightclient: Small EIP generally don't need both (e.g. EIP-3541). I don't think there is much value in someone building a table of EIPs with short summaries, because I'd like to move away from EIPs as they are today anyways.

MicahZoltu: While I agree with your desire to move away from EIPs in general, I am concerned that such things won't come to fruition for quite some time due to resource constraints.  I'm hesitant to make decisions that assume "this repository is dying anyway, we don't need to worry about long term maintenance/planning" until such time as there is a clear end of life schedule for it.

MicahZoltu: See https://github.com/ethereum/EIPs/pull/3564#issuecomment-840963361 for my argument for why I believe Simple Summary is valuable even when it mirrors the Abstract.  If anything, I would *rather* remove the Abstract in the case of duplication, though I believe that having both is valuable even if they overlap.

axic:
This was replaced by #3706.

PR #3567, opened 2021-05-11
Updates to 3541 (test cases and mark as review)

axic:


lightclient (approved):
Looks good to me - only potential issue is that the table isn't rendering properly on github. Maybe this is to render correctly on the EIPs website?

axic:
> Looks good to me - only potential issue is that the table isn't rendering properly on github. Maybe this is to render correctly on the EIPs website?

Good catch! It was rendering correctly in hackmd, but github is not so lenient on the extra column in the header separator. Fixed.

axic:
@MicahZoltu @lightclient can this merged by any chance before the ACD tomorrow?

axic:
Thanks @MicahZoltu!

PR #3575, opened 2021-05-14
EIP-2981 review comments

seibelj:
- Rename to `NFT Royalty Standard` to imply both 721 and 1155 compatibility
- Explain that it could potentially be a universal royalty standard, not just for NFTs
- Fix the 165 function calculation
- Rename the interface to _INTERFACE_ID_ERC2981
- Remove invalid return value from the constructor
- Use `supportsInterface()` for determining support in the example

### DIFF
--- a/EIPS/eip-2981.md
+++ b/EIPS/eip-2981.md
@@ -1,6 +1,6 @@
 ---
 eip: 2981
-title: ERC-721 Royalty Standard
+title: NFT Royalty Standard
 author: Zach Burks (@vexycats), James Morgan (@jamesmorgan), Blaine Malone (@blmalone), James Seibel (@seibelj)
 discussions-to: https://github.com/ethereum/EIPs/issues/2907
 status: Draft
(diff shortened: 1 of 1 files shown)


EIPS/eip-2981.md
```diff
@@ -139,7 +143,11 @@ This EIP mandates the inclusion of the `_data` parameter and the `_royaltyPaymen
 
 ## Backwards Compatibility
 
-This standard is completely compatible with current ERC-721 standards - in fact it requires it.
+This standard is compatible with current ERC-721 and ERC-1155 standards.
+
+## Universal Royalty Payments
+
+Although designed specifically with NFTs in mind, this standard does not require that a contract implementing EIP-2981 is compatible with either ERC-721 or ERC-1155 standards. Any other contract could integrate with this, provided it uniquely identifies assets by ID (like `tokenId`) to return royalty payment information. ERC-2981 is therefore a universal royalty standard for many other asset types.
```
lenifoti, line 150: >'Any other contract could integrate with this, provided it uniquely identifies assets by ID (like `tokenId`) to return royalty payment information.)'

This is not strictly true and open to misinterpretation.  I suggest a tighter definition:

>'Any other contract could support this interface to signal royalty payment information, with the limitation that assets can only be uniquely identified using the token id argument.'

EIPS/eip-2981.md
```diff
@@ -12,33 +12,35 @@ requires: 165
...
+This standard extends NFT standards such as the [ERC-721](./eip-721.md) and [ERC-1155]](./eip-1155.md) specifications to enable setting a royalty amount paid to the NFT creator or rights holder every time an NFT is sold and re-sold. This is intended for NFT marketplaces that want to support the ongoing funding of artists and other NFT creators. The royalty payment must be voluntary, as `transferFrom()` includes NFT transfers between wallets, and executing `transferFrom()` does not always imply a sale occurred. Marketplaces and individuals implement this standard by retrieving the royalty payment information with `royaltyInfo()`, which specifies how much to pay to which address for a given sale price. The exact mechanism for paying and notifying the recipient will be defined in future EIPs. This ERC should be considered a minimal, gas-efficient building block for further innovation in NFT royalty payments.
```
lenifoti, line 19: >This standard _extends_ NFT standards such as the [ERC-721](./eip-721.md) and [ERC-1155]](./eip-1155.md) specifications to _enable setting a royalty amount_ paid to the NFT creator or rights holder every time an NFT is sold and re-sold.'

I suggest:

''This standard allows contracts, such as NFTs that support [ERC-721](./eip-721.md) and [ERC-1155]](./eip-1155.md) interfaces, to signal a royalty amount to be paid to the NFT creator or rights holder every time the NFT is sold or re-sold.''

Rationale:

'extends' gives the impression that there is some dependency on NFT standards. That constraint has been removed form the backward compatibility section. The only crossover is the notion of indexing within the contract.

'setting' is not really the the function of this standard, it is about 'signaling'

EIPS/eip-2981.md
```diff
@@ -12,33 +12,35 @@ requires: 165
...
 
-This standard extends the [ERC-721 specification](./eip-721.md) to enable setting a royalty amount paid to the NFT creator or rights holder every time an NFT is sold and re-sold. If a marketplace chooses *not* to implement this EIP, then obviously no funds are paid for secondary sales. But as most NFT marketplaces have developed some unique royalty system themselves - and all of them are singular and only work within their own contracts - there should be an accepted standard for royalty payment information (if the creator chooses to set royalties on their NFTs). We believe the NFT marketplace ecosystem will voluntarily implement royalty payments to provide ongoing funding for artists and other creators, and NFT buyers will assess the royalty payment as a factor when making NFT purchasing decisions.
```
lenifoti, line 26: Same comment as above. RE: 'extends' and 'setting'.

EIPS/eip-2981.md
lenifoti, line 19: > 'The royalty payment must be voluntary, as `transferFrom()` includes NFT transfers between wallets, and executing `transferFrom()` does not always imply a sale occurred.'

Suggest: Remove references to transferFrom() and mention that enforced royalty payments can always be defeated by private arrangements.

Rationale:
Given that we are loosening the dependency on existing NFT standards, I think including specific transfer APIs is unnecessary and misleading. This section also does not address the more compelling argument that enforcing the royalty in a contract is actually impossible because extra payments can happen through a private arrangement.

lenifoti (commented):
See inline comments.  
I also think that backward compatibility and formal requirements (MUST/MUST NOT) need to be extended.  
Also some mention of the purpose of the extra data in the API needs to be stated, both descriptively and formally.

EIPS/eip-2981.md
jamesmorgan: I agree about the signalling part as well 👍🏻

jamesmorgan (commented):
LGTM in general - just few minor tweaks only, mainly around clarity as 
 suggested by the comments

YoutacRSVARs (approved):

lightclient:
Once this is ready to merge, please have an author approve it explicitly. Thanks.

PR #3576, opened 2021-05-14
Mark all London EIPs as Last Call

axic:
Suggested on ACD by @timbeiko, with the end-date set to the date of mainnet fork.

### DIFF
--- a/EIPS/eip-3529.md
+++ b/EIPS/eip-3529.md
@@ -3,7 +3,8 @@ eip: 3529
 title: Reduction in refunds
 author: Vitalik Buterin (@vbuterin), Martin Swende (@holiman)
 discussions-to: https://ethereum-magicians.org/t/eip-3529-reduction-in-refunds-alternative-to-eip-3298-and-3403-that-better-preserves-existing-clearing-incentives/6097
-status: Draft
+status: Last Call
+review-period-end: 2021-07-14
 type: Standards Track
 category: Core
 created: 2021-04-22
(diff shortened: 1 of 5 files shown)


vbuterin (approved):

MadeofTin (approved):

EIPS/eip-3529.md
```diff
@@ -3,7 +3,8 @@ eip: 3529
 title: Reduction in refunds
 author: Vitalik Buterin (@vbuterin), Martin Swende (@holiman)
 discussions-to: https://ethereum-magicians.org/t/eip-3529-reduction-in-refunds-alternative-to-eip-3298-and-3403-that-better-preserves-existing-clearing-incentives/6097
-status: Draft
```
axic, line 6: I think technically this should have gone through the `Review` state first, but it was extensively discussed during 2 (or 3) ACD meetings.

@timbeiko potentially the rule of considering only EIPs in the `Review` )or `Last Call`) state for final inclusion decision could be more strictly enforced.

axic:
Apparently the new automerge bot can't handle multiple files:
```
sorry only 1 file is supported right now
[
  'An Exception Occured While Linting: sorry only 1 file is supported right now'
]
```

EIPS/eip-3529.md
MadeofTin: It Technically okay to have skipped the step, but it shouldn't be the normal case. I remember writing something about what steps are allowed to be skipped, but I don't know it if ever got merged anywhere.

Being better with EIP-1 would help enforce some of these things as of now, what is and what isn't a rule isn't always clear.

MadeofTin:
Probably worth moving through this one and getting someone to hit the merge when we get the checks. Can we put in a feature request for multiple files? 😅

MadeofTin:
Actually he have all the checks we need as Vitalik is author on 3 of them. 👏

axic:
I tried triggering the old automerge bot, but the reason that won't work because it wants a clean CI state, i.e. no failures. Since the new bot displays as a failure, the old bot won't do anything. (It has been rendered useless.)

EIPS/eip-3529.md
timbeiko: @axic fair enough. I think it's more important to move them to `Last Call` now though given that's what triggers folks to look at the EIPs, so I would rather go Draft -> Last Call directly vs. delaying it by a few days/weeks.

MicahZoltu:
I don't think the old bot handled this situation either.

axic:
> I don't think the old bot handled this situation either.

It definitely worked for merging multiple files, but you are correct that since there was a status change it would not have merged it.

PR #3580, opened 2021-05-18
Make 191 and 234 final

lightclient:
While I don't want to require these old EIPs to be properly formatted to move them to final when they are effectively final already, in this case it was trivial to include the URLs while moving the EIPs to final.


fulldecent:
Just as well

fulldecent (approved):

MicahZoltu (approved):

PR #3591, opened 2021-05-24
Move EIP-2981 to Review with some final fixes

seibelj:
- Fix some broken links
- Remove incorrect example description @cliffhall
- Wrap `royaltyInfo()` to more lines @cliffhall
- Remove requirement that `transferFrom()` is used, and merely say it as an example @lenifoti
- Change the explanation to how this works to be 'signalling' rather than extending @lenifoti
- Require the same unit of exchange is passed to `royaltyInfo()`
- Move this EIP to **Review** 😎 

@blmalone @jamesmorgan @VexyCats for 👍


alita-moore:


	 - EIP 2981 state was changed from draft to last call
 @vexycats @jamesmorgan @blmalone @seibelj

VexyCats:
Goodjob!

jamesmorgan (approved):

jamesmorgan:
LGTM 👍🏻 ❤️

seibelj:
@jamesmorgan @VexyCats need another thumb. Had to change to `Review` from `Last Call` according to EIP flow spec. Also made a few tweaks based on latest comments.

EIPS/eip-2981.md
```diff
@@ -139,15 +149,15 @@ Additionally, we return a specific `_royaltyAmount` so there is no dispute with
...
 ### Unused data parameters in function signature
 
-This EIP mandates the inclusion of the `_data` parameter and the `_royaltyPaymentData` return value, and also mandates that these are unused. The purpose of this is to have a seamless path to build on this ERC by further EIP standards that specify payment notifications, fee splitting, and other advanced use cases.
+This EIP mandates the inclusion of the `_data` parameter and the `_royaltyPaymentData` return value, and also mandates that these are unused. The purpose of this is to have a seamless path to build on this ERC by further EIP standards that specify payment notifications, fee splitting, and other advanced use cases. Using them is considered undefined behavior.
 
 ## Backwards Compatibility
 
 This standard is compatible with current ERC-721 and ERC-1155 standards.
 
 ## Universal Royalty Payments
```
lightclient, line 158: This section should be made a subsection of a valid EIP section (see EIP-1 for a list).

lightclient (changes requested):
One small change request, then we can move to review. Thanks!

EIPS/eip-2981.md
cliffhall: This change seems out of step with the current discussion, which is leaning toward removal of these unused param/return values in favor of letting future EIPs simply overload or add methods as required.

lightclient: I'm not sure if I follow. I am requesting that you move the section "Universal Royalty Payments" from the `##` level to at least `###` under a valid EIP section. You can find a list of valid sections in EIP-1.

cliffhall: @lightclient No worries, I see what's happened. This PR from the other day was stuck, so I'm seeing all those changes, not changes you made.

dievardump:
2 questions:

1 / Shouldn't we provide the code using solidity 0.8 instead of 0.6?

2/ Solidity 8 introduced `type(InterfaceContract).interfaceId` to return an interface ID and facilitate introspection. 
However here if we add `supportsInterface` in the IERC2981, the interfaceId will differ from the one that is written in the contract.

- The bytes returned for the current implementation is `0x2baae9fd` 

- If we remove `supportsInterface` from it, then the interfaceId is the one expected: `0x2a55205a`

We should either remove `suppotsInterface` from the current implementation or add it to the interface id calculation.

jamesmorgan:
> 2 remarks:
> 
> 1 / Shouldn't we provide the code using solidity 0.8 instead of 0.6?
> 
> 2/ Solidity 8 introduced `type(InterfaceContract).interfaceId` to return an interface ID and facilitate introspection.
> However here if we add `supportsInterface` in the IERC2981, the interfaceId will differ from the one that is written in the contract.
> 
> * The bytes returned by `type(IERC2981).interfaceId` for the current implementation is `0x2baae9fd`
> * If we remove `supportsInterface` from it, then the interfaceId is the one expected: `0x2a55205a`
> 
> We should either remove `supportsInterface` from the current implementation or add it to the interface id calculation.

Not sure on this tbh, I assume it would be wise to use solidity 0.8 which means we should probably use the latest `type(InterfaceContract).interfaceId` format, however not sure what the typical path in this scenario. 

Doses anyone know of any solidity 0.8 defined standards to cross check?

wighawag:
since there already `is ERC165` we should not need to add `supportsInterface` again, no ?

cliffhall:
> since there already `is ERC165` we should not need to add `supportsInterface` again, no ?


Yes, I think it should be dropped. By including it on the interface, it is just another parent who needs overriding when you implement your NFT. This is what my current implementation looks like:

Screen Shot 2021-06-14 at 12 28 37 PM

cliffhall:
> since there already `is ERC165` we should not need to add `supportsInterface` again, no ?

Also, I [mentioned here](https://github.com/ethereum/EIPs/issues/2907#issuecomment-849136470) that we should change it to IERC165 and not ERC165.  

@jamesmorgan in that same post, I mentioned that we should move to 0.8.0, and thought that somewhere along the line someone had added it in a PR that never got merged or got stuck. Moving to 0.8.0 triggered the `supportsInterface` compiler  message that override wasn't present. If we ditch it from this interface and let it be inherited from IERC165 all should be well.

cliffhall:
To be more clear in the signature, can we also change `_value` to `_salePrice` since that's what it is described as?

cliffhall:
@jamesmorgan Thinking about this `supportsInterface` removal question, I noticed this example in the spec, where you're checking to see if a contract supports the royalties interface. 

Seems like we should cast `_contract` to `IERC165` and not `IERC2981` since we're testing a contract that may or may not be 2981. The most we can assume is IERC165 until we see that it supports `IERC2981`. 

Screen Shot 2021-06-14 at 1 07 40 PM

blmalone (approved):
LGTM 👍 

Great work.

lightclient (approved):

EIPS/eip-2981.md
```diff
@@ -44,7 +44,15 @@ RFC 2119.
...
+
+Implementers of this standard **MAY** use a different fixed-percentage royalty fee for each different `_tokenId`. This EIP does not require the same fixed-percentage royalty fee for every asset - it can vary for each asset.
+
+Marketplaces that support this standard **SHOULD NOT** send a zero-value transaction if the `_royaltyAmount` returned is `0`. This would waste gas and serves no useful purpose in this EIP.
+
+Marketplaces that support this standard **MUST** pay royalties no matter where the sale occured or in what currency, including on-chain sales, over-the-counter (OTC) sales, and off-chain sales such as at auction houses. As royalty payments are voluntary, entities that respect this EIP must pay no matter where the sale occurred - a sale conducted outside of the blockchain is still a sale. The exact mechanism for paying and notifying the recipient will be defined in future EIPs.
```
lightclient, line 55:
```suggestion
Marketplaces that support this standard **MUST** pay royalties no matter where the sale occurred or in what currency, including on-chain sales, over-the-counter (OTC) sales, and off-chain sales such as at auction houses. As royalty payments are voluntary, entities that respect this EIP must pay no matter where the sale occurred - a sale conducted outside of the blockchain is still a sale. The exact mechanism for paying and notifying the recipient will be defined in future EIPs.
```

lightclient (approved):

eip-automerger:

 - Trying to change EIP 2981 state from Draft to Review

PR #3605, opened 2021-06-08
Makes the EIPs site maximally responsive.

MicahZoltu:
If you don't like full-width pages, you can resize your browser window.  When looking at code, having artificial width limits can lead to unnecessary scroll bars.


alita-moore (approved):

lightclient (approved):
not a fan, but we can see how it goes I guess

lightclient:
I can't merge :(

PR #3607, opened 2021-06-10
EIP: Reject transactions from senders with deployed code

dankrad:
This EIP fixes the most sever problem with EOA and contract address collisions, by rejecting all transactions that come from accounts with deployed code (i.e. contracts)


alita-moore:


	 - File with name EIPS/eip-reject-transactions-from-senders-with-deployed-code.md is new and new files must be reviewed
	 - This PR requires review from one of [@MicahZoltu, @lightclient]
 @MicahZoltu @lightclient

dankrad:
Discussion to: https://github.com/ethereum/EIPs/issues/3608

EIPS/eip-reject-transactions-from-senders-with-deployed-code.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Test Cases
+
+Given a genesis allocation of  
+```
+Address: common.HexToAddress("0x71562b71999873DB5b286dF957af199Ec94617F7")
+Balance: big.NewInt(1000000000000000000), // 1 ether
+Nonce:   0,
+Code:    common.FromHex("0xB0B0FACE"),
+```
```
fjl, line 68: This example doesn't need to be Go code, you could just list the basic values.

EIPS/eip-reject-transactions-from-senders-with-deployed-code.md
```diff
@@ -0,0 +1,94 @@
...
+## Specification
+
+Any transaction where `tx.sender` has a `CODESIZE != 0` MUST be rejected as invalid.
+
+## Rationale
+
+We note that it was always the expected that a contract account's behaviour is constrainted by the code in that contract -- which means that the account's funds should not suddenly be spendable by some private key. It was just implicitly assumed in the past that a 160 bit address length is enough to provide collision resistance, and thus that this case could never occur. In that sense, this EIP should be seen as a clarification of protocol behaviour in a previously undefined case rather than an explicit upgrade of consensus rules.
+
+This does not exclude all possible attack vectors, only the most serious one. Further possible attack vectors via address collisions between contracts and EOAs are:
+1. An attacker can convince a user to send funds to an account before it is deployed. Some applications require this behaviour (e.g. state channels)
```
lightclient, line 48:
```suggestion
1. An attacker can convince a user to send funds to an account before it is deployed. Some applications require this behaviour (e.g. state channels).
```

lightclient (changes requested):
Nice EIP, just a couple nits.

EIPS/eip-3607.md
```diff
@@ -0,0 +1,94 @@
...
+if codesize := st.state.GetCodeSize(st.msg.From()); codesize != 0 {
+	return ErrSenderNoEOA
+}
+```
+
+An implementation in go-ethereum can be found [here](https://github.com/ethereum/go-ethereum/pull/23002)
+
+## Security Considerations
+
+This EIP is a strict security upgrade: It simply makes some transactions that were formely valid now invalid. There is no legitimate use for such transactions, so there should be no security downsides.
```
lightclient, line 88:
```suggestion
This EIP is a strict security upgrade: It simply makes some transactions that were formerly valid now invalid. There is no legitimate use for such transactions, so there should be no security downsides.
```

EIPS/eip-3607.md
```diff
@@ -0,0 +1,94 @@
...
+1. An attacker can convince a user to send funds to an account before it is deployed. Some applications require this behaviour (e.g. state channels).
+2. A chain reorg can happen after a contract is deployed. If the reorg removes the contract deployment transaction the funds can still be accessed using the private key.
+3. A contract can self desctruct, with the stated intention that ERC20s (or other tokens) in the contract would be burned. However, they can now be accessed by a key for that address.
+
+All these scenarios are much harder to exploit for an attacker, and likely have much lower yield making the attacks unlikely to be economically viable.
+
+## Backwards Compatibility
+
+It is unlikely that an attack like this has already occured on the Ethereum mainnet, or we would very likely have heard of it. It is inconceivable that someone would use this as a "feature" to make a contract an EOA at the same time, when they could simply do this by adding some methods to the contract instead of spending billions on building hardware to find hash collisions.
```
lightclient, line 56:
```suggestion
It is unlikely that an attack like this has already occurred on the Ethereum mainnet, or we would very likely have heard of it. It is inconceivable that someone would use this as a "feature" to make a contract an EOA at the same time, when they could simply do this by adding some methods to the contract instead of spending billions on building hardware to find hash collisions.
```

lightclient (changes requested):
sorry, missed these in the initial review

EIPS/eip-3607.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Reference Implementation
+
+The following check must be added to the state transition checks after checking that the nonce of the sender is correct.
+```go
+// Make sure the sender is an EOA
+if codesize := st.state.GetCodeSize(st.msg.From()); codesize != 0 {
+	return ErrSenderNoEOA
+}
+```
```
MicahZoltu, line 82: Consider writing this in pseudocode or a more well known/easier to read language than Go.  Also, consider extracting the assignment out of the conditional for a bit more clarity/readability.

EIPS/eip-3607.md
```diff
@@ -0,0 +1,94 @@
...
+
+An implementation in go-ethereum can be found [here](https://github.com/ethereum/go-ethereum/pull/23002)
```
MicahZoltu, line 84: Though we are no longer blocking PRs that include external links to transient things, I still recommend keeping these out of the EIP (same for similar links above).

MicahZoltu (approved):

EIPS/eip-3607.md
dankrad: This link in particular would not be transient -- however the links to Hackmd  would. What is the preferred format of these references? We could add them as an appendix? Or push them to a blog with a permalink?

EIPS/eip-3607.md
dankrad: @MariusVanDerWijden

EIPS/eip-3607.md
lightclient: I think the most preferable method is to add a PDF or markdown to this repository to the text. Here is the format for doing that: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md#auxiliary-files

youtacrsbyvars (commented):
###

EIPS/eip-3607.md
MicahZoltu: Text, JSON, PDF, code, etc. are all fine as assets, we just prefer they live *permanently* within the EIPs repository rather than living externally since organizations change names, repositories change names, GitHub may change its addressing format, etc.  Also, someone should be able to read and understand this EIP in isolation without having context about go-ethereum, Go, etc.

youtacrsbyvars (commented):
Yes

EIPS/eip-3607.md
```diff
@@ -0,0 +1,94 @@
...
+
+This does not exclude all possible attack vectors, only the most serious one. Further possible attack vectors via address collisions between contracts and EOAs are:
+1. An attacker can convince a user to send funds to an account before it is deployed. Some applications require this behaviour (e.g. state channels).
+2. A chain reorg can happen after a contract is deployed. If the reorg removes the contract deployment transaction the funds can still be accessed using the private key.
+3. A contract can self desctruct, with the stated intention that ERC20s (or other tokens) in the contract would be burned. However, they can now be accessed by a key for that address.
+
+All these scenarios are much harder to exploit for an attacker, and likely have much lower yield making the attacks unlikely to be economically viable.
```
3sGgpQ8H, line 52: The proposed solution doesn't solve the problem for those token contracts, that do implement [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612), as well as for other contracts that rely on the “ecrecover” function.  A very dirty way to fix this would be to modify the behavior of the “ecrecover” function to bever return an address where any code is deployed.
Another attack scenario, the proposed solution doesn't prevent, is when an attacker approves token transfer from an address before deploying a contract at this address.  However, preparation to such attack could be detected, as the attack requires a contract to be deployed at an address, that already has some outgoing transactions.

MicahZoltu: Please leave comments in the discussions-to link at the top of the EIP.  Messages left here are unlikely to be seen.

eulerbeat:
Is this (EIP-3607) deployed?

MicahZoltu:
Please use the `discussions-to` link for discussions/questions about an EIP.  You can find the link at the top of the EIP: https://eips.ethereum.org/EIPS/eip-3607

PR #2857, opened 2020-08-05
Setting ERC-1271 to Last Call

PhABC:
Setting the ERC-1271 status as Last Call

PR #3628, opened 2021-06-25
Setting ERC-1271 to Last Call

PhABC:
Setting this ERC to Last Call after addressing @MicahZoltu's [comments](https://github.com/ethereum/EIPs/pull/2857#pullrequestreview-462265956) on last attempt at setting to Last Call.


lightclient:
@PhABC generally we don't jump from draft status to last call. Could you please move EIP-1271 to review and then create a PR to move to last call?

PR #3632, opened 2021-06-28
3607: update to use empty code hash instead of codesize

MariusVanDerWijden:
As discussed on last ACD, we'd rather use the empty code hash instead of the code size cc @dankrad


dankrad:
Sounds good to me

YoutacVA (approved):

PR #3643, opened 2021-07-09
add eip TREX security token standard 

Joachim-Lebrun:
- Proposition of TREX security token standard, used by Tokeny.com to tokenize financial assets (already >8,5 billion € of tokenized assets with this standard) 
- interfaces detailed here https://github.com/TokenySolutions/EIP3643 with comments on the interfaces files describing each function
- full implementation available here https://github.com/TokenySolutions/T-REX/tree/develop
- whitepaper of T-REX available here https://tokeny.com/wp-content/uploads/2020/05/Whitepaper-T-REX-Security-Tokens-V3.pdf
- security audited by Kaspersky, report here https://tokeny.com/wp-content/uploads/2020/05/Tokeny-Solutions_T-REX-v3_Smart-Contract-Audit-Report_Kapersky.pdf

A discussion thread has been open on the ethereum magicians forum : https://ethereum-magicians.org/t/eip-3643-proposition-of-the-t-rex-token-standard-for-securities/6844


EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
+---
+eip: 42 (final number to be confirmed)
```
lightclient, line 2:
```suggestion
eip: 3643
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+title: T-REX - Token for Regulated EXchanges
+authors: Joachim Lebrun, Tony Malghem, Kevin Thizy, Luc Falempin, Adam Boudjemaa
+type: Standards Track
+category: ERC
+status: Draft
+require: ERC-20 , ERC-734, ERC-735, ERC-1822
```
lightclient, line 8:
```suggestion
require: 20, 734, 735, 1822
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
+---
+eip: 42 (final number to be confirmed)
+title: T-REX - Token for Regulated EXchanges
+authors: Joachim Lebrun, Tony Malghem, Kevin Thizy, Luc Falempin, Adam Boudjemaa
```
lightclient, line 4:
```suggestion
authors: Joachim Lebrun (@Joachim-Lebrun), Tony Malghem, Kevin Thizy, Luc Falempin, Adam Boudjemaa
```

lightclient: Please fill in other author's github handles.

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+type: Standards Track
+category: ERC
+status: Draft
+require: ERC-20 , ERC-734, ERC-735, ERC-1822
+created: 2021-07-09
+---
+
+### Note : we use the EIP n°42 in this proposition document but the final number needs to be determined by an EIP Editor, nonetheless the n°42 will be used until we have a final EIP number to refer to.
```
lightclient, line 12:
```suggestion
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Simple Summary
+
+The T-REX (ERC-42) token is an institutional grade security token standard, developed on Ethereum. <br>
```
lightclient, line 16:
```suggestion
The T-REX token is an institutional grade security token standard.
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+This standard provides a library of interfaces for the management and transfer of security tokens. It allows the contract owner/agent to manage the rules that the transfers have to fulfill to be considered as valid transfers as well as the rules related to an onchain identity to consider its owner as an eligible participant in the token ecosystem. The standard also provides a full set of functions to allow the token agent to manage the token properly.
```
lightclient, line 17: The simple summary should only be 1-2 sentences, please consolidate to that length.

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+
+
+
+## Abstract
+
+Standards should be backwards compatible with ERC-20 ([#20](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md)) and should be able to interact with ERC-735 ([#735](https://github.com/ethereum/EIPs/issues/735)) to validate the claims linked to an ONCHAINID, based on ERC-734 ([#734](https://github.com/ethereum/EIPs/issues/734)) and ERC-735. <br>
```
lightclient, line 23:
```suggestion
Standards should be backwards compatible with (ERC-20)[./eip-20.md] and should be able to interact with (ERC-735)[./eip-735.md] to validate the claims linked to an `ONCHAINID`, based on (ERC-734)[./eip-734.md] and ERC-735. 
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+The standard defines several interfaces that are described hereunder :
```
lightclient, line 24:
```suggestion
The standard defines several interfaces that are described hereunder:
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+- Token 
+- Identity Registry
+- Identity Registry Storage
+- Compliance 
+- Trusted Issuers Registry 
+- Claim Topics Registry
+
```
lightclient, line 31:
```suggestion
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+
+
+## Motivation
+
+Give standard interfaces for security tokens issued on Ethereum, through which any third party could interact with the security token. <br>
```
lightclient, line 36:
```suggestion
Give standard interfaces for security tokens issued on Ethereum, through which any third party could interact with the security token.
```

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+The functions described by these interfaces vary and allow the appropriate users to call a range of different actions, such as forced transfers, freeze tokens (partially or totally on a wallet or even freeze the entire token), minting, burning, recover lost tokens (if an investor loses access to his wallet), etc.
+
+The following requirements have been compiled following discussions with parties across financial institutions that are looking to issue securities on a DLT infrastructure such as ethereum.<br>
+
+- MUST be ERC-20 ([#20](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md)) compatible.
```
lightclient, line 41: Please follow the EIP-1 standard for linking to other EIPs.

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+- MUST be used in combination with an Identification system onchain (ONCHAINID)
```
lightclient, line 42: Please use a **bold** font for **MUST**.

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+- MUST define an Agent role and an Owner (token issuer) role
+- MUST be able to force transfers from an Agent wallet
+- MUST be able to issue transactions in batch (to save gas and to have all the transactions performed in the same block)
+- MUST be upgradeable (code of the smart contract should be upgradeable without changing the token smart contract address)
+
+## Rationale
+
+### Transfer Restrictions
+
+Transfers of securities can fail for a variety of reasons. This is in direct contrast to utility tokens, of which generally only require the sender to have a sufficient balance. <br>
```
lightclient, line 58:
```suggestion
Transfers of securities can fail for a variety of reasons. This is in direct contrast to utility tokens, of which generally only require the sender to have a sufficient balance.
```

lightclient: Please don't use `<br>`.

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Test Cases
+
+The standard is implemented and tested with full coverage on Tokeny's [T-REX repository](https://github.com/TokenySolutions/T-REX)
+
+## Security Considerations
+
+The suite of Smart Contracts has been audited by an external and independent company. The results can be found in [this document](https://tokeny.com/wp-content/uploads/2020/05/Tokeny-Solutions_T-REX-v3_Smart-Contract-Audit-Report_Kapersky.pdf).
+
+## Notes
```
lightclient, line 383: This is not a valid EIP section, please remove.

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+This standard is working completely on-chain. No off-chain validator is needed to perform transactions and token exchanges, thus introducing no external dependencies for the regulation of the exchanges.
+
+### KYC/AML Considerations
+
+Under some jurisdictions, a KYC/AML screening of one’s identity is necessary to hold tokens representing an underlying financial instrument. This standard relies heavily on identity contracts and validation through claims, enforcing the screening of identities before being able to hold tokens.
+
+### Cap-table immutability
+Through heavy identity integration, the standard is able to ensure the immutability of the cap-table on the blockchain. Since tokens are linked to identities instead of wallets, it is always possible to trace to whom the token belongs to, even in the case of lost wallets.
+
+## References
```
lightclient, line 396: This is not a valid EIP section, please remove.

EIPS/eip-42.md
```diff
@@ -0,0 +1,399 @@
...
+
+- [ONCHAINID documentation](https://docs.onchainid.com/)
+- [T-REX White Paper](https://tokeny.com/wp-content/uploads/2019/12/Whitepaper-T-REX-Security-tokens.pdf)
```
lightclient, line 399: Please be sure to include the copyright notice at the bottom as listed in the EIP template.

lightclient (changes requested):
Hi @Joachim-Lebrun, I left some feedback on your PR.

Joachim-Lebrun:
> Hi @Joachim-Lebrun, I left some feedback on your PR.

Hi, Thank you, i will take a look at it and update the md file accordingly!

EIPS/eip-42.md
wschwab: please add a `discussions-to: ` link (usually to a thread on https://ethereum-magicians.org) per eip-1

Joachim-Lebrun: discussion created here [EIP-3643 : Proposition of the T-REX token standard for securities](https://ethereum-magicians.org/t/eip-3643-proposition-of-the-t-rex-token-standard-for-securities/6844) i will update on the PR accordingly

Joachim-Lebrun:
normally, everything has been corrected following your recommendations @lightclient 
Thanks for your review

luziusmeisser:
I believe that this approach is fundamentally broken as it comingles concerns from different legal layer into one big monolithic technical blob. As an adverse side-effects, it makes basic operations like transfers significantly more expensive than a simple ERC-20 token. Furthermore, giving the issuer total control over the issued tokens does not only go against the spirit of decentralization, it might also violate investor protection laws designed to protect token holders from the issuer in case of conflicts of interest.

Here is a concrete example: let's assume that the founder of a successful company wants to use his tokenized shares in that company as a collateral to get a credit. He won't be able to a DAI-style DeFi protocol as they require the ability to take exclusive control over the deposited assets that serve as collateral. So he asks his bank for a credit. The bank would love to provide the credit, but their compliance department is reluctant to approve the collateral because they know that the founder has access to the admin key, enabling the founder to "steal" the collateral at any time. I say "steal" in quotes as such an action would not legally qualify as stealing, because the taken assets still belong to the founder when used as a collateral, so it is not exactly theft. Further, the bank's cyber insurance says they cannot provide insurance coverage for assets that are not truly under the control of the bank and the regulator is worried about the bank being able to auction off the asset in case of a default of the founder, so they do not recognize the collateral as risk-reducing when calculating capital requirements. All of this could be solved if the token offered the possibility to provide *exclusive control* over the token to the current holder.

As a side note, Swiss law explicitely recognizes the possibility to use crypto securities as a collateral, but it requires that it is technically ensured that the creditor has exclusive control over the crypto securities in case of a default. The reasoning behind this requirement are scenarios like the above.

So what is the correct way out?

The correct way out in my opinion is to depart from the monolithic "one size fits all" approach and to go for a layered design, distinguishing four layers.

image

The bottommost layer is the **infrastructure layer**. In our case, this would be the Ethereum blockchain, fulfilling basic requirements regarding data integrity, transparency, and so on. On top of that is the **register layer**, which essentially is an ERC-20 token that fulfills all constitutive legal requirements to represent a security token, but none of the circumstancial compliance rules that depend on the token holder, jurisdiction, involved intermediaries, etc. Legally, compliance rules are also added on a different layer than the basic rules that define what a security actually is and what it takes to create one. Also, compliance rules come and go whereas the basic rules about the issuance and transfer of securities last for decades. It makes sense to separate these different legal concerns also in the technical implementation. So the register layer ideally is a minimal ERC-20 contract with a few features added that are strictly necessary to make the tokens legally recognized securities (under Swiss law, the only feature that is missing in the ERC-20 standard is a link to a document called the "registration agreement" that specifies what the token actually represents and other general terms the tokens are subject to). Ideally, this base contract is not upgradable at all, potentially providing the token holders with a very high level security.

On top of the register layer, there is an **administrative layer** that adds all kind of administrative features, for example token recovery in case of a loss of private keys. Ideally, these features are added by wrapping the token and not by extending its functionality, making sure that it is possible to free the basic token on the register layer from all administrative backdoors if necessary, for example under the circumstances described in the example. Also, having this layer separated from the register layer creates a lot of flexibility, for example implementing a different set of rules for one jurisdiction than in another jurisdiction. One should also be careful to not call this layer "compliance" layer as "compliance" is usally used in the context of financial intermediation and AML rules, which in general do not apply directly to issuers. So the administrative layer is really reserved for token administration through the issuer, not for fulfilling the compliance requirements of individual custodians or other financial intermediaries that might hold the tokens at some point in time.

The latter requirements, the compliance rules of individual intermediaries, should be implemented on the **contractual layer**. Often, such rules are defined in the general terms applicable to the relation between an individual token holder and a financial intermediary. Even within jurisdictions, they might differ from intermediary to intermediary, for example requiring different levels of identity verification. All these kind of requirements are implemented on top of the other layers, usually without the involvement of the issuer. Also, this layer might contain smart contracts to automatically enforce shareholder agreements and the like.

It is important to have such a separation to cleanly deal with examples like the following (again coming from Swiss law, not sure how this works in other jurisdictions): assume an employee received tokenized shares from an issuer as part of an employee participation plan. The employee is not allowed to sell the shares according to his employment agreement, but does so anyway. The company refuses to execute the transfer of the tokens, saying that the transfer would violate the employment agreement. The empoyee goes to court and the court orders the company to approve the transfer because the refusal of a share transfer requires an according clause in the articles of association, which the company is lacking. Nonetheless, the employee is liable for violating the employment agreement. This example shows that the legal situation also has different layers, with the contractual agreements on upper layers not being able to supercede the applicable terms on lower layers. Of course, the company could have implemented an additional smart contract on the contractual layer to lock the shares of the employees and to enforce compliance with the employment agreement, but it is in general not allowed to use the terms of the register layer to enforce contractual agreements that have been made on a higher layer.

Let me know if you'd like to know more. The Swiss Blockchain Federation is working on an update to their Circular on Security Tokens, outlining these concerns in more detail.

Joachim-Lebrun:
Hi @luziusmeisser,
As you are probably  expecting, we cannot agree with you. We tried to provide a short and concise answer below to all the incongruences in your post:
- In regards of the law-related remarks, please note that the T-REX standard is already approved in its fundamental approach by all our customers lawyers (including some of the major European banks) and it has already been used to tokenise more than 8,5 billion € of assets across the globe. Also Tokeny Solutions, creator of the T-REX token standard is backed by Euronext who never rose any legal concerns of this kind. In facts what you have to understand is that the fundamental approach of T-REX is coming from years of discussion with the various actors of the industry, as well as decades of experience from our team members in the world of securities (stock exchanges, central banks, major CSDs, etc).
- When it comes to the gas use, we are well aware of the additional costs of such transactions, especially on Ethereum Main Net. The standard might not be cheap to use but if you had a better understanding of our customers (mostly handling private placements) and the traditional financial industry, you would know that it remains a much better options than already existing frameworks for many reasons. Also as it is developed in solidity it can be used on any EVM based blockchain, T-REX tokens are already deployed on the polygon network for instance when dealing with customer targeting retails and facing important number of transactions.
- Last but not least, T-REX is indeed a centralised and permissioned protocol deployed on top of a decentralized infrastructure. If you think it should be fully decentralized, we strongly advise you to DYOR about the legal obligations of issuers and agents of securities In Switzerland (as it seems to be your focus) and everywhere else in the world. Every T-REX token is deployed in parallel with a full legal paperwork describing the issuer and its agents obligations. Such actors are for most of them regulated by their respectives local authorities. They will not “steal” security tokens from investors. Wake up, we are not in DeFi here. Every actor is identified with a unique ONCHAINID and responsible for what they do, therefore it is not possible to “rugpull” a T-REX token.
Hopes it helps to bring a bit of light in your thoughts

luziusmeisser:
Hi @TokenySolutions, yes, it is no surprise that you are disagreeing. You are not in a position that would allow you to change your mind without calling into question all the work you and your company did in the past years.

I see that you talked to a lot of experts in the field of intermediated securities. And that's also what one can feel when reading your source code. It seems to be carefully crafted to cater towards the needs of banks, exchanges, securities depositories and other traditional securities. This is fine and I appreciate your effort to help them to tokenize the intermediated shares they are used to work with. However, in that case, you should be honest about naming this standard and call it "intermediated security standard" or the like.

Here are some statistics to put your proposal into perspective, again using the numbers from Switzerland I am familiar with. There are 230 Swiss companies listed on the stock market and a few more that have stored their shares with a CSD without having them listed on an exchange. These are the companies for which your standard can be useful. However, there are 600'000 Swiss companies in total, with 120'000 of them being public companies ("Aktiengesellschaften") with freely transferrable shares. And by freely transferrable I mean that their shares can be transferred without involving a notary or other official share register. Instead, the shareholder register typically is an excel sheet managed by the board of directors of each company and that is updated whenever a share transfer has taken place. For these 99.5% of all companies, the TREX standard should not be the preferred way of tokenization. Your standard seems very nice for use cases closely resembling the handling of intermediated securities, but it seems to be ill-suited for the vast majority of cases where there are no financial intermediaries involved and therefore the regulatory requirements are also very different.

--> To conclude: you are free to sell whatever blockchain-solution you want to banks and other traditional intermediaries seeking to be innovative, but you should not advertise your standard as a generally applicable tokenization standard for the vast majority of securities that exist without the involvement of a financial intermediary. Therefore, I propose to rename this EIP to "TREX **intermediated** security token standard".

Joachim-Lebrun:
@luziusmeisser Well, again we cannot agree with those statements, for the following reasons :
- We crafted this standard to be usable for all types of securities, private or public, freely transferrable or not, everything depends on the implementation that is done on the compliance smart contract that will basically determine what are the necessary checks to ensure the transfer is compliant with the legal paperwork related to the security token. Therefore in the case in which the securities are freely transferrable, the `canTransfer` function on the compliance contract simply returns `true` ... no more implementation required, that's what we call the "[default](https://github.com/TokenySolutions/T-REX/blob/master/contracts/compliance/DefaultCompliance.sol)" compliance contract on the T-REX implementation github.
- As the standard was developed to be usable by any security token issuer and for any type of asset, obviously some functions are not useful for all of them, especially regarding the compliance contract (that can be very complex or simply return `true` as explained before) or for the permissioning side. But if you can do more you can do less...
- Even if some securities are "freely" transferrable, there is still a need for KYCs on all stakeholders, that's the second part of our validation process, as the T-REX, in addition to the compliance rules (in the case of 100% freely transferrable securities there is none), is also checking the eligibility of the stakeholders, thanks to the ONCHAINID of each stakeholder and the KYC rules applied to this specific security token (basically the claim(s) required to be eligible and the claim issuers allowed to issue such claims)
- So, from what has been said above the only reason why the standard would not be the best in certain situations, where the compliance and eligibility rules are very light, is that it would use a bit more gas fees, for example to call the `canTransfer` function while the compliance always return `true` for this specific asset?  It is a very light concern in our opinion. As stated in the previous answer, in comparison to the existing state of securities, a small extra gas fee is not a huge issue, especially when you have the possibility to use this standard on other networks, such as polygon, binance smart chain, fantom, ...

luziusmeisser:
@TokenySolutions thanks for your reply. I take the opportunity to clear some misunderstandings regarding KYC and transaction costs.

Apparently, you are not a lawyer. Otherwise, you would know that the term "KYC" is not applicable to shareholders identifying themselves vis à vis the issuer. KYC means "know your client" and denotes the identity verification done by financial intermediaries with their clients. However, in general, issuers are not financial intermediaries and shareholders are not clients, so the term "KYC" is doubly wrong here. What you are probably referring to is the duty of shareholders to properly identify themselves and the duty of the company to keep a shareholder registry and an archive of all records related to that. The legal basis and requirements for the identification of shareholders by the issuer is entirely different than that for the identification of shareholders through financial intermediaries. Therefore, I would advise against using the term "KYC" in this context as it is not legally accurate.

Further, there are various choices for where the identification of shareholders takes place. You seem to be propagating an "onchain id" solution, with personal identifiers written onto the blockchain. From a data protection perspective, I personally find it more advisable to do the matching between blockchain-based addresses and personal identities off-chain. In this model, transactions happen exactly like traditional transfers with registered paper shares. Traditionally, registered paper shares are transferred as follows: first, the seller writes his name onto the share, signs it, and hands over the paper to the buyer. The buyer receives the paper certificate, shows it to the issuer and demands to be registered in the shareholder registry. If the issuer approves the transfer, the buyer is entered into the shareholder registry and is officially recognized as the new shareholder. If the issuer does not approve, the share has to be returned to the seller or sold on to a new buyer the isser approves of. Since blockchain based tokens share many properties of physical paper, this process can be adopted directly to the legal transfer of tokenized shares: the seller transfers them to the buyer, signing the transaction with his private key. The buyer then goes to the issuer, proving that he is the owner of the new address and providing all his personal data. If the issuer approves the transaction, the buyer is officially recognized as the new shareholder.

Note that this process is fully compliant with all applicable laws but much more light-weight than the process your token is imposing. If you prefer light-weight, off-chain verification of identities, the TREX standard seems not the best choice. Also, one should note that it is not possible to disable some of the questionable features such as the "forceTransfer" function, as that is baked into the token contract. Of course, the issuer can always promise to never abuse that function, but for some use-cases, it is desirable to be able to provide the holder exclusive control (I know you disagree here).

Further, you should note that the cost increase due to the overhead of the TREX contract is substiantial. Looking at the code, I would guess that transactions to new token holders will typically cost about three to four times as much as traditional ERC-20 transfers, even in the "minimal" configuration. Would be nice to have some actual measurements here.

Joachim-Lebrun:
@luziusmeisser I understand your point, but that’s not the way T-REX token works for several reasons :
- T-REX was built to use the blockchain as a primary registry for securities (it can also be used as a mirror of the primary registry that would be kept off-chain by authorized agents in case the jurisdiction doesn’t allow to use blockchain as a primary registry), that’s why transactions finality is necessary and that’s why we need to ensure all compliance & eligibility checks are done pre-transfer. T-REX is therefore a permissioned token as the transfer success is conditioned to the fulfilment of certain conditions, but once the conditions are met and the transfer is processed, the transaction is final and the issuer has all the necessary informations about token holders. Also we are glad to highlight the fact that in Luxembourg and many other EU countries, Blockchain can already be used as the primary registry for securities.
- The model you propose is not capable to ensure finality of a transaction in the way described above, and the permissionless nature comes with some drawbacks as well... with your system it is possible for securities to transit on the wallet of any individual, including (unfortunately) residents of banned countries (North Korea and others), terrorists, scammers, etc and even if those individuals cannot get the validation from the issuer to be included in the off-chain primary registry they can sell the security token to someone else and realise a profit from it.
- Regarding ONCHAINID, there is absolutely no privacy issue as we are not storing any private data onchain, only proofs of data verification and validation are issued onchain, in the form of claims issued by trusted third parties. ONCHAINID is a system of self-sovereign identity, the owner of the ONCHAINID contract has a full control over the data and can give/revoke access to some or all of them, upon request, to third parties. ONCHAINIDs are reusable on every T-REX token and on any application integrating ONCHAINID.
- For the management functions, yes indeed, management keys (owner & agent) have certain rights including mint, burn, freeze, forced transfers, yes it is very centralised, but it represents what the industry requires, especially if you consider blockchain as the primary registry. Note that it is possible to block some functions reserved to management by giving the role owner or agent to a smart contract instead of a wallet, and by implementing this smart contract in a way that this code is unreachable, that’s one of the purposes of permissioning contracts that we are currently developing and are available for the [owner role](https://github.com/TokenySolutions/T-REX/blob/develop/contracts/roles/OwnerManager.sol) and for the [agent role](https://github.com/TokenySolutions/T-REX/blob/develop/contracts/roles/AgentManager.sol).
- As previously stated, the gas fees are not a real issue for our clients, but we have some data to give you an example of the gas costs expected for a token with default compliance (returning true) and requiring 1 specific claim on the ONCHAINID issued by a specific claim issuer : the gas cost of a simple transfer is 142938 gas

luziusmeisser:
@TokenySolutions: thanks for the clarificiation. It seems to me you really should rename this standard, as it is not a standard designed for an open and permissionless setup, but rather a standard designed for intermediated securities that is typically deployed on permissioned blockchains.

Regarding the other points:

- Gas costs: thanks for confirming my guess. A plain ERC-20 transaction typically costs about 40'000 gas, so my guess of your system being three to four times more expensive even in the minimal configuration was spot on.

- Entry in the shareholder registry of the issuer is not required for finality of the transaction. When you trade on the stock market, transactions are also considered final long before your name has been put into the shareholder registry of the issuer. These are two independent steps. Therefore, also the system I described is capable to reach finality on the blockchain without relying on ONCHAINID.

- ONCHAINID is nice to prove that someone has provided his name and address with someone. But for a legally valid entry in the shareholder registry you actually need to actually have the name and address, not just a prove that it exists. Therefore, it seems to me your system also depends on an off-chain data to construct the shareholder registry. Same applies to compliance. Financial intermediaries are legally required to actually know their clients and not just have prove that their client was verified by someone else.

- Regarding the management keys: I prefer a system where the default state is free transferrability and not total centralized control. This also reflects the legal situation: here, the default state of things is that you can freely transfer a security and all the compliance and other restrictions are then added on top depending on the context in various specialized laws and regulations. This of course, requires a different token architecture with layers. Such an architecture would be capable of providing much stronger security and property guarantees.

- It is an illusion to think your level of control provides you with 100% certainty that no one ever sells a unit of your security to Osama bin Laden, Julien Assange, or another person not welcome in the financial system. First, it is always possible to transfer an ERC-20 token off-chain by simply handing over the private key to the next person. Of course, that person then would not be registered in the shareholder registry (neither it would be in our system), but that person would be perfectly capable of realizing the capital gains made while holding the private key. Second, the same problem exists when someone sends these tokens to a uniswap liquidity pool and then transfers the pool share. Third, the same problem also exists when the tokens are held by multisignature address and in many other setups. The idea that there is a 1:1 relation between persons and Ethereum-addresses is deeply flawed and breaks down in more elaborate setups. The right way to ensure that Osama bin Laden is not a shareholder is to refuse entry into the shareholder registry, thereby also exclusing him from dividend and voting rights. But it is impossible to guarantee that he can never hold a token or benefit from the share becoming more valuable. For example, someone could sell a terrorist an option to buy your share or a depository receipt. The issuer has no control over who the buyer is there and the buyer will clearly benefit from the share doing well. The goal here should be to deny the wrong people dividend and voting rights, and for that, it suffices to put the barrier at the entry into the shareholder registry, and not at the level of the transfer of the token.

It looks to me that there is a fundamental difference between our views. You want to put intermediated securities onto the blockchain and are aiming for a setup where complete centralized control is the default state. I am aiming for a free, open, inclusive and disintermediated financial system that is fundamentally decentralized, with only adding as little control and checks as legally required to be compliant with the law.

It might surprise you how high the costs of just a little bureacracy are in practice. A good example are the shares of Nestlé in the 80ies (see https://aktionariat.com/blog/efficency-of-whitelisting.html). Back then, all large Swiss companies had two types of shares listed on the stock market, easily transferrable bearer shares and a little less easily transferrable registered shares. For the transfer of the latter, some manual paperwork was necessary to actually execute the transfer, adding some extra cost. But apart from that, they were the same. The had the same nominal value, came with the same dividend, provided the same voting rights and where traded on the same exchange. The resulting price difference between the two was not 1% or 3% or so as one might expect. The freely transferrable shares traded almost twice as high as the less freely transferrable shares! After this difference persisted for more than 30 years, some legal changes made in the end of the 80ies, simplifying the transfer of registered shares, allowing the Nestlé and other Swiss companies to abandon bearer shares and merge them the registered shares. Once this decision was known, the price difference collapsed in the market. Nonetheless, I was very impressed to see what a big impact on the price of a security a simple registration requirement can have. Note that this was not about money laundering or criminals holding these shares, as all of them had to go through the same banking system to hold the shares. It was just about filling in a paper form and coming with somewhat higher deposit fees. Most people underestimate how much economic value is destroyed by unnecessary paperwork. So we should try to reduce it to a minimum, and not pro-actively endorse it.

Maybe we should have two different standards then. One for permissioned securities and one for freely transferrable securities.

Joachim-Lebrun:
@luziusmeisser Thanks for all your comments and our discussion of today. I understand what you are proposing, and as you said the standard that you are proposing for freely transferrable security tokens is fundamentally different from T-REX tokens and each approach has its pros and cons. I would like to highlight the following points to summarise :
- T-REX tokens are permissioned tokens, controlled by the issuer or authorized agents.
-They can represent/mirror the securities on the blockchain or be the native securities. In your proposition it is different as the tokens are representing securities are exchanged freely, using the token as a secondary registry, with a need to contact the issuer to modify the shareholders registry.
- Note that the 2 models are not incompatible, in the case it is possible (legally speaking) to use the freely transferrable tokens, it is possible to design a smart contract that would be used to wrap T-REX into freely transferrable tokens, T-REX remaining the primary registry of the security but with the possibility to use freely transferrable tokens for more convenience on the secondary market, but i think it is definitely another topic and i would like to keep discussing about T-REX and only T-REX on the comments feed of this PR
- As the 2 models are fundamentally different and not incompatible i believe it would be good if you propose another EIP for the freely transferrable security tokens on your side. 

I wish everything is clear now, for additional comments you can use the [discussion thread ](https://ethereum-magicians.org/t/eip-3643-proposition-of-the-t-rex-token-standard-for-securities/6844) as i feel the PR is becoming a bit overcharged with all the long comments... Thanks for your feedbacks

luziusmeisser:
Also thanks for the discussion, was interesting to hear what you background is and what you are working on.

EIPS/eip-3643.md
```diff
@@ -0,0 +1,383 @@
+---
+eip: 3643
+title: T-REX - Token for Regulated EXchanges
+authors: Joachim Lebrun (@Joachim-Lebrun), Tony Malghem (@TonyMalghem), Kevin Thizy (@Nakasar), Luc Falempin (@lfalempin) , Adam Boudjemaa (@Aboudjem)
+type: Standards Track
+category: ERC
+status: Draft
+require: 20 , 734, 735, 1822
```
lightclient, line 8:
```suggestion
require: 20, 734, 735, 1822
```

EIPS/eip-3643.md
```diff
@@ -0,0 +1,382 @@
+---
+eip: 3643
+title: T-REX - Token for Regulated EXchanges
+description: An institutional grade security token standard that provides interfaces for the management and compliant transfer of security tokens.
+authors: Joachim Lebrun (@Joachim-Lebrun), Tony Malghem (@TonyMalghem), Kevin Thizy (@Nakasar), Luc Falempin (@lfalempin), Adam Boudjemaa (@Aboudjem)
```
lightclient, line 5:
```suggestion
author: Joachim Lebrun (@Joachim-Lebrun), Tony Malghem (@TonyMalghem), Kevin Thizy (@Nakasar), Luc Falempin (@lfalempin), Adam Boudjemaa (@Aboudjem)
```

EIPS/eip-3643.md
```diff
@@ -0,0 +1,382 @@
...
+type: Standards Track
+category: ERC
+status: Draft
+require: 20, 734, 735, 1822
```
lightclient, line 9:
```suggestion
requires: 20, 734, 735, 1822
```

EIPS/eip-3643.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Simple Summary
+
+The T-REX token is 
+This standard  
+
+
+## Abstract
+
+Standards should be backwards compatible with [ERC-20](./eip-20.md) and should be able to interact with [ERC-735](./eip-735.md) to validate the claims linked to an [`ONCHAINID`](https://github.com/onchain-id/solidity), based on [ERC-734](./eip-734.md) and ERC-735.
```
lightclient, line 22:
```suggestion
Standards should be backwards compatible with [ERC-20](./eip-20.md) and should be able to interact with [ERC-735](https://github.com/ethereum/EIPs/issues/735) to validate the claims linked to an [`ONCHAINID`](https://github.com/onchain-id/solidity), based on [ERC-734](https://github.com/ethereum/EIPs/issues/734) and ERC-735. 
```

EIPS/eip-3643.md
```diff
@@ -0,0 +1,382 @@
+---
+eip: 3643
+title: T-REX - Token for Regulated EXchanges
+description: An institutional grade security token standard that provides interfaces for the management and compliant transfer of security tokens.
+author: Joachim Lebrun (@Joachim-Lebrun), Tony Malghem (@TonyMalghem), Kevin Thizy (@Nakasar), Luc Falempin (@lfalempin), Adam Boudjemaa (@Aboudjem)
+type: Standards Track
+category: ERC
+status: Draft
+requires: 20, 734, 735, 1822
```
lightclient, line 9:
```suggestion
requires: 20, 1822
```

lightclient: Not valid EIPs.

lightclient (approved):

PR #3668, opened 2021-07-19
Initial version of Durin spec

Arachnid:


EIPS/eip-3668.md
```diff
@@ -0,0 +1,184 @@
...
+
+Contracts must also implement sufficient validation of the data returned by the gateway to ensure it is valid. The validation required is application-specific and cannot be specified on a global basis.
+
+Client libraries implementing Durin may want to ensure that callers have to explicitly enable its use on a contract, so users are not surprised when their client library begins making HTTPS calls to external services.
+
+## Glossary
+ - Client: A process, such as JavaScript executing in a web browser, or a backend service, that wishes to query a blockchain for data. The client understands how to fetch data using Durin.
+ - Contract: A smart contract existing on Ethereum or another blockchain.
+ - Gateway: A service that answers application-specific Durin queries, usually over HTTPS.
+
```
MicahZoltu, line 182:
```suggestion
```
This isn't a valid top level EIP section.  Consider folding this into one of the standard sections like the specification (if these terms are referenced in the specification, this **SHOULD** be part of the specification).

EIPS/eip-3668.md
```diff
@@ -0,0 +1,184 @@
+---
+eip: 3668
+title: "Durin: Secure offchain data retrieval"
+author: Nick Johnson (@arachnid)
+discussions-to: <URL>
```
MicahZoltu, line 5: Make sure to fill this in before unmarking this as a draft PR.

MicahZoltu:
Will wait on a formal review until this is Ready for Review, though there is a bug right now in the bot that will cause it to auto-merge since @Arachnid is an EIP-editor.  Feel free to mention me in a comment if you want me to do a formal review prior to you marking it as Ready for Review so when it does auto-merge it at least has a review.

EIPS/eip-3668.md
Arachnid: EIP 1 specifies which sections an EIP *should* have. It doesn't say it's an exhaustive list!

If you strongly feel the glossary should be part of the specification section, I can move it - but that's not where glossaries are commonly put.

MicahZoltu: For a while now we have been enforcing that sections match the template because people were creating all sorts of random sections and EIP documents were becoming quite non-standard looking.

To get ahead of your next question: yes, we should update EIP-1 to match the effective rules.  However, I have been lazy about that because the *only* people who ever refer to EIP-1 when I suggest things are ex-EIP editors (you, Alex, Greg).  I don't think any regular users actually read EIP-1 so updating it would basically be a formality to get you three to do what everyone else does.  😄  I recognize that this is very much not an ideal situation, and I have been trying (along with others) to get more EIP-editors so we can tackle these problems but finding editors is hard.  😢

Arachnid: I'm skeptical that enforcing this restriction is actually useful. Would this spec be more readable if the glossary was a subheading of 'specification' to be technically compliant with the unwritten rules? This seems an awful lot like enforcing a rule because it's there rather than because it improves anything.

MicahZoltu: It is useful when you are optimizing for minimum amount of EIP Editor effort rather than maximum EIP quality, which is the current state of things.  At the moment, the *actual* options on the table are:
1. Strict but very easy/low cost to enforce rules that are uniformly applied.
2. No EIP editing happens at all.

It would be *great* if we could get to a more healthy place with regards to EIP editing where we had enough editor-hours available to be more thoughtful and nuanced in the review process, but unfortunately that isn't the world we live in right now.

The specific problem here is that if I say OK to you, I'll soon have people saying "but you let Nick submit an EIP with a non-standard section, so why can't I" (yes, this actually does happen when exceptions are made) and then I have to have a likely long nuanced debate with an author about why Nick's extraneous section was acceptable while theirs was not.  I think such conversations would be very educational and valuable for everyone involved, if we had an unbounded amount of editor-hours available to have them, but we do not sadly.

Arachnid:
PTAL. I think this unwritten rule requiring no sections not mentioned in EIP-1 is pointless, and in cases like this is making EIPs worse for no good reason. I don't see the harm in simply allowing people to create drafts with whatever additional sections they like; if the sections are bad the EIP won't make it to or past Last Call.

Nevertheless I've moved the section as requested in the interest of getting this merged and not further wasting EIP editors time with arguments.

MicahZoltu:
> if the sections are bad the EIP won't make it to or past Last Call.

What makes you think that a bad EIP wouldn't make it to/past Last Call?  Other than editors blocking on hard-line rules like this, there is almost nothing stopping an EIP from making it to the end of the process.  If we drop this rule, there will *definitely* be EIPs with bad sections that make it to final.

MicahZoltu (approved):

EIPS/eip-3668.md
```diff
@@ -0,0 +1,186 @@
...
+
+`prefix` specifies a prefix that the data returned by the gateway at `url` must have. Clients will check this value is a prefix of the data returned by the gateway, and cause the call to fail if it does not match.
+
+The contract must also implement a callback method for decoding and validating the data returned by the gateway. The name and signature of this method are implementation-specific and must be agreed upon between the contract and the gateway. If the client successfully calls the gateway, this contract function will be invoked by the client, with the callback data returned from the gateway. The contract must interpret this data and use it to validate the original request, returning the result of that request if successful, or reverting if validation fails.
+
+`prefix` provides a critical security measure against malicious gateways; by requiring the response returned by the gateway to start with a specific prefix, the contract can prevent the gateway from causing the client to call arbitrary methods on the contract, and require that the answer pertains to the original query, by including query parameters in the requried prefix.
```
MicahZoltu, line 60:
```suggestion
`prefix` provides a critical security measure against malicious gateways; by requiring the response returned by the gateway to start with a specific prefix, the contract can prevent the gateway from causing the client to call arbitrary methods on the contract, and require that the answer pertains to the original query, by including query parameters in the required prefix.
```

EIPS/eip-3668.md
```diff
@@ -0,0 +1,186 @@
...
+
+While specifying the call type in the ABI is a possible solution, this makes retrofitting existing interfaces to support offchain data awkward, and either results in contracts with the same name and arguments as the original specification, but with different return data - which will cause decoding errors for clients that do not expect this - or duplicating every function that needs support for offchain data with a different name (eg, `balanceOf -> offchainBalanceOf`). Neither solutions is particularly satisfactory.
+
+Using a revert, and conveying the `url` and `prefix` in the revert data, allows any function to be retrofitted to support lookups via Durin so long as the client understands the specification, and so facilitates translation of existing specificatiosn to use offchain data.
+
+### Passing address and abi data to `durin_call`
+Both of these parameters are required to be passed to the gateway in order to faciliate the writing of generic gateways, thus reducing the burden on contract authors to provide their own gateway implementations.
```
MicahZoltu, line 151:
```suggestion
Both of these parameters are required to be passed to the gateway in order to facilitate the writing of generic gateways, thus reducing the burden on contract authors to provide their own gateway implementations.
```

MicahZoltu (changes requested):
CI failing on a couple spelling errors.

Arachnid:
> What makes you think that a bad EIP wouldn't make it to/past Last Call? Other than editors blocking on hard-line rules like this, there is almost nothing stopping an EIP from making it to the end of the process. If we drop this rule, there will definitely be EIPs with bad sections that make it to final.

If we can't ensure that anything other than simple typographical rules are enforced on EIPs, then the EIP process is broken. I don't think adding an unwritten rule about what top-level headers are permissible will make the process noticeably better.

I've fixed the spelling errors, PTAL.

MicahZoltu:
> then the EIP process is broken

I agree!  Unfortunately, the fix requires human resources, of which we have scant few.  😢

PR #3672, opened 2021-07-22
ENS offchain resolution EIP

Arachnid:


EIPS/eip-3672.md
```diff
@@ -0,0 +1,72 @@
...
+category: ERC
+created: 2021-07-21
+requires: 137, 3668
+---
+
+## Simple Summary
+Specifies an updated resolution process for ENS that supports offchain resolution and wildcard lookups.
+
+## Abstract
+Instead of using the resolution process from EIP 137, clients should start with the provided name and check successively shorter suffixes until they find a name with a resolver. Then, they should use the process described in EIP 3668 to query the resolver.
```
lightclient, line 17:
```suggestion
Instead of using the resolution process from [EIP-137](./eip-137.md), clients should start with the provided name and check successively shorter suffixes until they find a name with a resolver. Then, they should use the process described in [EIP-3668](./eip-3668) to query the resolver.
```

lightclient: If possible, I think it's better for this info (e.g. how this EIP relates to other things) to be in the motivation and leave the abstract as strictly focused on what this EIP does.

EIPS/eip-3672.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Motivation
+Making ENS scalable requires providing ways to create and update ENS names that don't incur L1 transaction fees. Rather than picking one L2 and moving all of ENS to it, this EIP utilises EIP 3668 to support resolving records held on any external system, without introducing new trust assumptions (unless the owner of the name desires it). Combined with a new process for finding resolvers, this allows for moving entire subtrees of the ENS hierarchy to external systems.
```
lightclient, line 20:
```suggestion
Making ENS scalable requires providing ways to create and update ENS names that don't incur L1 transaction fees. Rather than picking one L2 and moving all of ENS to it, this EIP utilises EIP-3668 to support resolving records held on any external system, without introducing new trust assumptions (unless the owner of the name desires it). Combined with a new process for finding resolvers, this allows for moving entire subtrees of the ENS hierarchy to external systems.
```

lightclient (changes requested):
Couple nits - mainly waiting for a `discussions-to` link.

PR #3675, opened 2021-07-22
EIP-3675: Upgrade consensus to Proof-of-Stake

mkalinin:
Specification of the consensus mechanism upgrade on Ethereum Mainnet that introduces Proof-of-Stake (a.k.a. The Merge)

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -1,0 +1,244 @@
+---
+eip: 3675
+title: Upgrade consensus to Proof-of-Stake
+author: Mikhail Kalinin (@mkalinin), Danny Ryan (@djrtwo), Vitalik Buterin (@vbuterin)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-07-22
+requires: 2982
+---
+
+
+## Simple Summary
+
+Specification of the consensus mechanism upgrade on Ethereum Mainnet that introduces Proof-of-Stake.
+
+
+## Abstract
+
+This EIP deprecates Proof-of-Work (PoW) and supersedes it with the new Proof-of-Stake consensus mechanism (PoS) driven by the beacon chain. Information on the bootstrapping of the new consensus mechanism is documented in [EIP-2982](/EIPS/eip-2982). Full specification of the beacon chain can be found in [eth2.0-specs](https://github.com/ethereum/eth2.0-specs/) repository.
+
+This document specifies the set of changes to the block structure, block processing, the fork choice rule and the network interface introduced by the consensus upgrade.
+
+
+## Motivation
+
+The beacon chain network has been up and running since December 2020. Neither safety nor liveness failures were detected during this period of time. The long period of running without failures demonstrates sustainability of the beacon chain system and witnesses its readiness to start driving and become a security provider for the Ethereum Mainnet.
+
+To understand the motivation of introducing the Proof-of-Stake consensus see Motivation section of [EIP-2982](/EIPS/eip-2982#motivation).
+
+
+## Specification
+
+### Definitions
+
+* **PoW block**: Block that is built and verified by the existing proof-of-work mechanism. In other words, a block of the Ethereum network before the consensus upgrade.
+* **PoS block**: Block that is built and verified by the new proof-of-stake mechanism.
+* **Terminal PoW block**: A PoW block that satisfies the following conditions -- 
+`pow_block.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY` *and* `pow_block.parent_block.total_difficulty < TERMINAL_TOTAL_DIFFICULTY`.  
+There can be more than one terminal PoW block in the network, e.g. multiple children of the same pre-terminal block.
+* **`TERMINAL_TOTAL_DIFFICULTY`** The amount of total difficulty reached by the network that triggers the consensus upgrade.
+* **`TRANSITION_BLOCK`** The earliest PoS block of the canonical chain, i.e. the PoS block with the lowest block height.  
+`TRANSITION_BLOCK` **MUST** be a child of a terminal PoW block.
+* **`POS_CONSENSUS_VALIDATED`** An event occurring when the PoS block is validated with respect to the proof-of-stake consensus rules.
+* **`POS_CHAINHEAD_SET`** An event occurring when the PoS block is designated as the head of the canonical chain.
+* **`POS_BLOCK_FINALIZED`** An event occurring when the PoS block is considered as finalized.
+
+#### PoS events
+
+Events having the `POS_` prefix in the name (PoS events) are emitted by the new proof-of-stake consensus mechanism. They signify the corresponding assertion that has been made regarding a block specified by the event. The underlying logic of PoS events can be found in the beacon chain specification. On the occurence of each PoS event the corresponding action that is specified by this EIP **MUST** be taken.
+
+Each `POS_BLOCK_FINALIZED` event has a reference to the head of the canonical chain in addition to the finalized block and incorporates the semantics of the `POS_CHAINHEAD_SET` event. Thus, every action and effect that is specified on the occurrence of `POS_CHAINHEAD_SET` event according to this EIP **MUST** also be taken on every occurrence of `POS_BLOCK_FINALIZED` event.
+
+
+### PoW block processing
+
+PoW blocks that are descendants of any terminal PoW block **MUST NOT** be imported. This implies that a terminal PoW block will be the last PoW block in the canonical chain.
+
+
+### Block structure
+
+Beginning with `TRANSITION_BLOCK`, a number of previously dynamic block fields are deprecated by enforcing these values to instead be constants. Each block field listed in the table below **MUST** be replaced with the corresponding constant value.
+
+| Field | Constant value |
+|-|-|
+| **`ommersHash`** | `Keccak256(RLP(EMPTY_LIST))`  |
+| **`difficulty`** | `0` |
+| **`extraData`** | `EMPTY_LIST` |
+| **`mixHash`** | `0x0000000000000000000000000000000000000000000000000000000000000000` |
+| **`nonce`** | `0x0000000000000000` |
+| **`ommers`** | `EMPTY_LIST` |
+
+*Note*: `EMPTY_LIST` denotes `[]` respecting the semantics of the corresponding field.
+
+*Note*: Logic and validity conditions of block fields that are *not* specified here **MUST** remain unchanged. Additionally, the overall block format **MUST** remain unchanged.
+
+
+### Block validity
+
+Beginning with `TRANSITION_BLOCK`, the block validity conditions **MUST** be altered by the following:
+* Remove verification of the block's **`difficulty`** value with respect to the difficulty formula.
+* Remove verification of the block's **`nonce`** and **`mixHash`** values with respect to the Ethash function.
+* Remove all validation rules that are evaluated over the list of ommers and each member of this list, except for enforcing this list to be empty.
+* Add verification of the fields noted in [block structure](#Block-structure) as their specified constant value, including the enforcement of empty ommers list.
+* Add verification of block validity with respect to the PoS consensus rules, i.e. assert that there **MUST** be a corresponding event for the block to consider it valid. Such events are limited to the following:
+    * A `POS_CONSENSUS_VALIDATED` event considering the block or any of its descendants as valid.
+    * A `POS_CHAINHEAD_SET` event, including the `POS_BLOCK_FINALIZED` form of this event, designating the block or any of its descendants as the head of the canonical chain.
+
+*Note*: If one of the new rules fails then the block **MUST** be invalidated.
+
+*Note*: Validity rules that are not specified in the list above **MUST** remain unchanged.
+
+
+### Block and ommer rewards
+
+Beginning with `TRANSITION_BLOCK`, block and ommer rewards are deprecated. Specifically, the following actions **MUST** be taken:
+* Remove increasing the balance of the block's **`beneficiary`** account by the basic block reward.
+* Remove increasing the balance of the block's **`beneficiary`** account by the ommer inclusion reward per each ommer.
+* Remove increasing the balance of the ommer's **`beneficiary`** account by the ommer block reward per each ommer.
+
+*Note*: Transaction fee mechanics affecting the block's `beneficiary` account **MUST** remain unchanged.
+
+
+### Fork choice rule
+
+As of the first `POS_CHAINHEAD_SET` event, the fork choice rule **MUST** be altered in the following way: 
+* Remove the existing PoW heaviest chain rule.
+* Adhere to the new PoS LMD-GHOST rule. 
+
+The new PoS LMD-GHOST fork choice rule is specified as follows. On each occurrence of a `POS_CHAINHEAD_SET` event including the first one, the following actions **MUST** be taken:
+* Consider the chain starting at genesis and ending with the block nominated by the event as the canonical blockchain.
+* Set the head of the canonical blockchain to the block nominated by the event.
+
+
+### Network
+
+The networking stack **SHOULD NOT** send the following messages if they advertise the descendant of any terminal PoW block:
+* [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
+* [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
+
+Beginning with the first `POS_BLOCK_FINALIZED` event, the networking stack **MUST** remove the handlers corresponding to the following ETH protocol messages:
+* [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
+* [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
+
+
+## Rationale
+
+The changes specified in this EIP target a minimal requisite set of consensus and client software modifications to safely replace the existing proof-of-work consensus algorithm with the new proof-of-stake consensus represented by the already in-production beacon chain.
+
+Minimalism in this changeset was chosen to reduce the complexity of the unprecedented task of hot-swapping the live consensus of the Ethereum network in order both to increase  the safety of the operation and to reduce the time to production. Additionally, a minimal changeset helps ensure that *most* smart contracts and services will continue to function as intended during and after the transition with little to no required intervention.
+
+### Total difficulty triggering the upgrade
+
+See [Security considerations](#Terminal-total-difficulty-vs-block-number).
+
+### Halting the import of PoW blocks
+
+See [Security considerations](#Halt-the-importing-of-PoW-blocks).
+
+### Replacing block fields with constants
+
+Deprecated block fields are replaced with constant values to ensure the block format remains backwards compatible. Preserving the block format aids existing smart contracts and services in providing uninterrupted service during and after this transition.
+
+This change introduces an additional validity rule that enforces the replacement of deprecated block fields.
+
+### Replacing `difficulty` with `0`
+
+After deprecating the proof-of-work the notion of difficulty no longer exists and replacing the block header **`difficulty`** field with `0` constant is semantically sound.
+
+### Deprecating `extraData`
+
+The **`extraData`** field is deprecated in favour of the corresponding **`graffiti`** field of the beacon block.
+
+### Changing block validity rules
+
+The rule set enforcing the PoW seal validity is replaced with the corresponding PoS rules along with the consensus upgrade as the rationale behind this change.
+
+An additional rule validating a set of deprecated block fields is required by the block format changes introduced by this specification.
+
+### Removing block rewards
+
+Existing rewards for producing and sealing blocks are deprecated along with the PoW mechanism. The new PoS consensus becomes both responsible for sealing blocks and for issuing block rewards once this specification enters into effect.
+
+### Supplanting fork choice rule
+
+The fork choice rule of the PoW mechanism becomes completely irrelevant after the upgrade and is replaced with the corresponding rule of the new PoS consensus mechanism.
+
+### Removing block gossip
+
+After the upgrade of the consensus mechanism only the beacon chain network will have enough information to validate a block. Thus, block gossip provided by the `eth` network protocol will become unsafe and is deprecated in favour of the block gossip existing in the beacon chain network. 
+
+It is recommended for the client software to not propagate descendants of any terminal PoW block to reduce the load on processing the P2P component and stop operating in the environment with unknown security properties.
+
+## Test Cases
+*TBD*
+
+## Reference Implementation
+*TBD*
+
+
+## Backwards Compatibility
+
+This EIP introduces backward incompatibilities in block validity, block rewards and fork choice rule.
+
+The design of the consensus upgrade specified by this document does not introduce backward incompatibilities for existing applications and services built on top of Ethereum except for those that are described in the [EVM](#EVM) section below or heavily depends on the PoW consensus in any other way.
+
+
+### EVM
+
+Although this EIP does not introduce any explicit changes to the EVM there are a couple of places where it may affect the logic of existing smart contracts.
+
+#### DIFFICULTY
+
+`DIFFICULTY` operation will always return `0` after this EIP takes effect and deprecates the **`difficulty`** field by replacing it with `0` constant.
+
+*Note:* Altering the `DIFFICULTY` semantics to return randomness accumulated by the beacon chain is under consideration but will be introduced in a separate EIP.
+
+#### BLOCKHASH
+
+Pseudo-random numbers obtained as the output of `BLOCKHASH` operation become more insecure after this EIP takes effect and the PoW mechanism (which decreases the malleability of block hashes) gets supplanted by PoS.
+
+
+## Security Considerations
+
+### Beacon chain
+
+See Security Considerations section of [EIP-2982](/EIPS/eip-2982#security-considerations).
+
+
+### Transition process
+
+The transition process used to take this specification into effect is a more sophisticated version of a hardfork -- the regular procedure of applying backwards incompatible changes in the Ethereum network. This process has multiple successive steps instead of the normal block-height point condition of simpler hardforks.
+
+The complexity of this upgrade process stems from this fork targeting the underlying consensus mechanism rather than the application layer within the consensus mechanism. Although the design seeks simplicity where possible, safety and liveness considerations during this transition have been prioritized.
+
+#### Terminal total difficulty vs block number
+
+Using a pre-defined block number for the hardfork is unsafe in this context due to the PoS fork choice taking priority during the transition.
+
+An attacker may use a minority of hash power to build a malicious chain fork that would satisfy the block height requirement. Then the first PoS block may be maliciously proposed on top of the PoW block from this adversarial fork, becoming the head and subverting the security of the transition.
+
+To protect the network from this attack scenario, difficulty accumulated by the chain (total difficulty) is used to trigger the upgrade.
+
+#### Ability to jump between terminal PoW blocks
+
+There could be the case when a terminal PoW block is not observed by the majority of network participants due to (temporal) network partitioning. In such a case, this minority would switch their fork choice to the new rule provided by the PoS rooted on the minority terminal PoW block that they observed.
+
+The transition process allows the network to re-org between forks with different terminal PoW blocks as long as (a) these blocks satisfy the terminal PoW block conditions and (b) a `POS_BLOCK_FINALIZED` has not yet been received. This provides resilience against adverse network conditions during the transition process and prevents irreparable forks/partitions.
+
+#### Halt the importing of PoW blocks
+
+Suppose the part of the client software that is connected to the beacon chain network goes offline before the Ethereum network reaches the `TERMINAL_TOTAL_DIFFICULTY` and stays offline while the network meets this threshold. Such an event makes the client software unable to switch to PoS and allows it to keep following the PoW chain if this chain is being built beyond the terminal PoW block. Depending on how long the beacon chain part was offline, it could result in different adverse effects such as:
+* The client has no post-state for the terminal PoW block (the state has been pruned) which prevents it from doing the re-org to the PoS chain and leaving syncing from scratch as the only option to recover.
+* An application, a user or a service uses the data from the wrong fork (PoW chain that is kept being built) which can cause security issues on their side.
+
+Not importing PoW blocks that are beyond the terminal PoW block prevents these adverse effects on safety/re-orgs in the event of software or configuration failures *in favor* of a liveness failure.
+
+### Further threat analysis
+_TBD_
+
+
+## Copyright
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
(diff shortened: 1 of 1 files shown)


EIPS/eip-3675.md
```diff
@@ -0,0 +1,244 @@
...
+This EIP deprecates Proof-of-Work (PoW) and supersedes it with the new Proof-of-Stake consensus mechanism (PoS) driven by the beacon chain. Information on the bootstrapping of the new consensus mechanism is documented in [EIP-2982](/EIPS/eip-2982). Full specification of the beacon chain can be found in [eth2.0-specs](https://github.com/ethereum/eth2.0-specs/) repository.
+
+This document specifies the set of changes to the block structure, block processing, the fork choice rule and the network interface introduced by the consensus upgrade.
+
+
+## Motivation
+
+The beacon chain network has been up and running since December 2020. Neither safety nor liveness failures were detected during this period of time. The long period of running without failures demonstrates sustainability of the beacon chain system and witnesses its readiness to start driving and become a security provider for the Ethereum Mainnet.
+
+To understand the motivation of introducing the Proof-of-Stake consensus see Motivation section of [EIP-2982](/EIPS/eip-2982#motivation).
```
poojaranjan, line 30:
```suggestion
To understand the motivation of introducing the Proof-of-Stake consensus see Motivation section of [EIP-2982](/EIPS/eip-2982).
```

EIPS/eip-3675.md
```diff
@@ -0,0 +1,244 @@
...
+#### BLOCKHASH
+
+Pseudo-random numbers obtained as the output of `BLOCKHASH` operation become more insecure after this EIP takes effect and the PoW mechanism (which decreases the malleability of block hashes) gets supplanted by PoS.
+
+
+## Security Considerations
+
+### Beacon chain
+
+See Security Considerations section of [EIP-2982](/EIPS/eip-2982#security-considerations).
```
poojaranjan, line 208:
```suggestion
See Security Considerations section of [EIP-2982](/EIPS/eip-2982).
```

EIPS/eip-3675.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Removing block rewards
+
+Existing rewards for producing and sealing blocks are deprecated along with the PoW mechanism. The new PoS consensus becomes both responsible for sealing blocks and for issuing block rewards once this specification enters into effect.
+
+### Supplanting fork choice rule
+
+The fork choice rule of the PoW mechanism becomes completely irrelevant after the upgrade and is replaced with the corresponding rule of the new PoS consensus mechanism.
+
+### Removing block gossip
```
rauljordan, line 169: I think there should be a mention that there should still be tx gossip to avoid thinking that we should remove gossip on eth1 altogether

GregTheGreek:
:rocket: :rocket:

EIPS/eip-3675.md
olibearo:
```suggestion
To understand the motivation of introducing the Proof-of-Stake consensus see the Motivation section of [EIP-2982](/EIPS/eip-2982#motivation).
```

abcoathup: @poojaranjan why remove the anchors?  I find them helpful.

EIPS/eip-3675.md
anand-chat: is it possible to preserve the anchors?

MrUke:
Let's make history 🚀🚀

SjonHortensius (approved):

EIPS/eip-3675.md
mkalinin: Added a note that the logic of other message handlers must remain unchanged

EIPS/eip-3675.md
mkalinin: I would leave them as well. What is the reason to remove them?

EIPS/eip-3675.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Block structure
+
+Beginning with `TRANSITION_BLOCK`, a number of previously dynamic block fields are deprecated by enforcing these values to instead be constants. Each block field listed in the table below **MUST** be replaced with the corresponding constant value.
+
+| Field | Constant value |
+|-|-|
+| **`ommersHash`** | `Keccak256(RLP(EMPTY_LIST))`  |
+| **`difficulty`** | `0` |
+| **`extraData`** | `EMPTY_LIST` |
```
holiman, line 69: `extraData` is a `bytes`, no? So if  you want it empty, it should not be an rlp-list, IIRC, but rather a `0`-length byte array

EIPS/eip-3675.md
```diff
@@ -0,0 +1,246 @@
+---
+eip: 3675
+title: Upgrade consensus to Proof-of-Stake
+author: Mikhail Kalinin (@mkalinin), Danny Ryan (@djrtwo), Vitalik Buterin (@vbuterin)
+discussions-to: https://ethereum-magicians.org/t/eip-3675-upgrade-consensus-to-proof-of-stake/6706
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-07-22
+requires: 2982
```
axic, line 10: Meta question 1: can a core EIP depend on an informational EIP?

Meta question 2: should 2982 be finalised first?

EIPS/eip-3675.md
```diff
@@ -0,0 +1,246 @@
...
+This EIP deprecates Proof-of-Work (PoW) and supersedes it with the new Proof-of-Stake consensus mechanism (PoS) driven by the beacon chain. Information on the bootstrapping of the new consensus mechanism is documented in [EIP-2982](/EIPS/eip-2982). Full specification of the beacon chain can be found in [eth2.0-specs](https://github.com/ethereum/eth2.0-specs/) repository.
+
+This document specifies the set of changes to the block structure, block processing, the fork choice rule and the network interface introduced by the consensus upgrade.
+
+
+## Motivation
+
+The beacon chain network has been up and running since December 2020. Neither safety nor liveness failures were detected during this period of time. The long period of running without failures demonstrates sustainability of the beacon chain system and witnesses its readiness to start driving and become a security provider for the Ethereum Mainnet.
+
+To understand the motivation of introducing the Proof-of-Stake consensus see the Motivation section of [EIP-2982](/EIPS/eip-2982#motivation).
```
axic, line 30:
```suggestion
To understand the motivation of introducing the Proof-of-Stake consensus see the Motivation section of [EIP-2982](./eip-2982.md#motivation).
```

This should refer to a filename.  I am not sure if that supports anchors or not, but I have nothing against the anchor (probably it does not support the anchor though).

EIPS/eip-3675.md
MicahZoltu: The answer to the second is that it this just can't surpass the status of its dependency.  So if the dependency is in Last Call, this can only go up to Last Call.

For the first, I'm not a fan of a Core EIP depending on an Informational EIP, but I have largely lost that debate at this point I think.

EIPS/eip-3675.md
MicahZoltu: According to https://eth.wiki/fundamentals/specs/consensus-datastructures, which may be wrong, it is a `UInt256`, which is serialized as a leading-zero-truncated `byte[]`.  If that linked page is wrong we should update it as well as this EIP.

EIPS/eip-3675.md
lightclient: I don't have a strong opinion either way.

EIPS/eip-3675.md
```diff
@@ -0,0 +1,246 @@
...
+* [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
+
+*Note:* The logic of message handlers that are not affected by this section **MUST** remain unchanged.
+
+
+## Rationale
+
+The changes specified in this EIP target a minimal requisite set of consensus and client software modifications to safely replace the existing proof-of-work consensus algorithm with the new proof-of-stake consensus represented by the already in-production beacon chain.
+
+Minimalism in this changeset was chosen to reduce the complexity of the unprecedented task of hot-swapping the live consensus of the Ethereum network in order both to increase  the safety of the operation and to reduce the time to production. Additionally, a minimal changeset helps ensure that *most* smart contracts and services will continue to function as intended during and after the transition with little to no required intervention.
```
lightclient, line 133:
```suggestion
This EIP was designed to minimize the complexity of hot-swapping the live consensus of the Ethereum network. Both the safety of the operation and time to production were taken into consideration. Additionally, a minimal changeset helps ensure that *most* smart contracts and services will continue to function as intended during and after the transition with little to no required intervention.
```

lightclient (changes requested):
a few nits; glad to see it all coming together though

EIPS/eip-3675.md
mkalinin: From [EIP-1#linking-to-other-eips](https://eips.ethereum.org/EIPS/eip-1#linking-to-other-eips):
> For example, you would link to this EIP with `[EIP-1](/EIPS/eip-1)`.

EIPS/eip-3675.md
mkalinin: I am happy to remove this dependency as the specification part of this EIP does not depend on EIP-2982. References to EIP-2982 throughout the doc are purely informational.

EIPS/eip-3675.md
```diff
@@ -0,0 +1,246 @@
...
+* Set the head of the canonical blockchain to the block nominated by the event.
+
+
+### Network
+
+The networking stack **SHOULD NOT** send the following messages if they advertise the descendant of any terminal PoW block:
+* [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
+* [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
+
+Beginning with the first `POS_BLOCK_FINALIZED` event, the networking stack **MUST** remove the handlers corresponding to the following ETH protocol messages:
```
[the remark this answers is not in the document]
mkalinin: Good point! Yes, I think such peers should be disconnected as they are potentially following the PoW chain if it's kept being built and become useless for the node that follows Ethereum network.

EIPS/eip-3675.md
mkalinin: Will this type of links `./eip-2982.md` be converted into `/EIPS/eip-2982` for the website?

EIPS/eip-3675.md
mkalinin: Good catch! From YP: 
> **extraData**: An arbitrary byte array containing data relevant to this block.

mkalinin: Hmm, would it be better to either replace `EMPTY_LIST` with `0xc0` or replace `0x80` with `EMPTY_BYTE_ARRAY`? Having `EMPTY_LIST` aside with `0x80` is a bit messy.

OhArthits (approved):

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+### Block structure
+
+Beginning with `TRANSITION_BLOCK`, a number of previously dynamic block fields are deprecated by enforcing these values to instead be constants. Each block field listed in the table below **MUST** be replaced with the corresponding constant value.
+
+| Field | Constant value |
+|-|-|
+| **`ommersHash`** | `Keccak256(RLP(EMPTY_LIST))`  |
+| **`difficulty`** | `0` |
+| **`extraData`** | `0x80` |
+| **`mixHash`** | `0x0000000000000000000000000000000000000000000000000000000000000000` |
```
axic, line 69: Since now `extraData` shows the raw RLP value of 0x80, shouldn't all these fields do the same?

EIPS/eip-3675.md
djrtwo: yeah, fine by me

EIPS/eip-3675.md
mkalinin: IMO, this table should rather contain raw values, not their encoding. I am thinking about replacing `0x80` with `EMPTY_BYTE_ARRAY` or `EMPTY_BYTE_STRING`.

EIPS/eip-3675.md
MicahZoltu: Hmm... something is wrong, EIP-1 on the eips site doesn't match EIP-1 in GitHub...
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md#linking-to-other-eips
https://eips.ethereum.org/EIPS/eip-1#linking-to-other-eips

I have no idea how this happened, but it should be `./eip-2982.md#motivation`

MicahZoltu: My best guess is that the EIPs site renderer is doing a global find and replace of all relative links with absolute links and accidentally catching this and rewriting it (incorrectly).

EIPS/eip-3675.md
MicahZoltu: Maybe have multiple columns here, one that shows the post-serialization value and others that shows how to get that pre-serialization?  Different implementations may have an easier or harder time hard-coding things at different layers of the serialization process, and this ensures that everyone ends up with the same result no matter how they get there.
```
**`ommersHash`** | `0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347` | `keccak256([0xc0])` | `keccak256(rlp([]))
**`ommers`** | `0xc0` | `rlp([])`
```

EIPS/eip-3675.md
axic: > My best guess is that the EIPs site renderer is doing a global find and replace of all relative links with absolute links and accidentally catching this and rewriting it (incorrectly).

Oh, that might be it.

EIPS/eip-3675.md
lightclient: The correct empty value `0x80`, so that should be `EMPTY_BYTE_ARRAY`.

lightclient:
This is looking pretty good as a draft, please tag me when you're done making changes and are ready to merge.

EIPS/eip-3675.md
mkalinin: This table should look better now, please check.

axic:
These anchors are not working:
```
- ./_site/EIPS/eip-3675.html
  *  linking to internal hash #Block-structure that does not exist (line 287)
     <a href="#Block-structure">block structure</a>
  *  linking to internal hash #EVM that does not exist (line 499)
     <a href="#EVM">EVM</a>
  *  linking to internal hash #Halt-the-importing-of-PoW-blocks that does not exist (line 391)
     <a href="#Halt-the-importing-of-PoW-blocks">Security considerations</a>
  *  linking to internal hash #Terminal-total-difficulty-vs-block-number that does not exist (line 381)
     <a href="#Terminal-total-difficulty-vs-block-number">Security considerations</a>
htmlproofer 3.15.3 | Error:  HTML-Proofer found 4 failures!
```

I think they are made lowercase in the rendering, and/or the dashes are replaced with underscores.

mkalinin:
> These anchors are not working:
> 
> ```
> - ./_site/EIPS/eip-3675.html
>   *  linking to internal hash #Block-structure that does not exist (line 287)
>      <a href="#Block-structure">block structure</a>
>   *  linking to internal hash #EVM that does not exist (line 499)
>      <a href="#EVM">EVM</a>
>   *  linking to internal hash #Halt-the-importing-of-PoW-blocks that does not exist (line 391)
>      <a href="#Halt-the-importing-of-PoW-blocks">Security considerations</a>
>   *  linking to internal hash #Terminal-total-difficulty-vs-block-number that does not exist (line 381)
>      <a href="#Terminal-total-difficulty-vs-block-number">Security considerations</a>
> htmlproofer 3.15.3 | Error:  HTML-Proofer found 4 failures!
> ```
> 
> I think they are made lowercase in the rendering, and/or the dashes are replaced with underscores.

I have turned them all to lowercase, the dashes are in place on the site so they should work (if this output is an attempt to render for the EIP site)

mkalinin:
- Anchors are fixed and build passes
- Table with constant values in the block structure section is updated with more info
- Rationale behind no changes to block structure is extended with a more concrete example
- Added a statement that peers sending gossip after it's been deprecated should be disconnected

I think it's ready for the merge. cc @axic @lightclient @MicahZoltu

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+
+*Note*: If one of the new rules fails then the block **MUST** be invalidated.
+
+*Note*: Validity rules that are not specified in the list above **MUST** remain unchanged.
+
+
+### Block and ommer rewards
+
+Beginning with `TRANSITION_BLOCK`, block and ommer rewards are deprecated. Specifically, the following actions **MUST** be taken:
+* Remove increasing the balance of the block's **`beneficiary`** account by the basic block reward.
```
MicahZoltu, line 95: I think this is referred to just as `block reward` in most places.  Calling it `basic block reward` may lead to confusion since that phrase isn't commonly used.

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+
+The networking stack **SHOULD NOT** send the following messages if they advertise the descendant of any terminal PoW block:
+* [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
+* [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
+
+Beginning with the first `POS_BLOCK_FINALIZED` event, the networking stack **MUST** remove the handlers corresponding to the following ETH protocol messages:
+* [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
+* [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
+
+Beginning with the first `POS_BLOCK_FINALIZED` event, peers that keep sending these messages **SHOULD** be disconnected.
```
MicahZoltu, line 123: I think this will introduce a race condition.  Not all peers will receive `POS_BLOCK_FINALIZED` at the same time, so some may still be broadcasting `NewBlockHashes` and `NewBlock` for siblings of the terminal PoW block after some of their peers have received the `POS_BLOCK_FINALIZED` event.

Consider changing this rule to be that after the *second* `POS_BLOCK_FINALIZED` event peers are disconnected, as by that time all peers should have received the first `POS_BLOCK_FINALIZED` event and should no longer be gossping sibling terminal PoW blocks.

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Halting the import of PoW blocks
+
+See [Security considerations](#halt-the-importing-of-pow-blocks).
+
+### Replacing block fields with constants
+
+Deprecated block fields are replaced with constant values to ensure the block format remains backwards compatible. Preserving the block format aids existing smart contracts and services in providing uninterrupted service during and after this transition.
+
+Particularly, this is important for those smart contracts that verify Merkle proofs of transaction/receipt inclusion and state by validating the hash of externally provided block header string against the corresponding value returned by the `BLOCKHASH` operation.
```
MicahZoltu, line 146:
```suggestion
Particularly, this is important for those smart contracts that verify Merkle proofs of transaction/receipt inclusion and state by validating the hash of externally provided block header against the corresponding value returned by the `BLOCKHASH` operation.
```
The `string` there I think makes the sentence a bit more confusing, as you may receive a block header in many forms such as a structured format, a byte array, a hex string, etc.

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+After the upgrade of the consensus mechanism only the beacon chain network will have enough information to validate a block. Thus, block gossip provided by the `eth` network protocol will become unsafe and is deprecated in favour of the block gossip existing in the beacon chain network. 
+
+It is recommended for the client software to not propagate descendants of any terminal PoW block to reduce the load on processing the P2P component and stop operating in the environment with unknown security properties.
+
+## Test Cases
+*TBD*
+
+## Reference Implementation
+*TBD*
+
```
MicahZoltu, line 183:
```suggestion
```
Consider removing these until you are ready to fill them in.  They are optional sections, and thus not required to be present to merge this as a draft and if they never get filled in that is technically acceptable.  Generally, we want to write EIPs in a way that is "ready to merge as final" without TODOs.

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+Suppose the part of the client software that is connected to the beacon chain network goes offline before the Ethereum network reaches the `TERMINAL_TOTAL_DIFFICULTY` and stays offline while the network meets this threshold. Such an event makes the client software unable to switch to PoS and allows it to keep following the PoW chain if this chain is being built beyond the terminal PoW block. Depending on how long the beacon chain part was offline, it could result in different adverse effects such as:
+* The client has no post-state for the terminal PoW block (the state has been pruned) which prevents it from doing the re-org to the PoS chain and leaving syncing from scratch as the only option to recover.
+* An application, a user or a service uses the data from the wrong fork (PoW chain that is kept being built) which can cause security issues on their side.
+
+Not importing PoW blocks that are beyond the terminal PoW block prevents these adverse effects on safety/re-orgs in the event of software or configuration failures *in favor* of a liveness failure.
+
+### Further threat analysis
+_TBD_
+
```
MicahZoltu, line 244:
```suggestion
```
As above, recommend leaving out unfinished sections until you have something to put into them.

MicahZoltu (approved):
Left some comments that I recommend reviewing, but nothing blocking this from merging as a Draft.

EIPS/eip-3675.md
mkalinin: Fair, but there is also the following statement above:
- The networking stack **SHOULD NOT** send the following messages if they advertise the descendant of any terminal PoW block

Which means that effectively block gossip **SHOULD** be stopped before `POS_BLOCK_FINALIZED`.

EIPS/eip-3675.md
mkalinin: You're right, theoretically they may receive a block header as a set of field values, then RLP encode it, compute a hash and verify this hash. Although it does seem like a suboptimal solution we better remove this "string".

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+*Note*: Validity rules that are not specified in the list above **MUST** remain unchanged.
+
+
+### Block and ommer rewards
+
+Beginning with `TRANSITION_BLOCK`, block and ommer rewards are deprecated. Specifically, the following actions **MUST** be taken:
+* Remove increasing the balance of the block's **`beneficiary`** account by the basic block reward.
+* Remove increasing the balance of the block's **`beneficiary`** account by the ommer inclusion reward per each ommer.
+* Remove increasing the balance of the ommer's **`beneficiary`** account by the ommer block reward per each ommer.
```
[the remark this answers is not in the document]
mkalinin: It may fill redundant but the purpose of this statement is to be consistent with respect to the YP when removing PoW rewards, i.e. the intention is to explicitly remove each source of these rewards.

EIPS/eip-3675.md
MicahZoltu: If you need to pull data out of the header, it may be easier to serialize from raw inputs than it is to parse/decerialize to pull bits out of the header.

EIPS/eip-3675.md
```diff
@@ -0,0 +1,247 @@
...
+#### PoS events
+
+Events having the `POS_` prefix in the name (PoS events) are emitted by the new proof-of-stake consensus mechanism. They signify the corresponding assertion that has been made regarding a block specified by the event. The underlying logic of PoS events can be found in the beacon chain specification. On the occurrence of each PoS event the corresponding action that is specified by this EIP **MUST** be taken.
+
+Each `POS_BLOCK_FINALIZED` event has a reference to the head of the canonical chain in addition to the finalized block and incorporates the semantics of the `POS_CHAINHEAD_SET` event. Thus, every action and effect that is specified on the occurrence of `POS_CHAINHEAD_SET` event according to this EIP **MUST** also be taken on every occurrence of `POS_BLOCK_FINALIZED` event.
+
+
+### PoW block processing
+
+PoW blocks that are descendants of any terminal PoW block **MUST NOT** be imported. This implies that a terminal PoW block will be the last PoW block in the canonical chain.
```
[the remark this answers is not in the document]
mkalinin: You mean that there is a bit of inconsistency between the two sections? Like, this section says keep processing terminal PoW blocks and their ancestors and puts no time boundary on it while the network section implies more strict behaviour and deprecates gossip at a strict boundary. And we might want to reflect this restriction here as well?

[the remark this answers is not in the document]
mkalinin: I'd suggest to add the following kind of statement:
> Beginning with the first `POS_BLOCK_FINALIZED` event, client software **MUST NOT** import _any_ PoW block that doesn't belong to the canonical chain.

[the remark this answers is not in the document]
MicahZoltu: My concern is that currently I don't think clients gossip siblings until they see a new longest chain (maybe I'm wrong about this)?  If that assumption is true, then this will be a behavior change that clients need to implement.

EIPS/eip-3675.md
MicahZoltu: Why would block gossip stop before `POS_BLOCK_FINALIZED`?  I think we should assume that there will be a steady stream of new siblings until `POS_BLOCK_FINALIZED`, since the protocol rules as I understand them suggest you continue to gossip those.

EIPS/eip-3675.md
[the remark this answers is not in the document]
mkalinin: Hmm, interesting. I would assume they do verify PoW upon receiving a block and propagate this block further if the verification passed. In the mean time this block is stored. And here we have options, it either processed right away or processing is deferred until the fork that this block does belong to becomes the longest chain.

And this is where we may have a real issue (not related to gossip). If the terminal PoW block that is picked by the beacon chain proposer has been received but hasn't been processed by our node due to conditions described above (our node may receive another terminal block with the same TD and thus refuse to re-org). Then our node won't be able to verify the beacon block that has been received because it will stuck verifying the PoW block that is the parent of the execution payload. So, it means that the spec must force _every_ terminal PoW block to be processed, not only stored. 

cc @holiman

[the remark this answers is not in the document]
mkalinin: An alternative solution for this edge case would be the corresponding method in the Consensus API, let's say `verifyTerminalPoWBlock(block_hash: Hash32)`. This method would enforce the execution of a sibling terminal PoW block if it hasn't been executed yet. The downside of this approach is the delay between receiving a beacon block and considering it for further processing due to calling to this method, especially, if there are more than one block on this sibling fork that must be executed. But this approach seems to be less intrusive in terms of mainnet clients change set.

EIPS/eip-3675.md
mkalinin: Because there is enough time between the terminal PoW block event and the first `POS_BLOCK_FINALIZED` event that is issued when the first PoS block gets finalized (it is not issued when the beacon block with empty execution payload is finalized). The gap between these two events is at least 2 epochs and unless there is a party that is constantly mining terminal PoW blocks there should be a plenty of time to disseminate all terminal PoW blocks existing in the network.

EIPS/eip-3675.md
mkalinin: Arguably it may be cheaper given a string to verify it against the corresponding `BLOCKHASH` value by parsing out a `number` using some optimized RLP decoder that jumps between list elements and then read the needed field in the same fashion than RLP encoding the whole thing. Anyway, I am going to remove the "string" term.

EIPS/eip-3675.md
MicahZoltu: Ah, I feel like we should *assume* that one or more miners mine terminal PoW siblings.  They have a huge pile of hashing power and nothing to do with it.  Maybe we'll get lucky and they don't, but I feel like we should prepare for them doing this.

I wonder if a miner could do something interesting where once the first terminal PoW block is produced (the most likely one to be finalized) they switch to mining a forked chain some number of blocks back that lowers the difficulty with each block.  The goal here would be to drive the difficulty down as far as possible in this fork given their hashpower while still reaching the target total difficulty.  Once you get to one block shy of target total difficulty, you then produce terminal PoW blocks over and over as fast as you can in order to spam the network?

I'm not sure how much you could manipulate given 2 epochs.

EIPS/eip-3675.md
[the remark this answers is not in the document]
mkalinin: A solution with `verifyTerminalPoWBlock(block_hash: Hash32)` or a similar call aids mitigation of the DoS vector that is described in https://github.com/ethereum/EIPs/pull/3675#discussion_r679207367. If miners have a cheap way to produce terminal PoW blocks to spam the network it is better for a node to avoid executing them all and execute terminal PoW blocks from non-canonical chains on demand.

EIPS/eip-3675.md
mkalinin: > I'm not sure how much you could manipulate given 2 epochs.

Good question. There are possibilities depending on whether whole hashing power participates in anything like that, whether this attack started in advance by switching a portion of hashing power from mainnet to start producing such a chain before the transition. If all hashing power participates it may produce roughly `60` blocks during 2 epochs period but with lowering the difficulty this number increases; like 20 blocks with 1000 seconds gap between their timestamps is required to reduce the difficulty ~8 times and produce roughly `40*8=320` blocks during the rest of the period, it also doesn't count how many blocks needed to reach the terminal total difficulty threshold. If we assume that somehow miners become able to produce a block in a second then during two epochs we would got `768` terminal PoW blocks, I don't know how bad is it for the network if we assume that these blocks are not executed, only propagated with PoW verification.

As for the network stack, if handlers are removed won't peers sending unknown messages be disconnected by clients?

mkalinin: For the health of the network under this kind of attack it makes sense for clients to stop propagating *any* blocks upon receiving the first `POS_BLOCK_FINALIZED` and remove handlers with the corresponding penalty upon receiving the second `POS_BLOCK_FINALIZED`. So, the updated spec should looks like:
> The networking stack **SHOULD NOT** send the following messages if they advertise the descendant of any terminal PoW block:
> * [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
> * [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
>
> Beginning with the *first* `POS_BLOCK_FINALIZED` event, the networking stack **MUST** discard the following ingress messages:
> * [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
> * [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
> 
> Beginning with the *second* `POS_BLOCK_FINALIZED` event, the networking stack **MUST** remove the handlers corresponding to the following messages:
> * [`NewBlockHashes (0x01)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblockhashes-0x01)
> * [`NewBlock (0x07)`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#newblock-0x07)
>
> Peers that keep sending these messages after the handlers have been removed **SHOULD** be disconnected.

MicahZoltu: We should probably move this conversation to the discussions-to link.  PR comments should be assumed to be transient, especially on merged PRs.

mkalinin: done, https://ethereum-magicians.org/t/eip-3675-upgrade-consensus-to-proof-of-stake/6706/2

ptrinh:
hi all

erosbabyyyy31 (approved):

erosbabyyyy31 (approved):

PR #3681, opened 2021-07-27
1559: Change signer.balance assertion location

timbeiko:
After the [Ropsten consensus issue](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/retrospectives/london.md) it was noted that the ordering of the `assert` statements made it unclear that the `signer.balance` field in `assert signer.balance >= transaction.gas_limit * transaction.max_fee_per_gas` had already had `transaction.amount` deducted from it (L207). This change simply bring up this `assert` statement closer to other assertions for `signer.balance` to increase clarity.


abdelhamidbakhta (approved):
Looks good to me

PR #3685, opened 2021-07-28
Move EIP-2481 to "Last Call"

poojaranjan:
The author is ready to move the proposal to the next status.

### DIFF
--- a/EIPS/eip-2481.md
+++ b/EIPS/eip-2481.md
@@ -4,7 +4,7 @@ title: "eth/66: request identifier"
 author: Christoph Burgdorf (@cburgdorf)
 discussions-to: https://github.com/ethereum/EIPs/issues/2482
 status: Last Call
-review-period-ends: 2021-08-11
+review-period-end: 2021-08-11
 type: Standards Track
 category: Networking
 created: 2020-01-17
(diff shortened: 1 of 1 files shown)


cburgdorf (approved):

axic (approved):

EIPS/eip-2481.md
```diff
@@ -3,7 +3,8 @@ eip: 2481
 title: "eth/66: request identifier"
 author: Christoph Burgdorf (@cburgdorf)
 discussions-to: https://github.com/ethereum/EIPs/issues/2482
-status: Review
+status: Last Call
+review-period-ends: 2021-08-11
```
axic, line 7:
```suggestion
review-period-end: 2021-08-11
```

eip-automerger:

 - Trying to change EIP 2481 state from Review to Last Call

MicahZoltu:
This shouldn't have been merged to Last Call.  It `requires: 2464` which is still in draft, and EIPs should not advance past their dependencies.

axic:
> This shouldn't have been merged to Last Call. It requires: 2464 which is still in draft, and EIPs should not advance past their dependencies.

My bad. Also seems 2464 depends on 2364, which is also a draft. This will take quite a bit of time.

MicahZoltu:
*Hopefully* we can get both of those moved to Review and then Last Call on a relatively short timeframe.  All 3 can move through the phases essentially together, they just shouldn't "pass" their dependencies.

PR #3698, opened 2021-08-05
[PROPOSAL] Add `TERMINAL_TOTAL_DIFFICULTY` override to EIP 3675

lsankar4033:
In the case of [adversarial action to disrupt the point of transition](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3675.md#ability-to-jump-between-terminal-pow-blocks), we may want to specify a mechanism for clients to quickly change the transition point. 

This PR adds a client setting to the spec (`--terminal-total-difficulty-override`) that enables this.

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -126,8 +126,6 @@ Beginning with the first `POS_BLOCK_FINALIZED` event, peers that keep sending th
 
 ### `TERMINAL_TOTAL_DIFFICULTY` override
 
-Considering [network anomalies and/or attack scenarios](#ability-to-jump-between-terminal-pow-blocks), it may be necessary to quickly coordinate on a change in the transition point.
-
 Ethereum clients should support a `--terminal-total-difficulty-override DIFFICULTY` parameter that allows users to specify an override to `TERMINAL_TOTAL_DIFFICULTY`.
 
 *Note*: This client setting is only expected to be used in the event of unforseen exceptional scenarios. Docs should contain sufficient warnings of this flag's misusage.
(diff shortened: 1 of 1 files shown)


EIPS/eip-3675.md
```diff
@@ -124,6 +124,11 @@ Beginning with the first `POS_BLOCK_FINALIZED` event, peers that keep sending th
 
 *Note:* The logic of message handlers that are not affected by this section **MUST** remain unchanged.
 
+### `TERMINAL_TOTAL_DIFFICULTY` override
+
+Ethereum clients should support a `--terminal-total-difficulty-override DIFFICULTY` parameter that allows users to specify an override to `TERMINAL_TOTAL_DIFFICULTY`.
```
djrtwo, line 129:
```suggestion
Ethereum clients **SHOULD** support a `--terminal-total-difficulty-override DIFFICULTY` parameter that allows users to specify an override to `TERMINAL_TOTAL_DIFFICULTY`.
```

EIPS/eip-3675.md
```diff
@@ -175,6 +180,12 @@ After the upgrade of the consensus mechanism only the beacon chain network will
 
 It is recommended for the client software to not propagate descendants of any terminal PoW block to reduce the load on processing the P2P component and stop operating in the environment with unknown security properties.
 
+### Allowing for transition difficulty changes
+
+Considering [network anomalies and/or attack scenarios](#ability-to-jump-between-terminal-pow-blocks), it may be necessary to quickly coordinate on a change in the transition point.
```
djrtwo, line 185: suggest moving the link becuase the extent of  "attack scenarios" isn't really described in that section. In particular, one thing this helps fight against is the event of consistent 51% attacks/reorgs that prevent getting to terminal pow difficulty.
```suggestion
Considering [network anomalies](#ability-to-jump-between-terminal-pow-blocks) and/or attack scenarios, it may be necessary to quickly coordinate on a change in the transition point. 
```

djrtwo (commented):
looks good! two nits

EIPS/eip-3675.md
lsankar4033: fair enough!

MicahZoltu:
@djrtwo @mkalinin @vbuterin Looks like the author has addressed the comments, once one of you approves this should automerge.

djrtwo:
@mkalinin is out this week. I'd like to wait to quickly discuss with him before merge. Will leave in this state through the weekend

lsankar4033:
After discussing with @djrtwo and @mkalinin, have decided that a better way to express this is:
1. a client setting on the eth2spec for TTD overrides
2. a [PoS event](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3675.md#pos-events) here expressing changes in TTD coming from the consensus engine

Closing this PR in consideration of those!

PR #3705, opened 2021-08-06
3541: clarify the title

axic:
Recently on Twitter it seems a common misunderstanding is that the rejection is for the contract address. This clarifies it.

### DIFF
--- a/EIPS/eip-3541.md
+++ b/EIPS/eip-3541.md
@@ -1,6 +1,6 @@
 ---
 eip: 3541
-title: Reject new contracts starting with the 0xEF byte
+title: Reject new contract code starting with the 0xEF byte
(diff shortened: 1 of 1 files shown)


MicahZoltu:
I actually find the original more clear.

Maybe a simple summary would help?  😉

axic:
> I actually find the original more clear.

I'm happy with either, some people were confused by it. cc @nanexcool @alexvandesande

> Maybe a simple summary would help? 😉

Well, there's the abstract which explains it just as much. Still think the summary should be part of the header to serve the purpose you think it should (i.e. show up in the "embedded url views", etc.)

MicahZoltu:
Note: We don't have a strict policy about non-normative changes to final EIPs, but it would be nice to get one other author (since this is a multi-author EIP) to approve the change before merging.  Not as a formality, but just to make sure there is general agreement among people familiar with the specification.

chfast (approved):

PR #3715, opened 2021-08-10
Clarifications in 3540/3690

axic:


axic:
> An Exception Occured While Linting: 
> sorry only 1 file is supported right now

@alita-moore @MicahZoltu could this be improved in the future?

alita-moore:
> > An Exception Occured While Linting: 
> > sorry only 1 file is supported right now
> 
> @alita-moore @MicahZoltu could this be improved in the future?

Yes this is an item we have on the backlog. I can prioritize it higher if you'd like

axic:
@alita-moore for this PR I'm just splitting it, but it would be nice in the future :)

PR #3706, opened 2021-08-06
Add description field to the EIP header

axic:
Implementation of https://github.com/ethereum/EIPs/pull/3564#issuecomment-837367526.

The header:
Screenshot

Twitter before:
Screenshot

Twitter after:
Screenshot

**Could also render the `description` prominently after the header.**

PR #3717, opened 2021-08-11
Switch EOF EIPs to Review

axic:


lightclient (approved):

MicahZoltu:
How do you feel about being the first person to include a description in the front matter with this EIP?

axic:
> How do you feel about being the first person to include a description in the front matter with this EIP?

Sure, once we merge #3706 :)

axic:
@MicahZoltu added the field but I think we should still merge #3706 first.

axic:
@MicahZoltu okay to merge now?

MicahZoltu (changes requested):
Reference implementation shouldn't link out to third party GitHub repositories.  It should be either inline or in `../assets/eip-3540/*` or just left out entirely.  As I tell everyone, EIPs are not an appropriate place to collect useful links, they are meant to be a long lived and isolated specification document and people **SHOULD** create useful link repositories elsewhere that link to the specification (not the other way around).

Otherwise looks good.

axic:
> Reference implementation shouldn't link out to third party GitHub repositories. It should be either inline or in ../assets/eip-3540/* or just left out entirely. As I tell everyone, EIPs are not an appropriate place to collect useful links, they are meant to be a long lived and isolated specification document and people SHOULD create useful link repositories elsewhere that link to the specification (not the other way around).

Is this a blocker for review or only last call/final?  Review is still something rather in progress where these links can help the audience quite a bit.  Don't want rehash the previous arguments :)

MicahZoltu:
I consider it a blocker because I think all EIPs should be building toward final and they should not be used for a different purpose while in draft/review/last call (like a link repository, tutorial repository, etc.).  IMO, if the system is functioning well the burden of review should *decrease* with each status it makes it through because most of the issues that would prevent final will have been dealt with in the very early stages and final review is just "make sure no one slipped anything in at the last minute".

However, not all editors necessarily agree with this and maybe you can find one to merge that feels differently.

axic:
> However, not all editors necessarily agree with this and maybe you can find one to merge that feels differently.

It is only you who can merge since commit rights have been stripped off other editors 😓

MicahZoltu:
Ah, if another active editor (e.g., lightclient) says they think this should be merged with the external links I don't mind clicking the button.  I don't know why I have button rights and no one else does.  🤷

MicahZoltu:
Ah, if another active editor (e.g., lightclient) says they think this should be merged with the external links I don't mind clicking the button.  I don't know why I have button rights and no one else does.  🤷

axic:
> IMO, if the system is functioning well the burden of review should decrease with each status it makes it through because most of the issues that would prevent final will have been dealt with in the very early stages and final review is just "make sure no one slipped anything in at the last minute"

While this is a noble goal, I do not think it works like this. In practice, up until Last Call the EIP can change significantly. The point of the Review state is the authors *feel* the proposal is ready for external review, but that external review may cause a lot of changes. This means an editor can't think that if they done a review at the beginning of the Review status, they do not need to do one when approving for Last Call or Final.

MicahZoltu:
I agree that a lot *can* change between draft and final, especially in the normative content.  However, the editorial shape should remain of final quality throughout the process.

I think the core of my argument is centered around the idea that EIPs should not change their intended purpose between phases.  An EIP is a technical specification and it should remain a technical specification for the entire process.  It should not start out as a notepad for brainstorming, then evolve into a repository of links, and then finally switch into a specification.  It should be a specification from start to finish.

If you want a notepad for brainstorming, or a link repository, there are other great tools for this like HackMD, Notion, Discourse forums, Medium blogs, etc.  The EIPs repository is just not the right place to be doing brainstorming or link aggregation.

axic:
I think it is very disingenuous to claim that this EIP is a "notepad for brainstorming" or a "repository of links", when the only external links it has is in the implementation section. The specification is self contained.

There is no rule/recommendation in EIP-1 or the template against external links, but I know that you have been arguing against them. Yet even EIP-3074, which is partially authored by you, has an implementation section with outside links. It seems we are held to higher standards 🤷 

We are just rehashing the same arguments of #3540.

However, it was changed yesterday to include the Python validator in line, so you can just merge this. I am just unsure why other EIPs are allowed to do this.

MicahZoltu:
> even EIP-3074, which is partially authored by you, has an implementation section with outside links

https://github.com/ethereum/EIPs/pull/3397#discussion_r595238875
TL;DR: I argued against the inclusion of that link exactly the same as I am arguing against it here, I am not sure how it made it to review stage, but I'm guessing I missed it when that merge happened.  This is yet another reason why it is good to have the same rules throughout the process, because sometimes things get missed and having multiple checkpoints with the same ruleset reduces the chance of that happening.

> I think it is very disingenuous to claim that this EIP is a "notepad for brainstorming" or a "repository of links", when the only external links it has is in the implementation section. The specification is self contained.

I'm not saying your EIP is a link repository or notepad, but there are definitely people out there who write EIPs like this.  As I have stressed many times before, we are incredibly short staffed in terms of EIP editors and we don't have the time to go through and do a thorough evaluation of every link in a document.  The easiest path (for editors) is to just have hard and fast rules that are easy to follow, and "no external links" is trivially easy to follow.  As soon as we have a fleet of editors available to review EIPs and get into long nuanced debates about the quality/utility of each particular link I think it would be great to change this policy.

Here is an example of a slightly more extreme (but not the most extreme I have seen) example of what happens in reality when we don't have hardline rules:
https://eips.ethereum.org/EIPS/eip-2535#learning--references
Note that the author is currently arguing (reasonably so) in a PR that tries to make that EIP final that "other EIPs include links, why can't I".  I can tell them that older EIPs are grandfathered in, but if I approve this PR as is then I have to now have the *much* more difficult debate with them that this link is acceptable but their links are not, which will require me to evaluate every one of their external links individually.

Now you could argue (and I believe you are) that links early on in the process are acceptable but later on in the process they are not, but that just results in an author building up what they think is a good EIP over time only to have someone jump in at the last minute and tell them to basically rewrite their entire EIP (which is nearly what I have told the author of the above EIP).  Unsurprisingly, they are unhappy with being told this late in the game that their EIP doesn't match the quality standards we are shooting for after they have sunk so much time and energy into the process.

By having one set of rules that we follow for the entire EIP process, it results in authors building out a generally good EIP throughout the process and they aren't caught by any surprises at the end.

MicahZoltu (approved):

axic:
I am mostly arguing that in certain cases external links make a lot of sense. Two cases I see:
1) Linking to external references in motivation/rationale, where that link is not needed to understand the context, but helps expanding on it. EIP-2929 is an example doing this.
2) During development a reference implementation section with links is very useful, and often asked by the core devs. This potentially applies to tests too.

I understand your position claiming it is easier to outright reject links than trying to review them in context. What I really disagreed with is the seemingly different rules applied to different EIPs (even recent ones), where many are allowed to have links, but I felt from the start that the EOF EIPs are held to a higher standard somehow in this regard 😕

MicahZoltu:
There is definitely no intention of different rules applied, and any such treatment is entirely unintentional.  In fact this is why I argue so strongly about not special casing things, because when you do then people feel slighted (like you are, or Nick is in that other EIP).

FWIW, I do agree with you that there are definitely places where links are useful and likely to be reliable (not 404).  I would love it if we had enough EIP editors to evaluate everything and allow *good* links.

gcolvin:
I'm late to this discussion, but a few strong opinions:
* I've seen new EIPs emerge during editing that required significant changes to another EIP in progress.  I've seen criticisms and new ideas arise that required significant changes.  So long as the _intent_ of the EIP hasn't changed I really don't see a problem, and think the appropriate thing to do is move the EIP back to an earlier status if the changes are too large.
* I don't see a big problem with links, especially in earlier phases.  The author of an EIP can't be expected to include a textbook on the relevant subject matter, or give a long lecture during an AllCoreDevs call, but the critics of an EIP may need that context.  (Alternatively they should either do their own homework without links to start from or else abstain from the consensus, but that doesn't seem happen much.)
* I especially don't see a problem with links to PRs or repos in the implementation section, especially in earlier phases.  E.g. an implementation may involve dispersed changes to a client which cannot reasonably fit in the EIP itself, and moving the entire code base of a client to ../assets and leaving the reader to determine what has changed doesn't make much sense to me either.

PR #3723, opened 2021-08-13
Clarify validation conditions in EIP 3675 "The Merge"

ralexstokes:
This EIP provides events from the consensus client which indicate various validation meanings to the execution client.

Some of the language around these events is confusing or needs to be updated:

`POS_CONSENSUS_VALIDATED` should indicate that the referred block and any *ancestors* are considered valid, as a valid block implies all blocks reachable by `block.parent_root` are also valid.

`POS_CHAINHEAD_SET` has the same issue although as I understand this event, there will always be a single block referenced (even in the `FINALIZED` variant) that should be the unique head.

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -82,7 +82,7 @@ Beginning with `TRANSITION_BLOCK`, the block validity conditions **MUST** be alt
 * Add verification of the fields noted in [block structure](#block-structure) as their specified constant value.
 * Add verification of block validity with respect to the PoS consensus rules, i.e. assert that there **MUST** be a corresponding event for the block to consider it valid. Such events are limited to the following:
     * A `POS_CONSENSUS_VALIDATED` event considering the block or any of its ancestors as valid.
-    * A `POS_CHAINHEAD_SET` event, including the `POS_BLOCK_FINALIZED` form of this event, designating the head block as the head of the canonical chain.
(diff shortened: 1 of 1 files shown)


EIPS/eip-3675.md
```diff
@@ -81,8 +81,8 @@ Beginning with `TRANSITION_BLOCK`, the block validity conditions **MUST** be alt
 * Remove all validation rules that are evaluated over the list of ommers and each member of this list.
 * Add verification of the fields noted in [block structure](#block-structure) as their specified constant value.
 * Add verification of block validity with respect to the PoS consensus rules, i.e. assert that there **MUST** be a corresponding event for the block to consider it valid. Such events are limited to the following:
-    * A `POS_CONSENSUS_VALIDATED` event considering the block or any of its descendants as valid.
-    * A `POS_CHAINHEAD_SET` event, including the `POS_BLOCK_FINALIZED` form of this event, designating the block or any of its descendants as the head of the canonical chain.
+    * A `POS_CONSENSUS_VALIDATED` event considering the block or any of its ancestors as valid.
+    * A `POS_CHAINHEAD_SET` event, including the `POS_BLOCK_FINALIZED` form of this event, designating the head block as the head of the canonical chain.
```
mkalinin, line 85:
```suggestion
    * A `POS_CHAINHEAD_SET` event, including the `POS_BLOCK_FINALIZED` form of this event, designating the block as the head of the canonical chain.
```

mkalinin (approved):
LGTM!

PR #3732, opened 2021-08-16
Create auto-withdraw-bot.yml

alita-moore:
This is the implementation of the bot asked for here: https://github.com/ethereum-cat-herders/EIPIP/issues/73#issuecomment-879999394

This bot (every sunday at 00:00)
- Checks every eip file in EIPS for the last time it was edited longer than a year ago (from the time of running)
  - if no EIPs less than a year ago (stop)
  - if EIPs last edited longer than a year ago..
    -  if the EIP is in Draft, Stagnant, Last Call, or Review..
       - change the status to Withdrawn
- Assuming there are EIPs that should be withdrawn, it'll open a PR with those changes and mention the editors (example: https://github.com/alita-moore/EIPs/pull/20)

### DIFF
--- a/.github/workflows/auto-withdraw-bot.yml
+++ b/.github/workflows/auto-withdraw-bot.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           node-version: '14'
       - name: auto-withdraw-bot
-        uses: ethereum/EIP-Bot@a8bae07d5124287539524c6a2b7cc095e66c6811 # master
+        uses: ethereum/EIP-Bot@3d12afa46df84f7747cb5b01ba9abb0e748a9ee3 # mark-eips-stale
         id: auto-withdraw-bot
         with:
           GITHUB-TOKEN: ${{ secrets.TOKEN }}
--- a/.github/workflows/auto-withdraw-bot.yml
+++ b/.github/workflows/auto-withdraw-bot.yml
@@ -1,0 +1,22 @@
+on:
+  schedule:
+    # A job that runs every sunday at 00:00
+    - cron:  '0 0 * * 0'
+
+name: Auto Withdraw Bot
+jobs:
+  auto_merge_bot:
+    runs-on: ubuntu-latest
+    name: Auto Withdraw Bot
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Node.js Environment
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: auto-withdraw-bot
+        uses: ethereum/EIP-Bot@a8bae07d5124287539524c6a2b7cc095e66c6811 # master
+        id: auto-withdraw-bot
+        with:
+          GITHUB-TOKEN: ${{ secrets.TOKEN }}
--- a/.github/workflows/auto-withdraw-bot.yml
+++ b/.github/workflows/auto-withdraw-bot.yml
@@ -3,11 +3,11 @@ on:
     # A job that runs every sunday at 00:00
     - cron:  '0 0 * * 0'
 
-name: Auto Withdraw Bot
+name: Auto Stagnant Bot
 jobs:
   auto_merge_bot:
     runs-on: ubuntu-latest
-    name: Auto Withdraw Bot
+    name: Auto Stagnant Bot
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -15,8 +15,8 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: '14'
-      - name: auto-withdraw-bot
-        uses: ethereum/EIP-Bot@3d12afa46df84f7747cb5b01ba9abb0e748a9ee3 # mark-eips-stale
-        id: auto-withdraw-bot
+      - name: auto-stagnant-bot
+        uses: ethereum/EIP-Bot@b6a3ea6e7e1ca89391dc8c220905f7af72998193 # mark-eips-stale
+        id: auto-stagnant-bot
         with:
           GITHUB-TOKEN: ${{ secrets.TOKEN }}
--- a/.github/workflows/auto-withdraw-bot.yml
+++ b/.github/workflows/auto-withdraw-bot.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           node-version: '14'
       - name: auto-stagnant-bot
-        uses: ethereum/EIP-Bot@b6a3ea6e7e1ca89391dc8c220905f7af72998193 # mark-eips-stale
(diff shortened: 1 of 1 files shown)


MicahZoltu:
> if no EIPs less than a year ago (stop)

This will never evaluate to false.  Most Final EIPs won't ever get touched again.

MicahZoltu:
Bonus feature (can be added later) would be to auto-merge the withdraw PR after 1 week if one of the authors doesn't comment on it.

MicahZoltu:
What testing, if any, have we done or can we do?  I just want to make sure we don't run it for the first time and accidentally create 500 PRs mentioning every author because there was a bug that caused it to incorrectly flag every EIP or something.  😬

MicahZoltu:
Does it correctly handle not opening a new PR every day against the same stale EIP?  Does it somehow track which EIPs it already has a stale PR for?

Related: I wonder how we should handle the situation where an EIP gets a stale PR created, but the author comments saying they will update, but then they don't update.  Maybe for initial version we can just ignore that case and open a single stale PR per EIP ever and not worry about creating another if the author blocks the withdraw.

alita-moore:
> What testing, if any, have we done or can we do?  I just want to make sure we don't run it for the first time and accidentally create 500 PRs mentioning every author because there was a bug that caused it to incorrectly flag every EIP or something.  😬 

I've run it on a fork and it works as expected (I linked a PR on that fork)

alita-moore:
> Does it correctly handle not opening a new PR every day against the same stale EIP?  Does it somehow track which EIPs it already has a stale PR for?
> 
> Related: I wonder how we should handle the situation where an EIP gets a stale PR created, but the author comments saying they will update, but then they don't update.  Maybe for initial version we can just ignore that case and open a single stale PR per EIP ever and not worry about creating another if the author blocks the withdraw.

That's interesting. Currently it creates a single large single PR that just changes all without letting authors know (it only mentions the editors). This was the fastest solution. But I can also implement your feature. The feature of knowing which are open is pretty extensive, though.

alita-moore:
> > if no EIPs less than a year ago (stop)
> 
> This will never evaluate to true.  Most Final EIPs won't ever get touched again.

Ah it's if not changed for more than a year and a withdrawable status

MicahZoltu:
> That's interesting. Currently it creates a single large single PR that just changes all without letting authors know (it only mentions the editors). This was the fastest solution. But I can also implement this.

I'm a fan of expediency and "better than the current situation" over perfect.  😄  That being said, I do feel like we should be giving authors some kind of heads-up/notice that their EIP is about to be withdrawn, rather than just marking it as withdrawn.  I suspect many don't understand that "Draft" is not meant to be a permanent state and simply thought they were done with the process.

alita-moore:
> > That's interesting. Currently it creates a single large single PR that just changes all without letting authors know (it only mentions the editors). This was the fastest solution. But I can also implement this.
> 
> I'm a fan of expediency and "better than the current situation" over perfect.  😄  That being said, I do feel like we should be giving authors some kind of heads-up/notice that their EIP is about to be withdrawn, rather than just marking it as withdrawn.  I suspect many don't understand that "Draft" is not meant to be a permanent state and simply thought they were done with the process.

Sure, could always just mass mention every author in the set lmao. It'll be insane for the first run but if it runs frequently enough then it'll not really need multiple PRs

alita-moore:
Also I was thinking but a good approach to the merge after a week would be to, at the start of each run, check for open PRs and merge them. Then also clean the branches (this auto creates branches which may not be desirable).

MicahZoltu:
If we do a single monolithic PR, what happens if one of the authors indicates they will be updating (or they do update) their EIP?  How do we remove just that EIP from the PR?

How much harder would one PR per EIP be?  That feels like a more correct solution that will result in less management burden, but I am not in a position to judge how difficult that would be relative to this solution.

alita-moore:
Could just close it and let the other authors be caught next run

alita-moore:
The issue is that it'll create quite literally 58 PRs the first run. I don't mind this, but it's worth keeping in mind. This will also mean creating 58 new branches which again is pretty large. Totally feasible, though.

alita-moore:
I say we merge in this method because it's easy and then reevaluate if issues arise. What do you think? No need to waste energy if not necessary!

MicahZoltu:
I'm totally fine with 58 PRs and branches.  We can configure the repository to auto-delete branches on merge (which should have been done anyway).  I think a single monolithic PR is going to create far more of a headache than open branches because we'll have authors from 58 different EIPs all talking about their EIP in one mega-thread, which will trigger a GitHub notification to every one of them each time someone comments.

axic:
Wasn't the status supposed to change to "stale" as first? 🙊 

I'm perfectly fine with withdrawn, however someone made the rule (in EIP-1) that withdrawn can not be revived, it must be a new EIP. That I'm not sure why is a good rule, but in https://github.com/alita-moore/EIPs/pull/20 I see a lot of EIPs which are more like pending (waiting for other things to move), but their authors still consider them.

alita-moore:
alright, I'll do that @MicahZoltu

alita-moore:
yeah I'm not sure actually @axic, I feel like a year is a long time, no?

axic:
> yeah I'm not sure actually @axic, I feel like a year is a long time, no?

I'm not sure what you mean, but EIP-1 has very different rules to what is being implemented here.  I'm not necessarily in favour of those exact rules (*in fact I championed different rules*), but the bot should match EIP-1, and if changes are sought, it should be applied to EIP-1 too.

Screenshot

MicahZoltu:
Ah, good catch.  This bot should be labeling things as Stagnant, not Withdrawn, and it should use 6 months, rather than a year (to match EIP-1 as Alex indicated).

alita-moore:
sounds good, I'll make these changes 👍

alita-moore:
@axic @MicahZoltu I have addressed the requests, it now submits a new pull request / branch for every eip, marks it stagnant instead of withdrawn, and the cutoff is 6 months ago

Here are a few examples of an e2e test
https://github.com/alita-moore/EIPs/pull/199
https://github.com/alita-moore/EIPs/pull/202
https://github.com/alita-moore/EIPs/pull/200
https://github.com/alita-moore/EIPs/pull/199
https://github.com/alita-moore/EIPs/pull/198
https://github.com/alita-moore/EIPs/pull/197
https://github.com/alita-moore/EIPs/pull/196
https://github.com/alita-moore/EIPs/pull/195

alita-moore:
now it will create the pull requests but won't create new ones if there's one already open for that file. There's an edgecase here where if one is opened and then the file is edited then it'll stay open, but I supposed we can handle that manually.

alita-moore:
@MicahZoltu I'm going to merge this because it's been e2e tested, feel free to revert if you have other comments of course.

PR #3734, opened 2021-08-16
Update eip-2535.md

mudgen:
Two week Last Call review period has completed.

### DIFF
--- a/EIPS/eip-2535.md
+++ b/EIPS/eip-2535.md
@@ -3,8 +3,7 @@ eip: 2535
 title: Diamonds, Multi-Facet Proxy
 author: Nick Mudge (@mudgen)
 discussions-to: https://github.com/ethereum/EIPs/issues/2535
-status: Last Call
-review-period-end: 2020-08-15
+status: Final
 type: Standards Track
 category: ERC
 created: 2020-02-22


MicahZoltu (changes requested):
Simple Summary is too long.  Should be about one sentence.

The image at the top of the document should be removed.  EIPs should follow the template in terms of layout and avoid adding extraneous things that may break a future renderer.

There are a large number of sections that aren't part of the template.  See https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md for the set of standard sections.  If you want to include additional information, it should be in one of those sections.

There is no `Implementation` section, only `Reference Implementation` and it should contain code, not a link to external code.  This is an optional section though, so can be removed entirely if inlining a reference implementation doesn't make sense.

`Backward Compatibility` section should describe potential issues with prior existing code/specs in production, not describe how this standard might be forward compatible.  A single sentence indicating no known backward compatibility issues is acceptable if there aren't any.

All of the external links should be removed from this document.  EIPs are technical standards, not an appropriate place to collect links and materials.

In general, this EIP should probably be a lot shorter overall and a large amount of the material removed from it and instead hosted elsewhere.

mudgen:
@MicahZoltu Thanks, I'll get this fixed.

mudgen:
@MicahZoltu The EIP template says EIPs can include a link to test cases. Since that is allowed, caan I include a link to reference implementations?

mudgen:
@MicahZoltu It doesn't seem right/fair to me that some EIPs in Final status have links and/or Reference sections with links and URLs to resources including links to reference implementations but EIP-2535 is not allowed to.   Can this please be reconsidered? 

I think it is considerably helpful to provide readers of this standard with links to background material, implementations and related technical material to help them understand things related to the standard.

I can still fix sections and remove material and images. 

Examples:
* EIP-1155 has a References section with 15 links.  Four of the links are to implementations of the standard.
* EIP-721 has a Reference section that includes 35 URLs to various background information and resources including implementations.
* EIP-777 provides an external link to its reference implementation
* EIP-20 includes links to external implementations

MicahZoltu:
> @MicahZoltu The EIP template says EIPs can include a link to test cases. Since that is allowed, can I include a link to reference implementations?

Ah, someone should probably remove that from the template.  We really don't want external links in EIPs as they should be the root source of truth, and other things should link to them (not the other way around).  Things on the internet come and go with time, and a good standard is one that will still be fully functional/readable in the future when external resources may have moved, been deleted, been mutated, etc.

> It doesn't seem right/fair to me that some EIPs in Final status have links and/or Reference sections with links and URLs to resources including links to reference implementations but EIP-2535 is not allowed to.

Our best practices for EIPs has changed over time.  Unfortunately, we do not have the manpower to go back through history and update older EIPs to match.  Also, sometimes stuff slips through the cracks and doesn't get fixed for the same reason (lack of manpower).

> I think it is considerably helpful to provide readers of this standard with links to background material, implementations and related technical material to help them understand things related to the standard.

An EIP should be a pure technical specification.  Users should **NOT** be starting their research with an EIP, the EIP is where the user should end up after they have decided they want a thing (because of other resources) and then they look at the EIP for the exact technical specification.

There is a large misunderstanding that many people have which is to believe that EIPs are a place where people can go to learn about stuff that is built on Ethereum, but that is very much not its purpose.  The purpose of the EIPs repository is a place to house technical standards.  I highly encourage/recommend people build up documentation elsewhere on a docs site, HackMD, GitHub repository, etc. and then somewhere in those documents there is a link to the technical standard.

The examples you have provided of other EIPs that are poorly written are examples of situations where things have been either grandfathered in because we don't have the manpower to fix history, or things slipped through the cracks (also because we lack manpower).

mudgen:
@MicahZoltu Understood, thanks for explaining.

What about the situation that reference implementations included in a standard become out-dated and possibly unusable over time?  The standard remains perfectly fine but the reference implementation becomes bad. Because Solidity is a changing language. For example the EIP-2535 reference implementation written in Solidity 0.7.0 does not work when the code is changed to Solidity 0.8.0 because of a [bug in the Solidity optimizer introduced in Solidity 0.8.0](https://github.com/ethereum/solidity/issues/10930).  So the reference implementation had to change for Solidity 0.8.0. How should this be handled by Final standards?

The [EIP template](https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md) includes this text:

> EIPs can choose to include links to test cases if applicable.

Should that be removed from the EIP template?

MicahZoltu:
In theory, an update to a reference implementation to make it compile with a new compiler version would be a non-normative change and thus be allowed after the EIP is final.  In most cases though, the reference implementation is not meant to be used directly, but to give people some code to look at to better understand the specification.  The general expectation is that people will write their won implementation, or find an implementation written by someone else online.

For context, often times the reference implementation isn't a complete implementation and may even have some code stubbed out with `// implementation of this function is an exercise for the reader` or something.  It is certainly *better* when the reference implementation is complete, but that isn't always possible.

mudgen:
@MicahZoltu Understood, thanks.

lightclient (changes requested):
Hi @mudgen, it looks like I've missed some issues in this EIP. These should be addressed before moving to final:

* The image at the top should be removed.
* Only EIP-1 specified headings should be at the `##` level, all other headings should be `###` or below.
* Remove references to the old withdrawn Diamond Standard
* Remove the "Learning & References"

PR #3746, opened 2021-08-20
Removes recommendation to link externally.

MicahZoltu:
We should not be encouraging users to put tests elsewhere and link to them.  Instead, we should be recommending users inline or include in ../assets any test cases that are applicable to the EIP.


lightclient (approved):

azboyzranch (commented):
- [ ]

PR #3756, opened 2021-08-21
Create eip to cap gas limit

lightclient:
^


EIPS/eip-cap-gas-limit.md
```diff
@@ -0,0 +1,54 @@
...
+## Specification
+
+As of the fork block `N`, consider blocks with a `gas_target` greater than
+`15,000,000` invalid.
+
+## Rationale
+
+### Why Cap the Gas Target
+
+The gas target is currently under the control of block proposers. They have the
```
MicahZoltu, line 32: The gas target isn't under their control, the gas limit is.

EIPS/eip-cap-gas-limit.md
```diff
@@ -0,0 +1,54 @@
+---
+eip: <to be assigned>
```
MicahZoltu, line 2:
```suggestion
EIP: 3756
```

EIPS/eip-cap-gas-limit.md
lightclient: Silly me.

EIPS/eip-cap-gas-limit.md
lightclient: ty

lightclient:
@MicahZoltu updated.

EIPS/eip-3756.md
```diff
@@ -0,0 +1,55 @@
...
+
+A high gas limit increases pressure on the network. In the benign case, it
+increases the size of the state and history faster than we can sustain. In the
+malicious case, it amplifies the devestation of certain denial-of-service
+attacks.
+
+## Specification
+
+As of the fork block `N`, consider blocks with a `gas_limit` greater than
+`30,000,000` invalid.
```
MicahZoltu, line 27:
```suggestion
As of FORK_BLOCK_NUMBER, consider blocks with a `gas_limit` greater than `30,000,000` invalid.
```
This is the style we usually use for referring to the fork block number inline.  Unless there is an argument for deviating from that I would prefer to stick to the normal style here (note: I also personally like the previous style, so I'm a bit biased).

EIPS/eip-3756.md
```diff
@@ -0,0 +1,55 @@
...
+## Motivation
+
+A high gas limit increases pressure on the network. In the benign case, it
+increases the size of the state and history faster than we can sustain. In the
+malicious case, it amplifies the devestation of certain denial-of-service
+attacks.
+
+## Specification
+
+As of the fork block `N`, consider blocks with a `gas_limit` greater than
```
MicahZoltu, line 26: It may be valuable to be more explicit about what `exactly` is invalid here.  In this case it is the 10th RLP encoded item in the block header that must be a QUANTITY that is less than or equal to 30,000,000.

lightclient:
Can we go ahead and merge this as a draft? I'll update these things before review.

EIPS/eip-3756.md
```diff
@@ -0,0 +1,55 @@
...
+ability to increase the gas limit to whatever value they desire. This allows
+them to bypass the EIP and All Core Devs processes in protocol decisions that
+may negatively affect the security and/or decentralization of the network.
+
+### No Fixed Gas Limit
+
+A valuable property of proposers choosing the gas limit is they can scale it
+down quickly if the network becomes unstable or is undergoing certain types of
+attacks. For this reason, we maintain their ability to lower the gas limit
+_below_ 15,000,000.
```
MadeofTin, line 43: Do you mean Gas target here?

MicahZoltu (changes requested):
Only blocker is the 15M to 30M change (since currently it is incorrect).  The rest are just recommendations.

EIPS/eip-3756.md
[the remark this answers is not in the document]
JekaMas: If we'd like to restrict something, shall we have a good benchmark first for all clients? And then discuss real gas limits of the clients.

lightclient:
@MicahZoltu updated again.

MaxHastings:
I would highly suggest to add another section that addresses the concerns some people may have that the gas limit will never be raised again. We have seen dev limits in the past cause lots of controversy. Most notably the 1 MB block limit on Bitcoin. How will the Ethereum team reassure the community that this EIP will not lead to the same problem? Thank you :)

aliatiia:
Is the implementation simply [this](https://github.com/aliatiia/go-ethereum/commit/5876bc97c89f28b9d8cfb2cfcfd39e8b06cca977)?

EIPS/eip-3756.md
shemnon: Counting RLP fields is too pedantic for readability.  I think referencing EIP-1559 in the header and then "a `block` with `gas_limit` greater than `30,000,000`" reads much better.

EIPS/eip-3756.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Backwards Compatibility
+No backwards compatibility issues.
+
+## Test Cases
+TBD
+
```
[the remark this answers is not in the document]
holgerd77: I actually like this TBD especially for the Test Cases as a reminder that we should add at some point.

MicahZoltu:
Discussion about the general content of the EIP should be posted in the `discussions-to` link (in the EIP header).  PR comments are considered transient and meant just for editorial remarks and technical comments on the structure of the EIP.

MicahZoltu (approved):

prayank23:
> Most notably the 1 MB block limit on Bitcoin

@MaxHastings It's not 1 MB right now: https://bitcoin.stackexchange.com/questions/98810/whats-the-blocksize-limit-after-segwit-and-how-do-legacy-nodes-deal-with-segwit

PR #3763, opened 2021-08-24
updated implementation details in EIP-712

wschwab:


EIPS/eip-712.md
```diff
@@ -414,13 +414,39 @@ To be done before this EIP can be considered accepted:
...
+```javascript
+web3.currentProvider.sendAsync({
+  method: 'eth_signTypedData',
+  params: [msgParams, from]
+```
+In addition, the `signTypedData` function is of limited utility in web3 convenience libraries, as it is generally relevant to the interface that stores the private keys used for signing, and with the excepetionof burner wallets, this is almost always done in practice within a wallet.
+
+* `keccak256` struct hashing
+
+While the `keccak256` function in Solidity cannot be invoked directly on structs, the introduction of `abi.encode` gives a sufficient and well-accepted avenue to hashing structs.
```
Recmo, line 443: And when gas cost matters, dropping to assembly is always an option.

Recmo (approved):
Awesome! 🙌

EIPS/eip-712.md
wschwab: Thanks for taking a look! Do you think this needs to be included in the text, or should it be left as is?

MicahZoltu:
CI failing on:
> ./EIPS/eip-712.md:432: experimnetal ==> experimental
> ./EIPS/eip-712.md:447: sufficently ==> sufficiently

wschwab:
fixed typos

PR #3767, opened 2021-08-25
Update eip-2315.md

gcolvin:
*

### DIFF
--- a/EIPS/eip-2315.md
+++ b/EIPS/eip-2315.md
@@ -1,12 +1,12 @@
 ---
 eip: 2315
 title: Simple Subroutines for the EVM
-description: A small change to the EVM that provides native subroutines.
-status: Review
+status: Draft
 type: Standards Track
 category: Core
 author: Greg Colvin <greg@colvin.org>, Martin Holst Swende (@holiman)
 discussions-to: https://ethereum-magicians.org/t/eip-2315-simple-subroutines-for-the-evm/3941
+requires: [EIP-3540: EVM Object Format (EOF)](./eip-3540.md)
 created: 2019-10-17
 ---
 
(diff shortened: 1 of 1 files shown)


MicahZoltu (changes requested):
All opcode references in the specification must be accompanied by their respective number with the format `REVERT (0xfe)`.  You can choose to only use this style for the first reference in the specification section, or for all of them, but it must be included in the first reference to each opcode.

----

For the new opcodes (BEGINSUB, JUMPSUB, RETURNSUB), I recommend just picking some unused ones for now.  We can always change them prior to final, but having a straw man for people to argue against has value in aiding discussion, prototype implementation, etc.

----

EIP dependencies should be included in the frontmatter with:
```md
requires: 3540
```

----

`[EVM Object Format (EOF)](https://eips.ethereum.org/EIPS/eip-3540)` needs to change to `[EIP-3540](./eip-3540.md)`.  I wouldn't be opposed to `[EIP-3540: EVM Object Format (EOF)](./eip-3540.md)` either if you like having the title present (I do think the title adds value).  The key is that the link should be relative, and the EIP number should be included in the link title in its proper format (`EIP-####`).

----

External link to closed EVMOne pull request should be removed.  If this EIP depends on a standardized change, then that should be an EIP itself.  If the EVMOne change is just an example of a necessary change that clients will need to make, then I recommend briefly describing the necessary change here.

----

Reference to 615 in the Rationale and 3337 in the Backward compatibility should both be removed, or those EIPs should be moved to Review prior to this EIP moving to review.  I recommend just removing the links though, as neither add meaningful value in a world where they don't make it to final (which is the world we should be authoring in if they remain in stages lower than this EIP).

----

The Implementation section is no longer a standard EIP section.  It was replaced with the optional Reference Implementation section which should contain either an inline implementation or pseudocode implementation, or a link to `../assets/3767/*` which contains a reference implementation.  If you want to track implementation progress later in the process, https://github.com/ethereum/execution-specs is the appropriate place to do so.

----

The Costs and Codes section should be a subsection of the Specification.  It is impossible to implement this EIP in a consensus compatible way without gas costs, so it should go there.  Also, EIPs should not "suggest", they should assert.  My recommendation is to just put the gas costs in the ## Specification section directly along with each opcode's definition and then in the ## Rationale section you can describe your reasoning and methodology for each.

----

TODO, Appendix, and References sections are not standard EIP sections and should be removed.  EIPs are not the appropriate place to discuss future works or pass around acknowledgements.  They should be purely technical standards.  I strongly encourage all EIP authors to draft a separate document somewhere like hackmd or similar that tracks things that aren't explicitly related to the standard.  This includes collections of links, references, various external implementations, etc.

gcolvin:
Most of these are easy enough fixes, thanks for your careful review.

Some awkwardness probably needs to remain for a while.  What appears to be happening is that ideas from EIP-615 and related new ideas are being taken up by a few other EIPs in various stages of progress and separability.  So right now this EIP depends on other work that is still in progress.  We aren't yet sure how it will all shake out.  If the work this EIP depends on doesn't happen this EIP will need to be changed accordingly or perhaps withdrawn.  Not even reviewing this EIP until they are Final doesn't work for me, but I intended for it to remain under Review until those are ready.

> External link to closed EVMOne pull request should be removed. If this EIP depends on a standardized change, then that should be an EIP itself. If the EVMOne change is just an example of a necessary change that clients will need to make, then I recommend briefly describing the necessary change here.

The opcodes described in this EIP would still work without static jumps, but the safety analysis requires them.  But static jumps should be a separate EIP because they do not depend on subroutines.  That PR is the only place they are currently described.  I think all I can do is encourage @gumb0 to get his EIP going.

> Reference to 615 in the Rationale and 3337 in the Bacard compatibility should both be removed, or those EIPs should be moved to Review prior to this EIP moving to review. I recommend just removing the links though, as neither add meaningful value in a world where they don't make it to final (which is the world we should be authoring in if they remain in stages lower than this EIP).

I refer to EIP-615 because its safety properties have been discussed a lot and even formally verified over the six years of its life, and it's important that those properties are being retained.  (A major critique of the previously Accepted version of this proposal was that it failed to provide those properties.)

 ...
> TODO, Appendix, and References sections are not standard EIP sections and should be removed. EIPs are not the appropriate place to discuss future works or pass around acknowledgements. They should be purely technical standards. I strongly encourage all EIP authors to draft a separate document somewhere like hackmd or similar that tracks things that aren't explicitly related to the standard. This includes collections of links, references, various external implementations, etc.

It's an important, or at least a useful feature of this EIP that it supports the work described in the Appendix.  IETF RFCs allow Appendices.  (Acknowledgements too.)  But I can remove if it's that objectionable.

I'm not at all sure why we don't allow for References.  The material can all be found via links within the EIP, but a full list at the end is generally useful (and present in most docs.)  IETF RFCs can include both Normative and Informative reference sections.   But again, I can remove if it's that objectionable.

Tkb68 (approved):

gcolvin:
@MicahZoltu  I've commited changes for much of your critique.   A bunch of little stuff. The Reference section is (sadly) gone.  I've moved the links to EIP-615 to a single note in the Security section,  and replaced the link in the Abstract with detail as to what actual safety promises are being made.  I've taken Review off for now.  It can be discussed on the Magicians and in the #evm Discord without being in Review status.  And that gives @gumb0 and @axic time to move their EIPs further.

Tkb68 (approved):

Tkb68 (commented):
#3767

gcolvin:
The EIP Bot is dying with the following, which I don't understand.
```
 An Exception Occured While Linting: 
YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key at line 9, column 8:
    created: 2019-10-17
           ^
Error: can not read a block mapping entry; a multiline key may not be an implicit key at line 9, column 8:
    created: 2019-10-17
```

gcolvin:
> The EIP Bot is dying with the following, which I don't understand...

Seemed to be something wrong with url syntax on previous line, which I was removing anyway.

Tkb68:
#3768

MicahZoltu:
@alita-moore Any idea why the bot merged this?  It should have waited for a review approval.  Is this that bug we ran into before where if the author is an editor the bot doesn't wait for approval?

alita-moore:
I addressed that one @MicahZoltu but will investigate these now. BTW I can fix bugs faster without tests. If what was said during EIPIP was true and the bot will be obsolete in 6 months then it would be best to not write tests but I will otherwise. What do you think?

alita-moore:
> The EIP Bot is dying with the following, which I don't understand.
> ```
>  An Exception Occured While Linting: 
> YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key at line 9, column 8:
>     created: 2019-10-17
>            ^
> Error: can not read a block mapping entry; a multiline key may not be an implicit key at line 9, column 8:
>     created: 2019-10-17
> ```

This is a front matter error. It may be best to use a different package for parsing MD files though if there are consistent issues

alita-moore:
@MicahZoltu the reason that this was allowed to merge was because it's in draft state which is currently allowed to be merged without editor approval. Would you like me to make that a requirement>

alita-moore:
ah, but yeah it might be merging because greg is an author

MicahZoltu:
Oh, sorry!  I thought this was the PR that changed state to review.  It is fine, I must have been looking at the wrong PR.

MicahZoltu:
Ah, I see.  Originally it was a PR to change status to Review, but after feedback Greg moved it back to Draft.  So everything went as it should have. Sorry for the false alarm.

alita-moore:
no worries, I found a bug while looking into it (changing status didn't require editor approval and if status change was allowed it would still require editor approval).

PR #3783, opened 2021-08-30
Update interface definition code section

cliffhall:
* Removed supportsInterface from the IERC2981 interface
* Added a separate definition of the IERC165 interface with the original comment language
* For more details about this change, see; https://github.com/ethereum/EIPs/issues/2907#issuecomment-903283402


blmalone (approved):
LGTM 👍

PR #3788, opened 2021-09-02
Strict enforcement of chainId

GregTheGreek:

 - The build passes.

### DIFF
--- a/EIPS/eip-3788.md
+++ b/EIPS/eip-3788.md
@@ -1,7 +1,7 @@
 ---
 eip: 3788
 title: Strict enforcement of chainId
-description: Strictly enforce the chainId on transactions based on the chain.
+description: Reject transactions that do not explicitly have the same chainId as the node's configuration.
 author: Gregory Markou (@GregTheGreek)
 discussions-to: tbd
 status: Draft
--- a/EIPS/eip-3788.md
+++ b/EIPS/eip-3788.md
@@ -3,7 +3,7 @@ eip: 3788
 title: Strict enforcement of chainId
 description: Reject transactions that do not explicitly have the same chainId as the node's configuration.
 author: Gregory Markou (@GregTheGreek)
-discussions-to: tbd
+discussions-to: https://ethereum-magicians.org/t/discussion-to-eip-3788-strict-enforcement-of-chainid/7001
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-3788.md
+++ b/EIPS/eip-3788.md
@@ -19,7 +19,7 @@ Reject transactions that do not explicitly have the same chainId as the node's c
 
 Per [EIP-155](./eip-155.md) a transaction with a `chainId = 0` is considered to be a valid 
 transaction. This was a feature to offer developers the ability to sumbit replayable transactions 
-across different chains. With the rise of evm combatible chains, many of which use forks, or packages
+across different chains. With the rise of evm compatible chains, many of which use forks, or packages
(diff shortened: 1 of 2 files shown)


EIPS/eip-3788.md
```diff
@@ -0,0 +1,60 @@
+---
+eip: 3788
+title: Strict enforcement of chainId
+description: Strictly enforce the chainId on transactions based on the chain.
```
axic, line 4: I think the abstract explains the goal better. I had no idea what this EIP does based on this line.

EIPS/eip-3788.md
```diff
@@ -0,0 +1,60 @@
+---
+eip: 3788
+title: Strict enforcement of chainId
+description: Reject transactions that do not explicitly have the same chainId as the node's configuration.
+author: Gregory Markou (@GregTheGreek)
+discussions-to: tbd
```
MicahZoltu, line 6: This is a required field.  An ethereum-magicians forum thread is *strongly* recommended (you'll have to put up a bit of a fight to get other locations accepted).

MicahZoltu (changes requested):
discussions-to is required.

----

Personal note (not a blocker for EIP), which I can add to the discussions-to once it is available:

On a separate note, I disagree with this change.  It doesn't add value to the protocol and it breaks backward compatibility.  Any wallet/signing tool that wants this can easily add it and we don't benefit from adding it to the core protocol that we cannot achieve by just convincing wallets/signers to add it.

In general, anything that can be solved at a higher layer than the core protocol I believe should be solved at a higher layer than the core protocol, and the core protocol should be as permissive as possible when given a choice between permissive and restrictive.

EIPS/eip-3788.md
GregTheGreek: Just got to my computer, added it. Thanks for the reminder.

EIPS/eip-3788.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Abstract
+
+Reject transactions that do not explicitly have the same chainId as the node's configuration.
+
+## Motivation
+
+Per [EIP-155](./eip-155.md) a transaction with a `chainId = 0` is considered to be a valid 
+transaction. This was a feature to offer developers the ability to sumbit replayable transactions 
+across different chains. With the rise of evm combatible chains, many of which use forks, or packages
```
MicahZoltu, line 22:
```suggestion
across different chains. With the rise of evm compatible chains, many of which use forks, or packages
```

MicahZoltu (changes requested):
Bot is blocking due to spelling error.

MicahZoltu (approved):

sergiokessler:
> In general, anything that can be solved at a higher layer than the core protocol I believe should be solved at a higher layer than the core protocol, and the core protocol should be as permissive as possible when given a choice between permissive and restrictive.

can this be viewed as complete / incomplete info (instead of permissive / restrictive) ?

I mean, if there is the chainId the info is complete, incomplete otherwise.

GregTheGreek:
> > In general, anything that can be solved at a higher layer than the core protocol I believe should be solved at a higher layer than the core protocol, and the core protocol should be as permissive as possible when given a choice between permissive and restrictive.
> 
> 
> 
> can this be viewed as complete / incomplete info (instead of permissive / restrictive) ?
> 
> 
> 
> I mean, if there is the chainId the info is complete, incomplete otherwise.
> 
> 

Please use the discussion link :)

PR #3790, opened 2021-09-03
EIP-170: Name the MAX_CODE_SIZE constant

chfast:
Name the `0x6000` constant in EIP-170 as `MAX_CODE_SIZE`.

### DIFF
--- a/EIPS/eip-170.md
+++ b/EIPS/eip-170.md
@@ -12,12 +12,13 @@ created: 2016-11-04
 [Spurious Dragon](./eip-607.md)
 
 ### Parameters
+- `MAX_CODE_SIZE`: `0x6000` (`2**14 + 2**13`)
 - `FORK_BLKNUM`: 2,675,000
-- `CHAIN_ID`: 1 (main net)
+- `CHAIN_ID`: 1 (Mainnet)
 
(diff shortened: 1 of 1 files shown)


MicahZoltu (approved):
I'm fine with this.  A non-normative change that may help facilitate future conversations.

EIPS/eip-170.md
```diff
@@ -12,12 +12,13 @@ created: 2016-11-04
 [Spurious Dragon](./eip-607.md)
 
 ### Parameters
+- `MAX_CODE_SIZE`: `0x6000` (`2**14 + 2**13`)
 - `FORK_BLKNUM`: 2,675,000
 - `CHAIN_ID`: 1 (main net)
```
axic, line 17: Could also do this based on https://eips.ethereum.org/EIPS/eip-2228:
```suggestion
- `CHAIN_ID`: 1 (Mainnet)
```

axic (approved):

EIPS/eip-170.md
chfast: Fine, but these other params are irrelevant (we don't define hard fork blocks in the EIPs nowaways).

PR #3798, opened 2021-09-04
Stack depth definition was wrong

gcolvin:


EIPS/eip-3779.md
```diff
@@ -2,7 +2,7 @@
 eip: 3779
 title: Safer Control Flow for the EVM
 description: Ensure a minimal level of safety for EVM code.
-status: Review
+status: Draft
```
MicahZoltu, line 5: Was it intentional to change it back to draft here?

PR #3849, opened 2021-09-05
Add missing EIP states and bring en par with EIP-1

ligi:
Bring the states en par with EIP-1 - otherwise this can lead to confusion like here: https://github.com/ethereum/EIPs/pull/3838#discussion_r702400025 as the "Review" state is not reflected on https://eips.ethereum.org

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,7 +21,6 @@ title: Home
   <li><strong>Final</strong> - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.</li>
   <li><strong>Stagnant</strong> - Any EIP in Draft or Review if inactive for a period of 6 months or greater is moved to Stagnant. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to Draft.</li>
   <li><strong>Withdrawn</strong> - The EIP Author(s) have withdrawn the proposed EIP. This state has finality and can no longer be resurrected using this EIP number. If the idea is pursued at later date it is considered a new proposal.</li>
-
   <li><strong>Living</strong> - A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1.</li>
 </ul>
 
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@ title: Home
   <li><strong>Review</strong> - An EIP Author marks an EIP as ready for and requesting Peer Review.</li>
   <li><strong>Last Call</strong> - This is the final review window for an EIP before moving to FINAL. An EIP editor will assign Last Call status and set a review end date (review-period-end), typically 14 days later. If this period results in necessary normative changes it will revert the EIP to Review.</li>
   <li><strong>Final</strong> - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.</li>
-  <li><strong>Stagnant</strong> - Any EIP in DRAFT or REVIEW if inactive for a period of 6 months or greater is moved to STAGNANT. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to DRAFT.</li>
+  <li><strong>Stagnant</strong> - Any EIP in Draft or Review if inactive for a period of 6 months or greater is moved to Stagnant. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to Draft.</li>
   <li><strong>Withdrawn</strong> - The EIP Author(s) have withdrawn the proposed EIP. This state has finality and can no longer be resurrected using this EIP number. If the idea is pursued at later date it is considered a new proposal.</li>
 
   <li><strong>Living</strong> - A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1. Any changes to these EIPs will move between REVIEW and LIVING states.</li>
--- a/index.html
+++ b/index.html
@@ -14,11 +14,15 @@ title: Home
 
 <h2>EIP status terms</h2>
 <ul>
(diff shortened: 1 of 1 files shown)


ligi:
@eth-bot I cannot fulfill your wish - it needs to be this file changed - seems you are buggy :-)

index.html
```diff
@@ -14,11 +14,15 @@ <h2>Contributing</h2>
...
+  <li><strong>Idea</strong> - An idea that is pre-draft. This is not tracked within the EIP Repository.
+  <li><strong>Draft</strong> - The first formally tracked stage of an EIP in development. An EIP is merged by an EIP Editor into the EIP repository when properly formatted.</li>
+  <li><strong>Review</strong> - An EIP Author marks an EIP as ready for and requesting Peer Review.</li>
+  <li><strong>Last Call</strong> - This is the final review window for an EIP before moving to FINAL. An EIP editor will assign Last Call status and set a review end date (review-period-end), typically 14 days later. If this period results in necessary normative changes it will revert the EIP to REVIEW.</li>
+  <li><strong>Final</strong> - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.</li>
+  <li><strong>Stagnant</strong> - Any EIP in DRAFT or REVIEW if inactive for a period of 6 months or greater is moved to STAGNANT. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to DRAFT.</li>
```
MicahZoltu, line 22:
```suggestion
  <li><strong>Stagnant</strong> - Any EIP in Draft or Review if inactive for a period of 6 months or greater is moved to Stagnant. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to Draft.</li>
```

index.html
```diff
@@ -14,11 +14,15 @@ <h2>Contributing</h2>
...
+  <li><strong>Withdrawn</strong> - The EIP Author(s) have withdrawn the proposed EIP. This state has finality and can no longer be resurrected using this EIP number. If the idea is pursued at later date it is considered a new proposal.</li>
+
+  <li><strong>Living</strong> - A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1. Any changes to these EIPs will move between REVIEW and LIVING states.</li>
```
MicahZoltu, line 25:
```suggestiotn
  <li><strong>Living</strong> - A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1.</li>
```
I believe we decided to remove the toggling of living.  We should fix EIP-1 if that is what it says.

MicahZoltu (changes requested):
A few minor changes, but I'm fine with the change otherwise.

index.html
ligi: Yes - this is what EIP-1 says - I just copied from there to bring things en par

axic:
I wish this would be just dropped from the README. What its he point having it there?

axic:
There's also a version of this in the README. Can these be dropped from the README? It seems like calling for trouble having this duplicated in EIP-1, README, and the site.

ligi:
@axic agree on keeping it DRY - maybe here we should also just link to the section in EIP-1?

MicahZoltu:
We have discussed the problem of EIP-1, Readme, eips.ethereum.org, and ethereum.org/eips all being slightly different from each other and each slightly incorrect in the past.  It would be great to "fix" this problem but we haven't found someone with the time and willingness to dig in and fully resolve the issue, so we continue to stumble along with all 4 sources conflicting with each other more often than not.

index.html
MicahZoltu: Sorry, my comment was supposed to be a ```suggestion but I typod so it just ended up being a confusing comment.  Please have another look at the comment above.  I would rather have this be correct than copy EIP-1 (which is currently incorrect).  😊

MicahZoltu: https://github.com/ethereum/EIPs/pull/3856

lightclient:
Looks good to me, thanks @ligi. I am also onboard with removing from the `README`.

ligi:
Can this be merged? Sure the situation can be improved more - but I think this can be a good step on the way.

MicahZoltu:
I would still like https://github.com/ethereum/EIPs/pull/3849/files#r702928548 addressed.  EIP-1 has been fixed to no longer contain incorrect information, and I would prefer not to merge a change with known incorrect information.

ligi:
applied your suggestion - so I think this can be merged now

index.html
```diff
@@ -14,11 +14,15 @@ <h2>Contributing</h2>
...
+  <li><strong>Review</strong> - An EIP Author marks an EIP as ready for and requesting Peer Review.</li>
+  <li><strong>Last Call</strong> - This is the final review window for an EIP before moving to FINAL. An EIP editor will assign Last Call status and set a review end date (review-period-end), typically 14 days later. If this period results in necessary normative changes it will revert the EIP to Review.</li>
+  <li><strong>Final</strong> - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.</li>
+  <li><strong>Stagnant</strong> - Any EIP in Draft or Review if inactive for a period of 6 months or greater is moved to Stagnant. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to Draft.</li>
+  <li><strong>Withdrawn</strong> - The EIP Author(s) have withdrawn the proposed EIP. This state has finality and can no longer be resurrected using this EIP number. If the idea is pursued at later date it is considered a new proposal.</li>
+
```
MicahZoltu, line 24:
```suggestion
```

MicahZoltu (approved):

PR #3855, opened 2021-09-06
Add draft for PUSH0

axic:


lightclient (commented):
Looks great to me. Really interesting data on how often people are using `6000`!

EIPS/eip-3855.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Test Cases
+
+- `5F`
+- `5F5F..5F` (1024 times)
+- `5F5F..5F` (1025 times)
+
+## Security Considerations
+
+The authors are not aware of any impact on security.
```
tkstanczak, line 62: JUMPDEST analysis probably needs to be updated in all clients if PUSH0 is introduced

EIPS/eip-3855.md
```diff
@@ -0,0 +1,66 @@
...
+
+The main motivations for this change include:
+1) Reducing contract code size.
+2) Reducing the risk of contracts (mis)using various instructions as an optimisation measure. Repricing/changing those instructions can be more risky.
+3) Reduce the need to use `DUP` instructions for duplicating zeroes.
+
+To put the "waste" into perspective, across existing accounts 340,557,331 bytes are wasted on `PUSH1 00` instructions, which means 68,111,466,200 gas was spent to deploy them. In practice a lot of these accounts share identical bytecode with others, so their total stored size in clients is lower, however the deploy time cost must have been paid nevertheless.
+
+An example for 2) is [EIP-2733](./eip-2733.md) which changes the behaviour of `RETURNDATASIZE` such that it may not be guaranteed to be zero at the beginning of the call frame.
+
```
MicahZoltu, line 33: Not a blocker yet, but will be a blocker if you want to move this to review before EIP-2733 is moved to review.  My personal recommendation is to remove this line as it doesn't add significant value to the EIP.  An alternative would be to just describe the problem without mentioning the EIP like:
```suggestion
An example for 2) is a potential change to the behavior of `RETURNDATASIZE` such that it may not be guaranteed to be zero at the beginning of the call frame.
```

EIPS/eip-3855.md
```diff
@@ -0,0 +1,66 @@
...
+### Gas cost
+
+The `verylow` gas cost is used for instructions which place constant values onto the stack, such as `ADDRESS`, `ORIGIN`, and so forth.
+
+### Opcode
+
+`0x5f` means it is in a "contiguous" space with the rest of the `PUSH` implementations and potentially could share the implementation.
+
+If this argument is not strong enough, then `0x5c` seems like a good alternative choice.
+
```
MicahZoltu, line 49:
```suggestion
```
This comment is better suited for the discussions-to, rather than the EIP itself as it is a discussion about a potential alternative option.

EIPS/eip-3855.md
```diff
@@ -0,0 +1,66 @@
...
+## Backwards Compatibility
+
+This EIP introduces a new opcode which did not exists previously. Already deployed contracts using this opcode could change their behaviour after this EIP.
+
+## Test Cases
+
+- `5F`
+- `5F5F..5F` (1024 times)
+- `5F5F..5F` (1025 times)
```
MicahZoltu, line 58: I feel like these should each mention expected behavior?  I assume the 1025 pushes should expect failure, the first should expect one 0 on the stack, and the middle should expect 1024 0s on the stack?

MicahZoltu (approved):
Recommend reviewing the comments I left, but structurally this is fine to merge as a draft.  I can click the button when you are ready.

MicahZoltu:
I guess Alita implemented "editor approval auto-merges".  I still recommend reviewing the feedback I left though @axic.  😁

EIPS/eip-3855.md
chfast: I don't think so as PUSH0 does not have any data bytes.

dceleda: Not necessarily.  Most of the clients mark data segments and in case of PUSH0 there is no data segment , so the analysis will just iterate through this instruction like many others.

chfast: In other words: I don't think there is any functional change needed to jumpdest analysis. It still should only consider `PUSH1` to `PUSH32` and `JUMPDEST` opcodes. We can add this explainer to the draft.

EIPS/eip-3855.md
axic: I still think this rule is extremely counterproductive. I agree that the specification part should not rely on EIPs which have a lower status, but motivation and rationale are especially places where such references add a lot of value.

Based on earlier comments it seems I am not the only feeling such a way, but everyone gradually gives up.

MicahZoltu: The problem is that EIPs get rewritten a lot in early stages, and the reference may not make sense in the future.  I just gave some feedback to someone the other day on a Draft EIP and the outcome of the discussion that followed is likely to result in the entire EIP being gutted and mostly rewritten.  It still will probably keep the same title, but so much will change that any references to it externally likely will stop making sense.

axic: If you wouldn't complain about this other rule of not linking to ethmagicians in the motivation, I would have actually just linked to the discussion pointer where this was raised :)

MicahZoltu: That is really the same issue.  Stuff on the internet can change, and we want to ensure that EIPs are static documents that don't have their content mutate over time as the rest of the internet changes.  In almost all of these cases, I think it is generally reasonable to just inline the argument being made rather than linking to an external resource that makes the argument.

axic: Without giving a reference we just make random claims, then we need to repeatedly point people to the source of the information. You are asking us to work much harder because of a counterproductive rule.

axic: Btw, your suggested phrasing doesn't explain the problem we are referring to.

axic: I think a compromise is the following, where the reader can look up further information:
```suggestion
An example for 2) is changing the behaviour of `RETURNDATASIZE` such that it may not be guaranteed to be zero at the beginning of the call frame. This was proposed as a way to chain transactions (i.e. EIP-2733).
```

MicahZoltu: I would be fine with removing this example entirely.  IMO, the list of statements above do a sufficient explanatory job for an EIP.  The purpose of the motivation section isn't to provide an irrefutable argument for why the EIP is necessary, but rather to give people a quick overview of why this EIP is valuable in general.

I think this is still rooted in the same debate we have had repeatedly, which is that I don't think EIPs should be an information aggregator about a change, and you do.  I really liked how EIP-1559 had https://hackmd.io/@timbeiko/1559-resources and EIP 721 had http://erc721.org/ where the authors recorded all of the ancillary information related to the specification (including links, discussions, resources, etc.).  I really wish all EIP authors would do this, and people would no longer need to use the EIP as the "first place you look for information".

axic: You seem to have this fight on every new EIP because everyone assumes this is the logical thing to do 😉

axic:
> Recommend reviewing the comments I left, but structurally this is fine to merge as a draft. I can click the button when you are ready.

Pushed these changes in #3859.

EIPS/eip-3855.md
Arachnid: Removing explanatory content that requires a link because the link breaking will render it less useful seems incredibly counterproductive. If you do that, it's already strictly less useful than in the worst case situation otherwise, with a broken link.

MicahZoltu: The explanatory content can be moved inline is the point.  One can author an EIP such that it stands on its own without needing to link externally.  In this case, it is just a matter of making the case here rather than sending people elsewhere to find the argument being made.

Again, in this particular situation I think it is entirely unnecessary as the motivation already sufficiently explains why this change is useful and the link is just "supporting evidence" which would perhaps be valuable if you were engaging in a debate with someone, but EIPs are not the right place to form a long-form argument for why a standard is a good idea.

PR #3856, opened 2021-09-07
Removes Living toggling between review and living.

MicahZoltu:
I believe we decided in a previous EIPs meeting to just have Living stay living.  Having it toggle to review doesn't make sense, because then it means there is no living EIP-1 while it is in review.

cc @poojaranjan


axic (approved):

poojaranjan:
I suppose so. Thanks!

PR #3860, opened 2021-09-07
Add draft for initcode limiting & metering

axic:


EIPS/eip-3860.md
```diff
@@ -0,0 +1,123 @@
...
+## Motivation
+
+During contract creation the client has to perform jumpdest-analysis on the `initcode` prior to execution. The work performed scales linearly with the size of the `initcode`. This work currently is not metered, nor is there a protocol enforced upper bound for the size.
+
+There are three costs charged today:
+1. Cost for calldata aka `initcode`: 4 gas for a byte with the value of zero, and 16 gas otherwise.
+2. Cost for the resulting deployed code: 200 gas per byte.
+3. Cost of address calculation (hashing of code) in case of `CREATE2` only: 6 gas per word.
+
+Only the first cost applies to `initcode`, but only in the case of contract creation transactions. For the case of `CREATE`/`CREATE2` there is no such cost, and it is possible to programmatically generate variations of initcode in a relatively cheap manner. In the past it was possible to craft malicious `initcode` due to a vulnerability fixed in 2017 by [geth 1.6.5](https://github.com/ethereum/go-ethereum/releases/tag/v1.6.5).
```
MicahZoltu, line 31: Should I even bother commenting on the link here?  Motivation isn't the right place for history lessons.  Make the argument in text without relying on anecdotes.

MicahZoltu (approved):
Approving this under protest.

EIPS/eip-3860.md
axic: This was grandfathered in from the previous version of the EIP: https://eips.ethereum.org/EIPS/eip-2677

Are you saying a geth release is a risky link as it may change?

I'm happy to remove, but would like to get some confirmation from @holiman.

MicahZoltu: I'm saying the same thing that I always say, which is that external links should not be included in EIPs, and that hard line rules are easier for the 0.37 full time editors we have to enforce than fuzzy rules that make exceptions for some GitHub repositories but not others.

holiman: > Motivation isn't the right place for history lessons

Well, if there's a currently existing DoS attack due to how it works prior to the EIP, then that should be part of the motivation, no? 
And adding a reference "btw this actually happened in 2017" is a legit piece of useful information. 

I agree somewhat though, because the phrase "In the past it was possible to craft malicious `initcode` due to a vulnerability fixed in 2017" makes it sound like it has no bearing to the current state of things. A better phrasing would be  something like:

"Initcode-based attacks were carried out in 2017, prompting a new geth release to improve the resiliency. However, the root cause is has still not been addressed on the platform layer"

MicahZoltu: Why not just explain the attack vector being defended against and leave out the history?  What is the added value to someone wanting to implement this specification to have a discussion about prior attacks against the network?  Once this EIP is final and included in a hard fork it will just be a note for the history books, but not something future developers need to know about to implement this feature.

I'm a huge fan of keeping EIPs short and concise, and including historic details that are not relevant to someone implementing an Ethereum client (or tool) should (IMO) live elsewhere.

holiman: If the "motivation" section should explain why something is needed, then including historic details can help paint a fuller picture. It shows that it's not just a theoretical and/or future attack. To help the reader determine whether it's worth implementing this EIP, I think it's a relevant fact. 

I think you see "motivation" differently, or maybe you just have a higher opinion on the reader(s) ability to figure out the broader context by him/herself. 

I agree though, that the technical part of the eip should just be specification. In general, I wonder if the EIP workflow wouldn't have benefited from being split in two: one technical specification (highly detailed and technical, preferrably executable code), and one document describing the context, references, charts etc. Now we're in this spot which makes nobody happy :)

MicahZoltu: We are trying to move over to an executable specification, which @lightclient and others are working on.  I would definitely like to see EIPs be purely technical, which would mean dropping the motivation section which definitely is one of the primary sources of disagreement between various authors/editors.

You are correct that I see motivation differently than you from the sound of it.  I think the motivation section should be fairly short/concise and just give some rough ideas about "why this is useful".  I do not think that it should be a complete and thorough argument that is convincing on its own, which it seems like what some people desire.

My claim is that if you want to write a position essay with citations on an EIP, then use a blogging platform, not a standards repository.

PR #3981, opened 2021-09-17
EIP-3675: Hardcode terminal total difficulty

mkalinin:
### What's done?
Notes that `TERMINAL_TOTAL_DIFFICULTY` parameter must be hardcoded into client software distribution. See the corresponding change in the consensus layer for additional details and rationale https://github.com/ethereum/consensus-specs/pull/2605.


MicahZoltu:
This feels over-prescriptive for a specification.  How exactly clients go about doing releases and configuration should be up to the individual clients.  What is required for consensus is that clients get TTD from *somewhere*.  One can imagine a client that just fetches TTD from some centralized (trusted) source for example, or maybe from an on-chain DAO contract, or some other mechanism.

PR #4129, opened 2021-09-19
EIP 2844 stagnant (2021-Sep-19th@03.12.13)

alita-moore:
This EIP has not been active since (2020-Nov-25th@11.33.27); which, is greater than the allowed time of 6 months.

 authors: @oed 
 EIP Editors: @MicahZoltu, @lightclient, @arachnid, @cdetrio, @Souptacular, @vbuterin, @nicksavers, @wanderer, @gcolvin, @axic

### DIFF
--- a/EIPS/eip-2844.md
+++ b/EIPS/eip-2844.md
@@ -3,7 +3,7 @@ eip: 2844
 title: Add DID related methods to the JSON-RPC
 author: Joel Thorstensson (@oed)
 discussions-to: https://github.com/ethereum/EIPs/issues/2845
-status: Draft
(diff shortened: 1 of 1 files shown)


MicahZoltu:
@oed Currently, this EIP is just a draft, which means it is not a standard yet but rather just an idea for a standard.  If you would like to pursue turning this into a standard, you should make any adjustments you think are necessary and then create a PR that changes the status from `Draft` to `Review`.  Once you do that, please link the PR here and we'll close this out.  After that there are still additional steps before it becomes a standard, and you can check out EIP-1 for details on that process.

PR #4167, opened 2021-09-19
EIP 634 stagnant (2021-Sep-19th@04.20.59)

alita-moore:
This EIP has not been active since (2021-Jan-27th@03.18.27); which, is greater than the allowed time of 6 months.

 authors: @ricmoo 
 EIP Editors: @MicahZoltu, @lightclient, @arachnid, @cdetrio, @Souptacular, @vbuterin, @nicksavers, @wanderer, @gcolvin, @axic

### DIFF
--- a/EIPS/eip-634.md
+++ b/EIPS/eip-634.md
@@ -5,7 +5,7 @@ author: Richard Moore (@ricmoo)
 type: Standards Track
 discussions-to: https://github.com/ethereum/EIPs/issues/2439
 category: ERC
-status: Draft
+status: Stagnant
 created: 2017-05-17
 ---
 


ricmoo:
This should not be marked stagnant as it is still active. How do I move it back to draft? Or should I move it to review? @Arachnid

MicahZoltu:
@ricmoo If it is ready for review then you should submit a PR to the EIP to move it to review.  If it is not ready for review, then submit a PR moving it to draft.

When you say "it is still active", what do you mean by that?  The bot should only be flagging EIPs as stagnant when they haven't changed for 6 months and are not in Final.

ricmoo:
I meant that it is an EIP that has active implementations and dapps that are using it, like the ENS app. At this point it is mainly a place for new fields to be added. Is there a such thing as some sort of living EIP? For example, something akin to SLIP-0044 for tracking Chain ID’s?

MicahZoltu:
@ricmoo Any active implementations are unrelated to any formal standardization process, since this is a draft and there for there is no standard to follow.  If you would like to create a standard, then pushing this through the review => Last Call => Final process like normal would be the appropriate way to create such a standard.

As for living EIPs like SLIPs, we don't have a system for that.  The problem with such things is that they tend to not be maintained and fall out of date.  My recommendation is to create a website (managed externally by you or some other interested party) for tracking the list, and then create an EIP for each line item.  To simplify this process you could put together a nice template so the EIP process goes smoothly for each new addition.

alita-moore:
superceeded by https://github.com/ethereum/EIPs/pull/4310

PR #4189, opened 2021-09-19
2711: Withdrawing

MicahZoltu:
I still think this is a good idea, but I'm no longer actively championing it.  If anyone wants to take up ownership of this EIP or something similar to it please do so!


xinbenlv:
The idea is great, I see great usage. It would be great if someone could take over.

lightclient (approved):

PR #4194, opened 2021-09-20
3076: Move to Last Call + Interface

michaelsproul:
This PR moves EIP-3076: Slashing Protection Interchange Format to Last Call status.

The standard has been widely adopted by consensus clients, including Lighthouse, Teku, Prysm, Nimbus and Lodestar. No design issues have arisen during the course of implementation and use.

I have also reclassified the EIP as "Interface" rather than "Core", as it seems a better fit. Let me know if this change is not desirable.

I've set a review period for one month from today: 20th October 2021

### DIFF
--- a/EIPS/eip-3076.md
+++ b/EIPS/eip-3076.md
@@ -3,11 +3,12 @@ eip: 3076
 title: Slashing Protection Interchange Format
 author: Michael Sproul (@michaelsproul), Sacha Saint-Leger (@sachayves), Danny Ryan (@djrtwo)
 discussions-to: https://ethereum-magicians.org/t/eip-3076-validator-client-interchange-format-slashing-protection/
-status: Review
+status: Last Call
 type: Standards Track
-category: Core
+category: Interface
 created: 2020-10-27
-updated: 2021-02-08
+updated: 2021-09-20
+review-period-end: 2021-10-20
 ---
 
 ## Simple Summary


EIPS/eip-3076.md
```diff
@@ -3,11 +3,12 @@ eip: 3076
...
 type: Standards Track
-category: Core
+category: Interface
 created: 2020-10-27
-updated: 2021-02-08
+updated: 2021-09-20
+review-period-end: 2021-10-20
 ---
 
 ## Simple Summary
```
MicahZoltu, line 14: Simple summary section has been removed, and in its place a `description` header field has been added.  Please delete this section and add a frontmatter line item above like:
```md
description: A JSON interchange format for proof of stake validators to migrate slashing protection data between clients.
```

michaelsproul: Thanks, done.

MicahZoltu:
I'm leaving this for @axic or @lightclient to review.

Why is this going through the EIP process instead of the consensus client change control process?

michaelsproul:
> Why is this going through the EIP process instead of the consensus client change control process?

Honestly I'm not sure. As I understand it the EIP process will remain the governance process for the execution layer post merge, and changes won't "just" be made by pull request to [`execution-specs`](https://github.com/ethereum/execution-specs)? Conversely I think consensus is likely to continue to be governed by PRs alone, in which case maybe everything consensus-related should live in [`consensus-specs`](https://github.com/ethereum/consensus-specs).

The reason this was initially raised as an EIP is because it's a standalone component separate from the rest of consensus, and I was following the precedent set by [EIP-2333](https://eips.ethereum.org/EIPS/eip-2333) and [EIP-2335](https://eips.ethereum.org/EIPS/eip-2335) (and I was encouraged by @djrtwo). Those EIPs are similar in scope to this one, although perhaps with broader applicability to other chains. They're both inactive currently, and I think their fate should probably be similar to EIP-3076 (cc @CarlBeek)?

All three EIPs have some inertia behind their names in discussions + code, which would get slightly muddied by a rename/relocation, but it would be tolerable.

MicahZoltu:
> Honestly I'm not sure. As I understand it the EIP process will remain the governance process for the execution layer post merge, and changes won't "just" be made by pull request to [`execution-specs`](https://github.com/ethereum/execution-specs)? Conversely I think consensus is likely to continue to be governed by PRs alone, in which case maybe everything consensus-related should live in [`consensus-specs`](https://github.com/ethereum/consensus-specs).

I think most of the EIP editors would like to move to a PR process like consensus spec has (with some tweaks perhaps) long term.  Whether that work is possible/ready by the time The Merge happens is incredibly uncertain though as we first need an executable execution spec.

lightclient (approved):

PR #4329, opened 2021-09-27
Move EIP-1191 to Final

poojaranjan:
The author is willing to move the proposal to "Final". The  "Last call" - Review period ended on  2019-11-18.


juli (approved):
Thanks  @poojaranjan

MicahZoltu:
See discussion at https://github.com/ethereum/EIPs/issues/4335 about registries.

bumblefudge:
> as it relies on a list of chains that have "adopted" the proposal

does it RELY ON that list? can't the list just be explicitly declared to be non-normative and/or non-exhaustive and/or for reference purposes?

PR #4331, opened 2021-09-27
2315 cleanup

axic:


MicahZoltu:
This should be two separate PRs.  One PR per EIP allows all of the tooling to work, bundling multiple EIP changes into one PR breaks basically all of our tooling.

axic:
> This should be two separate PRs. One PR per EIP allows all of the tooling to work, bundling multiple EIP changes into one PR breaks basically all of our tooling.

@alita-moore said the bot now works with multiple files again. However this here is just an example for @gcolvin to see how it can be done.

alita-moore:
Yeah this shouldn't be a problem @MicahZoltu

gcolvin:
The tooling might handle it, but I'd prefer separate transactions, if any.  Without a Superseded status, people who follow links from various places expecting to find the work that has been ongoing since March as EIP-2315 will find no indication that the work has moved elsewhere.  And there is a good chance that portions of this will get moved to separate EIPs, leaving the remainder much closer to where it was in March.  So I'll close this for now, and reopen when and if appropriate rather than risk a lot of confusion.  Thanks, @axic

skaillet91 (commented):
EIPS/eip-4331.md

Rambonner:
> EIPS/eip-4331. md

gh pr checkout 4331

PR #4340, opened 2021-10-01
Move EIP-3607 to review

MariusVanDerWijden:

 - The build passes.

### DIFF
--- a/EIPS/eip-3607.md
+++ b/EIPS/eip-3607.md
@@ -3,7 +3,7 @@ eip: 3607
 title: Reject transactions from senders with deployed code
 author: Dankrad Feist (@dankrad), Dmitry Khovratovich (@khovratovich), Marius van der Wijden (@MariusVanDerWijden)
 discussions-to: https://github.com/ethereum/EIPs/issues/3608
-status: Draft
+status: Review
 type: Standards Track
 category: Core
 created: 2021-06-10


EIPS/eip-3607.md
```diff
@@ -3,7 +3,7 @@ eip: 3607
 title: Reject transactions from senders with deployed code
```
MicahZoltu, line 3:
```suggestion
 title: Reject transactions from senders with deployed code
 description: Do not allow transactions for which `tx.sender` has any code deployed.
```
The `Simple Summary` section has been removed from the template and replaced with a `description` field in the frontmatter.  Along with this suggestion, please remove the Simple Summary section entirely.

MicahZoltu (changes requested):
The only blockers are the `description` metadata field, the removal of the Simple Summary, and the removal of an external link.  The rest of the comments below are not required, but I recommend considering them at least.

----

Remove the external link to tests.  I encourage you to format test cases in a language agnostic way and include them as an asset (like you did with the implementation diff).  We should not be linking to external content though.

That being said, the one test case included already is fine and you could just remove the last line in the Test Cases section to fulfill this requirement.  As mentioned below, it would be nice if the test was a bit more clear about what it means to "fail".

----

Can you make the specification a bit more clear on how exactly the transaction should be rejected?  Consider wording like this:
> If a block contains a transaction whose signature recovers to an address where the `CODEHASH` of the address is not equal to `EMPTYCODEHASH`, then the block **MUST** be considered invalid by the client, similar to if the the block's timestamp was prior to its parent's timestamp.

The goal here is to clearly differentiate between "the transaction is entirely invalid and should not be allowed in the blockchain" and "the transcation is valid but reverts rather than executes".

----

In the reference implementation, consider being more clear about "transaction sender" and specifying that it is the address recovered from the transaction signature.

----

> An diff to implement ...

should be
> A diff to implement ...

----

> Every transaction send by the ...

should be
> Every transaction sent by the ...

VladLupashevskyi:
Hi, I'm not sure whether I should add the comment here or in the actual PR (#3607) which was already merged...

I didn't find in the specification how this change must or must not affect the execution of RPC calls that don't affect the state like `eth_call` or `eth_estimateGas`.

I think it would make sense not to change the behaviour of such RPC calls, because it's handy to check that some call from let's say a mulisig contract will pass or not. In this case it's possible just to put contract address as transaction sender, add data and call `eth_call`. Moreover this would be a breaking change for people who already use this feature and it's not mentioned in `Backwards Compatibility` section.

eulerbeat:
Where can I find the source code (preferably in geth) for this change? Is it deployed?

I have seen a few address collisions on ETH, BSC here:
...

MariusVanDerWijden:
@eulerbeat https://github.com/ethereum/go-ethereum/blob/356bbe343a30789e77bb38f25983c8f2f2bfbb47/core/state_transition.go#L230
I doubt that there are address collisions though, we full synced the chain with this change enabled. Don't know about BSC

PR #4345, opened 2021-10-05
Create Ice Age EIP

timbeiko:
EIP to push back the difficulty bomb from December 2021 to April/May 2022.


jdetychey:
As the Merge is getting ready, why not merging and canceling the Ice Age all together?

EIPS/eip-4345.md
```diff
@@ -0,0 +1,59 @@
...
+author: Tim Beiko (@timbeiko)
+discussions-to: TBA
+status: Final
+type: Standards Track
+category: Core
+created: 2021-10-05
+---
+
+## Simple Summary
+Delays the difficulty bomb to show effect in April 2022.
```
lightclient, line 13: This section has been deprecated in favor of the `description` preamble field. Please see the `eip-template.md` to see how it's used.

EIPS/eip-4345.md
```diff
@@ -0,0 +1,59 @@
+---
+eip: 4345
+title: Difficulty Bomb Delay to April 2022
+author: Tim Beiko (@timbeiko)
+discussions-to: TBA
```
lightclient, line 5: Will need a discussion link before merging.

timbeiko:
Ideally, we never "reach" this difficulty bomb because we've merged before. But, if we haven't, I think it's better to have to push it back again. The alternatives are (1) remove the bomb altogether, which I don't think we should do (makes it easier for the PoW chain to keep going, also to launch scam forks) or (2) push the bomb _way_ back, which is sort of equivalent to removing it in this context. 

We should move this conversation here: https://ethereum-magicians.org/t/eip-4345-difficulty-bomb-delay-to-april-2022/7209 😄

lightclient (changes requested):

EIPS/eip-4345.md
timbeiko: That's fixed now :-)

EIPS/eip-4345.md
timbeiko: Fixed, thanks!

timbeiko:
Thanks for the prompt review, @lightclient ! Comments should be addressed by the last commits.

EIPS/eip-4345.md
```diff
@@ -0,0 +1,57 @@
+---
+eip: 4345
+title: Difficulty Bomb Delay to April 2022
+description: Delays the difficulty bomb to show effect in April 2022.
```
lightclient, line 4: I think this could be slightly clearer. Maybe something like "until April 2022" or "take effect in April 2022".

lightclient (changes requested):

EIPS/eip-4345.md
timbeiko: Changed, along with the actual month to better describe when the bomb will actually be noticeable.

lightclient: ty

EIPS/eip-4345.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Specification
+#### Relax Difficulty with Fake Block Number
+For the purposes of `calc_difficulty`, simply replace the use of `block.number`, as used in the exponential ice age component, with the formula:
+```py
+    fake_block_number = max(0, block.number - 10_500_000) if block.number >= FORK_BLOCK_NUMBER else block.number
+```
+## Rationale
+
+The following script predicts a .25 second delay to blocktime May 2022 and a 1 second delay by June. This gives reason to address because the effect will be seen, but not so much urgency we don't have space to work around if needed.
```
lightclient, line 27:
```suggestion
The following script predicts a 0.25 second delay to block time May 2022 and a 1 second delay by June. This gives reason to address because the effect will be seen, but not so much urgency we don't have space to work around if needed.
```

lightclient (approved):

PR #4351, opened 2021-10-08
EIP-3675: Execution engine must validate length of the block's extraData

asn-d6:
The [merge beacon-chain spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/merge/beacon-chain.md) has set `MAX_EXTRA_DATA_BYTES` to 32. The execution engine must validate that this is the case.

/cc @djrtwo @mkalinin

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -67,7 +67,7 @@ PoW blocks that are descendants of any terminal PoW block **MUST NOT** be import
 
 | Name | Value |
 |-|-|
-| **`MAX_BYTES_PER_EXTRA_DATA`** | `32` |
+| **`MAX_EXTRA_DATA_BYTES`** | `32` |
 
 ### Block structure
 
@@ -81,7 +81,7 @@ Beginning with `TRANSITION_BLOCK`, a number of previously dynamic block fields a
 | **`nonce`** | `0x0000000000000000` |  |
 | **`ommers`** | `[]` | `RLP([]) = 0xc0`  |
 
-Beginning with `TRANSITION_BLOCK`, the validation of the block's **`extraData`** field changes: The length of the block's **`extraData`** **MUST** be less or equal to **`MAX_BYTES_PER_EXTRA_DATA`** bytes.
+Beginning with `TRANSITION_BLOCK`, the validation of the block's **`extraData`** field changes: The length of the block's **`extraData`** **MUST** be less than or equal to **`MAX_EXTRA_DATA_BYTES`** bytes.
 
 *Note*: Logic and validity conditions of block fields that are *not* specified here **MUST** remain unchanged. Additionally, the overall block format **MUST** remain unchanged.
 
@@ -199,7 +199,7 @@ It is recommended for the client software to not propagate descendants of any te
 
 ### Restricting the length of `extraData`
 
-The `extraData` field has been specified as unbounded in the PoW spec even though mainnet and most PoW testnets only utilize `32` bytes for it.  `extraData` fields of greater length are used by clique testnets and other networks to carry special signature/consensus schemes. This EIP restricts the length of `extraData` to `32` bytes since any network that is transitioning from another consensus mechanism to a beacon chain PoS consensus mechanism no longer needs extended or unbounded `extraData`.
+The `extraData` field is defined as a maximum of `32` bytes in the yellow paper. Thus mainnet and most PoW testnets cap the value at `32` bytes.  `extraData` fields of greater length are used by clique testnets and other networks to carry special signature/consensus schemes. This EIP restricts the length of `extraData` to `32` bytes because any network that is transitioning from another consensus mechanism to a beacon chain PoS consensus mechanism no longer needs extended or unbounded `extraData`.
 
 ## Backwards Compatibility
 


EIPS/eip-3675.md
```diff
@@ -88,6 +88,7 @@ Beginning with `TRANSITION_BLOCK`, the block validity conditions **MUST** be alt
 * Remove all validation rules that are evaluated over the list of ommers and each member of this list.
 * Add verification of the fields noted in [block structure](#block-structure) as their specified constant value.
 * Add verification of block validity with respect to the PoS consensus rules, i.e. assert that there **MUST** be a corresponding `POS_CONSENSUS_VALIDATED` event for the block to consider it valid.
+* Add verification of the block's **`extraData`** field. i.e. the length of the block's **`extraData`** **MUST** be less or equal to 32 bytes.
```
djrtwo, line 91: Suggest "32 bytes" be defined as a constant earlier in the spec. e.g. `MAX_BYTES_PER_EXTRA_DATA`

djrtwo:
Let's add a variable to contain the `32` for this. And let's also add a discussion in "rationale" for this point

Point out that mainnet and most (all?) PoW ethereum testnets utilize `32` for this value, but that in clique it is unbounded and other networks tend to use it for special consensus/signature schemes.  Any network that is transitioning from another consensus mechanism to a beacon chain PoS consensus mechanism will no longer need extended or unbounded extra-data so we reduce to 32-bytes at the transition to better bound the consensus types

EIPS/eip-3675.md
```diff
@@ -63,6 +63,12 @@ The `TERMINAL_TOTAL_DIFFICULTY` parameter is a part of client software configura
 PoW blocks that are descendants of any terminal PoW block **MUST NOT** be imported. This implies that a terminal PoW block will be the last PoW block in the canonical chain.
 
 
+### Constants
+
+| Name | Value |
+|-|-|
+| **`MAX_BYTES_PER_EXTRA_DATA`** | `32` |
```
lightclient, line 70:
```suggestion
| **`MAX_EXTRA_DATA_BYTES`** | `32` |
```
Maybe update to this to be consistent with the consensus spec?

EIPS/eip-3675.md
```diff
@@ -75,6 +81,8 @@ Beginning with `TRANSITION_BLOCK`, a number of previously dynamic block fields a
 | **`nonce`** | `0x0000000000000000` |  |
 | **`ommers`** | `[]` | `RLP([]) = 0xc0`  |
 
+Beginning with `TRANSITION_BLOCK`, the validation of the block's **`extraData`** field changes: The length of the block's **`extraData`** **MUST** be less or equal to **`MAX_BYTES_PER_EXTRA_DATA`** bytes.
```
lightclient, line 84:
```suggestion
Beginning with `TRANSITION_BLOCK`, the validation of the block's **`extraData`** field changes: The length of the block's **`extraData`** **MUST** be less or equal to **`MAX_EXTRA_DATA_BYTES`** bytes.
```
And here.

EIPS/eip-3675.md
lightclient, line 70:
```suggestion
| **`MAX_EXTRA_DATA_BYTES`** | `32` |
```
Maybe update to this to be consistent with the consensus spec?

lightclient, line 70:
```suggestion
| **`MAX_EXTRA_DATA_BYTES`** | `32` |
```
Maybe update to this to be consistent with the consensus spec?

djrtwo: no need for "PER" here. I was mistaken. When it is the length of a single list in a block, we don't use "PER". As in `MAX_ATTESTATIONS`. "PER" is used more for conversions

Let's use @lightclient's suggestion here

EIPS/eip-3675.md
djrtwo: suggest "less than or" rather than "less or"

EIPS/eip-3675.md
```diff
@@ -189,6 +197,9 @@ After the upgrade of the consensus mechanism only the beacon chain network will
 
 It is recommended for the client software to not propagate descendants of any terminal PoW block to reduce the load on processing the P2P component and stop operating in the environment with unknown security properties.
 
+### Restricting the length of `extraData`
+
+The `extraData` field has been specified as unbounded in the PoW spec even though mainnet and most PoW testnets only utilize `32` bytes for it.  `extraData` fields of greater length are used by clique testnets and other networks to carry special signature/consensus schemes. This EIP restricts the length of `extraData` to `32` bytes since any network that is transitioning from another consensus mechanism to a beacon chain PoS consensus mechanism no longer needs extended or unbounded `extraData`.
```
djrtwo, line 202: The yellow paper (which defines ethereum pow/mainnet consensus) notes that "... extraData must be at most
32 bytes." Thus it is not unbounded for PoW network spec as noted in the original comment.

Suggest using the following text (or similar):

```suggestion
The `extraData` field is defined as a maximum of `32` bytes in the yellow paper. Thus mainnet and most PoW testnets cap the value at `32` bytes.  `extraData` fields of greater length are used by clique testnets and other networks to carry special signature/consensus schemes. This EIP restricts the length of `extraData` to `32` bytes because any network that is transitioning from another consensus mechanism to a beacon chain PoS consensus mechanism no longer needs extended or unbounded `extraData`.
```

EIPS/eip-3675.md
asn-d6: ACK! Done in db1c213.

EIPS/eip-3675.md
asn-d6: Thanks for the suggestion. Adopted in db1c213.

asn-d6:
Thanks for the review! I pushed db1c213 as a fixup commit. You can squash it during rebase with `--autosquash` but if you would like me to squash it and force-push just let me know. Cheers!

PR #4357, opened 2021-10-10
remove POS_CONSENSUS_VALIDATED from EIP 3675

djrtwo:
Remove `POS_CONSENSUS_VALIDATED` from EIP 3675.

This event was removed for two reasons:
1. This event was an unnecessary optimization to allow for pruning of "bad" blocks from the block tree. This optimization was unnecessary because the PoS consensus would never send `POS_FORKCHOICE_UPDATED` for any such bad blocks or their descendants, and eventually any such blocks would be able to be pruned after a PoS finality event of an alternative branch in the block tree.
2. This event was dangerous in some scenarios because a block could be referenced by two *different* and conflicting PoS branches. Thus for the same block in some scenarios, both a `POS_CONSENSUS_VALIDATED == True` and `POS_CONSENSUS_VALIDATED == FALSE` event could sent, entirely negating the ability to safely perform the optimization in (1).

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -41,7 +41,6 @@ There can be more than one terminal PoW block in the network, e.g. multiple chil
 * **`TERMINAL_TOTAL_DIFFICULTY`** The amount of total difficulty reached by the network that triggers the consensus upgrade.
 * **`TRANSITION_BLOCK`** The earliest PoS block of the canonical chain, i.e. the PoS block with the lowest block height.
 `TRANSITION_BLOCK` **MUST** be a child of a terminal PoW block.
-* **`POS_CONSENSUS_VALIDATED`** An event occurring when the PoS block is validated with respect to the proof-of-stake consensus rules.
 * **`POS_FORKCHOICE_UPDATED`** An event occurring when the state of the proof-of-stake fork choice is updated.
 
 #### PoS events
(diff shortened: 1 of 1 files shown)


EIPS/eip-3675.md
```diff
@@ -177,6 +175,14 @@ Existing rewards for producing and sealing blocks are deprecated along with the
...
 
+### Remove of `POS_CONSENSUS_VALIDATED`
+
+In prior draft versions of this EIP, an additional POS event -- `POS_CONSENSUS_VALIDATED` -- was required as a validation condition for blocks. This event gave the signal to either fully incorporate or prune the block from the block tree.
+
+This event was removed for two reasons:
+1. This event was an unnecessary optimization to allow for pruning of "bad" blocks from the block tree. This optimization was unnecessary because the PoS consensus would never send `POS_FORKCHOICE_UPDATED` for any such bad blocks or their descendants, and eventually any such blocks would be able to be pruned after a PoS finality event of an alternative branch in the block tree.
+2. This event was dangerous in some scenarios because a block could be referenced by two *different* and conflicting PoS branches. Thus for the same block in some scenarios, both a `POS_CONSENSUS_VALIDATED == True` and `POS_CONSENSUS_VALIDATED == FALSE` event could sent, entirely negating the ability to safely perform the optimization in (1).
```
mkalinin, line 184: Nit: 
> `POS_CONSENSUS_VALIDATED == True` and `POS_CONSENSUS_VALIDATED == FALSE`

Should be `POS_CONSENSUS_VALIDATED == TRUE` and `POS_CONSENSUS_VALIDATED == FALSE` -- we can fix the inconsistency of the case between `True` and `FALSE` during the next update

djrtwo: whoops! I'll submit a tiny PR

PR #4361, opened 2021-10-11
Create EIP-4361: Sign-In with Ethereum

wyc:
This is the draft PR for [Sign-In with Ethereum](https://notes.ethereum.org/@djrtwo/sign-in-with-ethereum-RFP). Development is happening in the open on [community calls](https://forms.gle/bJX99EJqExq66J2c7). 

Latest updates about the project can be found at https://login.xyz

TODO:
- [x] Add links to first mentions of RFC/EIP specifications
- [x] Complete WIP sections
- [x] Fix failing CI build
- [x] Continue discussion and incorporate feedback

### DIFF
--- a/EIPS/eip-4361.md
+++ b/EIPS/eip-4361.md
@@ -40,9 +40,9 @@ I accept the ServiceOrg Terms of Service: https://service.org/tos
 
 URI: https://service.org/login
 Version: 1
-Chain ID: 1
-Nonce: 328917
+Nonce: 32891756
 Issued At: 2021-09-30T16:25:24Z
+Chain ID: 1
 Resources:
 - ipfs://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq/
 - https://example.com/my-web2-claim.json
@@ -58,12 +58,12 @@ ${statement}
 
 URI: ${uri}
 Version: ${version}
-Chain ID: ${chain-id}
 Nonce: ${nonce}
 Issued At: ${issued-at}
 Expiration Time: ${expiration-time}
 Not Before: ${not-before}
 Request ID: ${request-id}
+Chain ID: ${chain-id}
 Resources:
 - ${resources[0]}
 - ${resources[1]}
@@ -76,12 +76,12 @@ Resources:
 - `statement` (optional) is a human-readable ASCII assertion that the user will sign, and it must not contain `'\n'`.
 - `uri` is an RFC 3986 URI referring to the resource that is the subject of the signing (as in the _subject of a claim_).
 - `version` is the current version of the `message`, which MUST be `1` for this specification.
-- `chain-id` is the EIP-155 Chain ID to which the session is bound, and the network where Contract Accounts must be resolved.
 - `nonce` is a randomized token used to prevent replay attacks, at least 8 alphanumeric characters.
 - `issued-at` is the ISO 8601 datetime string of the current time.
 - `expiration-time` (optional) is the ISO 8601 datetime string that, if present, indicates when the signed authentication message is no longer valid.
 - `not-before` (optional) is the ISO 8601 datetime string that, if present, indicates when the signed authentication message will become valid.
 - `request-id` (optional) is an system-specific identifier that may be used to uniquely refer to the sign-in request.
+- `chain-id` is the EIP-155 Chain ID to which the session is bound, and the network where Contract Accounts must be resolved.
 - `resources` (optional) is a list of information or references to information the user wishes to have resolved as part of authentication by the relying party. They are expressed as RFC 3986 URIs separated by `"\n- "`.
 
 ### ABNF Message Format
@@ -89,20 +89,20 @@ Resources:
 The `message` to be signed MUST conform with the following Augmented Backus–Naur Form (ABNF, RFC 5234) expression (note that `%s` denotes case sensitivity for a string term, as per RFC 7405).
 ```abnf
 sign-in-with-ethereum =
-    domain %s"wants you to sign in with your Ethereum account:" LF
+    domain %s" wants you to sign in with your Ethereum account:" LF
     address LF
     LF
     [ statement LF ]
     LF
-    %s"URI: " url LF
+    %s"URI: " uri LF
     %s"Version: " version LF
     %s"Chain ID: " chain-id LF
     %s"Nonce: " nonce LF
-    %s"Issued At: " issued-at LF
-    [ %s"Expiration Time: " expiration-time LF ]
-    [ %s"Not Before: " not-before LF ]
-    [ %s"Request ID: " request-id LF ]
-    [ %s"Resources:" LF
+    %s"Issued At: " issued-at
+    [ LF %s"Expiration Time: " expiration-time ]
+    [ LF %s"Not Before: " not-before ]
+    [ LF %s"Request ID: " request-id ]
+    [ LF %s"Resources:"
     resources ]
 
 domain = dnsauthority
@@ -116,10 +116,10 @@ address = "0x" 40*40HEXDIG
     ; checksum encoding specified in EIP-55
     ; where applicable (EOAs).
 
-statement = *( reserved / unreserved )
+statement = *( reserved / unreserved / " " )
     ; See RFC 3986 for the definition
     ; of "reserved" and "unreserved".
-    ; The purpose is to exclude LF ('\n').
+    ; The purpose is to exclude LF (line break).
 
 uri = URI
     ; See RFC 3986 for the definition of "URI".
@@ -142,7 +142,7 @@ not-before = date-time
 request-id = *pchar
     ; See RFC 3986 for the definition of "pchar".
 
-resources = resource *( LF resource )
+resources = *( LF resource )
 
 resource = "- " URI
 ```
--- a/EIPS/eip-4361.md
+++ b/EIPS/eip-4361.md
@@ -2,7 +2,7 @@
 eip: 4361
 title: Sign-In with Ethereum.
 description: Standardizes off-chain authentication for Ethereum accounts to establish sessions.
-author: Wayne Chang (@wyc), Brantly Millegan (@brantlymillegan), Nick Johnson (@Arachnid)
+author: Wayne Chang (@wyc), Gregory Rocco (@obstropolos), Brantly Millegan (@brantlymillegan), Nick Johnson (@Arachnid) 
 discussions-to: https://ethereum-magicians.org/t/eip-4361-sign-in-with-ethereum/7263
 status: Draft
 type: Standards Track
@@ -245,9 +245,10 @@ The following concerns are out of scope for this version of the specification to
 - Requirements were gathered from existing implementations of similar sign-in workflows, including statements to allow the user to accept a Terms of Service, nonces for replay protection, and inclusion of the Ethereum address itself in the message.
 
 ## Forwards Compatibility
-- DIDs/VCs/Cross-Chain?
-- SIOPv2?
-- EIP-712? Which version?
+- Possible support for Decentralized Identifiers and Verifiable Credentials. 
+- Possible cross-chain support.
+- Possible SIOPv2 support.
+- Possible future support for EIP-712. 
 - Version interpretation rules, e.g., sign with minor revision greater than understood, but not greater major version.
 
 ## Test Cases
@@ -296,6 +297,16 @@ There are several cases where an implementer SHOULD check for state changes as t
 
 #### Maximum lengths for ABNF terms
 - While this specification does not contain normative requirements around maximum string lengths, implementers SHOULD choose maximum lengths for terms that strike a balance across the prevention of denial of service attacks, support for arbitrary use cases, and user readability.
+
+## Additional Resources
+Throughout the development process of this specification, community participation was encouraged and facilitated via weekly calls. This was to ensure coverage on major areas of concern, and accommodate a wide and diverse range of feedback from the Ethereum community. 
+
+Additionally, resources and research were made public both in the form of the community calls, the minutes being published, and blog posts. Those resources can be found at the following locations: 
+
+- [Login.xyz](https://login.xyz) - which served as the primary source of specification updates, news, and community call minutes.
+- [An initial overview of required workflows](https://blog.spruceid.com/sign-in-with-ethereum-proposed-architecture/) - which helped inform the initial kickoff and discussions on the specification. 
+- [Research on related EIPs](https://blog.spruceid.com/ethereum-identity-review-of-eips-over-time/) - which helped identify the necessary requirements around necessary EIPs to support.
+- [A comparison of EIP-191 vs. EIP-712](https://blog.spruceid.com/sign-in-with-ethereum-wallet-support-eip-191-vs-eip-712/) - which helped lead to the decision to use EIP-191 in the first version of EIP-4361. 
 
 ## Copyright Waiver
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/EIPS/eip-4361.md
+++ b/EIPS/eip-4361.md
@@ -1,0 +1,295 @@
+---
+eip: 4361
+title: Sign-In with Ethereum. 
+description: Standardizes off-chain authentication for Ethereum accounts to establish sessions.
+author: Wayne Chang (@wyc), Brantly Millegan (@brantlymillegan), Nick Johnson (@Arachnid)
+resolution: <TBD>
+discussions-to: https://github.com/ethereum/EIPs/pull/4361
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-10-11
+requires: 55, 191, 137, 155, 1271, 1328
+---
+
+## Abstract
+Sign-In with Ethereum describes how Ethereum accounts authenticate with off-chain services by signing a standard message format parameterized by scope, session details, and security mechanisms (e.g., a nonce). The goals of this specification are to provide a self-custodied alternative to centralized identity providers, improve interoperability across off-chain services for Ethereum-based authentication, and provide wallet vendors a consistent machine-readable message format to achieve improved user experiences and consent management.
+
+## Motivation
+When signing in to popular non-blockchain services today, users will typically use identity providers (IdPs) that are centralized entities with ultimate control over users' identifiers, for example, large internet companies and email providers. Incentives are often misaligned between these parties. Sign-In with Ethereum offers a new self-custodial option for users who wish to assume more control and responsibility over their own digital identity.
+
+Already, many services support workflows to authenticate Ethereum accounts using message signing, such as to establish a cookie-based web session which can manage privileged metadata about the authenticating address. This is an opportunity to standardize the sign-in workflow and improve interoperability across existing services, while also providing wallet vendors a reliable method to identify signing requests as Sign-In with Ethereum requests for improved UX.
+
+This work is sponsored by the Ethereum Foundation and Ethereum Name Service (ENS). It is being developed in the open through a series of recorded community calls and public repositories, and its development is informed by over twenty user interviews with a focus on currently-in-production uses, related prior EIPs, and fits within product roadmaps.
+
+## Specification
+Sign-In with Ethereum works as follows:
+
+1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the EIP-191 signature scheme (string prefixed with `\x19Ethereum Signed Message:\n<length of message>`). The `message` MUST incorporate an Ethereum `address`, `version` of the message, `uri` for scoping, `nonce` acceptable to the server, and `issued-at` timestamp.
+2. The signature is then presented to the server, which checks the signature's validity and message content.
+3. Additional fields, including `expiration-time`, `not-before`, `request-id`, `chain-id`, and `resources` may be incorporated as part of authentication for the session.
+4. The server may further fetch data associated with the Ethereum address, such as from the Ethereum blockchain (e.g., ENS, account balances, ERC-20/ERC-721/ERC-1155 asset ownership), or other data sources that may or may not be permissioned.
+
+
+### Example message to be signed
+```
+service.org wants you to sign in with your Ethereum account:
+0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
+
+I accept the ServiceOrg Terms of Service: https://service.org/tos
+
+URI: https://service.org/login
+Version: 1
+Chain ID: 1
+Nonce: 328917
+Issued At: 2021-09-30T16:25:24Z
+Resources:
+- ipfs://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq/
+- https://example.com/my-web2-claim.json
+```
+
+### Informal Message Template
+A Bash-like informal template of the full message is presented below for readability and ease of understanding. Field descriptions are provided in the following section. A full ABNF description is provided in the section thereafter.
+```
+${domain} wants you to sign in with your Ethereum account:
+${address}
+
+${statement}
+
+URI: ${uri}
+Version: ${version}
+Chain ID: ${chain-id}
+Nonce: ${nonce}
+Issued At: ${issued-at}
+Expiration Time: ${expiration-time}
+Not Before: ${not-before}
+Request ID: ${request-id}
+Resources:
+- ${resources[0]}
+- ${resources[1]}
+...
+- ${resources[n]}
+```
+### Message Field Descriptions
+- `domain` is the RFC 4501 `dnsauthority` that is requesting the signing.
+- `address` is the Ethereum address performing the signing conformant to capitalization encoded checksum specified in EIP-55 where applicable.
+- `statement` (optional) is a human-readable ASCII assertion that the user will sign, and it must not contain `'\n'`.
+- `uri` is an RFC 3986 URI referring to the resource that is the subject of the signing (as in the _subject of a claim_).
+- `version` is the current version of the `message`, which MUST be `1` for this specification.
+- `chain-id` is the EIP-155 Chain ID to which the session is bound, and the network where Contract Accounts must be resolved.
+- `nonce` is a randomized token used to prevent replay attacks, at least 8 alphanumeric characters.
+- `issued-at` is the ISO 8601 datetime string of the current time.
+- `expiration-time` (optional) is the ISO 8601 datetime string that, if present, indicates when the signed authentication message is no longer valid.
+- `not-before` (optional) is the ISO 8601 datetime string that, if present, indicates when the signed authentication message will become valid.
+- `request-id` (optional) is an system-specific identifier that may be used to uniquely refer to the sign-in request.
+- `resources` (optional) is a list of information or references to information the user wishes to have resolved as part of authentication by the relying party. They are expressed as RFC 3986 URIs separated by `"\n- "`.
+
+### ABNF Message Format
+
+The `message` to be signed MUST conform with the following Augmented Backus–Naur Form (ABNF, RFC 5234) expression (note that `%s` denotes case sensitivity for a string term, as per RFC 7405).
+```abnf
+sign-in-with-ethereum =
+    domain %s"wants you to sign in with your Ethereum account:" LF
+    address LF
+    LF
+    [ statement LF ]
+    LF
+    %s"URI: " url LF
+    %s"Version: " version LF
+    %s"Chain ID: " chain-id LF
+    %s"Nonce: " nonce LF
+    %s"Issued At: " issued-at LF
+    [ %s"Expiration Time: " expiration-time LF ]
+    [ %s"Not Before: " not-before LF ]
+    [ %s"Request ID: " request-id LF ]
+    [ %s"Resources:" LF
+    resources ]
+
+domain = dnsauthority
+    ; From RFC 4501:
+    ;     dnsauthority = host [ ":" port ]
+    ; See RFC 4501 for the fully contextualized
+    ; definition of "dnsauthority".
+
+address = "0x" 40*40HEXDIG
+    ; Must also conform to captilization
+    ; checksum encoding specified in EIP-55
+    ; where applicable (EOAs).
+
+statement = *( reserved / unreserved )
+    ; See RFC 3986 for the definition
+    ; of "reserved" and "unreserved".
+    ; The purpose is to exclude LF ('\n').
+
+uri = URI
+    ; See RFC 3986 for the definition of "URI".
+
+version = "1"
+
+chain-id = 1*DIGIT
+    ; See EIP-155 for valid CHAIN_IDs.
+
+nonce = 8*( ALPHA / DIGIT )
+    ; See RFC 5234 for the definition
+    ; of "ALPHA" and "DIGIT".
+
+issued-at = date-time
+expiration-time = date-time
+not-before = date-time
+    ; See RFC 3339 for the
+    ; definition of "date-time".
+
+request-id = *pchar
+    ; See RFC 3986 for the definition of "pchar".
+
+resources = resource *( LF resource )
+
+resource = "- " URI
+```
+
+#### Signing and Verifying with Ethereum Accounts
+- For Externally Owned Accounts (EOAs), the verification method specified in EIP-191 MUST be used.
+- For Contract Accounts,
+    - The verification method specified in EIP-1271 MUST be used.
+    - When performing EIP-1271 signature verification, the contract performing the verification MUST be resolved from the specified `chain-id`.
+    - Implementers SHOULD take into consideration that EIP-1271 implementations are not required to be pure functions, and can return different results for the same inputs depending on blockchain state. This can affect the security model and session validation rules. For example, a service with EIP-1271 signing enabled may rely on webhooks to receive notifications when state affecting the results is changed. When it receives a notification, it invalidates any matching sessions.
+
+### Resolving Ethereum Name Service (ENS) Data
+- The server or wallet MAY additionally perform resolution of ENS data, as this can improve the user experience by displaying a human-friendly information that is related to the `address`. Resolvable ENS data include:
+    - The [primary ENS name](https://docs.ens.domains/dapp-developer-guide/resolving-names#reverse-resolution).
+    - The [ENS avatar](https://gist.github.com/Arachnid/9db60bd75277969ee1689c8742b75182).
+    - Any other resolvable resources specified in [the documentation](https://docs.ens.domains/dapp-developer-guide/resolving-names#looking-up-other-resources).
+- If resolution of ENS data is performed, implementers SHOULD take precautions to preserve user privacy and consent, as their `address` could be forwarded to third party services as part of the resolution process.
+
+### Server-Side (Relying Party) Implementer Guidelines
+
+#### Verifying a signed `message`
+- The message MUST be checked for conformance to the ABNF above, checked against expected term values after parsing, and its signature must be verified.
+
+#### Creating sessions
+- Sessions MUST be bound to the `address` and not to further resolved resources that may change.
+
+#### Interpreting and resolving `resources`
+- The listed `resources` must be RFC 3986 URIs, but their interpretation is out of scope of this specification. 
+- Implementers SHOULD ensure that that URIs are human-friendly when expressed in plaintext form.
+
+### Wallet Implementer Guidelines
+
+#### Verifying `message`
+- The full `message` MUST be checked for conformance to the ABNF above.
+
+#### Verifying `domain` binding
+- Wallet implementers MUST prevent phishing attacks by matching on the `domain` term when processing a signing request. For example, when processing the message beginning with `"service.org wants you to sign in..."`, the wallet checks that the request actually originated from `service.org`.
+- The domain SHOULD be read from a trusted data source such as the browser window or over WalletConnect (EIP-1328) sessions for comparison against the signing message contents.
+
+#### Creating Sign-In with Ethereum interfaces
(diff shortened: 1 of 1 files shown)


EIPS/eip-4361.md
```diff
@@ -0,0 +1,295 @@
...
+- `nonce` is a randomized token used to prevent replay attacks, at least 8 alphanumeric characters.
+- `issued-at` is the ISO 8601 datetime string of the current time.
+- `expiration-time` (optional) is the ISO 8601 datetime string that, if present, indicates when the signed authentication message is no longer valid.
+- `not-before` (optional) is the ISO 8601 datetime string that, if present, indicates when the signed authentication message will become valid.
+- `request-id` (optional) is an system-specific identifier that may be used to uniquely refer to the sign-in request.
+- `resources` (optional) is a list of information or references to information the user wishes to have resolved as part of authentication by the relying party. They are expressed as RFC 3986 URIs separated by `"\n- "`.
+
+### ABNF Message Format
+
+The `message` to be signed MUST conform with the following Augmented Backus–Naur Form (ABNF, RFC 5234) expression (note that `%s` denotes case sensitivity for a string term, as per RFC 7405).
```
axic, line 89: Why is it preferable to have a new invented language and not some binary format, which is easier to parse (CBOR or even ERC-712)?

Is the only reason because the signed message will be "readily" displayed in wallets without implementing support for it? If so, I think hardware wallets may be a big exception to that where some other more easily parsable format is more likely to gain support.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,295 @@
+---
+eip: 4361
+title: Sign-In with Ethereum. 
+description: Standardizes off-chain authentication for Ethereum accounts to establish sessions.
+author: Wayne Chang (@wyc), Brantly Millegan (@brantlymillegan), Nick Johnson (@Arachnid)
+resolution: <TBD>
```
MicahZoltu, line 6:
```suggestion
```
This is not an EIP frontmatter field.

NizamShah7 (approved):

EIPS/eip-4361.md
wyc: Hi @axic, thanks for your concern. That is indeed the main reason, so that applications may adopt this specification without full buy-in from wallet vendors or significant degradation of their user experiences today. In our proposal response to the RFP, we actually specified EIP-712 as the signing format, but found the user experience across many wallets to be much worse than using EIP-191. Formats like CBOR, EIP-712, protobuf, etc. indeed specify structuring for data, but hardware wallets have spotty support for presentation of EIP-712 requests to the users already (please see ongoing issues https://github.com/LedgerHQ/app-ethereum/issues/105 and https://github.com/trezor/trezor-firmware/pull/1568). We believe that if adoption of this specification relies on all wallet vendors upgrading how they do signing first, it is far less likely to see success. While pure technical merit is important to consider, adoption depends far more on talking to downstream users and understanding their concerns, what they are likely to truly adopt, etc. We have completed over 30 interviews towards this conclusion. We welcome anyone to let us know of any further user research in support or incongruent with this claim. Hope that helps!

EIPS/eip-4361.md
wyc: @MicahZoltu from https://eips.ethereum.org/EIPS/eip-1#eip-header-preamble
```
` * resolution:` a url pointing to the resolution of this EIP
```

Is there a more updated version you could point me to that reflects its omission? Thanks

wyc:
Hi, I have opened a thread here to continue all discussion: https://ethereum-magicians.org/t/eip-4361-sign-in-with-ethereum/7263/2

EIPS/eip-4361.md
MicahZoltu: https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md I'll get a PR out to fix EIP-1 to match.

Ignore NizamShah7, it is just a spam account.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,304 @@
...
+Request ID: ${request-id}
+Chain ID: ${chain-id}
+Resources:
+- ${resources[0]}
+- ${resources[1]}
+...
+- ${resources[n]}
+```
+### Message Field Descriptions
+- `domain` is the [RFC 4501](https://datatracker.ietf.org/doc/html/rfc4501) `dnsauthority` that is requesting the signing.
```
SamWilsn, line 74: Should this be `dnsname`? Seems like `dnsauthority` is the server you're resolving against.

```suggestion
- `domain` is the [RFC 4501](https://datatracker.ietf.org/doc/html/rfc4501) `dnsname` that is requesting the signing.
```

> If an authority ("dnsauthority") is given in the URI string, this indicates the server that should receive the DNS query. Otherwise, the default DNS server should receive it.

_[source](https://datatracker.ietf.org/doc/html/rfc4501#page-4)_

EIPS/eip-4361.md
```diff
@@ -0,0 +1,304 @@
...
+- `address` is the Ethereum address performing the signing conformant to capitalization encoded checksum specified in [EIP-55](https://eips.ethereum.org/EIPS/eip-55) where applicable.
+- `statement` (optional) is a human-readable ASCII assertion that the user will sign, and it must not contain `'\n'`.
```
SamWilsn, line 76:
```suggestion
- `statement` (optional) is a human-readable ASCII assertion that the user will sign, and it must not contain `'\n'` (the byte `0x0a`).
```

SamWilsn, line 76: I might suggest also excluding `:` (colon) to avoid confusing the assertion with the subsequent metadata.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,304 @@
...
+- `uri` is an [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) URI referring to the resource that is the subject of the signing (as in the _subject of a claim_).
+- `version` is the current version of the `message`, which MUST be `1` for this specification.
+- `nonce` is a randomized token used to prevent replay attacks, at least 8 alphanumeric characters.
+- `issued-at` is the [ISO 8601](https://datatracker.ietf.org/doc/html/rfc3339) datetime string of the current time.
+- `expiration-time` (optional) is the [ISO 8601](https://datatracker.ietf.org/doc/html/rfc3339) datetime string that, if present, indicates when the signed authentication message is no longer valid.
```
SamWilsn, line 81: Maybe rename to `not-after` for symmetry with `not-before`?

---

Does if the time is exactly equal to `expiration-time`, is the signature valid?

EIPS/eip-4361.md
```diff
@@ -0,0 +1,304 @@
...
+
+This work is sponsored by the Ethereum Foundation and Ethereum Name Service (ENS). It is being developed in the open through a series of recorded community calls and public repositories, and its development is informed by over twenty user interviews with a focus on currently-in-production uses, related prior EIPs, and fits within product roadmaps.
+
+## Specification
+Sign-In with Ethereum works as follows:
+
+1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the [EIP-191](https://eips.ethereum.org/EIPS/eip-191) signature scheme (string prefixed with `\x19Ethereum Signed Message:\n<length of message>`). The `message` MUST incorporate an Ethereum `address`, `version` of the message, `uri` for scoping, `nonce` acceptable to the server, and `issued-at` timestamp.
```
SamWilsn, line 28: Does reusing `\x19Ethereum Signed Message:\n<length of message>` pose a security risk?

If an attacker on a visually similar phishing domain uses the existing `eth_signMessage` endpoint to sign a message, do we expect existing wallets to display a warning?

cfelde: I don't quite understand the question here related to the prefix (as mentioned in EIP-191) and domains?

If we want the wallet to display a warning then we'd need to require the wallet to parse a domain from the message and check this against the domain of the website in use. Should wallets do that? Sure, it would be nice and good UX as far as I'm concerned.

But how does this relate to the prefix?

EIPS/eip-4361.md
wyc: Hi, just wanted to follow up on this thread. I understand that people like structured signing and this decision may be unpopular, but we did the work and this is how it looks like in wallets today: https://blog.spruceid.com/sign-in-with-ethereum-wallet-support-eip-191-vs-eip-712/

Applications we spoke to are not happy to degrade their UX significantly, and people have little idea what they are signing if confronted with a JSON blob. Therefore, I think EIP-191 is still the right move here. 

If you'd like to consider CBOR, EIP-712, protobuf, etc., I would need some evidence that indeed broad wallet support will come for these soon so users can see what they're signing. We added a `version` term to allow this as an upgrade path in the future, but for this first iteration of the spec it seems premature. Thanks so much for your review and consideration.

EIPS/eip-4361.md
SamWilsn: If you use a separate prefix, existing RPC endpoints like `eth_signMessage` will generally refuse to sign, so the wallets don't need to implement any special checks on _all_ signed messages, only on ones that use the new prefix.

If you use the same prefix, wallets have to scan all `eth_signMessage` bodies for messages that parse as EIP-4361 compatible, and then check the domain.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,295 @@
...
+
+This work is sponsored by the Ethereum Foundation and Ethereum Name Service (ENS). It is being developed in the open through a series of recorded community calls and public repositories, and its development is informed by over twenty user interviews with a focus on currently-in-production uses, related prior EIPs, and fits within product roadmaps.
+
+## Specification
+Sign-In with Ethereum works as follows:
+
+1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the EIP-191 signature scheme (string prefixed with `\x19Ethereum Signed Message:\n<length of message>`). The `message` MUST incorporate an Ethereum `address`, `version` of the message, `uri` for scoping, `nonce` acceptable to the server, and `issued-at` timestamp.
```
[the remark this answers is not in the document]
kamescg: I tend to agree with @PhABC here. It's not completely obvious the requirement for a `nonce` and what benefit nonce actually providers for "the most developers" category.

Sure, for med/high security situations, it's important to re-authenticate a user before allowing roles based access control to resources (compute or otherwise), but I would argue "generic" authentication uses a timestamps for session management and if a re-authentication is required you invalidate the active session signature/JWT

Could easily see most services constantly using the `0` or `1` nonce, because they have too, but don't have a real need for it.

If a server is going to be "stateful it would be good to research what's the most common techniques for invalidation of active sessions. To my knowledge a lot services will just invalidate a previous session cookie and require a user re-authenticate, which nullifies the requirement for a nonce anyways, because they can be handled via timestamp parameters.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,295 @@
...
+...
+- ${resources[n]}
+```
+### Message Field Descriptions
+- `domain` is the RFC 4501 `dnsauthority` that is requesting the signing.
+- `address` is the Ethereum address performing the signing conformant to capitalization encoded checksum specified in EIP-55 where applicable.
+- `statement` (optional) is a human-readable ASCII assertion that the user will sign, and it must not contain `'\n'`.
+- `uri` is an RFC 3986 URI referring to the resource that is the subject of the signing (as in the _subject of a claim_).
+- `version` is the current version of the `message`, which MUST be `1` for this specification.
+- `chain-id` is the EIP-155 Chain ID to which the session is bound, and the network where Contract Accounts must be resolved.
```
kamescg: Why is the `chain-id` not an optional parameter? Also why is the session being bound to chainId to start with?

What does `resolved` in thus context mean for Contract Accounts?

EIPS/eip-4361.md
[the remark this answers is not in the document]
kdenhartog: Hmm maybe, I'm missing something here but can someone explain why login is necessary in a stateless app interaction? I want to make sure I understand the architecture of how this is being used since I think I may be missing the point in why this would be useful.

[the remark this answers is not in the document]
kamescg: @kdenhartog I think by stateless what's being expressed is no database in the authentication layer. Traditionally with a signed JWT you can get away with authenticating signatures without pinging a database to validate a session. If the signature is valid and the timestamp hasn't expired it's good to go.

EIPS/eip-4361.md
kamescg: Great naming suggestion!

Also, in regards to is the "signature valid" I think that's up to the implementer of the authentication service, no? Can't really be enforced in the message schema.

EIPS/eip-4361.md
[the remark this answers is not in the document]
kdenhartog: >Traditionally with a signed JWT you can get away with authenticating signatures without pinging a database to validate a session. If the signature is valid and the timestamp hasn't expired it's good to go.

Yes, but that's because the resource server has an inherent binding to the authorization server because it's directly retrieving the token from the OP in the authorization code flow. So in other words the application itself can be managed in a stateless manner because the OP can be trusted to not maliciously replay the token due to the trust between the two servers in that architecture. This means the nonce is unnecessary to be used for replay protection and it's only being used for the second purpose which is to associate a Client session with an ID Token. However, because the authorization code is also being used to associate the client session with the ID Token, the nonce no longer is necessary in this flow.

The issue here is that the resource server can't trust the authorization server in this case and so the architecture of SIWE is more in line with the implicit code flow which does require the usage of a nonce to be sent in the request. This is in section [3.2.2.1 of the OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest).

Building on this the spec says in section [3.2.2.11 of the OIDC Core spec](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation) it states:

> if a nonce value was sent in the Authentication Request, a nonce Claim MUST be present and its value checked to verify that it is the same value as the one that was sent in the Authentication Request. The Client SHOULD check the nonce value for replay attacks. The precise method for detecting replay attacks is Client specific. 

So in other words, because there's not an inherent trust between the RP and the wallet a client MUST be used. If an app wanted to go stateless with the current design they would most likely need to utilize the OAuth server that Spruce also has committed to so that the RP could establish a trust and defer authn and authz events to the OP.

EIPS/eip-4361.md
wyc: This is great clarification, thank you!

EIPS/eip-4361.md
wyc: @SamWilsn Thanks for your comments! We should encourage wallet vendors to display a separate interface once they have identified a valid SIWE message, but it's true they'd have to interpret the whole thing in the current form. Two possible improvements here, not mutually exclusive:

1. We could alternatively require the message to begin with `"EIP-4361. "` to allow quick prefix determination for parsers prior to attempting to parse any of the grammar, but I think the message should be short enough as not to pose a big risk. This would potentially set a precedence on how future signing requests will differentiate themselves, inadvertently setting a precedent for future EIP-191 extensions. It could also slightly degrade the user experience, but arguably also improve it by giving users a term that is likely to turn up with useful results in a search engine.

2. We can add a SHOULD clause to the spec that cannot be easily tested like: "Wallet vendors SHOULD display a confirmatory indicator if a valid EIP-4361 message is detected, and wallets MAY take reasonable steps to proactively identify similar but non-conformant messages to alert the user accordingly."

Thoughts on these?

EIPS/eip-4361.md
wyc: Great suggestion. I considered this, but some implementations including OpenSea may want to include a URL to their terms of service or similar, which would require `:` to construct. Is this exception reasonable enough to you to warrant its inclusion?

EIPS/eip-4361.md
wyc: Thanks for your suggestion, I like it a lot semantically. However, I think I'd prefer to stick to `expiration-time` to keep similarity to JWTs (https://datatracker.ietf.org/doc/html/rfc7519) which would make it more ergonomic to developers already familiar with the terminology there. A lot of the spec was inspired by fields set as precedence in JWTs. Does that make sense?

EIPS/eip-4361.md
wyc: Yes, it would be required for EIP-1271 based signing. Furthermore, by principle of least privilege, we wanted to strongly push implementers to lock the session to a specific `chain-id` or think carefully about its interpretation otherwise. I think keeping a required `chain-id` reduces one conditional check (for smart contract wallet or EOA) and also achieves the latter goal. Does that make sense?

EIPS/eip-4361.md
wyc: Agreed with your comment on naming to capture intents, especially as they are interpreted for TLS. However, we are setting up an application level session and not necessarily a comms channel. So, I think we should keep `dnsauthority`, which includes the host and port, to provide an opportunity for more granular binding. I'm thinking about the case where there may be multiple SIWE services exposed on the same host, but separated by port. This would allow the signing to be bound to an exact service and serve as an additional layer of protection on top of the nonce. Does that reasoning make sense?

EIPS/eip-4361.md
wyc: Also, note on the separate prefix--I don't think most wallets would be able to display the message cohesively if we removed the signing prefix `\x19Ethereum Signed Message:\n<length of message>` for wallets that don't make changes explicitly to support this spec. The goal is that applications should be able to create conforming implementations without forcing wallet vendors to do anything without substantially degrading the user experience. This is an important consideration for adoption, and at the same time your observation is valid.

EIPS/eip-4361.md
kdenhartog: The flip side of this is that it requires me to re-authenticate each time I change networks in my wallet with the current UX flows. E.g. say I want to use sushiswap on mainnet and then decide to use a L2 solution I'd have to re-authenticate after changing networks which doesn't make sense to me. Wouldn't this lead to greater consent fatigue which is already pretty high as is? Furthermore, is it likely that we'd have the same account on different networks owned by different people?

wyc: @kdenhartog An implementer is free to interpret a `chain-id` of `0` as across all networks for example, but that's their decision to make. From today's application, it does appear that networks aiming for EVM-compat indeed re-use the same account. The privacy and security implications of this are worth noting but out of scope for this spec.

EIPS/eip-4361.md
SamWilsn: I'm not a fan of (1) since it adds extra gibberish to the message that will make users even more likely to just blindly sign.

Is the `wants you to sign in with your Ethereum account` portion fixed and required? If so, that's probably sufficient for wallets to scan for.

I do think (2) is a good security recommendation. Maybe something as simple as: "Wallet software SHOULD display a high friction confirmatory indicator if the string `wants you to sign in with your Ethereum account` appears anywhere in an EIP-191 message signing request unless the message is a valid according to EIP-4361."

wyc: Thanks! I will make these changes, okay to add you as a contributor for your useful comments throughout?

EIPS/eip-4361.md
kdenhartog: It sounds like it's being made in scope though because the chain-id is being required so that's a bit confusing to me. Requiring it's usage, but not stating why it's necessary nor placing additional normative requirements on it seems a bit odd to me and is opening the door for potential interoperability issues based on wallets and dApps making different assumptions.

EIPS/eip-4361.md
SamWilsn: I disagree with your use of:

> `domain` is the [RFC 4501](https://datatracker.ietf.org/doc/html/rfc4501) `dnsauthority` that is requesting the signing.

In RFC 4501, `dnsauthority` is _who_ you are contacting to look up a domain, and not a domain itself.

If your intent is to capture the "part of the URL" which represents the server side, I'd recommend switching to `authority` from [RFC 3986 (Section 3)](https://datatracker.ietf.org/doc/html/rfc3986#section-3) or `host` from [WhatWG's URL Specification](https://url.spec.whatwg.org/#hosts-(domains-and-ip-addresses)).

EIPS/eip-4361.md
SamWilsn: They can percent-encode it right?

EIPS/eip-4361.md
wyc: Great suggestion, I think we'll go with one of those. Thanks for sharing your knowledge, I didn't think to look at WhatWG for a similar thing.

EIPS/eip-4361.md
wyc: Would you onboard with adding the recommendation of `chain-id` to be `0` if it's a "don't care"?

kdenhartog: Yeah I think normatively defining that would be a good way to prevent interop concerns. Strong +1 from me. Would chain-id 0 also need to be registered with EIP 115?

wyc: Hm, I don’t think it’s normatively defined in EIP-155, could be better to add the character `*` as a valid atom for this to avoid any future woes, okay with this?

wyc: It also begs the question of smart contract wallets being able to sign cross chain…in earlier iterations we had a list of `chain-id` but found that to be overkill…

kdenhartog: I'm good with a `*` character, but I think it may encounter some weird logic in the way EIP-115 was originally intended, but I'm not certain there's overlap at this point. Particularly this clause in EIP-115 makes me think it could break some logic because it's not a number type:

> If block.number >= FORK_BLKNUM and v = CHAIN_ID * 2 + 35 or v = CHAIN_ID * 2 + 36


>It also begs the question of smart contract wallets being able to sign cross chain…in earlier iterations we had a list of chain-id but found that to be overkill…

Yeah that seems like it would be a really useful feature, but I'd assume based on your initial response that including this may be more scope creep than you'll want to take on at this point. I'm good with leaving the cross contract signing off for now in order to simplify, but there's definitely tradeoffs that will come with this in terms of limited capabilities here.

EIPS/eip-4361.md
SamWilsn: Sure, that's fine!

EIPS/eip-4361.md
wyc: Removed in https://github.com/ethereum/EIPs/pull/4361/commits/0975d6c0d5a610dd8209d9911c715702c8600c62 , thanks for your advice! Feel free to reopen if there's anything else.

EIPS/eip-4361.md
wyc: Interpreting your emoji as acceptance, feel free to reopen if you'd like to talk about this more. Thanks!

EIPS/eip-4361.md
kamescg: > Would you onboard with adding the recommendation of `chain-id` to be `0` if it's a "don't care"?

Doesn't that imply `chain-id` is optional and the presumed default is 0? So no need to redeclare `0` with `0`.

EIPS/eip-4361.md
kamescg: I also found the `wants you to sign in with your Ethereum account` to be an "interesting" part of the specification.

How does this work with multiple languages in regards to the specification? Ideally the specification has no dependency on the English language and can be used in a universal setting. Can this be replaced with translated equivalents, and if so will wallet vendors be expected to parse for those strings as well? 

This is kind of secondary, but I found it just a little awkward to parse the current message template. Here is just a quick, naive approach. 

https://gist.github.com/kamescg/b0f2875881ff15ec976f6976cedca885

kamescg: > Maybe something as simple as: "Wallet software SHOULD display a high friction confirmatory indicator if the string wants you to sign in with your Ethereum account appears anywhere in an EIP-191 message signing request unless the message is a valid according to EIP-4361."

The best solution definitely feels like parsing the `eth_signMessage` to detect message `types` is the most flexible long-term solution. Also, it's hard to believe this will be the last EIP to use string parsing to detect the message `intention`.

Because of this I feel like detecting such a "generic message" like `wants you to sign in with your Ethereum account` is not the most scalable approach to dynamically detecting all incoming `eth_signMessage` potential future types.

In my opinion it would probably be best to drop the `wants you to sign in with your Ethereum account` because that's implied in the singing 😉  and replace it with a `type` and `language` keys that can be parsed by incoming wallets and also be used by future standards who need to have a `StructuredDataIsh` interface while still using the `eth_signMessage` call.

Easy to parse and identify.

And the intention message maybe has to be replaced with `Domain` and `Address` keys?

Also just curious as to why the `statement` param isn't a key/value pair like everything else? Is a bit of an odd data structure when the message has to be dynamically parsed and found by searching for the two surrounding `\n` on empty lines.

**Proposed Changes**
```
type: EIP-4361
language: en
Domain: http://localhost:3000 
Address: 0x000000000000000000000000000000000000dEaD
Statement: http://opensea.io/tos
URI: http://localhost:3000/authenticate
Version: 1
Chain ID: 1
Nonce: 12345678
Issued At: 2021-11-20T10:10:10Z
Not Before: 2021-11-19T10:10:10Z
Expiration Time: 2021-11-21T10:10:10Z
Request ID: 2
Resources:
 https://ipfs.com/hash
```

EIPS/eip-4361.md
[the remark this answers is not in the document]
kamescg: Sorry... I'm a little confused. What resource server? At the end of the day of this is minimally self-signed JWT-like token that's rooted to external sources via a verifiable address.

What are we getting from adopting that OIDC spec when we're removing the IDP and RP from the equation?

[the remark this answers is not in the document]
kamescg: Hmmm... yeah I just don't see why that matters with the self-signing of the authentication token. The trust is from signature verification and `issuedAt` key, which essentially acts as `nonce` so why add a field to semi-adopt the ODIC spec?

MicahZoltu:
Reminder to everyone: PR comments should mostly be limitd to feedback required to get the EIP merged as a draft.  Discussions about the specification itself should go in the discussions-to link found in the EIP header.

EIPS/eip-4361.md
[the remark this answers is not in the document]
kdenhartog: It's not about trying to adopt the OIDC spec. I only used that as an example in order to showcase this is standard for protocol designs like this and to highlight that trust boundaries between the different software systems matter here. In this particular case due to the nature in which the trust works, the appropriate method to be used is to require the nonce in order to allow the protocol to be used in a broader set of use cases without requiring dApp developers to have to assess their threat model each time and decide if it's necessary in order to support 1 architecture which I'm still not convinced makes legitimate sense.

EIPS/eip-4361.md
wyc: @kamescg I propose we keep it due to the reasoning here: https://github.com/ethereum/EIPs/pull/4361#discussion_r755524905

We can continue this discussion in the ETH Magicians forum if you feel strongly about it.

EIPS/eip-4361.md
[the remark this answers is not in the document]
wyc: Thanks for your comments, if you continue to feel strongly about this please continue the discussion in ETH Magicians.

EIPS/eip-4361.md
wyc: Hi Kames, the example you present is indeed more normalized, but I am concerned about readability to end users. We gathered feedback from over 30+ teams in the space over the past several weeks, and they were okay with the proposed format. We welcome your independent research to see if they prefer the format you propose for their users over the current one if you wish to move it forward.

Additionally, I would consider solving for a "scalable approach to dynamically detecting all incoming eth_signMessage potential future types" is out of scope for this spec.

Thanks again for your comments, and I look forward to continuing the conversation in ETH Magicians if you continue to feel strongly about this.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,312 @@
...
+
+This work is sponsored by the Ethereum Foundation and Ethereum Name Service (ENS). It is being developed in the open through a series of recorded community calls and public repositories, and its development is informed by over twenty user interviews with a focus on currently-in-production uses, related prior EIPs, and fits within product roadmaps.
+
+## Specification
+Sign-In with Ethereum works as follows:
+
+1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the [EIP-191](https://eips.ethereum.org/EIPS/eip-191) signature scheme (string prefixed with `\x19Ethereum Signed Message:\n<length of message>`). The `message` MUST incorporate an Ethereum `address`, `version` of the message, `uri` for scoping, `nonce` acceptable to the server, and `issued-at` timestamp.
+2. The signature is then presented to the server, which checks the signature's validity and message content.
+3. Additional fields, including `expiration-time`, `not-before`, `request-id`, `chain-id`, and `resources` may be incorporated as part of authentication for the session.
```
epappas, line 29: Could we consider the flexibility to support additional AuthN claims, for when it makes sense for the application? I can see some value in being able to switch between "wallet profiles" when it makes sense.

For example in a typical B2B App I'd also expect "department", or in another example, like a game, I could own multiple (in-game) characters under my wallet, would I be able to set an app-specific custom claim eg "character_name"?

EIPS/eip-4361.md
wyc: Updated in https://github.com/ethereum/EIPs/pull/4361/commits/1415bfb31f5647b428fd06e9c0f74f78d589bebd with credits too.

EIPS/eip-4361.md
wyc: I will keep what's there for now, due to our user research. If you feel strongly about this, please continue the conversation in ETH Magicians. Thanks again.

EIPS/eip-4361.md
wyc: Hi epappas, please see the Resources section to allow this degree of flexibility. We just want to achieve ETH-based authn with this spec, not anything past that, though we are trying to keep that open ended for other implementers. Please feel free to continue the conversation in ETH Magicians if you feel strongly about it. Thanks.

wyc:
Hi @MicahZoltu, I believe we are ready for review and merge here. Thank you for your patience.

MicahZoltu:
I only handle Core EIPs.  @axic or @lightclient are the ones who handle ERCs.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,313 @@
...
+
+Already, many services support workflows to authenticate Ethereum accounts using message signing, such as to establish a cookie-based web session which can manage privileged metadata about the authenticating address. This is an opportunity to standardize the sign-in workflow and improve interoperability across existing services, while also providing wallet vendors a reliable method to identify signing requests as Sign-In with Ethereum requests for improved UX.
+
+This work is sponsored by the Ethereum Foundation and Ethereum Name Service (ENS). It is being developed in the open through a series of recorded community calls and public repositories, and its development is informed by over twenty user interviews with a focus on currently-in-production uses, related prior EIPs, and fits within product roadmaps.
```
lightclient, line 22: Usually we don't include this kind of information in an EIP. I recommend removing it and instead posting it in the `discussions-to` thread.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Specification
+Sign-In with Ethereum works as follows:
+
+1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the [EIP-191](https://eips.ethereum.org/EIPS/eip-191) signature scheme (string prefixed with `\x19Ethereum Signed Message:\n<length of message>`). The `message` MUST incorporate an Ethereum `address`, `version` of the message, `uri` for scoping, `nonce` acceptable to the server, and `issued-at` timestamp.
```
lightclient, line 27:
```suggestion
1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the [EIP-191](./eip-191.md) signature scheme (string prefixed with `\x19Ethereum Signed Message:\n<length of message>`). The `message` MUST incorporate an Ethereum `address`, `version` of the message, `uri` for scoping, `nonce` acceptable to the server, and `issued-at` timestamp.
```
Please use relative links to other EIPs.

lightclient: (there are other instances in the EIP that would need to be updated)

EIPS/eip-4361.md
```diff
@@ -0,0 +1,313 @@
...
+- Wallet implementers MAY construct a custom Sign-In With Ethereum user interface by parsing the ABNF terms into data elements for use in the interface. The display rules above still apply to custom interfaces.
+
+#### Supporting internationalization (i18n)
+- After successfully parsing the message into ABNF terms, translation may happen at the UX level per human language.
+
+
+## Rationale
+
+### Requirements
+From the [RFP](https://notes.ethereum.org/@djrtwo/sign-in-with-ethereum-RFP):
```
lightclient, line 199: The rationale section should explain the reasoning behind certain decisions in the EIP. Linking to the RFP is not appropriate here.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,313 @@
...
+    - EIP-191 is simple to implement using a pre-set prefix prior to signing, while EIP-712 is more complex to implement requiring the further implementations of a bespoke Solidity-inspired type system, RLP-based encoding format, and custom keccak-based hashing scheme. **(2)**
+    - EIP-191 produces more human-readable messages, while EIP-712 creates signing outputs for machine consumption, with most wallets not displaying the payload to be signed in a manner friendly to humans. **(1)**![](https://i.imgur.com/yvhz2mD.png)
+
+    - EIP-712 has the advantage of on-chain representation and on-chain verifiability, such as for their use in metatransactions, but this feature is not relevant for the specification's scope. **(2)**
+- Why not use JWTs? Wallets don't support JWTs. The keccak hash function is [not assigned by IANA](https://www.iana.org/assignments/jose/jose.xhtml) for use as a JOSE algorithm. **(2, 3)**
+- Why not use YAML or YAML with exceptions? YAML is loose compared to ABNF, which can readily express character set limiting, required ordering, and strict whitespacing. **(2, 3)**
+
+### Discussions and Acknowledgements
```
lightclient, line 227:
```suggestion
### Discussions and Acknowledgements
```
This type of information should be tracked in the `discussions-to` thread, not in the EIP itself.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,313 @@
...
+- Interpretation of the URIs in the `resources` term as claims or other resources.
+- The specific mechanisms to ensure domain-binding.
+- The specific mechanisms to generate nonces and evaluation of their appropriateness.
+- Protocols for use without TLS connections.
+
+## Backwards Compatibility
+- Most wallet implementations already support EIP-191, so this is used as a base pattern with additional features.
+- Requirements were gathered from existing implementations of similar sign-in workflows, including statements to allow the user to accept a Terms of Service, nonces for replay protection, and inclusion of the Ethereum address itself in the message.
+
+## Forwards Compatibility
```
lightclient, line 251:
```suggestion
## Forwards Compatibility
```
This is not a valid EIP section. This sort of information should be incorporated into the rationale.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,313 @@
...
+#### Session invalidation
+There are several cases where an implementer SHOULD check for state changes as they relate to sessions.
+
+- If an EIP-1271 implementation or dependent data changes the signature computation, the server SHOULD invalidate sessions appropriately.
+- If any resources specified in `resources` change, the server SHOULD invalidate sessions appropriately. However, the interpretation of `resources` is out of scope of this specification.
+
+#### Maximum lengths for ABNF terms
+- While this specification does not contain normative requirements around maximum string lengths, implementers SHOULD choose maximum lengths for terms that strike a balance across the prevention of denial of service attacks, support for arbitrary use cases, and user readability.
+
+## Additional Resources
```
lightclient, line 305: This is not a valid EIP section. Please remove.

lightclient (changes requested):

EIPS/eip-4361.md
SamWilsn: Screenshot 2021-12-06 at 16-19-50 Meme Generator - Imgflip

EIPS/eip-4361.md
wyc: Addressed in https://github.com/ethereum/EIPs/pull/4361/commits/e59daacf95ecf9747b47921f7ce7b865eb139f75

EIPS/eip-4361.md
wyc: Addressed in https://github.com/ethereum/EIPs/pull/4361/commits/e59daacf95ecf9747b47921f7ce7b865eb139f75

EIPS/eip-4361.md
wyc: Addressed in https://github.com/ethereum/EIPs/pull/4361/commits/e59daacf95ecf9747b47921f7ce7b865eb139f75

EIPS/eip-4361.md
wyc: Addressed in https://github.com/ethereum/EIPs/pull/4361/commits/e59daacf95ecf9747b47921f7ce7b865eb139f75

EIPS/eip-4361.md
wyc: Addressed in https://github.com/ethereum/EIPs/pull/4361/commits/e59daacf95ecf9747b47921f7ce7b865eb139f75

wyc:
@lightclient thank you for your keen eye, in my latest commit I hope to have addressed all the points you brought up. Ready for another review when you're ready. Please let me know if you'd prefer that I squash everything into a single commit.

wyc:
Hi @lightclient, following up on this PR. Please let me know if there are any further necessary changes. I appreciate your review to date!

lightclient (approved):
Thanks for your patience and for addressing that feedback. LGTM.

EIPS/eip-4361.md
```diff
@@ -0,0 +1,293 @@
+---
```
ewinAT, line 1: 0x14D6485F5E3B109230563b1d443fD6B16539955c

ewinAT (commented):
want to register my ETH20 address

amzu4147:
spruceid:eip-4361

PR #4363, opened 2021-10-12
Transaction index OpCode

decanus:
## Abstract

This EIP adds an opcode that returns the current transactions.

## Motivation

Currently users are completely at the mercy of those who order transactions. This means that malicious actors can exploit the ordering of transactions at the users expense. A popular example of this is sandwhich attacks on AMMs, wherein a transaction is put in between 2 others that front-run it.

## Specification

Adds a new opcode `TXINDEX` at 0x49, which uses 0 stack arguments. It pushes the current transaction index onto the stack. Transaction Index is a 256-bit value. The operation costs G_base to execute.

## Reference Implementation

 - A dummy implementation in [go-ethereum@feature/tx-index](https://github.com/dialecticch/go-ethereum/tree/feature/tx-index).
 - Usage of the opcode in [UniswapV3](https://github.com/fxfactorial/uniswap-v3-core/commit/2e605d8f3fe4ccc8c8c0c78eb71674c220348b65).

## Copyright


eip-4363.md
```diff
@@ -0,0 +1,30 @@
...
+created: 2021-10-12
+---
+
+## Abstract
+
+This EIP adds an opcode that returns the current transactions.
+
+## Motivation
+
+Currently users are completely at the mercy of those who order transactions. This means that malicious actors can exploit the ordering of transactions at the users expense. A popular example of this is sandwhich attacks on AMMs, wherein a transaction is put in between 2 others that front-run it.
```
axic, line 17: I think it would make sense extending this to get an actual example how this helps. How is the user supposed to influence at which index this should be included? Only by paying a high enough premium so it gets included at the right index, or do you plan some kind of annotation for pools, or specific pools which prefer such transactions?

OisinKyne: > but I don't see how the end result is any different than priority gas auctions where bots bail out when they lose and cancel their pending transaction

I'd argue that one is a P2P layer solution, the other is an EVM layer solution for ordering, and I'd imagine an EVM layer approach is 100% successful versus late cancelling self-txs. 

And this EIP wouldn't prevent a centralised service (like flashbots) from pruning out reverting transactions to save them hitting the chain.

snario:
@decanus Given what @ncitron said, and that this makes EVM parallelization more complicated, would an alternate proposal to add something like an `IS_ZERO_IDX` opcode which returns `true` if the transaction index is 0 provide the same value?

decanus:
@snario I think that adding an opcode just for `IS_ZERO_IDX` wipes away a ton of other use-cases, for example doing certain things with a transaction depending on where it is in the block. Additionally EVM parallelization right now has larger issues to solve than the transaction index, so I think we can solve that when we get there.

eip-4363.md
```diff
@@ -0,0 +1,30 @@
+---
+eip: 4363
+title: Transaction index OpCode
+author: Edgar (@fxfactorial), Dean (@decanus)
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-10-12
+---
```
lightclient, line 9: It appears you're missing the `description` and `discussions-to` fields of the preamble.

eip-4363.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Specification
+
+Adds a new opcode `TXINDEX` at 0x49, which uses 0 stack arguments. It pushes the current transaction index onto the stack. Transaction Index is a 256-bit value. The operation costs G_base to execute.
+
+## Reference Implementation
+
+ - A dummy implementation in [go-ethereum@feature/tx-index](https://github.com/dialecticch/go-ethereum/tree/feature/tx-index).
+ - Usage of the opcode in [UniswapV3](https://github.com/fxfactorial/uniswap-v3-core/commit/2e605d8f3fe4ccc8c8c0c78eb71674c220348b65).
+
```
lightclient, line 27: Also missing Rationale + Security Considerations.

eip-4363.md
```diff
@@ -0,0 +1,30 @@
...
+author: Edgar (@fxfactorial), Dean (@decanus)
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-10-12
+---
+
+## Abstract
+
+This EIP adds an opcode that returns the current transactions.
```
lightclient, line 13:
```suggestion
Adds an opcode that returns the current transaction's index within the block.
```

CryptoGodfatherVA (approved):

Magicking:
Knowing the transaction counter doesn't prevent CPUs from scaling, not sure how passing a new read-only input would deepen the EVM parallelization possibilities.

Arguably, this EIP could incite to send (pay) transactions directly to miners instead of transacting through the public memory pool to ensure a certain place in the block.

Regarding other usages, here is an example of a lottery-based contract that makes you win if you have the highest tx index https://gist.github.com/Magicking/eaaaae36dcb40de19951df3802bb46d0 , miners can always win by putting the last executed transaction, in a way it widens and complexify MEV.

sambacha:
[direct link to the relevant implementation portion for the opcode](https://github.com/fxfactorial/uniswap-v3-core/commit/2e605d8f3fe4ccc8c8c0c78eb71674c220348b65#diff-060349bf35ee4391d7b6288dca701de30e4852c0385e80b5d1e56a7f745eaca7R505-R506)


Question: couldn't a proportion of realized *M.E.V* be negated by different contract implementations?

PR #4372, opened 2021-10-15
Update offset and calculations

timbeiko:
On ACD 124, we agreed to a 10,700,000 offset rather than 10,500,000.

### DIFF
--- a/EIPS/eip-4345.md
+++ b/EIPS/eip-4345.md
@@ -1,7 +1,7 @@
 ---
 eip: 4345
-title: Difficulty Bomb Delay to May 2022
-description: Delays the difficulty bomb to be noticeable in May 2022.
+title: Difficulty Bomb Delay to June 2022
+description: Delays the difficulty bomb to be noticeable in June 2022.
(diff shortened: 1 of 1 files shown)


EIPS/eip-4345.md
```diff
@@ -1,41 +1,41 @@
 ---
 eip: 4345
-title: Difficulty Bomb Delay to May 2022
-description: Delays the difficulty bomb to be noticeable in May 2022.
+title: Difficulty Bomb Delay to June 2022
+description: Delays the difficulty bomb to be noticeable in June 2022.
 author: Tim Beiko (@timbeiko), James Hancock (@MadeOfTin)
 discussions-to: https://ethereum-magicians.org/t/eip-4345-difficulty-bomb-delay-to-may-2022/7209
-status: Draft
+status: Review
```
axic, line 7: Actually this needs a `review-period-end` field.

MicahZoltu: Not for Review, only for Last Call.

axic: Yeah we realised after it, I thought it was Last Call as that's the status these go into when they are scheduled for a fork.

PR #4390, opened 2021-10-24
Update EIP Editors that get mentioned.

MicahZoltu:
Removed myself from everything other than Core EIPs.  Also removed all editors who haven't done any actual editorial work in the last ~year.

I would *love* it if more people were actively participating in the editorial process, but at the moment it is really just the three of us and I think the mentions should reflect that.


alita-moore:
Alert fatigue is a thing

lightclient:
@alita-moore bot is failing here because not an EIP file.

alita-moore:
Yep that's a planned feature. Think it's a higher priority than other features?

PR #4367, opened 2021-10-14
move eip-2464 from "stagnant" to "review"

poojaranjan:

PR #4366, opened 2021-10-14
Move eip-2364 from "stagnant" to "review"

poojaranjan:

PR #4395, opened 2021-10-28
Updates to EIPs 2464 and 2364.

timbeiko:
Combines the updates from https://github.com/ethereum/EIPs/pull/4367 and https://github.com/ethereum/EIPs/pull/4366, and addresses most comments from @MicahZoltu in a way which hopefully makes him happy, but recognizes the fact that these EIPs are already "ancient". Addressed all comments which were related to formatting/sections, but links are still present, and the specification has been left as is. 

Hopefully we can get a soft +1 on these before requiring Péter to review & approve.

### DIFF
--- a/EIPS/eip-2364.md
+++ b/EIPS/eip-2364.md
@@ -1,9 +1,10 @@
 ---
 eip: 2364
 title: "eth/64: forkid-extended protocol handshake"
-author: Péter Szilágyi <peterke@gmail.com>
+description: Introduces validation of the `forkid` when handshaking with peers.
+author: Péter Szilágyi <peterke@gmail.com>, Péter Szilágyi (@karalabe)
 discussions-to: https://github.com/ethereum/EIPs/issues/2365
-status: Stagnant
+status: Review
 type: Standards Track
 category: Networking
 created: 2019-11-08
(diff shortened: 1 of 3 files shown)


timbeiko:
Should have addressed your comments, @MicahZoltu! Would like @karalabe to approve before we merge given these are his EIPs and the modifications are a bit more extensive than I'd planned.

karalabe (approved):

axic:
There's also the `Backwards Compatibility` section as a requirement according to EIP-1.

MicahZoltu:
Both appear to have a backward compatibility section.

MicahZoltu (approved):

MicahZoltu:
Apparently I don't have authority to approve this anymore!  See #4418 to fix that, in the meantime @lightclient or @axic approving should result in auto-merge.

axic:
> Both appear to have a backward compatibility section.

Hmm, correct. The diff view was expanding weirdly yesterday.

EIPS/eip-2364.md
```diff
@@ -40,15 +41,17 @@ Whenever two peers connect using the `eth/64` protocol, the updated `Status` mes
 
 ## Rationale
 
-##### EIP-2124 mentions advertising the `forkid` in the discovery protocol too. How does that compare to advertising in the `eth` protocol? Why is the redundancy needed?
+The specification is tiny since most parts are already specified in EIP-2124. `eth/63` is not specified as an EIP, but is maintained [here](https://github.com/ethereum/devp2p/blob/master/caps/eth.md).
```
axic, line 44: Not sure why this needs to be here? Is this the way to make @MicahZoltu allow an external URL? :)

axic: And can this URL be a tagged one, and not `master`?

MicahZoltu: Ah, you are right, that should be removed.  It is not necessary for the specification.

PR #4396, opened 2021-10-29
Time-Aware Base Fee Calculation

adietrichs:
This EIP proposes accounting for time between blocks in the base fee calculation to target a stable throughput by time, instead of by block. Aiming to minimize changes to the calculation, it only introduces a variable block gas target proportional to the block time. The EIP can, in principle, be applied to either a Proof-of-Work or a Proof-of-Stake chain, however the security implications for the Proof-of-Work case remain unexplored.


lightclient (approved):

alita-moore (commented):
testing bot

alita-moore (commented):
testing bot

alita-moore:
Oh I see this is an edge case. Shall I treat this the same as eip-1?

MicahZoltu:
@alita-moore Why is this an edge case?  What happened that caused this to not merge?

alita-moore:
> @alita-moore Why is this an edge case?  What happened that caused this to not merge?

Ah looks like I got my mentions mixed up. I was thinking about the bot not parsing eip-###.md

In this case it's most likely because of the fact non-eip are not supported (this touched assets folder). Do you recommend I prioritize this above all else?

MicahZoltu:
Ah, I thought you already added support for `assets/eip-####.md`.  If that is still in the backlog then this failing makes sense.  It would be nice if the error message was a bit more clear, maybe we should *also* get an issue into the backlog for that.

PR #4399, opened 2021-10-30
EIP-4399: Supplant DIFFICULTY opcode with RANDOM

mkalinin:
A proposal to supplant the `DIFFICULTY` opcode with the `RANDOM` opcode during the upgrade to PoS (a.k.a. the Merge):
- enshrine the output of the beacon chain randomness beacon into the block header structure by supplanting the value of the deprecated `mixHash` field
- make the instruction with the number of the `DIFFICULTY` opcode return the value of the `mixHash` field
- rename the `mixHash` field and the `DIFFICULTY` opcode to `random` field and `RANDOM` opcode respectively

### DIFF
--- a/EIPS/eip-4399.md
+++ b/EIPS/eip-4399.md
@@ -2,7 +2,7 @@
 eip: 4399
 title: Supplant DIFFICULTY opcode with RANDOM
 author: Mikhail Kalinin (@mkalinin), Danny Ryan (@djrtwo)
-discussions-to:
+discussions-to: https://ethereum-magicians.org/t/eip-4399-supplant-difficulty-opcode-with-random/7368
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-4399.md
+++ b/EIPS/eip-4399.md
@@ -1,5 +1,5 @@
 ---
-eip: 4398
+eip: 4399
 title: Supplant DIFFICULTY opcode with RANDOM
 author: Mikhail Kalinin (@mkalinin), Danny Ryan (@djrtwo)
 discussions-to:
--- a/EIPS/eip-4398.md
+++ b/EIPS/eip-4398.md
@@ -1,0 +1,110 @@
+---
+eip: 4398
+title: Supplant DIFFICULTY opcode with RANDOM
+author: Mikhail Kalinin (@mkalinin), Danny Ryan (@djrtwo)
+discussions-to:
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-10-30
+requires: 3675
+---
+
+## Abstract
+
+This EIP supplants the semantics of the return value of existing `DIFFICULTY` opcode and renames the opcode to `RANDOM`.
+
+The return value of the `DIFFICULTY` opcode after this change is the output of the randomness beacon provided by the beacon chain.
+
+
+## Motivation
+
+Applications may benefit from using the randomness accumulated by the beacon chain. Thus, randomness outputs produced by the beacon chain should be accessible in the EVM.
+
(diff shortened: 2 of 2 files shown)


mkalinin:
idk what's happened but the PR opened with number `4399` instead of expected `4398`. Did a rename

EIPS/eip-4399.md
```diff
@@ -0,0 +1,110 @@
...
+### Renaming the field and the opcode
+
+The renaming should be done to make the field and the opcode names semantically sound.
+
+
+## Backwards Compatibility
+
+This EIP doesn't introduce any backward incompatible changes to the base protocol. Though, the changes it propose might be backward incompatible for the following categories of applications:
+* Applications that uses the value returned by the `DIFFICULTY` opcode as the PoW `difficulty` parameter
+* Applications which logic depending on the `DIFFICULTY` opcode return value to be relatively small number
```
MicahZoltu, line 85:
```suggestion
* Applications with logic that depends on the `DIFFICULTY` opcode returning a relatively small number
 ```

File: EIPS/eip-4399.md [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+
+### Renaming the field and the opcode
+
+The renaming should be done to make the field and the opcode names semantically sound.
+
+
+## Backwards Compatibility
+
+This EIP doesn't introduce any backward incompatible changes to the base protocol. Though, the changes it propose might be backward incompatible for the following categories of applications:
+* Applications that uses the value returned by the `DIFFICULTY` opcode as the PoW `difficulty` parameter
```
MicahZoltu, line 84:
```suggestion
* Applications that use the value returned by the `DIFFICULTY` opcode as the PoW `difficulty` parameter
```

EIPS/eip-4399.md
```diff
@@ -0,0 +1,110 @@
...
+* Applications which logic depending on the `DIFFICULTY` opcode return value to be relatively small number
+
+The first category is already affected by switching the consensus mechanism to PoS and no additional breaking changes are introduced by this specification.
+
+The second category comprises the applications that use the return value of the `DIFFICULTY` opcode in operations that might cause either overflow or underflow errors. Applications where the result of such errors may turn into security violations are not considered by this EIP.
```
MicahZoltu, line 89:
```suggestion
The second category is comprised of applications that use the return value of the `DIFFICULTY` opcode in operations that might cause either overflow or underflow errors.  While it is theoretically possible to author an application where a change in the range of possible values this OPCODE may return could lead to a security vulnerability, we believe the chances of that are very unlikely.
```

MicahZoltu (approved):
I left some feedback, but no blockers for merging as a draft.

EIPS/eip-4399.md
```diff
@@ -0,0 +1,110 @@
...
+## Reference Implementation
+
+*TBD*
+
+
+## Security Considerations
+
+The `RANDOM` opcode in the PoS network should be considered as a source of randomness output of a higher strength than the output provided by either the `BLOCKHASH` or the `DIFFICULTY` opcode in the PoW network.
+
+However, the randomness output returned by the `RANDOM` opcode is revealed within the parent block.
```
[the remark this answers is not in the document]
djrtwo: yes

and an early randao reveal slashing is on the table
was in fact in the phase 0 specs early on!

but it's a relatively low priority attack point so I don't expect we see a proposal for inclusion in the next 18 months, tbh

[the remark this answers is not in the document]
mkalinin: My try to be more clear on it is: "is revealed to the network within the parent block" -- https://github.com/ethereum/EIPs/pull/4407. I slightly prefer us do not describe an attack scenario with the collusion of multiple proposers in this document. The reveal to the network should create enough level of awareness for smart contract developers. But I don't have a strong opinion on that and we could and it might worth specifying from some perspective.

Early randao reveal slashing doesn't protect revealing the randao mix, does it? If it does then revealing after grandparent block is publish can't be punished unless the proposer punishes itself.

Lucid36 (approved):

PR #4400, opened 2021-10-30
Optional ERC721Consumable Extension

Daniel-K-Ivanov:
Proposes the addition of a Consumer role for ERC721 tokens. The purpose of the role is to enable multiple layers of integration on top of already existing dApps operating/issuing ERC721 tokens.


fulldecent:
Please set the discussion URL on your EIP.

---

"Using" an NFT is different in every context. For Su Squares, the owner can personalize. For Number Board, the owner can set a message. For CryptoKitties, the kitties can breed.

So being a consumer has many different meanings.

When standardizing something it is important to have a producer AND a consumer. I don't see who wants to consume this.

Daniel-K-Ivanov:
> Please set the discussion URL on your EIP.
> 
> "Using" an NFT is different in every context. For Su Squares, the owner can personalize. For Number Board, the owner can set a message. For CryptoKitties, the kitties can breed.
> 
> So being a consumer has many different meanings.
> 
> When standardizing something it is important to have a producer AND a consumer. I don't see who wants to consume this.

@fulldecent thank you for the heads-up. I will add a link to an official discussions URL once my post is approved by the Ethereum Magicians 🙏 

Appreciate the comment as well.
You are right that "using" an NFT is different in every context. If the standard explicitly states what the meaning of "use" is, it is rendered obsolete. My thinking is that the standard introduces the new role with few rules and leaves the "usage" of the NFT to the dApp. The same is true with ERC173 where it does not specify the exact functions that are restricted to the `owner` role, but the role is defined nevertheless.

In the Su Squares example, if the dApp were to implement the proposed `consumer` extension, that would enable owners of squares to delegate to other parties the personalisation of the square (enabling another protocol for renting/lending to integrate with Su Squares and build on top).

The same is true for Number Board or CryptoKitties, where if you are a consumer, you can manage the kitty (breeding).

I think that by introducing and implementing this extension we will enable the possibility for all NFTs that have utility to be rented out, which may be the next phase for NFTs, the same way we saw how renting/lending was a huge factor in making DeFi so successful

fulldecent:
I want to illustrate that there is difference between _templating_ something and _standardizing_ it.

Su Squares would have benefited from the proposed features, for sure.

And many contract like to be pausable as well.

OpenZeppelin Contracts offers the ERC721Pausable/Pausable contract https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Pausable.sol

This is a template, anybody can use it, it makes sense. Most people are just going to copy one line of code to get that feature.

But I don't know of any wallets that are planning to support OpenZeppelin Contracts `Pausable` contract and show that a contract is paused.

Because there are no wallets supporting it, there is no reason to standardize it. And it hasn't been.

You have not detailed anybody that wants to _read_ the consumer status for tokens. So you are in the same category~~no need to standardize this.

EIPS/eip-4400.md
```diff
@@ -0,0 +1,150 @@
+---
+eip: 4400
+title: ERC721 Consumer Extension
```
lightclient, line 3:
```suggestion
title: ERC-721 Consumer Extension
```

EIPS/eip-4400.md
```diff
@@ -0,0 +1,150 @@
...
+description: Standard interface extension for ERC721 Consumer role
```
lightclient, line 4:
```suggestion
description: Standard interface extension for ERC-721 consumer role  
```

EIPS/eip-4400.md
```diff
@@ -0,0 +1,150 @@
+---
+eip: 4400
+title: ERC721 Consumer Extension
+description: Standard interface extension for ERC721 Consumer role  
+author: Daniel Ivanov (@dani__ivanov) <daniel.k.ivanov95@gmail.com>
```
lightclient, line 5:
```suggestion
author: Daniel Ivanov (@dani__ivanov)
```

EIPS/eip-4400.md
```diff
@@ -0,0 +1,150 @@
...
+type: Standards Track
+category: ERC
+created: 2021-10-30
+requires: 721
+---
+
+## Abstract
+
+This specification defines standard functions outlining a `consumer` role for instance(s)
+of [ERC-721](https://eips.ethereum.org/EIPS/eip-721). An implementation allows reading the current `consumer` for a
```
lightclient, line 17:
```suggestion
of [ERC-721](./eip-721.md). An implementation allows reading the current `consumer` for a
```

Please link to all EIPs using this format.

EIPS/eip-4400.md
```diff
@@ -0,0 +1,150 @@
...
+for utilising/consuming a given NFT instance. The need for that role stems from the fact that other than owning the NFT
+instance, there are other actions that can be performed on an NFT. For example, various metaverses use`operator`
+/`contributor`
+roles for Land (ERC-721), so that owners of the land can authorise other addresses to deploy scenes to them (f.e.
+commissioning a service company to develop a scene).
+
+It is common for NFTs to have utility other than simply owning it. That being said, it requires a separate standardized
+consumer role, allowing compatibility with user interfaces and contracts, managing those contracts.
+
+Having a `consumer` role will enable protocols to integrate and build on top of dApps that issue ERC721 tokens.
```
lightclient, line 33:
```suggestion
Having a `consumer` role will enable protocols to integrate and build on top of dApps that issue ERC-721 tokens.
```

lightclient: Please replace `ERC721` with `ERC-721` in all places.

EIPS/eip-4400.md
```diff
@@ -0,0 +1,150 @@
...
+This specification defines standard functions outlining a `consumer` role for instance(s)
+of [ERC-721](https://eips.ethereum.org/EIPS/eip-721). An implementation allows reading the current `consumer` for a
+given NFT (`tokenId`) along with a standardized event for when an `consumer` has changed. The proposal depends on and
+extends the existing [ERC-721](https://eips.ethereum.org/EIPS/eip-721).
+
+## Motivation
+
+Many [ERC-721](https://eips.ethereum.org/EIPS/eip-721) contracts introduce their own custom role that grants permissions
+for utilising/consuming a given NFT instance. The need for that role stems from the fact that other than owning the NFT
+instance, there are other actions that can be performed on an NFT. For example, various metaverses use`operator`
```
lightclient, line 25:
```suggestion
instance, there are other actions that can be performed on an NFT. For example, various metaverses use `operator`
```

lightclient (changes requested):
Left a few comments - thanks!

Perseverance:
> I want to illustrate that there is difference between _templating_ something and _standardizing_ it.
> 
> Su Squares would have benefited from the proposed features, for sure.
> 
> And many contract like to be pausable as well.
> 
> OpenZeppelin Contracts offers the ERC721Pausable/Pausable contract https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Pausable.sol
> 
> This is a template, anybody can use it, it makes sense. Most people are just going to copy one line of code to get that feature.
> 
> But I don't know of any wallets that are planning to support OpenZeppelin Contracts `Pausable` contract and show that a contract is paused.
> 
> Because there are no wallets supporting it, there is no reason to standardize it. And it hasn't been.
> 
> You have not detailed anybody that wants to _read_ the consumer status for tokens. So you are in the same category~~no need to standardize this.

I think the `Pausable` example and the parallels drawn from it are not exactly in line with the idea of the suggested extension. 

While reasoning from a wallet perspective you are right - no one has wanted nor needed the `paused` state to be viewed. 

`Pausable`, however, is a strictly administrative "lever" - allowing for administrators to trigger it and grind a contract to a halt. No external entity - contract or wallet - has a great use case for reading the state of the contract. The `paused` state is only ever used internally in the contract as access restrictor.

The suggested extension however differs majorly in this regard. 

The `consumerOf` function is actually not intended for internal in-contract use. Its intended use case is for external entities - mainly contracts but also likely wallets - to reason with and base their logic on the consumer of an NFT rather than its `owner`. 

Contracts looking to build NFT legos based on consumer logic would actually need to know how to find out who is the consumer. Instead of allowing each project to name this interface (read function) for themselves, this extension will allow for a standard way of figuring out the `consumer`. 
In its essence this extension will enable NFT delegation. Some example use-cases of this extension would be:
- Metaverse land delegation - where a user can delegate the ability to build on metaverse land to another user. (Already exists in Decentraland, but would need to be built into the upcoming major metaverses)
- Virtual event passes delegation - where a user can delegate their access to a certain event or area to another user.
- In game assets leasing - where a user can delegate the abilities and effects of their in-game assets/wearables to another user in exchange for payment (or for free).

If we are looking for an analogy for this extension, I'd consider the `consumer`, `changeConsumer` and `consumerOf` much closer to `ERC721Metadata` extension of ERC721. Much like we don't want every project to name the function for access to the metadata URI differently, we don't want every project to name the function for access to the consumer address differently. (Unfortunately, this is already a fact with Decentraland using the term `operator` instead on their Land)

MindfulFroggie:
I completely agree this standard as a lot of benefit.
I commented [here](https://ethereum-magicians.org/t/erc-4400-erc-721-consumer-extension) regarding an improvement, so I'll add it here too:

With the current proposal, only the owner of the NFT can call the changeConsumer function. I think it would be good to add (or use) the mechanism of the ERC721 to aprove third parties to be able to use this function.

I imagine that there will be 3rd parties platforms (dapps) which NFT owners will use in order take advantage of this feature. Therefore the contract should give the 3rd parties the permission to change the consumers without the NFT owner having to do any action.

EIPS/eip-4445.md
```diff
@@ -0,0 +1,179 @@
...
+transfers & approvals.
+
+The `consumerOf(uint256 _tokenId)` function MAY be implemented as `pure` or `view`.
+
+The `changeConsumer(address _consumer, uint256 _tokenId)` function MAY be implemented as `public` or `external`.
+
+The `ConsumerChanged` event MUST be emitted when a consumer is changed.
+
+On every `transfer`, the consumer MUST be changed to the zero address the same way `approve` address is changed
+
```
wschwab, line 86: Hi! This is a pretty well written ERC from what I see. I'd recommend including EIP165 details here - that contracts implementing this ERC must respond `true` when `supportsInterface` is called with `0x953c8dfa`.

EIPS/eip-4445.md
```diff
@@ -0,0 +1,179 @@
...
+### Restriction on the Permissions
+
+There are numerous use-cases where a distinct role for NFTs is required that MUST NOT have owner permissions. A contract
+that implements the consumer role and grants ownership permissions to the consumer renders this standard pointless.
+
+## Backwards Compatibility
+
+There are no other standards that define a similar role for NFTs and the name (`consumer`) is not used by other ERC-721
+related standards.
+
```
wschwab, line 113: [EIP-1](https://eips.ethereum.org/EIPS/eip-1#what-belongs-in-a-successful-eip) specifies a "Test Cases" section. You can add tests to the implementation repo and link to them, if you'd like.

EIPS/eip-4445.md
```diff
@@ -0,0 +1,179 @@
...
+    function _beforeTokenTransfer(address _from, address _to, uint256 _tokenId) internal virtual override (ERC721) {
+        super._beforeTokenTransfer(_from, _to, _tokenId);
+
+        _changeConsumer(_from, address(0), _tokenId);
+    }
+
+}
+```
+
```
wschwab, line 176: There's also a "Security Considerations" section that needs to go here. It's ok if you say that there aren't any known considerations afaik, but I'd recommend talking about the potential cases where the permissions for the `consumer` aren't thought through clearly. (At least to me it does seem like there is a potential vector there - even if the letter of the standard is followed and the `consumer` can't transfer or burn the NFT, maybe they would otherwise be able to trigger things the owner wouldn't want.) This isn't a criticism - it's more of a warning to those implementing the standard to think things through.

Daniel-K-Ivanov: Good catch. Thank you for the feedback

Daniel-K-Ivanov:
@wschwab thank you for the review. I've addressed your comments 🙂

KKimos:
wschwab mentioned that our agreements are very similar, suggesting that our EIPs can be merged, hoping to further discuss.

KKimos:
wschwab mentioned that our agreements are very similar, suggesting that our EIPs can be merged, hoping to further discuss.

EIPS/eip-4445.md
```diff
@@ -0,0 +1,195 @@
+---
+eip: 4445
```
lightclient, line 2:
```suggestion
eip: 4400
```
Where did you get `4445` from?

lightclient (changes requested):

EIPS/eip-4445.md
Daniel-K-Ivanov: I thought that since 4400 has already passed, my PR should be updated to the next "available", but it seems that I was wrong. I've changed it back to the original one

lightclient (approved):

PR #4411, opened 2021-11-01
Add note about `FORK_NEXT` value re: EIP-2124

ralexstokes:
Concerning EIP-3675:

During the Amphora interop, it became clear we needed to agree on a value for `FORK_NEXT` as it pertains to EIP-2124.

This PR suggests a value of `0` and adds a short rationale for the decision.

PR #4415, opened 2021-11-03
Update specificiation on handling EIP-2124 during the consensus upgrade

ralexstokes:
Addressing PR feedback in #4411:

- Update link to EIP
- Specify what happens to the EIP-2124 values *after* the transition.

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -132,7 +132,7 @@ For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes impleme
 
 Once the transition has happened, `TRANSITION_POS_BLOCK` will have some block number `N`.
 This block number cannot be known ahead of time given the dynamic nature of the transition trigger condition.
-Once this block number is known, nodes can update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP 2124.
+Starting with `TRANSITION_POS_BLOCK`, nodes **MUST** update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP 2124.
 `FORK_NEXT` can remain `0` after the transition.
 
 #### devp2p
@@ -206,7 +206,6 @@ This event was removed for two reasons:
 
 The value of `FORK_NEXT` refers to the block number of the next fork a given node knows about and `0` otherwise.
 As the block number of the terminal PoW block will not be known until the transition, nodes have no value to use for `FORK_NEXT` and in light of this fact will use the default `0`.
-Nodes can retroactively update their node identifier with the correct data once the fork block is known.
 Using a value of `0` does imply extra networking overhead that EIP-2124 intends to eliminate, but given the mechanics of the transition process this decision seems the most pragmatic.
 
 ### Removing block gossip
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -128,7 +128,7 @@ Changes to the block tree store that are related to the above actions **MUST** b
 
 #### EIP-2124 fork identifier
 
-For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes implementing this EIP **SHOULD** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes implementing this EIP **MUST** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
 
 Once the transition has happened, `TRANSITION_POS_BLOCK` will have some block number `N`.
 This block number cannot be known ahead of time given the dynamic nature of the transition trigger condition.
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -132,7 +132,7 @@ For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes impleme
 
 Once the transition has happened, `TRANSITION_POS_BLOCK` will have some block number `N`.
 This block number cannot be known ahead of time given the dynamic nature of the transition trigger condition.
-Starting with `TRANSITION_POS_BLOCK`, nodes **MUST** update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP 2124.
+Starting with `TRANSITION_POS_BLOCK`, nodes **MUST** update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP-2124.
 `FORK_NEXT` can remain `0` after the transition.
 
 #### devp2p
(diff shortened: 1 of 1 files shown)


ralexstokes:
cc @axic

mkalinin:
Should this update be coordinated? The portion of nodes that has switched to the new `FORK_NEXT` value will drop all peers that doesn't support the new value because they haven't been updated yet. This updated should also set the value to the future block number, otherwise, it makes coordination infeasible.

We can set `FORK_NEXT` to a block number that according to our prediction will be the Merge, it shouldn't necessarily match the transition block number. The exact way of handling this value for the upgrades after the Merge is an open question and depends on what will trigger those upgrades, i.e. will it still be EL's block number or will it be CL epoch number. The latter seems more predictable as due to missed slots planned block number may hit later than CL epoch and could break the upgrade if it involves both layers. Basically, shift of the block number with respect to a slot number might not be an issue for the fork identifier functionality and it may still use the same approach for upgrades after the Merge and for the Merge itself.

EIPS/eip-3675.md
```diff
@@ -128,7 +128,12 @@ Changes to the block tree store that are related to the above actions **MUST** b
 
 #### EIP-2124 fork identifier
 
-For the purposes of the [EIP-2124 fork identifier](https://eips.ethereum.org/EIPS/eip-2124), nodes implementing this EIP **SHOULD** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes implementing this EIP **SHOULD** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+
+Once the transition has happened, `TRANSITION_POS_BLOCK` will have some block number `N`.
+This block number cannot be known ahead of time given the dynamic nature of the transition trigger condition.
+Once this block number is known, nodes can update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP 2124.
```
axic, line 135: Isn't this the correct wording?
```suggestion
Once this block number is known, nodes MUST update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP 2124.
```

EIPS/eip-3675.md
```diff
@@ -128,7 +128,12 @@ Changes to the block tree store that are related to the above actions **MUST** b
 
 #### EIP-2124 fork identifier
 
-For the purposes of the [EIP-2124 fork identifier](https://eips.ethereum.org/EIPS/eip-2124), nodes implementing this EIP **SHOULD** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes implementing this EIP **SHOULD** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
```
axic, line 131:
```suggestion
For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes implementing this EIP **MUST** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
```

EIPS/eip-3675.md
axic: Also instead of "Once this block number is known" shouldn't it just state the actual condition, i.e. "starting the block `TRANSITION_POS_BLOCK`..." ?

djrtwo (approved):

EIPS/eip-3675.md
```diff
@@ -128,7 +128,12 @@ Changes to the block tree store that are related to the above actions **MUST** b
 
 #### EIP-2124 fork identifier
 
-For the purposes of the [EIP-2124 fork identifier](https://eips.ethereum.org/EIPS/eip-2124), nodes implementing this EIP **SHOULD** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes implementing this EIP **MUST** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+
+Once the transition has happened, `TRANSITION_POS_BLOCK` will have some block number `N`.
+This block number cannot be known ahead of time given the dynamic nature of the transition trigger condition.
+Starting with `TRANSITION_POS_BLOCK`, nodes **MUST** update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP 2124.
```
axic, line 135: The style dictated by EIP-1:

```suggestion
Starting with `TRANSITION_POS_BLOCK`, nodes **MUST** update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP-2124.
```

EIPS/eip-3675.md
```diff
@@ -201,7 +206,6 @@ This event was removed for two reasons:
 
 The value of `FORK_NEXT` refers to the block number of the next fork a given node knows about and `0` otherwise.
```
axic, line 207: Is this section still needed? It seems to be properly described in the specification now.

axic (changes requested):
Sorry to interfere, but perhaps better to answer this here.

EIPS/eip-3675.md
```diff
@@ -128,7 +128,12 @@ Changes to the block tree store that are related to the above actions **MUST** b
 
 #### EIP-2124 fork identifier
 
-For the purposes of the [EIP-2124 fork identifier](https://eips.ethereum.org/EIPS/eip-2124), nodes implementing this EIP **SHOULD** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+For the purposes of the [EIP-2124 fork identifier](./eip-2124.md), nodes implementing this EIP **MUST** keep values of `FORK_NEXT` and `FORK_HASH` unaffected, i.e. `FORK_NEXT` is set to `0` and `FORK_HASH` retains the same value.
+
+Once the transition has happened, `TRANSITION_POS_BLOCK` will have some block number `N`.
+This block number cannot be known ahead of time given the dynamic nature of the transition trigger condition.
+Starting with `TRANSITION_POS_BLOCK`, nodes **MUST** update `FORK_HASH` by appending `uint64(N)` to the sequence of hashes used to calculate `FORK_HASH` as per EIP-2124.
```
mkalinin, line 135: If I understand EIP-2124 correctly, a local node will have to reject the connection with the remote node if the `FORK_HASH` of the remote node can't be completed with the locally known `FORK_NEXT`. This will be the case for the local node that is syncing and haven't reached the `TRANSITION_POS_BLOCK` yet while the transition has already happened in the network and the `TRANSITION_POS_BLOCK` is known to the remote peer, thus, the remote peer has updated its `FORK_HASH` value with the transition block.

PR #4416, opened 2021-11-03
Resurrect EIP-1328

ligi:
EIP-1328 is heavily used in WalletConnect - and very unlikely to change - but as I understand the process we now need to move it to Draft then to Review then to Last Call then to Final. Starting this process now - cc @pedrouid


pedrouid:
Seconded! We should also update the spec with parameters for v1.0 and v2.0

axic:
I think you can bring it to `Review` directly given it is mostly ready.

ligi:
@axic thanks - but according to this state-transition diagram I cannot:

image

axic (changes requested):
Please update this to the current EIP-1 standards:
1. Add `description:` preamble  and remove `Simple Summary`
2. Optionally add `Motivation`
3. Remove `References`
4. Add `Backwards Compatibility`
5. Add `Security Considerations`
6. Make sure at least one author has a GitHub handle and not an email

axic:
> @axic thanks - but according to this state-transition diagram I cannot:

Good catch. I think this is somewhat inconsistent, given if you have caught the bot at the time of making the change, you wouldn't need to do two steps.

ligi:
I actually caught it the first time: https://github.com/ethereum/EIPs/pull/4020#issuecomment-922449227
Unfortunately I did not catch the second attempt ..-)

Do we really need to add empty sections for backward compatibility and security considerations when they do not apply?

axic:
> Do we really need to add empty sections for backward compatibility and security considerations when they do not apply?

It seems so.

MicahZoltu:
I believe we discussed this and agreed that if the author wants to go from stagnant to Review that is fine.

The Backward Compatibility and Security Consideration sections are required.  At the least they should have a one-line blurb that says "There are no known backward compatibility issues introduced by this EIP" or "There are no known security issues presented by the introduction of this EIP."

lightclient:
Bumping as this is heavily used in WalletConnect and it would be great to move to final

ligi:
I did most things @axic pointed out changed from EIP-1 in style - just left

* Add Backwards Compatibility
* Add Security Considerations

Also @pedrouid can you add the 2.0 changes you mentioned?

EIPS/eip-1328.md
```diff
@@ -1,21 +1,18 @@
...
 ---
 
-## Simple Summary
-
-A standard to create WalletConnect URIs to initiate connections between applications and wallets.
-
 ## Abstract
 
-This standard defines how the data to connect some application and a wallet can be encoded with a URI. This URI can then be shown either as a QR code or for mobile to mobile as a link.
+This standard defines how the data to connect some application and a wallet can be encoded with a URI. This URI can then be shown either as a QR code or as a link.
```
ligi, line 15: @pedrouid do we perhaps want to remove "This URI can then be shown either as a QR code or as a link." - think it adds no value and restricts us without a need. e.g. how does the "copy" fall in place there.

pedrouid:
Thanks @ligi for resurrecting this EIP. I will look into this by Tuesday to include the missing sections and parameters for v2.0

pedrouid:
Sorry for the delay.... I have created a PR to Ligi's PR to include the v2.0 parameters: https://github.com/ligi/EIPs/pull/2

EIPS/eip-1328.md
pedrouid: I feel like this is already addressed when "link" is mentioned since it's a common pattern to include `copy to clipboard` for website links as well

ligi:
@pedrouid thanks for the final changes for WC2.0!
@MicahZoltu thanks for your suggestions - unfortunately I saw them after merging @pedrouid 's and then got this when trying to merge yours:
image
So I changed it myself and gave you credits in the commit msg - sorry for not seeing it earlier

Think this one should be mergeable now.

MicahZoltu:
@SamWilsn @lightclient @axic This has been sitting waiting for a review for quite a while.  Any chance one of you have some time to take a look?

EIPS/eip-1328.md
```diff
@@ -1,21 +1,18 @@
 ---
 eip: 1328
 title: WalletConnect Standard URI Format
```
SamWilsn, line 3:
```suggestion
title: WalletConnect URI Format
```

EIPS/eip-1328.md
```diff
@@ -1,21 +1,18 @@
...
-author: ligi <ligi@ligi.de>, Pedro Gomes <pedrouid@protonmail.com>
+description: Standard for WalletConnect URIs
```
SamWilsn, line 4:
```suggestion
description: Define URIs for initiating connections between applications and wallets
```

EIPS/eip-1328.md
```diff
@@ -28,26 +25,44 @@ WalletConnect request URI with the following parameters:
...
+# 1.0
 wc:8a5e5bdc-a0e4-4702-ba63-8f1a5655744f@1?bridge=https%3A%2F%2Fbridge.walletconnect.org&key=41791102999c339c844880b23950704cc43aa840f3739e365323cda4dfa89e7a
+
+# 2.0
+wc:c9e6d30fb34afe70a15c14e9337ba8e4d5a35dd695c39b94884b0ee60c69d168@2?relay-protocol=waku&symKey=7ff3e362f825ab868e20e767fe580d0311181632707e7c878cbeca0238d45b8b
 ```
 
 ## Rationale
 
 The need for this ERC stems from the discussion to move away from JSON format used in the alpha version of the WalletConnect protocol which makes for very inneficient parsing of the intent of the QR code, making it easier to create better QR code parsers APIs for Wallets to implement. Also by using a URI instead of a JSON inside the QR-Code the Android Intent system can be leveraged.
```
SamWilsn, line 57: This "Rationale" section reads more like it should belong in the "Motivation" section.

"Motivation" explains why the EIP as a whole needs to exist, while "Rationale" should explain individual choices made in the EIP itself.

SamWilsn (changes requested):

SamWilsn:
Opened https://github.com/ligi/EIPs/pull/5 with my suggestions.

ligi:
merged

SamWilsn (approved):
I think we're good to go!

SamWilsn:
@ligi is this good to go to review?

ligi:
yes!

Brandylee24 (approved):

PR #4424, opened 2021-11-05
Move EIPs 2363 & 2464 to Last Call

timbeiko:
^


MicahZoltu (changes requested):
>The specification is tiny since most parts are already specified in EIP-2124. `eth/63` is not specified as an EIP, but is maintained [here](https://github.com/ethereum/devp2p/blob/master/caps/eth.md).

Recommend deleting this.  It doesn't add value to the specification, and external links should not be included in EIPs (especially to devp2p GitHub, which have a history of turning into 404s).

I thought I mentioned this one before, but apparently I missed it.  😖

timbeiko:
@MicahZoltu addressed your comment :+1:

EIPS/eip-2364.md
```diff
@@ -41,7 +42,7 @@ Whenever two peers connect using the `eth/64` protocol, the updated `Status` mes
 
 ## Rationale
 
-The specification is tiny since most parts are already specified in EIP-2124. `eth/63` is not specified as an EIP, but is maintained [here](https://github.com/ethereum/devp2p/blob/master/caps/eth.md).
+The specification is tiny since most parts are already specified in EIP-2124. `eth/63` is not specified as an EIP, but is maintained in the `ethereum/devp2p` Github repository.
```
axic, line 45: I know @MicahZoltu has a thing against links, but this "non-link link" to me feels beyond pointless.

axic: Since the repository doesn't have tags, and the argument is that these repositories may change their structure, can we at least agree on pointing to https://github.com/ethereum/devp2p (as a proper link), because:
1) Repositories rarely are deleted, but rather renamed
2) archive.org can be used to find a copy
3) If that repository gets deleted and no information is to be found about what happened to the specification, then it is reasonable to believe the EIPs repository and other specification are useless too.

timbeiko: No strong opinion from me. Happy to change it to whatever is best.

axic: If it was only up to me, I'd pragmatically just allow links where there is no better alternative (which is the case here) 😅

MicahZoltu: I don't personally see the value in that paragraph in the context of a technical specification.  Just like we don't have every core EIP link to the yellow-paper, I think it doesn't make sense to have every Networking EIP link to the devp2p repository.

The size of the changeset is inconsequential IMO and not something that we need to spend space on the EIP explaining.  Small EIPs are generally always better, and we should never need to feel the need to defend them.  In fact, I would weakly argue that you should have to defend *large* EIPs and explain why it can't be broken up into smaller EIPs, but not the other way around.

timbeiko: Given these specs are already deprecated, can we optimize for merge speed and just remove the link :-) ?

axic: This is my preferred solution:
```suggestion
The specification is tiny since most parts are already specified in EIP-2124. `eth/63` is not specified as an EIP, but is maintained in the [ethereum/devp2p](https://github.com/ethereum/devp2p) Github repository.
```

MicahZoltu: My preferred solution:
```suggestion
```

MicahZoltu: As a reminder, despite appearances the EIP process is not actually a dictatorship.  If you want me to merge, remove the line.  If you want Alex to merge, change the line.  If you want something else, bribe @lightclient to allow some third option.  😆

timbeiko:
@MicahZoltu based on your latest comment, I'll go with @axic's suggestion. Main reason is this is an old EIP and I think his approach keeps the EIP closest to its original form. @axic, I need an "Approve" from you now, though 😅

axic (approved):

timbeiko:
@MicahZoltu I think you need to remove your "change requested" for it to merge.

EIPS/eip-2364.md
```diff
@@ -4,7 +4,8 @@ title: "eth/64: forkid-extended protocol handshake"
 description: Introduces validation of the `forkid` when handshaking with peers.
 author: Péter Szilágyi <peterke@gmail.com>, Péter Szilágyi (@karalabe), Tim Beiko (@timbeiko)
 discussions-to: https://github.com/ethereum/EIPs/issues/2365
-status: Review
+status: Last Call
+review-period-end: 2021-11-19
```
axic, line 8:
```suggestion
last-call-deadline: 2021-11-19
```

EIPS/eip-2464.md
```diff
@@ -4,7 +4,8 @@ title: "eth/65: transaction announcements and retrievals"
 description: Introduces `NewPooledTransactionHashes`, `GetPooledTransactions`, and `PooledTransactions`.
 author: Péter Szilágyi <peterke@gmail.com>, Péter Szilágyi (@karalabe), Gary Rong <garyrong0905@gmail.com>, Tim Beiko (@timbeiko)
 discussions-to: https://github.com/ethereum/EIPs/issues/2465
-status: Review
+status: Last Call
+review-period-end: 2021-11-19
```
axic, line 8:
```suggestion
last-call-deadline: 2021-11-19
```

axic (commented):
The field names have changed meanwhile 😅

PR #4427, opened 2021-11-07
Move EIP-3224 back from stagnant to draft

ricmoo:
Move EIP-3224 back from stagnant to draft

(I think something got messed up with my merge; one sec while I look into it)


ricmoo:
Definitely something got messed up. I'll close this and try again.

PR #4428, opened 2021-11-07
Unstagnant EIP-3224

ricmoo:
Move EIP-3224 from Stagnant back to Draft.


MicahZoltu (changes requested):
The Simple Summary section should be removed and a `description` header field should be added below `title` that contains essentially the same thing.  Try to keep it as short as possible, one sentence (and one line, don't hardcode newlines).

Remove external links from Test Cases.  Contracts that the tests are run against should be included either as an asset or inline, or just left out if they aren't required.

ricmoo:
I was planning on doing that once the PR was put back into draft, so that this PR would be as simple as possible. There are also a few spelling mistakes I was planning.

ricmoo:
(or should I do that first?)

MicahZoltu:
Those changes should be made before merging as draft, the same as for someone creating a new EIP.  The idea is that an EIP should be structurally sound and generally follow the various rules we have if it shows up on the EIPs page.

ricmoo:
I also removed all the external links in the security section.

ricmoo:
I've made the suggested changes.

EIPS/eip-3224.md
```diff
@@ -402,12 +403,12 @@ Wallets must be careful when displaying text provided by
 contracts and proper efforts must be taken to sanitize
 it, for example, be sure to consider:
 
-- HTML could be enbedded to attempt to trick web-based wallets into [executing code](https://en.wikipedia.org/wiki/Code_injection) using the script tag (possibly uploading any private keys to a server)
+- HTML could be enbedded to attempt to trick web-based wallets into executing code using the script tag (possibly uploading any private keys to a server)
```
axic, line 406:
```suggestion
- HTML could be embedded to attempt to trick web-based wallets into executing code using the script tag (possibly uploading any private keys to a server)
```

axic (changes requested):
Some issues:
- still has simple summary section.
- `function eipXXXDescribe` should probably be renamed.
- there's a TODO in rationale

And as a comment: I am not sure if this should be two EIPs, one for the on-chain element and the other for RPC (or wherever RPCs are maintained now).

ricmoo:
@axic 

It will be two EIPs, I’m just trying to get it back into draft status so I can make meaningful changes to it again. :)

For now I thought it’d be best to intentionally leave the `describeEipXXX` because I don’t wNt people jumping the gun and adding it yet. I don’t want another `eth_signTypedData_v4` situation.

I can remove the TODO. I was thinking that would be fine in a draft.

Thanks! :)

ricmoo:
Thanks. Done. :)

axic (approved):
Looks good enough to me. Not good enough to move past Draft though.

PR #4429, opened 2021-11-07
Unstagnant eip 634

ricmoo:
Moved EIP-634 from Stagnant back to Draft.


MicahZoltu (changes requested):
`description` frontmatter field is missing.  Should be a short description (about a sentence) of this EIP meant for humans to be reminded of the EIP's contents.

Remove Implementation section.  There is an optional Reference Implementation section, but since you don't appear to have one or it isn't applicable here it should be left out.

Add missing Security Consideration section just before the Copyright section.

Remove external link to Wikipedia for `phone`.

----

I'm going to keep saying this on each of these ENS EIPs, even though it isn't a blocker I do feel pretty strongly that these should not be in the EIPs repository.  ENS is an application that should have its own documentation.  It doesn't need standardization because there is one ENS and many ENS consumers.  This is a prime example of where documentation is required, not standardization.

ricmoo:
I've made the suggested changes.

axic:
> A new resolver interface is defined, consisting of the following method:
>     function text(bytes32 node, string key) constant returns (string text);
> The interface ID of this interface is 0x59d1d43c.

Suggestion:
1. Enclose the `text` line in code highlighting and add NatSpec.
2. Make it clear the interface ID refers to EIP-165, and also add `requires: 165`.
3. I think this should depends on EIP-137 otherwise it is out of context.

axic (changes requested):
I left some suggestions, but actually `requires: 137` I think is mandatory as this extends ENS.

EIPS/eip-634.md
```diff
@@ -29,9 +31,15 @@ associate arbitrary key-value text.
 
 A new resolver interface is defined, consisting of the following method:
 
-    function text(bytes32 node, string key) constant returns (string text);
+```solidity
```
axic, line 34: Actually, what many other EIPs do when they claim an interface ID is to actually describe this as an interface (because then Solidity can generate the interface IDs):
```solidity
interface IERC634 {
    function text(....
}
```

This is only a suggestion.

axic (commented):
Looks good to me, but please consider the suggestion and let me know the decision. Will approve then to avoid an early merge now.

EIPS/eip-634.md
```diff
@@ -114,8 +122,8 @@ Not applicable.
 ## Test Cases
```
axic, line 122: Actually this can be removed.

ricmoo: Removed.

EIPS/eip-634.md
ricmoo: I've made that change, thanks! :)

EIPS/eip-634.md
```diff
@@ -111,11 +121,9 @@ allowing applications to use custom keys for their own purposes.
 Not applicable.
 
 
-## Test Cases
-TBD
+## Security Consideration
```
axic, line 124:
```suggestion
## Security Considerations
```

axic (approved):

PR #4434, opened 2021-11-08
3675: refactor the rationale

ralexstokes:
Addressing PR feedback from this PR https://github.com/ethereum/EIPs/pull/4415 that was auto-merged a bit too early

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -128,11 +128,9 @@ Changes to the block tree store that are related to the above actions **MUST** b
 
 #### EIP-2124 fork identifier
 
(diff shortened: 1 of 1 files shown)


ralexstokes:
cc @axic

MicahZoltu:
This needs approval from the authors, not an editor.

mkalinin (approved):
Approving this PR but we should revisit a part of specification that covers EIP-2124 logic, see my concern [here](https://github.com/ethereum/EIPs/pull/4415/files#r745337814)

axic:
> Approving this PR but we should revisit a part of specification that covers EIP-2124 logic, see my concern here

@mkalinin I think this should be moved to the discussion URL as now this question has been chased through 3 PRs without resolution :(

mkalinin:
> > Approving this PR but we should revisit a part of specification that covers EIP-2124 logic, see my concern here
> 
> @mkalinin I think this should be moved to the discussion URL as now this question has been chased through 3 PRs without resolution :(

Done! https://ethereum-magicians.org/t/eip-3675-upgrade-consensus-to-proof-of-stake/6706/7

axic:
@mkalinin thanks!

PR #4436, opened 2021-11-09
Make the stale bot ignore discussion urls

axic:
See https://github.com/actions/stale#exempt-issue-labels

After this change, if an issue is labeled `discussion-url`, it will be exempt from closure.  We should reopen the actual discussion URLs which have been closed recently and label them too.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -18,6 +18,7 @@ jobs:
         days-before-stale: 60
         days-before-close: 7
         stale-pr-label: 'stale'
+        exempt-issue-labels: 'discussion-url'
   stale-issue:
     runs-on: ubuntu-latest
     name: "Mark stale issues"
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,12 +13,11 @@ jobs:
     - uses: actions/stale@v3
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-pr-message: 'There has been no activity on this pull request for two months. It will be closed in a week if no further activity occurs. If you would like to move this EIP forward, please respond to any outstanding feedback or add a comment indicating that you have addressed all required feedback and are ready for a review.'
-        close-pr-message: 'This pull request was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.'
         days-before-stale: 60
         days-before-close: 7
         stale-pr-label: 'stale'
-        exempt-issue-labels: 'discussion-url'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         days-before-stale: 60
         days-before-close: 7
         stale-pr-label: 'stale'
+        exempt-issue-labels: 'discussion-url'
```
axic, line 21: See https://github.com/actions/stale#exempt-issue-labels

.github/workflows/stale.yml
```diff
@@ -27,8 +27,7 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         days-before-stale: 180
         days-before-close: 14
-        stale-pr-label: 'stale'
+        exempt-issue-labels: 'discussion-url'
```
axic, line 30: Any better suggestion for the label?

MicahZoltu: `discussions-to` maybe (so it matches the header field name in the EIP)?

.github/workflows/stale.yml
```diff
@@ -27,8 +27,7 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         days-before-stale: 180
         days-before-close: 14
-        stale-pr-label: 'stale'
```
MicahZoltu, line 30: Why is this removed as well?  Feels unrelated to the addition of `exempt-issue-labels`?

MicahZoltu (approved):
Tentatively approve, but would like to hear a response to the `stale-pr-label` removal question before we merge.

.github/workflows/stale.yml
axic: We have two different configurations here: one for issues the other for PRs.

This section here is for the issue, and has the appropriate `stale-issue-label`. The `stale-pr-label` is a useless setting in this context.

.github/workflows/stale.yml
axic: I'm fine with that, though the value is an uri. Perhaps we should rename it to `discussion-uri` both label and in the EIPs? :)

MicahZoltu: I would approve a PR that changes it to `discussion-uri` everywhere.  If you are going to immediately follow this PR up with all of the necessary PRs to change to `discussion-uri` I'm fine with leaving this as is.  If that is just an idea to put on the backburner, I think we should maintain consistency for now and have the label match the field.

MicahZoltu: Let me know which way you would like to go, or if you would like to continue discussion, so I can know to merge or not.

axic: Renamed to `discussions-to` here just to get this merged. We can make a bigger PR later and renaming the label is once off on Github, all tagged issues carry it over.

PR #4437, opened 2021-11-09
2681: Clarify that 2^64-1 is invalid as a transaction nonce

axic:
Based on https://github.com/ethereum/go-ethereum/pull/23853.

### DIFF
--- a/EIPS/eip-2681.md
+++ b/EIPS/eip-2681.md
@@ -37,7 +37,7 @@ This mode of replay protection is out of fashion since [EIP-155](./eip-155.md) i
 
 3. Most clients already consider the nonce field to be 64-bit, such as go-ethereum.
 
-4. The reason a transaction with nonce `2^64-1` is invalid, because after inclusion the account's nonce would exceed `2^64-1`.
+4. The reason a transaction with nonce `2^64-1` is invalid, because otherwise after inclusion the sender account's nonce would exceed `2^64-1`.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-2681.md
+++ b/EIPS/eip-2681.md
@@ -25,7 +25,7 @@ Lastly, this facilitates a minor optimisation in clients, because the nonce no l
 
 Introduce two new restrictions retroactively from genesis:
 
-1. Consider any transaction invalid, where the nonce exceeds `2^64-1`.
+1. Consider any transaction invalid, where the nonce exceeds or equals to `2^64-1`.
 2. The `CREATE` and `CREATE2` instructions to abort with an exceptional halt, where the account nonce is `2^64-1`.
 
 ## Rationale
@@ -36,6 +36,8 @@ Introduce two new restrictions retroactively from genesis:
 This mode of replay protection is out of fashion since [EIP-155](./eip-155.md) introduced a more elegant way using chain identifiers.
 
 3. Most clients already consider the nonce field to be 64-bit, such as go-ethereum.
+
+4. The reason a transaction with nonce `2^64-1` is invalid, because after inclusion the account's nonce would exceed `2^64-1`.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-2681.md
+++ b/EIPS/eip-2681.md
@@ -45,7 +45,7 @@ While this is a breaking change, no actual effect should be visible:
 
 1. There is no account in the state currently which would have a nonce exceeding that value. As of November 2020, the account `0xea674fdde714fd979de3edf0f56aa9716b898ec8` is responsible for the highest account nonce at approximately 29 million.
 
-2. go-ethereum already has this restriction in place (`state.Account.Nonce` and `types.txdata.AccountNonce` it as a 64-bit number).
+2. go-ethereum already has this restriction partially in place (`state.Account.Nonce` and `types.txdata.AccountNonce` it as a 64-bit number).
 
 ## Security Considerations
 


axic:
@holiman @gumb0 can you please check?

EIPS/eip-2681.md
```diff
@@ -37,6 +37,8 @@ This mode of replay protection is out of fashion since [EIP-155](./eip-155.md) i
 
 3. Most clients already consider the nonce field to be 64-bit, such as go-ethereum.
 
+4. The reason a transaction with nonce `2^64-1` is invalid, because after inclusion the account's nonce would exceed `2^64-1`.
```
gumb0, line 40:
```suggestion
4. The reason a transaction with nonce `2^64-1` is invalid, because otherwise after inclusion the sender account's nonce would exceed `2^64-1`.
```

gumb0:
Point 2 in Backwards Compatibility becomes not exactly accurate now. Maybe just change to "restriction is partially in place".

(`types.txdata.AccountNonce` name looks also outdated by now, but maybe not worth to change it. But just in case it's `types.TxData.nonce()` function now)

holiman (approved):
LGTM (with @gumb0 's suggestion)

axic:
@gumb0 @holiman merged the suggestions

gumb0 (approved):

EIPS/eip-2681.md
```diff
@@ -25,7 +25,7 @@ Lastly, this facilitates a minor optimisation in clients, because the nonce no l
 
 Introduce two new restrictions retroactively from genesis:
 
-1. Consider any transaction invalid, where the nonce exceeds `2^64-1`.
+1. Consider any transaction invalid, where the nonce exceeds or equals to `2^64-1`.
```
MicahZoltu, line 28: This is a normative change, which makes me fairly uncomfortable making it as an edit to a final EIP.  One could create a new EIP that adds 2^64-1 to the set of allowed nonces, but policy is that final EIPs can have non-normative changes only.

MicahZoltu (changes requested):
I'm pretty uncomfortable with making a normative change to a final EIP, even if the change makes sense.  I would be more comfortable with a new EIP that makes 2^64-1 an invalid nonce (I doubt anyone would try to stop it from going through).

holiman:
> I'm pretty uncomfortable with making a normative change to a final EIP, even if the change makes sense. I would be more comfortable with a new EIP that makes 2^64-1 an invalid nonce (I doubt anyone would try to stop it from going through).

Not sure what you mean. As I see it, it's a clarification, not a change. 

So the base idea is that nonces are limited between `0-2^64-1`. This means that any transaction, which has nonce `2^64-1` is valid. However, when executing it, it will push the nonce to `2^64` (flipping the nonce back to `0` for the sender). This must be rejected, and the tx excluded. 

So: the tx is "atomically" valid, but adding it causes an invalid state. Hence, we can shortcut this and say "tx with nonce `2^64-1` is invalid"

MicahZoltu:
> when executing it, it will push the nonce to 2^64 (flipping the nonce back to 0 for the sender).

While I agree that the behavior you have described is reasonable, it is not spec compliant as currently written.  A spec compliant client would increment the nonce to 2^64 and then allow no more transactions from that account going forward.  I think the change proposed here is a good idea, I just argue that it is a normative change as it would cause a previously spec compliant client to become spec non-compliant and thus a new EIP is required.

The only specification details we have currently for this behavior is the following, which does not actually say that the nonce cannot be 2^64, it only says that a transaction submitted with a nonce >=2^64 is invalid.

> Introduce two new restrictions retroactively from genesis:
> 1. Consider any transaction invalid, where the nonce exceeds `2^64-1`.
> 2. The `CREATE` and `CREATE2` instructions to abort with an exceptional halt, where the account nonce is `2^64-1`.

holiman:
>  A spec compliant client would increment the nonce to 2^64 and then allow no more transactions from that account going forward.


The abstract of the EIP in full is literally:

> Limit account nonce to be between `0` and `2^64-1`.

MicahZoltu:
I would accept a PR to this EIP that updates the abstract to correctly reflect the specification.  The specification is the root of truth for the EIP, the abstract is meant to be a human readable summary.  In this case the abstract doesn't match, so it should be fixed.

holiman:
... Does it in any way change your opinion if I point out that the _title_ is "Limit account nonce to 2^64-1", and it's by that title it has been discussed and approved? 
Because if we change the abstract to be according to spec, we should also change the title to 

"Limit account nonce to be max 2^64-1 for contracts, and 2^64 for EOA accounts"

I don't think such a change is in line with the thinking that went into writing it (as the eip authors can confirm), and more importantly, nor with the discussions that led to it's approval. 

Both alternatives are pretty shitty really, IMO one is less shitty than the other.

lightclient (approved):
Ugh this is unfortunate we didn't catch it before it become final. I am personally in favor of accepting this PR because I think 1) it really is clarifying the true purpose of the EIP. However, with respect to the `Specification` section it *is* a non-normative change. 2) I think it will cause more harm than good to create a new EIP with the small modification as it will be another strange artifact in Ethereum history for future developers to try and follow.

axic:
In an editor capacity I agree with @lightclient that merging this is causing the least confusion in the mid/long term. Potentially bringing this to ACD for wider attention could be useful, or even just privately reaching out to other clients whether they actually started implementing this.

As an alternate solution I considered the idea of an `Errata` section a while back, which could be added to final EIPs listing post-finalisation fixes. Not sure we want to complicate matters that much for this change though, but could definitely work for something like the EIP-820 vs EIP-1820 case.

MicahZoltu:
As a reminder, EIP editorial process isn't a dictatorship and we (for the first time in a long time) seem to actually have 3 active editors.  Since @axic and @lightclient seem to think that a normative change to this EIP is reasonable in this case, I can just 🙈 and they can merge.  😬

PR #4441, opened 2021-11-09
Remove `updated` field

lightclient:
```console
$ rg "updated: " . | wc -l
0
```

Doesn't look like anyone uses the updated field.


axic (approved):
I liked it, but the process underwent substantial changes now.

MicahZoltu (approved):

PR #4444, opened 2021-11-09
Bound history size

lightclient:
Clients must stop serving historical blocks/receipts that are older than a year on the p2p layer. Clients may prune such historical data.

### DIFF
--- a/EIPS/eip-4444.md
+++ b/EIPS/eip-4444.md
@@ -2,7 +2,7 @@
 eip: 4444
 title: Bound Historical Data in Execution Clients
 description: Prune historical data in clients older than one year
-author: George Kadianakis (@asn-d6), lightclient (@lightclient), Alex Stokes (@ralexstokes)
+author: George Kadianakis (@asn_d6), lightclient (@lightclient), Alex Stokes (@ralexstokes)
 discussions-to: TBD
 status: Draft
 type: Standards Track
@@ -12,21 +12,15 @@ created: 2021-11-02
 
 ## Abstract
 
-Clients may prune history that is older than a year.
+Clients must stop serving historical blocks/receipts that are older than a year on the p2p layer. Clients may prune such historical data.
 
 ## Motivation
 
 Historical blocks, states, and receipts currently occupy more than 400GB of disk space (and growing!). Therefore, to validate the chain, users must typically have a 1TB disk.
 
-Historical data is not necessary for validating new blocks, so once a client has synced the tip of the chain, historical data is only retrieved when requested explicitly over the JSON-RPC or when a peer attempts to sync the chain. By pruning the history, this proposal reduces the disk requirements for users.
+Historical data is not necessary for validating new blocks, so once a client has synced the tip of the chain, historical data is only retrieved when requested explicitly over the JSON-RPC or when a peer attempts to sync the chain. By pruning the history, this proposal reduces the disk requirements for users. Pruning history also allows clients to remove code that processes historical blocks. This means that execution clients don't need to maintain code paths that deal with each upgrade's compound changes.
 
-Pruning history also allows clients to remove code that processes historical blocks. This means that execution clients don't need to maintain code paths that deal with each upgrade's compound changes.
-
-Finally, the p2p network also benefits from history pruning since there will be less networking strain as clients adopt more lightweight sync strategies.
-
-Although there are strong benefits to pruning client history, securely bootstrapping nodes without it is hard. The move to PoS presents a great opportunity to rely on the weak subjectivity assumption in PoS and simplify client syncing logic.
-
-
+This proposal also concretizes the availability guarantees of the p2p layer as clients won't be able to fetch historical data anymore. This results in less strain on the network as clients adopt more lightweight sync strategies based on the PoS weak subjectivity assumption.
 
 ## Specification
 
@@ -34,8 +28,9 @@ Although there are strong benefits to pruning client history, securely bootstrap
 | - | - | - |
 | `HISTORY_PRUNE_EPOCHS` | 82125 | A year in beacon chain epochs |
 
+Clients MUST NOT serve blocks/states/receipts that are older than `HISTORY_PRUNE_EPOCHS` epochs on the p2p network.
 
-Clients MAY prune block/state/receipt history that is older than `HISTORY_PRUNE_EPOCHS` epochs.
+Clients MAY locally prune block/state/receipt history that is older than `HISTORY_PRUNE_EPOCHS` epochs.
 
 #### Bootstrapping and syncing
 
@@ -46,6 +41,8 @@ Clients MUST use a valid Weak Subjectivity Checkpoint to bootstrap from a more r
 For the purposes of this proposal, we assume clients always start with a configured and valid weak subjectivity checkpoint. The way this is achieved is out-of-scope for this proposal.
 
 ## Rationale
+
+This proposal forces clients to stop serving old historical data over p2p. We make this explicit to force clients to seek historical data from other sources, instead of relying on the optional behavior of some clients which would result in quality degradation.
 
 ### Why a year?
 
--- a/EIPS/eip-4444.md
+++ b/EIPS/eip-4444.md
@@ -6,7 +6,7 @@ author: George Kadianakis (@asn_d6), lightclient (@lightclient), Alex Stokes (@r
 discussions-to: https://ethereum-magicians.org/t/eip-4444-bound-historical-data-in-execution-clients/7450
 status: Draft
 type: Standards Track
-category: Core
+category: Networking
 created: 2021-11-02
 ---
 
--- a/EIPS/eip-4444.md
+++ b/EIPS/eip-4444.md
@@ -1,0 +1,104 @@
+---
+eip: 4444
+title: Bound Historical Data in Execution Clients
+description: Prune historical data in clients older than one year
+author: George Kadianakis (@asn-d6), lightclient (@lightclient), Alex Stokes (@ralexstokes)
+discussions-to: TBD
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-11-02
+---
+
+## Abstract
+
+Clients may prune history that is older than a year.
+
+## Motivation
+
+Historical blocks, states, and receipts currently occupy more than 400GB of disk space (and growing!). Therefore, to validate the chain, users must typically have a 1TB disk.
+
+Historical data is not necessary for validating new blocks, so once a client has synced the tip of the chain, historical data is only retrieved when requested explicitly over the JSON-RPC or when a peer attempts to sync the chain. By pruning the history, this proposal reduces the disk requirements for users.
+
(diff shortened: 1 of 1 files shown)


EIPS/eip-4444.md
```diff
@@ -0,0 +1,101 @@
+---
+eip: 4444
+title: Bound Historical Data in Execution Clients
+description: Prune historical data in clients older than one year
+author: George Kadianakis (@asn_d6), lightclient (@lightclient), Alex Stokes (@ralexstokes)
+discussions-to: TBD
```
MicahZoltu, line 6: Required.

EIPS/eip-4444.md
```diff
@@ -0,0 +1,101 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-11-02
+---
+
+## Abstract
+
+Clients must stop serving historical blocks/receipts that are older than a year on the p2p layer. Clients may prune such historical data.
```
MicahZoltu, line 15: **MUST** seems overly prescriptive here.  Perhaps instead we could say **MAY**?

EIPS/eip-4444.md
```diff
@@ -0,0 +1,101 @@
+---
+eip: 4444
+title: Bound Historical Data in Execution Clients
+description: Prune historical data in clients older than one year
+author: George Kadianakis (@asn_d6), lightclient (@lightclient), Alex Stokes (@ralexstokes)
+discussions-to: TBD
+status: Draft
+type: Standards Track
+category: Core
```
MicahZoltu, line 9: I think this should go in `Networking` rather than `Core`.  Also, it should include details on what the expected response should be to a request for blocks older than the save duration.

EIPS/eip-4444.md
```diff
@@ -0,0 +1,101 @@
...
+
+This proposal also concretizes the availability guarantees of the p2p layer as clients won't be able to fetch historical data anymore. This results in less strain on the network as clients adopt more lightweight sync strategies based on the PoS weak subjectivity assumption.
+
+## Specification
+
+| Parameter | Value | Description |
+| - | - | - |
+| `HISTORY_PRUNE_EPOCHS` | 82125 | A year in beacon chain epochs |
+
+Clients MUST NOT serve blocks/states/receipts that are older than `HISTORY_PRUNE_EPOCHS` epochs on the p2p network.
```
MicahZoltu, line 31: Not everyone will agree on what time it is, should we have a bit of a no-mans-land between the time clients are expected to stop *requesting* ancient data and when clients are expected to stop *providing* ancient data?  This could be minutes/hours/days, just enough to deal with time synchronization issues.

Presumably, under such a regime any peer that requests data older than the longer of the two **SHOULD** be kicked.  Meanwhile, clients **SHOULD** be programmed to never ask for anything older than the shorter of the two.

EIPS/eip-4444.md
```diff
@@ -0,0 +1,101 @@
...
+
+Clients MAY locally prune block/state/receipt history that is older than `HISTORY_PRUNE_EPOCHS` epochs.
```
MicahZoltu, line 33: Specification should indicate what behavior should occur if a peer asks for data that is too old (which is why I think this should be a Networking spec).

MicahZoltu (changes requested):
A ballsy proposal, in need of a discussions-to link and I think a category change.

EIPS/eip-4444.md
lightclient: @MicahZoltu added.

EIPS/eip-4444.md
```diff
@@ -0,0 +1,101 @@
+---
+eip: 4444
```
MicahZoltu, line 2: /me goes and checks to see if you farmed this EIP number...

EIPS/eip-4444.md
lightclient: Given the definition of a `Core` EIP from EIP-1, I feel like this fits perfectly in the "non-consensus critical change, but relevant to core dev discussions". This EIP affects networking, RPC interface, and the possibility of full validating the chain.

> Core: improvements requiring a consensus fork (e.g. EIP-5, EIP-101), as well as changes that are not necessarily consensus critical but may be relevant to “core dev” discussions (for example, [EIP-90], and the miner/node strategy changes 2, 3, and 4 of EIP-86)

MicahZoltu: EIP-90 isn't an EIP... we should fix that example... I think functionally, all Core EIPs are consensus changes.  I would rather fix EIP-1 to reflect that then use what it says to allow a single EIP that doesn't follow the pattern that every other EIP (I think, at least everything in the last couple years) currently follows.

EIPS/eip-4444.md
lightclient: Crazy coincidence!

MicahZoltu: 🤨

EIPS/eip-4444.md
asn-d6: Hello. The idea here is that if we make this optional a client Alice might continue serving old data, and Bob will rely on Alice out of inertia. At some point, it's likely that Bob will get rugged by Alice since this is an optional feature and Alice might get bored of serving old data. By making this a **MUST** it forces Bob to find a long-term solution to his problem.

EIPS/eip-4444.md
lightclient: Meh fair. It still *feels* like a core change because it affecting many things like networking, RPC, and syncing trust assumptions. But I've updated it to `Networking` for now.

EIPS/eip-4444.md
lightclient: Why should anything specific occur? There are a small number of clients, we should work together to ensure that they follow the behavior of the EIP.

lightclient:
This should be ready to be reviewed again. It would be great to move any outstanding questions regarding the actual EIP to the discussion thread: https://ethereum-magicians.org/t/eip-4444-bound-historical-data-in-execution-clients/7450

EIPS/eip-4444.md
MicahZoltu: I can see **SHOULD** being applied along with rationale for it (though I don't agree with the rationale).  **MUST** should be used only if it is an error to not follow the rules.  In this case, it is not an error to serve old data so it shouldn't be a **MUST**.

MicahZoltu (approved):

EIPS/eip-4444.md
djrtwo: Note, that discussions with @karalabe have pointed to making this a **MUST** because otherwise users will still rely on it while quality of this feature degrades until it is unusable. A **MUST**, instead, forces users and dapps to actually "upgrade" how they utilize this at the point of this shipping, forcing everyone's hand so it doesn't silently get worse and worse.

If we go that path, then this spec should specify that devp2p *does* return errors on requests outside of the specified range of epochs/time

EIPS/eip-4444.md
djrtwo: So I can/should locally prune after these epochs, and thus I can be in a situation where our clocks have a disparity  (within reason) and you make what you think is an honest request that I do not have.  Are you suggesting in such a situation that although I cannot respond to the request, I *should not* consider your request as nefarious? Whereas if you are outside of the window, I can consider this as "bad" behavior and descore/kick you?

> under such a regime any peer that requests data older than the longer of the two SHOULD be kicked

I prefer **MAY** be downscored and/or kicked.

EIPS/eip-4444.md
```diff
@@ -0,0 +1,101 @@
...
+
+Clients MAY locally prune block/state/receipt history that is older than `HISTORY_PRUNE_EPOCHS` epochs.
+
+#### Bootstrapping and syncing
+
+This EIP impacts the way clients bootstrap and sync. Clients won't be able to full sync or snapshot sync since historical data won't be available on the p2p network.
+
+Clients MUST use a valid Weak Subjectivity Checkpoint to bootstrap from a more recent view of the chain. For the purpose of syncing, clients treat weak subjectivity checkpoints as the genesis block. We call this method "checkpoint sync".
+
+For the purposes of this proposal, we assume clients always start with a configured and valid weak subjectivity checkpoint. The way this is achieved is out-of-scope for this proposal.
```
axic, line 41: Actually back in April 2020 or so, within the Ipsilon team we thought about making a proposal to just hardcode given hashes for given blocks in an EIP. The idea would be to hardcode the hashes for past hard forks.

However then regenesis was proposed, which in practice does the same, but programatically.

Since regenesis as a concept is delayed ever so often, and given this EIP, would such a proposal make sense now?

EIPS/eip-4444.md
djrtwo: It's been argued that requesting and serving data beyond the limit should be explicitly incorrect. In such a case, you'd need to have devp2p return errors on such requests

EIPS/eip-4444.md
lightclient: Moved to magicians thread.

EIPS/eip-4444.md
lightclient: Moving to magicians thread.

EIPS/eip-4444.md
ytrezq: A new update on this but TheGraph just censored Tornado Cash.
The underlying problem is things like databased shared over ipfs or Google Bigquery is only for static content so wouldn’t allows us to have a realtime service as it would take a day to download updates. We also need very old data in order to let a user withdraw funds deposited several years ago.

This isn’t that much a problem currently as this just mean we will need our own home (not cloud) hosted node with parity_tracing, along for getting smart contract storage range at past blocks. Such thing is possible because full anciant data is broadcasted and updated in real time but show why it is important to have node/rpc be able to broadcast full chain history over the p2p network.
**Yet at the same time, it is important for censorship resistance to not have the protocol upgradable**

If third party service are made required, then well, Ethereum, will be decentralized at the currency level like Bitcoin while **Dapps like casinos or yield farming, will be fully permissionned have to register with authorities by paying an army of lawayers in order to be allowed to run which means, DeFi won’t be that much different than Fintechs on the tradionnal banking system which rent their computing hardware.**

I think keeping history on the p2p definitely worth the reduced transactions per seconds outputs or if we decide to behave like swift or MasterCard or Visa for being able to run as fast as them.

EIPS/eip-4444.md
ytrezq: > Not everyone will agree on what time it is, should we have a bit of a no-mans-land between the time clients are expected to stop _requesting_ ancient data and when clients are expected to stop _providing_ ancient data? This could be minutes/hours/days, just enough to deal with time synchronization issues.
> 
> Presumably, under such a regime any peer that requests data older than the longer of the two **SHOULD** be kicked. Meanwhile, clients **SHOULD** be programmed to never ask for anything older than the shorter of the two.

@MicahZoltu In my country, MasterCard and Visa while not blockchain related are required to keep past history of the last 10 years for law enforcement and they use a higher transaction throughout.

PR #4450, opened 2021-11-12
Add note about eip pronunciation

lightclient:
This will show them.


abcoathup:
Love it, but it doesn't cover the pronunciation of EIP,  eee-IP

PR #4476, opened 2021-11-22
Add note about PoW constant validation.

timbeiko:
Make it clear that the various 0 values in the EIP can be overridden by subsequent EIPs.


EIPS/eip-3675.md
```diff
@@ -98,6 +97,8 @@ Beginning with `TRANSITION_BLOCK`, the block validity conditions **MUST** be alt
 
 *Note*: Validity rules that are not specified in the list above **MUST** remain unchanged.
 
+*Note*: the value of `mixHash` is purposefully non-validated here to provide flexibility for future EIPs.
```
djrtwo, line 100: Future implies that it might be a while from now. Marginally prefer "subsequent"

```suggestion
*Note*: the value of `mixHash` is purposefully non-validated here to provide flexibility for subsequent EIPs.
```

MicahZoltu:
Note: This needs an approval by one of the authors to merge.

djrtwo (approved):

PR #4505, opened 2021-11-28
Make EIP 2364 & 2464 Final

timbeiko:
Last call period is over.


timbeiko:
@axic if you want to approve this, it'd be great because you pushed for having the devp2p link, with which I don't think Micah will make this final 😅

axic (changes requested):
I think the specification of 2364 should be improved. The part stating "Implement forbid" is a weird (it is more like a checklist), the "eth/63 should be kept alive" is more like a loose suggestion, but more importantly the "validate the remote forked, disconnecting at a detected incompatibility" part is I think too literal and not like a specification. As per my understanding this refers to the "Validation rules" part of EIP-2124.

PR #4510, opened 2021-11-30
Major update for EIP-3525

will-edge:
1. update the name of the standard to 'Semi-Fungible Token Standard'
2. Change the appearance of vNFT to EIP3525
3. Completing some metadata description


EIPS/eip-3525.md
```diff
@@ -257,7 +256,7 @@ The units level transfer has two types of interfaces, and both have safe and uns
 ```
 The main difference between the two kinds of interface is whether the application or the contract is responsible for determining/generating the target token ID in the transfer.
 
-Since partial transfer of a token will possibly result in new token id creation, it's important to give the implementing contract the ability to do that. On the other hand, since part of a token can be transferred to a token with the same slot, we want to keep the flexibility for dapps to determine whether to use this ability, resulting in less contract complexity and less gas consumption.
+Since transfering units of a token will possibly result in new token id creation, it's important to give the implementing contract the ability to do that. On the other hand, since part of a token can be transferred to a token with the same slot, we want to keep the flexibility for dapps to determine whether to use this ability, resulting in less contract complexity and less gas consumption.
```
MicahZoltu, line 259:
```suggestion
Since transferring units of a token will possibly result in new token id creation, it's important to give the implementing contract the ability to do that. On the other hand, since part of a token can be transferred to a token with the same slot, we want to keep the flexibility for dapps to determine whether to use this ability, resulting in less contract complexity and less gas consumption.
```

MicahZoltu:
Blockede by CI for:
> ./EIPS/eip-3525.md:259: transfering ==> transferring

EIPS/eip-3525.md
will-edge: Thanks for the suggestion!

PR #4511, opened 2021-11-30
EIP-4511: Execute Past Semantic Abort

tsutsu:
Proposes a "declared explicit intent to not standardize" a certain type of execution tracing, in order to leave EVM implementations free to optimize how often they detect certain error conditions.

### DIFF
--- a/EIPS/eip-4511.md
+++ b/EIPS/eip-4511.md
@@ -3,7 +3,7 @@ eip: 4511
 title: Execute Past Semantic Abort
 description: Allow clients to choose to continue to execute instructions after EVM bytecode interpretation would revert
 author: Levi Aul (@tsutsu)
-discussions-to: <URL>
+discussions-to: https://ethereum-magicians.org/t/eip-4511-execute-past-semantic-abort/7650
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-execute-past-semantic-abort.md
+++ b/EIPS/eip-execute-past-semantic-abort.md
@@ -1,0 +1,116 @@
+---
+eip: <to be assigned>
+title: Execute Past Semantic Abort
+description: Allow clients to choose to continue to execute instructions after EVM bytecode interpretation would revert
+author: Levi Aul (@tsutsu)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: Core
+created: 2021-11-30
+---
+
+## Abstract
+Rather than detecting/enforcing *interpreter-abort conditions* (Out-of-Gas, Stack Overflow/Underflow, etc.) on each EVM interpreter step, clients may choose to implement deferred detection/enforcement of these conditions at some higher-level instruction-sequence boundary point in the interpreter, under the condition that all transactions result in the same EVM post-transaction substate as they would with eager, per-interpreter-step *interpreter-abort condition* detection.
+
+In such clients, after the interpreter enters a state where it must revert execution, some unpredictable-but-bounded number of additional instructions from the execution frame will be executed. These instructions, as part of a reverted execution frame, will be *moot* (have no visible side-effects in the state.) However, the execution of these *moot* instructions may still be observed through client execution-tracing APIs.
+
+By removing these checks from the hot loop of EVM bytecode interpretation, these clients become amenable to a number of currently-impractical optimizations.
+
+## Motivation
+There are certain "interpreter abort" conditions within the EVM interpreter that are currently checked for on every interpretation step, such as Out-of-Gas, Stack Overflow, etc. When one of these conditions is found to apply, the currently-executing call frame is immediately escaped, with any state modifications made by that call frame and its descendants being reverted, just as if a user-triggered `REVERT` instruction were executed.
+
+EVM formal semantics *make no requirement* to perform these checks for each interpretation step. EVM transaction execution is effectively defined as a black box function upon the substate, where clients are free to implement the *means* by which the resulting substate is computed however they wish. Additionally, the revert on interpreter-abort "masks away" any information that a caller could use to determine what exactly the callee was doing when it aborted, by e.g. always using up all gas given to the callee.
+
+However, checking for interpreter-abort conditions on every interpretation step nevertheless seems to be a universal standard for Ethereum clients, perhaps out of mistaken assumption of this being a formal EVM requirement.
+
+Additionally, some ecosystem tooling may rely on cross-client predictable/deterministic output from client execution-tracing APIs, due to a similar mistaken assumption that the interpretation-step sequence for a given EVM transaction will always look the same.
+
+This implicit pseudo-requirement makes applying runtime optimizations to EVM interpretation, such as Just-In-Time compilation or threaded-code interpretation, mostly pointless, as either every instruction implementation needs to contain/generate redundant copies of these same checks, or trampolines to the interpreter to perform these checks need to be inserted between successive instructions.
+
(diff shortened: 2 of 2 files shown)


CryptoGGytrs52 (approved):

alita-moore:
Sorry I'll be looking into this issue later this week

alita-moore:
You need to get an eip number from a editor @axic @lightclient @MicahZoltu

EIPS/eip-execute-past-semantic-abort.md
```diff
@@ -0,0 +1,116 @@
+---
+eip: <to be assigned>
```
axic, line 2:
```suggestion
eip: 4511
```

EIPS/eip-execute-past-semantic-abort.md
```diff
@@ -0,0 +1,116 @@
...
+title: Execute Past Semantic Abort
+description: Allow clients to choose to continue to execute instructions after EVM bytecode interpretation would revert
+author: Levi Aul (@tsutsu)
+discussions-to: <URL>
```
axic, line 6: Need to create a discussion url.

axic:
Sorry I can't do a proper review now, but the above two are the minimum next steps.

EIPS/eip-execute-past-semantic-abort.md
MicahZoltu: Make sure to rename the file to `EIPS/eip-4511.md` as well.

EIPS/eip-execute-past-semantic-abort.md
tsutsu: Fixed

EIPS/eip-4511.md
```diff
@@ -0,0 +1,116 @@
...
+
+Because of this, BEAM VM languages are free to use *forward* branch instructions to implement non-loop control-structures, e.g. `if/else`, `case`, clause-head pattern matching, `goto fail`-type internal error handling, etc., without paying any bookkeeping overhead cost for these instructions. The instruction implementations for forward-branches remain just as simple as the instruction implementations for math ops *et al*, and so become amenable to optimization (e.g. branchless-code reduction) themselves.
+
+The EVM, as a trustless-multitenant runtime, cannot rely on the developer or compiler kindly limiting themselves to only deploying code that uses a subset of possible instruction sequences, the way BEAM compilers do. The EVM must ensure that Out-of-Gas conditions are checked at least once for each iteration of EVM-primitive branching loops.
+
```
MicahZoltu, line 56: Please follow the template, which means Specification comes before Rationale in this case.

https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md

MicahZoltu (changes requested):
Formatting generally looks fine aside from the order of Rationale and Specification.  Would be great to get @axic's eyes on this since they probably have a much better idea about the actual content than I, but for draft proper formatting is really the only requirement.

EIPS/eip-4511.md
tsutsu: I'm unclear on what to do in this case, where the Specification refers to information introduced in the Rationale. I originally had them in the opposite order, but the text was more confusing that way.

tsutsu: (Also, I would note that I ran `eipv` against this file and it passed; if this is a rule, shouldn't `eipv` enforce it?)

tsutsu:
One thing I'm not clear on myself is whether there's already existing terminology for some of the terms I introduced/made up, and whether that matters. E.g. is there an existing name for an "interpreter abort"? Seems like the kind of thing that should be in the yellow paper, but I couldn't find it (but maybe I didn't read closely enough.)

EIPS/eip-4511.md
MicahZoltu: The Specification should never depend on any other section.  The specification is the core of the entire document and should be an entirely self contained technical specification.  The specification should not be discussing the *why* of anything, it should merely assert "this is what you do if you want to implement this spec".  Any discussion of the why should happen in the rationale, and that can (and should) reference the specification.

One way to think of it is to write the specification first and imagine that the target audience of this document is some programmer that has been instructed to implement this spec and they don't care at all for the reasoning for anything, they only want to know what the requirements for spec compliance are and they want that information to be as easy to follow as possible.  Any side-tracked discussion about the *why* just makes it harder to implement as it is a distraction from what matters to that person.

MicahZoltu: > Also, I would note that I ran eipv against this file and it passed; if this is a rule, shouldn't eipv enforce it?

Yes, pull requests welcome.  😀

EIPS/eip-4511.md
```diff
@@ -0,0 +1,116 @@
...
+## Specification
+
+### Definitions
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
```
lightclient, line 60:
```suggestion
```
This is implied by EIP-1.

EIPS/eip-4511.md
```diff
@@ -0,0 +1,116 @@
...
+
+This implicit pseudo-requirement makes applying runtime optimizations to EVM interpretation, such as Just-In-Time compilation or threaded-code interpretation, mostly pointless, as either every instruction implementation needs to contain/generate redundant copies of these same checks, or trampolines to the interpreter to perform these checks need to be inserted between successive instructions.
+
+Explicitly allowing the deferral of interpreter-abort checks to execute only on arbitrary interpreter-chosen execution steps, removes the implicit constraint to include these checks "inside" or "between" instruction bodies, making these previously-pointless optimizations potentially worthwhile.
+
+## Rationale
+
+### Inspiration: Erlang BEAM VM
+
+This EIP is mostly inspired by looking at gas-tracking / Out-of-Gas detection specifically, and how other systems with similar constraints achieve efficient results.
```
lightclient, line 37:
```suggestion
This EIP is mostly inspired by looking at gas-tracking / out-of-gas detection specifically, and how other systems with similar constraints achieve efficient results.
```

EIPS/eip-4511.md
```diff
@@ -0,0 +1,116 @@
+---
+eip: 4511
+title: Execute Past Semantic Abort
+description: Allow clients to choose to continue to execute instructions after EVM bytecode interpretation would revert
+author: Levi Aul (@tsutsu)
+discussions-to: https://ethereum-magicians.org/t/eip-4511-execute-past-semantic-abort/7650
+status: Draft
+type: Standards Track
+category: Core
```
lightclient, line 9: I don't think this should be a core EIP as it doesn't affect consensus logic at all. I think either i) the information in the EIP should simply be applied as a PR into an existing client or possible ii) it is created as an `Informational` EIP.

lightclient (changes requested):

EIPS/eip-4511.md
MicahZoltu: We recently added this to the template, and I believe the consensus was that it should be included in every EIP?  @axic may remember better.  If we *do* want to say it is implied by EIP-1, then we should remove it from the template.

lightclient: Yep that's my mistake. Was thinking the RFC 2119 section in EIP-1 implied all EIPs were under that. We can leave as-is and I'll revert these requests.

PR #4513, opened 2021-12-01
Fixed typo on eip-721.md

sgc109:
Fixing typo for function name.
Actually I'm not sure if it is intended or just a typo, but there is no function with this name even though it is wrapped with backquotes.


lightclient:
@fulldecent would you mind providing some context on this PR?

fulldecent (approved):

fulldecent:
Thank you. This PR fixes a self-inconsistent reference to a `unsafeTransferFrom` function in the text. In an earlier draft, we were considering functions named like that. But when finalized, other names were chosen.

This PR corrects the descriptive text (not normative) to match references to the actual function names in the normative text.

PR #4529, opened 2021-12-07
fix: edit eip-1153.md to treat transient storage the same as persistent storage with regards to reverts

moodysalem:
IMO passing revert information from inner call frames is not as useful as being able to rely on the data written to transient storage as if it were regular storage

It can also be supported by not reverting and instead returning an error code

Tried to flesh out some of the time complexity stuff and it seems like it justifies a gas cost more similar to hot `SLOAD`/`SSTORE`

### DIFF
--- a/EIPS/eip-1153.md
+++ b/EIPS/eip-1153.md
@@ -2,35 +2,47 @@
 eip: 1153
 title: Transient storage opcodes
 author: Alexey Akhunov (@AlexeyAkhunov)
+description: Add opcodes for manipulating state that behaves identically to storage but is discarded after every transaction
 discussions-to: https://ethereum-magicians.org/t/eip-transient-storage-opcodes/553
-status: Stagnant
+status: Draft
 type: Standards Track
 category: Core
 created: 2018-06-15
 ---
 
-## Simple Summary
-Support for efficient transient storage in EVM. It is like regular storage (`SLOAD`/`SSTORE`), but with the lifetime limited to one Ethereum transaction.
-Notable use case is efficient reentrancy lock.
-
 ## Abstract
-This proposal introduces transient storage, which behaves similar to storage,
-but the updates will only persist within one Ethereum transaction. Transient storage is accessible to smart contracts via new opcodes: `TLOAD` and `TSTORE` (“T” stands for Transient).
+This proposal introduces transient storage opcodes, which manipulate state that behaves identically to storage, 
+but is discarded after every transaction. Transient storage is cheaper since it never requires disk access.
+Transient storage is accessible to smart contracts via 2 new opcodes: `TLOAD` and `TSTORE`, where “T” stands for "transient."
 
 ## Motivation
 Running a transaction in Ethereum can generate multiple nested frames of execution, each created by `CALL` (or similar) instructions.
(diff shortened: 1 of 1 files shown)


EIPS/eip-1153.md
```diff
@@ -1,36 +1,48 @@
...
 
-Language support could be added in relatively easy way. For example, in Solidity, a qualifier “transient” can be introduced (similar to the existing qualifiers “memory” and “storage”). Since addressing scheme of `TSTORE` and `TLOAD` is the same as for `SSTORE` and `SLOAD`, code generation routines that exist for storage variables, can be easily generalised to also support transient storage.
+Storage refunds accumulated due to inter frame communication are also limited to 20% of gas spent by a transaction as of [EIP-3529](./eip-3529.md) included in the London hard fork. 
+This greatly reduces the refunds for transiently-set storage slots in otherwise low-cost transactions. For example, in order to receive the full refund of one re-entrancy lock, 
+the transaction must spend ~80k gas on other operations.
+
+Language support could be added in relatively easy way. For example, in Solidity, a qualifier “transient” can be introduced 
+(similar to the existing qualifiers “memory” and “storage”, and Java's own `transient` keyword with a similar meaning).
```
AlexeyAkhunov, line 31: I would not bring Java here, as far as I remember `transient` in java simply means that a class member does not participate in the serialisation when built-in serialisation library is used. But I kind of see the similarity in the meaning :)

AlexeyAkhunov (approved):

EIPS/eip-1153.md
```diff
@@ -43,22 +55,36 @@ They use the same arguments on stack as `SLOAD` (`0x54`) and `SSTORE` (`0x55`).
...
+When transient storage is used in the context of `CALL` or `STATICCALL`, then the owning contract of the transient storage
+is the contract that is the target of the `CALL` or `STATICCALL` instruction (the callee).
 
-Transient storage does not interact with reverts or invalid transactions, that means if a frame reverts, its effects on the transient storage remain until the end of the transaction.
+If a frame reverts, all writes to transient storage within that frame are also reverted, as with persistent storage.
 
 ## Rationale
-There is a proposal to alleviate the cost of inter-frame communication by reducing the cost of `SSTORE` when it modifies the same item multiple times within the same transaction (EIP-1087).
+There is a proposal to alleviate the cost of inter-frame communication by reducing the cost of `SSTORE` 
+when it modifies the same item multiple times within the same transaction ([EIP-1087](./eip-1087.md)).
```
MicahZoltu, line 76: EIP-1087 is stagnant, which means this EIP should not reference it.  Generally speaking, you should only reference EIPs that are likely to become final *before* this EIP becomes final.

In this case, I recommend just giving a 2-line description of the alternative solution and then describe why you think this solution is superior.  No need to create a dependency on another EIP.

> Another option to solve a similar set of problems is X.  This EIP decided to do Y instead because of Z.

MicahZoltu (approved):
Approved, but the reference to a stagnant EIP will need to be removed before this moves on to the next stage (I recommend doing that ASAP).

EIPS/eip-1153.md
moodysalem: that's right yeah, but often in java a class would be serialized for writing to storage, so it did feel relevant in making the keyword more digestible

PR #4546, opened 2021-12-10
Wrapped deposit EIP

JChanceHud:
This EIP is designed to create a single contract that developers use when accepting deposits. Not sure if something similar exists.


EIPS/eip-wrapped_deposit.md
```diff
@@ -0,0 +1,137 @@
+---
+eip: <to be assigned>
```
wschwab, line 2:
```suggestion
eip: 4546
```

EIPS/eip-wrapped_deposit.md
```diff
@@ -0,0 +1,137 @@
...
+title: Wrapped Deposits
+description: A singleton contract for managing asset deposits.
+author: Justice Hudson (@jchancehud)
+discussions-to: https://ethereum-magicians.org/t/wrapped-deposit-contract-eip/7740
+status: Draft
+type: Informational
```
wschwab, line 8:
```suggestion
type: Standards Track
catrgory: ERC
```
This is an application-level standard, so it would belong in ERC (as opposed to Informational)

EIPS/eip-wrapped_deposit.md
```diff
@@ -0,0 +1,137 @@
...
+created: 2021-12-11
+---
+
+## Abstract
+The wrapped deposit contract handles deposits of assets (Ether, ERC20, ERC721) on behalf of a user. A user must only approve a spend limit once and then an asset may be deposited to any number of different applications that support deposits from the contract.
```
wschwab, line 13:
```suggestion
The wrapped deposit contract handles deposits of assets (Ether, [ERC-20](./eip-20.md), [ERC-721](./eip-721.md)) on behalf of a user. A user must only approve a spend limit once and then an asset may be deposited to any number of different applications that support deposits from the contract.
```
(EIP-1 specifies hyphenation - I generally forget about this myself, as other editors here can testify 🙂)

EIPS/eip-wrapped_deposit.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Motivation
+The current user flow for depositing assets in dapps is unnecessarily expensive and insecure. To deposit an ERC20 asset a user must either:
```
wschwab, line 16:
```suggestion
The current user flow for depositing assets in dapps is unnecessarily expensive and insecure. To deposit an ERC-20 asset a user must either:
```

EIPS/eip-wrapped_deposit.md
```diff
@@ -0,0 +1,137 @@
...
+
+  - send an approve transaction for the exact amount being sent, before making a deposit; for every deposit.
```
wschwab, line 18:
```suggestion
  - send an approve transaction for the exact amount being sent, before making a deposit, and then repeat this process for every subsequent deposit.
```

EIPS/eip-wrapped_deposit.md
```diff
@@ -0,0 +1,137 @@
...
+  - send an approve transaction for an infinite spend amount before making deposits.
+
+The first option is inconvenient, and expensive. The second option is insecure. Further, explaining approvals to new or non-technical users is confusing. This has to be done in _every_ dapp that supports ERC20 deposits.
```
wschwab, line 21:
```suggestion
The first option is inconvenient, and expensive. The second option is insecure. Further, explaining approvals to new or non-technical users is confusing. This has to be done in _every_ dapp that supports ERC-20 deposits.
```

wschwab: In addition, wouldn't the singleton face the same two issues? Either the singleton's user gives it an infinite approve, or faces approving each tx separately, no?

wschwab (commented):
Hi!

Thanks for submitting this EIP. I've left some comments, mainly just technical editing, below. I've also got some general questions about the EIP, though, which I'll leave here:
1) The depositor contract would be `msg.sender` on deposits to contracts, like you mention in the Security Considerations. Many deposit contracts issues rewards to depositors, and many check the depositor using `msg.sender`, which would result in the contract receiving rewards instead of the asset owner. Are you proposing a different check for deposit contracts that want to comply with this EIP? If so, I would make it explicit in Security Considerations.
2) You may wish to consider using a similar pattern to the specified receiver functions in ERC-721 and ERC-1155 instead of returning a bool.
3) If you are trying to be inclusive to the different forms of assets in the ecosystem at this point, I would recommend also including ERC-1155 NFTs.

Cheers!

JChanceHud:
Hi @wschwab thanks for the feedback!

> The depositor contract would be msg.sender on deposits to contracts, like you mention in the Security Considerations. Many deposit contracts issues rewards to depositors, and many check the depositor using msg.sender, which would result in the contract receiving rewards instead of the asset owner. Are you proposing a different check for deposit contracts that want to comply with this EIP? If so, I would make it explicit in Security Considerations.

I've interpreted this feedback in a few ways. The first is a question of backward compatibility. Existing contracts using `msg.sender` would not be able to use the proposed deposit contract. They would have to either update their implementation, or continue having users use the existing deposit flow. The second interpretation is "how will new contracts that offer dividends be built, and do old dividend contracts work?". The singleton contract works with most existing dividend implementations because it calls `transferFrom`. For example, the Compound contracts [rewards the `src`/`dist`, but not the `spender`](https://github.com/compound-finance/compound-protocol/blob/master/contracts/CToken.sol#L69). Finally, `Are you proposing a different check for deposit contracts that want to comply with this EIP`, yes, contracts that want to accept ERC-20 deposits should implement the `acceptERC20Deposit(address depositor, address token, uint amount)` and use the `depositor` for any balance changes.

There _may_ be some dividend contracts that reward the spender, not the src, in a `transferFrom`, in which case this contract should not be used. It might be worth researching other tokens and looking for places where this is the case, or at least providing a note like you said.

> You may wish to consider using a similar pattern to the specified receiver functions in ERC-721 and ERC-1155 instead of returning a bool.

Do you mean passing a bytes array as the final argument in the `acceptERC721Deposit` function? The `safeDepositERC721` function already accepts a bytes array and passes to the `onERC721Received` function. This should be called on the same contract as the `acceptDeposit`.

> If you are trying to be inclusive to the different forms of assets in the ecosystem at this point, I would recommend also including ERC-1155 NFTs.

Agree. I was planning to add support for these tokens as well.

--

I'm not sure if the wording or naming of this EIP is confusing. The deposit contract is never the receiver of assets (except momentarily in an Ether transfer). It calls into other contracts and uses `transferFrom`.

EIPS/eip-wrapped_deposit.md
JChanceHud: Sorta, the issue isn't having to do either of these, it's having to do either for _every_ contract I want to interact with. If there's a single contract that manages my deposits for many applications I'm more comfortable setting an infinite spend approval there, especially if I can easily digest the source (and it's not upgrade-able).

wschwab: I'm not sure I see how you get around the insecurity - if there's an infinite approve on your proposed contract, any layer that is interfaced with can still compromise the user's funds via the contract, no?

In addition, each asset would still need an infinite approve, so this doesn't reduce the amount of approvals to one.

JChanceHud: > any layer that is interfaced with can still compromise the user's funds via the contract

No, the contract is implemented only allowing funds to be deposited on behalf of `msg.sender`. This value changes with any external contract call. So, say I want to deposit to a compromised contract X. Contract X tries to re-enter the deposit contract and spend more of my tokens. Contract X just ends up spending it's own tokens.

The deposit contract should be safe even if the deposit receiver is malicious, this language should probably be in the EIP.

> each asset would still need an infinite approve, so this doesn't reduce the amount of approvals to one

Correct, one approval per asset.

lightclient (changes requested):
Hi @JChanceHud, could you please fill out the backwards compatibility section of the EIP and how this EIP would require contracts to upgrade and support the callback methods?

JChanceHud:
Thanks @lightclient, I added a note explaining backward compatibility a bit. Also added ERC-1155 token support.

EIPS/eip-4546.md
```diff
@@ -0,0 +1,174 @@
...
+User experience is improved because a simple global dapp can be implemented with the messaging: "enable token for use in other apps".
+
+## Backwards Compatibility
+
+This EIP is not backward compatible. Any contract planning to use this deposit system must implement specific functions to accept deposits. Existing contracts that are upgradeable can add support for this EIP retroactively by implementing one or more accept deposit functions.
+
+Upgraded contracts can allow deposits using both the old system (approving the contract itself) and the proposed deposit system to preserve existing approvals. New users should be prompted to use the proposed deposit system.
+
+## Reference Implementation
+https://github.com/jchancehud/wrapped-deposit
```
lightclient, line 75:
```suggestion
```

lightclient (approved):

PR #4557, opened 2021-12-13
EIP-4521: 721/20-compatible transfer

z0r0z:
adds suggestion and references for 721/20-compatible transfer

 - The build passes.

### DIFF
--- a/EIPS/eip-4521.md
+++ b/EIPS/eip-4521.md
@@ -3,7 +3,7 @@ eip: 4521
 title: 721/20-compatible transfer
 description: Recommends a simple extension to make NFTs compatible with apps and contracts that handle fungibles.
 author: Ross Campbell (@z0r0z)
-discussions-to: https://github.com/ethereum/EIPs/issues/4521
+discussions-to: https://ethereum-magicians.org/t/eip-4521-721-20-compatible-transfer/7903
 status: Draft
 type: Standards Track
 category: ERC
--- a/EIPS/eip-4521.md
+++ b/EIPS/eip-4521.md
@@ -1,0 +1,42 @@
+---
+eip: 4521
+title: 721/20-compatible transfer
+description: Recommends a simple extension to make NFTs compatible with apps and contracts that handle fungibles.
+author: Ross Campbell (@z0r0z)
+discussions-to: https://github.com/ethereum/EIPs/issues/4521
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-12-13
+requires 20, 721
+---
+
+## Abstract
+ERC-721, the popular standard for non-fungible tokens (NFTs), includes send functions, such as `transferFrom()` and `safeTransferFrom()`, but does not include a backwards-compatible `transfer()` found in fungible ERC-20 tokens. This standard provides references to add such a `transfer()`.
+
+## Motivation
+This standard proposes a simple extension to allow NFTs to work with contracts designed to manage ERC-20s and many consumer wallets which expect to be able to execute a token `transfer()`. For example, if an NFT is inadvertently sent to a contract that typically handles ERC-20, that NFT will be locked. It should also simplify the task for contract programmers if they can rely on `transfer()` to both handle ERC-20 and NFTs.
+
+## Specification
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+The interface for ERC-4521 `transfer()` MUST conform to ERC-20 and resulting transfers MUST fire the `Transfer` event as described in ERC-721.
+
+```sol
+function transfer(address to, uint256 tokenId) external returns (bool success);
+```
+
+## Rationale
+Replicating ERC-20 `transfer()` with just a minor change to accurately reflect that a unique `tokenId` rather than fungible sum is being sent is desirable for code simplicity and to make integration easier.
+
+## Backwards Compatibility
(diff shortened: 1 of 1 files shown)


EIPS/eip-4521.md
```diff
@@ -0,0 +1,42 @@
...
+eip: 4521
+title: 721/20-compatible transfer
+description: Recommends a simple extension to make NFTs compatible with apps and contracts that handle fungibles.
+author: Ross Campbell (@z0r0z)
+discussions-to: https://github.com/ethereum/EIPs/issues/4521
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-12-13
+requires 20, 721
```
wschwab, line 11:
```suggestion
requires: 721
```
I'd argue that this doesn't require ERC-20 since you don't need an ERC-20 interface for this, even if it's implementing a function based on ERC-20

wschwab (commented):
This EIP is well written, makes sense, and at least afaict, follows EIP-1 well (tho I'm a bit of a noob). Aside from the one correction I flagged, it looks good to go.

My only other point would be that you're setting yourself up to need to update what you've written in Security Considerations. If you know that the review period should finish soon, I don't personally think that's an issue, fwiw. If it could be some time, you might want to think about how to rewrite it.

ewinAT:
https://github.com/ethereum/EIPs/blob/master/eip-template.md

EIPS/eip-4521.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Rationale
+Replicating ERC-20 `transfer()` with just a minor change to accurately reflect that a unique `tokenId` rather than fungible sum is being sent is desirable for code simplicity and to make integration easier.
+
+## Backwards Compatibility
+There are no production instances of something like ERC-4521 that the author is aware of, but an implementation is [in review](https://github.com/Rari-Capital/solmate/pull/63) for inclusion in RariCapital’s SolMate library.
+
+## Reference Implementation
+A reference contract for an ERC-4521 NFT can be found [here](https://github.com/z0r0z/EIP-4521/blob/main/contracts/ERC4521.sol).
```
MicahZoltu, line 36: Don't include external links.  Instead, either inline the contract here or add it as an asset to `../assets/eip-4521/` and do a relative link to it here.

EIPS/eip-4521.md
```diff
@@ -0,0 +1,42 @@
...
+
+```sol
+function transfer(address to, uint256 tokenId) external returns (bool success);
+```
+
+## Rationale
+Replicating ERC-20 `transfer()` with just a minor change to accurately reflect that a unique `tokenId` rather than fungible sum is being sent is desirable for code simplicity and to make integration easier.
+
+## Backwards Compatibility
+There are no production instances of something like ERC-4521 that the author is aware of, but an implementation is [in review](https://github.com/Rari-Capital/solmate/pull/63) for inclusion in RariCapital’s SolMate library.
```
MicahZoltu, line 33: Don't link externally.  Recommend just removing this and saying something like "This EIP does not introduce any known backward compatibility issues".  If someone implements this prior to the standard becoming final, then the thing they implemented is not an EIP-4521 contract and thus doesn't need to be mentioned here IMO.

z0r0z: thanks - I made this update.

EIPS/eip-4521.md
z0r0z: thanks - I made this update.

EIPS/eip-4521.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Reference Implementation
+A reference contract for an ERC-4521 NFT can be found [here](https://github.com/z0r0z/EIP-4521/blob/main/contracts/ERC4521.sol).
+
+## Security Considerations
+Implementers must be sure to include the relevant return `bool` value for an ERC-4521 in order to conform with existing contracts that use ERC-20 interfaces, otherwise, NFTs may be locked unless a `safeTransfer` is used in such contracts.
+
```
ewinAT, line 60: contracts/LaunchpadNFTForBSW.sol

EIPS/eip-4521.md
```diff
@@ -0,0 +1,62 @@
+---
+eip: 4521
+title: 721/20-compatible transfer
+description: Recommends a simple extension to make NFTs compatible with apps and contracts that handle fungibles.
+author: Ross Campbell (@z0r0z)
+discussions-to: https://github.com/ethereum/EIPs/issues/4521
```
lightclient, line 6: Your `discussions-to` is 404ing. Can you create a topic on https://ethereum-magicians.org please?

EIPS/eip-4521.md
```diff
@@ -0,0 +1,62 @@
+---
+eip: 4521
+title: 721/20-compatible transfer
```
lightclient, line 3:
```suggestion
title: Add `transfer` function to NFT standards
```
This should also work with both 721 and 1155, correct?

lightclient (changes requested):

EIPS/eip-4521.md
wschwab: 1155 would be more complicated since it would require the insertion of another argument (since it has both `id` and `amount`), whereas 721s and 20s have the same number of arguments and congruent data types, though 20's uint256 is for the amount of tokens, whereas 721's is for `tokenId`

EIPS/eip-4521.md
z0r0z: yes @lightclient I have made a thread here: https://ethereum-magicians.org/t/eip-4521-721-20-compatible-transfer/7903

EIPS/eip-4521.md
lightclient: I see, in that case I still think the title needs to be revised to be more clear.

lightclient: I'll go ahead an merge this, but please revise the title before moving to review :)

lightclient (approved):

PR #4530, opened 2021-12-07
Update EIP to refer to Ethereum Mainnet, wording as per EIP-2228

fulldecent:
Author: @vbuterin

PR #4559, opened 2021-12-15
Update author format to preferred name + GitHub username format, Vitalik

fulldecent:
Updating EIPs from this author to use preferred author citation format.

REFERENCES:

- EIP template (uses name + GitHub username) https://github.com/ethereum/EIPs/blob/master/eip-template.md
- Invitation for PR from editor @axic  https://github.com/ethereum/EIPs/pull/4530#issuecomment-988406258

### DIFF
--- a/EIPS/eip-101.md
+++ b/EIPS/eip-101.md
@@ -1,7 +1,7 @@
 ---
 eip: 101
 title: Serenity Currency and Crypto Abstraction
-author: Vitalik Buterin <v@buterin.com>
+author: Vitalik Buterin (@vbuterin)
 status: Stagnant
 type: Standards Track
 category: Core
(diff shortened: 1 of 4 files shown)


axic (approved):
Fine from me, but probably @vbuterin should give a nod too.

axic:
@MicahZoltu @lightclient if you think this can merged as-is, please do. I think it should be fine as the author is identifyable, but this is a retroactive application of the EIP-1 rule if enforced by an editor.

lightclient (approved):

MicahZoltu:
Normally I am not a fan of changing the author line without author approval, but in this case the author is a high profile individual and their recent EIPs use the GitHub syntax suggesting that they are generally OK with that over their email address so I am willing to make an exception here.

PR #4560, opened 2021-12-15
Update author format to preferred name + GitHub username format, Gavin

fulldecent:
Updating EIPs from this author to use preferred author citation format.

REFERENCES:

- EIP template (uses name + GitHub username) https://github.com/ethereum/EIPs/blob/master/eip-template.md
- Invitation for PR from editor @axic  https://github.com/ethereum/EIPs/pull/4530#issuecomment-988406258

### DIFF
--- a/EIPS/eip-161.md
+++ b/EIPS/eip-161.md
@@ -1,7 +1,7 @@
 ---
 eip: 161
 title: State trie clearing (invariant-preserving alternative)
-author: Gavin Wood
+author: Gavin Wood (@gavofyork)
 type: Standards Track
 category: Core
 status: Final


axic (approved):
Fine from me, but probably @gavofyork should give a nod too.

axic:
@MicahZoltu @lightclient if you think this can merged as-is, please do. I think it should be fine as the author is identifyable, but this is a retroactive application of the EIP-1 rule if enforced by an editor.

MicahZoltu:
I am slightly less OK with this one versus the one for vbuterin because in the case of Vitalik, we have evidence that they are OK with their GitHub handle being associated with EIP authorship (based on their recent EIPs).  I don't believe there are any EIPs authored by Gavin that include their handle though, so we don't have a good precedent for acceptability.

That being said, it wouldn't be a significant doxxing of Gavin since as mentioned by @axic they are already well known and their GitHub handle is trivially easy to find (first result in an internet search for "Gavin Wood GitHub".

@lightclient do you have thoughts on the matter?

fulldecent:
^ this is all the right points to bring up for this consideration. 

Now it's up to editors to decide based on the evidence

axic:
It turns out this is the only EIP not having a username/email in the author line. If we can get this merged, we can have stronger rule enforcement: https://github.com/lightclient/eipv/issues/39

lightclient (approved):
I have no issues with merging this as-is.

MicahZoltu:
> It turns out this is the only EIP not having a username/email in the author line. If we can get this merged, we can have stronger rule enforcement: lightclient/eipv#39

I think this is compelling enough to convince me, when combined with support from other editors.

axic:
> > It turns out this is the only EIP not having a username/email in the author line. If we can get this merged, we can have stronger rule enforcement: https://github.com/ethereum/eipv/issues/39
> I think this is compelling enough to convince me, when combined with support from other editors.

Ah, we missed [EIP-158](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-158.md) from @vbuterin also lacking a handle.

PR #4564, opened 2021-12-15
Update EIP-1 to not use Issues for EIP number

poojaranjan:
Update EIP-1 to discourage use of Issues for "discussion-to" and "EIP number"

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -229,7 +229,7 @@ If the EIP isn't ready, the editor will send it back to the author for revision,
 
 Once the EIP is ready for the repository, the EIP editor will:
 
-- Assign an EIP number (generally the PR number)
+- Assign an EIP number (generally the PR number, but the decision is with the editors)
 
 - Merge the corresponding [pull request](https://github.com/ethereum/EIPs/pulls)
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-1.md
```diff
@@ -229,7 +229,7 @@ If the EIP isn't ready, the editor will send it back to the author for revision,
 
 Once the EIP is ready for the repository, the EIP editor will:
 
-- Assign an EIP number (generally the PR number or, if preferred by the author, the Issue # if there was discussion in [the Issues section of this repository](https://github.com/ethereum/EIPs/issues) about this EIP)
+- Assign an EIP number (generally the PR number)
```
axic, line 232:
```suggestion
- Assign an EIP number (generally the PR number, but the decision is with the editors)
```

MicahZoltu: I would also like to see the change proposed here by Alex, but if you are against it @poojaranjan we can merge this as is (it is an improvement even without Alex's suggestion) and then debate the addition of Alex's change to a separate PR.

MicahZoltu (approved):

EIPS/eip-1.md
poojaranjan: I am good!

lightclient (approved):
LGTM

PR #4572, opened 2021-12-16
Fix use of EIP-165 supportsInterface in EIP-2981

frangio:
There is a section in EIP-2981 that describes how to use EIP-165, but it's incorrect. Before checking for the relevant interface id, one has to first check that EIP-165 is supported (i.e. that `supportsInterface(0x01ffc9a7) = true` and `supportsInterface(0xffffffff) = false`).

I've replaced the instructions with a reference to the EIP where that is properly described.

### DIFF
--- a/EIPS/eip-2981.md
+++ b/EIPS/eip-2981.md
@@ -130,14 +130,7 @@ constructor (string memory name, string memory symbol, string memory baseURI) {
 
 #### Checking if the NFT being sold on your marketplace implemented royalties
 
-```solidity  
-bytes4 private constant _INTERFACE_ID_ERC2981 = 0x2a55205a;
-
-function checkRoyalties(address _contract) internal returns (bool) {
-    (bool success) = IERC165(_contract).supportsInterface(_INTERFACE_ID_ERC2981);
-    return success;
- }
-```
+To determine if an NFT implements royalties, integrators should check for the EIP-165 interface id `0x2a55205a` using the procedure specified in said EIP: ["How to Detect if a Contract Implements any Given Interface"](https://eips.ethereum.org/EIPS/eip-165#how-to-detect-if-a-contract-implements-any-given-interface).
 
 ## Rationale
 


cliffhall (commented):
The code example does exactly what the prose that replaces it instructs, but with the merit of being directly usable via copy and paste. The new paragraph would be fine to **add** before or after the code example, but I think clarity is negatively impacted by removing the example altogether.

frangio:
It differs in that it doesn't check for ERC165, and also that it calls with more than 30k gas (though to be fair it's arguable whether this is against the spec).

>since this is a just a view method, and an attempt to cast a bad contract to 165 and call one of its methods would revert

This isn't necessarily the case and that's the reason why checking for ERC165 is not this straightforward.

I appreciate the fact that this snippet can be copy-pasted, but as it is it's not really following the ERC165 spec.

cliffhall:
> > since this is a just a view method, and an attempt to cast a bad contract to 165 and call one of its methods would revert
> 
> This isn't necessarily the case and that's the reason why checking for ERC165 is not this straightforward.

In what case does casting a contract that doesn't support ERC-165 to the ERC-165 interface and calling `supportsInterface(bytes4 _interfaceId)` not revert?

frangio:
> also that it returns `false` for the arbitrary interfaceId' `0xffffffff` and that callers can only assume true support after those two static calls

Yes this is part of my point, and is what I thought you meant by "check for ERC165".

> their example code only does the first call

The `ERC165Query` contract in the spec does `noThrowCall(_contract, InvalidID)`, where `InvalidID` is `0xffffffff`.

> The extra check for a contract that falsely claims to support ERC-165 seems excessive.

It's not about a contract falsely claiming this. It isn't explained in the EIP but as I understand it, the procedure is meant for a situation where the `supportsInterface` function selector clashes with another function in the contract. Solidity will gladly cast any other non-zero value into a boolean and a contract would misinterpret it as support for an interface.

I understand the procedure may seem excessive, but that's not a good reason to ignore the spec. If EIP-2981 uses ERC165 interfaces it should stick to the specification.

lightclient:
@frangio I don't think this is a change that we can merge since it changes the functional standard of how to determine if a contract implements EIP-2981. I think we need to leave it as-is.

frangio:
@lightclient But do we agree that the current text of the standard contains an incorrect implementation of EIP-165? If we agree on that, how do we justify leaving it as is?

How should an integrator check if this EIP is supported: using the procedure as specified in EIP-165 or as in EIP-2981? They can lead to different results.

I see this as a bug in the specification. Is there no room at all for corrections once an EIP becomes Final?

MicahZoltu:
The Examples shouldn't be in the `## Specification` section IMO (better for `## Reference Implementation`).  In such a case, this would be a non-normative change to the EIP.

cliffhall:
In practice, what is the probability that a contract that **does not support** ERC-165 responds `true` to a function call of `supportsInterface(_INTERFACE_ID_ERC2981)`?

frangio:
Sorry I just don't think this is the place for those discussions. ERC-165 is already defined like that.

cliffhall:
> Sorry I just don't think this is the place for those discussions. ERC-165 is already defined like that.

Agreed. 

More to the point, the existing example is helpful if not complete. The developer can quickly understand the basic logic for checking if the contract supports this spec. Combining your paragraph with the example code would be instructive, making it clear that adding a test for ERC-165 itself is more involved, with a link to the appropriate reference material.

Would that solve the problem adequately?

frangio:
I'm satisfied with adding the code back with a disclaimer that it's only provided as an example and that a proper implementation should use the full procedure defined in EIP-165.

@cliffhall you don't show up as an EIP author though, can one of the authors with approval permissions mention if they're ok with this change?

frangio:
This needs an EIP author to chime in.

Issue #4526, opened 2021-12-06
Only run github actions on the canonical EIPs repository


SamWilsn:
As @gcolvin noted in the Discord, the bots for the EIPs process will run inside third-party forks if Actions are enabled. This can be kind of annoying if you're working on a draft or proposal that isn't ready for review yet.

I suggest adding something like:

```yaml
if: github.repository_owner == 'ethereum'
```

to the jobs.

alita-moore:
Send easy enough

mryalamanchi:
@alita-moore did you pick this task or can I move forward with it?

alita-moore:
Yea that's fine @mryalamanchi

mryalamanchi:
This issue is resolved. Can we close this @SamWilsn @alita-moore?

alita-moore:
Yep

PR #4595, opened 2021-12-20
Patch for preventing GH actions/bots to run on the forked branches

mryalamanchi:
This patch adds a condition to check and skip if the repository isn't 'ethereum/eips', solving the problem for forked branches.
Issue Ref: https://github.com/ethereum/EIPs/issues/4526
Sample test run: https://github.com/mryalamanchi/EIPs/runs/4579216649?check_suite_focus=true

Kindly review and merge

Note: The `if` checks don't allow for case-insensitive, as mentioned in the docs. 
So will have to keep an eye out if this can cause any problems later.

### DIFF
--- a/.github/workflows/auto-merge-bot.yml
+++ b/.github/workflows/auto-merge-bot.yml
@@ -4,7 +4,7 @@ jobs:
   auto_merge_bot:
     runs-on: ubuntu-latest
     name: EIP Auto-Merge Bot
-    if: github.event.pull_request.draft == false
+    if: github.event.pull_request.draft == false && github.repository == 'ethereum/eips'
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


alita-moore (approved):
It seems a bit overkill to me for us to mark workflows, and not just the main bots. But it's not a big deal to change and achieves the goal so 🤷‍♀️

lightclient:
@alita-moore what way would be less overkill here?

alita-moore:
Only add this to the relevant bots. I.e. the auto merge bot, auto stagnate bot

mryalamanchi:
@alita-moore do you suggest I push a commit only for the relevant bots?

alita-moore:
No, that's okay

azboyzranch (commented):
>.

PR #4601, opened 2021-12-22
EIP-165: Removes gas references.

MicahZoltu:
Gas prices can (and frequently do) change, and no contract or specification can make any guarantees on how much an operation will cost in the future.  Saying "this will cost less than 30,000 gas" is an assertion that this specification **CANNOT** make.

This is a normative change to a final EIP, but I think in this case it is acceptable because the EIP currently makes an assertion about the universe that it cannot reasonably make and it should be fixed to no longer make such an assertion.

### DIFF
--- a/EIPS/eip-165.md
+++ b/EIPS/eip-165.md
@@ -62,8 +62,7 @@ pragma solidity ^0.4.20;
 interface ERC165 {
     /// @notice Query if a contract implements an interface
     /// @param interfaceID The interface identifier, as specified in ERC-165
-    /// @dev Interface identification is specified in ERC-165. This function
-    ///  uses less than 30,000 gas.
+    /// @dev Interface identification is specified in ERC-165.
     /// @return `true` if the contract implements `interfaceID` and
     ///  `interfaceID` is not 0xffffffff, `false` otherwise
     function supportsInterface(bytes4 interfaceID) external view returns (bool);
@@ -85,7 +84,7 @@ Implementation note, there are several logical ways to implement this function. 
 
 ### How to Detect if a Contract Implements ERC-165
 
-1. The source contract makes a `STATICCALL` to the destination address with input data: `0x01ffc9a701ffc9a700000000000000000000000000000000000000000000000000000000` and gas 30,000. This corresponds to `contract.supportsInterface(0x01ffc9a7)`.
+1. The source contract makes a `STATICCALL` to the destination address with input data: `0x01ffc9a701ffc9a700000000000000000000000000000000000000000000000000000000`. This corresponds to `contract.supportsInterface(0x01ffc9a7)`.
 2. If the call fails or return false, the destination contract does not implement ERC-165.
 3. If the call returns true, a second call is made with input data `0x01ffc9a7ffffffff00000000000000000000000000000000000000000000000000000000`.
 4. If the second call fails or returns true, the destination contract does not implement ERC-165.
@@ -148,7 +147,7 @@ contract ERC165Query {
                 mstore(add(x, 0x04), _interfaceId) // Place first argument directly next to signature
 
                 success := staticcall(
-                                    30000,         // 30k gas
+                                    gas(),         // remaining gas
                                     _contract,     // To addr
                                     x,             // Inputs are stored at location x
                                     0x24,          // Inputs are 36 bytes long


gcolvin:
> This is a normative change to a final EIP, but I think in this case it is acceptable because the EIP currently makes an assertion about the universe that it cannot reasonably make and it should be fixed to no longer make such an assertion.

Can this be handled as an errata? 

And I can't tell, but is there a technical need for _some_  constant her?  It seems the 30,000 wasn't gratuitous.

axic:
~~First: IIRC the reason for this is that ERC-165 was made before the 63/64th rule was introduced.~~ 

Second: I think their intention was that ERC-165 targets finish within a determined limit. I do not think such a low limit provides any good utility as per your reasoning.

However I am against this change in the current form, and I'll explain why. 

There are two users of this ERC: 1) implementers of `supportsInterface` providers; 2) implementers of `supportsInterface` consumers. The (new) implementers should be aware of this limit so that they can support a wide range of consumers. Consumers on the contrary should probably consider doing what you suggest.

MicahZoltu:
> is there a technical need for some constant her? It seems the 30,000 wasn't gratuitous.

I'm against any constant/limit, because whatever we choose will be entirely arbitrary and may break in the future as gas costs for operations change over time.  Perhaps there is a future where the operations required to implement this function simply cannot happen in under 30,000 gas, or a contract can do what they need in 30,000 gas today but then a hardfork comes along and now they cannot and everything breaks.  Broadly speaking, I think it is a very bad idea to bake any assumptions about operational gas cost into any contract, and I definitely don't think we should be including such things in specifications (especially ones as widely used as 165).

> However I am against this change in the current form, and I'll explain why.

Is there another form of this change that you would be more agreeable to?

axic:
> > However I am against this change in the current form, and I'll explain why.
>
> Is there another form of this change that you would be more agreeable to?

Ultimately I think the decision is up to the authors of the EIP. I do not agree to a plain "remove all references to 30000" as a solution. I'd be positive if you can change the wording so that it leaves an explainer that many ERC-165 consumers expect a 30000 gas limit, and so new providers should aim to be within that limit, while new consumers should not impose such a limit.

gcolvin:
> Ultimately I think the decision is up to the authors of the EIP. I do not agree to a plain "remove all references to 30000" as a solution. I'd be positive if you can change the wording so that it leaves an explainer that many ERC-165 consumers expect a 30000 gas limit, and so new providers should aim to be within that limit, while new consumers should   impose such a limit.

I agree this is probably the right approach.  It probably belongs in the Rationale.  We need to hear from the authors. I'm guessing that @chriseth and @Arachnid are most recent champions.

fulldecent:
Everything discussed in this thread was known, debated and re-debated thoroughly during the ERC-165 standardization process.

Here is a summary of the arguments and the conclusion:

- Minimizing implementation gas usage is important
- High implementation gas usage will create DoS vulnerabilities, and on-chain consumers of ERC-165 need to limit attacks
- (Prediction from 2018) wallet implementations will continue to suck in the future and so we will continue to have ERC-165 clients on-chain (entirely true prediction)
- Some implementations on-chain have hard-coded 30k gas
- If there is a hardfork that changes gas costs, they must cite the impact on ERC-165 as a backward compatibility issue
- If you want to remove the 30k gas limit, that's great, but please do it as a new EIP

MicahZoltu:
> If there is a hardfork that changes gas costs, they must cite the impact on ERC-165 as a backward compatibility issue

It has been stated by the core devs repeatedly that gas costs can and will change and that contracts **SHOULD NOT** make any assumptions about gas costs.  While a backward compatibility note in a core EIP would be reasonable, it shouldn't be expected that core devs will avoid a gas cost change because it will break EIP-165.

fulldecent:
No such limitations on future innovation is expected.

Only requested is a backwards compatibiilty note, and hopefully a two-week, public comment period where I can give feedback and get a heads up before decisions are made.

Same situation as when ERC-165 was finalized.

wschwab:
Is there any way a caveat could be added explicitly that gas prices can change and that the EIP reflects the system as of the date it was written? I have personally benefited from the way EIP-165 was written, as it has enabled me to make better decisions about whether to use the mapping structure or not (for example), and assume others have and will benefit from it too.

MicahZoltu:
> Is there any way a caveat could be added explicitly that gas prices can change and that the EIP reflects the system as of the date it was written?

I think that would be an improvement over the current state of the EIP.  I still prefer just removing it all together, but I'm open to a compromise on the subject.

fulldecent:
There needs to retain a note in the text of the EIP which acknowledges that some existing implementations require a 30,000 limit.

I'm happy to see this change happen

PR #4603, opened 2021-12-23
Fix ERC721 Metadata JSON Schema descriptions

titusz:


fulldecent (approved):
Looks good, thank you

lightclient:
I am inclined to merge this, but I don't think I fully understand the ramifications of have multiple versions of the ERC-721 schema floating about. Only the description is being modified to read better, the typing is consistent. @axic @gcolvin @wschwab - any thoughts here?

PR #4609, opened 2021-12-27
Fix pseudocode on EIP-4337 paymaster reputation status

antico5:
The logic on the snippet is wrong (i.e. having opsSeen > 0 and opsIncluded = 0 would return "OK" every time).


yoavw (approved):
Approved.  Thanks for catching this.

PR #4616, opened 2021-12-29
Mark EIP1967 as Last Call with deadline 2022-01-16

spalladino:
Moving EIP1967 to last call. Let me know if there are any other changes needed before we can do this!


lightclient (changes requested):
A few things should be addressed before moving to last call
* Remove the external links from the EIP
* Update the reference to "EIP1822" and "EIP897" as "EIP-N". Also, EIP citations generally have the title following the number not before with the EIP in parentheses.
* Respond to @fulldecent's [comment](https://ethereum-magicians.org/t/eip-1967-standard-proxy-storage-slots/3185/14?u=matt) regarding the undefined storage slot

spalladino:
Hey @lightclient, thanks for reviewing!

- I've added a screenshot of how the proxy looks on a block explorer, so as not to depend on the external link to Etherscan. I don't think there are other external links where it's not clear their content from the text of the EIP. Should I remove every external link from the EIP? Or is it ok to keep them? I couldn't find any guidelines on external links on EIP-1, and EIP-1 itself has links to external sites, hence the question.
- Updated, I'm using exactly the same format as shown on EIP-1.
- Added the missing code from the reference implementation. I hadn't originally since it's not needed for understanding the code, but I'm ok including it for the sake of completeness.

fulldecent:
Please pause pending https://github.com/ethereum/solidity/issues/12547 so that we can link to Solidity project commitments rather than public comments from developers on that project.

wschwab:
@fulldecent I'm not sure I understand what you would need to see changed here - could you be a bit more specific?

fulldecent:
This EIP relies on a commitment of API stability made by the Solidity project.

This commitment is published on Twitter on the personal account of one of the developers of the project.

Let's wait until this commitment is published on the Solidity project's official website and then link to the commitment there instead.

lightclient (changes requested):
Please link to EIPs using the format `./eip-xxxx.md`. I realize there is an issue with the `eips.ethereum.org` website which has an old format listed. If you look in the github repository, you'll see the above format specified: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md#linking-to-other-eips

Bezzat:
Anlamiyorum lütfen destek olun parami almam için

5 Nis 2022 Sal 08:03 tarihinde github-actions[bot] ***@***.***>
şunu yazdı:

spalladino:
Hey @fulldecent @lightclient we've updated the link to the Solidity docs confirming no changes in storage layout, and changed the format of links to other EIPs. Let me know if there are any other changes pending, or if we are good to move to Last Call!

EIPS/eip-1967.md
```diff
@@ -4,7 +4,8 @@ title: Standard Proxy Storage Slots
 description: Standardise where proxies store the address of the logic contract they delegate to, as well as other proxy-specific information.
 author: Santiago Palladino (@spalladino)
 discussions-to: https://ethereum-magicians.org/t/eip-1967-standard-proxy-storage-slots/3185
-status: Review
+status: Last Call
+last-call-deadline: 2022-01-16
```
SamWilsn, line 8: I know that we editors have been holding up the process quite a bit, but this'll need to be 14 days from when this pull request is merged :sweat_smile:

EIPS/eip-1967.md
```diff
@@ -20,14 +21,20 @@ Delegating proxies are widely in use, as a means to both support upgrades and re
...
+A classic example of this is a block explorer. Here, the end user wants to interact with the underlying logic contract and not the proxy itself. Having a common way to retrieve the logic contract address from a proxy allows a block explorer to show the ABI of the logic contract and not that of the proxy. The explorer checks the storage of the contract at the distinguished slots to determine if it is indeed a proxy, in which case it shows information on both the proxy and the logic contract. As an example, this is how `0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48` is [shown on Etherscan](https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48#readProxyContract):
```
SamWilsn, line 24: You'll need to remove the external link:

```suggestion
A classic example of this is a block explorer. Here, the end user wants to interact with the underlying logic contract and not the proxy itself. Having a common way to retrieve the logic contract address from a proxy allows a block explorer to show the ABI of the logic contract and not that of the proxy. The explorer checks the storage of the contract at the distinguished slots to determine if it is indeed a proxy, in which case it shows information on both the proxy and the logic contract. As an example, this is how `0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48` is shown on Etherscan:
```

EIPS/eip-1967.md
```diff
@@ -20,14 +21,20 @@ Delegating proxies are widely in use, as a means to both support upgrades and re
...
+[Solidity](https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html) maps variables to storage based on the order in which they were declared, after the contract inheritance chain is linearized: the first variable is assigned the first slot, and so on. The exception are values in dynamic arrays and mappings, which are stored in the hash of the concatenation of the key and the storage slot. The Solidity development team has confirmed that the storage layout is to be preserved among new versions:
```
SamWilsn, line 33:
```suggestion
Solidity maps variables to storage based on the order in which they were declared, after the contract inheritance chain is linearized: the first variable is assigned the first slot, and so on. The exception are values in dynamic arrays and mappings, which are stored in the hash of the concatenation of the key and the storage slot. The Solidity development team has confirmed that the storage layout is to be preserved among new versions: 
```

EIPS/eip-1967.md
```diff
@@ -20,14 +21,20 @@ Delegating proxies are widely in use, as a means to both support upgrades and re
...
+
+> The layout of state variables in storage is considered to be part of the external interface of Solidity due to the fact that storage pointers can be passed to libraries. This means that any change to the rules outlined in this section is considered a breaking change of the language and due to its critical nature should be considered very carefully before being executed. In the event of such a breaking change, we would want to release a compatibility mode in which the compiler would generate bytecode supporting the old layout.
+
+Vyper seems to [follow the same strategy as Solidity](https://github.com/ethereum/vyper/issues/769). Note that contracts written in other languages, or directly in assembly, may incur in clashes.
```
SamWilsn, line 37:
```suggestion
Vyper seems to follow the same strategy as Solidity. Note that contracts written in other languages, or directly in assembly, may incur in clashes.
```

SamWilsn (changes requested):

fulldecent:
Thank you. I support to proceed with Last Call as-is or with the changes proposed [immediately above](https://github.com/ethereum/EIPs/pull/4616#pullrequestreview-934064524).

spalladino:
@SamWilsn thanks for the suggestions, I've just applied them. I also set the last call deadline to 17 days from today, so we have a few days buffer in case the PR is not merged today, and it also matches the 3-year anniversary since the EIP was created.

SamWilsn (approved):

PR #4626, opened 2022-01-04
EIP-4626: Tokenized Vault Standard

Joeysantoro:

 - The build passes.


eip-4626.md
```diff
@@ -0,0 +1,152 @@
+# Yield Bearing Vault Standard
+
+Author: Joey Santoro, t11s, Jet Jadeja
```
Joeysantoro, line 3:
```suggestion
Author: Joey Santoro <joeysantoro>, t11s <transmissions11>, Jet Jadeja <JetJadeja>
```

eip-4626.md
```diff
@@ -0,0 +1,155 @@
+# Yield Bearing Vault Standard
+
+eip: <to be assigned>
```
MicahZoltu, line 3:
```suggestion
eip: 4626
```

eip-4626.md
```diff
@@ -0,0 +1,155 @@
+# Yield Bearing Vault Standard
+
```
MicahZoltu, line 2: Please follow the template found at https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md

The formatting on this header is incorrect.

eip-4626.md
```diff
@@ -0,0 +1,155 @@
...
+eip: 4626
+title: Yield Bearing Vault Standard
+description: A standard for yield bearing vaults
+Author: Joey Santoro (@joeysantoro), t11s (@transmissions11), Jet Jadeja (@JetJadeja)
+Status: Draft
+Type: Standards Track
+Category: ERC
+Created: 2021-12-22
```
Joeysantoro, line 10:
```suggestion
eip: 4626
title: Yield Bearing Vault Standard
description: A standard for yield bearing vaults.
author: Author: Joey Santoro (@joeysantoro), t11s (@transmissions11), Jet Jadeja (@JetJadeja)
discussions-to: <URL>
status: Draft
type: Standards Track
category: ERC
created: 2021-12-22
```

eip-4626.md
```diff
@@ -0,0 +1,156 @@
+# Yield Bearing Vault Standard
+
+eip: 4626
+title: Yield Bearing Vault Standard
+description: A standard for yield bearing vaults.
+author: Author: Joey Santoro (@joeysantoro), t11s (@transmissions11), Jet Jadeja (@JetJadeja)
+discussions-to: <URL>
```
Joeysantoro, line 7:
```suggestion
discussions-to: https://github.com/ethereum/EIPs/pull/4626
```

eip-4626.md
Joeysantoro: done

eip-4626.md
```diff
@@ -0,0 +1,156 @@
+# Yield Bearing Vault Standard
+
```
Joeysantoro, line 2:
```suggestion
---
```

eip-4626.md
```diff
@@ -0,0 +1,156 @@
...
+eip: 4626
+title: Yield Bearing Vault Standard
+description: A standard for yield bearing vaults.
+author: Author: Joey Santoro (@joeysantoro), t11s (@transmissions11), Jet Jadeja (@JetJadeja)
+discussions-to: https://github.com/ethereum/EIPs/pull/4626
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-12-22
+
```
Joeysantoro, line 12:
```suggestion
---

# Yield Bearing Vault Standard

```

eip-4626.md
```diff
@@ -0,0 +1,158 @@
...
+`event Withdraw(address indexed _owner, addres indexed _to, uint256 _value)`
+
+Where `_from` is the user who triggered the withdrawal and held `_value` underlying tokens in the vault, and `_to` is the user who received the withdrawn tokens.
+
+
+## Rationale
+
+The vault interface is designed to be optimized for minimal implementation and integration logic while maintaining flexibility for both parties. Details such as accounting and allocation of deposited tokens are intentionally not specified, as vaults are expected to be treated as black boxes on-chain and inspected off-chain before use.
+
```
wschwab, line 145: There should be a "Backwards Compatability" section here, I think (someone will correct me if it isn't strictly necessary - even if it is, I would recommend spending a paragraph on what this means for platforms that are not following this EIP)

wschwab (commented):
I only had time for a quick passthrough rn, so I left some comments on some stuff that stuck out to me, mainly just technical stuff about how EIPs need to be structured. Don't get disheartened - especially the first time submitting an EIP it's hard to keep track of all the stuff that needs to be there and how it needs to be formatted. It might take a couple of rounds of iterations before it gets merged in as a Draft, but this looks like a good candidate imho.

eip-4626.md
```diff
@@ -0,0 +1,158 @@
...
+`function withdraw(address _to, uint256 _value) public returns (uint256 _shares)`
+
+Withdraws `_value` tokens from the vault and transfers them to `_to`. 
+
+MAY return a pro-rata ownership `_shares` value corresponding to `_value`,  if not MUST return `0`.
+
+
+#### totalHoldings
+`function totalHoldings() public view returns (uint256)`
```
bsamuels453, line 51: There ought to be some documentation here or elsewhere about whether this view should rely on external state. One 
 can imagine a naiive implementer making totalHoldings dependent on some kind of external oracle without realizing the security implications.

Joeysantoro: The interface is unopinionated in the same way totalSupply is unopinionated for ERC20. We expect most implementations to use a shares based model. It will always be up to the integration layer to control for "safe" vaults based on their own security model

eip-4626.md
```diff
@@ -0,0 +1,158 @@
+---
+eip: 4626
+title: Yield Bearing Vault Standard
+description: A standard for yield bearing vaults.
+author: Author: Joey Santoro (@joeysantoro), t11s (@transmissions11), Jet Jadeja (@JetJadeja)
+discussions-to: https://github.com/ethereum/EIPs/pull/4626
```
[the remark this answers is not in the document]
Joeysantoro: Opened a post but it needs mod approval

eip-4626.md
```diff
@@ -0,0 +1,158 @@
...
+## Motivation
+
+Yield bearing vaults have a lack of standardization leading to diverse implementation details. Some various examples include lending markets (Compound, Aave, Fuse), aggregators (Yearn, Rari Vaults, Idle), and intrinsically interest bearing tokens (xSushi). This makes integration difficult at the aggregator or plugin layer for protocols which need to conform to many standards. This forces each protocol to implement their own adapters which are error prone and waste development resources.
+
+A standard for yield bearing vaults will allow for a similar cambrian explosion to ERC-20, unlocking access to yield in a variety of applications with little specialized effort from developers.
+
+
+## Specification
+
+### Methods
```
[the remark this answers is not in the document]
Joeysantoro: I can add one later sure

eip-4626.md
Joeysantoro: We elected not to add one because its a pretty standalone standard and there isn't much to say. Its backwards compatible with ERC20 and all implementations which don't use this interface should either update or lose out on the network effects

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+`exchangeRate() * totalSupply()` MUST equal `totalHoldings()`.
+
+OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
+
+#### baseUnit
+
+`function baseUnit() public view returns(uint256)`
+
+The decimal scalar for vault shares and operations involving `exchangeRate()`.
+
```
transmissions11, line 112:
```suggestion

SHOULD equal `10 ** underlying().decimals()`

If pro-rata shares are represented as an ERC20 token, `baseUnit()` MUST be equal to equal to `10 ** decimals`

```

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
```
transmissions11, line 113:
```suggestion
#### decimals

`function decimals() public view returns (uint256)`

The decimal scalar for vault shares and operations involving `exchangeRate()`.

OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
```

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+
+MAY return a pro-rata ownership `_shares` value corresponding `_value`, if not MUST return `0`.
+
+OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
+
+#### exchangeRate
+
+`function exchangeRate() public view returns (uint256)`
+
+The amount of underlying tokens one `baseUnit` of vault shares is redeemable for.
```
transmissions11, line 99:
```suggestion
The amount of underlying tokens one "full unit" (`10 ** decimals()`) of Vault shares shares is redeemable for. 
```

transmissions11:
we need to be more clear about the decimal scalars for each of these functions. like is `exchangeRate` is scaled by `decimals` but `totalHoldings()` scaled by `underlying().decimals()`?

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+
+OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
+
+#### exchangeRate
+
+`function exchangeRate() public view returns (uint256)`
+
+The amount of underlying tokens one `baseUnit` of vault shares is redeemable for. 
+
+e.g. `_shares * exchangeRate() / baseUnit() = _value`.
```
transmissions11, line 101:
```suggestion
e.g. `_shares * exchangeRate() / (10 ** decimals()) = _value`.
```

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+
+`exchangeRate() * totalSupply()` MUST equal `totalHoldings()`.
```
transmissions11, line 103: we need to divide by the right scalar cuz totalHoldings decimals is scaled by underlying decimals

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+`event Deposit(address indexed _from, addres indexed _to, uint256 _value)`
+
+Where `_from` is the user who triggered the deposit and approved `_value` underlying tokens to the vault, and `_to` is the user who is able to withdraw the deposited tokens. 
+
+
+#### Withdraw
+
+MUST be emitted when tokens are withdrawn from the vault by a depositor.
+
+`event Withdraw(address indexed _owner, addres indexed _to, uint256 _value)`
```
transmissions11, line 131:
```suggestion
`event Withdraw(address indexed _from, addres indexed _to, uint256 _value)`
```

we use `_from` below

eip-4626.md
Joeysantoro: Great call, should we remove this MUST condition or add the decimals scalar? Leaning former

transmissions11: im not too opinionated either way tbh, just want to make sure we clarify the decimals one way or another

eip-4626.md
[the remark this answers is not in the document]
lightclient: Approved.

eip-4626.md
lightclient: Backwards Compatibility is a required section per EIP-1, so you will need to include something short even if it's just stating that there are no compatibility issues.

eip-4626.md
```diff
@@ -0,0 +1,152 @@
+---
+eip: 4626
+title: Yield Bearing Vault Standard
+description: A standard for yield bearing vaults.
+author: Joey Santoro (@joeysantoro), t11s (@transmissions11), Jet Jadeja (@JetJadeja)
+discussions-to: https://github.com/ethereum/EIPs/pull/4626
```
lightclient, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900
```

lightclient (changes requested):
Overall looks pretty good. Please fill out a `Backwards Compatibility` section and rerequest a review. Thanks!

transmissions11:
do we also want to specify logic about how deposits work? alberto was proposing a skimbatch compatible pattern but we should probably at the very least specify that it always supports using erc20 allowances

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+
+OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
+
+#### redeem
+
+`function redeem(address _to, uint256 _shares) public returns (uint256 _value)`
+
+Redeems a specific number of `_shares` for underlying tokens and transfers them to `_to`. 
+
+MAY return a pro-rata ownership `_shares` value corresponding `_value`, if not MUST return `0`.
```
Clearwood, line 91: I think the documentation here is reversed, the return parameter here is rightfully value

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### balanceOf
+
+`function balanceOf(address _owner) public view returns (uint256)`
+
+Returns the total amount of vault shares the `_owner` currently has.
+
+OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
+
+#### redeem
```
JTraversa, line 85: Given compound's `mint`/`redeemUnderlying` methods have been replaced by `deposit`/`withdraw` in this EIP, it might make sense to retain parallel structure, and replace `redeem` with `withdrawShares`.

I think this makes the method's use case a bit more clear at a glance, though `redeem` might have a nice ring to it... ¯\_(ツ)_/¯

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+OPTIONAL - This method is only needed for vaults that implement a pro-rata share mechanism for deposits.
+
+
+### Events
+
+#### Deposit
+
+MUST be emitted when tokens are deposited into the vault.
+
+`event Deposit(address indexed _from, addres indexed _to, uint256 _value)`
```
tonypiano, line 122: typo here `addres indexed _to`

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+
+Where `_from` is the user who triggered the deposit and approved `_value` underlying tokens to the vault, and `_to` is the user who is able to withdraw the deposited tokens. 
+
+
+#### Withdraw
+
+MUST be emitted when tokens are withdrawn from the vault by a depositor.
+
+`event Withdraw(address indexed _from, addres indexed _to, uint256 _value)`
```
tonypiano, line 131: typo here as well `addres indexed _to`

eip-4626.md
transmissions11: but this is not `redeemUnderlying` it's `redeem`, which has similar behavior even in compound:

Screen Shot 2022-01-04 at 9 28 55 PM

eip-4626.md
transmissions11: yup

eip-4626.md
alcueca: It's easier and more general to use a `calculateShares(uint256 underlyingAmount) returns (uint256 shares amount)`. No need to think about scalars, and supports non-linear exchange rates with precision.

Likewise, you should have the symmetrical `calculateUnderlying(uint256 sharesAmount) returns (uint256 underlyingAmount)`.

This pattern has been used in the yield oracles with great success, and partially in the lido staked ether contract where you have also symmetrical functions for asset conversions.

eip-4626.md
JTraversa: Yup its a 1-1 of `redeem` from compound! 

But if there is no `redeemUnderlying`, only `withdraw` and `withdraw`+`redeem` do the same thing, just one is scaled for shares, it might make sense to associate the two. (in the same way compound associates `redeem` and `redeemUnderlying`)

Again, I think `redeem` sounds nicer, `withdrawShares` describes what this does in context of other methods. Very minor hairs being split.

transmissions11: fair points

> Very minor hairs being split.

thats t11s for ya ;P

eip-4626.md
alcueca: To give another example, if a vault is implemented with a fee on deposits or withdrawals, it won't be possible for an interesting contract to know the deposit or withdrawal output, without specific knowledge of those fees.

By putting any conversion logic behind the `calculate` functions, vaults with non-linear math or fees can be integrated seamlessly.

Joeysantoro: I like this suggestion

JTraversa: Id say in addition to. 

Like @alcueca's saying, a `calculateX` is needed to ensure your own internal bookkeeping is in line for a single specific `deposit`/`withdraw` that could include unknown fees, but for more generalized bookkeeping that isnt tied to a single tx, it's still ideal to record the `exchangeRate`.

e.g. In our contracts, we benchmark `exchangeRate` at maturity (potentially regardless of any deposits/withdrawals) and use this as a point to begin divesting any additional yield from YT holders, back to PT holders.

alcueca: I would suggest this as instead `exchangeRate`. From the point of an integrating contract, you are not usually interested in the exchange rate itself, but in a specific conversion of a for b.

If you would still be interested in the exchange rate as an integrating contract, you can obtain it by passing one full unit as the amount parameter.

Furthermore, it removes the constraint on how tokens are converted for shares, which the EIP forces to a linear, no-fee formula (otherwise `exchangeRate` is true only for one full unit).

If still unconvinced, I'd like to hear the arguments in favour of a `exchangeRate` variable over two symmetrical functions.

alcueca:
`deposit`, `withdraw` and `redeem` might seem like familiar options for function names, but there is a gap.

 - `deposit`: User chooses an amount of underlying. Underlying goes into the vault. User mints shares (or some kind of receipt).
 - `withdraw`: User chooses an amount of underlying. Underlying gets out from the vault. User burns shares (or some kind of receipt).
 - `redeem`: User chooses an amount of shares (or some kind of receipt). Underlying gets out from the vault. User burns shares (or receipt).

There is a function missing:
 - `x`: User chooses an amount of shares (or some kind of receipt). Underlying goes into the vault. User mints shares (or receipt).

If this sounds a bit strange, it is easier to see it by comparing to a fully-fledged AMM. In Pool.sol from YieldSpace, there are four functions, which can be mapped onto a generic yield-bearing vault:
 - sellBase: Choose an exact amount of asset A to give away for an undetermined amount of asset B
 - buyBase: Choose an exact amount of asset A to receive in exchange for an undetermined amount of asset B
 - sellFYToken: Choose an exact amount of asset B to give away for an undetermined amount of asset A
 - buyFYToken: Choose an exact amount of asset B to receive in exchange for an undetermined amount of asset A

Similar functionality is present in other AMMs like Uniswap, usually merged into one sell and one buy functions because both assets in a regular AMM pair are identical.

In the case of a generic Yield-Bearing Vault, it might help to organize the asset transfer functions in the same way, to ensure that the features provide complete and consistent coverage of use cases.

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-12-22
+---
+
+## Abstract
+
+The following standard allows for the implementation of a standard API for yield bearing vaults within smart contracts. This standard provides basic functionality for depositing and withdrawing tokens and reading balances with an optional extension for tokenized vaults using ERC-20.
```
alcueca, line 15: Why is the ERC20 tokenized vault optional?

I understand that there might be an interest to describe the Yield-Bearing Vault specification separately, to be able to combine it later on with ERC-20, ERC-721, ERC-1155, or some other custom tokenization standard.

However, the EIP does a halfway job of dissociating itself from ERC-20, stating it as optional, which turns it confusing.

I would suggest to either remove the optionality of tokenized vaults, and specify a pure Yield-Bearing Vault EIP, or make ERC-4626 Vaults to inherit from ERC-20 as a requirement. Either case would make the EIP clearer and more robust.

My preference would be to make ERC-4626 Yield-Bearing Vaults to compulsorily inherit from ERC-20, and to provide a complete standard that also matches closely the existing implementations. I think it will be quite hard to specify a Yield-Bearing Vault with no relationship to any tokenizing standard, yet with enough meaning to be useful.

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+MAY return a pro-rata ownership `_shares` value corresponding `_value`, if not MUST return `0`.
+
+MAY represent `_shares` using internal accounting or an ERC-20 token. 
+
+If pro-rata shares ownership is implemented, the vault SHOULD implement `balanceOf`, `redeem`, `totalSupply` and `exchangeRate`.
+
+
+#### withdraw
+
+`function withdraw(address _to, uint256 _value) public returns (uint256 _shares)`
```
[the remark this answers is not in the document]
Joeysantoro: I am not as opinionated on the naming conventions as @transmissions11 and am open to reordering or making `redeem` required which is the shares based model. I do like the symmetry of deposit/withdraw both using the same underlying token accounting and having redeem or withdrawShares use the shares model. I'm also comfortable having withdraw be the shares and withdrawUnderlying be for underlying. Combined with @alcueca's feedback, I believe the standard can be somewhat more opinionated on some of these details.

eip-4626.md
Joeysantoro: I'm writing a generic ERC4626 router and all of your advice has become very appropriate as I encounter some of the nuances of the implementation details between vaults of different types

eip-4626.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Methods
+
+#### deposit
+
+`function deposit(address _to, uint256 _value) public returns (uint256 _shares)`
+
+Deposits `_value` tokens into the vault and grants ownership of them to `_to`.
+
+MAY return a pro-rata ownership `_shares` value corresponding `_value`, if not MUST return `0`.
```
mrenoon, line 34: As it stands now, the inputs to an EIP-4626 pool must be the **underlying**, which is the unit for valuing the pool and the user's position.

I would like to make a case to allow the inputs to be something else, called the **deposit token**. There are a few cases where the **deposit token** could be different from the **underlying**:
1. A vault could take in aUSDC instead of USDC, due to certain design choices. The **underlying** unit of accounting here, is still USDC
2. A wrapper contract could be made to take in gOHM or xSUSHI, use OHM/SUSHI as the underlying unit of accounting, and essentially produce a wrapped_gOhm that adheres to EIP-4626
3. A vault could take in a UniswapV2 LP token, while choosing to use liquidity (sqrt(token1Amount*token2Amount)) as the underlying unit of accounting.

To allow for these use cases, we just need to make one additional view function and three semantic changes:
* Additional function:
     * `depositToken() public view returns (address)`
     * Meaning: the token for depositing and withdrawing, which could be the same as `underlying()`
* Semantic change for depositing:
    * Rename variables from 
       ```
       function deposit(address _to, uint256 _value) public returns (uint256 _shares)
       ```
       to
       ```
       function deposit(address _to, uint256 _amount) public returns (uint256 _shares)
       ```
    * Meaning: deposits `_amount` of deposit tokens, which is worth `_value = _shares * exchangeRate()` in value into the vault and grants ownership of the value to `_to`
* Semantic change for redeeming (shares):
    * Rename variables from 
       ```
       function redeem(address _to, uint256 _shares) public returns (uint256 _value)
       ```
       to
       ```
       function redeem(address _to, uint256 _shares) public returns (uint256 _amount)
       ```
    * Meaning: Redeems a specific number of `_shares` for deposit tokens and transfers them to `_to`. The value of the redeemed tokens will be `_shares * exchangeRate()`
* Semantic change for withdrawing (deposit tokens):
    * Rename 
       ```
       function withdraw(address _to, uint256 _value) public returns (uint256 _shares)
       ```
       to
       ```
       function withdraw(address _to, uint256 _amount) public returns (uint256 _shares)
       ```
    * Meaning: withdraws `_amout` deposit tokens from the vault and transfers them to `_to`.


**In short**: These semantic changes do not alter how things currently work but will make the EIP more general and future-proof.

transmissions11: there is no need for this, as what the vault does with the deposited tokens internally is completely irrelevant to this ERC interface. 

underlying signifies what comes in and out of the vault and what totalHoldings() and such is measured in, we do not specify what happens to deposited underlying tokens. you can impl the vaults you proposed without additional metadata.

eip-4626.md
transmissions11: i am ok with enforcing tokenization if consensus is there but we strayed away from specifying it cuz there are many different approaches to doing it beyond a simple non-rebasing erc20 and we want the core deposit and withdraw logic to be adopted widely

eip-4626.md
mrenoon: Let's say we want to implement this vault A:
- User deposits cUSDC
- The vault claims COMP over time and convert to more cUSDC

Option 1: If we use the current standard:
- underlying() = cUSDC
- exchangeRate() = rate of how much more cUSDC a user will get, due to COMP
- underlyingBalanceOf(user) = cUSDC balance

Option 2: if we apply the proposed semantic changes here, it can be:
- underlying() = USDC
- exchangeRate() = rate of how much more USDC user is getting, from both Compound's rate and claiming COMP in this vault
- underlyingBalanceOf(user) = USDC-equivalent balance of the user

If we take the perspective of let's say a UI to show the details of this vault, it makes more sense to show the user balance in USDC using underlyingBalanceOf(), and easier to compare this vault with other vaults with the same USDC underlying.
Another example is from the perspective of a protocol to tokenise the yield of this vault with respect to USDC, where Option 2 makes a lot more sense as well.

eip-4626.md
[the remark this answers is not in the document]
UncleGrandpa925: Agree on this, I think the norm of most protocols is to use `value` for depositing & `shares` for withdrawing, so making the default `withdraw` function to use `shares` will minimise unexpected surprises to devs. 
Since the interface is for a vault, having an additional `withdrawValue` is maybe necessary? But I think the devs can easily do some math and use the core `withdraw` function if necessary. This is probably better than forcing the function in the interface.

eip-4626.md
alcueca: I think that you can require the vault to be an ERC20 in which balances represent ownership of the vault (shares), without specifying how shares are allocated.

eip-4626.md
alcueca: I think that explicitly specifying a multi-layer approach adds complexity to the standard, which should be kept as simple as possible. Not all existing vaults will be immediately compliant and that must be accepted.

As such, I don't think that a `depositToken()` method should be added.

On specifying inputs based on `value` or `shares`, that I agree for the sake of completeness.

1. Deposit exact `value` underlying, receive some `shares`
2. Receive exact `shares`, deposit underlying for some `value`
3. Withdraw exact `shares`, receive underlying for some `value`
4. Receive exact `value` of underlying, by withdrawing some `shares`.

This follows the pattern seen in AMMs that also deal with two tokens (in this case the underlying and the vault tokens).

Externally, the vault does the same job of exchanging between the two, and it is irrelevant the internal mechanism used to determine the exchange rate.

I'm not arguing to provide an interface identical to any existing AMM, but a familiar set of methods, proven to work in an equivalent use case, would definitely help with integration.

eip-4626.md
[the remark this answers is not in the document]
alcueca: Please, see above. I think that both the `withdraw` and `deposit` features should accept both `value` or `shares` as inputs.

eip-4626.md
alcueca: Why `totalHoldings` and not `totalUnderlying`, `underlyingHoldings`, or `underlyingReserves`. I think this is the only instance where "Holdings" is used to refer to underlying.

Joeysantoro: I like totalUnderlying as well. @transmissions11 @JetJadeja wdyt?

eip-4626.md
Joeysantoro: In favor, will enforce ERC20 compatibility

eip-4626.md
Joeysantoro: I think it makes a lot of sense to add a "mint" method which does `3. Withdraw exact shares, receive underlying for some value` @transmissions11 @JetJadeja wdyt?

eip-4626.md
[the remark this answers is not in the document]
GalloDaSballo: Happy to agree to disagree, I'm not making any statement about what I think the code should look like.

The code that is being used in production by the majority of protocols looks like the following
`function deposit(uint256 amount)`
`function withdraw(uint256 shares)`

Happy to go into detail as to why that is the case, but this interface has survived over hundreds of forks on all EVM compatible chains, and as such shows wisdom in its simplicity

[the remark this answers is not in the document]
WyseNynja: Also, withdrawing a fixed amount of underlying (rather than shares) from vaults always ends up leaving some dust behind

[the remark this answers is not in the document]
transmissions11: hence why we have a redeem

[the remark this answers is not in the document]
transmissions11: withdraw standing for shares is flat out illogical. you don't "withdraw" a check at the bank, or "withdraw" a winning lottery ticket.

eip-4626.md
```diff
@@ -0,0 +1,148 @@
...
+`function redeemFrom(address _from, address _to, uint256 _shares) public returns (uint256 _value)`
+
+Redeems a specific number of `_shares` from owner `_from` for underlying tokens and transfers them to `_to`. 
+
+MUST share the same authorization mechanism as transferFrom.
+
+MUST emit the `Withdraw` event.
+
+#### totalHoldings
+`function totalHoldings() public view returns (uint256)`
```
kbrizzle, line 75: Can we make this `totalUnderlying()` to conform with `balanceOfUnderlying()`, `calculateUnderlying()`, `underlying()` etc

eip-4626.md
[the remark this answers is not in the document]
kbrizzle: All of the cToken-based protocols (Compound, Rari, Cream, etc) use `mint(uint amount)`, `redeem(uint shares)`, `redeemUnderlying(uint amount)` so I think whichever way is chosen there's going to be a group of existing protocols that aren't natively compatible.

eip-4626.md
davidcallanan: @mrenoon Perhaps whatever is using this interface can go through two interfaces to determine the value in USDC. First determine SHARES/aUSDC value via ERC-4626 and then determine aUSDC/USDC via say a DEX or another ERC-4626. This is kind of composing two "valuation" interfaces which have "calculate" functions. Could that be standardized independent of the rest of EIP-4626? Both ERC-4626 and a DEX could implement this "valuation" interface. Then the logic which composes two valuation interfaces could be wrapped up into its own valuation interface, achieving your goal of going through this one interface which calculates the SHARES/USDC valuation directly.

eip-4626.md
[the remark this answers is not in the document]
GalloDaSballo: > withdraw standing for shares is flat out illogical. you don't "withdraw" a check at the bank, or "withdraw" a winning lottery ticket. you **redeem** them. you withdraw $100

You can't withdraw to someone else either, the bank analogy falls flat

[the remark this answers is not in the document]
GalloDaSballo: > withdraw standing for shares is flat out illogical. you don't "withdraw" a check at the bank, or "withdraw" a winning lottery ticket. you **redeem** them. you withdraw $100

As for withdrawing $100 dollars, the reason why fork over fork never killed the `withdraw(shares)` is the level of complexity or close to impossibility to withdraw all shares.

Illustrated by literally the first withdraw I found on Rari's smart contract

https://etherscan.io/tx/0xcada09b35d35aaee76fdaf2a00b8f275b07ced7eb8c80cf48cc5dac42c4832ea

Screenshot 2022-01-07 at 15 40 02

Screenshot 2022-01-07 at 15 40 14

Did the depositor really want to keep those 0.02 tokens? Or is the implementation that forces them into something they don't want?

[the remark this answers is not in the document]
davidcallanan: @GalloDaSballo

**Definitions**
(Google) withdraw: remove or take away (something) from a particular place or position
(Merriam Webster) withdraw: to remove (money) from a place of deposit

It doesn't seem to say anything about where the money ends up (the destination), so I think you **can** withdraw into *someone else's* account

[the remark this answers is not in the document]
davidcallanan: > the reason why fork over fork never killed the `withdraw(shares)` is the level of complexity or close to impossibility to withdraw all shares

Regardless, the term "redeem" is perhaps better-fitted than "withdraw".

eip-4626.md
Joeysantoro: Yes we will change this

Joeysantoro, line 75:
```suggestion
#### totalUnderlying
`function totalUnderlying() public view returns (uint256)`
```

eip-4626.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Specification
+
+All tokenized vaults MUST implement ERC20. If a vault is to be non-transferrable, it MAY revert on calls to transfer or transferFrom. The ERC20 operations balanceOf, transfer, totalSupply, etc. operate on the vault "shares" which represent ownership in the underlying.
+
+### Methods
+
+#### deposit
+
+`function deposit(address _to, uint256 _value) public returns (uint256 _shares)`
```
fubuloubu, line 32: Any way that address to send shares to can be the second parameter? In yearn vaults, we have them in this order because in decreasing order of usage:
- deposit everything for yourself
- deposit a specific amount for yourself
- deposit a specific amount and give to someone else

Depositing everything to another person is rare

fubuloubu: Same for `withdraw`

transmissions11: cool with me, makes sense in languages (vyper) that have optional args

Joeysantoro: I have a decent preference for the other order given it follows ERC20 spec. Wouldn't yearn need to still make adapters for the remaining functions anyway? What is the gain by switching if its only limited backward compatibility

fubuloubu: As a side note, the likelihood any protocol makes an adapter is quite low, I didn't make the suggestion for backwards compatibility reasons.

Anyways, I don't think ERC20 plays a huge role here in this spec in terms of driving the interface of these operations. Merely consider that the addition of override functions might be a good way to reduce gas costs for common operations, if the ordering is correct.

So adding:
```solidity
function deposit() public returns (uint256 _shares) // Deposit all to self
function deposit(uint256 _value) public returns (uint256 _shares) // Deposit some to self
function deposit(uint256 _value, address _to) public returns (uint256 _shares) // Deposit some to other
```

As another side note... I really wish people would start writing ERC specifications in YAML per [EIP-2069](https://eips.ethereum.org/EIPS/eip-2069) but that never really took off :disappointed:

eip-4626.md
```diff
@@ -0,0 +1,148 @@
+---
+eip: 4626
+title: Tokenized Vault Standard
+description: A standard for tokenized vaults with a single underlying ERC20 token.
```
lightclient, line 4:
```suggestion
description: A standard for tokenized vaults with a single underlying ERC-20 token.
```

eip-4626.md
```diff
@@ -0,0 +1,148 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900
+status: Draft
+type: Standards Track
+category: ERC
+created: 2021-12-22
+---
+
+## Abstract
+
+The following standard allows for the implementation of a standard API for tokenized vaults with a single underlying ERC20 within smart contracts. This standard is an extension on the ERC20 token that provides basic functionality for depositing and withdrawing tokens and reading balances.
```
lightclient, line 15:
```suggestion
The following standard allows for the implementation of a standard API for tokenized vaults with a single underlying [ERC-20](./eip-20.md) within smart contracts. This standard is an extension on the ERC-20 token that provides basic functionality for depositing and withdrawing tokens and reading balances.
```

eip-4626.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Rationale
+
+The vault interface is designed to be optimized for minimal implementation and integration logic while maintaining flexibility for both parties. Details such as accounting and allocation of deposited tokens are intentionally not specified, as vaults are expected to be treated as black boxes on-chain and inspected off-chain before use.
+
+ERC20 is forced because implementation details like token approval and balance calculation directly carry over to the shares accounting. This standardization makes the vaults immediately compatible with all ERC20 use cases in addition to ERC4626.
+
+## Backwards Compatibility
+
+ERC4626 is fully backward compatible with the ERC20 standard and has no known compatibility issues with other standards. For production implementations of vaults which do not use ERC4626, wrapper adapters can be developed and used.
```
lightclient, line 134:
```suggestion
ERC-4626 is fully backward compatible with the ERC-20 standard and has no known compatibility issues with other standards. For production implementations of vaults which do not use ERC-4626, wrapper adapters can be developed and used.
```

lightclient (changes requested):

lightclient (approved):

lightclient:
Can the discussions here please be moved to the discussion thread? https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900

aallamaa:
Would it be possible to add a deposit with permit (ERC2612 permit) function “depositWithPermit” in the standard ?

MicahZoltu:
Please use the discussions-to URL at the top of the EIP to discuss this.

PR #4649, opened 2022-01-09
New EIP request

jeyakatsa:
Created this EIP after making some discoveries while working on the [Altair Minimal Light Client Server for Teku(Consensys) for the Ethereum 2 Beacon Chain](https://github.com/jeyakatsa/Altair----Minimal-Light-Client-Prototype).

### DIFF
--- a/EIPS/eip-XXXX.md
+++ b/EIPS/eip-XXXX.md
@@ -1,19 +1,38 @@
+---
+Eip: XXXX *Number to be determined by Editor*
+Title: ERC Token in Multiple Languages
+Author: Jeremy Akatsa (@jeyakatsa)
+Status: Draft
+Type: Standards Track
+Category: ERC
+Created: 1/8/2022
+---
+
 ## Simple Summary
 An ERC token/currency capable of being created in multiple languages (Java, Rust, Python, Go, etc).
 
 ## Abstract
-Today, hundreds of thousands of websites are created everyday in many languages. What if the Ethereum ecosystem could also do the same, but with currencies and tokens?
+Today, millions of websites are created every week in many languages. What if the Ethereum ecosystem could also do the same, but with currencies and tokens?
 
 ## Motivation
-While building the *[Light Client Infrastructure for Teku](https://github.com/jeyakatsa/teku/tree/master/light-client),* a discovery was made. Clients serve as arbiters or "bridges" to the Ethereum main chain. Such clients offer differing languages that connect with the transactions offered on Ethereum, thus concluding if said clients can offer a base layer capable of inferring transactions from Ethereum's base layer, can tokens created in said languages also offer the same results?
+While building the *[Light Client Infrastructure for the Teku Client (for Ethereum 2)](https://github.com/jeyakatsa/teku/tree/master/light-client),* a discovery was made; clients serve as arbiters or "bridges" to the Ethereum main chain.
+| ETH2 Client   | Language     | Team                |
+|:--------------|:------------ |:------------------- |
+| Teku          | Java         | Consensys           |
+| Lighthouse    | Rust         | Sigma Prime         |
+| Trinity       | Python       | Ethereum Foundation |
+| Prysm         | Go           | Prysmatic Labs      |
+
+Such clients offer differing languages that connect with the transactions offered on Ethereum, thus concluding if said clients can offer a base layer capable of inferring transactions from Ethereum's base layer, can tokens created in said languages also offer the same results?
 
 ## Specification
+>The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
 
 #### Starting Point:
-Creating a library of creating smart-contracts in these languages: Java, Rust, Python & Go, beginning with Java. The first agenda being creating a simple smart-contract out of Java by using Solidity as the base layer.
+Creating a library of implementing smart-contracts in these languages: Java, Rust, Python & Go, beginning with Java. The first agenda being creating a simple smart-contract out of Java by using Solidity as the base layer.
 
 ##### Smart-Contract Storage Example in *Solidity*:
-```
+```solidity
 pragma solidity >=0.4.16 <0.9.0;
 
 contract SimpleStorage {
@@ -30,7 +49,7 @@ contract SimpleStorage {
 ```
 
 ##### Smart-Contract Storage Example in *Java*:
-```
+```java
 public class SimpleStorage {
     private Uint256 storedData;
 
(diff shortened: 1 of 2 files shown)


EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
+---
+Eip: XXXX *Number to be determined by Editor*
+Title: ERC Token in Multiple Languages
+Author: Jeremy Akatsa (@jeyakatsa)
+Status: Draft
+Type: Standards Track
+Category: ERC
+Created: 1/8/2022
```
SamWilsn, line 8:
```suggestion
created: 2022/01/08
```

Needs to be in ISO 8601 (yyyy-mm-dd) format.

EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
...
+While building the *[Light Client Infrastructure for the Teku Client (for Ethereum 2)](https://github.com/jeyakatsa/teku/tree/master/light-client),* a discovery was made; clients serve as arbiters or "bridges" to the Ethereum main chain.
+| ETH2 Client   | Language     | Team                |
+|:--------------|:------------ |:------------------- |
+| Teku          | Java         | Consensys           |
+| Lighthouse    | Rust         | Sigma Prime         |
+| Trinity       | Python       | Ethereum Foundation |
+| Prysm         | Go           | Prysmatic Labs      |
+
+Such clients offer differing languages that connect with the transactions offered on Ethereum, thus concluding if said clients can offer a base layer capable of inferring transactions from Ethereum's base layer, can tokens created in said languages also offer the same results?
+
```
SamWilsn, line 24: This section seems like content for the motivation section, rather than an abstract.

Instead, the abstract should be a technical summary of the EIP itself. Someone should be able to read only the abstract to get the gist of what this specification does.

EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
+---
+Eip: XXXX *Number to be determined by Editor*
+Title: ERC Token in Multiple Languages
+Author: Jeremy Akatsa (@jeyakatsa)
+Status: Draft
+Type: Standards Track
+Category: ERC
+Created: 1/8/2022
+---
```
SamWilsn, line 9: For consistency with other EIPs, the header keys should be lower case (`title` instead of `Title`.)

Before merging, you'll need to add:

 - `description`
 - `discussions-to`

EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Simple Summary
+An ERC token/currency capable of being created in multiple languages (Java, Rust, Python, Go, etc).
```
SamWilsn, line 12: I think your simple summary is a bit too vague. Keep it short, but try to be a bit more specific, and touch on exactly what you want to accomplish.

EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
...
+
+Such clients offer differing languages that connect with the transactions offered on Ethereum, thus concluding if said clients can offer a base layer capable of inferring transactions from Ethereum's base layer, can tokens created in said languages also offer the same results?
+
+## Motivation
+Today, millions of websites are created every week in many languages. What if the Ethereum ecosystem could also do the same, but with currencies and tokens?
+
+## Specification
+>The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+#### Starting Point:
```
SamWilsn, line 31: I can't tell if you're specifying a way to define smart contracts in languages other than Solidity, or specifying a standard way to represent interactions with a smart contract in other languages.

EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
...
+Creating a library of implementing smart-contracts in these languages: Java, Rust, Python & Go, beginning with Java. The first agenda being creating a simple smart-contract out of Java by using Solidity as the base layer.
```
SamWilsn, line 32: Try not to write the EIP as a set of instructions, but rather as a document describing the final outcome.

EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public Uint256 getStoredData () {
+        return storedData;
+    }
+}
+```
+
+## Rationale
+![](https://imgur.com/hs8WlUw.jpg)
```
SamWilsn, line 67: Please upload external auxiliary files in the assets directory (see [EIP-1](https://eips.ethereum.org/EIPS/eip-1#auxiliary-files)).

EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Minting
+Overall scope/goal for this proposal is for these tokens to be minted as easy as NFTs. This could be accomplished simply from a "button" within a dApp/App that mints a new token, *other simpler possibilities open to discussion*. The Total & Circulating Supply of these tokens are to be minted by dApps/Apps integrated within the logic-base implemented by the developer.
+
+### Exchanging
+These New-ERC Tokens are meant to act as ***"Personal Currencies"*** so in order for them to be successfully exchanged, they will need the consent of both parties *(like trading one NFT for another OR specifically speaking, trading 1 Yuan for 1 Dollar, or 50 Yuans for 1 Dollar, etc)*.
```
SamWilsn, line 73: The rationale section should explain choices made in the specification section, not introduce new specifications.

SamWilsn (commented):
I would suggest splitting this EIP into two parts, one about the new personal token standard and one about writing tokens in other languages.

That said, overall, I think these ideas need to be developed further before coming to the EIP stage. Might be best to start some discussions on the R&D discord or the ECH forum and seek input from other developers.

lightclient (commented):
@jeyakatsa it feels like this idea should live in the Ethereum Magicians forum until it's more fleshed out. The specification doesn't detail how one would write smart contracts in other languages.

Regarding the EIP itself, it feels like you're conflating some different ideas, so maybe let me clarify a couple things:
- An ERC is an application standard for Ethereum smart contracts. There is of course the ERC tokens, ERC-20 and ERC-721. But there also non-token ERCs, such as ERC-165 which describes how a contract can signal what functions it implements. Therefore, your title "ERC Token in Multiple Languages" confuses me because it isn't clear what ERC you're referring to.
- Token standards on Ethereum define an _interface_ for interacting with a deployed contract and the behavior of each method. That interface typically follows the [contract ABI](https://docs.soliditylang.org/en/v0.8.11/abi-spec.html) standard. It seems a bit futile to re-implement well defined token standards in another language when there are good implementations in solidity and viper that can be used off the shelf.
- Maybe what you're more getting to is that it could be valuable to allow java, go, rust, etc. devs to directly write smart contracts in their preferred language. This makes more sense to me than re-implementing token contracts. In that case, I don't think that makes sense as an ERC. Solidity isn't an ERC, neither is Viper. The EVM is already well defined so you are able to design whatever compilation pipeline you like, as long as it targets EVM bytecode. IIRC, there was actually an effort from ETC Core to maintain a version of LLVM that could compile EVM. This would allow many languages to compile direction to EVM.
- As someone who has spent a fair bit of time looking at and considering how general programming languages could be used in a blockchain env (mainly via wasm), I found that usually didn't make a lot of sense. Compilation pipelines are very complex and typically designed for compiling to physically hardware. There are also lots of EVM-specific operations that don't have first class respect in typical languages. Adding support for these things often ended creating a [frankenstein](https://github.com/quilt/sheth/blob/master/src/lib.rs#L46) combination of macros, libraries, etc.

jeyakatsa:
@lightclient @SamWilsn 

Thanks for the responses, these are very good insights.

Artem Vorotnikov also had some pretty good insights within the EF R&D discord about this as well. I also opened a R&D discussion within the ethresear.ch forum (https://ethresear.ch/t/a-new-erc-token-proposal/11540) before I went ahead with opening this EIP.

Thus (as it seems), WASM is the best way forward (if I were to continue down the path of implementing this entire EIP), but as it seems due to all the responses I've received in aggregate, it would be extremely complex to compile an entire token standard into another language.

Nevertheless, I've already started the process of how developers can implement their own smart-contracts in Java for example when I first started this process: https://github.com/jeyakatsa/New-ERC-Token-Java-Abstraction so it seems that this is the next-best course of action for this, then maybe expanding from there?

I'll see after I've completed the first iteration of my Solidity-to-Java Smart Contract conversion process...

Thank you both so very much for your input again. 

It's helped focus my efforts even more. I'll get back with a more focused implementation (and EIP).

EIPS/eip-XXXX.md
jeyakatsa: Will implement in next EIP.

EIPS/eip-XXXX.md
jeyakatsa: Understood.

EIPS/eip-XXXX.md
jeyakatsa: Got you.

EIPS/eip-XXXX.md
jeyakatsa: I'll me more specific next time.

PR #4652, opened 2022-01-09
Move EIP-2612 to Review

wschwab:
This PR moves EIP-2612 to Review. EIP-2612 enjoys widespread adoption, and imho should be moved forward in the EIP process. I have spoked with the EIP's author, @MrChico , and has told me that he is content with the specification as-is, and would be interested and it seeing it move forward.


wschwab:
I'm closing in order to reopen - this branch seems to have additional content in it.

PR #4691, opened 2022-01-17
[EIP-3978] Add LOGs gas refunds

k06a:
Adding gas refunds for reverted LOGs since the blockchain database was never modified.


k06a:
Maybe having a full refund is not that great idea, but let's discuss.

PR #4692, opened 2022-01-18
Move EIP-2315 to Review

gcolvin:

 - The build passes.

### DIFF
--- a/EIPS/eip-2315.md
+++ b/EIPS/eip-2315.md
@@ -2,7 +2,7 @@
 eip: 2315
 title: Simple Subroutines for the EVM
 description: This proposal introduces two opcodes to support simple subroutines.
-status: Draft
+status: Review
 type: Standards Track
 category: Core
 author: Greg Colvin (@gcolvin), Greg Colvin <greg@colvin.org>, Martin Holst Swende (@holiman), Brooklyn Zelenka (@expede)
@@ -29,7 +29,7 @@ As an alternative, we propose to provide for subroutines with two simple operati
 * `RJUMPSUB` to call a subroutine and
 * `RETURNSUB` to return from it.
 
-Taken together with other recent propoposals they provide an _efficient, static, and safe_ control-flow facility.
+Taken together with other required propoposals they provide an _efficient, static, and safe_ control-flow facility.
 
 *Efficient*.  Substantial reductions in the complexity and the gas costs of calling and optimizing simple subroutines -- as much as 56% savings in gas in the analysis below.  Substantial performance advantages for static jumps are [reported elsewhere](./eip-4200.md).
 
@@ -39,7 +39,7 @@ Taken together with other recent propoposals they provide an _efficient, static,
 
 ### History
 
-Facilities to directly support subroutines are provided by all but one of the real and virtual machines we have programmed, including the Burroughs 5000, CDC 7600, IBM 360, DEC PDP 11 and VAX, Motorola 68000, Sun SPARC, a few generations of Intel silicon, ARM, UCSD p-machine, Sun JVM, Wasm, and the sole exception -- the EVM.  In whatever form, these operations provide for
+Facilities to directly support subroutines are provided by all but one of the real and virtual machines we have programmed. Including physical machines like the Burroughs 5000, CDC 7600, IBM 360, DEC PDP 11 and VAX, Motorola 68000, Sun SPARC, ARM, and a few generations of Intel silicon.  And virtual machines like the UCSD p-machine, Sun JVM, Wasm, and the sole exception -- the EVM.  In whatever form, these operations provide for
 * capturing the current context of execution,
 * transferring control to a new context, and 
 * returning to the original context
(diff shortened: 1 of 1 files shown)


EIPS/eip-2315.md
```diff
@@ -2,7 +2,7 @@
 eip: 2315
 title: Simple Subroutines for the EVM
 description: This proposal introduces two opcodes to support simple subroutines.
```
lightclient, line 4: I think the "this proposal" can be dropped as it can be inferred by the context of the description.

lightclient (changes requested):
Please move the `Test Cases` section above `Security Considerations`

MicahZoltu (changes requested):
This currently depends on 3540 (Review), 3670 (Review), 3779 (Draft), and 4200 (Draft).  Any dependencies need to be at least as far along in the EIP process as the dependent, which means this can't move to Review until 3779 and 4200 are in Review.  I didn't dig deep into the technical details for those dependencies, but if they aren't strictly required then an appropriate course of action may be to just remove the dependency.  If you decide to go that route, you'll need to remove the external link to EIP-3779 and EIP-4200 in the Specification and Backward Compatibility sections.

----

External links should be removed before merging into Review.  Specifically this line:
> The [concept](https://people.cs.clemson.edu/~mark/subroutines.html) goes back to [Turing's Automatic Computing Engine of 1946](http://www.alanturing.net/turing_archive/archive/p/p01/P01-001.html):

I recommend just keeping the text and the attached quote, without the links.  If someone wants to look up the source of the quote they can just do an internet search for it, though that shouldn't matter because the argument should stand on its own without needing to appeal to an authority (especially one from 80 years ago).

----

I personally recommend removing the entire History section.  I generally let people include just about whatever they want in the Motivation (aside from external links), so this isn't a blocker but I think history lessons are way out of scope for an EIP.  The motivation section should be brief in explaining the value this standard adds to the ecosystem and leave it up to the user or the discussions-to for additional argumentation/context.

----

Test Cases section should be above the Security Considerations section.

----

Editor note: I just wanted to say how I appreciate your Test Cases section.  Most EIPs don't include any test cases, or if they do they are ill defined.  The test cases here are quite above the bar and I'm very happy to see them!

gcolvin:
I'll take care of the issues noted by @lightclient.  Thanks.

I'll rework the dependencies.  The other EIPs are only needed to provide the degree of the completeness and safety that I was asked for last year, so I think I can just refer to them in the safety section rather than create a dependency.  This EIP could go in without them.

 @MicahZoltu, I think the references to Turing's original work need to stay.  Turing's document is a specification (for an actual computer) published in 1946 by the National Physical Laboratory of the United Kingdom -- best practices for real and virtual machines are long-established.  Turing's design remains a best practice, and a big part of my motivation is simply to bring the EVM into line with best practices.   So his specification directly informs this one, as I state in the both the Motivation and the Rationale.

gcolvin:
As for having a History section at all.  In general, I agree.  But it became painfully clear to me last year that some critics of this proposal were not aware of the very long history of the well-established design that I am proposing here.

gcolvin:
And all thanks go to @holiman for the Test Cases.

holiman:
> And all thanks go to @holiman for the Test Cases.

Speaking of which. As of these new changes, the testcases are moot -- I think they should no longer be in the EIP. Alternatively, modified to the new semantics.

gcolvin:
I've tried to modify them to account for the changes, @holiman.  We can deal with my mistakes during review.  As I recall it these got generated by the test code itself, so really getting them right means writing the test code.  Eventually I need to do that, but not unless there is consensus that we want to do this.

gcolvin:
@lightclient @MicahZoltu   I've merged a new draft that tries to respond to your points.

EIPS/eip-2315.md
```diff
@@ -29,7 +29,7 @@ As an alternative, we propose to provide for subroutines with two simple operati
 * `RJUMPSUB` to call a subroutine and
 * `RETURNSUB` to return from it.
 
-Taken together with other recent propoposals they provide an _efficient, static, and safe_ control-flow facility.
+Taken together with other required propoposals they provide an _efficient, static, and safe_ control-flow facility.
```
gumb0, line 32:
```suggestion
Taken together with other required proposals they provide an _efficient, static, and safe_ control-flow facility.
```

MicahZoltu:
> As for having a History section at all. In general, I agree. But it became painfully clear to me last year that some critics of this proposal were not aware of the very long history of the well-established design that I am proposing here.

The information is useful, just not in the EIP itself.  I recommend including it in the `discussions-to`.  Ethereum Magicians allows you to link directly to individual comments, so if anyone needs a history lesson you can just link them directly to the comment with the history lesson.

gcolvin:
@MicahZoltu 
> The information is useful, just not in the EIP itself.

There is a tension here between an EIP being pure specification, and an EIP providing the information needed to evaluate and implement it.  As author - and my coauthor Brooklyn concurs - I really do insist that this section remain, at least through the Review period.  Note that we specifically call out the design we have chosen as Turing's design.  Without some context the reviewer may not understand.  So we have included a brief history and two stable links to the original source that we quote.

axic:
> 4200 (Draft)

Filed #4716 to solve this.

MicahZoltu:
> There is a tension here between an EIP being pure specification, and an EIP providing the information needed to evaluate and implement it.

You are correct on where the tension lies.  I am of the opinion that there is great value in having a document that contains useful information, history, links, discussion, etc. and one of those links would be the EIP.  In many cases, the EIP should *not* be the primary source of information on a particular topic.

I personally liked what Tim did for EIP-1559 which was to create this document that he maintained: https://hackmd.io/@timbeiko/1559-resources.  During the discussions around EIP-1559, I would link people to that document rather than the EIP itself, and if someone wanted they were able to click through to the EIP from there.  This gave Tim unconstrained decision making power over what was included in the document that was passed around to everyone and linked everywhere, while the EIP was able to remain pure throughout its lifecycle.

EIPS/eip-2315.md
```diff
@@ -376,7 +379,16 @@ Bytecode: `0x6005565b5d5b60035e` (`PUSH1 0x05, JUMP, JUMPDEST, RETURNSUB, JUMPDE
 
 Consumed gas: `30`
 
+## Security Considerations
+
+These changes introduce new flow control instructions, so any software which does static/dynamic analysis of EVM code needs to be modified accordingly. The `RJUMPSUB` semantics are similar to `JUMP` whereas the `RETURNSUB` instruction is different, since it can 'land' on any opcode (but the possible destinations can be statically inferred).
+
+## References
```
lightclient, line 386: This section is not valid per EIP-1 and should be removed.

lightclient (changes requested):

EIPS/eip-2315.md
gcolvin: Gone.

gcolvin:
> I personally liked what Tim did for EIP-1559 ...

I did too.  This is a different situation.  There isn't that much extra material here, part of it is normative, and removing the rest would only make the Motivation appear weak and the Rationale appear arbitrary.

Recall that the previous version of this proposal was met with last-minute objections, semantic disputes, design disagreements, and demands for justifications that could not be answered directly from the EIP, or based on my authority as an expert.  I don't want to happen again.  So I cite an unimpeachable authority and try to make only statements that I can justify from my personal experience.  I can't count on people reading any document besides this EIP, so I need the relevant material here.

So if you cannot in good conscience move this EIP to Review, @MicahZoltu, I'd ask that you not block it and allow another editor to merge this PR, so these issues can be resolved by a larger consensus on the way to Final status.

PR #4694, opened 2022-01-18
Update eip-1051.md

alita-moore:

 - The build passes.

### DIFF
--- a/EIPS/eip-1051.md
+++ b/EIPS/eip-1051.md
@@ -21,7 +21,13 @@ In the interests of facilitating efficient and secure contracts, we propose new 
 
 Two new flags are added to the EVM state: overflow (`ovf`) and signed overflow (`sovf`).
 
-The `ovf` flag is set in the following circumstances:
+The `ovf` flag is 
+
+
+
+
+
+set in the following circumstances:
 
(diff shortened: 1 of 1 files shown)


jasoncoker925:
gh pr checkout 4694

mohammadj44:
@mohammadj44 @eth-bot 

> https://github.com/mohammadj44/yaran444.git

RenanSouza2:
This may be too late, but I have some toughth on thi EIP wich is one crucial to join security and effiency

I like how signed and unsigned are treated separetely
how you can perform a batch of operations and the flags only need to be checked at the end (with some exception if you mix signed and unsigned operations)

but I also have some remarks:

The ovf flag should trigger in the operations EXP and SHL

an signed sum of a positive and a negative number should not trigger the signed and should trigger the unsigned flag. The specification says "The sovf flag is set whenever the ovf flag is set, and additionally in the following circumstances"

Division by zero, and this include operations div, mod, sdiv, smod, addmodm, mulmod as well; should be treated in the flags

The signed division -(2**255) / -1 should trigger the sovf flag

The flags related to signed multiplication is wrong, it should trigger when the modulo of the ideal multiplication is bigger equal than 2**255

The signed flag should trigger with the SHL operation, this is trickier to define when

PR #4705, opened 2022-01-20
EIP-4626: Preview functions

alcueca:
After an intense internal discussion, we have decided that 4 preview functions are needed, one for each of the 4 mutable functions. Significant use cases would either not be possible or would be not compliant if using a single `exchangeRate` function.

On the other hand, to allow easier integration of the simpler use cases, `exchangeRate()` has been kept as an alias for `redeemPreview(10**decimals())`

### DIFF
--- a/EIPS/eip-4626.md
+++ b/EIPS/eip-4626.md
@@ -98,31 +98,31 @@ Returns the total amount of underlying tokens managed by the vault.
 #### balanceOfUnderlying
 `function balanceOfUnderlying(address owner) public view returns (uint256)`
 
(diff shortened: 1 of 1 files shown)


abcoathup:
@alcueca it would be helpful to rename the PR to include the EIP so anyone looking at the PRs can see which EIP this refers to.

transmissions11:
@lightclient looks like auto merge got disabled for some reason, mind helping us out here? 😅

transmissions11:
alright nvm, i just dup'd it in #4710 instead 😅

PR #4383, opened 2021-10-19
move eip-712 to Last Call

wschwab:

PR #4709, opened 2022-01-21
Update eip-712.md

moodysalem:
correct the byte length of the return data from `eth_signTypedData`

### DIFF
--- a/EIPS/eip-712.md
+++ b/EIPS/eip-712.md
@@ -225,7 +225,7 @@ Typed data is a JSON object containing type information, domain separator parame
 
 ##### Returns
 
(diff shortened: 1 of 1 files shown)


lightclient:
cc @recmo

wschwab:
I've got a PR open for moving 712 to Last Call: #4383 , would it be okay with you @moodysalem if I include this change there so that there's just that PR for it?

SamWilsn:
@wschwab Not sure if you saw the reaction, but you got a :+1: from moodysalem to move the changes into your other PR.

wschwab:
> @wschwab Not sure if you saw the reaction, but you got a +1 from moodysalem to move the changes into your other PR.

I didn't, thanks for flagging - I don't know exactly when I'll get around to it, will try to make it today

SamWilsn:
No rush! I've just been going through the EIPs from oldest to newest, so you'll probably get pings from me periodically ;)

wschwab:
I've made the changes and will push them to #4383 , but would like some confirmation that numbers proposed here are definitely correct (and that isn't just a difference of two-digit words vs single or something like that)

lightclient:
closing in favor of #4383

Issue #2613, opened 2020-04-24
EIP-2612: 712-signed token approvals


MrChico:
This is a place for discussing https://github.com/ethereum/EIPs/pull/2612, which proposes an additional method `permit` for making approvals by way of signed messages rather than direct transactions

maurelian:
IMO the 'security considerations' section should mention the front running attack that can occur when an allowance is reduced, either by `approve` or the new `permit` function. I don't think this proposal changes the risk or possible mitigations in any way, but I haven't thought hard about it. 

If you have a strong desire to do some bike shedding, you could even suggest a [specific mitigation](https://arxiv.org/pdf/1907.00903v1.pdf) for it. 😝

Nipol:
I was looking forward to talking about this. I'd like to talk about the need for a "deadline"

What aspects of Tx Relayer do we need to take into account "deadline"? The Tx Relayer may have a response that causes tx while setting tx fee low. or no response.

And when a user signs for an interface, it requires more information. or more knowledge. Is usability not compromised? and more complexity for Tx Relayer.

So I think the deadline is unnecessary. 🤔

maurelian:
I like `deadline`, as it solves the problem of worrying about [all the approvals](https://tac.dappstar.io/#/) one has floating out in the world. 
That said, I am sympathetic to the argument that it is inconsistent with the existing `approve` interface.

MrChico:
Thanks @maurelian, added a comment about the approval front running attack. 
@nipol The `deadline` parameter is a way to mitigate the [free option](https://blog.althea.net/the-free-option-problem/) that the relaying party gets when receiving a signed message (it is just a deadline for the `Permit` itself, not for the approval so I don't think it solves the problem of random tokens being given allowance). In cases where the free option is not a concern, `deadline` can simply be set to `uint(-1)`, so it should be seen as an optional parameter

MrChico:
Actually @maurelian I was just made aware of the [xDai staking token implementation of `permit`](https://etherscan.io/address/0x0Ae055097C6d159879521C384F1D2123D1f195e6#code) which I think does what exactly what you are alluding to here. It shares the same ABI as Dai and Chai, but not the same semantics. I guess we better document all three flavors in the ERC

Nipol:
In terms of "Free Options," I think `deadline` has sufficient requirements.

But in terms of EIP-712, deadline is not convincing.
[...]

PR #4728, opened 2022-01-28
Specify that permit must revert on invalid signatures

frangio:
As suggested in https://github.com/ethereum/EIPs/issues/2613#issuecomment-1024589197.

The [recent Multichain/AnySwap vulnerability](https://media.dedaub.com/phantom-functions-and-the-billion-dollar-no-op-c56f062ae49f) that put almost half a billion at risk was due to a combination of permit, infinite approvals, and "phantom functions" (i.e. a fallback function that accepts any function call).

A literal reading of the current text of the EIP shows that the last part is not strictly necessary. A compliant implementation of `permit` isn't required to revert if the permit is invalid or old, and such an implementation could lead to the same vulnerability that was found there.

This PR adds to the specification that `permit` must revert if the signature is invalid.

### DIFF
--- a/EIPS/eip-2612.md
+++ b/EIPS/eip-2612.md
@@ -64,6 +64,8 @@ if and only if the following conditions are met:
 - `nonces[owner]` (before the state update) is equal to `nonce`.
 - `r`, `s` and `v` is a valid `secp256k1` signature from `owner` of the message:
 
+If any of these conditions are not met, the `permit` call must revert.
+
 ```sol
 keccak256(abi.encodePacked(
    hex"1901",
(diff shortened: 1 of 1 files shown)


MrChico (approved):

MrChico:
I approve of this change. I am using the language `if and only if` in the sense that they are used in https://github.com/ethereum/act to signal this already, but I'm aware that thats a little niche

lightclient:
@frangio can you also update the status back to draft?

frangio:
@lightclient Done.

lightclient:
@frangio, it appears there are now some merge conflicts.

PR #4736, opened 2022-01-31
Consensus Layer Withdrawal Protection

benjaminchodroff:


eip-draft_clwp.md
```diff
@@ -0,0 +1,202 @@
...
+eip: <to be assigned>
+title: Consensus Layer Withdrawal Protection
+description: Provides additional security in the "change withdrawal address" operation when a consensus layer mnemonic may be compromised, without changing consensus
+author: Benjamin Chodroff (@benjaminchodroff), Jim McDonald (@mcdee)
+discussions-to: https://ethereum-magicians.org/t/consensus-layer-withdrawal-protection/8161
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): Interface
+created: 2022-01-30
+requires (*optional): <EIP number(s)>
```
SamWilsn, line 11:
```suggestion
category: Interface
created: 2022-01-30
```

Can drop the `requires` section, and the stuff in parentheses is just a note for authors.

pk910:
Hmm, interesting.. Are there so many withdrawal keys compromised that this is a relevant problem already?
Validators should know how to handle mnemonics.. If someone compromises your wallet mnemonic that's your problem and there is no withdrawal delay for non-community-whitelisted transactions ;)

How should the "delay rebroadcast change withdrawal address request" part of your proposal work? 
If a valid change withdrawal address request that is not "whitelisted" gets delayed by 5 epochs on every node that receives it it'll probably take a very long time till all nodes got it.. But it's still a valid request.
That's probably what you want to achieve, so a "whitelisted" request gets propagated faster as "non-whitelisted" ones are delayed.
Wouldn't it be a better approach to rebroadcast immediately, but delay inclusion in locally built blocks for 5 epochs instead?
That way non-whitelisted but valid requests are just delayed by a maximum of 5 epochs, but not longer.
An attacker is probably smart enough to broadcast his malicious "change withdrawal address request" to as many nodes as possible to get around that double/tripple/... re-broadcast delay anyway..

This proposal needs a high participation rate to have an effect. Just a single validator would be enough to include many non-whitelisted change withdrawal address requests in a block if chosen as a proposer..
The "Change Withdrawal Address Broadcast File" needs to be updated automatically to keep in sync with the claim-repository.
Btw. Who should be in control of this repository and who should be allowed to decide in a "contested claim" and who in general is responsible for security of that repository?
Is all that necessary for a hopefully small group of validators that handled their withdrawal mnemonic insecurely?

(Just commenting what came in my head as a dev interested staking user here 😉 Happy to be proven wrong)

froid1911:
This is indeed an awesome proposal, but I have seen various Stakers who lost access to their cl withdrawal keys and only have access to the deposit and verifier keys. Therefore I propose to change the consensus_layer_withdrawal_signature to consensus_layer_verifier_signature.

benjaminchodroff:
> (Just commenting what came in my head as a dev interested staking user here 😉 Happy to be proven wrong)
Sincerely, thank you! I really appreciate any feedback -- and welcome the discussion.

Yes, the intent of the rebroadcast delay is a slight inconvenience for everyone, but not a permanent denial for anyone. While delayed, it also assumed that the request would not be included in locally built blocks too. Perhaps that isn't clear in the EIP, and thank you for the suggestion. We do not need all nodes to stop rebroadcasting, but the more that participate, the more the proposal helps. Even a slight statistical edge may help someone.

> This proposal needs a high participation rate to have an effect. 
Not necessarily. If the attackers are lazy and don't get broadcast their change withdrawal address on the first available block, it may help many legitimate users who were unaware they were compromised. If an attacker makes a public claim that is contested, at least it lets the legitimate user know they are indeed under attack -- and potentially the social mechanism could expose the hacker...

> The "Change Withdrawal Address Broadcast File" needs to be updated automatically to keep in sync with the claim-repository. 
Not necessarily. Simply not being included in the "change withdrawal address broadcast file" means by default they will be slightly delayed in being accepted if they are requesting an address that isn't their original deposit address. Additionally, this is really only for 0x00 CL addresses. It is expected new validators would default to specifying their 0x01 address upon starting their validator. 

The proposal assumes there is not a single repository. Anyone using this proposal could create a "contested claim" repository and could create a list by any social enforcement mechanism they find acceptable (aka: I'm a hacker, and I only include my own addresses as acceptable evidence!) That being said, I do intend to run my own repository and provide verifiable claims that others can verify themselves, or safely ignore.

benjaminchodroff:
> therefore I propose to change the consensus_layer_withdrawal_signature to consensus_layer_validator_signature

That idea is not "protection" and sadly outside the scope of this proposal. I do agree that a similar "Consensus Layer Recovery" EIP could be written, but would almost certainly require a change to consensus - and that is going to certainly face a harder challenge to get the community to agree.

eip-draft_clwp.md
```diff
@@ -0,0 +1,201 @@
+---
+eip: <to be assigned>
```
lightclient, line 2:
```suggestion
eip: 4736
```

lightclient (changes requested):
Overall looks pretty good to me.

lightclient:
Please move discussions to the `discussions-to` thread.

lightclient (changes requested):

EIPS/eip-4736.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Backwards Compatibility
+As there is currently no existing "change withdrawal address" operation in existence, there is no documented backwards compatibility. As all of the proposal is OPTIONAL in both implementation and operation, it is expected that client beacon nodes that do not implement this functionality would still remain fully backwards compatible with any or all clients that do implement part or all of the functionality described in this proposal. Additionally, while users are RECOMMENDED to enable these OPTIONAL features, if they decide to either disable or ignore some or all of the features, or even purposefully load content contrary to the intended purpose, the beacon node client will continue to execute fully compatible with the rest of the network as none of the proposal will change core Ethereum consensus. 
+
+## Test Cases
+This proposal does not change consensus. However, beacon node client test cases will be added here once the proposal is formalized.
+
```
MicahZoltu, line 75:
```suggestion
```
This is an optional section, so it can be left out until such time as you actually have something to put into it.

EIPS/eip-4736.md
```diff
@@ -0,0 +1,201 @@
...
+	
+It is critical to understand that this proposal is not a consensus change. Nothing in this proposal restricts the validity of withdrawal credential change operations within the protocol. It is a voluntary change by client teams to build this functionality in to their beacon nodes, and a voluntary change by node operators to accept any or all of the restrictions and broadcasting capabilities suggested by end users.
+
+Because of the above, even if fully implemented, it will be down to chance as to which validators propose blocks, and which voluntary restrictions those validators’ beacon nodes are running. Node operators can do what they will to help the community prevent attacks on any compromised consensus layer keys, but there are no guarantees of success this will prevent a successful attack. 
+
+## Specification
+The Consensus Layer change withdrawal credentials operation is [not yet fully specified](https://github.com/ethereum/consensus-specs/pull/2759), but MUST have at least the following fields:
```
MicahZoltu, line 30: While linking to Ethereum specifications is borderline acceptable, linking to a pull request to a specification is not.  Please remove this link until such time as the specification is actually updated.

EIPS/eip-4736.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Reference Implementation
+### Change Withdrawal Address Acceptance File
+A file intended to be preloaded with all consensus layer withdrawal credentials and verifiable execution layer deposit addresses. This file will be generated by a script and able to be independently verified by community members using the consensus and execution layers, and intended to be included by all clients, enabled by default. Client nodes are encouraged to enable packaging this independently verifiable list with the client software, and enable it by default to help further protect the community from unsuspected attacks. 
+
+depositAddress.csv format (both fields required):
+```withdrawal credential, withdrawal address```
```
MicahZoltu, line 78:
```suggestion
`withdrawal credential, withdrawal address`
```

MicahZoltu:
Why is this an EIP?  I thought consensus layer changes had their own specification system/process and EIPs were used exclusively for execution layer specifications?

benjaminchodroff:
> Why is this an EIP?
Fair point, we are in an unchartered area with Ethereum being both EL and CL, but this is a proposal to improve Ethereum and others guided that this was the best way to do so. I hope to use this proposal to get agreement on the specification before we start to submit pull requests to implement it.

MicahZoltu:
There is certainly disagreement within the community around the future of change control post merge.  My understanding of the unstable schelling point was that we would continue to keep CL and EL change control processes separate until such time as we figure out how to consolidate them.

MicahZoltu:
I personally don't think an EIP is the best way to get community feedback on an idea.  I would prefer a HackMD article, ethresear.ch post, Ethereum Magicians post, etc. (or some combination thereof).  IMO, EIPs serve one very specific purpose well, which is standardization/specifications of Ethereum Execution Layer consensus changes, and it is a pretty poor process for basically everything else.  This all being said, this is not a dictatorship and one of the other editors may disagree with me and agree to merge this.  You will want to get CI passing first though:
> eip-4736.md:	description exceeds max length of 140 characters

lightclient (approved):

PR #4749, opened 2022-02-03
663: Update SWAP/DUP to use EIP-3540

axic:


axic:
Oh wow, why was this automerged? It is a status change from Stagnant to Draft.

cc @alita-moore @MicahZoltu

alita-moore:
Oof, most likely because you're an editor and so can override the bot. I'll patch it asap

alita-moore:
~~Oh wait actually is because of the fact Kodiak considers skipped 
~~

Nevermind I found a real issue, patching...

alita-moore:
the problem was that in draft state the bot is skipped and when you go from draft -> review the bot is not re-run so it remains in the skipped state. The kodiak auto-merger considers skipped as success. There are a few ways for us to address this but the easiest is to disable / remove kodiak until it no longer does this.

see https://github.com/chdsbd/kodiak/issues/780 for an investigation on this topic

PR #4752, opened 2022-02-03
Revert "663: Update SWAP/DUP to use EIP-3540"

alita-moore:
Reverts ethereum/EIPs#4749


axic:
Well, I guess there wasn't too much of a need to revert it unless @MicahZoltu stands up, but oh well, will submit a new PR :)

PR #4765, opened 2022-02-04
EIP-3675: Move to Review

mkalinin:
Change EIP-3675 status to Review

### DIFF
--- a/EIPS/eip-3675.md
+++ b/EIPS/eip-3675.md
@@ -3,7 +3,7 @@ eip: 3675
 title: Upgrade consensus to Proof-of-Stake
 author: Mikhail Kalinin (@mkalinin), Danny Ryan (@djrtwo), Vitalik Buterin (@vbuterin)
 discussions-to: https://ethereum-magicians.org/t/eip-3675-upgrade-consensus-to-proof-of-stake/6706
-status: Draft
+status: Review
 type: Standards Track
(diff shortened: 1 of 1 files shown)


axic (changes requested):
Please make sure this follows the latest EIP-1 rules. Two things to note at a glance: lack of `description` preamble and deprecated `Simple Summary` section.

axic (commented):
This is okay for `Review` from perspective, but there is a big outstanding question regarding the relation to EIP-2982 and the external links to the devp2p spec.

mkalinin:
> This is okay for `Review` from perspective, but there is a big outstanding question regarding the relation to EIP-2982 and the external links to the devp2p spec.

I was thinking about whether it makes sense if an EIP `require` an informational EIP. Technically EIP-3675 implicitly depends on all consensus specs and on `bellatrix` upgrade in particular. Not sure what would be the best way to reflect this dependency. Currently, it's mentioning the beacon chain as the new consensus driver and have a link to consensus specs. I'd leave it as it is if it's appropriate.

Network section could be in a separate EIP that depends on EIP-3675. But the changes to the networking stack are tightly coupled to the transition process; not sure if it would read better if they were placed in a separate document. Though, I don't have strong arguments against doing so.

lightclient (approved):
LGTM

PR #4766, opened 2022-02-04
Update eip-747.md

ayuthmang:
When I'm trying to play in the developer console this extra semi-colon make this snippet unable to execute.

image

This PR tend to remove unnecessary semi-colon from code example that makes the snippet works.


lightclient (commented):
This change is good on my end. @danfinlay, @estebanmino: if you're okay with this change, can you approve please?

PR #4777, opened 2022-02-05
4626: clarify behavior

transmissions11:


EIPS/eip-4626.md
```diff
@@ -115,93 +131,105 @@ The `assetsOf` method SHOULD be used for more accurate calculations of a user's
...
 ```
 
 #### previewDeposit
 
 Allows an on-chain or off-chain user to simulate the effects of their deposit at the current block,
 given current on-chain conditions.
 
-MUST return the *exact* amount of Vault shares that would be minted if the caller were to deposit
-a given *exact* amount of underlying assets using the `deposit` method.
+MUST return as close to the _exact_ amount of Vault shares that would be minted if the caller were
```
fubuloubu, line 193: Does it under shoot or overshoot, or is it within some bound of what the correct answer might be? What properties are different versus `assetsPerShare`?

EIPS/eip-4626.md
```diff
@@ -115,93 +131,105 @@ The `assetsOf` method SHOULD be used for more accurate calculations of a user's
...
+to deposit a given _exact_ amount of underlying assets using the `deposit` method as technically possible.
```
axic, line 194: I think thit is still not explicitly stating that balance check cannot be made.

EIPS/eip-4626.md
```diff
@@ -88,23 +99,28 @@ MUST *NOT* revert.
...
-MAY *NOT* be completely accurate according to slippage or other on-chain conditions,
-when performing the actual exchange.
-MUST *NOT* return 0.
-MUST *NOT* revert.
-
-In certain types of fee calculations, this calculation MAY *NOT* reflect the "per-user" price-per-share,
-and instead should reflect the "average-user's" price-per-share,
-meaning what the average user should expect to see when exchanging to and from.
+
+MAY _NOT_ be completely accurate according to slippage or other on-chain conditions, when performing the actual exchange.
```
fubuloubu, line 114: I think should add the note about manipulation resistance

EIPS/eip-4626.md
```diff
@@ -23,8 +23,9 @@ and withdrawing tokens and reading balances.
 Tokenized Vaults have a lack of standardization leading to diverse implementation details.
 Some various examples include lending markets (Compound, Aave, Fuse),
 aggregators (Yearn, Rari Vaults, Idle), and intrinsically interest bearing tokens (xSushi).
-This makes integration difficult at the aggregator or plugin layer for protocols which need to conform to many standards.
-This forces each protocol to implement their own adapters which are error prone and waste development resources.
+
+This makes integration difficult at the aggregator or plugin layer for protocols which need to conform to many standards and
+forces each protocol to implement their own adapters which are error prone and waste development resources.
```
fubuloubu, line 28:
```suggestion
This makes integration difficult at the aggregator or plugin layer for protocols which need to conform to many standards,
and forces each protocol to implement their own adapters which are error prone and waste development resources.
```

EIPS/eip-4626.md
```diff
@@ -45,21 +46,28 @@ which may affect precision for computing the value of Vault shares
...
-         exchanged on mint/deposit/withdraw/redeem (defined by the Vault).
+  exchanged on mint/deposit/withdraw/redeem (defined by the Vault).
+
 - slippage: Any difference between advertised share price and economic realities of
-            deposit to or withdrawal from the Vault, which is not accounted by fees.
-- fee: An amount of assets or shares charged to the Vault. Can be a fee on deposits, yield, AUM, withdrawal, or any other kind of fee.
+  deposit to or withdrawal from the Vault, which is not accounted by fees.
+
+- fee: An amount of assets or shares charged to the Vault. Can be a fee on deposits,
+  yield, AUM, withdrawal, or any other kind of fee.
```
fubuloubu, line 60: Besides the extra spacing between the defs, I don't think this indentation changes anything in how it's displayed

EIPS/eip-4626.md
```diff
@@ -88,23 +99,28 @@ MUST *NOT* revert.
...
-MUST *NOT* return 0.
-MUST *NOT* revert.
-
-In certain types of fee calculations, this calculation MAY *NOT* reflect the "per-user" price-per-share,
-and instead should reflect the "average-user's" price-per-share,
-meaning what the average user should expect to see when exchanging to and from.
+
+MAY _NOT_ be completely accurate according to slippage or other on-chain conditions, when performing the actual exchange.
+
+MUST _NOT_ return 0 to indicate the Vault is empty.
```
fubuloubu, line 116:
```suggestion
MUST _NOT_ return 0 the Vault has no assets.
```

but potentially change to:
```
SHOULD return a 1:1 exchange rate when the Vault is empty
```

EIPS/eip-4626.md
fubuloubu, line 194: Should be one of the following:
```suggestion
MUST return the _minimum_ amount of Vault shares that would be minted if the caller were
to deposit a given _exact_ amount of underlying assets using the `deposit` method.
```

fubuloubu:
```suggestion
MUST return the _maximum_ amount of Vault shares that would be minted if the caller were
to deposit a given _exact_ amount of underlying assets using the `deposit` method.
```

fubuloubu:
```suggestion
MUST return an amount of Vault shares that would be minted if the caller were to deposit a given
 _exact_ amount of underlying assets using the `deposit` method, within an `X` bps range of accuracy.
```
where `X` is something like 1 bps (basis point e.g. 0.01%)

EIPS/eip-4626.md
```diff
@@ -115,93 +131,105 @@ The `assetsOf` method SHOULD be used for more accurate calculations of a user's
...
 Allows an on-chain or off-chain user to simulate the effects of their deposit at the current block,
 given current on-chain conditions.
 
-MUST return the *exact* amount of Vault shares that would be minted if the caller were to deposit
-a given *exact* amount of underlying assets using the `deposit` method.
+MUST return as close to the _exact_ amount of Vault shares that would be minted if the caller were
+to deposit a given _exact_ amount of underlying assets using the `deposit` method as technically possible.
+
+MUST _NOT_ account for deposit limits like those returned from `maxDeposit` and should always act
+as though the deposit would be accepted, regardless if the user has enough tokens approved, etc.
```
fubuloubu, line 197: Make a note how `max*` accounts for this instead?

EIPS/eip-4626.md
```diff
@@ -115,93 +131,105 @@ The `assetsOf` method SHOULD be used for more accurate calculations of a user's
...
+
 MUST be inclusive of deposit fees. Integrators should be aware of the existence of deposit fees.
-MUST *NOT* revert.
 
-Note that any discrepancy between `shares * assetsPerShare` and `assets` SHOULD be considered slippage
-in share price or some other type of condition, meaning the depositor will lose assets by depositing.
+MUST _NOT_ revert.
+
+Note that any unfavorable discrepancy between `assetsPerShare` and the ratio of `assets` to `shares` SHOULD
```
fubuloubu, line 203: This is where having the exchange function directly would be helpful

Fitrah92: 🤩

EIPS/eip-4626.md
transmissions11: the suggestion makes no sense read it back lol

transmissions11: > SHOULD return a 1:1 exchange rate when the Vault is empty

what if ppl want 2:1 share ratios n such? idt worth restricting

EIPS/eip-4626.md
transmissions11: idt its realistic to enforce a numerical bound here

EIPS/eip-4626.md
transmissions11: exactly :)

Fitrah92 (commented):
Received

EIPS/eip-4626.md
fubuloubu: > > SHOULD return a 1:1 exchange rate when the Vault is empty
> 
> what if ppl want 2:1 share ratios n such? idt worth restricting

I agree. How about "MUST return the initial share price when empty"?

fubuloubu:
```suggestion
MUST return the initial share price when empty.
```

EIPS/eip-4626.md
fubuloubu: So how about "MUST return some number between 0 and `2 ** 256 - 1` that represents an approximation of the exchange rate for the exact number of shares (but really it's just a random number lol, don't ask for specifics)"?

Joeysantoro:
https://github.com/ethereum/EIPs/pull/4806 included many of the changes in this PR. Some style changes were omitted. Also, the language for the preview methods should indicate that the preview methods are "no worse than" so the user should be getting a better rate if there is a discrepancy

PR #4778, opened 2022-02-05
4626: Tiny clarification for ABI

axic:
The function name is `totalAssets` and uses a name return value of `totalAssets`. If this YAML is used to auto-generate Solidity, it will result in a shadowing warning/error.


transmissions11 (approved):
nice find!

PR #4788, opened 2022-02-11
EIP-4788: Beacon state root in EVM

ralexstokes:
Expose the beacon state root in the EVM by including it in each execution block and introduce an opcode to facilitate reading the roots.

### DIFF
--- a/EIPS/eip-4788.md
+++ b/EIPS/eip-4788.md
@@ -1,0 +1,157 @@
+---
+eip: 4788
+title: Beacon state root in the EVM
+description: Expose beacon chain state roots in the EVM
+author: Alex Stokes (@ralexstokes), Danny Ryan (@djrtwo)
+discussions-to: https://ethereum-magicians.org/t/eip-4788-beacon-state-root-in-evm/8281
+status: Draft
+type: Standards Track
+category: Core
+created: 2022-02-10
+---
+
+## Abstract
+
+Commit to the state root of the beacon chain in the `ommers` field in the post-merge execution block. Reflect the changes in the `ommersHash` field of the execution block header.
+
+Store each beacon chain state root into a contract and add a new opcode that reads this contract.
+
+## Motivation
+
+Exposing the beacon chain state root allows for proofs about the beacon state to be verified inside the EVM. This functionality supports a wide variety of use cases in smart contracts involving validator status and finality produced by the consensus layer.
+
+In particular, this functionality is required for beacon chain validator withdrawals to the EVM.
+
+## Specification
+
+| constants                 | value                                        | units
+|---                        |---                                           |---
+| `FORK_TIMESTAMP`          | TBD                                          |
+| `FORK_EPOCH`              | TBD                                          |
+| `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffd` |
+| `OPCODE_VALUE`            | `0x48`                                       |
+| `G_beacon_state_root`     | 20                                           | gas
+
+### Background
+
+The method of injecting the beacon state root in this EIP follows the general strategy of [EIP-4399](./eip-4399.md) to make a post-merge change to the EVM integrating information from the beacon chain. This EIP along with [EIP-3675](./eip-3675.md) should be taken as relevant background to understand the particular approach of this EIP.
+
+The method for exposing the state root data via opcode is inspired by [EIP-2935](./eip-2935.md).
+
+### Block structure and validity
+
+Beginning at the execution timestamp `FORK_TIMESTAMP`, execution clients **MUST**:
+
+1. set the value of the `ommers` field in the block to an RLP list with one element: the 32 byte [hash tree root](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#merkleization) of the [beacon state](https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#beaconstate) from the previous slot to this block.
+
+2. set the value of the `ommersHash` field in the block header to the Keccak256 hash of the `ommers` field.
+
+```python
+beaconStateRoot = <32 byte value> # provided by consensus client
+ommers = RLP([beaconStateRoot])   # in the block body
+ommersHash = Keccak256(ommers)    # in the block header
+```
+
+3. Add the block validation that the `ommersHash` does indeed match the expected committment given the `ommers` value.
+
+### EVM changes
+
+#### Block processing
+
+At the start of processing any execution block where `block.timestamp >= FORK_TIMESTAMP` (i.e. before processing any transactions), write the beacon state root provided in the block into the storage of the smart contract at `HISTORY_STORAGE_ADDRESS`. This data is keyed by the block number.
+
+In pseudocode:
+
+```python
+beacon_state_root = block.ommers[0]
+sstore(HISTORY_STORAGE_ADDRESS, block.number, beacon_state_root)
+```
+
+#### New opcode
+
+Beginning at the execution timestamp `FORK_TIMESTAMP`, introduce a new opcode `BEACON_STATE_ROOT` at `OPCODE_VALUE`. This opcode consumes one word from the stack encoding the block number for the root. The opcode has a gas cost of `G_beacon_state_root`.
+
+The result of executing this opcode leaves one word on the stack corresponding to a read of the history contract's storage; in pseudocode:
+
+```python
+block_number = evm.stack.pop()
+sload(HISTORY_STORAGE_ADDRESS, block_number)
+```
+
+If there is no root stored at the requested block number, the opcode follows the existing EVM semantics of `sload` returning `0`.
+
+## Rationale
+
+### General strategy
+
+See the rationale for EIP-4399 for discussion about this general strategy of reusing execution block elements for beacon chain data.
+
+### Fork mechanics
+
+This EIP requires the consensus layer and execution layer to execute a network upgrade in lockstep.
+To carry out this task, a `FORK_EPOCH` (of the beacon chain) will be chosen and then used to compute a timestamp `FORK_TIMESTAMP`.
+This `FORK_TIMESTAMP` can be used in the execution layer to identify when the protocol change should be deployed.
+
+This technique works because the timestamps in post-merge execution blocks are aligned to beacon chain slots and thus serve as a proxy for the slot number.
+
+Another option for the fork definition would be to pick a beacon chain epoch and an execution payload block number.
+This design however is not reliable due to the presence of skipped slots on the beacon chain.
+
+### Execution layer validations
+
+By including the beacon state root in the execution block in the deprecated `ommers` field, execution clients can still verify the chain in a self-contained way without relying on an available consensus client.
+This property is important during syncing (and likely other phases of execution node operation).
+
+### Minimizing client code change
+
+By including the `ommersHash` validation, clients can use existing code with only minimal changes (supplying the actual state root) during block production and verification.
+Having the beacon state root value in the `ommers` field means that it is fairly straightforward to provide the value from the block data to the EVM execution context for client implementations as they stand today.
+
+### Gas cost of opcode
+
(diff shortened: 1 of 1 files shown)


vbuterin:
I would still favor using the slot number and not the block number, and copying the most recent block hash into multiple slots if there are skips the same way as the [consensus chain history accumulator does it](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#beacon-chain-state-transition-function) (see the `process_slot` function).

Reasons:

1. Block numbers are an execution-layer concept; they have no meaning in the consensus layer. Using the block number for this would introduce the concept of "the beacon block at execution block # N", which doesn't exist today. So using execution block numbers as an index of the beacon block would hurt layer separation.
2. Using the block number would also mean that we would have two accumulators of historical beacon blocks that work differently, whereas using slot number means that the accumulators work the same way, increasing cleanness (and potentially opening the door to eventually getting rid of one of those accumulators at some point in the future).
3. Slot numbers are more useful than block numbers, because they correspond directly to time. If you want something like "tell me the state at 00:00 GMT on Jan 1, 2023", then slot numbers are the only thing you could use (unless you incur the logarithmic overhead and complexity of a binary search).

I don't think the extra complexity of having an SSTORE loop is that large; it's a simple loop, and the beacon chain already does it.

mcdee:
If I'm reading it correctly, it appears that `ommersHash` will be a 32-byte hash of the RLP encoding of a single-element list of the 32-byte beacon state root.  I get the idea that you can continue to calculate `ommersHash` from `ommers`, but does this feature need to be retained?  Is altering the block structure and/or validation logic that much of an issue that it needs to be avoided at all costs?

EIPS/eip-beacon_state_root_in_evm.md
```diff
@@ -0,0 +1,169 @@
+---
+eip: <to be assigned>
```
lightclient, line 2: 4788

EIPS/eip-beacon_state_root_in_evm.md
```diff
@@ -0,0 +1,169 @@
...
+title: Beacon state root in the EVM
+description: Expose beacon chain state roots in the EVM
+author: Alex Stokes (@ralexstokes), Danny Ryan (@djrtwo)
+discussions-to: <TODO: eth-magicians URL>
```
lightclient, line 6: Please submit a discussion link.

EIPS/eip-beacon_state_root_in_evm.md
```diff
@@ -0,0 +1,169 @@
...
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): Core
```
lightclient, line 9:
```suggestion
category: Core
```

EIPS/eip-beacon_state_root_in_evm.md
```diff
@@ -0,0 +1,169 @@
...
+|---                        |---                                           |---
+| `FORK_TIMESTAMP`          | TBD                                          |
+| `FORK_EPOCH`              | TBD                                          |
+| `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffd` |
+| `OPCODE_VALUE`            | `0x48`                                       |
+| `G_beacon_state_root`     | 20                                           | gas
+
+### Background
+
+The method of injecting the beacon state root in this EIP follows the general strategy of [EIP-4399](./eip-4399.md) to make a post-merge change to the EVM integrating information from the beacon chain. This EIP along with [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) should be taken as relevant background to understand the particular approach of this EIP.
```
lightclient, line 37:
```suggestion
The method of injecting the beacon state root in this EIP follows the general strategy of [EIP-4399](./eip-4399.md) to make a post-merge change to the EVM integrating information from the beacon chain. This EIP along with [EIP-3675](./eip-3675.md) should be taken as relevant background to understand the particular approach of this EIP.
```

lightclient (dismissed):
gm

EIPS/eip-beacon_state_root_in_evm.md
```diff
@@ -0,0 +1,169 @@
...
+discussions-to: <TODO: eth-magicians URL>
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): Core
+created: 2022-02-10
+---
+
+## Abstract
+
+Commit to the state root of the beacon chain in the `ommers` field in the post-merge execution block. Reflect the changes in the `ommersHash` field of the execution block header.
```
MicahZoltu, line 15: Recommend referring to these fields by their position, since the names are meaningless (especially post-merge since it is just a fixed value).  You could say something like "in position `x`, which formerly held the `ommersHash`".

MicahZoltu:
There is no discussions-to URL yet so posting this comment here:
I strongly advocate for block root rather than state root.  There is value in being able to prove block data such as historic RANDAOs.

EIPS/eip-beacon_state_root_in_evm.md
```diff
@@ -0,0 +1,169 @@
...
+```python
+block_number = evm.stack.pop()
+sload(HISTORY_STORAGE_ADDRESS, block_number)
+```
+
+If there is no root stored at the requested block number, the opcode follows the existing EVM semantics of `sload` returning `0`.
+
+###  Other considerations
+
+The [Engine API `PayloadAttributes` message](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#payloadattributesv1) will be updated with an additional field to send the relevant beacon state root from the consensus client to the execution client when preparing a block.
```
MicahZoltu, line 85: Recommend removing this section and leaving it up to a separate changeset against the execution API specification.  This EIP can stand on its own without reference to *how* the execution client gets the data, and in general I prefer to keep changesets as isolated as possible.

EIPS/eip-beacon_state_root_in_evm.md
```diff
@@ -0,0 +1,169 @@
...
+
+Another option for the fork definition would be to pick a beacon chain epoch and an execution payload block number.
+This design however is not reliable due to the presence of skipped slots on the beacon chain.
+
+### Consensus layer validations
+
+The [`ExecutionPayload` and `ExecutionPayloadHeader`](https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#executionpayload) will be updated with a 32 byte field `beacon_state_root` for the relevant beacon state root for the corresponding execution block.
+This way the beacon state root committed to in the block can be verified against the beacon state root recorded in the beacon chain (e.g. `state.state_roots[state.slot-1]`).
+A future PR to the `consensus-specs` repo with this change is forthcoming.
+
```
MicahZoltu, line 111:
```suggestion
```
See above, this should be a separate change against the consensus spec and is out of scope of this EIP (which specifies behavior of the execution client).

ralexstokes:
@mcdee 

> If I'm reading it correctly, it appears that `ommersHash` will be a 32-byte hash of the RLP encoding of a single-element list of the 32-byte beacon state root. I get the idea that you can continue to calculate `ommersHash` from `ommers`, but does this feature need to be retained? Is altering the block structure and/or validation logic that much of an issue that it needs to be avoided at all costs?

The primary goal here is to get the beacon accumulator into the execution block header (or at least committed to there). Execution clients already have this functionality to include ommer headers (and commit to them via the `ommersHash`). 

This EIP suggests to reuse this machinery as it can simply be used as is, without having to introduce further development and testing requirements for client developers going into Shanghai (that already has a number of EIPS already competing for scarce developer attention).

There are a couple of alternatives if we don't want to go this route:

1) Retain the EIP-3675 behavior that the ommers part of the execution block is empty. Repurpose the `ommersHash` as the `beaconStateRoot` field. The state root is available now via the header. The downside is that if we need to include further data into the execution block we will no longer have a variable length portion of the block to take advantage of. The route this EIP suggests leaves open the possibility that we can place further info into the deprecated `ommers` field if needed.

2) Leave the ommers alone entirely per the EIP-3675 functionality and add an entirely new field to the block header. As per my understanding, there is some desire to not disturb the block header for reasons of backwards compatibility (although we already added the EIP-1559 base fee during the London upgrade so this point is weakened *some*).

I want to get client developer input on this point to see what the appetite is for the various options.

ralexstokes:
@vbuterin 

> I don't think the extra complexity of having an SSTORE loop is that large; it's a simple loop, and the beacon chain already does it.

I agree with all of your reasoning -- the concern is efficient reads and writes in the case where we have many skipped slots. Unlikely to happen on mainnet, but we should still consider worst case behavior. Will think this over some more...

ralexstokes:
@MicahZoltu 

> There is no discussions-to URL yet so posting this comment here:

updated! https://ethereum-magicians.org/t/eip-4788-beacon-state-root-in-evm/8281

feel free to migrate this there if you want :)

> I strongly advocate for block root rather than state root. There is value in being able to prove block data such as historic RANDAOs.

we could go either way here -- the important context is that all of the block roots are committed to in the beacon state (so only a few extra hashes in a merkle proof to go from state root to any block root) and so it is fairly easy to e.g. prove some RANDAO value in a block via the state root.

the opposite also holds: each block has its state root, so we could just as easily use the block root and then (with a few extra hashes) reach the state root and then prove anything in the state we wanted...

my personal feeling is that there is more in the *state* that a user would want to make proofs against (e.g. finalized or justified blocks for light client-esque bridges) and so we should favor that use case more (by requiring fewer hashes to process in a proof) -- knowing this do you still see a case for block roots?

EIPS/eip-beacon_state_root_in_evm.md
ralexstokes: makes sense

EIPS/eip-beacon_state_root_in_evm.md
ralexstokes: fair  -- this was a hold over from a previous document explaining the "full picture" idea, no problem to drop here now

MicahZoltu:
> we could go either way here -- the important context is that all of the block roots are committed to in the beacon state (so only a few extra hashes in a merkle proof to go from state root to any block root) and so it is fairly easy to e.g. prove some RANDAO value in a block via the state root.
> 
> the opposite also holds: each block has its state root, so we could just as easily use the block root and then (with a few extra hashes) reach the state root and then prove anything in the state we wanted...
> 
> my personal feeling is that there is more in the _state_ that a user would want to make proofs against (e.g. finalized or justified blocks for light client-esque bridges) and so we should favor that use case more (by requiring fewer hashes to process in a proof) -- knowing this do you still see a case for block roots?

This would be a good bit of text to put in the Rationale.

MicahZoltu:
CI blocked on:
```
./EIPS/eip-4788.md:55: committment ==> commitment
```

MicahZoltu (approved):
No blockers, but recommend applying or commenting on the remaining feedback.

EIPS/eip-4788.md
```diff
@@ -0,0 +1,157 @@
...
+
+2. set the value of the `ommersHash` field in the block header to the Keccak256 hash of the `ommers` field.
+
+```python
+beaconStateRoot = <32 byte value> # provided by consensus client
+ommers = RLP([beaconStateRoot])   # in the block body
+ommersHash = Keccak256(ommers)    # in the block header
+```
+
+3. Add the block validation that the `ommersHash` does indeed match the expected committment given the `ommers` value.
```
lightclient, line 55:
```suggestion
3. Add the block validation that the `ommersHash` does indeed match the expected commitment given the `ommers` value.
```

lightclient (changes requested):

lightclient (approved):

jambestwick:
> There is no discussions-to URL yet so posting this comment here: I strongly advocate for block root rather than state root. There is value in being able to prove block data such as historic RANDAOs.

agreeed

EIPS/eip-beacon_state_root_in_evm.md
xiaobaiskill: Thank you for your suggestions. I have made targeted modifications based on your feedback and have added a file containing detailed information related to the proposal. Thank you very much for your contribution.

leveyp94-hash (approved):

PR #4820, opened 2022-02-17
EIP 4626 - Fix parameter names

alcueca:
- Use `owner` and `receiver` consistently, and rework the sentences where they appear for better grammatical flow and precision.
 - Talk about vault shares in lowercase
 - Introduce a couple of missing "exactly" remarks.


fubuloubu (approved):
I used capital Vault to refer to spec implementation, perhaps it could be in the definitions section that it refers to "the implementation of ERC4626"

PR #4826, opened 2022-02-19
Changes to event fields

alcueca:
`Deposit(caller, receiver, assets, shares)` -> `Deposit(caller, owner, assets, shares)`
`Withdraw(caller, receiver, assets, shares)` -> `Withdraw(caller, receiver, owner, assets, shares)`

### DIFF
--- a/EIPS/eip-4626.md
+++ b/EIPS/eip-4626.md
@@ -479,7 +479,7 @@ Note that some implementations will require pre-requesting to the Vault before a
 
 #### Deposit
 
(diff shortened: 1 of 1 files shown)


fubuloubu (approved):

transmissions11:
i dont like the naming of `owner` for the recipient in deposit, but wont die for it

also think `owner` should come right after `caller`` in Withdraw, not after `receiver`

alcueca:
Done that way to match function signatures

transmissions11:
we discussed that a while ago, common for events not to match function params to be more intuitive

PR #4830, opened 2022-02-22
EIP-4803: Move to review

axic:


MicahZoltu (approved):
I still think that `2^31-1` (~2 billion) is better on just about every front with the *exception* of it theoretically being reachable one day in the distant future.

Clients that currently limit uint64 and int64 already break this rule today, and clients that use uint64 will continue to break this rule after this is merged so the argument that "`2^63-1` is easier to implement" falls flat for me, since clearly no one *actually* cares as no one implements this correctly today.

PR #4837, opened 2022-02-22
EIP-4399: Move to Review

mkalinin:


MicahZoltu (approved):
Test cases are required for Core EIPs.  This isn't a blocker for now, but it eventually will become a blocker.  That being said, I don't actually know how to structure a test case for this.  Maybe something along the lines of describing the state of the stack before and after the 0x44 opcode is called, given a known previous RANDAO?

PR #4844, opened 2022-02-26
Shard Blob Transactions

protolambda:
This EIP introduces shard blob transactions. See EIP abstract.

TODO: update `discussions-to` link


protolambda:
Force pushes log: Got number 4844 unexpectedly, 4843 doesn't exist, had to change the name. Updated discussions link. Fixed a remaining TODO (`tx.header` -> `tx.message` to follow naming pattern established in CL).

MicahZoltu:
I'm very suspicious of an EIP numbered 4844 with @lightclient as an author and issue/PR number 4843 completely missing from GitHub.  As soon as I figure out how you managed to skip 4843 I will decide your fate.

EIPS/eip-4844.md
```diff
@@ -0,0 +1,564 @@
...
+The work that is already done in this EIP includes:
+
+* A new transaction type, of the exact same format that will need to exist in "full sharding"
+* _All_ of the execution-layer logic required for full sharding
+* _All_ of the execution / consensus cross-verification logic required for full sharding
+* Layer separation between `BeaconBlock` verification and data availability sampling blobs
+* Most of the `BeaconBlock` logic required for full sharding
+* A self-adjusting independent gasprice for blobs
+  ([multidimensional EIP 1559](https://ethresear.ch/t/multidimensional-eip-1559/11651) with an
+  [exponential pricing rule](https://ethresear.ch/t/make-eip-1559-more-like-an-amm-curve/9082))
```
MicahZoltu, line 465: External links shouldn't be included in an EIP.  The EIP should stand on its own.  I recommend moving these links to the discussions-to URL (perhaps in the original post) and changing this to:
```suggestion
* A self-adjusting independent gasprice for blobs
```

EIPS/eip-4844.md
```diff
@@ -0,0 +1,564 @@
...
+Rollups need data to be available once, long enough to ensure honest actors can construct the rollup state, but not forever.
+
+Optimistic rollups only need to actually proof the underlying data when fraud proofs are being submitted.
+The fraud proof submission function would require the full contents of the fraudulent blob to be submitted as part of calldata.
+It would use the blob verification function to verify the data against the versioned hash that was submitted before,
+and then perform the fraud proof verification on that data as is done today.
+
+ZK rollups would provide two commitments to their transaction or state delta data:
+the kzg in the blob and some commitment using whatever proof system the ZK rollup uses internally.
+They would use the [commitment proof of equivalence protoco](https://ethresear.ch/t/easy-proof-of-equivalence-between-multiple-polynomial-commitment-schemes-to-the-same-data/8188),
```
MicahZoltu, line 495: Same comment as above:
```suggestion
They would use the commitment proof of equivalence protocol,
```
(also, `protocol` was spelled wrong)

EIPS/eip-4844.md
```diff
@@ -0,0 +1,564 @@
...
+
+Work in progress. Prototype implementations are actively being tested and used to understand the testing surface in more depth.
+
+## Reference implementation
+
+The reference implementation is a work in progress while the specification is being reviewed and improved.
+
+Execution layer implementation: https://github.com/protolambda/go-ethereum/pull/1
+
+Consensus layer implementation: https://github.com/prysmaticlabs/prysm/compare/kiln...blobs
```
MicahZoltu, line 549: Either include a reference implementation inline, or include one in `./assets/eip-4844/*`, or just leave this section out entirely if a self contained reference implementation isn't possible or doesn't make sense.

EIPS/eip-4844.md
```diff
@@ -0,0 +1,564 @@
...
+
+## Security Considerations
+
+This EIP increases the storage requirements per Beacon block by a maximum of ~2 MB.
+This is equal to the theoretical maximum size of a block today (30M gas / 16 gas per calldata byte = 1.875M bytes), and so it will not greatly increase worst-case bandwidth.
+Post-merge, block times are expected to be static rather than an unpredictable Poisson distribution, giving a guaranteed period of time for large blocks to propagate.
+
+The _sustained_ load of this EIP is much lower than alternatives (eg. EIP-4488 and EIP-4490),
```
MicahZoltu, line 557:
```suggestion
The _sustained_ load of this EIP is much lower than alternatives (eg. [EIP-4488](./eip-4488.md) and [EIP-4490](./eip-4490.md)),
```

EIPS/eip-4844.md
```diff
@@ -0,0 +1,564 @@
...
+where each fee is the 10% minimum increment higher than the previous, and finishes it off with a 21000-gas basic transaction with fee `f10`.
+Hence, an attacker could impose millions of gas worth of load on the network and only pay 21000 gas worth of fees.
+
+We recommend a simple solution: both for blob transactions and for transactions carrying a large amount of calldata,
+increase the minimum increment for mempool replacement from 1.1x to 2x, decreasing the number of resubmissions an attacker can do at any given fee level by ~7x.
+
+## Test Cases
+
+Work in progress. Prototype implementations are actively being tested and used to understand the testing surface in more depth.
+
```
MicahZoltu, line 542:
```suggestion
## Test Cases

TBD
```
Recommend simplifying this so it stands out more as being incomplete and doesn't get overlooked before moving to last call/final.  You don't have to explain why you don't have tests in a draft, you just have to make sure they exist before last call/review (for Core EIPs).

MicahZoltu (changes requested):
External links and switching EIP links to relative are the only blockers.

protolambda:
@MicahZoltu thank you for your review. I implemented your suggestions and moved the background links to the eth magicians discussion thread.

EIPS/eip-4844.md
```diff
@@ -0,0 +1,554 @@
...
+TBD
+
+## Security Considerations
+
+This EIP increases the storage requirements per Beacon block by a maximum of ~2 MB.
+This is equal to the theoretical maximum size of a block today (30M gas / 16 gas per calldata byte = 1.875M bytes), and so it will not greatly increase worst-case bandwidth.
+Post-merge, block times are expected to be static rather than an unpredictable Poisson distribution, giving a guaranteed period of time for large blocks to propagate.
+
+The _sustained_ load of this EIP is much lower than alternatives (eg. [EIP-4488](./eip-4488.md) and [EIP-4490](./eip-4490.md)),
```
MicahZoltu, line 547: Any EIPs you link to are considered "dependencies", and thus require those EIPs be as far along in the EIP process as this one.  `EIP-4490` isn't an EIP at all, and `EIP-4488` is a draft I think (and may not reach final before this).  I recommend just describing (very briefly) the alternatives and then including additional links out in the discussions-to thread.

As mentioned previously, EIPs should stand on their own aside from their dependencies, and the EIP document isn't an appropriate place to make comparisons against other proposals aside from a high level, "we could do X, but we decided this is superior because of Y".

MicahZoltu (changes requested):
CI failures:
> - ./_site/EIPS/eip-4844.html
>   *  internally linking to ./eip-4490.md, which does not exist (line 941)
>      <a href="./eip-4490.md">EIP-4490</a>
>   *  linking to internal hash #Networking that does not exist (line 430)
>      <a href="#Networking">Networking</a>

For the anchor link, maybe try `###Networking`?  I have seen people use anchor links in the past in EIPs, but I never looked into how it works.  The rendering system uses Jekyll, so that would be where I would start if you want to dig deeper.

> eip-4844.md:	description exceeds max length of 140 characters

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+eip: 4844
+title: Shard Blob Transactions
+description: Shard Blob Transactions scale data-availability of ethereum in a simplified manner while being forwards-compatible with the sharding roadmap.
+author: Vitalik Buterin (@vbuterin), Dankrad Feist (@dankrad), Diederik Loerakker (@protolambda), George Kadianakis (@asn-d6), Matt Garnett (@lightclient)
+discussions-to: https://ethereum-magicians.org/t/eip-4844-shard-blob-transactions/8430
+status: Draft
+type: Standards Track
+category: Core
+created: 2022-02-25
+requires: 2718, 2930, 1559
```
lightclient, line 11:
```suggestion
requires: 1559, 2718, 2930
```
Must be in ascending order.

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+class ECDSASignature(Container):
+    v: uint8
+    r: uint256
+    s: uint256
+```
+
+The `priority_fee_per_gas` and `max_basefee_per_gas` fields follow [EIP-1559](./eip-1559.md) semantics,
+and `access_list` as in [`EIP-2930`](./eip-2930.md).
+
+[`EIP-2718`](./eip-2718.md) is extended with a "wrapper data", the typed transaction can be encoded in two forms, dependent on the context:
```
djrtwo, line 164: If this is to be a general purpose extension to 2718, we might want to strip it out into a separate EIP

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+Execution-payloads / blocks use the minimal encoding of transactions.
+In the transaction-pool and local transaction-journal the network encoding is used.
+
+For previous types of transactions the network encoding is no different, i.e. `TransactionNetworkPayload == TransactionPayload`.
+
+The `TransactionNetworkPayload` wraps a `TransactionPayload` with additional data:
+this wrapping data SHOULD be verified directly before or after signature verification.
+
+When a blob transaction is passed through the network (see the [Networking](#Networking) section below),
+the `TransactionNetworkPayload` version of the transaction also includes `blobs` and `kzgs` (commitments list).
```
djrtwo, line 177: Might consider `Minimal` versions of blob-type-txs as gossiped and the `blobs`/`kzgs` extension gotten via req/resp. To avoid the gossip amplification factor

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+    return ecrecover(tx_hash(tx), sig.v, sig.r, sig.s)
+```
+
+### Beacon chain validation
+
+On the consensus-layer the blobs are now referenced, but not fully encoded, in the beacon block body.
+Instead of embedding the full contents in the body, the contents of the blobs are propagated separately, as a "sidecar".
+
+This "sidecar" design provides forward compatibility for further data increases:
+with data-availability-sampling (DAS) not all blobs will have to be downloaded by all beacon nodes on the network.
```
djrtwo, line 207:
```suggestion
This "sidecar" design provides forward compatibility for further data increases by black-boxing `is_data_available()`:
with full sharding `is_data_available()` can be replaced by data-availability-sampling (DAS) thus avoiding all blobs being downloaded by all beacon nodes on the network.
```

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+def tx_hash(tx: SignedBlobTransaction) -> Bytes32:
+    # The pre-image is prefixed with the transaction-type to avoid hash collisions with other tx hashers and types
+    return keccak256(BLOB_TX_TYPE + ssz.hash_tree_root(tx.message))
+
+def get_origin(tx: SignedBlobTransaction) -> Address:
+    sig = tx.signature
+    return ecrecover(tx_hash(tx), sig.v, sig.r, sig.s)
+```
+
+### Beacon chain validation
```
djrtwo, line 201: Suggest moving CL code changes out into `consensus-specs` repo and use "events" and other types of mechanisms to just reference the outer CL functionality

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+    beacon_block_slot: Slot
+    blobs: List[Blob, MAX_OBJECT_LIST_SIZE]
+```
+
+Nodes broadcast the `SignedBeaconBlock` along with its corresponding
+`BlobsSidecar` over the network instead of plain beacon blocks.
+
+### Required blobs downloading
+
+When a node receives both objects, it first calls `verify_blobs` and if this call passes it runs the usual processing on the beacon block.
```
djrtwo, line 243: Might consider using a function `is_data_available()` to abstractly talk about this verification. and then define it as all the data being in the sidecar for this EIP

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+
+### P2P changes
+
+Full peer-to-peer specification is out of scope for this EIP, the consensus-specs expands on this instead.
+
+In summary, the peer to peer layer changes involve:
+- Along other consensus-layer upgrade modifications to `SignedBeaconBlock`, the `beacon_block` gossip topic is updated 
+  and the `beacon_blocks_by_range/2/` and `beacon_blocks_by_root/2/` type tables are updated.
+- Blobs sidecars are relayed through a new global GossipSub topic: `blobs_sidecar`
+- Blobs sidecars are synced through ReqResp protocols:
```
djrtwo, line 269: Important to define how long these are expected to be served (e.g. on the order of a month?) to put the bound on when this new data can be pruned

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+  and paves the precedent that all new transaction types should be SSZ
+* It defines `TransactionNetworkPayload` to separate network and block encodings of a transaction type
+* Its (cleaner) gas price update rule could be applied to the primary basefee.
+  It also starts the trend toward using the state to store system state instead of having "implied state" such as historical blocks and hashes.
+
+### How rollups would function
+
+Instead of putting rollup block data in transaction calldata, rollups would expect rollup block submitters
+to put the data into blobs. This guarantees availability (which is what rollups need) but would be much cheaper than calldata.
+Rollups need data to be available once, long enough to ensure honest actors can construct the rollup state, but not forever.
```
djrtwo, line 488: Thus why we can have different p2p/node pruning rules than the rest of block data

djrtwo (commented):
nice!

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+    v: uint8
+    r: uint256
+    s: uint256
+```
+
+The `priority_fee_per_gas` and `max_basefee_per_gas` fields follow [EIP-1559](./eip-1559.md) semantics,
+and `access_list` as in [`EIP-2930`](./eip-2930.md).
+
+[`EIP-2718`](./eip-2718.md) is extended with a "wrapper data", the typed transaction can be encoded in two forms, dependent on the context:
+- Network (default): `TransactionType || TransactionNetworkPayload`, or `LegacyTransaction`
```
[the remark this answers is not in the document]
MicahZoltu: Haven't we had this debate before?  There is no nested RLP encoding, it is just RLP.

EIPS/eip-4844.md
```diff
@@ -0,0 +1,558 @@
...
+* The KZG commitments match the blob contents. (Note: this can be optimized with additional data, using a proof for a 
+  random evaluation at two points derived from the commitment and blob data)
+
+
+The signature is verified and `tx.origin` is calculated as follows:
+
+```python
+def tx_hash(tx: SignedBlobTransaction) -> Bytes32:
+    # The pre-image is prefixed with the transaction-type to avoid hash collisions with other tx hashers and types
+    return keccak256(BLOB_TX_TYPE + ssz.hash_tree_root(tx.message))
```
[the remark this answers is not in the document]
MicahZoltu: IIUC `ssz.hash_tree_root` lets you do fancy proofs of specific items within an object, rather than needing to provide the entire object.

EIPS/eip-4844.md
protolambda: Before creating a whole new EIP I want to get feedback in ACD first (there may be different ways we split this EIP).

EIPS/eip-4844.md
protolambda: Will work on a draft for the CL side this week. More clarification about the consensus specs and EIP relation would be appreciated.

EIPS/eip-4844.md
protolambda: Added as suggested point for feedback to the EIP-4844 discussion in ACD. For now I'll default to not extend p2p communication.

MicahZoltu (approved):

Pandapip1:
> As soon as I figure out how you managed to skip 4843 I will decide your fate.

@MicahZoltu if someone opens an issue or PR and then gets banned from GitHub, the issue then 404's (see, or rather fail to see #4928)

MicahZoltu:
Why didn't I even receive an email if that was the case?  For 4928 I at least got an email, since the person got banned *after* the email went out (which is immediate).

Pandapip1:
Okay, I have no idea then.

PR #4863, opened 2022-02-28
EIP-4863: Beacon chain push withdrawals

ralexstokes:


EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,121 @@
+---
+eip: TBD
```
ralexstokes, line 2:
```suggestion
eip: 4863
```

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Block validity
+
+If a block contains *any* transactions with `WITHDRAWAL_TX_TYPE` type, they **MUST** come before **ALL** other transactions in the block.
+
+If the execution client receives a block where this is not the case, it **MUST** consider the block invalid.
+
+### Transaction processing
+
+When processing a transaction with `WITHDRAWAL_TX_TYPE` type, the implementation should make a balance transfer for `amount` Gwei to the `address` specified by the `WithdrawalTransaction`.
```
MicahZoltu, line 58:
```suggestion
When processing a transaction with `WITHDRAWAL_TX_TYPE` type, the implementation should make a balance transfer for `amount` nanoether to the `address` specified by the `WithdrawalTransaction`.
```
My hill of the day.

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,121 @@
...
+There is no source for this balance transfer, much like the coinbase transfer.
+
+This balance transfer is unconditional and **MUST** not fail.
```
MicahZoltu, line 61:
```suggestion
When processing a transaction with `WITHDRAWAL_TX_TYPE` type, the implementation **MUST** increase the balance of the `address` by the `amount` nanoether specified by the `WithdrawalTransaction`.
No ether deducted from any other accounts, much like block mining rewards under Proof of Work.

This transaction is unconditional and **MUST** not fail.
```
I don't think we should call this a `transfer`, since that implies there is a source and a destination.  This is a mint from the context of the EVM.

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,121 @@
...
+### Why a new transaction type? And why no gas?
+
+Special semantics
+
+Firewall off generic EVM excution from this type of processing to simplify testing, security review.
+
+### Why no gas costs for new transaction type?
+
+The maximum number of this transaction type that can reach the execution layer at a given time is bounded (enforced by the consensus layer) and this limit is kept small so that
+any gas required is negligible in the context of the broader block gas limit.
```
MicahZoltu, line 84:
```suggestion
any execution layer operational costs is negligible in the context of the broader block execution.
```
For clarity, we should be talking about execution layer operational costs here, since that is what matters (gas is just a proxy for that, and it isn't even being used in this case).

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,121 @@
...
+
+ensure only those receipts genuinely made on the beacon chain are passed to the EVM
+
+provide efficient means for beacon chain to check this
+
+### SSZ vs RLP, and their boundary
+
+< explain where we put the translation boundary from SSZ to RLP >
+
+point out that the transaction hash is hash of RLP of SSZ encoding, not the hash tree root of the SSZ
```
MicahZoltu, line 102: The transaction hashing mechanism historically hasn't been part of the specification for new transaction types, since transaction hashes are not actually part of consensus (they are just a networking layer thing).  Also, I think the statement here is incorrect, the transaction hash is over `rlp(TransactionType | TransactionPayload)` (I may be wrong on this, someone should double check).

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,121 @@
...
+## Backwards Compatibility
+
+No issues.
+
+## Reference Implementation
+
+A draft PR containing a prototype implementation in Geth can be found here:
+
+https://github.com/ethereum/go-ethereum/pull/24468
+
```
MicahZoltu, line 113:
```suggestion
```
Either inline a reference implementation, or include a standalone reference implementation in `../assets/eip-4863/*`, or leave this section out.  Linking to a pull request to an entire execution client does not qualify as a "reference implementation" for an EIP, that is just called a full production client.

MicahZoltu (changes requested):
Only blocker is the Reference Implementation section.

I would like to have a debate about using the term `Gwei` in our specifications and documentation at some point, probably on a broader stage though.  I think we are doing our users (especially future users) a disservice by perpetuating the term `gwei` and `wei` instead of `nanoether` and `attoether` because most users don't understand that `wei` is just a small amount of ether and they think it is some other unit/measure.

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,121 @@
...
+The `TransactionPayload` is an SSZ-encoded container given by the following schema:
+
+```python
+class WithdrawalTransaction(Container):
+    address: ExecutionAddress
+    amount: Gwei
+```
+
+where `ExecutionAddress` is an alias for a `Bytes20` SSZ type and `Gwei` is an alias for a `uint64` SSZ type.
+Refer to the [SSZ specs](https://github.com/ethereum/consensus-specs/blob/master/ssz/simple-serialize.md) for further details on layout and encoding.
```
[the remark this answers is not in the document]
ralexstokes: nope

EIPS/eip-beacon_chain_push_withdrawals.md
ralexstokes: agreed, now it simply reads as a balance increase

ralexstokes:
> I would like to have a debate about using the term Gwei in our specifications and documentation at some point

yeah i also think we should go ahead and do the conversion Gwei -> Wei on the beacon so there is one less thing to do at the execution layer

EIPS/eip-beacon_chain_push_withdrawals.md
ralexstokes: at this point i'm leaning towards not reaching for a new way to compute the tx hash -- which AIUI will be `hash(rlp(Type | Payload))`

EIPS/eip-beacon_chain_push_withdrawals.md
ralexstokes: i think a link to a PR here is a great way to showcase the relative complexity of this change (esp. if you are familiar w/ Geth implementation).
 
i can remove it if you think it blocks this EIP but i'll just link it everywhere else

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,130 @@
...
+When processing a transaction with `WITHDRAWAL_TX_TYPE` type, the implementation should increase the balance of the `address` specified by
+the `WithdrawalTransaction` by the `amount` of wei specified.
+
+This balance change is unconditional and **MUST** not fail.
+
+This transaction type has no associated gas costs.
+
+Implementations should take care to decode the correct amount of wei from the `EncodedWei` value, which represents a 256-bit unsigned integer value in little-endian byte order.
+
+TODO: add logs?
```
protolambda, line 68: Does this transaction have a receipt? Current APIs assume receipts can be accessed by tx-index, so those would break if this transaction does not output a receipt. Also, if the receipt is defined as the standard `type ++ rlp([status, cumulative_transaction_gas_used, logs_bloom, logs])`, then the receipt is a constant bytestring, since this always succeeds, has 0 gas, and no logs.

EIPS/eip-beacon_chain_push_withdrawals.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Reference Implementation
+
+A draft PR containing a prototype implementation in Geth can be found here:
+
+https://github.com/ethereum/go-ethereum/pull/24468
+
+## Security Considerations
+
+Consensus-layer validation of withdrawal transactions is critical to ensure that the proper amount of ETH is withdrawn back into the execution layer.
```
protolambda, line 125: Additional security concern: this type of transaction may not be propagated through anything else than the engine API or trusted blocks (i.e. part of a chain recognized by the consensus layer). The tx-pool could be spammed with useless non-processable withdrawal transactions, and execution payload builders shouldn't consider these during building, or the payload will be invalid.

EIPS/eip-beacon_chain_push_withdrawals.md
ralexstokes: i think it is likely we add some kind of logging and along with those changes we'd introduce some kind of receipt format as well.

EIPS/eip-beacon_chain_push_withdrawals.md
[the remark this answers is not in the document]
djrtwo: But the executioin-layer can't actually do the validation. It is entirely reliant upon CL validating that the proper receipts were dequeued from the beacon state.

The execution-layer can enforce some things such as type-3 TXs only being at the front of the block instead of interspersed in normal TXs, but it can't know the proper count or values (unless we added requisite merkle proofs and counters)

so specifically --  what do you want to see in the block-validity section? If a block is being inserted (via the CL events) it is assumed requisite CL validations have occured

djrtwo: good call! 

It might be worth defining this more broadly as a "system TX" or a "CL TX" and none of those types of TXs can be gossiped

EIPS/eip-beacon_chain_push_withdrawals.md
ralexstokes: done!

EIPS/eip-4863.md
```diff
@@ -0,0 +1,124 @@
...
+
+This push approach gives validators a small subsidy with respect to processing, in lieu of needing to buy gas via normal EVM processing that would be required for a pull-based approach.
+
+This style also happens automatically when the requisite conditions are met on the beacon chain which is nicer UX for validators.
+
+### Why a new transaction type?
+
+This EIP suggests a new transaction type as it has special semantics different from other existing types of EVM transactions.
+
+An entirely new transaction type firewalls off generic EVM excution from this type of processing to simplify testing and security review of withdrawals.
```
MicahZoltu, line 82:
```suggestion
An entirely new transaction type firewalls off generic EVM execution from this type of processing to simplify testing and security review of withdrawals.
```

MicahZoltu (approved):

Kaktus259 (commented):
Q

Skylerspencer (commented):
Thanks

PR #4871, opened 2022-03-04
fix the link to eip-2929

ZumZoom:


MicahZoltu:
@ZumZoom This EIP is currently stagnant.  If you would like to shepherd this EIP toward final, please change the status from Stagnant to either Draft or Review and add yourself as an author and then get @wjmelements to approve such a PR.  If you don't intend to shepherd this EIP through the process to final, then it is best to just leave this as is in Stagnant and not iterate on it.

wjmelements (approved):

gcolvin (approved):
Looks good.

PR #4877, opened 2022-03-06
add MetadataUpdate event for interface ERC721Metadata

0xanders:
Add MetadataUpdated event for ERC721Metadata. 

This event emits when the result of tokenURI(_tokenID) will be  changed. So that the third parties such as NFT market could timely  update the NFT images and related attributes.

### DIFF
--- a/EIPS/eip-721.md
+++ b/EIPS/eip-721.md
@@ -180,6 +180,11 @@ The **metadata extension** is OPTIONAL for ERC-721 smart contracts (see "caveats
 /// @dev See https://eips.ethereum.org/EIPS/eip-721
 ///  Note: the ERC-165 identifier for this interface is 0x5b5e139f.
 interface ERC721Metadata /* is ERC721 */ {
+    /// @dev This event emits when the result of tokenURI(_tokenID) will be 
+    /// changed. So that the third parties such as NFT market could timely 
+    /// update the NFT images and related attributes.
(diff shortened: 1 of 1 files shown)


Sagi1521:
> Hi! I'm a bot, and I wanted to automerge your PR, but couldn't because of the following issue(s):
> 
> 
> ---
> ## (fail) eip-721.md
> | classification |
> | ------------- |
> | `updateEIP` |
> - eip-721.md is in state final at the head commit, not draft or last call or review; an EIP editor needs to approve this change
> - This PR requires review from one of [@lightclient, @axic, @samwilsn]
> - eip-721.md requires approval from one of (@fulldecent, @dekz, @nastassiasachs)

Pandapip1:
This should probably be a new EIP.

MicahZoltu:
> Only one event is added, and the event does not affect the value of the interfaceId.

It doesn't matter.  Normative changes (such as this one) are now allowed on final EIPs.  You may create a new EIP that introduces the event and has a dependency on this EIP, but you cannot make a normative change to a final EIP.

0xanders:
> > Only one event is added, and the event does not affect the value of the interfaceId.
> 
> It doesn't matter. Normative changes (such as this one) are now allowed on final EIPs. You may create a new EIP that introduces the event and has a dependency on this EIP, but you cannot make a normative change to a final EIP.

OK, thanks.

PR #4880, opened 2022-03-07
Update language to reflect the reason of why we should not link to external resources, and propose exceptions

xinbenlv:
Update language to reflect the reason of why we should not link to external resources, and propose exceptions

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -185,9 +185,9 @@ The `created` header records the date that the EIP was assigned a number. Both h
 
 EIPs may have a `requires` header, indicating the EIP numbers that this EIP depends on.
 
-## Linking to External Resources
+## Direct Linking to External Resources
 
-Links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
+Excersise extreme caution when adding external resources. Direct link e.g. [URI](https://datatracker.ietf.org/doc/html/rfc3986) **SHOULD NOT** be included. External resources, if hosted centralized, may disappear, move, or change unexpectedly or manipulated by adversary, creating security and abuse harm. Exceptions to this rule are: other well recognized standards such as IETF RFCs or content hosted in decentralized persistant storage.
 
 ## Linking to other EIPs
 


xinbenlv:
Requesting reviews from SamWilsn@ MicahZoltu@, gcolvin@ who authored / reviewed #4872

EIPS/eip-1.md
```diff
@@ -185,9 +185,9 @@ The `created` header records the date that the EIP was assigned a number. Both h
 
 EIPs may have a `requires` header, indicating the EIP numbers that this EIP depends on.
 
-## Linking to External Resources
+## Direct Linking to External Resources
```
SamWilsn, line 188: For consistency with the other header, I dislike this change.

```suggestion
## Linking to External Resources
```

EIPS/eip-1.md
```diff
@@ -185,9 +185,9 @@ The `created` header records the date that the EIP was assigned a number. Both h
...
 
-Links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
+Excersise extreme caution when adding external resources. Direct link e.g. [URI](https://datatracker.ietf.org/doc/html/rfc3986) **SHOULD NOT** be included. External resources, if hosted centralized, may disappear, move, or change unexpectedly or manipulated by adversary, creating security and abuse harm. Exceptions to this rule are: other well recognized standards such as IETF RFCs or content hosted in decentralized persistant storage.
```
SamWilsn, line 190: Your suggestion is actually pretty close to my original philosophy for this change. After some discussion with Micah, I came to agree that reviewing what services fulfill the requirements (durable, compatible licenses, etc) is probably too much work for the editors we have.

That said, it might be worth opening a discussion thread on eth magicians to hammer out a couple exceptions (like IETF RFCs.)

MicahZoltu: I'm very weakly in favor of allowing RFC and IETF specifications, but I am a bit worried that just the act of *creating* such a list is a slippery slope to having to maintain such a list.  People will definitely want to add <insert source here> to that list once it exists.  Also, I currently do allow links to consensus specs, so that should be on the list too...

gcolvin: Here is what the IETF has to say about the Reference Section.  It's brief:

> The use of URIs in references is acceptable, as long as the URI is the most stable (i.e., unlikely to change and expected to be continuously available) and direct reference possible.  The URI will be verified as valid during the RFC editorial process.
>
> If a dated URI (one that includes a timestamp for the page) is available for a referenced web page, its use is required.
>
> Note that URIs may not be the sole information provided for a  reference entry.

The entire style guide is worth a gander.

> RFC  7322 
> RFC Style Guide
> https://www.rfc-editor.org/rfc/rfc7322.html
> https://www.rfc-editor.org/rfc/rfc7322.html#page-14

lightclient (commented):
I am weakly against. **SHOULD NOT** is a flexible enough directive in my opinion, no need to start naming resources. I'm particularly against the linking to decentralized "persistent" storage.

PR #4904, opened 2022-03-11
EIP-4844: v -> y_parity like EIP-1559, and fix intrinsic gas

protolambda:
- `v` is now encoded as `y_parity`, like EIP-1559 (and the AL tx type also) does.
- intrinsic gas now adds `Gtxcreate` (yellow paper gas cost for contract creation by tx) and access-list intrinsic gas.

Thanks to [yperbasis](https://ethereum-magicians.org/u/yperbasis) on eth-magicians for reporting these two issues.

This PR is getting auto-merged, but happy to adjust again if anything is wrong/missing.

The `21000` -> `20000` intrinsic gas cost reduction was already there since the original blob-transactions proposal by Vitalik ([here](https://notes.ethereum.org/@vbuterin/blob_transactions)). I realize we may need to add an explanation, I asked the other authors for clarification. In general it should encourage usage of the new SSZ-encoded transaction type.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -153,7 +153,7 @@ class AccessTuple(Container):
     storage_keys: List[Hash, MAX_ACCESS_LIST_STORAGE_KEYS]
 
 class ECDSASignature(Container):
-    v: uint8
+    y_parity: boolean
     r: uint256
     s: uint256
 ```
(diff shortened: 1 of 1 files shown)


FlaShib1fan813:
0x2b591e99afE9f32eAA6214f7B7629768c40Eeb39

PR #4913, opened 2022-03-14
EIP-3156: fix return type in implementation

hjorthjort:


MicahZoltu:
@lightclient The reference implementation doesn't match the specification.  The reference implementation has a return type of `bytes32` while the reference implementation has a return type of `bool`.  The reference implementation does correctly return a `bytes32` but as written it will be collapsed into a `bool` (incorrectly).  I do think this change is correct and reasonable for a final EIP, as it corrects the reference implementation to match the specification (and to be internally consistent).

fubuloubu (approved):

SamWilsn (approved):

PR #4925, opened 2022-03-16
Eip 3651: Stagnant -> Review

wjmelements:
3651 is being considered for inclusion in Shanghai.
#### Changes
Changes EIP-3651 from Stagnant to Review


gcolvin (approved):

wjmelements:
automerging seems to have failed

MicahZoltu (commented):
Simple summary isn't a valid EIP section.  Header needs a `description` section (which probably would just contain the simple summary text in this case).

PR #4937, opened 2022-03-23
docs: add function overloading

YamenMerhi:
- Adding function overloading to `getData(...)` and `setData(...)` to be backward compatible and provide a more efficient way to get and set single data from/to a smart contract.
- Will be cheaper by **2k~3k** gas unit.


frozeman (approved):

frozeman:
The reasoning behind this update is to make it cheaper for smart contracts to read and set data on a ERC726Y supporting smart contract. As this standard will likely be the basis of a lot of other standards in the future.

PR #4942, opened 2022-03-24
mark EIP-1153 as review

moodysalem:


Pandapip1:
@MicahZoltu are external links to eth magicians allowed?

MicahZoltu (changes requested):
External link in Security Considerations should be removed.  I recommend extracting the critical information from that thread and including it in the EIP directly.

Otherwise this looks good to move to Review.

MicahZoltu:
Good catch @Pandapip1!

Pandapip1:
Test cases should be removed until there is text

Pandapip1:
I've suggested a few changes: #4947

moodysalem:
> * There are missing `requires` in the header

This header is optional according to eip-template.md

I've approved changes from @Pandapip1 and added some small language nits/some additional context to security considerations to this PR

Pandapip1:
> This header is optional according to eip-template.md

It's optional because not every EIP references other EIPs. If you reference an EIP, you should put the header in.

MicahZoltu (approved):

PR #4951, opened 2022-03-28
Remove mention of rationale containing "proof of consensus".

MicahZoltu:
We have a pretty strict policy that EIPs are purely technical standards and *NOT* the place to establish (or prove) consensus.  There are many EIPs that have not achieved consensus and they are perfectly valid technical standards, even though they are not implemented in Ethereum today.


MicahZoltu:
cc @lightclient @axic @gcolvin @SamWilsn

gcolvin:
I'd approve this, but don't seem to have that option. Consensus can't be "proved", and is generally reached as part of the process of review, after the Rationale is written.  Exceptions can be handled by the more general policy of making reasonable exceptions, which is a different discussion.

gcolvin:
> @gcolvin You should be able to click the files tab, then "Review Changes" in the top right and submit a comment/approve/reject review. That being said, no bot will be involved with this one, so a "verbal" approval is fine. 😄

The unmistakable big green button is there now.  Must have been a Github glitch that made it go missing.

gcolvin (approved):

MicahZoltu:
Ah, right.  I forgot that the bot currently allows changes to EIP-1 with a single editor approval.  😬   we really should fix that...

PR #4960, opened 2022-03-31
eip 1459: stagnant -> review

holiman:
This PR moves EIP 1459 from `Stagnant` to `Review`. I would like to move it to final, but I suspect that would not be allowed. 

The DNS discovery has been live on Ethereum mainnet, goerli, ropsten and rinkeby since February 2020.

A crawler is traversing all those networks, and harvesting node records (ENRs), which it the makes public in two ways: in a [git repository](https://github.com/ethereum/discv4-dns-lists) and via DNS. Examples of such dns lists are `all.goerli.ethdisco.net` for all types of goerli-nodes, `snap.goerli.ethdisco.net` for all snap-capable goerli nodes, or `les.goerli.ethdisco.net` for all public les-servers on goerli. 

(The EIP itself does not specify the naming conventions, which in the case of `ethdisco.net` uses different top level domains to denote different capability lists)

### DIFF
--- a/EIPS/eip-1459.md
+++ b/EIPS/eip-1459.md
@@ -4,7 +4,7 @@ title: Node Discovery via DNS
 author: Felix Lange <fjl@ethereum.org>, Péter Szilágyi <peter@ethereum.org>
 type: Standards Track
 category: Networking
-status: Stagnant
+status: Review
 created: 2018-09-26
 requires: 778
 discussions-to: https://github.com/ethereum/devp2p/issues/50


holiman:
cc @fjl

MicahZoltu:
I created a PR to fixup some things that are now required of EIPs, sadly GitHub doesn't let me apply them as comments to this PR since they are not related to changes being made in this pull request.

https://github.com/holiman/EIPs/pull/3

MicahZoltu (changes requested):
You are correct, stagnant => review is appropriate.  I think we have also allowed stagnant => last call for things that are already deployed if you wanted, though you can easily just move from Review to Last Call once this PR is merged.

MicahZoltu (approved):
Need an approval from @fjl or @karalabe , but otherwise looks good to me.

holiman:
@MicahZoltu I applied your changes but reverted the addition of me as author.

karalabe (approved):

MicahZoltu:
Overriding the bot and manually merging because the original only had email addresses for authors, not GitHub handles, so the bot won't auto-merge without Felix's approval which isn't necessary here (since Peter already approved).

PR #4959, opened 2022-03-31
Add EIP: Standardize Ethereum Drink Menu

SamWilsn:

PR #4964, opened 2022-04-01
Fix problem with EIP editor overwork

Pandapip1:
This might need a bit of fine-tuning


Pandapip1:
I'm thinking about this, and it would actually be simpler for the EIP editors if they didn't have to do anything. Is there a way to make the @eip-bot enable auto-merge for every PR so that the EIP editors can just enjoy #4959?

PR #4967, opened 2022-04-02
Finally add signer to auth [EIP-3074]

SamWilsn:


lightclient (approved):
LGTM!

EIPS/eip-3074.md
```diff
@@ -56,39 +53,66 @@ The variable has the same scope as the program counter -- `authorized` persists
...
-| `top - 1`  | `yParity`    |
-| `top - 2`  | `r`          |
-| `top - 3`  | `s`          |
+| `top - 1`  | `authority`  |
+| `top - 2`  | `offset`     |
+| `top - 3`  | `length`     |
+
+##### Memory
+
+The final two stack arguments (`offset` and `length`) describe a range of memory. The format of the contents of that range is:
```
jochem-brouwer, line 71: The rationale for this is that it is now possible to also support other signature types, right? If that is the case, how should these be distinguished? It is clear that `memory[offset : offset+32]` is either all zeros or 31 zero bytes and 1 byte representing `1`, so it would be possible to do some versioning there to signal whatever signature type is used. I am not sure if this versioning should be baked in from the start, though :thinking:

EIPS/eip-3074.md
```diff
@@ -56,39 +53,66 @@ The variable has the same scope as the program counter -- `authorized` persists
...
 | ---------- | -------------|
 | `top - 0`  | `authorized` |
 
+##### Memory
+
+Memory is not modified by this instruction.
+
 #### Behavior
 
+If `length` is greater than 96, the extra bytes are ignored. Bytes outside the range (in the event `length` is less than 96) are treated as if they had been zeroes.
```
jochem-brouwer, line 91: This is only the case when we are using an ECDSA signature, right?

EIPS/eip-3074.md
SamWilsn: > If that is the case, how should these be distinguished?

The target address will eventually know how to validate its own signature format.

EIPS/eip-3074.md
SamWilsn: Right, if we ever change the signature format, this paragraph would have to be changed.

EIPS/eip-3074.md
jochem-brouwer: Ah right, this makes sense.

EIPS/eip-3074.md
```diff
@@ -56,39 +53,66 @@ The variable has the same scope as the program counter -- `authorized` persists
...
-| `top - 3`  | `s`          |
+| `top - 0`  | `authority`  |
+| `top - 1`  | `offset`     |
+| `top - 2`  | `length`     |
+
+##### Memory
+
+The final two stack arguments (`offset` and `length`) describe a range of memory. The format of the contents of that range is:
+
+ - `memory[offset    : offset+32 ]` - `yParity`
```
jochem-brouwer, line 72: Can this be reduced to one byte to save memory?

EIPS/eip-3074.md
```diff
@@ -56,39 +53,66 @@ The variable has the same scope as the program counter -- `authorized` persists
...
 
-The gas cost for `AUTH` is `3100`. This is equal to the cost for the `ecrecover` precompile, plus a bit extra to cover a keccak256 hash and some additional logic.
+The gas cost for `AUTH` is equal to the sum of:
+
+ - fixed fee `3100`.
+ - memory expansion gas cost (`auth_memory_expansion_fee`)
+
+The fixed fee is equal to the cost for the `ecrecover` precompile, plus a bit extra to cover a keccak256 hash and some additional logic.
+
+The memory expansion gas cost (`auth_memory_expansion_fee`) shall be calculated in the same way as `RETURN`, where memory is expanded if the specified range is outside the current allocation.
```
jochem-brouwer, line 115: So "standard" quadratic expansion costs, right?

EIPS/eip-3074.md
```diff
@@ -170,6 +194,16 @@ As with `CALL`, the full gas cost is charged immediately, independently of actua
...
 
+### Signature in Memory
+
+The signature format (`yParity`, `r`, and `s`) is fixed, so it might seem curious that `auth` accepts a dynamic memory range. The signature is placed in memory so that `auth` can be upgraded in the future to work with contract accounts (which might use non-ECDSA signatures) and not just EOAs.
+
+### Signing Address `auth` Argument
+
+Including `authority` (the signing address) as an argument to `auth` allows future upgrades to the instruction to work with contract accounts, and not just EOAs.
+
+If `authority` were not included and multiple signature schemes allowed, it would not be possible to compute the authorizing account's address from just the signature alone.
```
jochem-brouwer, line 205: I'm no expert here, this does make sense, but if there is a signing scheme which would also return 20-byte addresses then it is not known by the address format which signing scheme should be used. I'm not sure if this is a limitation?

EIPS/eip-3074.md
```diff
@@ -56,39 +53,66 @@ The variable has the same scope as the program counter -- `authorized` persists
...
+
 The arguments (`yParity`, `r`, `s`) are interpreted as an ECDSA signature on the secp256k1 curve over the message `keccak256(MAGIC || chainId || paddedInvokerAddress || commit)`, where:
  - `chainId` is the current chain's [EIP-155](./eip-155.md) unique identifier padded to 32 bytes.
  - `paddedInvokerAddress` is the address of the contract executing `AUTH` (or the active state address in the context of `CALLCODE` or `DELEGATECALL`), left-padded with zeroes to a total of 32 bytes (ex. `0x000000000000000000000000AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA`).
  - `commit`, one of the arguments passed into `AUTH`, is a 32-byte value that can be used to commit to specific additional validity conditions in the invoker's pre-processing logic (e.g. a nonce for replay protection).
 
-Signature validity and signer recovery is handled analogous to transaction signatures, including the stricter `s` range for preventing ECDSA malleability. Note that `yParity` is expected to be `0` or `1`.
+Signature validity and signer recovery is handled analogously to transaction signatures, including the stricter `s` range for preventing ECDSA malleability. Note that `yParity` is expected to be `0` or `1`.
```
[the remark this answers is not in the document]
SamWilsn: The note here is just to differentiate `yParity` from `v` (which is often not `0` or `1`.)

Pretty sure that you can't actually use `26`/`27` with ECDSA signatures, so yeah, this'll just fail.

EIPS/eip-3074.md
SamWilsn: Exactly.

EIPS/eip-3074.md
SamWilsn: You're totally right. If you don't know the scheme, you don't know how to derive the address. That's why you have to pass in `authority` into `AUTH`.

EIPS/eip-3074.md
```diff
@@ -56,39 +53,66 @@ The variable has the same scope as the program counter -- `authorized` persists
...
-| `top - 0`  | `authorized` |
+| `top - 0`  | `success`    |
+
+##### Memory
+
+Memory is not modified by this instruction.
 
 #### Behavior
 
+If `length` is greater than 128, the extra bytes are ignored. Bytes outside the range (in the event `length` is less than 128) are treated as if they had been zeroes.
```
jochem-brouwer, line 91: @SamWilsn this implies a limit of 128 bytes on the memory of this opcode. What if there is a signature which uses more than 128 bytes?

jochem-brouwer: This is also not very clear regarding the gas schedule. What is meant by "ignored"? I am implementing this, and this question has to do with the gas schedule. Is the memory length capped at 128? Or, if the length is larger, and the memory length (relative to offset) is smaller than that, should that memory be filled with zeros? Or is it the length indeed capped at 128?

SamWilsn: > @SamWilsn this implies a limit of 128 bytes on the memory of this opcode.

I hope that it implies that the opcode only reads the first 128 bytes.

> What if there is a signature which uses more than 128 bytes?

That would be defined in the EIP which introduces the new signature scheme.

maciej-spec (commented):
@

PR #4906, opened 2022-03-13
EIP4906: ERC-721 Metadata Update Extension

0xanders:
This standard is an extension of ERC-721. It proposes adding a `MetadataUpdate` event to ERC-721 tokens.

PR #5008, opened 2022-04-15
EIP5008: ERC-721 Nonce Extension

emojidao:
This standard is an extension of ERC-721.  It proposes adding a `nonce` function to ERC-721 tokens.


EIPS/eip-5008.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Abstract
+
+This standard is an extension of [ERC-721](./eip-721.md). It proposes adding a `nonce` property and `MetadataUpdate` event to ERC-721 tokens.
+
+## Motivation
+
+Some orders of NFT marketplace has been attacked and the NFTs have been sold in a lower price than market floor price. One reason is that users transfer NFT to another wallet and then, after a certain period of time, transfer it back to the original wallet, and the order becomes valid again.
+
+This EIP proposes adding an `nonce` property to ERC-721 tokens, and the `nonce` will be changed when transfer. If `nonce` is added to an order, the order can be checked to avoid attacks.
```
OnahProsperity, line 22: Will suggest since this PR has been proposed for merge the following is what I noticed.
`nonce is a word` so his EIP proposes adding `an` before `nonce` according to [https://www.merriam-webster.com/dictionary/nonce](nonce).
Thanks for the new improvement.

EIPS/eip-5008.md
```diff
@@ -0,0 +1,174 @@
+---
+eip: 5008
+title: ERC-721 Nonce and Metadata Update Extension
+description: Add a `nonce` function an a `MetadataUpdate` event to ERC-721.
```
SamWilsn, line 4:
```suggestion
description: Add a `nonce` function and a `MetadataUpdate` event to ERC-721.
```

EIPS/eip-5008.md
```diff
@@ -0,0 +1,174 @@
...
+requires: 165, 721
+---
+
+## Abstract
+
+This standard is an extension of [ERC-721](./eip-721.md). It proposes adding a `nonce` function and `MetadataUpdate` event to ERC-721 tokens.
+
+## Motivation
+
+Some orders of NFT marketplace has been attacked and the NFTs have been sold in a lower price than market floor price. One reason is that users transfer NFT to another wallet and then, after a certain period of time, transfer it back to the original wallet, and the order becomes valid again.
```
SamWilsn, line 20:
```suggestion
Some orders of NFT marketplaces have been attacked and the NFTs sold at a lower price than the current market floor price. The can happen when users transfer an NFT to another wallet and, later, back to the original wallet. This reactivates the order, which may list the token at a much lower price than the owner would have intended.
```

EIPS/eip-5008.md
```diff
@@ -0,0 +1,174 @@
...
+
+This EIP proposes adding a `nonce` property to ERC-721 tokens, and the `nonce` will be changed when transfer. If a `nonce` is added to an order, the order can be checked to avoid attacks.
```
SamWilsn, line 22:
```suggestion
This EIP proposes adding a `nonce` property to ERC-721 tokens, and the `nonce` will be changed when a token is transferred. If a `nonce` is added to an order, the order can be checked to avoid attacks.
```

EIPS/eip-5008.md
```diff
@@ -0,0 +1,174 @@
...
+
+Many [ERC-721](./eip-721.md) contracts emit their custom event when metadata changed. It is easy to update metadata of one NFT by specific event, but it is difficult for third-party platforms such as NFT marketplace to update metadata of many NFTs based on custom events.
```
SamWilsn, line 24:
```suggestion
Many [ERC-721](./eip-721.md) contracts emit a custom event when a token's metadata is changed. It is easy to update metadata of one NFT by specific event, but it is difficult for third-party platforms such as NFT marketplace to update metadata of many NFTs based on custom events.
```

EIPS/eip-5008.md
```diff
@@ -0,0 +1,174 @@
...
+## Specification
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY" and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+```solidity
+interface IERC5008  {
+    /// @dev This event emits when the metadata of a token is changed.
+    /// So that the third-party platforms such as NFT market could
+    /// timely update the images and related attributes of the NFT
+    event MetadataUpdate(uint256 tokenId);
```
SamWilsn, line 37: It almost feels like `MetadataUpdate` should be in a separate EIP from `nonce`. They _seem_ completely independent?

0xanders: > 

Yes, `MetadataUpdate`  in a separate EIP  is better. I reopened  https://github.com/ethereum/EIPs/pull/4906

EIPS/eip-5008.md
```diff
@@ -0,0 +1,174 @@
...
+    /// @return The nonce of the NFT
+    function nonce(uint256 tokenId) external view returns(uint256);
+}
+```
+The `nonce(uint256 tokenId)` function MUST be implemented as `view`.
+
+The `UpdateNonce` event MUST be emitted when the nonce of a NFT is changed.
+
+The `MetadataUpdate` event MUST be emitted when the metadata of a token is changed.
+
```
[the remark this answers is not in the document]
emojidao: > 
Updated.

The `supportsInterface` method MUST return `true` when called with `0xce03fdab`.

wschwab:
fwiw this makes sense to me, and I'm for it getting merged in as Draft

SamWilsn (approved):

0xanders (approved):

PR #5022, opened 2022-04-20
EIP-5022: Increase price of SSTORE from zero to non-zero to 40k gas

greenlucid:


Pandapip1:
@greenlucid - please close this PR and then immediately re-open it

EIPS/eip-5022.md
```diff
@@ -0,0 +1,67 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2022-04-20
+discussions-to: https://ethereum-magicians.org/t/eip-proposal-increase-cost-of-sstore-from-20k-to-x-when-creating-new-storage/7614
+---
+
+### Abstract
+
+Increase the price of the SSTORE opcode from `20_000` gas to `40_000` gas when the original slot is zero and the resultant slot is non-zero.
```
MicahZoltu, line 14:
```suggestion
Increase the price of the SSTORE (0x55) opcode from `20_000` gas to `40_000` gas when the original slot is zero and the resultant slot is non-zero.
```
Opcodes should be referenced at least the first time in the document in the form:
> REVERT (0xfe)

EIPS/eip-5022.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Specification
+
+| Constant | Value |
+| - | - |
+| `FORK_BLOCK` | TBD |
+| `NEW_STORAGE_PRICE` | `40_000`
+
+For blocks where `block.number >= FORK_BLOCK`, a new gas schedule applies. Make `SSTORE_SET_GAS`, the price when a slot is set from zero to non-zero, equal `NEW_STORAGE_PRICE`. All other costs remain the same.
+
```
MicahZoltu, line 54: Specification should come before rationale.
```suggestion
## Specification

| Constant | Value |
| - | - |
| `FORK_BLOCK` | TBD |
| `NEW_STORAGE_PRICE` | `40_000`

For blocks where `block.number >= FORK_BLOCK`, a new gas schedule applies. Make `SSTORE_SET_GAS`, the price when a slot is set from zero to non-zero, equal `NEW_STORAGE_PRICE`. All other costs remain the same.

## Rationale

### Why not also raise the cost of non-zero to non-zero?

Rewriting storage does not affect state growth, which is the main issue this EIP is addressing. Rewriting storage may also be underpriced.
Increasing the price of state growth will, at least, incentivize developers to reuse storage instead.

### Why not also increase the gas refund from setting non-zero to zero?

More discussion is needed on this.

### Why not a better state solution?

Whereas solutions like state rent, or state expiry have been researched for a long time, they will not be ready on the short to medium term. So, it is desirable to patch pricing for the short term. Opcode repricing has been done before, so it should not impose a large development time investment for clients.

### Why was that specific amount chosen?

The current pricing was made off a naive approach of benchmarking opcodes in a laptop. Not only it did not consider the long term problem of having the same price for a resource that costs more over time, the benchmark itself was wrong. This price is closer to what the naive original benchmark should have been. It could go higher, but that may be too disruptive.

### Is this too distruptive?

This change will severely impact the gas cost of many applications. The network does not have to subsidize state growth at the expense of more expensive regular transactions, so even if it is too disruptive, it will increase the health of the network.

```

EIPS/eip-5022.md
```diff
@@ -0,0 +1,67 @@
...
+### Backwards compatibility
```
MicahZoltu, line 55:
```suggestion
## Backwards Compatibility
```

EIPS/eip-5022.md
```diff
@@ -0,0 +1,67 @@
...
+
+Contracts that depend on hardcoded gas costs will break if they create state.
+
+It is a gas schedule change, so transactions from an epoch before FORK_BLOCK should be treated with previous gas costs.
+
+## Implementation
+
+https://github.com/ethereum/go-ethereum/pull/24725
+
```
MicahZoltu, line 64: Implementation isn't a valid EIP section.  There is a `Reference Implementation` section, but it should contain an inline implementation, not a link to an external one.  Reference Implementation is optional, so if it doesn't make sense to include something inline, then it is best to just leave the section out.
```suggestion
```

EIPS/eip-5022.md
```diff
@@ -0,0 +1,67 @@
...
+## Security considerations
+
+TODO
```
MicahZoltu, line 67: Copyright is required.
```suggestion
## Security Considerations
TBD

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```

EIPS/eip-5022.md
```diff
@@ -0,0 +1,67 @@
+---
+eip: 5022
+title: Increase price of SSTORE from zero to non-zero to 40k gas
+author: Green (@greenlucid)
+status: Draft
+type: Standards Track
+category: Core
+created: 2022-04-20
+discussions-to: https://ethereum-magicians.org/t/eip-proposal-increase-cost-of-sstore-from-20k-to-x-when-creating-new-storage/7614
```
MicahZoltu, line 9: Needs a description and `discussions-to` must come after the `author`.
```suggestion
eip: 5022
title: Increase price of SSTORE from zero to non-zero to 40k gas
description: <required>
author: Green (@greenlucid)
discussions-to: https://ethereum-magicians.org/t/eip-proposal-increase-cost-of-sstore-from-20k-to-x-when-creating-new-storage/7614
status: Draft
type: Standards Track
category: Core
created: 2022-04-20
```

MicahZoltu (approved):
Just a few minor adjustments, generally seems fine for draft though.

MicahZoltu (changes requested):
Oops, I meant request changes.  😖

PR #5020, opened 2022-04-19
Fix broken link & automatically redirect

Pandapip1:
- https://eips.ethereum.org/EIPS/eip-20.md actually 404s now. The backup link has been moved to https://eips.ethereum.org/EIPS/eip-20.
- Adds a script tag that simulates an HTTP redirect on page load.

Issue #5054, opened 2022-04-29
Redirect legacy URLs


Pandapip1:
I'm not sure about the feasibility of this, but I imagine that it should be possible with a middleware (perhaps Cloudflare rules?):

The previous format for EIP files (eip-number-postfix.md) is no longer used. However, there may be dead links to those pages (such as with eip-20). I suggest that all paths that match the following regex (which are legacy URLs)

```regex
^(.*\/EIPS\/eip-\d+)[^\d\s\/][^\/]*$
```

be redirected to the value of the first capturing group (which will be the current EIP path).

See the following link for a demonstration: https://regexr.com/6klaf

See also: #5052

MicahZoltu:
Closing this in favor of focusing the discussion on #5056.

Issue #5052, opened 2022-04-29
Should `eip-20-token-standard.md` be deleted?


Pandapip1:
`eip-20-token-standard.md` is the only file in the `EIPS` folder with a postfix. The file sticks out like a sore thumb, and it should be removed for consistency.

Are there data that show how many page views would be lost by removing it?

Pandapip1:
Superseded by #5056

PR #5055, opened 2022-04-29
Force usage of included LICENSE file

Pandapip1:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

Fixes https://github.com/ethereum/EIPs/issues/5031

PR #5056, opened 2022-04-29
Automatic redirection of old-type URLs

Pandapip1:
Fixes #5020, #5054, #5052

### DIFF
--- a/404.html
+++ b/404.html
@@ -2,6 +2,15 @@
 layout: default
 ---
 
+<script>
+  let theUrl = window.location.href;
+  let theRegex = /^([a-zA-z][a-zA-z0-9\-]*:\/\/[a-zA-z][a-zA-z0-9\-.\/]*\/EIPS\/eip-[0-9]+)[^0-9\s].*$/gm;
+  let theNewUrl = theRegex.exec(theUrl);
+  if (theNewUrl !== null) {
+    window.location.href = theNewUrl[1];
+  }
+</script>
+
 <style type="text/css" media="screen">
   .container {
     margin: 10px auto;


Pandapip1:
Huh, I did not expect that to work first try. I am pleasantly surprised.

Tests:
https://pandapip1.github.io/EIPs/EIPS/eip-20
https://pandapip1.github.io/EIPs/EIPS/eip-20.md
https://pandapip1.github.io/EIPs/EIPS/eip-20adsfgfdafgsdfg
https://pandapip1.github.io/EIPs/EIPS/eip-721testingthatthisworksforeipsotherthan20
https://pandapip1.github.io/EIPs/EIPS/eip-notaneip

Pandapip1:
I'm temporarily changing the GitHub pages branch to #5055 to debug why the LICENCE file isn't working.

MicahZoltu:
This feels like unnecessary complexity.  I think EIP-20 is the only EIP that would match the regex, and no one should be using that link anymore.  I would rather just set an EOL on the mis-named EIP-20 and delete it once we reach that EOL.

Pandapip1:
Is there any harm in merging this? It's not adding _too_ much additional complexity.

MicahZoltu:
> This also has the advantage of also redirecting links that mistakenly add `.md`.

Is this a common problem?  Do we have examples of live links that include `.md`?  Why `.md` but not `.txt` (or any other possible typo)?

While I agree that 6 lines isn't much, many instances of "just 6 lines" can really add up in a project over time and that can turn into technical debt and bit rot.  In general, I think that any feature addition, even if it is a single character code change, must reach a certain bar of necessity before being included in a project and to me it feels like this particular issue doesn't quite meet that bar.

Pandapip1:
> Is this a common problem? Do we have examples of live links that include .md? Why .md but not .txt (or any other possible typo)?

It fixes the second typo too. If the user puts `.html`, `.markdown`, or even `.thisisnotevenarealfileextensionoratleastitprobablyisntusedbyanyapplication`, it'll properly redirect. Same with extra endings (e.g. `eip-20-token_standard` or something that never existed such as `eip-4834-never_had_an_ending`. Same with combinations: `eip-20-token_standard.md` also works.

I doubt it's a common problem. I very much don't think my code will contribute to code rot at all (it's pretty well documented), but I can add another comment.

Editor CC: @lightclient @SamWilsn @axic @gcolvin

404.html
```diff
@@ -2,6 +2,15 @@
 layout: default
 ---
 
+<script>
+  let thePath = window.location.pathname;                   // Fetches the path (e.g. "/EIPS/eip-20-token-standard" or "/core")
+  let theRegex = /^(.*\/EIPS\/eip-\d+)[^\d\s\/][^\/]*$/gm;  // This regex matches the EIP path (e.g. "/EIPS/EIP-20") if the original path isn't valid
```
SamWilsn, line 7: While I don't think this is exploitable (because you use `pathname`), I am uncomfortable with a redirect to a location in the URL.

lightclient:
@Pandapip1 thank you for this contribution! I think it is a good idea, maybe something to consider if we move the ERCs out of this repository. However, it doesn't seem terribly important given the current state of the repo. As most editors are against the change, I'm going to close this PR.

PR #5072, opened 2022-05-05
Update EIP-725

YamenMerhi:
- Adding function signature for ERC725Y functions
- Add an example to use specific notations when calling ERC725Y overloaded functions
- Mark standard as Review instead of draft
- Change the terms keys and values to dataKeys and dataValues.
- Change the term key/value store to data key/value store. (Use key/value store instead of key-value store)
- Edit `Executed` event to use bytes4 instead of bytes

### DIFF
--- a/EIPS/eip-725.md
+++ b/EIPS/eip-725.md
@@ -12,7 +12,7 @@ created: 2017-10-02
 
 ## Simple Summary
 
-A standard interface for a smart contract based account with attachable key value store.
+A standard interface for a smart contract based account with attachable data key/value store.
 
 ## Abstract
 
@@ -25,7 +25,7 @@ The standard is divided into two sub standards:
 Can execute arbitrary smart contracts and deploy other smart contracts.
 
 `ERC725Y`:
-Can hold arbitrary data through a generic key/value store.
+Can hold arbitrary data through a generic data key/value store.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


frozeman (approved):

frozeman:
@lightclient, @axic, @SamWilsn

EIPS/eip-725.md
```diff
@@ -1,9 +1,9 @@
 ---
 eip: 725
-title: General key-value store and execution standard
+title: General data key/value store and execution standard
```
SamWilsn, line 3: We're discouraging "standard" in titles and replaced "Simple Summary" with descriptions now:

```suggestion
title: General data key/value store and execution
description: An interface for a smart contract based account with attachable data key/value store
```

EIPS/eip-725.md
```diff
@@ -1,9 +1,9 @@
...
 author: Fabian Vogelsteller (@frozeman), Tyler Yasaka (@tyleryasaka)
 discussions-to: https://github.com/ethereum/EIPs/issues/725
-status: Draft
+status: Review
 type: Standards Track
 category: ERC
 requires: 165, 173
```
SamWilsn, line 9:
```suggestion
category: ERC
created: 2017-10-02
requires: 165, 173
```

Header fields are out of order (and I can't grab the existing `created:` header in this suggestion :sweat_smile:)

EIPS/eip-725.md
```diff
@@ -12,7 +12,7 @@ created: 2017-10-02
 
 ## Simple Summary
```
SamWilsn, line 14:
```suggestion
```

EIPS/eip-725.md
```diff
@@ -12,7 +12,7 @@ created: 2017-10-02
...
 
-A standard interface for a smart contract based account with attachable key value store.
+A standard interface for a smart contract based account with attachable data key/value store.
```
SamWilsn, line 15:
```suggestion
```

SamWilsn (changes requested):
Other required changes I can't comment directly on:
 - The "Rationale" section should describe individual choices made in this EIP. It currently sounds like it should be in the "Motivation" section, which describes the reason why the EIP as a whole is necessary.
 - You'll need to add a "Backwards Compatibility" section. Can be quite brief.
 - You'll need to add a "Security Considerations" section. Can be quite brief.
 - Rename the "Implementation" section to "Reference Implementation".
 - Remove all external links. You can include your implementation and documents with compatible licenses in the `../assets/eip-725` directory.

frozeman:
@SamWilsn we will update the document with all requested changes. 

Though there is one that I do not agree with and thats the removal of all external links:

ERC725 is a base standard, that references other smart contract standards, that are not standardised here in the ERC repository. Specifically [LSPs, which are "LUKSO Standard Proposals"](https://github.com/lukso-network/LIPs/tree/main/LSPs).

The reason why they are not part of the ERC canon, is that the LSP standards are part of a larger set of core standards that are all interoperable and cohesive together. Nonetheless they do function equally for any EVM chain, the ability to make these standards understood requires a less noisy and neutral place, than the current ERC space (that I started). Therefore I deem a separate repo (LIPs/LSPs) as necessary for the success and transparency around these new sets of standards (many of which use ERC725Y as a core basis).

ERC725Y should reference [LSP-2](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md), which specifically defines a way of *how* ERC725Y data keys and values are encoded and decoded. As well as link to [LSP0 ERC725Account](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-0-ERC725Account.md), which is the first proper followup standard of ERC725.

Additionally, the ERC725 alliance also maintains an updated implementation of the ERC725 standard, that needs to be referenced from the standard document for guidance. Additionally a basic implementation will be added to the `../assets/eip-725` as well.

MicahZoltu:
Why not write this as a LUKSO Standard instead of an Ethereum standard if it depends on LUKSO standards?

The reason we have a no external links policy is because of:
* mutability - EIPs are immutable once final, not everything on the web follows that
* simplicity - an EIP that links to a bunch of external resources is more complex than an EIP that doesn't
* limited editor bandwidth - we don't have enough editors to get into a deeply nuanced argument about every external link source that people want to reference.

SamWilsn:
I have to agree with @MicahZoltu here. We try to make it so EIPs can be implemented with only this repository. If you want, you could convert LSP0 and LSP2 to EIPs instead, and link to them that way?

z0r0z:
it seems helpful to reorder the preferred operations with logic of call, delegatecall, staticcall, create, create2? @frozeman 
compare,
image

SamWilsn:
Please take a look at the check results over here: https://github.com/ethereum/EIPs/runs/7369111505?check_suite_focus=true

You can use \`backticks\` around interface names (ex. \`IERC725Y\`) to get `eipw` to ignore them.

Pandapip1:
Closing since the author hasn't been seen in >3 months (the stale threshold).

PR #5085, opened 2022-05-09
EIP-3540: Remove Simplified Implementation section

chfast:
This implementation variant does not handle some of the cases we
currently care about. E.g. section size should never be 0.

### DIFF
--- a/EIPS/eip-3540.md
+++ b/EIPS/eip-3540.md
@@ -296,8 +296,6 @@ All cases should be checked for creation transaction, `CREATE` and `CREATE2`.
 
 ## Reference Implementation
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-3540.md
```diff
@@ -350,32 +350,6 @@ def validate_eof(code: bytes):
     assert len(code) == (pos + section_sizes[S_CODE] + section_sizes[S_DATA])
 ```
 
-### Simplified Implementation
```
axic, line 353: Please remove line 299 (`### Generic Implementation`) as well.

axic (approved):

PR #5086, opened 2022-05-09
Update EIP-1271: Fix example

PhABC:


Ivshti (approved):

MicahZoltu:
I'm very confused by this PR.  It allegedly is a PR against `master`, but in `master` this EIP is final rather than last call and this PR shows the EIP still in Last Call.

@PhABC Would you mind checking out your fork and making sure it is otherwise in sync with `ethereum/EIPs` and that your branch is rebased on master?  GitHub shouldn't be having trouble here... so I'm just guessing at what might be wrong.

SamWilsn:
There's a merge commit in this PR that could be messing things up? I'd recommend resetting this branch to the latest `master` and cherry-picking the changes you want (without the merge commit.)

MicahZoltu:
Is there some reason why you recommend cherry picking rather than `git rebase` @SamWilsn?

SamWilsn:
> Is there some reason why you recommend cherry picking rather than `git rebase` @SamWilsn?

I've had bad experiences rebasing branches that contain merge commits, but that might just be a me thing.

MicahZoltu:
Ah, good point.  One way to deal with that would be to do:
```
# reset branch pointer to point at `origin/master`, but leave working directory untouched.
git reset --soft origin/master
# commit working directory charges to local branch
git add .
git commit -m '...'
# force push local branch to origin
git push --force-with-lease origin
```

Ivshti:
hey @MicahZoltu @SamWilsn @PhABC, is this now good to merge?

MicahZoltu:
This PR still has something weird going on with it as it is allegedly against a Last Call EIP but this EIP is final.  Someone needs to fix this, or submit a new issue that is correctly against the current head EIP.

SamWilsn:
Are we still interested in getting this merged? I'd prefer getting the git history cleaned up before doing so.

Pandapip1 (approved):

Pandapip1:
The Git history is fine for me.

EIPS/eip-1271.md
```diff
@@ -127,12 +127,7 @@ Example implementation of a signing contract:
     }
 
     // Recover ECDSA signer
-    signer = ecrecover(
-      keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", _hash)),
-      v,
-      r,
-      s
-    );
+    signer = ecrecover(_hash, v, r, s);
```
SamWilsn, line 130: Probably a dumb question, but is the old version broken in some way, or just less-than-ideal? From a quick read, it _looks_ like the code from before this PR would be secure, but would display a hash while signing instead of human readable text.

SamWilsn:
> The Git history is fine for me.

You've got to stop making merge commits :rofl:

EIPS/eip-1271.md
xinbenlv: I support this editorial change from @PhABC . 

I think the EIP-1271 is more general. 

Original text adding the "\x19Ethereum Signed Message:\n32" part of signature makes it confusing for reader. Some readers might deem such additional msg packing for `_hash` as necessary, but in fact it's not necessary in the spec of 1271.

xinbenlv (approved):

xinbenlv:
@Pandapip1 @SamWilsn I think the latest editorial policy is to ignore EIP Walidator warnings for _Final_ EIPs, right?

xinbenlv:
Waiting for manual merge

PR #5093, opened 2022-05-13
EIP-3074: Clarify that ignored bytes still cost gas

SamWilsn:
@jochem-brouwer hopefully this clarifies the wording a bit!


jochem-brouwer:
Great, thanks!!

PR #5098, opened 2022-05-17
ERC4907： Change status to Last Call

emojidao:
Change status to Last Call


0xanders (approved):

lightclient (changes requested):
In "Backward Compatibility" please replace "ERC721" with "ERC-721". Looks good otherwise.

ed738 (approved):

ed738 (approved):

0xanders:
Hi @lightclient  ,  I see this pr has been approved at [EIP editor apprenticeship meeting](https://github.com/ethereum-cat-herders/EIPIP/issues/135), could you pls approve this pr and set the last call date?

EIPS/eip-4907.md
```diff
@@ -4,7 +4,7 @@ title: Rental NFT, ERC-721 User And Expires Extension
 description: Add a time-limited role with restricted permissions to ERC-721 tokens.
 author: Anders (@0xanders), Lance (@LanceSnow), Shrug <shrug@emojidao.org>
 discussions-to: https://ethereum-magicians.org/t/idea-erc-721-user-and-expires-extension/8572
-status: Review
+status: Last Call
```
MicahZoltu, line 7:
```suggestion
status: Last Call
last-call-deadline: YYYY-MM-DD
```
This needs a last call deadline.  It should be *at least* 14 days in the future.  Due to editor response times being a bit slow for ERCs, I recommend setting it 3-4 weeks in the future so you don't run into problems.

lightclient (approved):

MicahZoltu (commented):
````
run in Terminal：
```
truffle test ./test/test.js
```
````
the chances that this works at an undefined point in the future approach 0 as the distance between now and then approach infinity.  In other words, please remove this as it will almost certainly break, and doesn't provide significant value toward standardization.

0xGh:
Hi there folks! I just learnt about this proposal and I wanted to chime in and ask what would be the best place to chat about design decisions eg. `user` vs `holder` or why `useOf` doesn't return a tuple or a struct

MicahZoltu:
> what would be the best place to chat about design decisions

The discussions-to link, which is in the header of the EIP: https://ethereum-magicians.org/t/idea-erc-721-user-and-expires-extension/8572

PR #5105, opened 2022-05-24
Fixes "spelling error" in diff

MicahZoltu:
I was hoping that this file would fail once, not every time someone submits a PR.  Unfortunately I was wrong.  I have no idea if this will invalidate the patch, but I'm optimistic that the diff is lenient on line lengths and having a few more characters than the diff's default line length won't break anything.  😬


MicahZoltu:
Could I get a quick review from @axic @SamWilsn @lightclient or @gcolvin please?  Currently, Travis is failing on every PR because of this "spelling error".

SamWilsn (approved):

PR #5106, opened 2022-05-24
EIP-4844: Change to versioned hashes

ppopth:
In the `validate_blob_transaction_wrapper` function, it uses `blob_commmitments` which doesn't make sense. It makes more sense when it uses `blob_versioned_hashes` instead.


asn-d6:
Looks good to me! Thanks!

asn-d6:
Looks good to me! Thanks!

asn-d6 (approved):

PR #5109, opened 2022-05-26
Update EIP-4750: Static stack use validation for EOF functions.

gcolvin:

 - The build passes.

### DIFF
--- a/EIPS/eip-4750.md
+++ b/EIPS/eip-4750.md
@@ -123,7 +123,7 @@ In addition to container format validation rules above, we extend code section v
 
 Rules 1 through 4 can be validated with a simple scan of the bytecode.
 
-Rules 5, 6, and 7 require a symbolic execution of the code.  This algorithm does not doesn't consider the code’s data and computations, only its control flow and stack use.  This means we will reject programs with invalid code paths, even if those paths are not reachable.
+Rules 5, 6, and 7 require a symbolic execution of the code.  This algorithm does not consider the code’s data and computations, only its control flow and stack use.  This means we will reject programs with invalid code paths, even if those paths are not reachable.
 
 _Note: a similar algorithm is sketched out in EIP-615, and will eventually be given here._
 
--- a/EIPS/eip-4750.md
+++ b/EIPS/eip-4750.md
@@ -141,7 +141,7 @@ Rules 5, 6, and 7 require a symbolic execution of the code -- a Reference Implen
 
 ### `RETF` in the top frame ends execution vs exceptionally halts
 
-execution, because there is arguably no caller for the starting function. This would mean that return stack is initialized as empty, and `RETF` exceptionally aborts when return stack is empty.
+Alternative logic for executing `RETF` in the top frame could be to exceptionally halt execution, because there is arguably no caller for the starting function. This would mean that the return stack is initialized as empty, and `RETF` exceptionally aborts when the return stack is empty.
 
 We have decided in favor of always having at least one item in the return stack, because it allows to avoid having a special case for empty stack in the interpreter loop stack underflow check. We keep the stack underflow rule general by having `caller_stack_height = 0` in the top frame.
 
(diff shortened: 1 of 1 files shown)


Pandapip1 (commented):
Just to test the new CI change

EIPS/eip-4750.md
```diff
@@ -2,7 +2,7 @@
 eip: 4750
 title: EOF - Functions
 description: Individual sections for functions with `CALLF` and `RETF` instructions
-author: Andrei Maiboroda (@gumb0), Alex Beregszaszi (@axic), Paweł Bylica (@chfast)
+author: Andrei Maiboroda (@gumb0), Alex Beregszaszi (@axic), Paweł Bylica (@chfast), Greg Colvin, Brooklyn Zelenka (@expede)
```
Pandapip1, line 5: CC'ing existing authors: @gumb0 @axic @chfast
CC'ing potential new authors: @gcolvin @expede
Can you all confirm you are okay with this change?

EIPS/eip-4750.md
```diff
@@ -95,21 +94,36 @@ Under `PC_post_instruction` we mean the PC position after the entire immediate a
...
-2. If data stack has less than `caller_stack_height + types[code_section_index].outputs`, execution results in exceptional halt.
-3. Pops nothing and pushes nothing to data stack.
-4. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.
+2. Pops nothing and pushes nothing to data stack.
+3. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.
     4.1. If return stack is empty after this, execution halts with success.
 
 ### Code Validation
 
+If the execution of an instruction would violate a condition the execution is in an exceptional halting state.  The [Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) defines five such states.
```
Pandapip1, line 103:
```suggestion
If the execution of an instruction would violate a condition, then the execution is in an exceptional halting state. The [Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) defines five such states.
```
No external links.

EIPS/eip-4750.md
```diff
@@ -137,9 +151,94 @@ This change poses no risk to backwards compatibility, as it is introduced only f
...
+```
+
+ _Note: Python translation welcomed with gratitude._
+
 ## Security Considerations
 
-TBA
+For our purposes we define a safe EVM contract as one that cannot encounter an exceptional halting state. From the standpoint of security it would be best if unsafe contracts were never placed on the blockchain. Unsafe code can attempt to overflow stack, underflow stack, execute invalid instructions, and jump to invalid locations.  Unsafe code is an exploit waiting to happen.  We cannot guarantee that all valid code is safe, but we can guarantee that valid contracts will not halt with an exception unless they throw out of gas or overflow stack.
+
+
```
Pandapip1, line 241:
```suggestion
```

EIPS/eip-4750.md
```diff
@@ -127,7 +141,7 @@ In addition to container format validation rules above, we extend code section v
 
 ### `RETF` in the top frame ends execution vs exceptionally halts
 
-Alternative logic for executing `RETF` in the top frame could be to exceptionally halt execution, because there is arguably no caller for the starting function. This would mean that return stack is initialized as empty, and `RETF` exceptionally aborts when return stack is empty.
+execution, because there is arguably no caller for the starting function. This would mean that return stack is initialized as empty, and `RETF` exceptionally aborts when return stack is empty.
```
Pandapip1, line 144:
```suggestion
Alternative logic for executing `RETF` in the top frame could be to exceptionally halt execution, because there is arguably no caller for the starting function. This would mean that the return stack is initialized as empty, and `RETF` exceptionally aborts when the return stack is empty.
```
(I'm pretty sure you didn't mean to remove that first part).

EIPS/eip-4750.md
MicahZoltu: Personally, I'm generally fine with one existing author approving (don't need all of them).  I tend to agree that the authors being added should all agree.

Pandapip1: > one existing author approving
I would prefer half, rounded up.

gumb0: I would prefer these changes to be a separate EIP, requiring this one.

gumb0:
We generally agree with the direction taken by these changes, and in fact plan to publish our own separate EIP with very similar idea, maybe with somewhat different implementation.

@gcolvin However I'm a bit confused whether this one is still your preferred apporach or did you switch to implementing static validation on top of EIP-2315, and this one should be just closed?

EIPS/eip-4750.md
gcolvin: I put this up here so that we could have a precise proposal to discuss.  I don't expect it to be merged in its present form.

gcolvin:
> @gcolvin  is still your preferred apporach or did you switch to implementing static validation on top of EIP-2315, and this one should be just closed?

I'm not concerned with the exact approach, just that as much be checked at validation time as possible.   If you have an EIP that does the job then by all means close this PR.  EIP-2315 and EIP-4759 are independent  -- one can go in whether or not the other goes in.  EIP-2315 now specifies it's own validation, so if they both go in there will be some work to harmonize the redundant specs.

Issue #735, opened 2017-10-09
ERC: Claim Holder


frozeman:
```
EIP: 735
Title: Claim Holder
Author: Fabian Vogelsteller (@frozeman)
Type: Standard
Category: ERC
Status: Discussion
Created: 2017-10-09
```

**NOTE: Due to the changes in [ERC725](https://github.com/ethereum/EIPs/issues/725), this spec is not fully compatible with the current ERC725v2. If you're interested in adopting this spec to work with 725v2, please comment below, or send a gist with changes.**

# Abstract

The following describes standard functions for adding, removing and holding of claims.
These claims can attested from third parties (issuers) or self attested.

# Motivation

This standardised claim holder interface will allow Dapps and smart contracts to check the claims about a claim holder. Trust is here transfered to the issuers of claims.

# Definitions

- `claim issuer`: is another smart contract or external account, which issues claims about this identity. The claim issuer can be an identity contract itself.
- `claim`: A claim is an information an issuer has about the identity holder. This contains the following:
    - `topic`: A `uint256` number which represents the topic of the claim. (e.g. 1 biometric, 2 residence  (ToBeDefined: number schemes, sub topics based on number ranges??))
    - `scheme `: The scheme with which this claim SHOULD be verified or how it should be processed. Its a `uint256` for different schemes. E.g. could `3` mean contract verification, where the `data` will be call data, and the `issuer` a contract address to call (ToBeDefined). Those can also mean different key types e.g. 1 = ECDSA, 2 = RSA, etc. (ToBeDefined)
    - `issuer`: The issuers identity contract address, or the address used to sign the above signature. If an identity contract, it should hold the key with which the above message was signed, if the key is not present anymore, the claim SHOULD be treated as invalid. The issuer can also be a contract address itself, at which the claim can be verified using the call `data`.
    - `signature`: Signature which is the proof that the claim issuer issued a claim of `topic ` for this identity. it MUST be a signed message of the following structure: `keccak256(address identityHolder_address, uint256 _ topic, bytes data)` // or `keccak256(abi.encode(identityHolder_address, topic, data))` ?
    - `data`: The hash of the claim data, sitting in another location, a bit-mask, call data, or actual data based on the claim scheme.
    - `uri`: The location of the claim, this can be HTTP links, swarm hashes, IPFS hashes, and such.

# Specification

#### claim structure

The claims issued to the identity. Returns the claim properties.

``` js
struct Claim {
    uint256 topic;
    uint256 scheme;
[...]

PR #5114, opened 2022-05-30
Creates EIP-5114: Soulbound Token

MicahZoltu:


EIPS/eip-5114.md
```diff
@@ -0,0 +1,64 @@
...
+	event Mint(uint256 indexed tokenId, address indexed erc721Token, uint256 indexed erc721TokenId);
+
+	// returns the ERC-721 token that owns this token.
+	// this function **MUST** throw if the token hasn't been minted yet
+	// this function **MUST** always return the same result every time it is called after it has been minted
+	// this function **MUST** return the same value as found in the original `Mint` event for the token
+	function ownerOf(uint256 index) external view returns (address erc721Token, uint256 erc721TokenId);
+	
+	// returns a censorship resistant URI with details about this token collection
+	// the metadata returned by this is merged with the metata return by `tokenUri(uint256)`
```
TimDaub, line 31: "metadata"

EIPS/eip-5114.md
```diff
@@ -0,0 +1,64 @@
...
+	// data from `tokenUri` takes precedent over data returned by this method
+	// any external links referenced by the content at `tokenUri` also **MUST** follow all of the above rules
+	function collectionUri() external view returns (string tokenUri);
+	
+	// returns a censorship resistant URI with details about this token instance
+	// the tokenUri **MUST** be immutable and content addressable (e.g., ipfs://)
+	// the tokenUri **MUST NOT** point at mutable/censorable content (e.g., https://)
+	// data from this takes precedent over data returned by `collectionUri`
+	// any external links referenced by the content at `tokenUri` also **MUST** follow all of the above rules
+	function tokenUri(uint256 tokenId) external view returns (string tokenUri);
```
TimDaub, line 43: Wouldn't using `tokenURI` here be better than `tokenUri` as then the bytes4 signature of the function address in Solidity is equal to e.g. that of EIP-721?

TimDaub:
What is your take on EIP-165 and do you plan to use it here?

EIPS/eip-5114.md
MicahZoltu:
```suggestion
	// the metadata returned by this is merged with the metadata return by `tokenUri(uint256)`
```

EIPS/eip-5114.md
MicahZoltu: I considered it, but I don't think 721 compatibility matters much because the `ownerOf` signature is different and owner lookups require custom code, so this token won't natively work with existing ERC-721 tooling.

Perhaps there are specific tools that don't need to do owner lookups but do need to URI lookups that would benefit from consistency across methods?

*Personally* I would like to see a compelling reason to break the superior (IMO) naming convention of capitalizing only the first letter of acronyms (it avoids situations like `XMLHTTPRPC` with `XmlHttpRpc`).

MicahZoltu:
> Could an EIP-5114 Soulbound token be bound to another EIP-5114 token? Are hierarchies permitted? What about loops?

I am a big fan of allowing hierarchies.  In fact, I wonder if this EIP should be modified such that it allows for binding to not only ERC-721 tokens but also to ERC-5114 tokens so you can have hierarchies?

In theory loops are possible I suppose, but it would given the strict rules about immutability, the whole loop would need to be created at once.  At first glance, this doesn't seem to be a problem worth resolving, but it may be good to mention it in the security considerations section that tooling should look out for it and protect itself from getting caught in an infinite loop/recursion.

MicahZoltu:
I have added a note about loops to security considerations.  I also changed it to just require binding to an NFT, with EIP-721 being merely an example.  The only real requirement is that the thing being bound to has an `address` and a `tokenId`.

EIPS/eip-5114.md
TimDaub: IMO naming convention arguments should be out of scope of this EIP. But a separate EIP could make normative statements on function capitalization conventions.

I way mainly arguing for consistency with EIP-721 but it's not a hard objection from my side.

MicahZoltu:
> to send a PR on EIP-165, would I send it to this origin/branch: MicahZoltu-patch-1?

The simplest option is to just click on a line or drag-select a handful of lines and leave a comment with a suggestion block like:
````
```suggestion
<proposed changes go here>
```
````
That will put the suggested change inline in the PR that can then be one-click committed.

Another option is to do as you suggested and submit a PR against my branch.

A final option is to wait until this is merged as a Draft, and then submit a PR to modify the EIP once it actually exists.

File: EIPS/eip-5114.md [hunk 3]
```diff
@@ -0,0 +1,66 @@
...
+### Immutability
+By requiring that tokens can never move, we both guarantee non-seperability and non-mergability among collections of soulbound tokens that are bound to a single NFT while simultaneously allowing users to aggressively cache results.
+### Content Addressible URIs Required
+Soulbound tokens are meant to be permanent badges/indicators attached to a persona.  This means that not only can the user not transfer ownership, but the minter also cannot withdraw/transfer/change ownership as well.  This includes mutating or removing any remote content as a means of censoring or manipulating specific users.
+### No Specification for tokenUri Data Format
+The format of the data pointed to by `collectionUri()` and `tokenUri(uint256)` is intentionally left out of this standard in favor of separate standards that can be iterated on in the future.  The immutability constraints are the only thing defined by this to ensure that the spirit of this token is maintained, regardless of the specifics of the data format.
+
+## Backwards Compatibility
+This is a new tokene type and is not meant to be backward compatible with any existing tokens other than using ERC-721 as souls.
```
KillariDev, line 56: tokene -> token (typo)

EIPS/eip-5114.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Security Considerations
+Users of tokens that claim to implement this EIP must be diligent in verifying they actually do.  A token author can create a token that, upon initial probing of the API surface, may appear to follow the rulse when in reality it doesn't.  For example, the contract could allow transfers via some mechanism and simply not utilize them initially.
```
KillariDev, line 59: rulse-> rules (typo)

EIPS/eip-5114.md
```diff
@@ -0,0 +1,66 @@
...
+	// returns the NFT token that owns this token.
+	// this function **MUST** throw if the token hasn't been minted yet
+	// this function **MUST** always return the same result every time it is called after it has been minted
+	// this function **MUST** return the same value as found in the original `Mint` event for the token
+	function ownerOf(uint256 index) external view returns (address nftAddress, uint256 nftTokenId);
+	
+	// returns a censorship resistant URI with details about this token collection
+	// the metadata returned by this is merged with the metadata return by `tokenUri(uint256)`
+	// the tokenUri **MUST** be immutable and content addressable (e.g., ipfs://)
+	// the tokenUri **MUST NOT** point at mutable/censorable content (e.g., https://)
```
wschwab, line 33: as much as I appreciate the sentiment, it seems out-of-scope to me to specify the nature of the storage, as it is outside the Ethereum network

the requirement for immutability is also questionable to me - I could see a use for a URI that can deliberately be changed or change itself over time

EIPS/eip-5114.md
```diff
@@ -0,0 +1,66 @@
...
+eip: 5114
+title: Soulbound Token
+description: A token that is attached to a "soul" at mint time and cannot be transferred after that.
+author: Micah Zoltu (@MicahZoltu)
+discussions-to: https://ethereum-magicians.org/t/eip-5114-soulbound-token/9417
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-05-30
+requires: 721
```
KillariDev, line 11: Is this requirement necessary? I believe you can bind EIP5114 to any address?

EIPS/eip-5114.md
```diff
@@ -0,0 +1,66 @@
...
+---
+
+## Abstract
+A soulbound token is a token that is bound to another Non Fungible Token (NFT; e.g., a EIP-721 token) when it is minted, and cannot be transferred/moved after that.
```
lightclient, line 15:
```suggestion
A soulbound token is a token that is bound to another Non-Fungible Token (NFT; e.g., a EIP-721 token) when it is minted, and cannot be transferred/moved after that.
```

EIPS/eip-5114.md
lightclient, line 59: Please use one space after each period.

lightclient (changes requested):

EIPS/eip-5114.md
MicahZoltu:
```suggestion
This is a new token type and is not meant to be backward compatible with any existing tokens other than using ERC-721 as souls.
```

EIPS/eip-5114.md
MicahZoltu: The EIP references EIP-721, so it requires it (not all editors agree on this policy).

EIPS/eip-5114.md
MicahZoltu: If we don't constrain the storage, then the tokens functionally become mutable.  By asserting immutability throughout, we enable applications/users to build caching systems that they cannot built if we only encourage the behavior.

EIPS/eip-5114.md
MicahZoltu: Fixed.

EIPS/eip-5114.md
TimDaub: Immutable data doesn't exist. From what I understand, only content-integrity assuring identifiers exist that can indicate mutation. Replication can be used to have non-mutated backups.

MicahZoltu: I'm not following your claim here that immutable data doesn't exist.  If you ask IPFS for something by content hash, you will get either that exact thing, or you will get nothing (if it can't be found on the network).  There is no situation where you would get a different result from someone else or at a different point in time.

Regarding censorable content: I tend to agree that censorable content is fuzzy here, and I wish we could nail down something a bit more concrete.  What I want to avoid is people linking to content that a provider can remove at a future point in time.  The goal with these badges is that they are permanent, they aren't something that comes and goes over time.  If you link out to a traditional `https` resource, the host of that resource/owner of the domain can remove it.  Similarly, the domain may be seized and changed or taken away.  Something like IPFS on the other hand allows anyone to ensure the content survives, you don't have to rely on a single operator.

TimDaub: "immutable data"/content-addressable criteria:

I agree with you that some networks of computers can create the illusion that some data is immutable. But on the basis of a principle, the data isn't immutable. E.g. on IPFS/Ethereum/Bitcoin when you run a full node, there's nothing stopping you from going into the chain data directory and switching a few bits. You can always mutate the data. It's likely though that the system maintaining the data will notice and e.g. notably fail or repair itself. That's what I meant by "Replication can be used to fallback on non-mutated backups (in a distributed system)."

How I'd describe it is that by requesting data using content-addressing, a user can guarantee to themselves that they'll either receive the exact data corresponding to the content-address, or nothing at all (e.g. when content-address and data don't align).

I think "content-addressing" is a better term than e.g. "immutable".

data availability guarantees/non-censorable:

I'd strongly opt for e.g. "highly-available data" rather than "non-censorable." From my perspective, censorship is about withholding information in the interest of a few. Data availability, on the other hand, is a term from the blockchain research world that describes the concept of persisting data within e.g. a p2p. E.g. all EVM storage has high data availability, and anything stored in the Bitcoin blockchain too.

For IPFS, actually, it doesn't have data availability built-in. To pin an IPFS content hash, one has to personally deploy a dedicated node that continues to make the content hash available within the distributed hash table. IMO, if you wanted the ultimate data availability for Soulbound Badges, it'd probably be best to store the contents of `tokenUri` on-chain as e.g. a struct. I'm not kidding. It's because if you outsource delivering extremely high-quality data availability to other networks not aligned with Ethereum (e.g. IPFS), it may be that one day they just go down/stop working while Ethereum still works.

Now talking about storing the badge's metadata on-chain, I think it's obligatory I bring up https://github.com/ethereum/EIPs/issues/735

MicahZoltu: Lets move this conversation over to the discussions-to link as I'm hoping this will get merged as a draft soon and I don't want this conversation to get lost.

EIPS/eip-5114.md
KillariDev: Hmm, in abstract it says that EIP5114 requires it to be bound to another NFT, and gives example of EIP721. In section "Backwards Compatibility" Its stated that EIP721 HAS to be used as a soul. I think the EIP should be more specific if EIP721 as a soul is mandatory, or is some other token or EIP also acceptable.

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+	// the tokenUri **MUST** be immutable and content addressable (e.g., ipfs://)
+	// the tokenUri **MUST NOT** point at mutable/censorable content (e.g., https://)
+	// data from `tokenUri` takes precedent over data returned by this method
+	// any external links referenced by the content at `tokenUri` also **MUST** follow all of the above rules
+	function collectionUri() external view returns (string tokenUri);
+	
+	// returns a censorship resistant URI with details about this token instance
+	// the tokenUri **MUST** be immutable and content addressable (e.g., ipfs://)
+	// the tokenUri **MUST NOT** point at mutable/censorable content (e.g., https://)
+	// data from this takes precedent over data returned by `collectionUri`
```
KillariDev, line 41: precedent - > precedence

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+	// any external links referenced by the content at `tokenUri` also **MUST** follow all of the above rules
+	function tokenUri(uint256 tokenId) external view returns (string tokenUri);
+}
+```
+
+## Rationale
+### Immutability
+By requiring that tokens can never move, we both guarantee non-seperability and non-mergability among collections of soulbound tokens that are bound to a single NFT while simultaneously allowing users to aggressively cache results.
```
KillariDev, line 49: non-seperability -> non-separability

KillariDev: non-mergability -> non-mergeability

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+### Content Addressible URIs Required
```
KillariDev, line 50: Addressible -> Addressable

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+	// this function **MUST** throw if the token hasn't been minted yet
+	// this function **MUST** always return the same result every time it is called after it has been minted
+	// this function **MUST** return the same value as found in the original `Mint` event for the token
+	function ownerOf(uint256 index) external view returns (address nftAddress, uint256 nftTokenId);
+	
+	// returns a censorship resistant URI with details about this token collection
+	// the metadata returned by this is merged with the metadata return by `tokenUri(uint256)`
+	// the tokenUri **MUST** be immutable and content addressable (e.g., ipfs://)
+	// the tokenUri **MUST NOT** point at mutable/censorable content (e.g., https://)
+	// data from `tokenUri` takes precedent over data returned by this method
```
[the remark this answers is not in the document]
MicahZoltu:
```suggestion
	// data from `tokenUri` takes precedence over data returned by this method
```

EIPS/eip-5114.md
MicahZoltu:
```suggestion
	// data from this takes precedence over data returned by `collectionUri`
```

EIPS/eip-5114.md
MicahZoltu:
```suggestion
By requiring that tokens can never move, we both guarantee non-separability and non-mergeability among collections of soulbound tokens that are bound to a single NFT while simultaneously allowing users to aggressively cache results.
```

EIPS/eip-5114.md
MicahZoltu:
```suggestion
### Content Addressable URIs Required
```

EIPS/eip-5114.md
MicahZoltu: Any EIP that matches the interface `ownerOf(uint256 id)` I think is the actual requirement.  EIP-721 is an example of that, as is EIP-5114.

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+### Content Addressible URIs Required
+Soulbound tokens are meant to be permanent badges/indicators attached to a persona.
+This means that not only can the user not transfer ownership, but the minter also cannot withdraw/transfer/change ownership as well.
+This includes mutating or removing any remote content as a means of censoring or manipulating specific users.
+### No Specification for tokenUri Data Format
+The format of the data pointed to by `collectionUri()` and `tokenUri(uint256)` is intentionally left out of this standard in favor of separate standards that can be iterated on in the future.
+The immutability constraints are the only thing defined by this to ensure that the spirit of this token is maintained, regardless of the specifics of the data format.
+
+## Backwards Compatibility
+This is a new token type and is not meant to be backward compatible with any existing tokens other than using ERC-721 as souls.
```
MicahZoltu, line 59:
```suggestion
This is a new token type and is not meant to be backward compatible with any existing tokens other than existing viable souls (like EIP-721 tokens).
```

EIPS/eip-5114.md
KillariDev: Does this mean that if EIP-5114 gets minted to an address that does not implement `ownerOf(uint256 id)`, then the EIP-5114 contract is no longer EIP-5114? Or can it be said that the contract is still EIP-5114 but the standard is not obeyed for this single instance?

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+	// this function **MUST** return the same value as found in the original `Mint` event for the token
+	function ownerOf(uint256 index) external view returns (address nftAddress, uint256 nftTokenId);
+	
+	// returns a censorship resistant URI with details about this token collection
+	// the metadata returned by this is merged with the metadata return by `tokenUri(uint256)`
+	// the tokenUri **MUST** be immutable and content addressable (e.g., ipfs://)
+	// the tokenUri **MUST NOT** point at mutable/censorable content (e.g., https://)
+	// data from `tokenUri` takes precedence over data returned by this method
+	// any external links referenced by the content at `tokenUri` also **MUST** follow all of the above rules
+	function collectionUri() external view returns (string tokenUri);
```
SamWilsn, line 36: Should all of these `tokenUri` references (and return variable name) be `collectionUri`? Seems a bit odd to return a `tokenUri` from the `collectionUri` function.

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-05-30
+requires: 721
+---
+
+## Abstract
+A soulbound token is a token that is bound to another Non-Fungible Token (NFT; e.g., a EIP-721 token) when it is minted, and cannot be transferred/moved after that.
+
```
SamWilsn, line 16: I'm not sure why this EIP needs to exist. Are there many use cases for nontransferable tokens? Why should implementers prefer this EIP over, say, [EIP-4973](https://eips.ethereum.org/EIPS/eip-4973)?

Seems like this EIP needs a Motivation section ;)

EIPS/eip-5114.md
```diff
@@ -0,0 +1,75 @@
...
+## Backwards Compatibility
+This is a new token type and is not meant to be backward compatible with any existing tokens other than existing viable souls (like EIP-721 tokens).
+
+## Security Considerations
+Users of tokens that claim to implement this EIP must be diligent in verifying they actually do.
+A token author can create a token that, upon initial probing of the API surface, may appear to follow the rules when in reality it doesn't.
+For example, the contract could allow transfers via some mechanism and simply not utilize them initially.
+
+It should also be made clear that Soulbound tokens are not bound to a human, they are bound to a persona.
```
SamWilsn, line 66: Should check for consistent capitalization of Soulbound

EIPS/eip-5114.md
MicahZoltu:
```suggestion
	// returns a censorship resistant URI with details about this token collection
	// the metadata returned by this is merged with the metadata return by `tokenUri(uint256)`
	// the collectionUri **MUST** be immutable and content addressable (e.g., ipfs://)
	// the collectionUri **MUST NOT** point at mutable/censorable content (e.g., https://)
	// data from `tokenUri` takes precedence over data returned by this method
	// any external links referenced by the content at `collectionUri` also **MUST** follow all of the above rules
	function collectionUri() external view returns (string collectionUri);
```

EIPS/eip-5114.md
MicahZoltu: I see what you did there.  I may consider adding a motivation after this has merged as a draft, but as it isn't required I would like to move forward without it for now.

EIPS/eip-5114.md
MicahZoltu:
```suggestion
It should also be made clear that soulbound tokens are not bound to a human, they are bound to a persona.
```

yu-tou:
Some suggestions and opinions:
1. Does soulbound token binding bind humans to address, or address to address? Or is tokenId and address bound? This is a topic worth exploring.
2. How to unbind soulbound token after binding? Unreleasable binding may cause some problems, but this kind of release is definitely not arbitrary release, what is it?
3. Does soulbound token need to be compatible with widely used standards such as erc721 and erc20? I think it is necessary, otherwise it will bring a lot of ecological compatibility problems to create a token standard.
4. Who initiated the binding of the soulbound token? How to ensure that no one can generate a soulbound token for an address without consent, it is a serious topic, if someone can generate a soulbound token for you without consent, it will be a very serious disaster.
5. What are the specific scenarios of soulbound token? It needs to be supplemented by some specific scenarios in different fields to define the exact characteristics that SBTs should have.

btw. We are also exploring some scenarios about soulbound token, which can be used as a reference, but our scenario is a further Paired bounding scenario, which creates a new identity for the address and stores the relationship between the paired address at the same time. It is an extend of the standard erc721, and its characteristics are mainly:

* implement from erc721, most NFT usage scenarios are seamlessly compatible
* non-transferable and non-sellable, one person can only have one valid Token at the same time.
* mint paired soulbound token through multi-signature flow, Mint will issue 2 Soulbound Tokens at one time.
* can be destroyed through multi-signature flow, and new Soulbound Token can be minted with other addresses after burn.

Behind the token, we store the relationship between two tokens, and the bounding relationship can be queried through the contract interface.
But we haven't considered making it a standard yet, because his scene is still very limited, and it's more of a social network attempt.

we write a contract at: https://github.com/marryinweb3/ERC721-520
and implement a dapp at: https://marry3.love

we can talk about  soulbound tokens more at here

MicahZoltu:
@yu-tou These are good questions/comments, but they should really be made over at this EIP's discussions-to link (https://ethereum-magicians.org/t/eip-5114-soulbound-badges/9417/17).  Would you mind posting the same thing over there so it doesn't get lost when this PR is merged?

yu-tou:
> 

ok ,thanks

EIPS/eip-5114.md
5cent-AI: We've talked to some developers, and I'm guessing the logic is beyond many people's comprehension. But if you think deeply about some social phenomena, it is the only feasible solution to the problem of SBTs.

5cent-AI: New news, we have improved ERC721QS-1 and proposed ERC721QS-1P, which can also be used for NFT anti-theft, NFT leasing, NFT lending and designing SBTs! We even found cases where a similar approach was already used. We will update the documentation as soon as possible. But ERC721QS-2 saves gas more.

SamWilsn (approved):

PR #5121, opened 2022-06-01
Finalize EIP-2364

### LINKED ISSUE #2365: EIP-2364: eth/64: forkid-extended protocol handshake
This is a discussion thread for https://github.com/ethereum/EIPs/pull/2364

---

**Abstract**

The [`forkid` (EIP-2124)](https://eips.ethereum.org/EIPS/eip-2124) was designed to permit two Ethereum nodes to quickly and cheaply decide if they are compatible or not, not only at a genesis/networking level, but also from the perspective of the currently passed network updates (i.e. forks).

EIP-2124 only defined how the `forkid` is calculated and validated, but due to the numerous possible applications, it did not itself spec out how the `forkid` should be exchanged between peers. This current EIP requests the inclusion of the `forkid` as a new field in the Ethereum wire protocol (`eth`) handshake (releasing a new version, `eth/64`).

By cross validating each others' forks already during the handshake, incompatible nodes can disconnect before expensive block exchanges and validations take place (PoW check, EVM execution, state reconstruction). This furher prevents peer slots from being taken up by nodes that are incompatible, but have not yet been detected as such.

Pandapip1:
CC: @karalabe, @timbeiko

Closes #2365

### DIFF
--- a/EIPS/eip-2364.md
+++ b/EIPS/eip-2364.md
@@ -4,8 +4,7 @@ title: "eth/64: forkid-extended protocol handshake"
 description: Introduces validation of the `forkid` when handshaking with peers.
 author: Péter Szilágyi <peterke@gmail.com>, Péter Szilágyi (@karalabe), Tim Beiko (@timbeiko)
 discussions-to: https://github.com/ethereum/EIPs/issues/2365
-status: Last Call
-last-call-deadline: 2021-11-19
+status: Final
 type: Standards Track
 category: Networking
 created: 2019-11-08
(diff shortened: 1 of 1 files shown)


timbeiko (approved):

MicahZoltu (changes requested):
> Additionally, `forkid` validation via the discovery protocol requires ENR implementation ([EIP-778](./eip-778.md)) and ENR extension support ([EIP-868](./eip-868.md)), which is not mandated by the Ethereum network currently. Lastly, the discovery protocol is just one way to find peers, but systems that cannot use UDP or that rely on other mechanism (e.g. DNS discovery ([EIP-1459](./eip-1459.md))) still need a way to filter connections.

EIP-1459 is still in review, not final, which means it is blocking this from being merged.  Since it is just used as an example, can we just remove it?

> Additionally, `forkid` validation via the discovery protocol requires ENR implementation ([EIP-778](./eip-778.md)) and ENR extension support ([EIP-868](./eip-868.md)), which is not mandated by the Ethereum network currently. Lastly, the discovery protocol is just one way to find peers, but systems that cannot use UDP or that rely on other mechanism (e.g. DNS discovery) still need a way to filter connections.

MicahZoltu:
@timbeiko Can you review/approve again?  The reference to EIP-1459 was removed and I would like to get author sign-off on that change before approving.

timbeiko (approved):

MicahZoltu (approved):

PR #5129, opened 2022-06-02
Add protocol for author list changes

Pandapip1:
Most publications have a protocol for handling author list changes. This PR adds a protocol for this type of change.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -203,6 +203,10 @@ It occasionally becomes necessary to transfer ownership of EIPs to a new champio
 
 If you are interested in assuming ownership of an EIP, send a message asking to take over, addressed to both the original author and the EIP editor. If the original author doesn't respond to the email in a timely manner, the EIP editor will make a unilateral decision (it's not like such decisions can't be reversed :)).
 
+## Changes to the Author List
+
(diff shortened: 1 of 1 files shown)


MicahZoltu:
> Most publications have a protocol for handling author list changes. This PR adds a protocol for this type of change.

I would like to see a more compelling argument for why this is necessary than "because other publications do it".

My default position is that the author list follows the same rules as any other changes to the EIP and doesn't need to be called out as a special case.

EIPS/eip-1.md
```diff
@@ -203,6 +203,10 @@ It occasionally becomes necessary to transfer ownership of EIPs to a new champio
 
 If you are interested in assuming ownership of an EIP, send a message asking to take over, addressed to both the original author and the EIP editor. If the original author doesn't respond to the email in a timely manner, the EIP editor will make a unilateral decision (it's not like such decisions can't be reversed :)).
 
+## Changes to the Author List
+
+Changes to the author list of an EIP in Draft require either approval from either an editor or every author listed in either the old or new author list. Changes to EIPs in Review or Last Call require approval from both an editor and authors, unless the authors cannot be contacted. Changes to the author list of a Stagnant EIP follow the rules for the new status. Changes to the author list of a Final EIP are not allowed. Changes to the author list of a Living EIP only require editor approval.
```
SamWilsn, line 208: > Changes to the author list of an EIP in Draft require either approval from either an editor or every author listed

Too many "either" in this sentence.

If I'm understanding this correctly this would be an increase from any author approving to all authors approving. Why increase the threshold?

> the old or new author list

So a PR with no old authors and ten new authors would be merged if all ten new authors approved without any old authors approving?

---

Overall I think these rules are complex, and I don't see what problem they're solving. Maybe you could explain a bit more why this is necessary?

Pandapip1: Honestly, I don't know what rule they are meant to solve. Now that I think of it, it's probably actually a legal thing to prevent libels and/or misattribution. I'll close this.

PR #5131, opened 2022-06-03
Add EIP-5131: ENS Subdomain Authentication

wwhchung:

 - The build passes.

### DIFF
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -6,7 +6,7 @@ author: Wilkins Chung (@wwhchung)
 discussions-to: https://ethereum-magicians.org/t/eip-603-ens-authentication-link/9458
 status: Draft
 type: Standards Track
-category: Interface
+category: ERC
 created: 2022-06-03
 requires: 137
 --- 
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -12,27 +12,28 @@ requires: 137
 --- 
 
 ## Abstract
-Proving the ownership of an asset to a third party application in the Ethereum ecosystem is a common occurrence. Users frequently sign payloads of data presented to them via these applications in order to prove ownership. The process of proving ownership of an address is commonly used to authenticate users before allowing them to perform some operation.
+This EIP uses the Ethereum Name Service Specification (EIP-137) as a way to link one or more signing wallets (presumably less secure) to prove control and asset ownership of a main wallet.
 
-Examples:
+## Motivation
+Proving the ownership of an asset to a third party application in the Ethereum ecosystem is a common occurrence. Users frequently sign payloads of data presented to them via these applications in order to prove ownership. The process of proving ownership of an address is commonly used to authenticate users before allowing them to perform some operation.  However, the current method of using a person's main wallet for authentication of control and asset ownership is both insecure and inconvenient.  It is as if, in order to authenticate, you must give an application root access to your wallet.
+
+*** Examples: ***
  - In order for you to edit your profile on OpenSea, you must sign a message with your wallet address.
  - In order to access NFT gated content, you must sign a message with the wallet containing the NFT.
  - In order to claim an airdrop, you must interact with the smart contract with the qualifying wallet address.
  - In order to prove ownership of an NFT, you must sign a payload with the address that owns that NFT.
 
+### Problems with existing methods and solutions ###
 Unfortunately, we've seen many cases where users have accidentally signed a malicious payload. The result is almost always a significant loss of assets associated with the signing address.
 
 In addition to this, many users keep significant portions of their assets in 'cold storage'. With the increased security from 'cold storage' solutions, we usually see decreased accessibility because users naturally increase the barriers required to access these wallets.
 
-This EIP proposes a solution which uses the Ethereum Name Service Specification (EIP-137) as a way to link one or more signing wallets (presumably less secure) to authenticate a main wallet.
+Some solutions propose dedicated registry smart contracts to create this link, or new protocols to be supported. This is problematic from an adoption standpoint, and there have not been any standards created for them. 
+
+### Proposal: Use the Ethereum Name Service (EIP-137) ###
+Rather than 're-invent the wheel', this proposal aims to use the widely adopted Ethereum Name Service in order to bootstrap a safer and more convenient way to sign and authenticate, and provide 'read only' access to a main wallet.
 
 From there, the benefits are twofold. This EIP gives users increased security via outsourcing potentially malicious signing operations to wallets that are more accessible (hot wallets), while being able to maintain the intended security assumptions of wallets that are not frequently used for signing operations.
-
-
-## Motivation
-As explained in the abstract, the current method of using a person's main wallet for authentication of control and asset ownership is both insecure and inconvenient.  It is as if, in order to authenticate, you must give an application root access to your wallet.  This proposal provides an easy way to do 'read only' access to a wallet by leveraging existing infrastructure.
-
-Some solutions propose dedicated registry smart contracts to create this link, or new protocols to be supported.  Rather than 're-invent the wheel', this proposal aims to use the widely adopted Ethereum Name Service in order to bootstrap a safer and more convenient way to sign and authenticate.
 
 
 ## Specification
@@ -43,13 +44,13 @@ Let:
  - `mainAddress` represents the wallet address we are trying to authenticate or prove asset ownership for.
  - `mainENS` represents the reverse lookup ENS string for `mainAddress`.
  - `authAddress` represents the address we want to use for signing in lieu of `mainAddress`.
- - `authENS` represents the reverse lookup ENS string for `authAddress`.  It must be in the format `auth[0-9]*.<mainENS>`.
+ - `authENS` represents the reverse lookup ENS string for `authAddress`.  It must be in the format `auth[0-9A-za-z]*.<mainENS>`.
 
 
 ### Setting up one or many `authAddress` records
 The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record and reverse record configured.
 
-1. Using your `mainAddress` wallet, sign into ens.domains and create a subdomain record for `mainENS` called `auth[0-9]*`.  This becomes the `authENS`
+1. Using your `mainAddress` wallet, sign into ens.domains and create a subdomain record for `mainENS` called `auth[0-9A-za-z]*`.  This becomes the `authENS`
 2. Set the ETH resolver record for the subdomain created in step 1 (`authENS`) to the `authAddress`.
 3. Using `authAddress`, sign into ens.domains and set the ENS reverse record to `authENS`
 
@@ -60,7 +61,7 @@ Control of `mainAddress` and ownership of `mainAddress` assets is proven if any 
 
 Practically, this would work by performing the following operations:
 1. Get the reverse ENS record for `authAddress`
-2. Parse `auth[0-9]*.<mainENS>` to determine the linked ENS
+2. Parse `auth[0-9A-za-z]*.<mainENS>` to determine the linked ENS
 3. Do a lookup on the linked ENS record to determine the linked `mainAddress`
 4. MUST get the reverse ENS record for `mainAddress` and verify that it matches `<mainENS>` 
     - Otherwise one could set up other ENS nodes (with auths) that point to mainAddress and authenticate via those.
@@ -101,7 +102,7 @@ async function getLinkedAddress(provider: ethers.providers.Provider, address: st
   const addressENS = await provider.lookupAddress(address);
   if (!addressENS) return null;
 
-  const authMatch = addressENS.match(/^(auth[0-9]*)\.(.*)/);
+  const authMatch = addressENS.match(/^(auth[0-9A-za-z]*)\.(.*)/);
   if (!authMatch) return null;
 
   const linkedENS = authMatch[2];
@@ -153,7 +154,7 @@ library LinkedAddress {
     /**
      * Validate that the message sender is an authentication address for mainAddress
      * @param ensRegistry    Address of ENS registry
-     * @param authENSLabel   The ENS label of the authentication wallet (must be `auth[0-9]*`)
+     * @param authENSLabel   The ENS label of the authentication wallet (must be `auth[0-9A-Za-z]*`)
      * @param mainAddress    The main address we want to authenticate for.
      * @param mainENSParts   The array of the main address ENS domain parts (e.g. wilkins.eth == ['wilkins', 'eth']).
      *                       This is used vs. the full ENS a a single string name hash computations are gas efficient.
@@ -172,7 +173,7 @@ library LinkedAddress {
      *
      * @param ensRegistry    Address of ENS registry
      * @param authAddress    The address of the authentication wallet
-     * @param authENSLabel   The ENS label of the authentication wallet (must be `auth[0-9]*`)
+     * @param authENSLabel   The ENS label of the authentication wallet (must be `auth[0-9A-Za-z]*`)
      * @param mainAddress    The main address we want to authenticate for.
      * @param mainENSParts   The array of the main address ENS domain parts (e.g. wilkins.eth == ['wilkins', 'eth']).
      *                       This is used vs. the full ENS a a single string name hash computations are gas efficient.
@@ -215,12 +216,18 @@ library LinkedAddress {
         require(authResolver != address(0), "Auth ENS not registed");
         require(authAddress == Resolver(authResolver).addr(authNameHash), "Not authenticated");
 
-        // Check that the subdomain name has the correct format auth[0-9]*.
+        // Check that the subdomain name has the correct format auth[0-9A-Za-z]*.
         bytes4 authPart = bytes4(authENSLabel[:4]);
         require(authPart == "auth", "Invalid prefix");
         unchecked {
             for (uint256 i = authENSLabel.length; i > 4; i--) {
-                require(authENSLabel[i] >= 0x30 && authENSLabel[i] <= 0x39, "Invalid char");
+                bytes1 char = authENSLabel[i];
+                require(
+                    (char >= 0x30 && char <= 0x39) ||
+                    (char >= 0x41 && char <= 0x5A) ||
+                    (char >= 0x61 && char <= 0x7A),
+                    "Invalid char"
+                );
             }
         }
 
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -79,8 +79,8 @@ This is made possible because the attendee completed the necessary upfront autho
 
 ### Example 2: Multiple Devices
 I have multiple devices and wish to sign with any of them for convenience.  In this situation:
- - My iPhone has mobile metamask hot wallet, and I can add it as an auth account by setting auth1.wilkins.eth to that wallet (and the corresponding reverse record). 
- - My iPad has the same and can be set to auth2.wilkins.eth, etc. 
+ - My phone has mobile metamask hot wallet, and I can add it as an auth account by setting auth1.wilkins.eth to that wallet (and the corresponding reverse record). 
+ - My tablet has the same and can be set to auth2.wilkins.eth, etc. 
 
 Lost Access to Signing Key/Device: 
 In the event that a user losses access to the signing key of an `authAddress` they can simply delete the corresponding ENS record. An attractive side effect of this is that there is no need to import/reimport seed phrases. 
@@ -119,7 +119,7 @@ If your application operates a secure backend server, you could run the client/s
 #### Without a secure server (web client only)
 Provided is a refrence implementation for an internal function to verify that the message sender has an authentication link to the main address.
 
-``````
+```
 // SPDX-License-Identifier: MIT
 
 pragma solidity ^0.8.0;
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -74,17 +74,16 @@ Practically, this would work by performing the following operations:
 
 Note that this specification allows for both contract level and client/server side validation of signatures.  It is not limited to smart contracts, which is why there is no proposed external interface definition.
 
-## Rationale
-The proposed specification allows one to link multiple addresses as 'authentication addresses' to a core main address. This is beneficial from a security standpoint (if the authentication address is compromised, the assets held by the main address is not), and convenience (if the authentication address is a simple MetaMask wallet but the main address is a hardware wallet).
+### Example Use Cases
 
-### Example 1: Event Access
+#### Example 1: Event Access
 An event requires attendees to prove ownership of an NFT to gain access. The attendee in question only has their phone available to perform a signing operation upon the presentation of a challenge. The phone is considered the attendee's 'hot wallet' which is linked to `auth.wilkins.eth`.
 The NFT is owned by the Ethereum address that corresponds to `wilkins.eth`, which is controlled by the attendee's cold storage device (Ledger, Trezor etc). The attendee does not have this device present at the event.
 The attendee is able to prove ownership of the NFT by signing with the 'hot wallet' key.
 
-This is made possible because the attendee completed the necessary upfront authorization steps with the cold storage device to allow the 'hot wallet' to be used for this operation. 
+This is made possible because the attendee completed the necessary upfront ENS configuration with the cold storage device to allow the 'hot wallet' to be used for this operation. 
 
-### Example 2: Multiple Devices
+#### Example 2: Multiple Devices
 I have multiple devices and wish to sign with any of them for convenience. In this situation:
  - My phone has mobile metamask hot wallet, and I can add it as an auth account by setting auth1.wilkins.eth to that wallet (and the corresponding reverse record). 
  - My tablet has the same and can be set to auth2.wilkins.eth, etc. 
@@ -93,6 +92,19 @@ Lost Access to Signing Key/Device:
 In the event that a user losses access to the signing key of an `authAddress` they can simply delete the corresponding ENS record. An attractive side effect of this is that there is no need to import/reimport seed phrases. 
 In the unfortunate event that you lose access to the signing key of the `mainAddress`, users will lose access to that top-level domain and all related subdomains. As always, this private key data should be treated as highly sensitive. 
 
+## Rationale
+
+### Usage of EIP-137
+The proposed specification makes use of EIP-137 rather than introduce another registry paradigm. The reason for this is due to the existing wide adoption of EIP-137 and ENS.
+
+However, the drawback to EIP-137 is that any linked `authAddress` must contain some ETH in order to set the `authENS` reverse record. This can be solved by a separate reverse lookup registry that enables `mainAddress` to set a reverse record with a message signed by `authAddress`.
+
+### One-to-Many Authenitcation Relationship
+This proposed specification allows for a one (`mainAddress`) to many (`authAddress`) authentication relationship.  i.e. one `mainAddress` can authorize many `authAddress` to authenticate, but an `authAddress` can only authenticate itself or a single `mainAddress`.
+
+The reason for this design choice is to allow for simplicity of authentication via client and smart contract code. You can determine which `mainAddress` the `authAddress` is signing for without any additional user input.
+
+Further, you can design UX without any user interaction necessary to 'pick' the interacting address by display assets owned by `authAddress` and `mainAddress` and use the appropriate address dependent on the asset the user is attempting to authenticate with.
 
 ## Reference Implementation
 
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -287,7 +287,7 @@ library LinkedAddress {
 ```
 
 ## Security Considerations
-The core purpose of this EIP is to enhance security and promote a safer way to authenticate wallet control and asset ownership when the main wallet is not needed and assets held by the main wallet do not need to be moved.  Consider it a way to do 'read only' authentication.
+The core purpose of this EIP is to enhance security and promote a safer way to authenticate wallet control and asset ownership when the main wallet is not needed and assets held by the main wallet do not need to be moved. Consider it a way to do 'read only' authentication.
 
 
 ## Copyright
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -47,6 +47,12 @@ Let:
  - `authENS` represent the reverse lookup ENS string for `authAddress`.  It must be in the format `auth[0-9A-za-z]*.<mainENS>`.
 
 
+Control of `mainAddress` and ownership of `mainAddress` assets by `authAddress` is proven if all the following conditions are met:
+ - `mainAddress` has an ENS ETH resolver record and a reverse record set to `mainENS`.
+ - `mainENS` has a subdomain record matching the format `auth[0-9A-Za-z]*.<mainENS>`. This subdomain record is referred to as `authENS`.
+ - `authAddress` has an ENS reverse record set to `authENS`
+ 
+
 ### Setting up one or many `authAddress` records
 The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record and reverse record configured.
 
@@ -57,7 +63,7 @@ The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record 
 Currently this EIP does not enforce an upper-bound on the number of `authAddress` entries you can include. Users can repeat this process with as many address as they like. 
 
 ### Authenticating `mainAddress` via `authAddress`
-Control of `mainAddress` and ownership of `mainAddress` assets is proven if any one of associated `authAddress` is the msg.sender or has signed the message.
+Control of `mainAddress` and ownership of `mainAddress` assets is proven if any associated `authAddress` is the msg.sender or has signed the message.
 
 Practically, this would work by performing the following operations:
 1. Get the reverse ENS record for `authAddress`
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -74,7 +74,7 @@ Lost my device? No problem, just delete the record.  No need to import/reimport 
 
 ## Reference Implementation
 
-### Client Side
+### Client/Server Side
 In typescript, the validation function, using ethers.js would be as follows:
 ```
 async function getLinkedAddress(provider: ethers.providers.Provider, address: string): Promise<LinkedAddress | null> {
@@ -96,10 +96,42 @@ async function getLinkedAddress(provider: ethers.providers.Provider, address: st
 }
 ```
 
-### Solidity
-In solidity, the validation signature could look like:
-```
+### Contract side
+
+### With a secure server
+If there is a secure, you could run the client/server code above, then use the result in conjunction with EIP-1721 for a cheap and secure way to validate that the the message signer is indeed authenticated for the main address.
+
+#### Without a secure server (web client only)
+Provided is a refrence implementation for an internal function to verify that the message sender has an authentication link to the main address.
+
+``````
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/// @author: manifold.xyz
+
 /**
+ * ENS Registry Interface
+ */
+interface ENS {
+    function resolver(bytes32 node) external view returns (address);
+}
+
+/**
+ * ENS Resolver Interface
+ */
+interface Resolver{
+    function addr(bytes32 node) external view returns (address);
+    function name(bytes32 node) external view returns (string memory);
+}
+
+/**
+ * Validate a signing address is associtaed with a linked address
+ */
+abstract contract LinkedAddress {
+
+    /**
      * Validate that the message sender is an authentication address for the mainAddress
      *
      * @param ensRegistry    Address of ENS registry
@@ -108,10 +140,90 @@ In solidity, the validation signature could look like:
      * @param mainENSParts   The array of the main address ENS domain parts (e.g. wilkins.eth == ['wilkins', 'eth']). 
      *                       This is used vs. the full ENS a a single string name hash computations are gas efficient.
      */
-function validate(address ensRegistry, bytes calldata senderENS, address mainAddress, string[] memory mainENSParts) view returns(bool);
-```
+    function validate(address ensRegistry, bytes calldata senderENS, address mainAddress, string[] memory mainENSParts) internal view returns(bool) {
+        bytes32 mainNameHash = _computeNamehash(mainENSParts);
+        address mainResolver = ENS(ensRegistry).resolver(mainNameHash);
+        require(mainResolver != address(0), "Invalid");
+        require(mainAddress == Resolver(mainResolver).addr(mainNameHash), "Invalid");
+        bytes32 senderReverseNameHash = _computeReverseNamehash();
+        address senderResolver = ENS(ensRegistry).resolver(senderReverseNameHash);
+        require(senderResolver != address(0), "Invalid");
+        string memory senderENSLookup = Resolver(senderResolver).name(senderReverseNameHash);
+        require(keccak256(senderENS) == keccak256(bytes(senderENSLookup)), "Invalid");
 
-This can be an internal validate function to the smart contract, that is used in conjunction with specs like EIP-1721: Standard Signature Validation Method for Contracts and Merkle Proof Verification.
+        // Quick substring comparison
+        // Get the total theoretical length of mainENS
+        bytes memory ensCheckBuffer;
+        for (uint i = mainENSParts.length; i > 0;) {
+            ensCheckBuffer = abi.encodePacked('.', mainENSParts[i-1], ensCheckBuffer);
+            unchecked {
+              i--;
+            }
+        }
+        bytes32 ensCheck = keccak256(ensCheckBuffer);
+        
+        // Length of senderENS must be >= ensCheckBuffer.length+4 (since it needs to be of format auth[0-9]*.mainENS)
+        require(senderENS.length >= ensCheckBuffer.length+4, "Invalid");
+        // Check ending substring of the senderENS matches
+        require(ensCheck == keccak256(senderENS[senderENS.length-ensCheckBuffer.length:]), "Invalid");
+        // Check prefix matches auth[0-9]*.
+        require(keccak256(abi.encodePacked('auth')) == keccak256(senderENS[:4]), "Invalid");
+        for (uint i = senderENS.length-ensCheckBuffer.length; i > 4;) {
+          require(senderENS[0] >= 0x30 && senderENS[i] <= 0x39, "Invalid");
+          unchecked {
+            i--;
+          }
+        }
+        return true;
+    }
+
+    // *********************
+    //   Helper Functions
+    // *********************
+
+    function _computeNamehash(string[] memory _nameParts) private pure returns (bytes32 namehash) {
+        namehash = 0x0000000000000000000000000000000000000000000000000000000000000000;
+        for (uint i = _nameParts.length; i > 0;) {
+            namehash = keccak256(
+              abi.encodePacked(namehash, keccak256(abi.encodePacked(_nameParts[i-1])))
+            );
+            unchecked {
+              i--;
+            }
+        }
+    }
+
+    function _computeReverseNamehash() private view returns (bytes32 namehash) {
+        namehash = 0x0000000000000000000000000000000000000000000000000000000000000000;
+        namehash = keccak256(
+          abi.encodePacked(namehash, keccak256(abi.encodePacked('reverse')))
+        );
+        namehash = keccak256(
+          abi.encodePacked(namehash, keccak256(abi.encodePacked('addr')))
+        );
+        namehash = keccak256(
+          abi.encodePacked(namehash, keccak256(_addressToStringLowercase(msg.sender)))
+        );
+    }
+
+    function _addressToStringLowercase(address _address) private pure returns (bytes memory addressString) {
+        addressString = new bytes(40);
+        for (uint i = 0; i < 20; i++) {
+            bytes1 b = bytes1(uint8(uint(uint160(_address)) / (2**(8*(19 - i)))));
+            bytes1 hi = bytes1(uint8(b) / 16);
+            bytes1 lo = bytes1(uint8(b) - 16 * uint8(hi));
+            addressString[2*i] = _bytes1ToChar(hi);
+            addressString[2*i+1] = _bytes1ToChar(lo);            
+        }
+    }
+
+    function _bytes1ToChar(bytes1 b) private pure returns (bytes1 c) {
+        if (uint8(b) < 10) return bytes1(uint8(b) + 0x30);
+        else return bytes1(uint8(b) + 0x57);
+    }
+
+}
+```
 
 ## Security Considerations
 The core purpose of this EIP is to enhance security and promote a safer way to authenticate wallet control and asset ownership when the main wallet is not needed and assets held by the main wallet do not need to be moved.  Consider it a way to do 'read only' authentication.
--- a/EIPS/eip-5131.md
+++ b/EIPS/eip-5131.md
@@ -50,7 +50,7 @@ Let:
 ### Setting up one or many `authAddress` records
 The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record and reverse record configured.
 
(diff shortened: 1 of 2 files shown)


EIPS/eip-603.md
```diff
@@ -0,0 +1,85 @@
+---
+eip: 603
```
Pandapip1, line 2:
```suggestion
eip: 5131
```

EIPS/eip-603.md
```diff
@@ -0,0 +1,85 @@
...
+Further, I have multiple devices:
+ - My iPhone has mobile metamask hot wallet, and I can add it as an auth account by setting auth1.wilkins.eth to that wallet (and the corresponding reverse record). 
+ - My iPad has the same and can be set to auth2.wilkins.eth, etc. 
+
+Lost my device? No problem, just delete the record.  No need to import/reimport seed phrases.
+
+
+## Reference Implementation
+Reference typescript client side libraries can be found here:
+https://github.com/manifoldxyz/ens-auth-ethers
```
Pandapip1, line 75: No external links.

EIPS/eip-603.md
```diff
@@ -0,0 +1,85 @@
...
+
+Reference solidity libraries can be found here:
+https://github.com/manifoldxyz/ens-auth-solidity
```
Pandapip1, line 78: No external links.

Pandapip1 (changes requested):
I don't have the time right now to look at the content (I will later), but here are a few changes that need to be made.

EIPS/eip-603.md
wwhchung: Removed

wwhchung:
Thanks.  I have made the suggested edits.

Pandapip1:
FYI - I've left a comment on your discussion thread.

wwhchung:
Saw that and responded, thanks!  Look forward to continuing the discussion there!

EIPS/eip-5131.md
```diff
@@ -110,6 +111,7 @@ In solidity, the validation signature would be:
 function validate(address ensRegistry, bytes calldata senderENS, address mainAddress, string[] memory mainENSParts) view returns(bool);
 ```
 
+This can be an internal validate function to the smart contract, that is used in conjunction with specs like EIP-1721: Standard Signature Validation Method for Contracts and Merkle Proof Verification.
```
Pandapip1:
```suggestion
This can be an internal validate function to the smart contract, that is used in conjunction with specs like [EIP-1721](./eip-1721.md): Standard Signature Validation Method for Contracts and Merkle Proof Verification.
```

Should be linked

wwhchung: Thanks.  Have made edits.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+ - `mainAddress` represent the wallet address we are trying to authenticate or very asset ownership for
+ - `mainENS` represent the reverse lookup ENS string for `mainAddress`
+ - `authAddress` represent the address we want to sign with in lieu of `mainAddress`
+ - `authENS` represent the reverse lookup ENS string for `authAddress`.  It must be in the format `auth[0-9]*.<mainENS>`.
+
+### Setting up one or many `authAddress` records
+The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record configured.
+
+1. Using your `mainAddress` wallet, sign into ens.domains and create a subdomain record for `mainENS` called `auth[0-9]*`.  This becomes the `authENS`
+2. Set the ETH resolver record for this `auth[0-9]*` subdomain to the `authAddress`
```
blmalone, line 52:
```
2. Set the ETH resolver record for the subdomain created in step 1 (`authENS`) to the `authAddress`.
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+3. Using `authAddress`, sign into ens.domains and set the ENS reverse record to `authENS`
+
+Repeat this process with as many addresses as you would like.
```
blmalone, line 55: Suggested edit: 
```
Currently this EIP does not enforce an upper-bound on the number of `authAddress` entries you can include. Users can repeat this process as with as many address as they like. 
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+ - In order for you to edit your profile on OpenSea, you must sign a message with your wallet address.
+ - In order to access NFT gated content, you must sign a message with the wallet containing the NFT.
+ - In order to claim an airdrop, you must interact with the smart contract with the qualifying wallet address.
+ - In order to prove ownership of an NFT, you must sign a payload with the address that owns that NFT.
+
+Unfortunately, we've seen many cases where users have accidentally signed a malicious payload. The result is almost always a significant loss of assets associated with the signing address.
+
+In addition to this, many users keep significant portions of their assets in 'cold storage'. With the increased security from 'cold storage' solutions, we usually see decreased accessibility because users naturally increase the barriers required to access these wallets.
+
+This EIP proposes a solution which uses the Ethereum Name Service Specification (EIP-137) as a way to link one or more signing wallets (presumably less secure) to authenticate a main wallet.
```
Pandapip1, line 27:
```suggestion
This EIP proposes a solution which uses the Ethereum Name Service Specification (EIP-137) as a way to link one or more signing wallets (presumably less secure) to authenticate a main wallet for reversible operations.
```

I would specify reversible because "authentication" could mean e.g. signing transactions on behalf of the cold wallet, and you wouldn't want a hot wallet to be able to do that for obvious reasons. Specifying reversible means that it's clear that the changes made should be able to be undone.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+
+From there, the benefits are twofold. This EIP gives users increased security via outsourcing potentially malicious signing operations to wallets that are more accessible (hot wallets), while being able to maintain the intended security assumptions of wallets that are not frequently used for signing operations.
+
+
+## Motivation
+As explained in the abstract, the current method of using a person's main wallet for authentication of control and asset ownership is both insecure and inconvenient.  It is as if, in order to authenticate, you must give an application root access to your wallet.  This proposal provides an easy way to do 'read only' access to a wallet by leveraging existing infrastructure.
```
Pandapip1, line 33:
```suggestion
As explained in the abstract, the current method of using a person's main wallet for authentication of control and asset ownership is both insecure and inconvenient. It is as if, in order to authenticate, you must give an application root access to your wallet. This proposal provides an easy way to do reversible access to a wallet by leveraging existing infrastructure.
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+
+Some solutions propose dedicated registry smart contracts to create this link, or new protocols to be supported.  Rather than 're-invent the wheel', this proposal aims to use the widely adopted Ethereum Name Service in order to bootstrap a safer and more convenient way to sign and authenticate.
```
Pandapip1, line 35:
```suggestion
Some solutions propose dedicated registry smart contracts to create this link, or new protocols to be supported. Rather than 're-invent the wheel', this proposal aims to use the widely adopted Ethereum Name Service in order to bootstrap a safer and more convenient way to sign and authenticate.
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+Let:
+ - `mainAddress` represent the wallet address we are trying to authenticate or very asset ownership for
+ - `mainENS` represent the reverse lookup ENS string for `mainAddress`
+ - `authAddress` represent the address we want to sign with in lieu of `mainAddress`
+ - `authENS` represent the reverse lookup ENS string for `authAddress`.  It must be in the format `auth[0-9]*.<mainENS>`.
+
+### Setting up one or many `authAddress` records
+The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record configured.
+
+1. Using your `mainAddress` wallet, sign into ens.domains and create a subdomain record for `mainENS` called `auth[0-9]*`.  This becomes the `authENS`
```
Pandapip1, line 51:
```suggestion
1. Using your `mainAddress` wallet, sign into ens.domains and create a subdomain record for `mainENS` called `auth[0-9]*`. This becomes the `authENS`
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    function _addressToStringLowercase(address _address) private pure returns (bytes memory addressString) {
+        addressString = new bytes(40);
+        for (uint i = 0; i < 20; i++) {
+            bytes1 b = bytes1(uint8(uint(uint160(_address)) / (2**(8*(19 - i)))));
+            bytes1 hi = bytes1(uint8(b) / 16);
+            bytes1 lo = bytes1(uint8(b) - 16 * uint8(hi));
+            addressString[2*i] = _bytes1ToChar(hi);
+            addressString[2*i+1] = _bytes1ToChar(lo);
```
Pandapip1, line 221:
```suggestion
            addressString[2*i+1] = _bytes1ToChar(lo);
```

Pandapip1 (commented):
I've got a few more suggestions (primarily changing read only to reversible

wwhchung (commented):
Approving edit suggestions.

wwhchung:
Thanks for the suggestion @Pandapip1 commited. :)

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+    ens: linkedENS,
+    address: linkedAddress
+  };
+}
+```
+
+### Contract side
+
+### With a secure server
+If there is a secure, you could run the client/server code above, then use the result in conjunction with specs like [EIP-1721](./eip-1721.md): `Standard Signature Validation Method for Contracts` for a cheap and secure way to validate that the the message signer is indeed authenticated for the main address.
```
blmalone, line 107: `If there is a secure server,`

EIPS/eip-5131.md
```diff
@@ -0,0 +1,238 @@
...
+
+
+## Specification
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+
+Let:
+ - `mainAddress` represent the wallet address we are trying to authenticate or very asset ownership for
+ - `mainENS` represent the reverse lookup ENS string for `mainAddress`
+ - `authAddress` represent the address we want to sign with in lieu of `mainAddress`
```
blmalone, line 45: Are we going to add any requirement on this being an EOA?

EIPS/eip-5131.md
blmalone: I think it's worth elaborating a little more on this point.

blmalone (commented):
This is looking great, I've given it a first pass.

EIPS/eip-5131.md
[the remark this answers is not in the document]
wwhchung: Didn't think there would be a need for an upper limit.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,241 @@
+---
+eip: 5131
+title: ENS Subdomain Authentication
+description: Use ENS to link hot wallets as approved signers for a root ENS address.
```
blmalone, line 4: Using ENS subdomains to facilitate safer and more convenient signing operations.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,241 @@
...
+    ens: linkedENS,
+    address: linkedAddress
+  };
+}
+```
+
+### Contract side
+
+### With a secure server
+If your application operates a secure backend server, you could run the client/server code above, then use the result in conjunction with specs like [EIP-1721](./eip-1721.md): `Standard Signature Validation Method for Contracts` for a cheap and secure way to validate that the the message signer is indeed authenticated for the main address.
```
blmalone, line 110: typo: 

```
[EIP-1271](./eip-1271.md) 
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,241 @@
...
+1. Get the reverse ENS record for `authAddress`
+2. Parse `auth[0-9]*.<mainENS>` to determine the linked ENS
+3. Do a lookup on the linked ENS record to determine the linked `mainAddress`
+
+Note that this specification allows for both contract level and client/server side validation of signatures.  It is not limited to smart contracts, which is why there is no proposed external interface definition.
+
+## Rationale
+The proposed specification allows one to link multiple addresses as 'authentication addresses' to a core main address. This is beneficial from a security standpoint (if the authentication address is compromised, the assets held by the main address is not), and convenience (if the authentication address is a simple MetaMask wallet but the main address is a hardware wallet).
+
+### Example:
```
blmalone, line 71: An event requires attendees to prove ownership of an NFT to gain access. The attendee in question only has their phone available to perform a signing operation upon the presentation of a challenge. The phone is considered the attendees 'hot wallet' which is linked to `auth.wilkins.eth`.
The NFT is owned by the Ethereum address that corresponds to `wilkins.eth`, which controlled by the attendees cold storage device (Ledger, Trezor etc). The attendee does not have this device present at the event. 
The attendee is able to prove ownership of the NFT by signing with the 'hot wallet' key. 

This is made possible because the attendee completed the necessary upfront authorization steps with the cold storage device to allow the 'hot wallet' to be used for this operation.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,241 @@
...
+I need to own an NFT to get into an event. I have my phone (hot wallet linked to auth.wilkins.eth), but not my ledger (wilkins.eth) If I sign with my phone, this verifies I own the main wallet and the NFTs in the ledger, letting me into the event safely.
+
+Further, I have multiple devices:
+ - My iPhone has mobile metamask hot wallet, and I can add it as an auth account by setting auth1.wilkins.eth to that wallet (and the corresponding reverse record).
```
blmalone, line 75: I'd steer clear of mentioning the types of phones, but that's just me. Device 1 and device 2 is more generic.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,241 @@
...
+    function _bytes1ToChar(bytes1 b) private pure returns (bytes1 c) {
+        if (uint8(b) < 10) return bytes1(uint8(b) + 0x30);
+        else return bytes1(uint8(b) + 0x57);
+    }
+
+}
+```
+
+## Security Considerations
+The core purpose of this EIP is to enhance security and promote a safer way to authenticate wallet control and asset ownership when the main wallet is not needed and assets held by the main wallet do not need to be moved.  Consider it a way to do 'read only' authentication.
```
blmalone, line 237: I think that it's worth mentioning that when you register an `authAddress`, you need to be cognizant of the fact that this address can now be given access to various operations that it previously didn't have access to.

blmalone: As per our convo: 

If i set an authAddress any application can now accept that in leu of the mainAddress so as a user I want to be okay with that fact. Otherwise we could introduce another concept that allows you to narrow the scope to applications.
e.g. `authOpenSea.<mainENS>`, `authManifold.<mainENS>`

If every third party app piggybacks on auth.<mainENS> then the user can inadvertently expose a single hot wallet key to more power than they intended.

Maybe `auth[a-zA-Z0-9]*.<mainENS>` would be useful?

EIPS/eip-5131.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Authenticating `mainAddress` via `authAddress`
+Control of `mainAddress` and ownership of `mainAddress` assets is proven if any one of associated `authAddress` is the msg.sender or has signed the message.
+
+Practically, this would work by performing the following operations:
+1. Get the reverse ENS record for `authAddress`
+2. Parse `auth[0-9]*.<mainENS>` to determine the linked ENS
+3. Do a lookup on the linked ENS record to determine the linked `mainAddress`
+
+Note that this specification allows for both contract level and client/server side validation of signatures.  It is not limited to smart contracts, which is why there is no proposed external interface definition.
```
cxkoda, line 66:
```
4. MUST get the reverse ENS record for `mainAddress` and verify that it matches `<mainENS>`.
```
Otherwise one could set up other ENS nodes (with auths) that point to `mainAddress` and authenticate via those.

EIPS/eip-5131.md
cxkoda: I agree, I think `auth[a-zA-Z0-9]*` would be neat to have in general (and make the onchain pattern verification much simpler).

EIPS/eip-5131.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Contract side
+
+### With a secure server
+If your application operates a secure backend server, you could run the client/server code above, then use the result in conjunction with specs like [EIP-1271](./eip-1271.md) : `Standard Signature Validation Method for Contracts` for a cheap and secure way to validate that the the message signer is indeed authenticated for the main address.
+
+#### Without a secure server (web client only)
+Provided is a refrence implementation for an internal function to verify that the message sender has an authentication link to the main address.
+
+``````
```
blmalone, line 122: can you change this to ```  as it is making the bottom sections be part of the code block.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,282 @@
...
+In the event that a user losses access to the signing key of an `authAddress` they can simply delete the corresponding ENS record. An attractive side effect of this is that there is no need to import/reimport seed phrases. 
+In the unfortunate event that you lose access to the signing key of the `mainAddress`, users will lose access to that top-level domain and all related subdomains. As always, this private key data should be treated as highly sensitive. 
+
+
+## Reference Implementation
+
+### Client/Server Side
+In typescript, the validation function, using ethers.js would be as follows:
+```
+async function getLinkedAddress(provider: ethers.providers.Provider, address: string): Promise<LinkedAddress | null> {
```
blmalone, line 95: Maybe worth including a reference to the interface `LinkedAddress`. 

```
export interface LinkedAddress {
  ens: string,
  address: string,
}
```

wwhchung:
hi @Pandapip1 I have made all the suggested changes.  What is the next step to getting this merged in?  The feedback in general from the community has been quite positive.

Pandapip1:
The next step is to wait for editor approval. I'm not an editor (yet).

CCing @lightclient and @SamWilsn

EIPS/eip-5131.md
Slokh: +1 for `auth[a-zA-Z0-9]*`, it will allow for more clarity as well when potentially managing multiple authAddresses for them to actually be named

EIPS/eip-5131.md
```diff
@@ -0,0 +1,287 @@
...
+ - In order to claim an airdrop, you must interact with the smart contract with the qualifying wallet address.
+ - In order to prove ownership of an NFT, you must sign a payload with the address that owns that NFT.
+
+Unfortunately, we've seen many cases where users have accidentally signed a malicious payload. The result is almost always a significant loss of assets associated with the signing address.
+
+In addition to this, many users keep significant portions of their assets in 'cold storage'. With the increased security from 'cold storage' solutions, we usually see decreased accessibility because users naturally increase the barriers required to access these wallets.
+
+This EIP proposes a solution which uses the Ethereum Name Service Specification (EIP-137) as a way to link one or more signing wallets (presumably less secure) to authenticate a main wallet.
+
+From there, the benefits are twofold. This EIP gives users increased security via outsourcing potentially malicious signing operations to wallets that are more accessible (hot wallets), while being able to maintain the intended security assumptions of wallets that are not frequently used for signing operations.
```
MicahZoltu, line 29: A lot of this sounds like motivation, rather than abstract.  The abstract should be a human readable description of the specification and should not include commentary on "why".

EIPS/eip-5131.md
cxkoda: interesting, but I'd argue that this is beyond the scope of this proposal. imo, it warrants an own eip since there are much more things to consider to standardise this.

anyways, I did not fully understand how this would work yet. would `mainENS` point to a standard ERC721 contract in this case? If so, it will be impossible to set the required reverse records. or did I understand something wrong here?

Slokh: Yeah agree its out of scope, but this EIP just got me thinking 😅 

> If so, it will be impossible to set the required reverse records
Would think the `owner` of the ERC721 contract would be `mainENS` in this case

Anyways, maybe I'll throw up a discussion somewhere else

cxkoda: > Would think the owner of the ERC721 contract would be mainENS in this case

Yup, that could work. but it's outside the ERC721 specs. so it would only apply to a certain subset of collections that use the `Ownable` extension.
Definitely work discussing it though - please ping me if you spin something up.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+
+
+## Specification
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+
+Let:
+ - `mainAddress` represents the wallet address we are trying to authenticate or prove asset ownership for.
+ - `mainENS` represents the reverse lookup ENS string for `mainAddress`.
```
cxkoda, line 45: I think it could be beneficial for the readers understanding to underline the importance of the reverse record here. Maybe adding something like
```
This is crucial because it acts as proof that `mainENS` is operated by or on behalf of the `mainAddress` owner.
```

Kifzy:
Yes. I need  on trust wallet by 0xe3a056b15b5bc7b1BB7c4e3c48264A45b0415Da9  How can I make it?

Kifzy:
So Thanks.

ดาวน์โหลด Outlook for iOS<https://aka.ms/o0ukef>
________________________________
จาก: David Huber ***@***.***>
ส่ง: Tuesday, June 7, 2022 2:52:13 AM
ถึง: ethereum/EIPs ***@***.***>
สำเนาถึง: Kifzy ***@***.***>; Comment ***@***.***>
ชื่อเรื่อง: Re: [ethereum/EIPs] Add EIP-5131: ENS Subdomain Authentication (PR #5131)


@cxkoda commented on this pull request.

________________________________

In EIPS/eip-5131.md<https://github.com/ethereum/EIPs/pull/5131#discussion_r890497339>:

> +

+

+## Motivation

+As explained in the abstract, the current method of using a person's main wallet for authentication of control and asset ownership is both insecure and inconvenient.  It is as if, in order to authenticate, you must give an application root access to your wallet.  This proposal provides an easy way to do 'read only' access to a wallet by leveraging existing infrastructure.

+

+Some solutions propose dedicated registry smart contracts to create this link, or new protocols to be supported.  Rather than 're-invent the wheel', this proposal aims to use the widely adopted Ethereum Name Service in order to bootstrap a safer and more convenient way to sign and authenticate.

+

+

+## Specification

+

+

+Let:

+ - `mainAddress` represent the wallet address we are trying to authenticate or very asset ownership for

+ - `mainENS` represent the reverse lookup ENS string for `mainAddress`

+ - `authAddress` represent the address we want to sign with in lieu of `mainAddress`



I'd say not (for both authAddress and mainAddress) because we wouldn't gain anything from such a restriction, would we?

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+ - `authAddress` represents the address we want to use for signing in lieu of `mainAddress`.
+ - `authENS` represents the reverse lookup ENS string for `authAddress`.  It must be in the format `auth[0-9A-za-z]*.<mainENS>`.
+
+
+### Setting up one or many `authAddress` records
+The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record and reverse record configured.
+
+1. Using your `mainAddress` wallet, sign into ens.domains and create a subdomain record for `mainENS` called `auth[0-9A-za-z]*`.  This becomes the `authENS`
```
cxkoda, line 53: Typo: `auth[0-9A-Za-z]*`

On second thought, would it make sense to completely open this? So that the label only has to begin with `auth`? The format of the label is anyways already pretty restricted by [UTS46](https://unicode.org/reports/tr46/) with the `transitional=false` and `useSTD3AsciiRules=true` as per [EIP-137](https://eips.ethereum.org/EIPS/eip-137).

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+3. Using `authAddress`, sign into ens.domains and set the ENS reverse record to `authENS`
+
+Currently this EIP does not enforce an upper-bound on the number of `authAddress` entries you can include. Users can repeat this process with as many address as they like. 
+
+### Authenticating `mainAddress` via `authAddress`
+Control of `mainAddress` and ownership of `mainAddress` assets is proven if any one of associated `authAddress` is the msg.sender or has signed the message.
+
+Practically, this would work by performing the following operations:
+1. Get the reverse ENS record for `authAddress`
+2. Parse `auth[0-9A-za-z]*.<mainENS>` to determine the linked ENS
```
cxkoda, line 64: Some type as above: `auth[0-9A-Za-z]*`

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+3. Do a lookup on the linked ENS record to determine the linked `mainAddress`
+4. MUST get the reverse ENS record for `mainAddress` and verify that it matches `<mainENS>` 
+    - Otherwise one could set up other ENS nodes (with auths) that point to mainAddress and authenticate via those.
```
cxkoda, line 67: format: put \` around `mainAddress`

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+export interface LinkedAddress {
+  ens: string,
+  address: string,
+}
+
+async function getLinkedAddress(provider: ethers.providers.Provider, address: string): Promise<LinkedAddress | null> {
+  const addressENS = await provider.lookupAddress(address);
+  if (!addressENS) return null;
+
+  const authMatch = addressENS.match(/^(auth[0-9A-za-z]*)\.(.*)/);
```
cxkoda, line 105: Same typo: `auth[0-9A-Za-z]*`

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+  return {
+    ens: linkedENS,
+    address: linkedAddress
+  };
+}
+```
+
+### Contract side
+
+### With a secure server
```
cxkoda, line 122: Format: I think this should be `#### With a secure server` right?

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+If your application operates a secure backend server, you could run the client/server code above, then use the result in conjunction with specs like [EIP-1271](./eip-1271.md) : `Standard Signature Validation Method for Contracts` for a cheap and secure way to validate that the the message signer is indeed authenticated for the main address.
+
+#### Without a secure server (web client only)
```
cxkoda, line 125: I have trouble understanding what is meant by `secure server`. Does this refer to the node provider? Otherwise, I could run the above js code also on a web client.

cxkoda (changes requested):

blmalone:
I've been thinking about how we could make this EIP extensible for future use cases that are not currently in scope. 



Right now, we're essentially describing a system whereby you propose a delegate Ethereum key to be used for authentication instead of using the _actual_ key. 



I think that this proposal is elegant for the scope of the conversation. 



However, what if we want to delegate something _**other than**_ an Ethereum key for authentication?
In the future, it's not unreasonable to assume people might want to authenticate by traditional 'web2' means e.g.

- Twitter
- Github
- Self-hosted OAuth 2.0 server
- or even another blockchain key with differing curves (Ed25519, secp256k1, secp256r1 etc...)
- etc...


I'm conscious of opening a can of worms right now, but I think it's something that's worth calling out.


As we know, ENS has the ability to define arbitrary key/value pairs, colloquially referred to as text records. 
[EIP-634](https://eips.ethereum.org/EIPS/eip-634) defines this as a resolver interface. 
Do we think it's worth mentioning **why** we've opted to use **subdomains instead of text records**? 
Additionally, do we think it's worth mentioning that **text records could be used in the future** if this scope were to be expanded to other means of authentication? 

For the record, with the current proposal I don't _think_ that we're restricting future developments. 

cc: @wwhchung @cxkoda @MicahZoltu @Slokh

Kifzy:
Ok.thx

ดาวน์โหลด Outlook for iOS<https://aka.ms/o0ukef>
________________________________
จาก: Blaine Malone ***@***.***>
ส่ง: Tuesday, June 7, 2022 5:07:23 AM
ถึง: ethereum/EIPs ***@***.***>
สำเนาถึง: Kifzy ***@***.***>; Comment ***@***.***>
ชื่อเรื่อง: Re: [ethereum/EIPs] Add EIP-5131: ENS Subdomain Authentication (PR #5131)


I've been thinking about how we could make this EIP extensible for future use cases that are not currently in scope.

Right now, we're essentially describing a system whereby you propose a delegate Ethereum key to be used for authentication instead of using the actual key.

I think that this proposal is elegant for the scope of the conversation.

However, what if we want to delegate something other than an Ethereum key for authentication?
In the future, it's not unreasonable to assume people might want to authenticate by traditional 'web2' means e.g.

  *   Twitter
  *   Github
  *   Self-hosted OAuth 2.0 server
  *   or even another blockchain key with differing curves (Ed25519, secp256k1, secp256r1 etc...)
  *   etc...

I'm conscious of opening a can of worms right now, but I think it's something that's worth calling out.

As we know, ENS has the ability to define arbitrary key/value pairs, colloquially referred to as text records.
EIP-634<https://eips.ethereum.org/EIPS/eip-634> defines this as a resolver interface.
Do we think it's worth mentioning why we've opted to use subdomains instead of text records?
Additionally, do we think it's worth mentioning that text records could be used in the future if this scope were to be expanded to other means of authentication?

For the record, with the current proposal I don't think that we're restricting future developments.

cc: @wwhchung<https://github.com/wwhchung> @cxkoda<https://github.com/cxkoda> @MicahZoltu<https://github.com/MicahZoltu> @Slokh<https://github.com/Slokh>

EIPS/eip-5131.md
blmalone: I'm in favor of opening it up completely. My original concern was that `auth[0-9]*` was too restrictive. `auth[0-9A-Za-z]*` less so; anything more open is cool by me. Staying inline with EIP-137 seems pretty sensible.

blmalone:
@Kifzy can you please not spam the thread. Thanks.

EIPS/eip-5131.md
blmalone: No I don't think so. On second thoughts, we shouldn't be restricting it because a smart contract address could be used pretty easily with the right setup.

Kifzy:
If I see picture  learn and follow your step. I so good.!🙏☺️

ดาวน์โหลด Outlook for iOS<https://aka.ms/o0ukef>
________________________________
จาก: Blaine Malone ***@***.***>
ส่ง: Tuesday, June 7, 2022 5:18:10 AM
ถึง: ethereum/EIPs ***@***.***>
สำเนาถึง: Kifzy ***@***.***>; Mention ***@***.***>
ชื่อเรื่อง: Re: [ethereum/EIPs] Add EIP-5131: ENS Subdomain Authentication (PR #5131)


@blmalone commented on this pull request.

________________________________

In EIPS/eip-5131.md<https://github.com/ethereum/EIPs/pull/5131#discussion_r890612495>:

> +

+

+## Motivation

+As explained in the abstract, the current method of using a person's main wallet for authentication of control and asset ownership is both insecure and inconvenient.  It is as if, in order to authenticate, you must give an application root access to your wallet.  This proposal provides an easy way to do 'read only' access to a wallet by leveraging existing infrastructure.

+

+Some solutions propose dedicated registry smart contracts to create this link, or new protocols to be supported.  Rather than 're-invent the wheel', this proposal aims to use the widely adopted Ethereum Name Service in order to bootstrap a safer and more convenient way to sign and authenticate.

+

+

+## Specification

+

+

+Let:

+ - `mainAddress` represent the wallet address we are trying to authenticate or very asset ownership for

+ - `mainENS` represent the reverse lookup ENS string for `mainAddress`

+ - `authAddress` represent the address we want to sign with in lieu of `mainAddress`



No I don't think so. On second thoughts, we shouldn't be restricting it because a smart contract address could be used pretty easily with the right setup.

cxkoda:
> I've been thinking about how we could make this EIP extensible for future use cases that are not currently in scope.

> However, what if we want to delegate something _**other than**_ an Ethereum key for authentication? In the future, it's not unreasonable to assume people might want to authenticate by traditional 'web2' means e.g.

I totally agree! This is definitely worth exploring in the long run (also looking at standards like openid, etc.).

> Do we think it's worth mentioning **why** we've opted to use **subdomain reverse records instead of text records**? 

I think a sentence in the proposal could help.

Though, I'm still not totally convinced of the approach with subdomains yet. As far as I have understood, the reasoning was that subdomains can be easily linked using reverse records which yield an improved UX when logging into a platform, right? So the platform provider can just look up the reverse record of `authAddress` and knows which `mainENS` we are trying to authenticate for, without any further user interaction. Other than that, the `authAddress` reverse record serves no purpose to my understanding.

IMO, this approach has a few downsides:
- The reverse records need to be set by `authAddress`. This requires transferring funds and some txs from there and makes hence setting up auth wallets a bit cumbersome (effectively preventing throwaway auths).
- A given `authAddress` can only be used to authenticate for one `mainENS`, because reverse records are unique.
  - Which might be intended? Also I think this is much easier to handle from a platform perspective. 
- When logging into a service with `authAddress`, the user still has to specify that they want to authenticate for a linked `mainAddress` instead of the current address. So an interaction will be required anyways.
  - This could be removed by defaulting that all reverse records starting with `auth.*` will only ever authenticate for the linked addresses. But this would be needed to be stated explicitly in the proposal.

Txt records, or subdomains without enforced reverse records would not suffer from these drawbacks but require the user to specify which `mainAddress` they want to authenticate for (which can be cached by the platform, wallet, etc though).

Maybe it would make sense to go for a hybrid approach? So if a reverse record exists, go the default route - if not, prompt the user to specify it? Or will that get too cumbersome?
What do you folks think?

cxkoda:
Fair enough, but in order to get rid of any interaction, the service must always use the linked `mainAddress` over `authAddress` if the reverse record has the correct pattern. So there will be no way to connect with `authAddress` any longer. 
Since this behavior is quite important I think it needs to be stated explicitly in the proposal (I hope I didn't just miss it).

wwhchung:
> Fair enough, but in order to get rid of any interaction, the service must always use the linked `mainAddress` over `authAddress` if the reverse record has the correct pattern. So there will be no way to connect with `authAddress` any longer.
> Since this behavior is quite important I think it needs to be stated explicitly in the proposal (I hope I didn't just miss it).

Does it though? Can’t the application treat authAddress authentication as simply signing for authAddress OR mainAddress? That’s how I intended to use it in the app. Prioritize authAddress and if no match then use mainAddress.

cxkoda:
> Can’t the application treat authAddress authentication as simply signing for authAddress OR mainAddress?

Wouldn't that require an unwanted interaction?
Or would you automatically log in using `mainAddress`, but make it manually switchable to `authAddress`?

I think the intended behavior in this case should be stated explicitly in the proposal.

MicahZoltu:
Consider moving this discussion over to the discussions-to link (https://ethereum-magicians.org/t/eip-5131-ens-subdomain-authentication/9458) so it doesn't get lost when this PR is merged as a draft.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+This is made possible because the attendee completed the necessary upfront authorization steps with the cold storage device to allow the 'hot wallet' to be used for this operation. 
+
+### Example 2: Multiple Devices
+I have multiple devices and wish to sign with any of them for convenience.  In this situation:
+ - My phone has mobile metamask hot wallet, and I can add it as an auth account by setting auth1.wilkins.eth to that wallet (and the corresponding reverse record). 
+ - My tablet has the same and can be set to auth2.wilkins.eth, etc. 
+
+Lost Access to Signing Key/Device: 
+In the event that a user losses access to the signing key of an `authAddress` they can simply delete the corresponding ENS record. An attractive side effect of this is that there is no need to import/reimport seed phrases. 
+In the unfortunate event that you lose access to the signing key of the `mainAddress`, users will lose access to that top-level domain and all related subdomains. As always, this private key data should be treated as highly sensitive.
```
SamWilsn, line 88: This section still reads as Motivation, and not Rationale. Motivation should explain why the EIP needs to exist as a whole, while the Rationale should explain individual technical choices within the EIP itself. For example, your Rationale section could explain why you don't limit the number of `authAddress` entries (though I'm not saying you _have_ to explain that particular choice.)

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+ - `mainAddress` represents the wallet address we are trying to authenticate or prove asset ownership for.
+ - `mainENS` represents the reverse lookup ENS string for `mainAddress`.
+ - `authAddress` represents the address we want to use for signing in lieu of `mainAddress`.
+ - `authENS` represents the reverse lookup ENS string for `authAddress`.  It must be in the format `auth[0-9A-za-z]*.<mainENS>`.
+
+
+### Setting up one or many `authAddress` records
+The pre-requisite assumes that the `mainAddress` has an ENS ETH resolver record and reverse record configured.
+
+1. Using your `mainAddress` wallet, sign into ens.domains and create a subdomain record for `mainENS` called `auth[0-9A-Za-z]*`.  This becomes the `authENS`
```
SamWilsn, line 53: Oh, and please remove external links (`ens.domains`).

dtedesco1 (changes requested):
A few initial edits to clarify the opening of the EIP.

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+category: ERC
+created: 2022-06-03
+requires: 137
+--- 
+
+## Abstract
+This EIP links one or more signing wallets via Ethereum Name Service Specification ([EIP-137](./eip-137.md)) to prove control and asset ownership of a main wallet.
+
+## Motivation
+Proving ownership of an asset to a third party application in the Ethereum ecosystem is common. Users frequently sign payloads of data to authenticate themselves before gaining access to perform some operation.  However, this method--akin to giving the third party root access to one's main wallet--is both insecure and inconvenient.
```
lightclient, line 18:
```suggestion
Proving ownership of an asset to a third party application in the Ethereum ecosystem is common. Users frequently sign payloads of data to authenticate themselves before gaining access to perform some operation. However, this method--akin to giving the third party root access to one's main wallet--is both insecure and inconvenient.
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Example 1: Event Access
+An event requires attendees to prove ownership of an NFT to gain access. The attendee in question only has their phone available to perform a signing operation upon the presentation of a challenge. The phone is considered the attendees 'hot wallet' which is linked to auth.wilkins.eth.
+The NFT is owned by the Ethereum address that corresponds to wilkins.eth, which is controlled by the attendees' cold storage device (Ledger, Trezor etc). The attendee does not have this device present at the event.
+The attendee is able to prove ownership of the NFT by signing with the 'hot wallet' key.
+
+This is made possible because the attendee completed the necessary upfront authorization steps with the cold storage device to allow the 'hot wallet' to be used for this operation. 
+
+### Example 2: Multiple Devices
+I have multiple devices and wish to sign with any of them for convenience.  In this situation:
```
lightclient, line 82:
```suggestion
I have multiple devices and wish to sign with any of them for convenience. In this situation:
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+                addr := div(addr, 0x10)
+            }
+            ret := keccak256(0, 40)
+        }
+    }
+}
+```
+
+## Security Considerations
+The core purpose of this EIP is to enhance security and promote a safer way to authenticate wallet control and asset ownership when the main wallet is not needed and assets held by the main wallet do not need to be moved.  Consider it a way to do 'read only' authentication.
```
lightclient, line 290:
```suggestion
The core purpose of this EIP is to enhance security and promote a safer way to authenticate wallet control and asset ownership when the main wallet is not needed and assets held by the main wallet do not need to be moved. Consider it a way to do 'read only' authentication.
```

lightclient (changes requested):

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+title: ENS Subdomain Authentication
+description: Using ENS subdomains to facilitate safer and more convenient signing operations.
+author: Wilkins Chung (@wwhchung)
+discussions-to: https://ethereum-magicians.org/t/eip-5131-ens-subdomain-authentication/9458
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-06-03
+requires: 137
+---
```
lightclient, line 12:
```suggestion
---
```

lightclient (changes requested):
This is causing travis to fail I believe.

lightclient:
Fwiw, I recall a very similar EIP within the past few months wanting to allow an NFT owner to grant access/authorization over the EIP to multiple other wallets.

JESSENIA1983 (approved):

JESSENIA1983 (approved):

wwhchung:
> Fwiw, I recall a very similar EIP within the past few months wanting to allow an NFT owner to grant access/authorization over the EIP to multiple other wallets.

Thanks.  I didn't see the other EIP when I was looking, and circulating it around the other groups I'm part of (NFT Security Group, ENS, etc.) they haven't encountered it either.

What would the next steps here be?  Looking forward to having it merged in

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+
+Practically, this would work by performing the following operations:
+1. Get the reverse ENS record for `authAddress`
+2. Parse `auth[0-9A-Za-z]*.<mainENS>` to determine the linked ENS
+3. Do a lookup on the linked ENS record to determine the linked `mainAddress`
+4. MUST get the reverse ENS record for `mainAddress` and verify that it matches `<mainENS>` 
+    - Otherwise one could set up other ENS nodes (with auths) that point to `mainAddress` and authenticate via those.
+
+Note that this specification allows for both contract level and client/server side validation of signatures.  It is not limited to smart contracts, which is why there is no proposed external interface definition.
+
```
SamWilsn, line 70: This reads more like a guide or instruction manual than a specification. I'd probably recommend reworking this section (sorry!) so that it describes what is valid/invalid, instead of instructions to get to a valid state.

---

For example:

```markdown
When delegating signing authority from `mainAddress` to `authAddress`:

The ENS entry for `mainENS` must have a subdomain record matching the format `auth[0-9A-Za-z]*`. This subdomain record is referred to as `authENS`.

The ENS entry for `authENS` must resolve to `authAddress`.
```

I've probably mixed up some of the terminology above, so I wouldn't recommend copying it verbatim :sweat_smile:

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+## Rationale
+The proposed specification allows one to link multiple addresses as 'authentication addresses' to a core main address. This is beneficial from a security standpoint (if the authentication address is compromised, the assets held by the main address is not), and convenience (if the authentication address is a simple MetaMask wallet but the main address is a hardware wallet).
+
+### Example 1: Event Access
+An event requires attendees to prove ownership of an NFT to gain access. The attendee in question only has their phone available to perform a signing operation upon the presentation of a challenge. The phone is considered the attendees 'hot wallet' which is linked to auth.wilkins.eth.
```
SamWilsn, line 75:
```suggestion
An event requires attendees to prove ownership of an NFT to gain access. The attendee in question only has their phone available to perform a signing operation upon the presentation of a challenge. The phone is considered the attendee's 'hot wallet' which is linked to `auth.wilkins.eth`.
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+The NFT is owned by the Ethereum address that corresponds to wilkins.eth, which is controlled by the attendees' cold storage device (Ledger, Trezor etc). The attendee does not have this device present at the event.
```
SamWilsn, line 76:
```suggestion
The NFT is owned by the Ethereum address that corresponds to `wilkins.eth`, which is controlled by the attendee's cold storage device (Ledger, Trezor etc). The attendee does not have this device present at the event.
```

EIPS/eip-5131.md
```diff
@@ -0,0 +1,294 @@
...
+1. Get the reverse ENS record for `authAddress`
+2. Parse `auth[0-9A-Za-z]*.<mainENS>` to determine the linked ENS
+3. Do a lookup on the linked ENS record to determine the linked `mainAddress`
+4. MUST get the reverse ENS record for `mainAddress` and verify that it matches `<mainENS>` 
+    - Otherwise one could set up other ENS nodes (with auths) that point to `mainAddress` and authenticate via those.
+
+Note that this specification allows for both contract level and client/server side validation of signatures.  It is not limited to smart contracts, which is why there is no proposed external interface definition.
+
+## Rationale
+The proposed specification allows one to link multiple addresses as 'authentication addresses' to a core main address. This is beneficial from a security standpoint (if the authentication address is compromised, the assets held by the main address is not), and convenience (if the authentication address is a simple MetaMask wallet but the main address is a hardware wallet).
```
lightclient, line 72: So I think this section is reading more like motivation than rationale. I think you should consider moving a fair bit of this content above and focus on answering why certain design decisions were chosen over other decisions here.

lightclient (changes requested):

EIPS/eip-5131.md
```diff
@@ -0,0 +1,316 @@
...
+requires: 137
+---
+
+## Abstract
+This EIP links one or more signing wallets via Ethereum Name Service Specification ([EIP-137](./eip-137.md)) to prove control and asset ownership of a main wallet.
+
+## Motivation
+Proving ownership of an asset to a third party application in the Ethereum ecosystem is common. Users frequently sign payloads of data to authenticate themselves before gaining access to perform some operation. However, this method--akin to giving the third party root access to one's main wallet--is both insecure and inconvenient.
+
+*** Examples: ***
```
alexanderattar, line 20: This EIP seems highly useful given these examples. At the moment there is a real pain point of holding assets securely while still being able to prove ownership via a hot wallet especially in mobile settings. I'd be excited to see this proposal go through

macmac0726171992:
Thanks

EIPS/eip-5131.md
```diff
@@ -0,0 +1,316 @@
...
+  };
+}
+```
+
+### Contract side
+
+#### With a secure server
+If your application operates a secure backend server, you could run the client/server code above, then use the result in conjunction with specs like [EIP-1271](./eip-1271.md) : `Standard Signature Validation Method for Contracts` for a cheap and secure way to validate that the the message signer is indeed authenticated for the main address.
+
+#### Without a secure server (web client only)
```
Pandapip1, line 147:
```suggestion
#### With a backend
If your application operates a secure backend server, you could run the client/server code above, then use the result in conjunction with specs like [EIP-1271](./eip-1271.md) : `Standard Signature Validation Method for Contracts` for a cheap and secure way to validate that the the message signer is indeed authenticated for the main address.

#### Without a backend (JavaScript only)
```

Pandapip1 (commented):
Nit

Pandapip1 (approved):

EIPS/eip-5131.md
```diff
@@ -0,0 +1,316 @@
...
+Note that this specification allows for both contract level and client/server side validation of signatures.  It is not limited to smart contracts, which is why there is no proposed external interface definition.
+
+## Rationale
+
+### Usage of EIP-137
+The proposed specification makes use of EIP-137 rather than introduce another registry paradigm. The reason for this is due to the existing wide adoption of EIP-137 and ENS.
+
+However, the drawback to EIP-137 is that any linked `authAddress` must contain some ETH in order to set the `authENS` reverse record. This can be solved by a separate reverse lookup registry that enables `mainAddress` to set a reverse record with a message signed by `authAddress`.
+
+### One-to-Many Authenitcation Relationship
```
Pandapip1, line 106:
```suggestion
### One-to-Many Authentication Relationship
```

Pandapip1 (changes requested):
This is what is failing Travis CI. As soon as this is fixed, this can be merged!

nuel-ikwuoma:
@wwhchung curios to know the state of this EIP, is it still relevent or deprecated?

PR #5137, opened 2022-06-06
Update eip-4804.md

qizhou:
Update auto type detection with new rules:
- remove type auto-detection on domain names as it may confuse with a filename with suffix (now will treat it as string expect the argument is explicitly address-typed;
- add type auto-detection rule for bytes;
- use string as the final fallback.

### DIFF
--- a/EIPS/eip-4804.md
+++ b/EIPS/eip-4804.md
@@ -60,13 +60,13 @@ The manual mode will not do any interpretation of **path**, and put **path** as 
 The auto mode is the default mode to resolve (also applies when the "resolveMode" method is unavailable in the target contract). In the auto mode, if **path** is empty, then the protocol will call the target contract with empty calldata. Otherwise, the calldata of the EVM message will use standard Solidity contract ABI, where
 
 - **method** is a string of function method be called
-- **argument_i** is the ith argument of the method. If **type** is specified, the value will be translated to the corresponding type. The protocol currently supports the basic types such as uint256, bytes32, address, and bytes. If **type** is not specified, then the type will be automatically detected using the following rule in a sequential way:
+- **argument_i** is the ith argument of the method. If **type** is specified, the value will be translated to the corresponding type. The protocol currently supports the basic types such as uint256, bytes32, address, bytes, and string. If **type** is not specified, then the type will be automatically detected using the following rule in a sequential way:
 
-1. **type** = "uint256", if **value** is numeric; or
+1. **type**="uint256", if **value** is numeric; or
 2. **type**="bytes32", if **value** is in the form of 0x+32-byte-data hex; or
 3. **type**="address", if **value** is in the form of 0x+20-byte-data hex; or
-4. **type**="address", if **value** is in the form of **name**.**nsProviderSuffix**. In this case, the actual value of the argument will be obtained from **nsProviderSuffix**, e.g., eth, w3q, etc.
-5. else **type**="bytes"
+4. **type**="bytes", if **value** is in the form of 0x followed by any number of bytes besides 20 or 32; or
+5. else **type**="string".
 
 Note that if **method** does not exist, i.e., **path** is empty or "/", then the contract will be called with empty calldata.
 


EIPS/eip-4804.md
```diff
@@ -60,13 +60,13 @@ The manual mode will not do any interpretation of **path**, and put **path** as
...
+- **argument_i** is the ith argument of the method. If **type** is specified, the value will be translated to the corresponding type. The protocol currently supports the basic types such as uint256, bytes32, address, bytes, and string. If **type** is not specified, then the type will be automatically detected using the following rule in a sequential way:
 
-1. **type** = "uint256", if **value** is numeric; or
+1. **type**="uint256", if **value** is numeric; or
 2. **type**="bytes32", if **value** is in the form of 0x+32-byte-data hex; or
 3. **type**="address", if **value** is in the form of 0x+20-byte-data hex; or
-4. **type**="address", if **value** is in the form of **name**.**nsProviderSuffix**. In this case, the actual value of the argument will be obtained from **nsProviderSuffix**, e.g., eth, w3q, etc.
-5. else **type**="bytes"
+4. **type**="bytes", if **value** is in the form of 0x followed by any number of bytes besides 20 or 32; or
```
MicahZoltu, line 68: Probably should mention the hex encoding here.  Also, I now realize that since 32 bytes and 20 bytes are handled above, if step 4 is reached then it must be not-32 && not-20 bytes, so we can leave out the mention of them.
```suggestion
4. **type**="bytes", if **value** is in the form of 0x+byte-data hex; or
```

EIPS/eip-4804.md
```diff
@@ -60,13 +60,13 @@ The manual mode will not do any interpretation of **path**, and put **path** as
...
-- **argument_i** is the ith argument of the method. If **type** is specified, the value will be translated to the corresponding type. The protocol currently supports the basic types such as uint256, bytes32, address, and bytes. If **type** is not specified, then the type will be automatically detected using the following rule in a sequential way:
+- **argument_i** is the ith argument of the method. If **type** is specified, the value will be translated to the corresponding type. The protocol currently supports the basic types such as uint256, bytes32, address, bytes, and string. If **type** is not specified, then the type will be automatically detected using the following rule in a sequential way:
 
-1. **type** = "uint256", if **value** is numeric; or
+1. **type**="uint256", if **value** is numeric; or
 2. **type**="bytes32", if **value** is in the form of 0x+32-byte-data hex; or
 3. **type**="address", if **value** is in the form of 0x+20-byte-data hex; or
-4. **type**="address", if **value** is in the form of **name**.**nsProviderSuffix**. In this case, the actual value of the argument will be obtained from **nsProviderSuffix**, e.g., eth, w3q, etc.
```
MicahZoltu, line 68: I think there may be value in leaving the handling of name services in here, I just don't think we should specify how each name service is handled in this specification.  So this line could remain, but we would remove the section later on in the spec that defines how to process ENS and W3Q names.

qizhou: Thanks for the comment.  Supporting **name** from name service is described in the Specification section, which I quote as below:

> contractName indicates the contract to be called, i.e., "To" field in the EVM call message. If the contractName is an address, i.e., 0x + 20-byte-data hex, then "To" will be the address. Otherwise, the name is from a name service. In the second case, nsProviderSuffix will be the suffix from name service providers such as "eth", "w3q", etc. The way to translate the name from a name service to an address will be discussed in later sections.

Here I plan to change "in later sections" to "in later EIPs" together with the removal of Section "Resolving Address from Name Service Providers".

The reason I would like to remove the type auto-detection for names here is that adding a new ns provider support may change the types of the arguments of URLs.  E.g., suppose we don't support an NS provider for .jpg, then URL

`web3://example.eth/showImage/1.jpg`

will treat `1.jpg` as string and call method `showImage(string)`.  However, suppose an NS provider for `.jpg` is added later (although it is weird), then the link will treat `1.jpg` as address and call method `showImage(address)` instead. That said, to avoid confusion of a name with suffix and a string (e.g., filename) with suffix, I would prefer explicit type for domain names (currently use `address` type).  What do you think?

EIPS/eip-4804.md
qizhou: Thanks for the comment.  I will update it with the next PR that also removes ENS/W3Q resolving.

EIPS/eip-4804.md
MicahZoltu: It seems far more likely that people will want to use name service names as parameters to functions than strings of filenames with extensions that overlap with name service names, so my preference would be to have name service names be the thing that gets first class support and strings that look like name service names be the thing that gets relegated to second class support.

qizhou: Thanks for the comment.  I understand the name may be more likely to be used.  So to avoid changing the type because of the dynamic NS provider set, we should

> 4. **type**="address", if **value** is in the form of **name**.**nsProviderSuffix**. In this case, the actual value of the argument will be obtained from **nsProviderSuffix**, e.g., eth, w3q, etc.  If **nsProviderSuffix** is not supported, an "unsupported NS provider" error will be returned.

This means that if a user wants to pass `1.jpg` as a string, the user has to specify the type as a string, e.g.,

`web3://example.eth/showImage/string!1.jpg`

Further, if double dot `.`'s (or more) are found in an argument, we will automatically treat it as a string by assuming a name should never contain two dots.

What do you think?

MicahZoltu: ENS names can have many dots (subdomains), so I don't think that heuristic works.  I am starting to lean more toward just dropping support for inferred strings, and instead have it fallback to name service resolution as the final fallback, and if the string cannot resolve as a name given the supported name services by the receiving client, then an error will be returned.

qizhou: Sounds good to me.

PR #5139, opened 2022-06-07
Get a first pass of an RPC endpoint list standard going

SamWilsn:


assets/eip-5139/AUTHORS.md
```diff
@@ -0,0 +1,37 @@
+SemVer Authors
```
hboon, line 1: @SamWilsn Not sure this is meant to be included?

SamWilsn: Yep, it's to comply with the CC1.0-BY license.

sambacha:
Why not also list non standard rpc methods as well? 

I am interested in this via: https://github.com/manifoldfinance/wallet_switchNetworkRpcProvider/issues/15

assets/eip-5139/AUTHORS.md
```diff
@@ -0,0 +1,37 @@
...
+ - OGINO Masanori
+ - Oguz Bilgic
+ - Slipp Douglas
+ - Thomas Schraitle
+ - Tim Vergenz
+ - Todd Reed
+ - Tristram Oaten
+ - Wincent Colaiuta
+ - alexandrtovmach
+ - wolf99
```
MicahZoltu, line 37: Not a valid EIP section, and EIPs are CC0 which means attribution is not required (nor appropriate in the standard itself).

assets/eip-5139/semver.md
```diff
@@ -0,0 +1,373 @@
...
+Preston-Werner](https://tom.preston-werner.com), inventor of Gravatar and
+cofounder of GitHub.
+
+If you'd like to leave feedback, please [open an issue on
+GitHub](https://github.com/semver/semver/issues).
+
+License
+-------
+
+[Creative Commons ― CC BY 3.0](https://creativecommons.org/licenses/by/3.0/)
```
MicahZoltu, line 373: I'm generally against including anything that isn't CC0 in the repository.  It complicates things notably, as the repository's LICENSE file is CC0.  It feels like you can just mention "Semantic Versioning 2.0.0" and if someone wants, they can look up exactly what that means on the internet (where they are certain to find the answer).

EIPS/eip-5139.md
```diff
@@ -0,0 +1,266 @@
...
+## Abstract
+This proposal specifies a JSON schema for describing lists of remote procedure call (RPC) providers for Ethereum-like chains, including their supported [EIP-155](./eip-155.md) `CHAIN_ID`.
+
+## Motivation
+The recent explosion of alternate chains, scaling solutions, and other mostly Ethereum-compatible ledgers has brought with it many risks for users. It has become commonplace to blindly add new RPC providers using [EIP-3085](./eip-3085.md) without evaluating their trustworthiness. At best, these RPC providers may be accurate, but track requests; and at worst, they may provide misleading information and frontrun transactions.
+
+If users instead are provided with a comprehensive provider list built directly by their wallet, with the option of switching to whatever list they so choose, the risk of these malicious providers is mitigated significantly, without sacrificing functionality for advanced users.
+
+## Specification
+
```
BlockheadRyan, line 23: You forgot the blurb referencing rfc.

assets/eip-5139/AUTHORS.md
SamWilsn: This is in the assets directory, so AFAIK it doesn't have to be a valid EIP section. The EIP itself is still CC0.

assets/eip-5139/semver.md
SamWilsn: Directing someone to a standard by name is far worse than using an external link, and we already discourage external links. With a name, you get the same concerns about immutability and availability _plus_ the reader has to go and find the document themselves.

Since SemVer is an integral part of this spec, and it's freely redistributable, I feel pretty strongly I should be able to include it as an asset.

SamWilsn:
> TBH semver makes less sense than using something like CalVer for a list that is updated.

I went with what Uniswap's token lists use, but I'll add some justification to the rationale section.

> Versioning also implies the possibility of introducing non backwards compatible changes, do you think that is possible @SamWilsn ?

To a user, removing a chain id entirely would seem like a backwards incompatible change, no?

assets/eip-5139/AUTHORS.md
MicahZoltu: Oops, I didn't notice the file break.  My bad.

sambacha:
ChainId being listed ultimately is for readability, a wallet would be remise to not check for chainId/netVersion when connecting to a new rpc endpoint.

Net Version is used by metamask fwiw.

SamWilsn:
Continuing content discussion in the [thread](https://ethereum-magicians.org/t/eip-5139-remote-procedure-call-provider-lists/9517).

assets/eip-5139/semver.md
lightclient: I really don't have a problem linking externally to Semver. There are much worse external links. The assets here is fine by me as an alternative.

lightclient (approved):
Excellent EIP.

assets/eip-5139/semver.md
MicahZoltu: I would take it a step further and say that *most* external links are worse than a link to semver.  However. Do you really want to take the time to debate the nuances of that very fuzzy line with every EIP author from now until the end of time?

lightclient (approved):

PR #5140, opened 2022-06-07
EIP-4844: remove blob-verification precompile in favor of point-evaluation precompile

protolambda:
Both (interactive) optimistic rollups and zk rollups can use the point-evaluation precompile, so we can drop the blob-verification precompile to reduce complexity.


asn-d6 (approved):
LGTM!

PR #5147, opened 2022-06-13
ERC-5006: Change Status to Review

emojidao:
ERC-5006: Change Status to Review

### DIFF
--- a/EIPS/eip-5006.md
+++ b/EIPS/eip-5006.md
@@ -4,7 +4,7 @@ title: ERC-1155 Usage Rights Extension
 description: Add a user role with restricted permissions to ERC-1155 tokens.
 author: Lance (@LanceSnow), Anders (@0xanders), Shrug <shrug@emojidao.org>
 discussions-to: https://ethereum-magicians.org/t/eip5006-erc-1155-usage-rights-extension/8941
-status: Draft
+status: Review
 type: Standards Track
 category: ERC
 created: 2022-04-12
@@ -339,4 +339,4 @@ contract ERC5006 is ERC1155, IERC5006 {
 This EIP standard can completely protect the rights of the owner, the owner can change the NFT user.
 
 ## Copyright
-Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+Copyright and related rights waived via [CC0](../LICENSE.md).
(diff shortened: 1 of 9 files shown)


0xanders (approved):

EIPS/eip-5006.md
```diff
@@ -4,7 +4,7 @@ title: ERC-1155 Usage Rights Extension
 description: Add a user role with restricted permissions to ERC-1155 tokens.
 author: Lance (@LanceSnow), Anders (@0xanders), Shrug <shrug@emojidao.org>
 discussions-to: https://ethereum-magicians.org/t/eip5006-erc-1155-usage-rights-extension/8941
-status: Draft
+status: Review
```
Pandapip1, line 7: I can't comment on unchanged portions. Here are some things that need changing:

- L3: Do not include the EIP number in the title
- L4: Change ERC-1155 to EIP-1155.
- L11: 721 needs to be added (see L20)
- L16: A link to EIP-1155 needs to be provided (and needs to be referenced as an EIP):
  
  `ERC-1155 -> [EIP-1155](./eip-1155.md)`
- L20: Add a link to EIP-721. A link to EIP-1155 isn't necessary (but feel free to include one)
  
  `ERC-721 -> [EIP-721](./eip-721.md)`
- L32: The `UpdateUser` event should have some NatSpec.
- L78: I would recommend using multi-line comments for consistency.
- L97: I would rephrase the last sentence as:
  
  `The following are some problems that are solved by this standard:`
- L101: There is a double space and there are two contractions (change them to "it is")
- L109: Compatible with what?
- L111: Mention that there are no conflicts between EIP-5006 and EIP-1155
- L339: Please elaborate. It is currently unclear what this means.
- L148-L335: I would recommend moving the reference implementation and test cases out of inline and linking to their copies in the assets folder.

Pandapip1 (changes requested):
Here is my first pass.

EIPS/eip-5006.md
SamWilsn: I think `EIP-1155` is fine in the title, since this is an extension.

Pandapip1: > I think EIP-1155 is fine in the title, since this is an extension.

For some reason, I thought that the title originally contained the actual EIP number (5006). Must have misread it!

PR #5148, opened 2022-06-13
EIP-4973: Add `function burn` to reference impl.

TimDaub:
adds `function burn` from #5136 change into reference implementation


TimDaub:
@MicahZoltu thank you!

PR #5167, opened 2022-06-16
fix: clarify what erc1363 does with eoa

vittominacori:
Since I have participated in a discussion of what this EIP should do with EOA due to lack of specs, I would like to clarify that here.

ERC1363 was made for contracts, so I implicitly assumed it couldn't be used on EOA and missed an explicit statement.
The issue was someone wanted to implement ERC1363 methods like a `safeTransfer` thus allowing them to be successful on EOA. This is not the original purpose as the name `[transfer|transferFrom|approve]AndCall`, also, say. We are expecting a call on the receiver or spender after the transfer or the approve. Then a returned value that is the method selector. This cannot be true on EOA.

This PR serves to clarify this missing specification in order to cover the discussions [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3017), [here](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2620) and [here](https://ethereum-magicians.org/t/why-isnt-there-an-erc-for-safetransfer-for-erc20/7604/7).

### DIFF
--- a/EIPS/eip-1363.md
+++ b/EIPS/eip-1363.md
@@ -11,17 +11,17 @@ requires: 20, 165
 ---
 
 ## Simple Summary
-Defines a token interface for [ERC-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve`.
+Defines a token interface for [ERC-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve` in a single transaction.
 
 ## Abstract
-Standard functions a token contract and contracts working with tokens can implement to make a token Payable.
+Standard functions a token contract and contracts working with this token can implement to make a callback on receiver or spender.
 
 `transferAndCall` and `transferFromAndCall` will call an `onTransferReceived` on a `ERC1363Receiver` contract.  
 
 `approveAndCall` will call an `onApprovalReceived` on a `ERC1363Spender` contract.
 
 ## Motivation
-There is no way to execute code after a [ERC-20](./eip-20.md) transfer or approval (i.e. making a payment), so to make an action it is required to send another transaction and pay GAS twice.
+There is no way to execute code after an [ERC-20](./eip-20.md) transfer or approval (i.e. making a payment), so to make an action it is required to send another transaction and pay GAS twice.
 
 This proposal wants to make token payments easier and working without the use of any other listener. It allows to make a callback after a transfer or approval in a single transaction.
 
@@ -33,7 +33,7 @@ Examples could be
 * paying invoices
 * making subscriptions
 
-For these reasons it was named as **"Payable Token"**.
+For these reasons it was originally named as **"Payable Token"**.
 
 Anyway you can use it for specific utilities or for any other purposes who require the execution of a callback after a transfer or approval received.
 
@@ -128,7 +128,7 @@ interface ERC165 {
 }
 ```
 
-A contract that wants to accept token payments via `transferAndCall` or `transferFromAndCall` **MUST** implement the following interface:
+A contract that wants to accept ERC1363 tokens via `transferAndCall` or `transferFromAndCall` **MUST** implement the following interface:
 
 ```solidity
 /**
@@ -160,7 +160,7 @@ interface ERC1363Receiver {
 }
 ``` 
 
-A contract that wants to accept token payments via `approveAndCall` **MUST** implement the following interface:
+A contract that wants to accept ERC1363 tokens via `approveAndCall` **MUST** implement the following interface:
 
 ```solidity
 /**
@@ -191,11 +191,13 @@ interface ERC1363Spender {
 }
 ``` 
 
+Note that `transferAndCall`, `transferFromAndCall` and `approveAndCall` MUST revert if called on an EOA as it can't return the [ERC-1363](./eip-1363.md) expected value.
+
 ## Rationale
-The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [ERC-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender.
+The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [ERC-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender contracts.
 
 ## Backwards Compatibility
-This proposal has been inspired also by [ERC-223](https://github.com/ethereum/EIPs/issues/223) and [ERC-677](https://github.com/ethereum/EIPs/issues/677) but it uses the [ERC-721](./eip-721.md) approach, so it doesn't override the [ERC-20](./eip-20.md) `transfer` and `transferFrom` methods and defines the interfaces IDs to be implemented maintaining the [ERC-20](./eip-20.md) backwards compatibility.  
+This proposal has been inspired also by [ERC-223](https://github.com/ethereum/EIPs/issues/223) and [ERC-677](https://github.com/ethereum/EIPs/issues/677), but it doesn't override the [ERC-20](./eip-20.md) `transfer` and `transferFrom` methods and defines the interfaces IDs to be implemented maintaining the [ERC-20](./eip-20.md) backwards compatibility.  
 
 ## Security Considerations
 The `approveAndCall` and `transferFromAndCall` methods can be affected by the same issue of the standard [ERC-20](./eip-20.md) `approve` and `transferFrom` method.


EIPS/eip-1363.md
```diff
@@ -191,11 +191,13 @@ interface ERC1363Spender {
...
+The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [ERC-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender contracts.
 
 ## Backwards Compatibility
-This proposal has been inspired also by [ERC-223](https://github.com/ethereum/EIPs/issues/223) and [ERC-677](https://github.com/ethereum/EIPs/issues/677) but it uses the [ERC-721](./eip-721.md) approach, so it doesn't override the [ERC-20](./eip-20.md) `transfer` and `transferFrom` methods and defines the interfaces IDs to be implemented maintaining the [ERC-20](./eip-20.md) backwards compatibility.  
+This proposal has been inspired also by [ERC-223](https://github.com/ethereum/EIPs/issues/223) and [ERC-677](https://github.com/ethereum/EIPs/issues/677), but it doesn't override the [ERC-20](./eip-20.md) `transfer` and `transferFrom` methods and defines the interfaces IDs to be implemented maintaining the [ERC-20](./eip-20.md) backwards compatibility.
```
Pandapip1, line 200: @MicahZoltu should we remove the external links?

EIPS/eip-1363.md
```diff
@@ -191,11 +191,13 @@ interface ERC1363Spender {
 }
 ``` 
 
+Note that `transferAndCall`, `transferFromAndCall` and `approveAndCall` MUST revert if called on an EOA as it can't return the [ERC-1363](./eip-1363.md) expected value.
```
Pandapip1, line 194:
```suggestion
Note that `transferAndCall`, `transferFromAndCall` and `approveAndCall` MUST revert if called on an EOA as it can't return the [EIP-1363](./eip-1363.md) expected value.
```

EIPS/eip-1363.md
```diff
@@ -191,11 +191,13 @@ interface ERC1363Spender {
...
+
 ## Rationale
-The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [ERC-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender.
+The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [ERC-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender contracts.
```
Pandapip1, line 197:
```suggestion
The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [EIP-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender contracts.
```

EIPS/eip-1363.md
```diff
@@ -33,7 +33,7 @@ Examples could be
 * paying invoices
 * making subscriptions
 
-For these reasons it was named as **"Payable Token"**.
+For these reasons it was originally named as **"Payable Token"**.
```
Pandapip1, line 36:
```suggestion
For these reasons it was originally named **"Payable Token"**.
```

EIPS/eip-1363.md
```diff
@@ -11,17 +11,17 @@ requires: 20, 165
...
-Standard functions a token contract and contracts working with tokens can implement to make a token Payable.
+Standard functions a token contract and contracts working with this token can implement to make a callback on receiver or spender.
 
 `transferAndCall` and `transferFromAndCall` will call an `onTransferReceived` on a `ERC1363Receiver` contract.  
 
 `approveAndCall` will call an `onApprovalReceived` on a `ERC1363Spender` contract.
 
 ## Motivation
-There is no way to execute code after a [ERC-20](./eip-20.md) transfer or approval (i.e. making a payment), so to make an action it is required to send another transaction and pay GAS twice.
+There is no way to execute code after an [ERC-20](./eip-20.md) transfer or approval (i.e. making a payment), so to make an action it is required to send another transaction and pay GAS twice.
```
Pandapip1, line 24:
```suggestion
There is no way to execute code after an [EIP-20](./eip-20.md) `transfer` or `approval` (i.e. making a payment), so to make an action it is required to send another transaction.
```

EIPS/eip-1363.md
```diff
@@ -11,17 +11,17 @@ requires: 20, 165
 ---
 
 ## Simple Summary
-Defines a token interface for [ERC-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve`.
+Defines a token interface for [ERC-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve` in a single transaction.
 
 ## Abstract
-Standard functions a token contract and contracts working with tokens can implement to make a token Payable.
+Standard functions a token contract and contracts working with this token can implement to make a callback on receiver or spender.
```
Pandapip1, line 17:
```suggestion
The following are the transfer and callback functions standardized in this EIP:
```

EIPS/eip-1363.md
```diff
@@ -11,17 +11,17 @@ requires: 20, 165
 ---
 
 ## Simple Summary
-Defines a token interface for [ERC-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve`.
+Defines a token interface for [ERC-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve` in a single transaction.
```
Pandapip1, line 14:
```suggestion
Defines a token interface for [EIP-20](./eip-20.md) tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve` in a single transaction.
```

EIPS/eip-1363.md
```diff
@@ -191,11 +191,13 @@ interface ERC1363Spender {
...
 
 ## Backwards Compatibility
-This proposal has been inspired also by [ERC-223](https://github.com/ethereum/EIPs/issues/223) and [ERC-677](https://github.com/ethereum/EIPs/issues/677) but it uses the [ERC-721](./eip-721.md) approach, so it doesn't override the [ERC-20](./eip-20.md) `transfer` and `transferFrom` methods and defines the interfaces IDs to be implemented maintaining the [ERC-20](./eip-20.md) backwards compatibility.  
+This proposal has been inspired also by [ERC-223](https://github.com/ethereum/EIPs/issues/223) and [ERC-677](https://github.com/ethereum/EIPs/issues/677), but it doesn't override the [ERC-20](./eip-20.md) `transfer` and `transferFrom` methods and defines the interfaces IDs to be implemented maintaining the [ERC-20](./eip-20.md) backwards compatibility.  
 
 ## Security Considerations
 The `approveAndCall` and `transferFromAndCall` methods can be affected by the same issue of the standard [ERC-20](./eip-20.md) `approve` and `transferFrom` method.
```
Pandapip1, line 203:
```suggestion
The `approveAndCall` and `transferFromAndCall` methods can be affected by the same issue of the standard [EIP-20](./eip-20.md) `approve` and `transferFrom` method.
```

Pandapip1 (changes requested):
A few changes I think need to be made to bring it back up to the current EIP format.

vittominacori:
@Pandapip1 just pushed  your suggestion changes

lukehutch:
@vittominacori Do you need a PR for the changes I suggested?

vittominacori:
@lukehutch sorry, what changes?

lukehutch:
@vittominacori see my several comments above.

vittominacori:
@lukehutch check the updated file. I added all the suggested changes here. Can't find your comments here. Please check the updated file, if you want to suggest something just comment on line

Noosom (approved):

Noosom (approved):

lukehutch:
> @lukehutch check the updated file. I added all the suggested changes here. Can't find your comments here. Please check the updated file, if you want to suggest something just comment on line

I did comment on the file already. Please search this webpage for "lukehutch started a review". I left several comments.

MicahZoltu:
@lukehutch I also don't see your comments.  Did you perhaps forget to click the button that actually submits the review?

lukehutch:
Strange... they're showing inline for me just like everybody else's comments, and I don't see a button I need to click...

I'll paste screenshots below.

image

image

lukehutch (changes requested):
OK, I see a button "Submit review" -- does this make the comments visible?

By the way some of the lines I clicked on to leave comments aren't the right line, it's only letting me comment on lines that have changes in the PR that I'm commenting on, not on other lines that weren't changed.

MicahZoltu:
Yep, clicking submit review was the issue.  Sometimes I'll leave comments in the Review Comments box if the feedback is for an unchanged line.

EIPS/eip-1363.md
```diff
@@ -160,7 +160,7 @@ interface ERC1363Receiver {
 }
 ``` 
 
-A contract that wants to accept token payments via `approveAndCall` **MUST** implement the following interface:
+A contract that wants to accept ERC1363 tokens via `approveAndCall` **MUST** implement the following interface:
```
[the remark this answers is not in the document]
vittominacori: changing to sender as it it who initiate the transaction

EIPS/eip-1363.md
```diff
@@ -128,7 +128,7 @@ interface ERC165 {
 }
 ```
 
-A contract that wants to accept token payments via `transferAndCall` or `transferFromAndCall` **MUST** implement the following interface:
+A contract that wants to accept ERC1363 tokens via `transferAndCall` or `transferFromAndCall` **MUST** implement the following interface:
```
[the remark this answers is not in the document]
vittominacori: changing  `operator` to `spender` and `from` to `sender`

EIPS/eip-1363.md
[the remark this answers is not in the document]
vittominacori: I uncommented the ERC20 and ERC165 interface and listed as extension. 
I think it should not have the IERC20 interfaceId ad ERC20 doesn't.

EIPS/eip-1363.md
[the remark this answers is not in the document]
vittominacori: Updating notes text

vittominacori:
@lukehutch just pushed the requested changes.
I updated params name. I've already done it in my own implementation but not updated the EIP. 
As we are updating now it should be safe to rename some params.

EIPS/eip-1363.md
```diff
@@ -11,41 +11,41 @@ requires: 20, 165
...
 
 ## Specification
-Implementing contracts **MUST** implement the [ERC-1363](./eip-1363.md) interface as well as the [ERC-20](./eip-20.md) and [ERC-165](./eip-165.md) interfaces.
+Implementing contracts **MUST** implement the [EIP-1363](./eip-1363.md) interface as well as the [EIP-20](./eip-20.md) and [EIP-165](./eip-165.md) interfaces.
 
 ```solidity
 pragma solidity ^0.8.0;
 
-interface ERC1363 /* is ERC20, ERC165 */ {
+interface ERC1363 is ERC20, ERC165 {
```
lukehutch, line 48: I think you should specify whether just `0xb0202a11 == type(ERC1363).interfaceId` should be registered via ERC-165, or also `type(ERC1363).interfaceId ^ type(ERC20).interfaceId ^ type(ERC165).interfaceId`. I actually think the latter is more relevant, because both ERC20 and ERC165 are required to use ERC1363.

EIPS/eip-1363.md
```diff
@@ -181,24 +181,27 @@ interface ERC1363Spender {
...
+  function onApprovalReceived(address sender, uint256 amount, bytes memory data) external returns (bytes4);
 }
 ``` 
 
+Note that `transferAndCall`, `transferFromAndCall` MUST revert if the recipient is an EOA address, because EOA recipients do not implement the required ERC1363Receiver interface.  
+Note that `approveAndCall` MUST revert if the spender is an EOA address, because EOA spenders do not implement the required ERC1363Spender interface.
+
 ## Rationale
-The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [ERC-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender.
+The choice to use `transferAndCall`, `transferFromAndCall` and `approveAndCall` derives from the [EIP-20](./eip-20.md) naming. They want to highlight that they have the same behaviours of `transfer`, `transferFrom` and `approve` with the addition of a callback on receiver or spender contracts.
```
lukehutch, line 198: Small suggested tweak: "with the addition of a callback" => "with the addition of a required callback".

Also, I would add to the end of this paragraph an explanation for why this even matters:

> Requiring a callback on the receiver prevents tokens from accidentally being sent to contracts that do not know how to receive or spend them (which has caused the loss of millions of dollars worth of tokens previously). Requiring a callback on a spender contract ensures that intended spenders are notified that they have been granted an allowance to spend tokens on the sender's behalf, while ensuring that the spender supports this functionality.

EIPS/eip-1363.md
```diff
@@ -181,24 +181,27 @@ interface ERC1363Spender {
...
    *  unless throwing
    */
-  function onApprovalReceived(address owner, uint256 value, bytes memory data) external returns (bytes4);
+  function onApprovalReceived(address sender, uint256 amount, bytes memory data) external returns (bytes4);
 }
 ``` 
 
+Note that `transferAndCall`, `transferFromAndCall` MUST revert if the recipient is an EOA address, because EOA recipients do not implement the required ERC1363Receiver interface.  
+Note that `approveAndCall` MUST revert if the spender is an EOA address, because EOA spenders do not implement the required ERC1363Spender interface.
+
```
lukehutch, line 196: I think it is important to be explicit about the following point here, even though the correct order is obvious in the NatSpec comments:

> `transferAndCall` and `transferFromAndCall` MUST call the `onTransferReceived` method in the receiver only AFTER updating the token contract's internal state, such as transferring the requested amount from the sender's balance to the recipient's balance, and reducing the ERC20 allowance by the approved amount (in the case of `transferFromAndCall`), in order to eliminate the chance of reentrancy attacks.

> `approveAndCall` MUST call the `onApprovalReceived` method in the spender only AFTER all updating the token contract's internal state, such as granting the spender the approved allowance for spending the sender's tokens, in order to eliminate the chance of reentrancy attacks.

EIPS/eip-1363.md
```diff
@@ -181,24 +181,27 @@ interface ERC1363Spender {
    * approval. Return of other than the magic value MUST result in the
    * transaction being reverted.
    * Note: the token contract address is always the message sender.
-   * @param owner address The address which called `approveAndCall` function
-   * @param value uint256 The amount of tokens to be spent
+   * @param sender address The address which called `approveAndCall` function
+   * @param amount uint256 The amount of tokens to be spent
    * @param data bytes Additional data with no specified format
    * @return `bytes4(keccak256("onApprovalReceived(address,uint256,bytes)"))`
```
lukehutch, line 187: I think you should give the easy way to calculate this, using a function selector:

`* @return ERC1363Spender.onApprovalReceived.selector == bytes4(keccak256("onApprovalReceived(address,uint256,bytes)"))`;

Same for the receiver interface too.

lukehutch (changes requested):
A few more suggested tweaks and clarifications

EIPS/eip-1363.md
lukehutch: Maybe even also be explicit about the fact that the transaction should revert if a call to `onTransferReceived` or `onApprovalReceived` reverts. This is important because there is a big move towards using `address.call` now for calling external contracts, and `call` does not automatically revert if the call fails for some reason.

JESSENIA1983 (approved):

EIPS/eip-1363.md
vittominacori: I think that `type(ERC20).interfaceId` is not relevant as it is clearly expressed that 
`Implementing contracts MUST implement the EIP-1363 interface as well as the EIP-20 and EIP-165 interfaces.`

EIPS/eip-1363.md
vittominacori: These are already expressed in code comments 

```
   * @dev Any ERC1363 smart contract calls this function on the recipient
   * after a `transfer` or a `transferFrom`. This function MAY throw to revert and reject the
   * transfer. Return of other than the magic value MUST result in the
   * transaction being reverted.`
   ```

```
   * @dev Any ERC1363 smart contract calls this function on the recipient
   * after an `approve`. This function MAY throw to revert and reject the
   * approval. Return of other than the magic value MUST result in the
   * transaction being reverted.
   ```

Don't want to make the spec so repetitive.

EIPS/eip-1363.md
vittominacori: Uhm yes, right. But `bytes4(keccak256("onApprovalReceived(address,uint256,bytes)"))` will always remain the same. The interface name could be different in different implementations instead. 
I mean I do not want to create confusion neither I want to block the interface name.

In my own implementation I named the interface as `IERC1363Spender` (not `ERC1363Spender`) so the returned value should be `IERC1363Spender(this).onApprovalReceived.selector` that is different from what you suggested also if it is the right value. I would leave the returned calculated value method to the implementer. He MUST return the `bytes4(keccak256("onApprovalReceived(address,uint256,bytes)"))` value.

Note that I used the one you suggested in [my implementation](https://github.com/vittominacori/erc1363-payable-token/blob/master/contracts/payment/ERC1363Payable.sol#L75) but I prefer to leave it as is in the spec.

EIPS/eip-1363.md
vittominacori: Updated motivation here https://github.com/ethereum/EIPs/pull/5167/commits/3c3627831c05a061c87283fdc917bd4f6afd2f4b

EIPS/eip-1363.md
```diff
@@ -11,36 +11,36 @@ requires: 20, 165
...
-Standard functions a token contract and contracts working with tokens can implement to make a token Payable.
+The following are the transfer and callback functions standardized in this EIP:
 
 `transferAndCall` and `transferFromAndCall` will call an `onTransferReceived` on a `ERC1363Receiver` contract.  
 
 `approveAndCall` will call an `onApprovalReceived` on a `ERC1363Spender` contract.
 
 ## Motivation
-There is no way to execute code after a [ERC-20](./eip-20.md) transfer or approval (i.e. making a payment), so to make an action it is required to send another transaction and pay GAS twice.
+There is no way to execute code after an [EIP-20](./eip-20.md) `transfer` or `approval` (i.e. making a payment), so to make an action it is required to send another transaction and pay GAS twice.
```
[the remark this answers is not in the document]
vittominacori: This has been updated with a new text. Please check here.

https://github.com/ethereum/EIPs/pull/5167/files#diff-d03e52deda369b57949a2bf014c8b41fb8a5c84e072f7fed33b192ed748c2714R24

vittominacori:
@lightclient yes in fact my original commit only changes few words and typos https://github.com/ethereum/EIPs/commit/7dc4bb7af35880b0a4e075bb5726757c9ff36fd9 but later reviewer has requested many text changes and to remove any external link to be compliant to new guideline.

Consider that nothing in the original EIP was changed except for a clarification on what it does with EOA. 
I contributed to deploy several ERC1363 tokens with my own implementation package and with token generator and all of them already behave as per the current EIP.

Now it is more clean and follow the last guideline on EIP creation.

lukehutch:
Correct, the core issue was only that the original EIP said that receiver contracts MUST implement the receiver interface and spender contracts MUST implement the spender interface, but it did not specify what happens if the receiver or spender is an EOA. ERC 777 and ERC 4524 explicit allow for successfully sending to EOAs, but the reference implementation of ERC1363 did not. This led to a disagreement over the OpenZeppelin implementation of ERC1363, where I was saying OpenZeppelin should look to the reference implementation to resolve the ambiguity in wording, but the OpenZeppelin guy who created the implementation interpreted the ambiguous wording the opposite way, allowing for sending to, and spending by, EOAs. He said that the reference implementation shouldn't count for anything, because it was not linked from the spec (the link had to be removed before the spec was finalized, at the request of a reviewer), and he pointed out that git repos could be changed at any time anyway.

All the other changes (including any that I suggested, at least the ones that were accepted) are only clarifications to wording or resolutions of ambiguity that do not materially change the semantics of this EIP.

MicahZoltu:
The question one must ask for changes to final EIPs is:
Is it possible for someone to have deployed something that was compliant before the change and non-compliant after the change.  If so, then the change is normative and cannot be included.  Generally speaking, this means that there are functionally no changes allowed, including clarification of ambiguous behavior.

I think the best you can do for this EIP in this situation is to add a note in the Rationale or Specification indicating that both interpretations are valid and people interacting with or implementing this EIP cannot make assumptions about which one is correct.

The person you mentioned is correct here, anything not in the Specification section of the EIP is purely supplemental, and the specification is the only thing that matters as far as what the specification is.

vittominacori:
@MicahZoltu
yes but the original EIP says

`Defines a token interface for ERC-20 tokens that supports executing recipient code after transfer or transferFrom, or spender code after approve.` and EOA does not allow to execute code.

and then

image

here there are no references to EOA but it refers to **contracts**.

And also `It allows to make a callback after a transfer or approval in a single transaction.` that is not possible with EOA. EOA will never make a callback in a single transaction.

`A contract that wants to accept token payments via transferAndCall or transferFromAndCall MUST implement the following interface: ...` and `A contract that wants to accept token payments via approveAndCall MUST implement the following interface: ...` they always refer to contracts.

And again `...with the addition of a callback on receiver or spender.` that is not possible on EOA.

And this is the original implementation linked in EIP discussion topic https://github.com/vittominacori/erc1363-payable-token.

I mean that also if the behavior with EOA was not explicit, it is really clear that this EIP does not consider addresses because the purpose was to make contracts able to receive callbacks.

At OpenZeppelin they added a PR to introduce EIP1363 because they want to have a `safeTransfer` method as expressed [here](https://ethereum-magicians.org/t/why-isnt-there-an-erc-for-safetransfer-for-erc20/7604/7). They want to force EIP1363 `transferAndCall` to use the `safeTransfer` behavior taken from ERC4524 in order to not to add both in their library.
I think we must not change the EIP1363 original behavior to just "win" the battle with ERC4524 and be added in OpenZeppelin library. ERC1363 original purpose was to be used with contracts. If we want a safeTransfer that do not revert on EOA we can use ERC4524. 

I just wanted to clarify that behavior adding a note in specification (if for some people that was not already clear), then reviewers requested typo and semantic changes so I accepted. If this PR is more than what can be accepted I will close and then open again just with the notes on EOA behavior.

lukehutch:
@MicahZoltu I disagree that the current EIP should be interpreted this way -- I interpret it the opposite way -- and I am not even the author. The author @vittominacori has made the intent behind the EIP clear here.

vittominacori:
@MicahZoltu 

```
    function transferAndCall(
        address to,
        uint256 amount,
        bytes memory data
    ) public virtual override returns (bool) {
        transfer(to, amount);
        require(_checkOnTransferReceived(_msgSender(), to, amount, data), "ERC1363: _checkOnTransferReceived reverts");
        return true;
    }
```

```
    function _checkOnTransferReceived(
        address sender,
        address recipient,
        uint256 amount,
        bytes memory data
    ) internal virtual returns (bool) {
        if (!recipient.isContract()) {
            return false;
        }
        bytes4 retval = IERC1363Receiver(recipient).onTransferReceived(_msgSender(), sender, amount, data);
        return retval == IERC1363Receiver.onTransferReceived.selector;
    }
```

this is my suggested implementation. 

Also if we remove the `if (!recipient.isContract()) { return false; }` the retval will never be the requested value `retval == IERC1363Receiver.onTransferReceived.selector`. So why we are dealing on EOA when it is required to return the selector value? We are speculating on a missing declaration that was clear for me when I introduced the EIP. I do not mention EOA because they are not part of this spec. The only one requesting for EOA to not revert is OpenZeppelin that wanted to use EIP1363 (that is final) with the ERC4524 (draft) behavior. 

The EIP1363 was born to make developers able to build callbacks on contracts after a token transfer or approve. They must be sure that the callback has been executed. So there are no reasons to make the token able to transfer and call an EOA because an EOA cannot make any action and this may break the token workflow as it expects the callback to be executed. To transfer to an EOA we already have the `transfer` method.
If people wanted to use an ambiguous method that sometimes returns a bytes4 and makes a callback and sometimes does nothing without notify nobody they could use ERC4524.

I think that this PR just clarify a thing that was implied in the original EIP.

lukehutch:
That's a good point -- what if nothing were changed about the EIP other than the suggested implementation be added?

vittominacori:
That suggested implementation is taken from my own implementation repo https://github.com/vittominacori/erc1363-payable-token

I think we must not edit the EIP adding implementation, but just adding the note about the EOA behavior. 
Others are only typos or changes to phrase wording.

vittominacori:
That suggested implementation is taken from my own implementation repo https://github.com/vittominacori/erc1363-payable-token

I think we must not edit the EIP adding implementation, but just adding the note about the EOA behavior. 
Others are only typos or changes to phrase wording.

MicahZoltu:
I can appreciate that this EIP may not represent the *intent* of the authors, but once an EIP is final intent no longer matters.  The specification currently says that you need to take a specific action when `transferAndCall` is made, and it doesn't say that the action is conditional on whether the recipient is an EOA or a contract so that means the action must be taken regardless of whether the recipient is an EOA or a contract.

Unfortunately, the EIP also doesn't specify how the token contract should deal with recipients who do not return the expected return value, so we have to assume that compliant tokens can interpret the return value however they want (including either failing the transfer or allowing the transfer).

My recommendation at this point would be to create a new EIP that makes your intent very clear.  When authoring this new EIP, I recommend you go into great detail how a compliant implementation should deal with all scenarios.  For example, specify the behavior when the recipient doesn't return the expected value,e and specify the behavior when they do, and specify the behavior when the recipient throws.

vittominacori:
> I can appreciate that this EIP may not represent the intent of the authors, but once an EIP is final intent no longer matters. The specification currently says that you need to take a specific action when transferAndCall is made, and it doesn't say that the action is conditional on whether the recipient is an EOA or a contract so that means the action must be taken regardless of whether the recipient is an EOA or a contract.

>Unfortunately, the EIP also doesn't specify how the token contract should deal with recipients who do not return the expected return value, so we have to assume that compliant tokens can interpret the return value however they want (including either failing the transfer or allowing the transfer).

Yes it doesn't specify what to do with other return values, but token specification says to `return true unless throwing` and that it must call `onTransferReceived` on receiver:

```
  /**
   * @notice Transfer tokens from `msg.sender` to another address and then call `onTransferReceived` on receiver
   * @param to address The address which you want to transfer to
   * @param value uint256 The amount of tokens to be transferred
   * @return true unless throwing
   */
  function transferAndCall(address to, uint256 value) external returns (bool);
```

And receiver says:

```
  /**
   * @notice Handle the receipt of ERC1363 tokens
   * @dev Any ERC1363 smart contract calls this function on the recipient
   * after a `transfer` or a `transferFrom`. This function MAY throw to revert and reject the
   * transfer. Return of other than the magic value MUST result in the
   * transaction being reverted.
   * Note: the token contract address is always the message sender.
   * @param operator address The address which called `transferAndCall` or `transferFromAndCall` function
   * @param from address The address which are token transferred from
   * @param value uint256 The amount of tokens transferred
   * @param data bytes Additional data with no specified format
   * @return `bytes4(keccak256("onTransferReceived(address,address,uint256,bytes)"))`
   *  unless throwing
   */
  function onTransferReceived(address operator, address from, uint256 value, bytes memory data) external returns (bytes4);
```

So basically that `Any ERC1363 smart contract calls this function on the recipient after a 'transfer' or a 'transferFrom'.` and `Return of other than the magic value MUST result in the transaction being reverted.`.

Anyway, I think that what I'm saying is already specified. 
In my opinion calling `transferAndCall` on EOA must revert like a non compliant contract.
Token contracts that do not revert are not compliant with the ERC1363 specification also without adding the below notes.

```
Note that `transferAndCall`, `transferFromAndCall` MUST revert if the recipient is an EOA address, because EOA recipients do not implement the required ERC1363Receiver interface.  
Note that `approveAndCall` MUST revert if the spender is an EOA address, because EOA spenders do not implement the required ERC1363Spender interface.
```

So, if we can't add the note to clarify you can also close this PR but it doesn't mean that we can use `transferAndCall` to transfer tokens to EOA.

lukehutch:
@MicahZoltu the fact is that both you and the OlenZeppelin implementer interpreted the EIP wrongly, despite what (to me) seems quite clear language that specifies the opposite intent. So please give some latitude to fix this, as allowed by the EIP errata rules.

MicahZoltu:
The fact that there are multiple interpretations from multiple people is precisely why a change should not be allowed.  Standards that reach final are final, even if they include ambiguities that result in multiple implementations.

----

@vittominacori A token can correctly implement this this EIP and interact with a non-compliant receivers since this EIP doesn't specify how a compliant token should behave when interacting with a non-compliant receiver.  Essentially, since the EIP doesn't specify a specific behavior when interacting with a non-compliant receiver (such as an EOA) it is up to the individual token to decide how to behave in that scenario.

vittominacori:
The reason because OpenZeppelin wanted to force use the non intended behavior was to avoid including both 1363 and 4524. There are tons of tokens built using my implementation and no one reported that issue. If a dev wants to use the `transferAndCall` he should have clear what the call means so basically that the callback must be executed and not dealing with addresses hoping that the callback was executed. This was my initial point of view.

Anyway if the EIP cannot be updated, please close that PR. 
Everyone will be able to build their own implementation and both will be compliant.

PR #5178, opened 2022-06-24
Update EIP-4494: Include EIP-1271 and EIP-2098 as required

TimDaub:
EIP-1271:
- EIP-4494 currently makes no mention of EIP-1271 but I strongly assume that there is interest to support signature validation through a smart contract
- There are strong reasons to believe so as [@dievardump's reference implementation](https://github.com/dievardump/erc721-with-permits/blob/main/contracts/ERC721WithPermit.sol#L116) uses OZ's [`SignatureChecker` that in turn implements EIP-1271](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol#L36)
- Unclear to me: Whether or not the authors of the EIP document want to further mention 1271. Anyways, at least I think it should be mentioned in the header as required.

EIP-2098:
- document mentions: "sig is a valid secp256k1 or [EIP-2098](https://eips.ethereum.org/EIPS/eip-2098) signature from owner of the tokenId:"

### DIFF
--- a/EIPS/eip-4494.md
+++ b/EIPS/eip-4494.md
@@ -12,19 +12,19 @@ requires: 165, 712, 721, 1271, 2098
 ---
 
 ## Abstract
-The "Permit" approval flow outlined in [ERC-2612](./eip-2612.md) has proven a very valuable advancement in UX by creating gasless approvals for ERC20 tokens. This EIP extends the pattern to ERC-721 NFTs. This EIP borrows heavily from ERC-2612.
+The "Permit" approval flow outlined in [EIP-2612](./eip-2612.md) has proven a very valuable advancement in UX by creating gasless approvals for [EIP-20](./eip-20.md) tokens. This EIP extends the pattern to [EIP-721](./eip-721.md) NFTs. This EIP borrows heavily from EIP-2612.
 
-This requires a separate EIP due to the difference in structure between ERC-20 and ERC-721 tokens. While ERC-20 permits use value (the amount of the ERC-20 token being approved) and a nonce based on the owner's address, ERC-721 permits focus on the `tokenId` of the NFT and increment nonce based on the transfers of the NFT.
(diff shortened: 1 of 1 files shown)


Pandapip1:
@MicahZoltu @lightclient could one of you manually merge this?

MicahZoltu:
We need to get approval from one of the authors before merging.  The author may prefer to remove the reference to 2098 to resolve that one, and they may not want to require 1271 for some reason.  We should not let 4494 go to review without resolving the 2098 issue though.

TimDaub:
@dievardump @wschwab ☝🏼

wschwab:
I don't see why 1271 or 2098 should be required. We have designed the EIP to be compatible with them, but at no point in the implementation are they required. Feel free to change my mind.

Pandapip1:
> Feel free to change my mind.

It should be explicitly stated that they may be supported, then, because EIP-1271 is assumed to not be supported.

TimDaub:
> I don't see why 1271

fair.

> 2098 should be required

You mention EIP-2098 as also @dievardump points out, and it is how you format the signature as `bytes`, and in that case, IMO, it should be added to the requirements.

TimDaub:
> we point it out, but we never say that a 4494-compliant implementation MUST support either, imho a requirement should be something required, not something that was merely accommodated

fine. defining it like this can be a pro and a con.

Pro: if a better compact signature scheme arrives it may be usable

con: being loose can lead to undefined behavior.

I tend to prefer being very explicit and specific but it's ur guy's call, feel free to close

EIPS/eip-4494.md
```diff
@@ -1,30 +1,30 @@
 ---
 eip: 4494
-title: Permit for ERC-721 NFTs
-description: ERC-712-singed approvals for ERC-721 NFTs
+title: Permit for EIP-721 NFTs
+description: ERC-712-singed approvals for EIP-721 NFTs
```
xinbenlv, line 4: @Pandapip1 it seems a false positive warning for EIPW? My understanding is that preamble can't add Markdown Links.

xinbenlv (approved):

EIPS/eip-4494.md
Pandapip1, line 4:
```suggestion
description: EIP-712-singed approvals for EIP-721 NFTs
```

Pandapip1 (changes requested):

PR #5198, opened 2022-07-02
move eip-173 to Final

z0r0z:


Pandapip1:
This is going to need to be updated to the current EIP format.

PR #5209, opened 2022-07-08
Adjust EIP-4973 interface to add `function give(...)` and `function take(...)`

TimDaub:
This PR isn't ready for merging or review yet. But if you wanna help, feel free to comment below anyways.

TODOs:

- [x] Address lint errors of EIP bot
- [x] Add requirements how give and take functions are supposed to be implemented
- [x] Add a section on how structured data hash can be generated from JS and Solidity (and potentially add the type JSON structure itself
- [x] Adjust all old language around `event Revoke` and `event Attest`
- [x] Change all mentions of `function burn` to `function unequip`
- [ ] Consider adding a recommendation for content-addressable `string tokenURIs` as was outlined here https://github.com/otterspace-xyz/EIPs/pull/1/files#diff-78eda02f8170e73250f20c44a63a94888296116f3fec90171d5e766ce312f5afR228
- [x] Add flat file reference implementation: https://github.com/rugpullindex/ERC4973/blob/25bb347db0448842964d375e887b295663ee97ed/assets/ERC4973-flat.sol


Pandapip1:
How about `grant`/`revoke`?

TimDaub:
> How about `grant`/`revoke` instead of `give`/`take`?

neither give or take "revoke" something. For both give and take the agreement of a second party is required:

- take(from,... signature) takes from "address from" (in case signature is valid) and mints to the msg.sender
- give(to,..., signature) mints from the msg.sender to "address to" if their signature is valid

TimDaub:
> * Consider adding a recommendation for content-addressable `string tokenURIs` as was outlined here https://github.com/otterspace-xyz/EIPs/pull/1/files#diff-78eda02f8170e73250f20c44a63a94888296116f3fec90171d5e766ce312f5afR228

Decided against implementing this change for now. Will maybe follow up later when we have better information.

PR #5210, opened 2022-07-08
Move EIP-173 to final with minor (normal) changes

Pandapip1:
CC: @mudgen

### DIFF
--- a/EIPS/eip-173.md
+++ b/EIPS/eip-173.md
@@ -8,7 +8,6 @@ type: Standards Track
 category: ERC
 status: Final
 created: 2018-06-07
-requires: 2767
 ---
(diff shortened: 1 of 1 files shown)


mudgen:
@Pandapip1 
Looks good, except the standard does not require EIP-2767, and I think could cause some confusion to says so.   So just remove the mentioning of EIP-2767 if the standard can't mention it without requiring it.

Pandapip1:
The `requires` header is a bit misleading in its name: it contains all the EIPs that are _referenced_.

Pandapip1:
CC @mudgen

mudgen (approved):

PR #5219, opened 2022-07-11
PR-5219: Contract Resource Requests

Pandapip1:

 - The build passes.


EIPS/eip-5219.md
```diff
@@ -0,0 +1,72 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-07-10
+requires: 20, 137, 165, 4834
+---
+
+## Abstract
+
+Ethereum is the most-established blockchain for building decentralized applications (referred to as `DApp`s). Due to this, the Ethereum DApp ecosystem is very diverse. However, one issue that plagues DApps is the fact that they are not fully decentralized. Specifically, to interface a "decentralized" application, one first needs to access a *centralized* website containing the DApp's front-end code. This presents a few issues, discussed in the Motivation.
```
MicahZoltu, line 16: The abstract currently describes a problem, not a solution.  The abstract should try to describe the solution and leave a description of the problem to the Motivation section.

Pandapip1: I think I probably got sidetracked and forgot to write about the solution. I'll write that now.

Pandapip1:
Re-trigger CI

EIPS/eip-5219.md
```diff
@@ -0,0 +1,75 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-07-10
+requires: 137, 165, 4804, 4834
+---
+
+## Abstract
+
+Ethereum is the most-established blockchain for building decentralized applications (referred to as `DApp`s). Due to this, the Ethereum DApp ecosystem is very diverse. However, one issue that plagues DApps is the fact that they are not fully decentralized. Specifically, to interface a "decentralized" application, one first needs to access a *centralized* website containing the DApp's front-end code, presenting a few issues (discussed in the Motivation). To solve these issues, this EIP introduces an interface allowing clients to make HTTP-like requests (containing a method, a resource to request, a request body, and headers) to smart contracts, and to receive HTTP-like responses (containing headers and a body).
```
MicahZoltu, line 16: This feels like mostly a motivation rather than abstract.  The abstract should describe the specification in a pithy and human readable way, and leave the *why* to the motivation.

EIPS/eip-5219.md
```diff
@@ -0,0 +1,75 @@
...
+### Name Resolution
+
+It is RECOMMENDED for clients implementing this EIP to recognize the following standards:
+
+| Name                          | Specification                                        |
+| ---                           | ---                                                  |
+| [EIP-137](./eip-137.md) (ENS) | See [EIP-137](./eip-137.md#resolver-specification)   |
+| [EIP-4834](./eip-4834.md)     | See [EIP-4834](./eip-4834.md#specification)          |
+
+**Changes to this list:** This list MAY be appended to even when this EIP becomes final. New additions should be made at the bottom of the list.
```
MicahZoltu, line 40: EIPs are immutable once final, so you will not be able to add items to the specification in the future.  Generally speaking, it is recommended to keep lists of things out of EIPs because they can't be maintained over time and to instead use some other mechanism for maintaining a list of things.

EIPS/eip-5219.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Separation of Concerns
+
+It is RECOMMENDED to separate the application logic from the front-end logic (the contract implementing the interface defined in [Contract Interface](#contract-interface)).
+
+### Contract Interface
+
+DApp contracts MUST implement the interface defined in the following file: [Contract Interface](../assets/eip-5219/IDecentralizedApp.sol).
+
+In addition, DApp contracts MUST implement [EIP-165](./eip-165.md).
```
MicahZoltu, line 54: Why is this required?  It doesn't feel like it is strictly necessary for this to be a useful specification, and one can imagine a future mechanism for interface detection that is superior that the ecosystem wants to switch to, and it would be unfortunate for anything using that newer/better system to no longer be compatible with this EIP.

assets/eip-5219/IDecentralizedApp.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: CC0-1.0
+pragma solidity ^0.8.0;
+
+interface IDecentralizedApp {
+    /// @notice                     Send an HTTP-like request to this contract
+    /// @param  method              The HTTP method to use (e.g. GET, POST, PUT, DELETE)
+    /// @param  resource            The resource to request (e.g. "/asdf/1234" turns in to ["asdf", "1234"], and "/demo?qwerty=uiop" turns in to ["demo"])
+    /// @param  parameters          The parameters to send with the request (e.g. "querty=uiop&1234=5678" would be ["qwerty", "1234"])
+    /// @param  values              The values to send with the request (e.g. "querty=uiop&1234=5678" would be ["uiop", "5678"])
```
MicahZoltu, line 9: Why not an array of tuples rather than two arrays?

assets/eip-5219/IDecentralizedApp.sol
```diff
@@ -0,0 +1,16 @@
...
+    /// @param  headers             A list of header names (e.g. ["X-Foo-Bar"])
+    /// @param  headerValues        A list of header values (e.g. ["asdf"])
```
MicahZoltu, line 11: Same question as above, why not an array of tuples?

assets/eip-5219/IDecentralizedApp.sol
```diff
@@ -0,0 +1,16 @@
...
+    /// @return body                The body of the response
```
MicahZoltu, line 12:
```suggestion
    /// @return statusCode
    /// @return body                The body of the response
```

MicahZoltu (commented):
It is unclear to me how exactly this solves the problems you proposed.  We already have a mechanism for calling functions against Ethereum contracts and getting responses, along with an enchoding scheme for all of that (the solidity function call ABI).  I don't understand what we gain by having a function with an abstract function calling mechanism in it.

assets/eip-5219/IDecentralizedApp.sol
Pandapip1: Huh, I'm not sure why I didn't think of that.

Pandapip1:
> I don't understand what we gain by having a function with an abstract function calling mechanism in it.

Think about this EIP as standardizing a way for smart contracts to become websites. HTTP just happens to be the current standard for that, and so this EIP tries to mimic that.

lightclient (approved):
LGTM

PR #5236, opened 2022-07-12
Migrate Travis CI to GitHub Actions

Pandapip1:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #5241, opened 2022-07-13
Add Jekyll Caching Plugin

Pandapip1:
Rationale: The Jekyll build and HTML check is the rate-limiting factor of the CI.

~~Blocked by #5236, as I have no way of locally testing these changes.~~

Never mind, forgot I could just point github pages to another branch. Here is the result: https://pandapip1.github.io/EIPs/

The build takes about ~25s faster.


Pandapip1:
😓 that was close

Gemfile.lock
```diff
@@ -1,274 +0,0 @@
-GEM
```
SamWilsn, line 1: Why remove the lockfile?

Pandapip1: The alternative is that I add something to the CI to keep it up to date. Would that be preferable?

SamWilsn: Something like dependabot? Yeah, I'd be okay with that, as long as a human has to approve the change.

Pandapip1:
Let's try this again, but this time, the CI makes the gemfile.

Pandapip1:
I give up.

PR #5253, opened 2022-07-14
Editorial suggestions for EIP-3541

xinbenlv:
Understood that EIP-3541 is already final, but wondering if it makes more sense to specify EVM Object Format (EOF) with what it refer to.

FYI: @axic


axic:
I think this can't be merged because EIP-3540 is of lesser state (i.e. not `Final`) than EIP-3541.

MicahZoltu:
Going to close this for the reason Axic indicated.  Feel free to open it again after 3540 is final or withdrawn.

PR #5256, opened 2022-07-14
Revert "Small fix for when the Gemfile is updated"

Pandapip1:
Reverts ethereum/EIPs#5249

It didn't work 😢


MicahZoltu (approved):

MicahZoltu:
We should probably be careful about introducing external bot dependencies.  The bot has quite a bit of power, and it would suck to have the repository ruined with a supply chain attack.

PR #5258, opened 2022-07-14
EIP-4750: Make type section optional

gumb0:
Type section in the container spec should better be optional in order to maintain backwards-compatibility with EOF1 spec.


JESSENIA1983 (approved):

SamWilsn:
Just checking if this is still supposed to be a draft.

chfast (approved):

PR #5266, opened 2022-07-15
Force LF line endings

Pandapip1:
This is required for `eipw` to successfully run locally.


.gitattributes
```diff
@@ -1,3 +1,7 @@
 # GitHub highlighting for Solidity files
 # See https://github.com/github/linguist/pull/3973#issuecomment-357507741
 *.sol linguist-language=Solidity
+
+# Force Linux line endings on all files
+# Necessary for running eipw locally
+* text=auto eol=lf
```
SamWilsn, line 7: Why not just for `*.md` files? Would be weird if a `*.bat` got converted to unix line endings :P

MicahZoltu: I believe this line is pretty standard, and I don't think `.bat` files break if they get `LF` newlines.  While we *could* try to come up with a custom list of all possible files that may need special casing, I don't think we really gain anything from that.

Pandapip1: I would prefer if all the files use the same line ending, for consistency reasons (why add unnecessary complexity?).

SamWilsn: Alright, go for it then!

Pandapip1:
This will need to be manually merged. ÇC: @MicahZoltu @lightclient

PR #5272, opened 2022-07-15
Add RFC Information

Pandapip1:
- Adds links RFC 822 and RFC 2119
- Adds rules to deal with RFCs

Change to EIP-1: See files tab


Pandapip1:
Huh, who knew? RFCs have spelling mistakes.

EIPS/eip-1.md
```diff
@@ -265,11 +265,15 @@ The `description` field in the preamble:
 
 When referring to an EIP by number, it should be written in the hyphenated form `EIP-X` where `X` is the EIP's assigned number.
 
+### IETF RFCs
+
+When referring to an IETF RFC by number, it must be written in the non-hyphenated form `RFC X`, where `X` is the RFC's assigned number. Each RFC that is referenced in an EIP **MUST** be accompanied by a relative markdown link to the RFC (in the `assets/eip-1/RFCs` directory) the first time it is referenced, and **MAY** be accompanied by a link on subsequent references. If the RFC is not currently in that directory, a separate pull request must be submitted which only includes the RFC in that directory.
```
SamWilsn, line 270: Why a separate pull request?

SamWilsn, line 270: Also, we want to store all RFCs in EIP-1's asset directory? That will reduce duplicates for sure, but it feels weird.

I'm not opposed to this pattern, just wanted to bring it up.

EIPS/eip-1.md
```diff
@@ -265,11 +265,15 @@ The `description` field in the preamble:
...
+
 ### RFC 2119
 
-EIPs are encouraged to follow [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) for terminology and to insert the following at the beginning of the Specification section:
+EIPs are encouraged to follow [RFC 2119](../assets/eip-1/RFCs/rfc2119.txt) for terminology and to insert the following at the beginning of the Specification section:
```
SamWilsn, line 274: It looks like you're using a lowercase directory, but an uppercase link. Not sure if that's bad or not.

EIPS/eip-1.md
Pandapip1: I would prefer if the git history wasn't polluted with random unrelated PRs. @MicahZoltu's the editor that's probably going to have the strongest opinion on this front.

Pandapip1: Would it be better to have a separate `RFCs` directory in the assets folder? I strongly would like there to be a single place that contains all the RFCs.

EIPS/eip-1.md
Pandapip1: I thought I renamed the directory, huh. I'll fix that.

EIPS/eip-1.md
gcolvin: So just `assets/RFCs`?   Fetching them is yet another thing a bot could do ;)

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         with:
           check_filenames: true
           ignore_words_file: .codespell-whitelist
-          skip: .git,Gemfile.lock,**/*.png,**/*.gif,**/*.jpg,**/*.svg,.codespell-whitelist,vendor,_site,_config.yml,style.css
+          skip: .git,Gemfile.lock,**/*.png,**/*.gif,**/*.jpg,**/*.svg,.codespell-whitelist,vendor,_site,_config.yml,style.css,assets/eip-1/RFCs/*.txt
```
MicahZoltu, line 48: This ignores RFCs in `assets/eip-1/RFCs` but the RFCs have been added to `assets/RFCS`.

assets/RFCs/README.md
```diff
@@ -0,0 +1,9 @@
+# Internet Engineering Task Force (IETF) Request for Comments (RFC)
+
+This folder is dedicated to storing relevant RFCs. If you would like to add an RFC to this folder, please submit a pull request adding it!
+
+According to the [IETF's FAQ](https://trustee.ietf.org/about/faq/#collapse3103):
+
+> Am I allowed to reproduce whole RFCs?
+> 
+> Yes. Since the beginning of the RFC series, reproduction of whole RFCs (including translation into a language other than English) has been allowed and encouraged. The IETF Trust and the RFC Editor place no restrictions on this. Most RFCs include the standard phrase “Distribution of this memo is unlimited” to indicate this.
```
MicahZoltu, line 9: I think I would prefer to leave this file out, as I would rather not encourage people to add RFCs (it should be a special case, not a common case I think).

MicahZoltu: Hmm, it appears the proposal here isn't to just add these two RFCs, but rather introduce a process by which any RFC can be imported.  I'm in favor of bringing in the two RFCs, I will need to think more on creating a path toward inclusion of *any* RFC.  I am not strongly against it, though I do worry a bit about "you allow RFCs, why not also support my-new-standard-repository-that-i-created-yesterday?"

.github/workflows/ci.yml
Pandapip1: Fixed.

assets/RFCs/README.md
Pandapip1: > I do worry a bit about "you allow RFCs, why not also support my-new-standard-repository-that-i-created-yesterday?"

I think this is more of a theoretical problem than one that is actually likely to occur.

MicahZoltu: I wish that was the case, but I have had multiple people who are owners of other standards repositories ask to reference their standards repository (e.g., for some other blockchain) in their EIPs.  😖

gcolvin:
There  is also a good argument for _not_ copying RFCs.  The RFC Editor maintains a canonical list at <https://www.rfc-editor.org>.  For each RFC there is an Info page with TXT, PDF, and HTML renderings of the RTF and a host of information about the RFC.  The RFC style guide at <https://www.rfc-editor.org/info/rfc7322> requires that links to RFCs be to their Info page, _not_ to a particular rendering.  The required form is <https://www.rfc-editor.org/info/rfc#>.  Why not just do as they ask?

MicahZoltu:
@gcolvin Where in the style guide does it say that third parties referring to RFCs should not include the raw text?

gcolvin:
> @gcolvin Where in the style guide does it say that third parties referring to RFCs should not include the raw text?

@MicahZoltu  It doesn't.  It requires that links to RFCs be to their Info page, presumably because it is more informative.
> https://www.rfc-editor.org/rfc/rfc7322.html#section-4.8.6.2
We are free to choose a particular rendering and copy it anywhere we like, but I am doubting that we should.

lightclient:
I am generally against enforcing the "no external links" decree on RFCs.

Pandapip1:
So here's my thought: we create a page that redirects to the specified RFC info page, while keeping text backups of the RFCs. If for some reason the info page dies, we can have it redirect to the text backups instead.

assets/RFCs/README.md
Pandapip1: Well then, we can just... say no to them (or have them convert all of their standards repo into EIPs!)

lightclient:
I think that solution is too hacky. It would require users either type the entire `eips.ethereum.org/rfc-X` url for links (which is in contrast of our current relative link policy) or we use relative links that are resolved only once the site has been deployed, meaning the links can't be accessed when viewing on github.

SamWilsn:
I guess an option would be to allow direct links to the RFC, but have the linter fail if a backup isn't made?

gcolvin:
Why are we afraid of RFC links failing?  The whole Internet is built to IETF Standards, and IETF Standards themselves use those links.  They aren't going away.

What I would suggest is that we provide a Normative Reference section with full citations for both EIPs and RFCs, as is required by IETF for RFCs.

Pandapip1:
> Why are we afraid of RFC links failing? The whole Internet is built to IETF Standards, and IETF Standards themselves use those links. They aren't going away.

This argument could be made for many links. By not allowing any external links, we avoid a lot of unneeded discussions. And after all, the internet _could_ die...

What's the problem with linking to text-based RFCs? Yes, it might not be perfect according to their style guide. But I feel like keeping all information in the EIPs repository is worth the small tradeoff.

MicahZoltu:
I'm pretty sure the style guide linked by @gcolvin above is specifically for people drafting RFCs, not for people referring to RFCs from elsewhere.

In the past we have discussed managing a whitelist of "acceptable" external links (like RFCs), but the challenge is in maintaining that list and dealing with people requesting that their thing be added to it.  While RFCs certainly seem like an *obviously* right choice, it opens the doors to everyone claiming their standardization thing is also a right choice.

I think the primary difference here (including the RFC text in this repository) is that anyone can do this for any source of standard being referenced (assuming it is license compatible) and we don't have to get into the weeds about what each different standards committee's policy is on immutability and whatnot.

gcolvin:
> @MicahZoltu  I'm pretty sure the style guide linked by @gcolvin above is specifically for people drafting RFCs, not for people referring to RFCs from elsewhere.  

[RFC 7322](https://www.rfc-editor.org/info/rfc7322) is the style guide for how to refer to RFCs from within another RFC.  They also have guidance for other types of references.  I strongly advise that we follow their guidance.  In essence
> ... The use of URIs in references is acceptable, as long as the URI is the most stable (i.e., unlikely to change and expected to be continuously available) and direct reference possible.
> ... Note that URIs may not be the sole information provided for a reference entry.

So like the IETF I just don't have a problem with the judicious use of external references, normative and informative. RFC 7322 itself has a lot of informative references.  I don't find "we are afraid they might cause too much work in the future" a convincing argument for imposing unnecessary restrictions on EIP authors.  Authors, reviewers, and editors can simply exercise good judgment, and if we can't get the resources we need to create high-quality documents we are in trouble regardless.

I also don't have a problem with putting an extra copy in our repo when we are allowed to.  The citation in the Reference section can include both links.  But course that means being sure that we have the legal right to publish a copy, which creates its own problems.

Pandapip1:
> as long as the URI is the most stable (i.e., unlikely to change and expected to be continuously available) and direct reference possible

For the purposes of the EIPs repository, the EIPs repository itself is the most stable and direct reference possible.

gcolvin:
It's stable and direct, and also misses the full context of the Info page.  I've suggested we put both the canonical link to the IETF and a link to our copy in the Reference section.  And please let's use the HTML rendering of the RFC, which includes links to the other renderings and to the Info page.

JESSENIA1983 (approved):

Pandapip1:
> And please let's use the HTML rendering of the RFC, which includes links to the other renderings and to the Info page.

I'd be okay with that.

Pandapip1:
@SamWilsn EIPW bug?

gcolvin:
> If the HTML rendering includes links to the other renderings, then we're good, right?

Good enough, thanks.

gcolvin:
> with all anchors replaced with internal anchors

What does this mean?

Pandapip1:
> What does this mean?

By default, all the anchors in the HTML rendering of the RFCs have an "external" link to themselves before the anchor part.

gcolvin:
> > > with all anchors replaced with internal anchors
> > 
> > What does this mean?

> By default, all the anchors in the HTML rendering of the RFCs have an "external" link to themselves before the anchor part.

So you are asking the EIP author to edit the links inside of the RFC?

Regardless, this is getting to be a lot more hassle than just linking directly to the canonical RFC Info page that the IETF maintains.

And regardless, this is getting to be (please forgive my annoyance) a silly level of paranoia about external links. But I'll need to bring this up at a more general level because - like the IETF - I don't in general have any problem with external links, and strongly object to our draconian policy.

Pandapip1:
The problem isn't that the EIP repository might outlive the internet. That's unlikely. What I do believe to be the problem is that the EIP repository might outlive the HTTP and HTTPS standards.

gcolvin:
> EIP repository might outlive the HTTP and HTTPS standards.
I'm sorry, but that's exactly what I mean by silly paranoia.  If HTTPS goes away the modern world will stop working, including our repo.

Pandapip1:
By outlive the standards, I meant that a new standard, say super-http-variant://, becomes the norm, to the point that even the IETF website has migrated.

There's the issue.

gcolvin:
If the IETF migrates their website then every single RFC that refers to another RFC will need to be updated.  And the many, many other links to the RFCs will be broken.   And again, if HTTP changes that much then every link on the web would break.  It's far more like that Github will go away.

So please, if insist on linking to our own copy let's just use the IETF's HTML rendering as it is.

Pandapip1:
Stahp!

gcolvin:
I think we don't have the right to publish modified copies (derivative works) of RFCs.
> The IETF Trust does not in general grant the right to create derivative works of RFCs; in fact it does not have the right to do so, under the current IETF rules (RFC 5378).
https://trustee.ietf.org/about/faq/#reproducing-rfcs

Pandapip1:
> I think we don't have the right to publish modified copies (derivative works) of RFCs.

Well, we aren't doing that anymore anyways.

gcolvin:
> Well, we aren't doing that anymore anyways.
Just noticed that - thanks!

EIPS/eip-1.md
```diff
@@ -267,7 +267,7 @@ When referring to an EIP by number, it should be written in the hyphenated form
 
 ### IETF RFCs
 
-When referring to an IETF RFC by number, it must be written in the non-hyphenated form `RFC X`, where `X` is the RFC's assigned number. Each RFC that is referenced in an EIP **MUST** be accompanied by a relative markdown link to the RFC (in the `assets/RFCs` directory) the first time it is referenced, and **MAY** be accompanied by a link on subsequent references. If the RFC is not currently in that directory, a separate pull request must be submitted which ONLY includes the HTML rendering of that RFC, with all anchors replaced with internal anchors, in that directory.
+When referring to an IETF RFC by number, it must be written in the non-hyphenated form `RFC X`, where `X` is the RFC's assigned number. Each RFC that is referenced in an EIP **MUST** be accompanied by a relative markdown link to the RFC (in the `assets/RFCs` directory) the first time it is referenced, and **MAY** be accompanied by a link on subsequent references. If the RFC is not currently in that directory, a separate pull request must be submitted which ONLY includes the HTML rendering of that RFC.
```
xinbenlv, line 270: If the RFC is not currently in that directory, a separate pull request ~~must~~ **MUST** be submitted which ONLY includes the HTML rendering of that RFC.

Pandapip1: Yes.

Pandapip1:
@MicahZoltu @lightclient @axic @SamWilsn @gcolvin is everyone happy with this in its current state?

xinbenlv:
@lightclient I agree with you on this

Pandapip1:
If most people are in agreement that external links to RFCs are fine, then I guess we could do that. It's definitely the low-maintenance option!

MicahZoltu:
My only concern with externally linking to RFCs is that it opens the doors for people saying, "why can't I link to <my personal standards repository> too?"  And this isn't a far fetched outlier scenario, we actually have had multiple people with brand new "standards" organizations that they want to link externally to.  By requiring we embed the content here, it makes it so the process isn't special for RFCs, as anyone can add documents to their assets and link to them relatively.

xinbenlv:
If EM meet our bar of trust I find it hard to convince not to trust ietf.org ...

MicahZoltu:
> If EM meet our bar of trust I find it hard to convince not to trust ietf.org ...

They serve quite different purposes.  The discussion link isn't part of the standard/specification, and if it does disappear from the internet the standards are all still valid/usable.  On the other hand, if an external link disappears the standard may no longer make sense.

MicahZoltu:
In the meeting, we discussed maybe allowing links to third party standards libraries, but only ones that have been around for X years.  Both 3 years and 30 years were proposed, though I think both fairly flippantly.

I personally support this technique, though I'm not a fan of 3 years.  30 years basically would limit us to a very small number of standards groups, but 3 I feel like opens us up to a lot of stuff that is too new to know if it will survive.

How would people feel about "10 years".  That would allow us to start including BIPs in about a little over a year from now, which feels just about right to me in that I think BIPs are the borderline case for me.

xinbenlv:
On Wed, Jul 27, 2022 at 23:45 Micah Zoltu ***@***.***> wrote:

> In the meeting, we discussed maybe allowing links to third party standards
> libraries, but only ones that have been around for X years. Both 3 years
> and 30 years were proposed, though I think both fairly flippantly.
>
> I personally support this technique, though I'm not a fan of 3 years. 30
> years basically would limit us to a very small number of standards groups,
> but 3 I feel like opens us up to a lot of stuff that is too new to know if
> it will survive.
>
> How would people feel about "10 years". That would allow us to start
> including BIPs in about a little over a year from now, which feels just
> about right to me in that I think BIPs are the borderline case for me.
>
10year seems a good line to me. Also would we consider internet archive.org
for the newer outlinks? (coming from my editor experience on Wikipedia, in
which a time machine bot will automatically issue archive request for a
link to archive.org and when it becomes unopenable, update the link to
point to archived version

MicahZoltu:
I feel like if we limit to standards organizations that have been around and publishing standards for at least 10 years, then we shouldn't need to explicitly back anything up as they are well enough established that even if the main site goes down we can find copies elsewhere.

That being said, it would be cool if someone were to write a bot that automatically snapshotted the whole EIPs repository plus all external links to standards to IPFS on a regular basis and periodically posted the IPFS hash somewhere.  This wouldn't have to be anything official, anyone could do this as a separate useful project.

Pandapip1:
@MicahZoltu @axic @lightclient @SamWilsn @gcolvin any objections?

lightclient:
Is `rfc-editor.org` the correct resource to point to? Is there any documentation on how to web reference an RFC?

Pandapip1:
> Is rfc-editor.org the correct resource to point to?

There's also the datatracker (which I would prefer) but @gcolvin suggested using rfc-editor.org instead.

> Is there any documentation on how to web reference an RFC?

Yes, but none of it specifies which version to link to.

xinbenlv:
I agree with @MicahZoltu on that
> That being said, it would be cool if someone were to write a bot that automatically snapshotted the whole EIPs repository plus all external links to standards to IPFS on a regular basis and periodically posted the IPFS hash somewhere. This wouldn't have to be anything official, anyone could do this as a separate useful project.

I am happy to take on this task to build such bot, if no one has yet

SamWilsn:
My biggest issue with allowing links to other standards bodies is that they may not enforce the same immutability rules that we do. For example, [WHATWG](https://whatwg.org/) has existed for over ten years, so linking to their standards would be allowed. Unfortunately their standards are all living, so we'd need to make sure EIPs link to a snapshot and not the most recent version.

Pandapip1:
Well, the 10-year rule is an informal rule for now. All this EIP adds is RFCs.

xinbenlv:
Per https://github.com/ethereum/EIPs/pull/5272#issuecomment-1196279935, I filed a feature request to build a archive bot for #5408

Pandapip1:
Moving to draft until the PR with the external link changes is made.

Pandapip1:
Closing in favor of a new PR once the external link changes are done.

PR #5289, opened 2022-07-18
PR-5289: Ethereum Notary Interface

Pandapip1:
Allows interoperability between Ethereum smart contracts and the real world.


Pandapip1:
That was close 😓.

xinbenlv:
This is cool.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
...
+created: 2022-07-16
+requires: 165
+---
+
+## Abstract
+
+Currently, the real-world applications of smart contracts are limited by the fact that they aren't legally binding. This EIP proposes a standard that allows smart contracts to be legally binding by providing IPFS links to legal documents and ensuring that the users of the smart contract have privity with the relevant legal documents.
+
+(NOTE: None of the authors is a lawyer. If you are a lawyer and would like to contribute to this EIP, please contact Pandapip1 on Discord (Pandapip1#8943). This text will be removed by the time that this EIP goes into last call.)
+
```
MicahZoltu, line 19:
```suggestion
```
This type of content should be in the discussions-to link, not in the EIP itself.  EIPs should be authored toward their final form and not contain transient information.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
...
+    uint8 size;
+    bytes32 digest;
+}
+
+interface ILegalDocument is IERC165 {
+    /// @notice The IPFS multihash of the legal document. This MUST resolve to a PDF.
+    function legalDocument() public pure returns (string memory);
+    
+    /// @notice The date the legal document was published (in days since January 1, 1970).
+    function datePublished() public pure returns (uint64);
```
MicahZoltu, line 45: It feels like this should just be a unix timestamp, which is well standardized as a way to store a date in a number.  Also, why isn't the publish date in the PDF itself rather than stored separately?

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
...
+
+    /// @notice The ECDSA signature of the given user signing the IPFS hash.
+    function documentSignature(address user) public view returns (bytes memory);
+
+    /// @notice Provide an ECDSA signature of the given user signing the IPFS hash.
+    /// @dev This MUST be validated by the smart contract.
+    function signDocument(address user, bytes memory signature) public;
```
MicahZoltu, line 52: Does signing the document carry some meaning, or is that left up to the implementations?  Currently, anyone can sign with this method, including anonymous people.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
...
+
+interface ILegalDocument is IERC165 {
+    /// @notice The IPFS multihash of the legal document. This MUST resolve to a PDF.
+    function legalDocument() public pure returns (string memory);
+    
+    /// @notice The date the legal document was published (in days since January 1, 1970).
+    function datePublished() public pure returns (uint64);
+
+    /// @notice The ECDSA signature of the given user signing the IPFS hash.
+    function documentSignature(address user) public view returns (bytes memory);
```
MicahZoltu, line 48: Since there is no way to iterate users, it means you need to get a user from somewhere else off-chain.  Why not just store the signature there instead of in contract storage?

EIPS/eip-5289.md
MicahZoltu: It feels like this function should fire an event, so logs can be searched for signatures and users.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
...
+All EIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The EIP must explain how the author proposes to deal with these incompatibilities. EIP submissions without a sufficient backwards compatibility treatise may be rejected outright.
+
+## Test Cases
+
+Test cases for an implementation are mandatory for EIPs that are affecting consensus changes.  If the test suite is too large to reasonably be included inline, then consider adding it as one or more files in `../assets/eip-####/`.
+
+## Reference Implementation
+
+An optional section that contains a reference/example implementation that people can use to assist in understanding or implementing this specification.  If the implementation is too large to reasonably be included inline, then consider adding it as one or more files in `../assets/eip-####/`.
+
```
MicahZoltu, line 71:
```suggestion
```
Optional sections, can be left out until they contain something.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
...
+}
+```
+
+## Rationale
+
+The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages.
+
+## Backwards Compatibility
+
+All EIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The EIP must explain how the author proposes to deal with these incompatibilities. EIP submissions without a sufficient backwards compatibility treatise may be rejected outright.
```
MicahZoltu, line 62:
```suggestion
TBD
```
TBD stands out more as "incomplete" than the placeholder text.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
...
+
+    /// @notice Provide an ECDSA signature of the given user signing the IPFS hash.
+    /// @dev This MUST be validated by the smart contract.
+    function signDocument(address user, bytes memory signature) public;
+}
+```
+
+## Rationale
+
+The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages.
```
MicahZoltu, line 58:
```suggestion
TBD
```
TBD stands out more as "incomplete" than the placeholder text.

Pandapip1: Huh, I thought I had done the rationale. Oops!

EIPS/eip-5289.md
Pandapip1: I'm almost certain I did the backward compatibility. I'm starting to think that I forgot to push a commit 😆

EIPS/eip-5289.md
Pandapip1: Okay, definitely forgot to push a commit.

EIPS/eip-5289.md
Pandapip1: Why do users need to be iterated through?

EIPS/eip-5289.md
Pandapip1: Great idea!

EIPS/eip-5289.md
Pandapip1: > Does signing the document carry some meaning?

Yes, it does, but it is indeed left up to the implementations.

> Currently, anyone can sign with this method, including anonymous people.

This is by design. IIUC the document can be worded in such a way that it can be used by anonymous people.

EIPS/eip-5289.md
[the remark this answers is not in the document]
Pandapip1: No. 64 bits will last us until the year 570776255712 - i.e. longer than the universe has been around.

EIPS/eip-5289.md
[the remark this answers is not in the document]
Pandapip1: 1. Good idea. I'll probably make that change.
2. I would like it, but since I'm supporting EIP-1271 for smart contracts to be able to sign, it won't work.
3. Why?

EIPS/eip-5289.md
```diff
@@ -0,0 +1,78 @@
+---
+eip: 5289
+title: Legal Smart Contracts
```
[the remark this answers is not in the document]
Pandapip1: Yup, that sounds better.

EIPS/eip-5289.md
MicahZoltu: They don't, I'm just arguing that this function is unnecessary and it increases the cost of adding signatures to the system because it forces implementations to store the signatures on-chain unnecessarily.  Whatever mechanism an implementation has for finding a signer address, that same mechanism could be used to find the signer signature.

If this interface supported user iteration, then it would make more sense to have the signatures on-chain, because it would be reasonable to do signer discovery on-chain and (per above) the signatures should be where the user list is.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,146 @@
...
+    function legalDocument(uint48 documentId) public pure returns (bytes memory);
+    
+    /// @notice Returns whether or not the given user signed the document
+    function isDocumentSigned(address user, uint48 documentId) public view returns (boolean signed);
+
+    /// @notice The block timestamp of when the user signed. Throws if not signed.
+    function documentSignedAt(address user, uint48 documentId) public view returns (uint64 timestamp);
+
+    /// @notice Provide a signature
+    /// @dev This MUST be validated by the smart contract. This MUST emit DocumentSigned.
```
MicahZoltu, line 60:
```suggestion
    /// @dev This MUST be validated by the smart contract. This MUST emit DocumentSigned or throw.
```
Might be redundant, but I like overly explicit EIPs personally.

MicahZoltu:
> FYI - I changed the category to interface as it is probably more appropriate for what I am trying to accomplish.

Standards Track: ERC is appropriate here.  All ERCs are interfaces.  `interface` is generally used for this like client APIs like the JSON-RPC API or the new Engine API (though, both of those aren't managed in this repository, which is why Interface is almost never used).

EIPS/eip-5289.md
Pandapip1: As it happens, I replaced this method in a later version with one that only gives a timestamp. I agree with your assessment.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,139 @@
+---
+eip: 5289
+title: Notary Interface
```
xinbenlv, line 3:
```suggestion
title: Smart Notary Interface
```
After a second thought, I think we can add a **Smart** keyword before **Notary**: the benefit is that if the interface gets adoption, the wording can be easier to be used in area outside of blockchain (e.g. popular media) to refer to the fact this concept comes from smart contract, yet for such reader it also shows the relevance to the concept of notary, making it more spreadable and understandable. Just my two cents, feel free to reject.

MicahZoltu: Why not `Ethereum Notary Interface` or something?  I don't think `Smart` really means much these days, I have a "Smart Light" which has nothing to do with Ethereum.

Pandapip1:
```suggestion
title: Ethereum Notary Interface
```

Pandapip1: Ethereum notary interface it is!

EIPS/eip-5289.md
```diff
@@ -0,0 +1,139 @@
+---
+eip: 5289
+title: Smart Notary Interface
```
Pandapip1, line 3:
```suggestion
title: Ethereum Notary Interface
```

EIPS/eip-5289.md
```diff
@@ -0,0 +1,139 @@
...
+## Reference Implementation
+
+### Legal Contract Library
+
+```solidity
+/// SPDX-License-Identifier: CC0-1.0
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/utils/introspection/ERC165Storage.sol";
+import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
```
MicahZoltu, line 97:
```suggestion
import "ERC165Storage.sol";
import "SignatureChecker.sol";
```
It should be an exercise for the reader to decide where to get their dependencies.  That being said, `SignatureChecker` isn't an EIP so this should probably be expanded out at least into an inline interface with comments.

EIPS/eip-5289.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Reference Implementation
+
+### Legal Contract Library
+
+See [`ERC5289Library`](../assets/eip-5289/ERC5289Library.sol).
+
+## Security Considerations
+
+No security considerations.
```
SamWilsn: Not required for a draft, but a note about the definition of privity and legal jurisdictions might be nice before final

SamWilsn (approved):

PR #5252, opened 2022-07-14
Add EIP-5252: Account-bound Finance

hskang9:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #5290, opened 2022-07-18
Make the CI Commenting Cleaner

Pandapip1:
Deletes the previous comment before posting a new one.

### DIFF
--- a/.github/workflows/ci-comment.yml
+++ b/.github/workflows/ci-comment.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event.workflow_run.conclusion == 'failure' }}
     steps:
-      - name: 'Download artifact'
+      - name: Download Artifact
         uses: actions/github-script@v6
         with:
           script: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-comment.yml
```diff
@@ -32,10 +32,17 @@ jobs:
...
-      - name: 'Unzip artifact'
+      - name: Unzip Artifact
         run: unzip pr_number.zip
+      
+      - name: Delete Previous Comment
+        uses: izhangzhihao/delete-comment@master
+        with: 
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          delete_user_name: github-actions[bot]
+          issue_number: ${{ github.event.number }}  # remove comments from the current PR
```
SamWilsn, line 43: Why remove previous comments? I've always preferred keeping everything in a coherent timeline, so you can see the history, but I guess I could be convinced otherwise.

Pandapip1: It clutters history. Very fast.

If you want to see the history of which checks passed and failed, the checks tab allows you to pull up the run results for a specific commit.

MicahZoltu:
This doesn't appear to have worked:
https://github.com/ethereum/EIPs/pull/5252#issuecomment-1189958932

Pandapip1:
Sam reverted it.

SamWilsn:
Do we want to bring it back? The rate limit has reset, so could try it again.

Pandapip1:
Honestly, I would prefer that it not comment at all.

The history of checks is easily viewable at a glance by clicking the status indicator beside each commit, and if more detail is required, from the checks panel of the GitHub PR UI. Having the bot comment every time there's a commit is... a lot of unnecessary comments.

SamWilsn:
My understanding is that it should only comment once per push, even if there are multiple commits. That said, I haven't tested it, and from what I've seen, it certainly seems to run on every applied suggestion.

I would be quite happy to not have it comment, but it is something @MicahZoltu asked for so I declare this a _contentious issue_ :rofl:

MicahZoltu:
I do think there is value because most authors don't actually look at CI status or the files section once their PR is created, so for most authors I think the commenting by the bot will be invisible.  By having the bot comment, it lets the author know "hey, you need to go look at your PR for it to move forward".

Pandapip1:
> By having the bot comment, it lets the author know "hey, you need to go look at your PR for it to move forward"

Okay, but then there should only ever be one such comment at any time, and it shouldn't trigger when the auto-merge bot "fails."

MicahZoltu:
I agree, we want to keep the comments to a minimum and only when actionable.  Ideally, only when there is a *new* thing the author needs to check but I am guessing that is hard.  I agree that deleting old comments is a good idea so the issue doesn't get filled with them.

PR #5302, opened 2022-07-21
Remove old EIP-20 file

Pandapip1:
Bye!


Pandapip1:
(I will need either @MicahZoltu or @lightclient to manually merge this once someone else approves)

MicahZoltu:
We had this here so that people who had links to it would not get a 404.  We never decided on an EOL strategy at the time, so it has sat here ever since.  It is not clear to me what the right deprecation strategy is here, whether now or sometime in the future.

Pandapip1:
Why not remove it then, and if people complain we re-add it?

MicahZoltu:
I'm not against that strategy.  It *has* been quite a while since we renamed this EIP.  Would like to hear thoughts from other editors.

Pandapip1:
Other Editor CC: @lightclient @axic @SamWilsn @gcolvin

Pandapip1:
@MicahZoltu everyone was okay with removing this file. Please *manually* merge.

PR #5311, opened 2022-07-22
Update EIP-173: Add EIP-165 dependency

fulldecent:
This is included in the text and this PR adds it to the metadata.


Pandapip1:
Personally, I wouldn't approve it because the standard would have to be brought up to current EIP formatting (and that involves non-normal changes). Alternatively, this could be manually merged. @MicahZoltu @lightclient thoughts?

fulldecent:
I went ahead and fixed other non-normative issues (section order, and metadata order) complained from the bot. 

I have not changed any text or normative things.

Pandapip1:
> I have not changed any text or normative things.

I think you have normative vs non-normative flipped around:

Normative: the meaning is unchanged.
Non-normative: the meaning is changed.

The concern was that updating the standard to current guidelines might have to result in normative changes, which we try to avoid in Final EIPs.

fulldecent:
I have not changed normative things. My changes are non-normative.

---

And in plain English: I didn't change any specs.

Pandapip1:
Sorry, let me be a bit more specific:

Some of the changes required by eipw here _require_ non-normative changes to the spec.

fulldecent:
Got it.

I try to get all non-normative changes in one PR so that gets merged with less oversight required.

And then other stuff I try to separate out, because that might get held up.

MicahZoltu (commented):
With approval by one of the authors and buy-in from the other editors, I think I would be OK with making all suggested eipw changes *except* for the `discussions-to` link change.

Regarding the currently proposed changes, `type` header field should come immediately before `category` header field so at least we aren't making the situation worse.

lightclient:
@fulldecent if you can update the `type` field to be above `category`, I will approve.

fulldecent:
@lightclient Got it, fixed at e0e0c1871072e0a93c89b0ded5c46acf9997f890

xinbenlv:
173 is FINAL, I lean towards follow IETF RFC approach to create a new EIP that "updates" EIP-165

Pandapip1:
ERC also needs to be replaced with EIP.

Pandapip1:
> 173 is FINAL, I lean towards follow IETF RFC approach to create a new EIP that "updates" EIP-165

This doesn't introduce any backwards compatibilities, and is therefore fine.

fulldecent:
This PR does not touch any mention of "ERC" or "EIP".

I am requesting that this here PR be reviewed and accepted as an incremental improvement and that any discussion of "ERC"/"EIP" be considered in a separate, isolated PR.

Pandapip1:
> This PR does not touch any mention of "ERC" or "EIP".

`eipw` still requires that you fix the instances.

xinbenlv:
> > This PR does not touch any mention of "ERC" or "EIP".
> 
> 
> 
> `eipw` still requires that you fix the instances.

I suggest EIPW to be updated so that it only check modified lines rather than full file.

Pandapip1:
> I suggest EIPW to be updated so that it only check modified lines rather than full file.

If the effort to bring an EIP up to date is not worth the change introduced in a PR, then the PR shouldn't be opened. `eipw` is specifically to *ensure* that all EIPs, regardless of age, are still of the correct format when they are changed. The errors must be fixed.

xinbenlv:
@fulldecent , You are correct. I take that back: the 173 already mandates 165. It's reasonable to add 165 as `requires`.

I am in favor of this commit to be merged without fixing all new requirement.

xinbenlv:
> > I suggest EIPW to be updated so that it only check modified lines rather than full file.
> 
> If the effort to bring an EIP up to date is not worth the change introduced in a PR, then the PR shouldn't be opened. `eipw` is specifically to _ensure_ that all EIPs, regardless of age, are still of the correct format when they are changed. The errors must be fixed.

While I can see the good heart behind the requirement, I think requiring this may unnecessarily turn away contributors who just have time to make small improvements. Let me file a issue and we can discuss it in our next EIPIP meeting

Pandapip1:
> While I can see the good heart behind the requirement, I think requiring this may unnecessarily turn away contributors who just have time to make small improvements. Let me file a issue and we can discuss it in our next EIPIP meeting

That's the point. We *want* to discourage small contributions as they take up precious editor time.

fulldecent:
Rather than telling editors what they can't merge, in an effort to save their time... let's instead tell editors what they can merge, while minimizing the amount of synchronous back-and-forth required for them to do so.

Basically all my PRs get merged in this repo, that's why I keep making them.

Pandapip1:
I'm fine with @SamWilsn merging this then.

Pandapip1 (approved):
I think the consensus has changed to "EIPW doesn't matter for Final EIPs."

CC @SamWilsn for manual merge

xinbenlv (approved):

Pandapip1:
CC (@mudgen, @danfinlay) for approval

Issue #5310, opened 2022-07-22
GitHub actions don't run for first time contributors


SamWilsn:
### Pull Request

https://github.com/ethereum/EIPs/pull/5164

### What happened?

`eipw` and friends didn't run automatically for this PR since this is the authors first time contributing.

Pandapip1:
This is due to a branch protection rule. CC @MicahZoltu @lightclient

SamWilsn:
I don't think so. I think this is just the default for GitHub actions.

MicahZoltu:
Any idea what happens if a first-time-contributor submits a PR that touches a `.workflow` file, such as to print out secrets?  While the number of second time editors to this repo is incredibly high, at least the current setup makes it a *little* harder to execute such an attack...

MicahZoltu:
We should probably limit the set of actions that can be run to protect against that attack vector more broadly, though I think there are actions that let you run arbitrary shell commands so I'm not sure that would help much.

Pandapip1:
Needs reopening.

MicahZoltu:
> We should probably limit the set of actions that can be run to protect against that attack vector more broadly, though I think there are actions that let you run arbitrary shell commands so I'm not sure that would help much.

The prescience of this comment is alarming.  😬

Pandapip1:
Still an issue.

Pandapip1:
Still an issue :|

Pandapip1:
Still an issue

Pandapip1:
Still an issue.

Pandapip1:
Still an issue.

Pandapip1:
Still an issue.

lightclient:
What do we need to do to resolve this?

Pandapip1:
Unsure.

Pandapip1:
Still an issue :|

Pandapip1:
I think this is still an issue.

Pandapip1:
I actually haven't seen this happen recently, so I'll assume that something solved this.

Issue #5298, opened 2022-07-20
Pin all GitHub Action dependencies to hash version.


MicahZoltu:
All GitHub action references should be like https://github.com/ethereum/EIPs/blob/a182311ee869fa8e5c3e7d4717a1ba5d856be8b4/.github/workflows/auto-merge-bot.yml#L16
Not like https://github.com/ethereum/EIPs/blob/a182311ee869fa8e5c3e7d4717a1ba5d856be8b4/.github/workflows/ci-comment.yml#L15

This protects us from supply chain attacks against this repository as only GitHub (who already has the ability to completely screw us) could sneak in a malicious change to a dependency, which has privileged access.

Pandapip1:
I feel like the more pressing issue is the fact that any of us editors can just arbitrarily change the CI on a whim. Want to add an action that deletes EIP-1? Go ahead! Want to add and/or remove other editors for the fun of it? Nobody's stopping you (except for the two editors with manual merge permissions who can undo it).

Pandapip1:
> This protects us from supply chain attacks against this repository as only GitHub (who already has the ability to completely screw us) could sneak in a malicious change to a dependency, which has privileged access.

Tags can't be modified. If we target a release, then this would provide the required protection without being too... hashy.

MicahZoltu:
> I feel like the more pressing issue is the fact that any of us editors can just arbitrarily change the CI on a whim. Want to add an action that deletes EIP-1? Go ahead! Want to add and/or remove other editors for the fun of it? Nobody's stopping you (except for the two editors with manual merge permissions who can undo it).

Yes, we should fix this as well.  I think there is already an issue for this somewhere?  If not, we should create one.

> Published tags can't be modified, as they are locked to a specific commit. If we target a release, then this would provide the required protection without being too... hashy.

Are you sure about this?  GitHub will happily let you delete a git tag and publish a new one pointing at a new commit.  I would be mildly surprised to hear that they follow a different set of rules for their actions system.

Pandapip1:
> GitHub will happily let you delete a git tag and publish a new one pointing at a new commit. 

Huh, I didn't know this. Good to know!

Pandapip1:
Needs reopening.

Pandapip1:
Still an issue

Pandapip1:
Anyone want to do this?

Pandapip1:
Still an issue

Pandapip1:
See #5797.

PR #5312, opened 2022-07-22
Improvify GitHub Action

Pandapip1:
Fixes https://github.com/ethereum/EIPs/issues/5298, Fixes https://github.com/ethereum/EIPs/issues/5310, and fixes a lot of ugliness

Yes, I know improvify isn't a word and that I should use improve.

Will need manual merge from @MicahZoltu or @lightclient


Pandapip1:
I keep on forgetting these things will auto merge....

Pandapip1:
Never mind!

.github/workflows/ci.yml
```diff
@@ -80,13 +86,18 @@ jobs:
...
         with:
           name: pr_number
           path: pr/
+
       - name: Checkout EIP Repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-      - uses: ethereum/eipw-action@master
+
+      - name: Run EIPW
+        uses: ethereum/eipw-action@ca327b6089abbb281e4c51eda75c3bdbac0e01f8
```
SamWilsn, line 100: Until we're happy with the lints for `eipw`, would it be okay to keep this pointing at `master`?

It's under the `ethereum` organization, so I think the risk of supply chain attacks is a bit lower.

.github/workflows/ci.yml
```diff
@@ -1,45 +1,52 @@
...
+
     runs-on: ubuntu-latest
-    
+
     steps:
       - name: Checkout EIP Repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-      
+        with:
+          ref: "refs/pull/${{ github.event.number }}/merge"
```
SamWilsn, line 17: Oh this is neat, so it'll run with the workflow file from `master`, but fetch from the PR?

.github/workflows/ci.yml
```diff
@@ -49,26 +56,31 @@ jobs:
...
+      - name: Set Up Rust Toolchain
+        uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f
+        with:
+          toolchain: stable
+
+      - name: Install EIP Validator
+        run: cargo install eipv --version=0.4.0
+
+      - name: Run EIP Validator
+        run: eipv EIPS/ --ignore=title_max_length,missing_discussions_to --skip=eip-20-token-standard.md
```
MicahZoltu, line 77: Why was this indented?  At a glance, the original indentation was also correct I think?  YAML is dumb, but I believe both of the following are correct forms for a list
```yml
key:
- item1
- item2
- item3
```
```yml
key:
  - item1
  - item2
  - item3
```

.github/workflows/ci.yml
```diff
@@ -1,45 +1,52 @@
...
     name: CodeSpell
+
     runs-on: ubuntu-latest
+
     steps:
       - name: Checkout EIP Repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-        
+        with:
+          ref: "refs/pull/${{ github.event.number }}/merge"
```
MicahZoltu, line 48: What does this do?

.github/workflows/ci.yml
Pandapip1: Yes

.github/workflows/ci.yml
Pandapip1: Because the indentation was the latter for every single other job, so I matched them for consistency.

.github/workflows/ci.yml
Pandapip1: Pulls the files from the PR. I changed the event to `pull_request_target`, which always runs (even for new contributors) but also changes the default checkout behaviour to checkout the master branch of the original repo.

MicahZoltu (approved):

Rachi26:
Nice

PR #5314, opened 2022-07-22
Move to last call

fulldecent:
I have no changes left to make here.

Requesting last call status to raise visibility here and see if anybody objects here.

### DIFF
--- a/EIPS/eip-5313.md
+++ b/EIPS/eip-5313.md
@@ -4,8 +4,7 @@ title: Light Contract Ownership
 description: An interface for identifying ownership of contracts
 author: William Entriken (@fulldecent)
 discussions-to: https://ethereum-magicians.org/t/eip-5313-light-contract-ownership/10052
-status: Last Call
-last-call-deadline: 2022-08-05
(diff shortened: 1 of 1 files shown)


EIPS/eip-5313.md
```diff
@@ -4,7 +4,8 @@ title: Light Contract Ownership
 description: An interface for identifying ownership of contracts
 author: William Entriken (@fulldecent)
 discussions-to: https://ethereum-magicians.org/t/eip-5313-light-contract-ownership/10052
-status: Draft
+status: Last Call
+last-call-deadline: 2022-08-05
```
Pandapip1, line 8: I would like to push this back a bit further (maybe to end of month). Draft -> Last Call is quite the jump!

fulldecent: Updated this PR to move to Review. This can merge now. And then will make a separate one for Last Call. What can do later.

SamWilsn (approved):
I think this is fine to move to review. I did leave a comment on your discussion thread too.

PR #5319, opened 2022-07-23
Re-Add Comment Cleaner

Pandapip1:
Only needs to be one workflow file because now we use `pull_request_target`!


Pandapip1:
(Obligatory requires manual merge from @MicahZoltu or @lightclient)

.github/workflows/ci.yml
```diff
@@ -82,23 +82,6 @@ jobs:
...
-          echo $PR_NUMBER > ./pr/pr_number
-          echo $PR_SHA > ./pr/pr_sha
-          echo $MERGE_SHA > ./pr/merge_sha
-
-      - name: Upload Artifact
-        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8
-        with:
-          name: pr_number
-          path: pr/
-
```
MicahZoltu, line 101: This feels unrelated to the other change.  Am I missing something?  If not, can this be split into two separate PRs so they can be discussed independently?

MicahZoltu: Oh, I see.  Previously we saved an artifact and then used that to determine whether to comment or not, but now we comment in the same workflow so it is easier to see results and comment accordingly.

MicahZoltu (approved):
Waiting to merge until @SamWilsn comments, since EIPW is Sam's thing.

.github/workflows/ci.yml
```diff
@@ -109,3 +92,31 @@ jobs:
...
+
+    runs-on: ubuntu-latest
+    if: needs.htmlproofer.result == 'failure' || needs.codespell.result == 'failure' || needs['eip-validator'].result == 'failure' || needs['eipw-validator'].result == 'failure'
+    needs: [ "htmlproofer", "codespell", "eip-validator", "eipw-validator", "clean-comments" ]
+    
+    steps:
+      - name: Comment on PR
+        uses: thollander/actions-comment-pull-request@686ab1cab89e0f715a44a0d04b9fdfdd4f33d751
+        with:
+          message: The commit ${{ github.event.pull_request.head.sha }} (as a parent of ${{ github.sha }}) contains errors. Please inspect the [Files Tab](https://github.com/ethereum/EIPs/pull/${{ github.event.number }}/files) for actionable items.
```
SamWilsn, line 121: In the event multiple commits get run simultaneously, and a later commit finishes before an earlier commit, will the non-deleted comment point to the latest check run?

I was thinking of adding an if-statement to my script to only comment if the check run is for the latest commit. Should we do that here too?

SamWilsn (approved):
Looks fine to me, but left a question

.github/workflows/ci.yml
Pandapip1: > In the event multiple commits get run simultaneously, and a later commit finishes before an earlier commit, will the non-deleted comment point to the latest check run?

No, but the chance that this will happen is negligible.

.github/workflows/ci.yml
Pandapip1: This also means that the rate limit is lower.

Pandapip1:
@MicahZoltu ready to merge

Pandapip1:
The merge conflicts can't be resolved. I'll re-do this.

.github/workflows/ci.yml
MicahZoltu: It does seem like a good idea to halt the check if it isn't against the latest commit in the PR.  At the least, it could save us some CPU time.

PR #5320, opened 2022-07-24
Add EIP-5320: Harberger Taxes NFT

greenlucid:


EIPS/eip-5320.md
```diff
@@ -0,0 +1,158 @@
...
+type: Standards Track
+category: ERC
+created: 2022-07-23
+discussions-to: todo
+---
+
+### Simple Summary
+
+A standard interface for Non-Fungible Tokens subject to Harberger Taxes.
+
```
MicahZoltu, line 15: Not a valid section.  Please see https://raw.githubusercontent.com/ethereum/EIPs/master/eip-template.md for the current layout.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+### Caveats
+
+The taxes have to be collected manually, either through `collect`, or by interacting with the Harberger tokens.
+
+The taxes are paid upfront, which is less capital efficient.
+
+The functions of the standard are not passing the buyer or owner as parameter, instead, they use `msg.owner`.
```
MicahZoltu, line 109: `msg.owner` isn't a thing.  Do you mean `msg.sender`?

MicahZoltu (commented):
Make sure you address all of the feedback from the bot (see inline comments in files tab in GitHub), and address any CI failures that appear to be problems with the EIP.

veteransmod:
Solve issues here https://direct.lc.chat/14345826/

Pandapip1 (commented):
This EIP is looking good.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+The functions of the standard are not passing the buyer or owner as parameter, instead, they use `msg.sender`.
+
+The funds are per token, not per account.
+
+## Rationale
+
+### EIP-20 compatibility
+
+EIP-20 compliant tokens are a widespread industry standard and using native tokens is limiting.
```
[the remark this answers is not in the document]
greenlucid: That's for the Harberger Tax contract to decide. The deployer can define what the allowed currency is. For example, in [this implementation](https://github.com/kleros-crime-syndicate/harbs/blob/master/contracts/src/HarbergerAdsFull.sol) the currency is defined at construction time, so you can, for example, make the allowed currency be WETH.

Having a single, enforced currency is not enforced at the standard level, but ideally a contract shouldn't allow *any* EIP-20 token because it opens up these attacks.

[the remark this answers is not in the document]
Pandapip1: Could you put this in the Security Considerations?

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+  /// @dev call this to manually send pending taxes to the collector.
+  /// This function should be callable by any external party.
+  /// @param _tokenId The identifier of the token to pay taxes for.
+  function collect(uint256 _tokenId) external;
+  
+  /// @dev call this to change the perceived valuation of the token.
+  /// @param _tokenId The identifier of the token to change valuation of.
+  /// @param _valuation The new valuation of the token.
+  function changeValuation(uint256 _tokenId, uint256 _valuation) external;
```
mdnorman, line 97: since there's no "tax period", it's hard to know when taxes must be paid for someone who owns this token. I was working with CityDAO on this a proposal for using Harberger taxes on a parcel of land, and one of the things that folks need to know is when the collection period is. You don't want to make it a single date because then folks will try to game the system, and you don't want to make it "all year long" because you don't know when it will actually be assessed.

I wrote up quite a bit of detail [here](https://docs.google.com/document/d/1ufqjUU1ZSrCUlOfHJudyhF4MVTimHcEU5L_swNvXCYk/edit#heading=h.6db975ulp1ep) if you want to take a look. It may inform this EIP in some ways to make tax collection more standardized.

MicahZoltu: @mdnorman This comment would be better suited for the discussions-to link so it doesn't get lost.  Comments on the PR itself should be limited to editorial type things, while comments on the content itself should go in the discussions-to.  PRs are very transient in nature, while the discussion link is much more long lived.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+---
+
+## Abstract
+
+The following standard allows for implementation of a standard API for NFTs subject to Harberger Taxes within smart contracts. Harberger Taxes work in the following way, admitting some variants:
+
+1. Alice owns an Item, and values at an amount X. 
+2. Alice pays taxes according to a tax rate R to a collector, depending on this self-assesed value.
+3. Bob offers Alice an amount Y. Because Y is higher than X, Alice is forced to sell the Item to him.
+4. Bob self-asseses the value too at Z, and now pays taxes over it.
```
MicahZoltu, line 21:
```suggestion
4. Bob self-asseses the value at Z, and now pays taxes over it.
```

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+}
+```
+
+### Caveats
+
+The taxes have to be collected manually, either through `collect`, or by interacting with the Harberger tokens.
+
+The taxes are paid upfront, which is less capital efficient.
+
+The functions of the standard are not passing the buyer or owner as parameter, instead, they use `msg.sender`.
```
MicahZoltu, line 107:
```suggestion
The functions don't pass the buyer or owner as parameter, instead, they use `msg.sender`.
```
A standard should only define the publicly visible surface area, not implementation details.  If there is some behavior that must be standardized, then that should be called out explicitly in the specification with a **MUST** and very clear directions on what implementations must do.  This statement seems to imply that implementations must use `msg.sender`, but it doesn't specify where or how it must be used.  Either remove this line (recommended if it isn't required for interoperability) or improve its specificity.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+  /// @param _tokenId The identifier of the token that was taxed.
+  /// @param _value The taxes that have been paid.
+  event TaxPaid(uint256 indexed _tokenId, uint256 _value);
+
+  /// @dev buy a token. This function must succeed if 
+  /// @param _tokenId The identifier of the token to buy.
+  /// @param _offer The maximum amount offered to buy this token.
+  /// @param _valuation The valuation the token will have after a successful purchase.
+  /// @param _fund The fund put in to pay for the taxes of the token.
+  function buy(uint256 _tokenId, uint256 _offer, uint256 _valuation, uint256 _fund) external;
```
MicahZoltu, line 72: The specification should make it explicit that when this function is called, `collect` must be called internally before the sale completes.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+  function defund(uint256 _tokenId, uint256 _value) external;
+  
+  /// @dev call this to revoke ownership of a token.
+  /// @param _tokenId The identifier of the token to revoke.
+  function revoke(uint256 _tokenId) external;
+
+  /// @dev call this to manually send pending taxes to the collector.
+  /// This function should be callable by any external party.
+  /// @param _tokenId The identifier of the token to pay taxes for.
+  function collect(uint256 _tokenId) external;
```
MicahZoltu, line 92: The specification should assert what happens when `collect` fails for any reason.  Presumably the remaining funds would be collected and the NFT's valuation would be set to 0 (which would allow anyone to acquire the property for free)?

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+  /// @dev call this to remove excess funding for the token.
+  /// if the resultant fund for the token was zero, it must be revoked.
+  /// @param _tokenId The identifier of the token to defund.
+  /// @param _value The amount to retrieve from the token's fund.
+  function defund(uint256 _tokenId, uint256 _value) external;
+  
+  /// @dev call this to revoke ownership of a token.
+  /// @param _tokenId The identifier of the token to revoke.
+  function revoke(uint256 _tokenId) external;
```
MicahZoltu, line 87: This feels unnecessary. The owner can just set the valuation to 0, which would make the property free to acquire by anyone.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+  /// @dev call this to fund the item, to pay for the maintenance taxes.
+  /// @param _tokenId The identifier of the token to fund.
+  /// @param _value The amount sent to fund the token.
+  function fund(uint256 _tokenId, uint256 _value) external;
+
+  /// @dev call this to remove excess funding for the token.
+  /// if the resultant fund for the token was zero, it must be revoked.
+  /// @param _tokenId The identifier of the token to defund.
+  /// @param _value The amount to retrieve from the token's fund.
+  function defund(uint256 _tokenId, uint256 _value) external;
```
MicahZoltu, line 83: This should probably assert that the valuation gets set to 0 as a side effect of defunding.

EIPS/eip-5320.md
MicahZoltu, line 97: Security Considerations section should mention that `collect` **SHOULD** be called internally from this method prior to changing the valuation to avoid tax evasion.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+Tax rates could be more complex than a simple flat rate, for example, they could be progressive, or be lower or higher if the address meets some specific conditions. The taxes could have discounts if they are collected more frequently. The possibilities are too high to justify making the tax rate a standard view, globally or per item.
+
+### No commit reveal for buying tokens
+
+Commit reveal would add too much complexity to an ERC. It could be added as an extension, but it should not be part of this standard, since it already has too many functions.
+
+### Taxes and funds are per token, not per account
+
+This made for a simpler implementation, but it may be too naive. Please give feedback on how to implement this.
```
JasoonS, line 144: I still need to give my thoughts (at a retreat this week, so might have to wait).

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+```solidity
+interface ERC5320 is ERC721 {
+
+  /// @dev emits when a token is forcibly bought through the Harberger mechanism.
+  /// @param _tokenId The identifier of the token that was bought.
+  /// @param _owner The owner of the token after the purchase.
+  /// if equal to address(0), it means the token was revoked.
+  /// @param _amount The amount that was paid in this transaction.
+  event TokenBought(uint256 indexed _tokenId, address indexed _owner, uint256 _amount);
```
will-holley, line 49: "Lease" and "Leasor" are a better fit here; one of the core ideas of HT is that an asset within this system cannot be "owed" (as commonly understood within the prevailing contexts of private ownership).

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+  /// @dev emits when an token is funded with currency to pay the taxes, or when this
+  /// fund was updated.
+  /// @param _tokenId The identifier of the token that had a fund change.
+  /// @param _amount The updated amount of the fund.
+  event TokenFunded(uint256 indexed _tokenId, uint256 _amount);
```
will-holley, line 55: The standard should not assume that a deposit model is necessary.  Deposit vs. non-deposit are alternative methods with their own respective tradeoffs (which should be documented as part of this standard).

EIPS/eip-5320.md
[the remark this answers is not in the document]
will-holley: HT may be philosophically incompatible with fungibility because each token, by nature of having a self-assessed valuation, is non-fungible.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-07-23
+requires: 20, 721
+---
+
+## Abstract
+
+The following standard allows for implementation of a standard API for NFTs subject to Harberger Taxes within smart contracts. Harberger Taxes work in the following way, admitting some variants:
```
SamWilsn, line 16:
```suggestion
The following standard defines an API for NFTs subject to Harberger Taxes within smart contracts. This particular flavor of Harberger Taxes works in the following way:
```

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+1. Alice owns an Item, and values at an amount X. 
+2. Alice pays taxes according to a tax rate R to a collector, depending on this self-assesed value.
+3. Bob offers Alice an amount Y. Because Y is higher than X, Alice is forced to sell the Item to him.
+4. Bob self-asseses the value too at Z, and now pays taxes over it.
+
+This mechanic sets a stage in which actors won't overvalue the item to avoid paying excessive taxes, but won't undervalue the item either, to avoid a third party to force them to sell it.
```
SamWilsn, line 23:
```suggestion
This mechanic encourages actors to correctly value an item: overvaluing leads to excessive taxes, while undervaluing lets third parties forcefully purchase at a discount.
```

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
+The main obvious case of Harberger Taxes is to be used as a coordination mechanism for land allocation, but, to give some examples:
+
+- Real-world land allocation.
+- Digital land allocation (videogames, social experiments...).
+- Advertisement Space in websites.
+- Governance mechanisms that use "seats" subject to Harberger.
+- Social media that requires holding a token to interact.
```
SamWilsn, line 31:
```suggestion
The obvious use case of Harberger Taxes is to be used as a coordination mechanism for real world land allocation. Some other examples include:

- Digital land allocation (ex. video games, or social experiments).
- Advertisement space in websites.
- Governance mechanisms that use "seats" subject to Harberger.
- Social media that requires holding a token to interact.
```

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+
```
SamWilsn, line 32: These use cases can go in the Motivation section. They don't belong in a technical abstract.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+## Motivation
+
+Provide smart contract interoperability to interact with Harberger Taxes, which is a pricing mechanism that can provide utility to many different applications.
```
SamWilsn, line 35:
```suggestion
This proposal provides smart contract interoperability to interact with Harberger Taxes.
```

I think the abstract sufficiently defines Harberger Taxes.

EIPS/eip-5320.md
```diff
@@ -0,0 +1,156 @@
...
+  /// @param _tokenId The identifier of the token whose valuation changed.
+  /// @param _valuation The new valuation of the token.
+  event ValuationSet(uint256 indexed _tokenId, uint256 _valuation);
+  
+  /// @dev emits when a token has tax paid for them.
+  /// @param _tokenId The identifier of the token that was taxed.
+  /// @param _value The taxes that have been paid.
+  event TaxPaid(uint256 indexed _tokenId, uint256 _value);
+
+  /// @dev buy a token. This function must succeed if
```
SamWilsn, line 67: Might want to finish this sentence?

EIPS/eip-5320.md
SamWilsn: > Presumably the remaining funds would be collected and the NFT's valuation would be set to 0 (which would allow anyone to acquire the property for free)?

Shouldn't this transfer the token to the collector, not set the valuation to zero? The collector is owed taxes, but didn't receive them.

tyntyn1:
@greenlucid @eth-bot @mdnorman @MicahZoltu @JasoonS @atlanticcrypto

tyntyn1:
@eth-bot

EIPS/eip-5320.md
tyntyn1: @greenlucid @eth-bot @JasoonS

tyntyn1: @mdnorman

tyntyn1 (commented):
hoangphuong15091996@gmail.com @eth-bot

EIPS/eip-5320.md
```diff
@@ -0,0 +1,155 @@
...
+- Governance mechanisms that use "seats" subject to Harberger.
+- Social media that requires holding a token to interact.
+
+## Motivation
+
+This proposal provides smart contract interoperability to interact with Harberger Taxes.
+
+## Specification
+
+Every contract implementing this EIP must implement the `ERC5320` and `ERC721` interfaces.
```
xinbenlv, line 38:
```suggestion
Every contract implementing this EIP must implement the [EIP-5320](./eip-5320.md) and [EIP-721](./eip-721.md) interfaces.
```

xinbenlv: EIP-5320 is this EIP itself so it's ok to omit

EIPS/eip-5320.md
```diff
@@ -0,0 +1,155 @@
...
+  /// @param _tokenId The identifier of the token to buy.
+  /// @param _offer The maximum amount offered to buy this token.
+  /// @param _valuation The valuation the token will have after a successful purchase.
+  /// @param _fund The fund put in to pay for the taxes of the token.
+  function buy(uint256 _tokenId, uint256 _offer, uint256 _valuation, uint256 _fund) external;
+
+  /// @dev call this to fund the item, to pay for the maintenance taxes.
+  /// @param _tokenId The identifier of the token to fund.
+  /// @param _value The amount sent to fund the token.
+  function fund(uint256 _tokenId, uint256 _value) external;
```
xinbenlv, line 76: Which amount of token is this `_value` here representing?

EIPS/eip-5320.md
Pandapip1: Self-references should be avoided, but since this is referencing a solidity interface defined directly in this EIP, this is fine.

PR #5329, opened 2022-07-24
Test Exploit

SamWilsn:


SamWilsn:
It worked!

image

MicahZoltu:
This doesn't appear to have exfiltrated any secrets.  Am I misunderstanding something here, were you able to extract a secret?

Aldekein:
This did a POST but didn't reveal any ENV variables as I see.

SamWilsn:
I didn't include a secret in my POST, since I was using a public service to receive the message. My test was to verify that the attacker's `Gemfile` was loaded, which presumably would've had access to all the environment variables.

MicahZoltu:
We have a `TEST` secret that you can use to find out if you can successfully extract secrets.  You are rewarded with a "good job" if you can tell me what the contents of the `TEST` secret are!

PR #5332, opened 2022-07-25
Replace kodiak with GitHub action

Pandapip1:


lightclient (approved):

MicahZoltu:
I believe we tried this previously, and the reason we switched to Kodiak was because we couldn't get this to work properly.  @alita-moore would have more details on their attempts if they are available to discuss.

MicahZoltu:
I think we probably should have merged this with `dry-run` enabled first before removing Kodiak.  Too late now though.  😬

Pandapip1:
Well it seems to work properly, so there's that.

alita-moore:
Githib auto merge occasionally failed to enable. But Kodiak offers the ability to auto merge any PR without needing to enable it. But Kodiak was also unreliable, so I enabled both githib auto merge and Kodiak to try and address edgecases.

It's a difficult problem to solve, and the solution I implemented is hacky. But it's the best I could find. So I advise not removing it until github fixes their auto merge issues.

alita-moore:
Though, if you found a better solution then 👏

But I did do a lot of searching, so it's probably worth monitoring it closely for a bit just to be sure.

PR #5335, opened 2022-07-25
Resurrect EIP-831 and add "eth" shorhand

ligi:
This EIP was stagnant but used - was contacted by @3esmit to resurrect it and add a eth shorthand as alternative to the verbose "ethereum" scheme.
Resurrecting is a no-brainer - the shortcut could use some discussion as it might be problematic e.g. when registering the scheme - so this might need a check.
Let's discuss it here when there is time: https://github.com/ethereum-cat-herders/EIPIP/issues/159

### DIFF
--- a/EIPS/eip-831.md
+++ b/EIPS/eip-831.md
@@ -8,6 +8,7 @@ status: Draft
 type: Standards Track
 category: ERC
 created: 2018-01-15
+requires: 67, 681
 ---
 
 ## Abstract
--- a/EIPS/eip-831.md
+++ b/EIPS/eip-831.md
@@ -1,16 +1,14 @@
 ---
 eip: 831
 title: URI Format for Ethereum
-author: ligi <ligi@ligi.de>
+description: A way of creating Ethereum URIs for various use-cases.
+author: ligi (@ligi)
+discussions-to: https://ethereum-magicians.org/t/eip-831-uri-format-for-ethereum/10105
+status: Draft
 type: Standards Track
 category: ERC
-status: Stagnant
 created: 2018-01-15
 ---
-
-## Simple Summary
-
-A standard way of creating Ethereum URIs for various use-cases.
 
 ## Abstract
 
@@ -20,27 +18,25 @@ URIs embedded in QR-codes, hyperlinks in web-pages, emails or chat messages prov
 
 ### Syntax
 
-Ethereum URIs contain "ethereum" in their schema (protocol) part and are constructed as follows:
+Ethereum URIs contain "ethereum" or "eth" in their schema (protocol) part and are constructed as follows:
 
-    request                 = "ethereum" ":" [ prefix "-" ] payload
+    request                 = "eth" [ "ereum" ] ":" [ prefix "-" ] payload
     prefix                  = STRING
     payload                 = STRING
 
 ### Semantics
 
-`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to ERC-67. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
+`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to [EIP-67](./eip-67.md). When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
 
-`payload` is mandatory and the content depends on the prefix. Structuring of the content is defined in the ERC for the specific use-case and not in the scope of this document. One example is ERC-681 for the pay- prefix.
-
+`payload` is mandatory and the content depends on the prefix. Structuring of the content is defined in the ERC for the specific use-case and not in the scope of this document. One example is [EIP-681](./eip-681) for the pay- prefix.
 
 ## Rationale
 
-The need for this ERC emerged when refining ERC-681. We need a container that does not carry the weight of the use-cases. ERC-67 was the first attempt on defining Ethereum-URIs. This ERC tries to keep backward compatibility and not break existing things. This means ERC-67 URIs should still be valid and readable. Only if the prefix feature is used, ERC-67 parsers might break. No way was seen to avoid this and innovate on the same time. This is also the reason this open prefix approach was chosen to being able to adopt to future use-cases and not block the whole "ethereum:" scheme for a limited set of use-cases that existed at the time of writing this.
+The need for this ERC emerged when refining EIP-681. We need a container that does not carry the weight of the use-cases. EIP-67 was the first attempt on defining Ethereum-URIs. This ERC tries to keep backward compatibility and not break existing things. This means EIP-67 URIs should still be valid and readable. Only if the prefix feature is used, EIP-67 parsers might break. No way was seen to avoid this and innovate on the same time. This is also the reason this open prefix approach was chosen to being able to adopt to future use-cases and not block the whole "ethereum:" scheme for a limited set of use-cases that existed at the time of writing this.
 
-## References
+## Security Considerations
 
-1. ERC-681, ./eip-681.md
-2. ERC-67, ./eip-67.md
+There are no known security considerations at this time.
 
 ## Copyright
 


EIPS/eip-831.md
```diff
@@ -4,8 +4,9 @@ title: URI Format for Ethereum
 author: ligi <ligi@ligi.de>
 type: Standards Track
 category: ERC
-status: Stagnant
+status: Draft
```
Pandapip1, line 7: Please fix the EIPw errors

EIPS/eip-831.md
```diff
@@ -20,27 +18,25 @@ URIs embedded in QR-codes, hyperlinks in web-pages, emails or chat messages prov
...
+    request                 = "eth" [ "ereum" ] ":" [ prefix "-" ] payload
     prefix                  = STRING
     payload                 = STRING
 
 ### Semantics
 
-`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to ERC-67. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
-
-`payload` is mandatory and the content depends on the prefix. Structuring of the content is defined in the ERC for the specific use-case and not in the scope of this document. One example is ERC-681 for the pay- prefix.
+`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to [EIP-67](https://github.com/ethereum/EIPs/issues/67). When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
```
MicahZoltu, line 29: There is no EIP-67, so you cannot link to EIP-67.  Recommend just removing this comment about compatibility.  You can't be backward incompatible with a thing that doesn't exist.  😄

ligi: I really want to leave this in as it gives important context.

Pandapip1:
```suggestion
`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility with other standards. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
```
Fixed.

EIPS/eip-831.md
```diff
@@ -20,27 +18,25 @@ URIs embedded in QR-codes, hyperlinks in web-pages, emails or chat messages prov
...
     prefix                  = STRING
     payload                 = STRING
 
 ### Semantics
 
-`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to ERC-67. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
-
-`payload` is mandatory and the content depends on the prefix. Structuring of the content is defined in the ERC for the specific use-case and not in the scope of this document. One example is ERC-681 for the pay- prefix.
+`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to a former URL format by @alexvandesande that was commonly used but never formalized and so cannot be mentioned by it's name without making the EIPW bot angry. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
```
SamWilsn, line 29: Come on, the rules aren't _that_ bad :rofl:

You could try:

```suggestion
`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility with preexisting implementations (including a draft EIP by @alexvandesande). When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
```

If you want to keep this text literally, I suggest a small grammar edit:

```suggestion
`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to a former URL format by @alexvandesande that was commonly used but never formalized and so cannot be mentioned by its name without making the EIPW bot angry. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
```

EIPS/eip-831.md
ligi: done

EIPS/eip-831.md
MicahZoltu: What other standards are being referred to by "with other standards"?  If there are no standards then this feels like a vacuous statement.

Perhaps instead, "backward compatibility with existing contracts in the ecosystem" to make it clear that there are no existing standards but there does exist contracts in the wild that there is a desire to be compatible with?

EIPS/eip-831.md
MicahZoltu: I'm not personally a fan of name dropping like that.  It doesn't add value to some future implementer who wants to implement this specification.

EIPS/eip-831.md
```diff
@@ -20,27 +18,25 @@ URIs embedded in QR-codes, hyperlinks in web-pages, emails or chat messages prov
...
     prefix                  = STRING
     payload                 = STRING
 
 ### Semantics
 
-`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to ERC-67. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
-
-`payload` is mandatory and the content depends on the prefix. Structuring of the content is defined in the ERC for the specific use-case and not in the scope of this document. One example is ERC-681 for the pay- prefix.
+`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility to a former URL format by @alexvandesande (#67) that was commonly used but never formalized and so cannot be mentioned by it's name without making the EIPW bot angry. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
```
MicahZoltu, line 29:
```suggestion
`prefix` is optional and defines the use-case for this URI. If no prefix is given: "pay-" is assumed to be concise and ensure backward compatibility with a common pattern used prior to the introduction of this standard. When the prefix is omitted, the payload must start with `0x`. Also prefixes must not start with `0x`. So starting with `0x` can be used as a clear signal that there is no prefix.
```

ligi: Thanks - but it was never used on "Ethereum mainnet" as this is used on another layer and not really on a network - and seems I can reference EIP-67 again soon as for: https://github.com/ethereum/EIPs/pull/5338
This kinda solves it.

MicahZoltu: *IF* that gets approved by the author you can reference EIP-67.  Note that this PR will be blocked from merging until then though, and be prepared for that to never happen as many EIP authors never respond.  😢

EIPS/eip-831.md
Pandapip1: > What other standards are being referred to by "with other standards"? If there are no standards then this feels like a vacuous statement.

There are other standards out there that aren't EIPs, or even official standards. "EIP-223" (not an EIP) is a good example of this.

ligi:
I removed the commits that tried to trick EIPW by not mentioning EIP-67 as @SamWilsn now submitted a PR toadd EIP-67 so we can reference it

EIPS/eip-831.md
MicahZoltu: At best, that is a defacto standard which I personally wouldn't call a standard.

ligi:
Rebased to master so it picks up EIP-67 now - so from my side it would be good to go

EIPS/eip-831.md
```diff
@@ -1,17 +1,15 @@
...
 title: URI Format for Ethereum
-author: ligi <ligi@ligi.de>
+description: A way of creating Ethereum URIs for various use-cases.
+author: ligi (@ligi)
+discussions-to: https://ethereum-magicians.org/t/eip-831-uri-format-for-ethereum/10105
+status: Draft
 type: Standards Track
 category: ERC
-status: Stagnant
 created: 2018-01-15
```
MicahZoltu, line 10:
```suggestion
created: 2018-01-15
requires: 67, 681
```
Note: At the moment `requires` means "required reading to fully understand this EIP" and it doesn't necessarily mean that someone implementing this EIP needs to also implement either of these EIPs

MicahZoltu (commented):
On the `requires` line, there is active discussion among the editors to remove this requirement, but if you want to get this merged sooner rather than later you'll need to require any referenced EIPs.

SamWilsn (approved):

SamWilsn:
@ligi can you close / re-open this to get the renamed (:rage:) `eipw` check to pass?

PR #5337, opened 2022-07-25
Ping editors with write permissions when they are needed

Pandapip1:
CCs @MicahZoltu and @lightclient when a PR is in need of manual merging.

### DIFF
--- a/.github/workflows/auto-merge-bot.yml
+++ b/.github/workflows/auto-merge-bot.yml
@@ -23,4 +23,4 @@ jobs:
           INTERFACE_EDITORS: "@lightclient,@axic,@SamWilsn,@Pandapip1"
           META_EDITORS: "@lightclient,@axic,@gcolvin,@SamWilsn,@Pandapip1"
           INFORMATIONAL_EDITORS: "@lightclient,@axic,@gcolvin,@SamWilsn,@Pandapip1"
-          MAINTAINERS: "@alita-moore,@mryalamanchi"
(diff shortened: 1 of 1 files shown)


MicahZoltu:
Those pings are meant for people who are actually working on the bot as sort of "unexpected error occurred, investigate and fix the bug".  They aren't meant for "needs manual merge".

Pandapip1:
They do happen to coincide 100% of the time, however.

MicahZoltu:
Only because the current maintainers are no longer maintaining, so bugs aren't getting fixed and we are manually overriding instead of fixing bot bugs.  😬 What *should* be happening is the maintainers get notified, they investigate while the PR is open, eventually fix the bug, and then the PR gets automerged correctly without Micah or Lightclient getting involved.

Pandapip1:
> What should be happening is the maintainers get notified, they investigate while the PR is open, eventually fix the bug, and then the PR gets automerged correctly without Micah or Lightclient getting involved.

But that isn't what is happening.

MicahZoltu:
> But that isn't what is happening.

Yeah, definitely, and we should fix that root problem rather than repurposing this mechanism for something else.  Currently, I *believe* I see/address every instance of these bugs happening so I don't think we are in dire need of a mechanism to further surface the problem.  Are you aware of a pattern of some of these slipping through the cracks and not getting caught at the moment?

Pandapip1:
No.

Issue #5082, opened 2022-05-07
One-time NFT Sale of unused EIP numbers.


MicahZoltu:
# Problem
1. We could use funding to pay for full time editors.
2. Users want to be able to choose their EIP numbers.
3. We want to maintain credibly neutrality and not give number preference to "friends of editors".

# Solution:
We do a one time auction of all unused EIP numbers between 1 and 5000 in the form of NFTs.  When a user wants to create an EIP, they can either get a number assigned to their EIP (following current or new selection strategy), or they can burn one of the NFTs just before their draft is merged and we'll let them have that number.

The proceeds would go toward funding EIP editors (specifically, funding to hire someone to replace me).  Additional things we could spend money on should the proceeds be particularly high:
* Full time staff to work on process automation.
* Fund development/enhancement/maintenance of the executable and consensus specs repositories.
* White glove service assisting people with EIP authoring.
* Authoring and maintaining of external resources related to upcoming Ethereum standards.

Pandapip1:
What chain would this be on? If it was on Ethereum, the fees would be prohibitive. If it was on e.g. Gnosis chain, then, well, it wouldn't be on Ethereum.

MicahZoltu:
It would be on Ethereum, because none of the current L2 chains are censorship resistant (they don't align with the ethos of Ethereum currently).  I don't think the fees would be too prohibitive on L1 though.  We would need to think through auction design, and potentially we could do the auction off-chain but proven on-chain with some kind of ZK prover system (high engineering cost, but very cool).

Pandapip1:
I'm very much down for designing and implementing this. I'd suggest an eventual goal as maybe having all the EIPs accessed via IPFS with the hash stored/updated on-chain.

abcoathup:
Making the EIP process more sustainable is a great aim.  

I would want to see any number with potential value up for auction rather than just limited to below 5000.  

Auto assigned numbers could be large, e.g. 6 or 7 digits, with no repeating single digit, double digit or triple digit patterns and not ending in zero.  Remove the majority of value from such auto-assigned numbers and avoid EIP editors having to make subjective choices on number usage.

MicahZoltu:
There are many numbers that are "special" like 1337, palindromes, etc.  I think figuring out what numbers people will want in advance will likely result in missing some.  Also, it makes it harder to assign new numbers because we now have to check to see if a particular number has already been auctioned or not.

[...]

PR #5340, opened 2022-07-25
Update EIP-1: Decentralize the preferred choice of `discussion-to`

xinbenlv:
Since there is no guarantee the Ethereum-Magician is going to last forever or never lock post, or delete user account, same as Github, and *requiring* Ethereum-Magician seems a centralize decision, I move to decentralize the recommendation.

FYI: Here is the [WHOIS of ethereum-magicians](https://who.is/whois/ethereum-magicians.org)
Last time it was renewed was on 2022/06/25, and currently is to expire on Feb 2023. Most of the domain information are redacted (why?)

I am curious why are we trusting ethereum-magicians more than ANY OTHER choices?

```
Registrant Contact Information:
NameREDACTED FOR PRIVACY
Organization
AddressREDACTED FOR PRIVACY
CityREDACTED FOR PRIVACY
State / ProvinceParis
Postal CodeREDACTED FOR PRIVACY
CountryFR
PhoneREDACTED FOR PRIVACY
FaxREDACTED FOR PRIVACY
Email

Administrative Contact Information:
NameREDACTED FOR PRIVACY
Organization
AddressREDACTED FOR PRIVACY
CityREDACTED FOR PRIVACY
State / ProvinceParis
Postal CodeREDACTED FOR PRIVACY
CountryFR
PhoneREDACTED FOR PRIVACY
FaxREDACTED FOR PRIVACY
Email

Technical Contact Information:
NameREDACTED FOR PRIVACY
Organization
AddressREDACTED FOR PRIVACY
CityREDACTED FOR PRIVACY
State / ProvinceParis
Postal CodeREDACTED FOR PRIVACY
CountryFR
PhoneREDACTED FOR PRIVACY
FaxREDACTED FOR PRIVACY
Email

Information Updated: 2022-07-25 21:03:03
```

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -167,7 +167,7 @@ At least one author must use a GitHub username, in order to get notified on chan
 
 While an EIP is a draft, a `discussions-to` header will indicate the URL where the EIP is being discussed.
 
-A preferred discussion URL is a that URL shall not point to any URL which is ephemeral, and any URL which can get locked over time. The following is a list of recommended choices: 
+A preferred discussion URL is one that is NOT ephemeral, or could get locked in a forseable future. The following is a list of recommended choices: 
 - A post on [Ethereum Magicians](https://ethereum-magicians.org/) that does not intend to be locked.
 - An issue or pull request on [Ethereum/EIPs Github repository](https://github.com/ethereum/EIPs).
 - A post on [ETH Research](https://ethresear.ch)
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -167,7 +167,11 @@ At least one author must use a GitHub username, in order to get notified on chan
 
 While an EIP is a draft, a `discussions-to` header will indicate the URL where the EIP is being discussed.
 
-The preferred discussion URL is a topic on [Ethereum Magicians](https://ethereum-magicians.org/). The URL cannot point to Github pull requests, any URL which is ephemeral, and any URL which can get locked over time (i.e. Reddit topics).
+A preferred discussion URL is a that URL shall not point to any URL which is ephemeral, and any URL which can get locked over time. The following is a list of recommended choices: 
+- A post on [Ethereum Magicians](https://ethereum-magicians.org/) that does not intend to be locked.
+- An issue or pull request on [Ethereum/EIPs Github repository](https://github.com/ethereum/EIPs).
+- A post on [ETH Research](https://ethresear.ch)
+
 
 #### `type` header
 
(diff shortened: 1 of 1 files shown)


MicahZoltu:
In this case, we actually have a *desire* to have the discussions centralized as it makes it so people can reliably participate in all discussions without needing to have accounts in multiple locations.  We specifically don't want to use GitHub issues/pull requests because we want to utilize them for other purposes, and we cannot back them up and we don't control hosting of them.

Ethereum Magicians can be (and I think is) backed up, and I believe it is self hosted (or at least it can be if needed).  If GitHub decides to censor Ethereum, due to the nature of `git` it isn't that big of a deal for the repository contents itself, but all of the value-adds like issues and pull requests would be lost.

lightclient:
People involved with the EIP process also help maintain the FEM forum, so I don't consider it likely for FEM to disappear. Keeping everything uniform here IMO is a bigger win than allowing other mediums of discussion threads, so I'm going to close this for now.

xinbenlv:
@lightclient thank you for reply.

Could we keep thid issue open for more discussion so other people can respond their opinions?

Pandapip1:
I'd be okay with leaving this open and adding it to the EIPIP adgenda. CC @poojaranjan

lightclient:
IMO we have already spent a lot of time debating on the requirement that all new EIPs will use FEM for `discussions-to`. I don't think we should rehash things over and over again, so I would rather just let this rest and accept the FEM requirement for the near term (the next 1-2 years) and then reevaluate. Otherwise we're going to get whiplash from flip flopping.

MicahZoltu:
While I share @lightclient's position on this subject, I'm fine with leaving this open for a bit longer to incentivize additional feedback.  My *guess* though is that since most of the editors I believe are more or less aligned on this subject someone would need to put forward a pretty strong argument to change the policy at this point.  If we don't make significant progress on this in the next week or two though, I think we should probably close this issue.

Greg put forth what I thought was a pretty good argument back when we previously discussed switching to FEM (the fact that you can't edit original post), but we were able to resolve that by changing the editing rules on the forum so people now could edit their original post.

gcolvin:
Can we help make sure that the Magician's forum is at least as financially stable as our repo?

MicahZoltu:
We could achieve that with something like https://github.com/ethereum/EIPs/issues/5082  😆

gcolvin:
> We could achieve that with something like #5082 😆

To which I have strong objections.

Pandapip1:
We could use ENS, or my alternative, depending on what we want out of it.

PR #5342, opened 2022-07-26
Update Stale Bot

Pandapip1:
- Massively reduces the time required for an issue or PR to become stale or be closed because of it
- Reduces the verbosity of the message
- Reduces the frequency that the bot is run at
- Adds a "long-term" label that can be applied to PRs or issues that are awaiting e.g. author approval (will need to manually be created).
- Other stuff

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,13 +20,15 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         ascending: true # Since we have so many issues, the stale bot finds it hard to keep track. This makes sure that at least the oldest are removed.
         # Issue config
-        stale-issue-message: There has been no activity on this issue in a month. It will be closed in a week if no further activity occurs
+        stale-issue-message: There has been no activity on this issue in a month. It will be closed in a week if no further activity occurs.
+        close-issue-message: This issue was closed due to inactivity. If you are still pursuing it, feel free to reopen it.
         days-before-issue-stale: 30
         days-before-issue-close: 7
         exempt-issue-labels: discussions-to,long-term
         stale-issue-label: stale
         # PR config
         stale-pr-message: There has been no activity on this pull request in a month. It will be closed in a week if no further activity occurs.
+        close-pr-message: This pull request was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.
         days-before-pr-stale: 30
         days-before-pr-close: 7
         exempt-pr-labels: long-term
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Mark stale PRs
     steps:
-    - uses: actions/stale@v3
+    - uses: actions/stale@532554b8a8498a0e006fbcde824b048728c4178f
       with:
         # General
         repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,35 +1,33 @@
-name: "Mark stale PRs & Issues"
+name: Mark stale PRs & Issues
 
 on:
   schedule:
-  # Run this every hour, so we are not spammed with changes at once. Later we could consider changing this to once a day.
-  - cron: "0 * * * *"
+  - cron: 0 0 * * *  # Runs at 00:00 UTC every day.
+
+permissions:
+  issues: write
+  pull-requests: write
 
 jobs:
-  stale-pr:
+  stale:
     if: github.repository == 'ethereum/eips'
     runs-on: ubuntu-latest
     name: "Mark stale PRs"
     steps:
     - uses: actions/stale@v3
       with:
+        # General
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: 60
-        days-before-close: 7
-        stale-pr-label: 'stale'
-        stale-pr-message: 'There has been no activity on this pull request for two months. It will be closed in a week if no further activity occurs. If you would like to move this EIP forward, please respond to any outstanding feedback or add a comment indicating that you have addressed all required feedback and are ready for a review.'
-        close-pr-message: 'This pull request was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.'
-  stale-issue:
-    if: github.repository == 'ethereum/eips'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -1,35 +1,33 @@
...
-        stale-pr-message: 'There has been no activity on this pull request for two months. It will be closed in a week if no further activity occurs. If you would like to move this EIP forward, please respond to any outstanding feedback or add a comment indicating that you have addressed all required feedback and are ready for a review.'
-        close-pr-message: 'This pull request was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.'
-  stale-issue:
+  stale:
     if: github.repository == 'ethereum/eips'
     runs-on: ubuntu-latest
-    name: "Mark stale issues"
+    name: Mark stale PRs
     steps:
     - uses: actions/stale@v3
```
MicahZoltu, line 17: This should be using a hash rather than a tag.

.github/workflows/stale.yml
```diff
@@ -1,35 +1,33 @@
...
+        # Issue config
+        stale-issue-message: There has been no activity on this issue in a month. It will be closed in a week if no further activity occurs
+        days-before-issue-stale: 30
+        days-before-issue-close: 7
+        exempt-issue-labels: discussions-to,long-term
+        stale-issue-label: stale
+        # PR config
+        stale-pr-message: There has been no activity on this pull request in a month. It will be closed in a week if no further activity occurs.
+        days-before-pr-stale: 30
+        days-before-pr-close: 7
```
MicahZoltu, line 31: I don't think we should be changing this to 1 month from 2.

.github/workflows/stale.yml
```diff
@@ -1,35 +1,33 @@
...
-        days-before-close: 14
-        exempt-issue-labels: 'discussions-to'
-        stale-issue-label: 'stale'
-        stale-issue-message: 'There has been no activity on this issue for two months. It will be closed in a week if no further activity occurs. If you would like to move this EIP forward, please respond to any outstanding feedback or add a comment indicating that you have addressed all required feedback and are ready for a review.'
-        close-issue-message: 'This issue was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.'
+        ascending: true # Since we have so many issues, the stale bot finds it hard to keep track. This makes sure that at least the oldest are removed.
+        # Issue config
+        stale-issue-message: There has been no activity on this issue in a month. It will be closed in a week if no further activity occurs
+        days-before-issue-stale: 30
+        days-before-issue-close: 7
```
MicahZoltu, line 25: I don't think we should be changing this from 3 months to 1.

.github/workflows/stale.yml
```diff
@@ -1,35 +1,33 @@
...
+        exempt-issue-labels: discussions-to,long-term
+        stale-issue-label: stale
+        # PR config
+        stale-pr-message: There has been no activity on this pull request in a month. It will be closed in a week if no further activity occurs.
```
MicahZoltu, line 29: Why did you remove the helper text that gives advice to the author on how to proceed?

.github/workflows/stale.yml
```diff
@@ -1,35 +1,33 @@
...
+        days-before-pr-stale: 30
+        days-before-pr-close: 7
+        exempt-pr-labels: long-term
```
MicahZoltu, line 32: Would be nice if we could get the bot to add a label if a PR is an EIP vs if it is touching something like workflow files, EIP-1, etc.  That way we could have this only apply to EIPs.

.github/workflows/stale.yml
```diff
@@ -1,35 +1,33 @@
...
-    name: "Mark stale PRs"
-    steps:
-    - uses: actions/stale@v3
-      with:
-        repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: 60
-        days-before-close: 7
-        stale-pr-label: 'stale'
-        stale-pr-message: 'There has been no activity on this pull request for two months. It will be closed in a week if no further activity occurs. If you would like to move this EIP forward, please respond to any outstanding feedback or add a comment indicating that you have addressed all required feedback and are ready for a review.'
-        close-pr-message: 'This pull request was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.'
```
MicahZoltu, line 21: Why did you remove the close message?

.github/workflows/stale.yml
Pandapip1: We can. I'll do that now.

.github/workflows/stale.yml
Pandapip1: Re-added.

Pandapip1:
So my rationale for lowering the time is that, as mentioned, authors may reopen their PRs and issues. It isn't too much of a hassle to click that button, even if it is a slight inconvenience.

What *is* a big inconvenience is having 72 open PRs where about a quarter of them are unlikely to be touched by their authors unless prompted and having to determine whether or not they need reviewing. If a PR hasn't been touched in *over a month*, it might be because the author is no longer interested.

Pandapip1:
I just had a wacky thought: what if we used the [waiting: response from author](https://github.com/ethereum/EIPs/issues?q=is%3Apr+is%3Aopen+label%3A%22waiting%3A+response+from+author%22) label as the stale label for PRs, set the period before it becomes "stale" to one week, and the period before it gets *closed* as 1 month?

Pandapip1:
Note that the stale bot counts commits *and* comments. If the author hasn't added any new commits *or* commented that they will be addressing editor feedback for a week I would argue it's probably fair to give a 30-day warning :)

.github/workflows/stale.yml
Pandapip1: Fixed. I changed it to 7 days, instead.

MicahZoltu:
My recommendation is to split this into multiple PRs.  One PR to just cleanup the workflow file, and then subsequent PRs to propose each of your individual suggestions for how we should alter the bot's behavior.  At the moment, I think this PR is touching too many things at once which makes it hard to rally around/discuss.  I think the cleanup PR (assuming it changes no behavior, just cleans stuff up) can be merged almost immediately while the others will each require separate lines of discussion among editors.

MicahZoltu:
I'm not sure if it was intentional or not, but this got merged without approval from another editor and it appears it still had a bunch of functional changes (not just cleanup).  We should probably revert it and try again.

MicahZoltu:
Reverted, but I would like to see this pursued as I think several of the changes are good ideas (at the least, the cleanup is good).

PR #5349, opened 2022-07-27
EIP-4626

vdusart:
Fix a typo by changing the name of the variable in the explanation of the methods `deposit` and `mint`


vdusart:
IDK but It might be better to keep amount and replace asset.

fubuloubu (approved):

Pandapip1 (approved):
I think that this modification is fine. Someone will need to manually merge this to fix the eipw errors though.

PR #5353, opened 2022-07-27
Update EIP-4844: Remove system contract concept

lightclient:
This PR removes the "system contract" concept from EIP-4844. Although the approach has several pros, it tends to be a sticking point in the discussions around EIP-4844 due to the fact that we don't currently have the concept of a system contract with storage and that concept would need to be implemented into clients - whereas the header extension approach has already been performed to add `baseFee` to the header in EIP-1559.

Additionally, it resolves an issue in the original draft that conflated blobs and blob txs.

Will leave in draft until some consensus among authors is reached.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -235,7 +235,7 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
 ### Header extension
 
 The current header encoding is extended with a new 256-bit unsigned integer field `blob_txs`. This is the running
-total of blob txs included on chain since this EIP was activated. The resulting RLP encoding of the header is therefore:
+total of excess blob txs included on chain since this EIP was activated. The resulting RLP encoding of the header is therefore:
 
 ```
 rlp([
@@ -262,12 +262,9 @@ rlp([
 The value of `blob_txs` can be calculated using the parent header and number of blob transactions in the block.
 
 ```python
-def calc_blob_gas(parent: Header, blob_txs_in_block: int) -> int:
-    # Don't let the new total fall behind the targeted total to avoid
-    # accumulating a long period of very low fees
-    blocks_since_start = header.Number - FORK_BLKNUM
-    targeted_total = blocks_since_start * TARGET_BLOB_TXS_PER_BLOCK
-    return max(header.blob_txs + blob_txs_in_block, targeted_total)
+def calc_excess_blob_txs(parent: Header, blobs: int) -> int:
+    adjusted = parent.blob_txs + blobs
+    return adjusted - min(TARGET_BLOB_TXS_PER_BLOCK, adjusted)
 ```
 
 ### Beacon chain validation
@@ -346,7 +343,7 @@ def get_intrinsic_gas(tx: SignedBlobTransaction, pre_state: ExecState) -> int:
 def get_blob_gas(header: Header) -> int:
     blocks_since_start = get_current_block(pre_state) - FORK_BLKNUM
     targeted_total = blocks_since_start * TARGET_BLOB_TXS_PER_BLOCK
-    actual_total = header.blob_txs
+    actual_total = targeted_total + header.blob_txs
     if actual_total < targeted_total:
         return 0
     else:
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -49,7 +49,7 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `POINT_EVALUATION_PRECOMPILE_ADDRESS` | `Bytes20(0x14)` |
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
 | `MAX_BLOBS_PER_BLOCK` | `16` |
-| `TARGET_BLOB_TXS_PER_BLOCK` | `8` |
+| `TARGET_BLOBS_PER_BLOCK` | `8` |
 | `MAX_BLOBS_PER_TX` | `2` |
 | `GASPRICE_UPDATE_FRACTION_PER_BLOB` | `64` |
 | `MAX_VERSIONED_HASHES_LIST_SIZE` | `2**24` |
@@ -264,7 +264,7 @@ The value of `blob_txs` can be calculated using the parent header and number of 
 ```python
 def calc_excess_blob_txs(parent: Header, blobs: int) -> int:
     adjusted = parent.blob_txs + blobs
-    return adjusted - min(TARGET_BLOB_TXS_PER_BLOCK, adjusted)
+    return adjusted - min(TARGET_BLOBS_PER_BLOCK, adjusted)
 ```
 
 ### Beacon chain validation
@@ -342,7 +342,7 @@ def get_intrinsic_gas(tx: SignedBlobTransaction, pre_state: ExecState) -> int:
 
 def get_blob_gas(header: Header) -> int:
     blocks_since_start = get_current_block(pre_state) - FORK_BLKNUM
-    targeted_total = blocks_since_start * TARGET_BLOB_TXS_PER_BLOCK
+    targeted_total = blocks_since_start * TARGET_BLOBS_PER_BLOCK
     actual_total = targeted_total + header.blob_txs
     if actual_total < targeted_total:
         return 0
@@ -500,12 +500,12 @@ sequencers would simply have to switch over to using a new transaction type at t
 ### Blob gasprice update rule
 
 The blob gasprice update rule is intended to approximate the formula `blob_gas = 2**(excess_blobs / GASPRICE_UPDATE_FRACTION_PER_BLOB)`,
-where `excess_blobs` is the total "extra" number of blobs that the chain has accepted relative to the "targeted" number (`TARGET_BLOB_TXS_PER_BLOCK` per block).
+where `excess_blobs` is the total "extra" number of blobs that the chain has accepted relative to the "targeted" number (`TARGET_BLOBS_PER_BLOCK` per block).
 Like EIP-1559, it's a self-correcting formula: as the excess goes higher, the `blob_gas` increases exponentially, reducing usage and eventually forcing the excess back down.
 
 The block-by-block behavior is roughly as follows.
-If in block `N`, `blob_gas = G1`, and block `N` has `X` blobs, then in block `N+1`, `excess_blobs` increases by `X - TARGET_BLOB_TXS_PER_BLOCK`,
-and so the `blob_gas` of block `N+1` increases by a factor of `2**((X - TARGET_BLOB_TXS_PER_BLOCK) / GASPRICE_UPDATE_FRACTION_PER_BLOB)`.
+If in block `N`, `blob_gas = G1`, and block `N` has `X` blobs, then in block `N+1`, `excess_blobs` increases by `X - TARGET_BLOBS_PER_BLOCK`,
+and so the `blob_gas` of block `N+1` increases by a factor of `2**((X - TARGET_BLOBS_PER_BLOCK) / GASPRICE_UPDATE_FRACTION_PER_BLOB)`.
 Hence, it has a similar effect to the existing EIP-1559, but is more "stable" in the sense that it responds in the same way to the same total usage regardless of how it's distributed.
 
 ## Backwards Compatibility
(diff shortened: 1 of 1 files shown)


protolambda (changes requested):
I was expecting that we'd add a `blob_basefee` field, not a total `blob_txs` counter. Since the counter does not capture the fee-adjustment result of previous history, only the total number of blobs.

In EIP-1559 we weigh recent history higher than older history, but now it's basically averaging the usage all-time, which is not the same as the exponential adjustment we have in EIP-1559; adjustments will be smaller and smaller the longer the fork is active.

And the `ExecutionPayloadHeader` and `ExecutionPayload` type in the CL spec will need to match the EL definition of the block.

protolambda:
After re-reading the EIP, and the `Blob gasprice update rule` text, I do think your implementation is correct, but I think the thing in question for me is this part of the EIP:

> Hence, it has a similar effect to the existing EIP-1559, but is more "stable" in the sense that it responds in the same way to the same total usage regardless of how it's distributed.

EIP-1559 responds based on the distribution of blobs throughout history, whereas the current EIP draft handles excess blobs over all time, making excess blobs have long-term effects on the price.

With these effects, we get two unfortunate side-effects:
- Excess blobs (i.e. over target) that were included long ago, and already pruned, will cause the blob price to be higher even if all blocks within the non-pruned period perfectly match the target.
- Few bursts vs many small excesses: stable bandwidth usage and processing cost may be preferable, yet the total cost is lower if you distribute them in bursts.
  - Example:
    - For 8 blocks, if you have 1 excess each time:  `2**(1/64)+2**(2/64)+2**(3/64)+2**(4/64)+2**(5/64)+2**(6/64)+2**(7/64)+2**(8/64) = ~ 8.40` total cost (note how the excess keeps growing)
    - For 8 blocks, if you have no excess, and then 8 all at once: `2**(8/64) = ~ 1.09` (or `~ 8.09` if the no-excess blocks charge `1` each, unlike the EIP code which makes those free...)
  - Maybe we embrace it, since batching is good for CPU performance? Would unstable bandwidth hurt the network more than less efficient batch processing?

Also, the EIP describes "`blob_gas` of block `N+1` increases by a factor of `2**((X - TARGET_BLOB_TXS_PER_BLOCK) / GASPRICE_UPDATE_FRACTION_PER_BLOB)`." but there's no increase, the code uses this directly as gas price per blob (and it can even be zero).

EIPS/eip-4844.md
```diff
@@ -234,6 +232,44 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
...
+    nonce,
+    base_fee,
+    blob_txs
+])
+```
+
+The value of `blob_txs` can be calculated using the parent header and number of blob transactions in the block.
+
+```python
+def calc_blob_gas(parent: Header, blob_txs_in_block: int) -> int:
```
vbuterin, line 265: Rename to `calc_blob_txs`?

vbuterin: On second thoughts, a potential even better idea: instead of storing a "total" of `blob_txs`, which is not actually even a total because of how it gets topped up if it gets too low, why not store `excess_blob_txs`?

```python
def calc_excess_blob_txs(parent: Header, blob_txs_in_block: int) -> int:
    output = parent.blob_txs + blob_txs_in_block
    output = output - min(TARGET_BLOB_TXS_PER_BLOCK, output)
```

vbuterin:
> With these effects, we get two unfortunate side-effects:

I think that if we had gone with an EIP 1559 based on excess (so what this EIP does) from the start, no one would seriously consider adding a weird rule where you penalize excess more if it's spread out across blocks than if it's concentrated in a single block (which is what status quo EIP 1559 gives us in practice). The complexity is just not worth it, and the value of having a clean simple invariant of what total usage is targeted is really high.

More concretely, the value that the clean invariant of the excess-based rule gives us includes:

* We would not have to worry about bad proposers manipulating the basefee by strategically delaying transactions, because we know that strategically delaying transactions never affects the basefee
* A cleaner upper bound on how much storage clients need for blobs (`target_per_block * (storage_duration / slot_duration + GASPRICE_UPDATE_FRACTION_PER_BLOB * 50)`, instead of needing 3% extra just in case you get a worst-case distribution
* A 51% attack on the basefee mechanism would actually require the attacker to have 50.01% of proposer slots, instead of ~46.9% with an EIP-1559-style rule

protolambda:
In favor of less complexity, let's go ahead with this PR. Pruning/batching incentives are interesting to me, but don't have to block this PR, and may just make it worse.

The change from `total_blob_txs` to `excess_blob_txs` like Vitalik proposed would be nice though; it's less complex and we may not have to compute what the target `total_blob_txs` would be (which gets weird once you change the allowed throughput with a hardfork).

EIPS/eip-4844.md
lightclient: I've updated the PR to use this technique, PTAL!

EIPS/eip-4844.md
```diff
@@ -307,14 +340,10 @@ def get_intrinsic_gas(tx: SignedBlobTransaction, pre_state: ExecState) -> int:
...
-def get_blob_gas(pre_state: ExecState) -> int:
+def get_blob_gas(header: Header) -> int:
     blocks_since_start = get_current_block(pre_state) - FORK_BLKNUM
     targeted_total = blocks_since_start * TARGET_BLOB_TXS_PER_BLOCK
-    actual_total = read_storage(
-        pre_state,
-        SYSTEM_STATE_ADDRESS,
-        TOTAL_BLOB_TXS_STORAGE_SLOT
-    )
+    actual_total = targeted_total + header.blob_txs
```
lightclient, line 346: Please note that I've updated the gas calculation to sum the running excess of blob txs and the targeted total value.

EIPS/eip-4844.md
```diff
@@ -234,6 +232,41 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
...
+    blob_txs
+])
+```
+
+The value of `blob_txs` can be calculated using the parent header and number of blob transactions in the block.
+
+```python
+def calc_excess_blob_txs(parent: Header, blobs: int) -> int:
+    adjusted = parent.blob_txs + blobs
+    return adjusted - min(TARGET_BLOB_TXS_PER_BLOCK, adjusted)
```
0xGabi, line 267: Is `TARGET_BLOB_TXS_PER_BLOCK` the same as `TARGET_BLOBS_PER_BLOCK` constant? If that's the case, can we define only one?

0xGabi (commented):
After reading the EIP and this Draft I notice the above.

lightclient:
@0xGabi thanks for the review. I was previously considering renaming the constant but decided against it in this PR. Looks like I didn't revert the change fully! Should be okay now though.

EIPS/eip-4844.md
```diff
@@ -234,6 +232,41 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
...
+    blob_txs
+])
+```
+
+The value of `blob_txs` can be calculated using the parent header and number of blob transactions in the block.
+
+```python
+def calc_excess_blob_txs(parent: Header, blobs: int) -> int:
+    adjusted = parent.blob_txs + blobs
+    return adjusted - min(TARGET_BLOBS_PER_BLOCK, adjusted)
```
adietrichs, line 267: This changes the mechanism, because it only stores the diff if positive. To keep the mechanism intact, `blob_txs` would have to be a signed integer. If this is impractical, I would recommend sticking to the original `actual_total` accounting variant.

lightclient: So my understanding of the current mechanism is that the `max` function clips any negative deltas vs. the target amount and thus causes the same behavior:

```python
def update_blob_gas(state: ExecState, blob_txs_in_block: int):
    current_total = read_storage(
        state,
        SYSTEM_STATE_ADDRESS,
        TOTAL_BLOB_TXS_STORAGE_SLOT
    )
    # Don't let the new total fall behind the targeted total to avoid
    # accumulating a long period of very low fees
    blocks_since_start = get_current_block(pre_state) - FORK_BLKNUM
    targeted_total = blocks_since_start * TARGET_BLOB_TXS_PER_BLOCK
    new_total = max(current_total + blob_txs_in_block, targeted_total)
    write_storage(
        state,
        SYSTEM_STATE_ADDRESS,
        TOTAL_BLOB_TXS_STORAGE_SLOT,
        new_total
    )
```

ndengeyingoma:
hey i need blockchain project for using in my defense and graduation
who are able to give to me???

EIPS/eip-4844.md
adietrichs: Ah, you are correct, I overlooked that in the original EIP. Those two mechanisms are equivalent after all then.

Inphi (commented):
LGTM from an inplementer's POV.

EIPS/eip-4844.md
```diff
@@ -234,6 +232,44 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
...
+    blobs
+])
+```
+
+The value of `blobs` can be calculated using the parent header and number of blobs in the block.
+
+```python
+def calc_excess_blobs(parent: Header, blobs: int) -> int:
+    adjusted = parent.blobs + blobs
+    return adjusted - min(TARGET_BLOBS_PER_BLOCK, adjusted)
```
adietrichs, line 270: suggested more expressive naming:
* `new_blobs` instead of `blobs`
* `new_total` instead of `adjusted`

EIPS/eip-4844.md
```diff
@@ -307,45 +343,16 @@ def get_intrinsic_gas(tx: SignedBlobTransaction, pre_state: ExecState) -> int:
...
-    targeted_total = blocks_since_start * TARGET_BLOB_TXS_PER_BLOCK
-    actual_total = read_storage(
-        pre_state,
-        SYSTEM_STATE_ADDRESS,
-        TOTAL_BLOB_TXS_STORAGE_SLOT
-    )
-    if actual_total < targeted_total:
+def get_blob_gas(header: Header) -> int:
+    if header.blobs == 0:
         return 0
```
adietrichs, line 348: I don't think we need to introduce this special case, in the old version 0 was just passed into the `fake_exponential` function normally.

EIPS/eip-4844.md
```diff
@@ -307,45 +343,16 @@ def get_intrinsic_gas(tx: SignedBlobTransaction, pre_state: ExecState) -> int:
...
     else:
         return fake_exponential(
-            actual_total - targeted_total,
+            header.blobs,
             GASPRICE_UPDATE_FRACTION_PER_BLOB
         )
 ```
 
-We update at the end of a block, as follows:
```
adietrichs, line 327: We might need something equivalent for setting the new excess blobs header field, using `calc_excess_blobs`. I.e. specify how to calculate the total number of blobs in the block by iterating through all blob txs and summing up their blobs counts.

adietrichs (commented):
Some more small recommended changes, then this should be good to merge.

EIPS/eip-4844.md
lightclient: I think `new_total` is actually incorrect though, because it is the `excess_blobs` + `new_blobs` which isn't always the total number of blobs.

EIPS/eip-4844.md
lightclient: I think this is covered by the link "The value of `excess_blobs` can be calculated ... ".

adietrichs: I think we need to specify where the input value `new_blobs` to the `calc_excess_blobs` function comes from though.

lightclient: "... using the parent header and number of blobs in the block."

lightclient: ;)

adietrichs (approved):
lgtm

EIPS/eip-4844.md
```diff
@@ -290,8 +328,8 @@ For early draft implementations, we simply change `get_blob_gas(pre_state)` to a
 ### Gas price update rule (Full version)
 
 We propose a simple independent EIP-1559-style targeting rule to compute the gas cost of the transaction.
-We use the `TOTAL_BLOB_TXS_STORAGE_SLOT` storage slot of the `SYSTEM_STATE_ADDRESS` address
-to store persistent data needed to compute the cost.
+We use the `blobs` header field to store persistent data needed to compute the cost.
```
axic, line 331: Is this `blobs` here `excess_blobs`?

axic: cc @adietrichs @lightclient

lightclient: fixed in #5629, thanks for catching

Issue #5158, opened 2022-06-14
Example relative link in EIP-1 is rendered as absolute


SamWilsn:
In the [markdown source](https://github.com/ethereum/EIPs/blob/78bc7a8ecc0bea3c73e21a43258e4de5748afbfc/EIPS/eip-1.md?plain=1#L194), this link is relative, but in the rendered version of [EIP-1](https://eips.ethereum.org/EIPS/eip-1), it is absolute:

![image](https://user-images.githubusercontent.com/57262657/173617031-7ce3d9a5-2d01-4ebf-b738-1917c9c61a97.png)

MicahZoltu:
WTF Jekyll markdown renderer....

Pandapip1:
This is still an issue

SamWilsn:
Still broken.

PR #5361, opened 2022-07-28
CI: Add Link Checker

Pandapip1:
This adds another action that checks links much quicker than htmlproofer does, and works with GitHub's automatic file annotations (which htmlproofer does not and can not).

Why the redundant check? Faster feedback to authors writing in real-time, and the ability to actually see the correct file and line numbers instead of having to guess. Fixing htmlproofer errors is a pain.

This does *not* replace htmlproofer.


Pandapip1:
Well, that fixed it. I've submitted a PR to their repo: https://github.com/gaurav-nelson/github-action-markdown-link-check/pull/148

.github/workflows/ci.yml
```diff
@@ -1,37 +1,23 @@
...
-        run: |
-          bundle exec jekyll doctor
-          bundle exec jekyll build
-      
-      - name: HTML Proofer
-        run: bundle exec htmlproofer ./_site --check-html --check-opengraph --report-missing-names --log-level=:debug --assume-extension --empty-alt-ignore --timeframe=6w --disable-external
-      
-      - name: DNS Validator
-        run: bundle exec github-pages health-check
+      - name: Link Checker
```
SamWilsn, line 15: I guess one benefit to checking after the rendering instead of before is that any weirdness with Jekyll will be caught (see #5158 for an example of weirdness.) I'm not opposed to switching though, just wanted to raise the concern.

You've checked this against the repo already, I assume?

Pandapip1: > You've checked this against the repo already, I assume?

Yes. I even had to submit a PR to fix a small bug lol.

Pandapip1: > any weirdness with Jekyll will be caught

#5158 didn't get caught by HTMLProofer, though!

SamWilsn: It wasn't a link!

.github/workflows/ci.yml
```diff
@@ -1,37 +1,23 @@
...
-      - name: Build Website
-        run: |
-          bundle exec jekyll doctor
-          bundle exec jekyll build
-      
-      - name: HTML Proofer
-        run: bundle exec htmlproofer ./_site --check-html --check-opengraph --report-missing-names --log-level=:debug --assume-extension --empty-alt-ignore --timeframe=6w --disable-external
-      
-      - name: DNS Validator
-        run: bundle exec github-pages health-check
```
MicahZoltu, line 34: What did this health-check do previously, and does the new tool do it or do you believe this is no longer necessary?

.github/workflows/ci.yml
```diff
@@ -1,37 +1,23 @@
...
-      - name: Install Ruby
-        uses: ruby/setup-ruby@08245253a76fa4d1e459b7809579c62bd9eb718a
-        with:
-          ruby-version: '2.6.0'
-          bundler-cache: true
         
-      - name: Build Website
-        run: |
-          bundle exec jekyll doctor
-          bundle exec jekyll build
```
MicahZoltu, line 28: It feels like we should still be building the website *somewhere* in our CI, as we don't want to merge a change that breaks eips.ethereum.org.

.github/workflows/ci.yml
Pandapip1: It checked that the CNAME `eips.ethereum.org` actually pointed to `ethereum.github.io`. I assume the Ethereum foundation isn't going to rug us (and if they did, the last thing we would want is our CI failing because of it).

.github/workflows/ci.yml
```diff
@@ -1,37 +1,23 @@
...
-      
-      - name: HTML Proofer
-        run: bundle exec htmlproofer ./_site --check-html --check-opengraph --report-missing-names --log-level=:debug --assume-extension --empty-alt-ignore --timeframe=6w --disable-external
-      
-      - name: DNS Validator
-        run: bundle exec github-pages health-check
+      - name: Link Checker
+        uses: Pandapip1/github-action-markdown-link-check@b36c1679e855535e05e56038813289331f095f3f
```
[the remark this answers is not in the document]
Pandapip1: > Can you help me understand what changing git's safe.directory has to do with this?

I have no idea, but the actual action straight-up doesn't work with this repo without it.

.github/workflows/ci.yml
Pandapip1: > We don't want to merge a change that breaks eips.ethereum.org

If a change does get made that breaks `eips.ethereum.org`, then it won't deploy.

.github/workflows/ci.yml
MicahZoltu: I would worry less about being rugged, and more about someone accidentally breaking something when updating DNS records.

Do we gain some tangible benefit from removing this that makes it worth losing out on robustness?

.github/workflows/ci.yml
MicahZoltu: For PR merging, I think 2 minutes of CI is reasonable.  Shorter is certainly better, but I am pretty apprehensive about removing the build step entirely from CI.

Is there anything we can do to improve the performance of the Jekyll build?  Is there perhaps some caching we could benefit from, or settings we can tweak to improve that?

MicahZoltu: If build performance is the concern, we should probably run a Jekyll `--profile` to see where time is being spent.  Maybe it is something simple/small we can fix.

Pandapip1: HTMLproofer takes about as long as jekyll.

Pandapip1: > Is there perhaps some caching we could benefit from

I looked at that earlier, but gave up. I could give it a go again, if you want.

.github/workflows/ci.yml
Pandapip1: If anything, the above should run as a periodic check on the repo, instead of on PR.

MicahZoltu: I agree that a periodic check would be better.  At the moment I don't believe we have the infrastructure to do such things though, and I would want to see a replacement put into place that gives us confidence that should something break the appropriate parties will be notified before removing this.  Maybe @jpitts knows if someone already has this sort of monitoring in place though and if so then we can remove this?

.github/workflows/ci.yml
MicahZoltu: While I would appreciate faster CI as much as the next guy, I don't have a strong desire/need for it to warrant thinking anyone should sink a significant amount of time into it.  Another option would be to switch away from Jekyll to something else (with perhaps faster build times).

@lightclient tried to do this a while back, moving to some Rust based system, but I think it didn't support yaml frontmatter which would have required pretty significant changes to every EIP for it to run successfully so we dropped it.  Maybe that tool has added YAML frontmatter support and we could re-evaluate that change?  Or maybe there is another Jekyll-compatible renderer that wouldn't be too hard to hot-swap in?

Pandapip1:
@MicahZoltu how about this CI is added alongside HTMLProofer, but have HTMLProofer not required and this one required? That way if there is an issue with Jekyll we can just not approve it, but for irrelevant PRs we don't have to worry?

.github/workflows/ci.yml
Pandapip1: > At the moment I don't believe we have the infrastructure to do such things though,

We do. I can write it if you want!

SamWilsn:
I have to side with Micah on this one. I think building with Jekyll as part of CI is reasonable, and if we're doing that, it makes more sense to check the links in the rendered HTML. Seems like doing it that way has more potential to catch errors (like misconfigured a renderer.)

I wouldn't be opposed to a faster markdown tool and/or faster HTML link checker.

SamWilsn:
I haven't investigated at all, but perhaps upgrading to Ruby 3 and adding the `--jit` or `--yjit` options would help?

Pandapip1:
I think that having this (even as a non-required check) is still useful because of the ability for it to actually comment on the files view.

lightclient:
I don't think we should have duplicate services in CI checking links.

lightclient:
Going to close this as it appears the majority is against this change.

Pandapip1:
Reopening, since the decision that was made was on a previous revision of this PR that replaced the jekyll action.

.github/workflows/ci.yml
```diff
@@ -35,31 +35,44 @@ jobs:
   htmlproofer:
     name: HTMLProofer
     runs-on: ubuntu-latest
-    
+
     steps:
       - name: Checkout EIP Repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-      
-      - name: Install OpenSSL
```
xinbenlv, line 43: Do we no longer need OpenSSL?

Pandapip1: No, we don't. Not sure why it was there in the first place.

xinbenlv: SG, consider adding "also remove OpenSSL" to PR title?

Pandapip1:
Would still like this merged.

Pandapip1:
@SamWilsn what do you think of this? Personally, I would find it helpful because of the annotations.

SamWilsn (approved):

SamWilsn:
Since this is in addition to the legacy link checker, I'm good to try this out. I'll revert if I see any problems.

PR #5362, opened 2022-07-28
Attempt to fix the Automerge action

Pandapip1:


.github/workflows/enable-automerge.yml
```diff
@@ -2,12 +2,6 @@ name: Automerge
 
 on:
   pull_request_target:
-    types:
-      - opened
-      - synchronized
-      - reopened
-      - edited
-      - ready_for_review
```
MicahZoltu, line 10: I *think* the proper format for this would be: `on: [ pull_request_target ]`?  At least, that is how I have seen it in the past (and YML is ridiculous).

PR #5365, opened 2022-07-28
Update Stale Bot Formatting

Pandapip1:
- Uses a single invocation of the action

PR #5366, opened 2022-07-28
Update wording for stale bot

Pandapip1:
~~Blocked by #5365~~

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,13 +20,15 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         ascending: true # Since we have so many issues, the stale bot finds it hard to keep track. This makes sure that at least the oldest are removed.
         # Issue config
-        stale-issue-message: There has been no activity on this issue in a month. It will be closed in a week if no further activity occurs
+        stale-issue-message: There has been no activity on this issue in a month. It will be closed in a week if no further activity occurs.
+        close-issue-message: This issue was closed due to inactivity. If you are still pursuing it, feel free to reopen it.
         days-before-issue-stale: 30
         days-before-issue-close: 7
         exempt-issue-labels: discussions-to,long-term
         stale-issue-label: stale
         # PR config
         stale-pr-message: There has been no activity on this pull request in a month. It will be closed in a week if no further activity occurs.
+        close-pr-message: This pull request was closed due to inactivity. If you are still pursuing it, feel free to reopen it and respond to any feedback or request a review in a comment.
         days-before-pr-stale: 30
         days-before-pr-close: 7
         exempt-pr-labels: long-term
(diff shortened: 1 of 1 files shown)


Pandapip1:
CC @MicahZoltu

.github/workflows/stale.yml
```diff
@@ -20,14 +20,14 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         ascending: true # Since we have so many issues, the stale bot finds it hard to keep track. This makes sure that at least the oldest are removed.
         # Issue config
-        stale-issue-message: There has been no activity on this issue for six months. It will be closed in a week if no further activity occurs. If you would like to move this EIP forward, please respond to any outstanding feedback or add a comment indicating that you have addressed all required feedback and are ready for a review.
+        stale-issue-message: There has been no activity on this issue in six months. It will be closed in two weeks if no further activity occurs. If you are still pursuing it, feel free to respond to any feedback.
```
MicahZoltu, line 23: For the record, I believe `for` is just as correct as `in` [in|for] this case.  I don't really care which we use, but I'm curious why the desire to change it?

Regarding the rest of the change, I feel like the original wording gave authors more clear instructions on how to proceed.  What is the perceived advantage of using the more concise statement?

Pandapip1: > For the record, I believe for is just as correct as in [in|for] this case. I don't really care which we use, but I'm curious why the desire to change it?

In the stale bot documentation, the phrasing I used above is roughly what is used (including the in).

MicahZoltu: Is there a more complete argument for using the proposed wording over our current wording besides "it is how someone else on the internet does it"?

Pandapip1: Uh, no, but "It is how most other people on the internet do it" works better.

MicahZoltu: Most humans are wrong about most things, so I don't find that to be a particularly compelling argument.  😆

Pandapip1: > Most humans are wrong about most things, so I don't find that to be a particularly compelling argument. 😆

[citation needed]

Pandapip1:
@MicahZoltu any objections?

.github/workflows/stale.yml
Pandapip1: I think most people would disagree with you, FYI.

Pandapip1: > Regarding the rest of the change, I feel like the original wording gave authors more clear instructions on how to proceed. What is the perceived advantage of using the more concise statement?

Somehow I missed this. I think that in either scenario, it's clear that there is an automated system that is noting that the EIP has a lack of activity, and that the way to resolve this is by creating some activity. The advantage of being more concise is that it's more concise.

MicahZoltu:
My position is still the same: I think the old wording was more clear and gave users better directions on how to proceed compared to the new wording.  Given that we disagree on which is preferred, I would like to get more editor feedback before moving forward with this change.

Issue #5315, opened 2022-07-22
Standardize April 1 EIPs


Pandapip1:
I know this is a random time to post about this but... can we please have April 1st EIPs? April 1st RFCs are a well-known thing: https://en.wikipedia.org/wiki/April_Fools%27_Day_Request_for_Comments

I suggest we do the same thing as IETF and merge April 1st EIPs based on cleverness and humor.

The following PRs would also be re-opened and merged on April 1, 2023:
https://github.com/ethereum/EIPs/pull/4963
https://github.com/ethereum/EIPs/pull/4962
https://github.com/ethereum/EIPs/pull/4961
https://github.com/ethereum/EIPs/pull/4959
https://github.com/ethereum/EIPs/pull/3455

abcoathup:
We could standardize on a category with a prefix EAF (EIP April First/Fools)

MicahZoltu:
It creates an unnecessary burden on EIP editors, which are already sorely understaffed.  If you want to create an April Fools joke standard, why not just use Hack MD and post a link to it on Twitter?  You can say you are "in the process of drafting a new EIP and looking for feedback" if you want it to be more "believable".

I like a good joke as much as the next guy, but I don't like it when a joke creates work for someone.  Jokes should ideally (IMO) be setup in a way that they don't interfere with anyone's ability to be productive and don't force themselves on anyone's workflow or life who isn't interested in participating in the joke.

Pandapip1:
> I like a good joke as much as the next guy, but I don't like it when a joke creates work for someone. Jokes should ideally (IMO) be setup in a way that they don't interfere with anyone's ability to be productive and don't force themselves on anyone's workflow or life who isn't interested in participating in the joke.

If I didn't like the work, I wouldn't be volunteering :)

SamWilsn:
I volunteer to clean up the mess from april fools jokes.

For the record, I still want to get [EIP-4959](https://github.com/ethereum/EIPs/pull/4959) merged, since it satisfies the requirements for an Informational EIP.

MicahZoltu:
If we had a reliable mechanism for isolating jokes from real work I would care far, far less.  However, at the moment I get an email for *every single thing* that happens on this repository and my inbox fills up every day of the week except Sunday because of that.  On April Fools day, I get an even more full inbox but it is of stuff that I am supposed to ignore, but the effort required to identify which items should be ignored and which shouldn't is often *higher* than the effort required to do my regular work.

[...]

PR #5370, opened 2022-07-29
Corrects the name of EIPW.

MicahZoltu:
This appears to just be a typo, fixing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
       run: eipv EIPS/ --ignore=title_max_length,missing_discussions_to --skip=eip-20-token-standard.md
 
   eipw-validator:
-    name: EIPW Validator
+    name: EIP Walidator
     runs-on: ubuntu-latest
     steps:
       - name: Save PR number


SamWilsn:
The name of the validator is `eipw`, but I can certainly live with this joke.

SamWilsn (approved):

MicahZoltu:
I would be OK with renaming it to `EIPW` if you prefer @SamWilsn.  I feel like `EIPV Validator` is redundant because it expands to "Ethereum Improvement Proposal Validator Validator", and since EIPW is just EIPV with an increment of the `V`, the expansion would be "Ethereum Improvement Proposal Walidator Validator", which still feels redundant.

.github/workflows/ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       run: eipv EIPS/ --ignore=title_max_length,missing_discussions_to --skip=eip-20-token-standard.md
 
   eipw-validator:
-    name: EIPW Validator
+    name: EIP Walidator
```
Pandapip1, line 70:
```suggestion
    name: EIPU Ualidator
```
If it's a w (double-u) then I believe that it should really be corrected to this.

In all seriousness, I wasn't aware that you had agreed with https://github.com/ethereum/EIPs/issues/5315. Good to know :)

MicahZoltu: I haven't!  I really do want to address the problem of EIPW being neither/both an acronym and a word at the same time.  Either it should be named `eipw` (not an acronym, the letters don't mean anything), or it should be `EIP W???` where the W is the first letter of some word.  I don't actually care what that word is, and it can be a backronym.  I just don't think it should be `EIPW Validator` because that implies that it is something that validates EIPWs, and those aren't a thing.

Pandapip1:
```suggestion
    name: WIP Validator
```
How about this? It does validate WIP EIPs :)

Pandapip1:
Also, it's worth noting that the branch protection settings will have to be **immediately** updated once this PR is merged.

Pandapip1 (approved):
LGTM otherwise

MicahZoltu:
@SamWilsn Did you prefer just changing the name to `EIPW` or `eipw` instead of `EIP Walidator`?  I'm fine with any of those options.

Pandapip1:
I'm fine with `EIP Walidator`, and Sam has already agreed to live with the joke.

MicahZoltu:
Manually merging since this seems to be getting stuck on failing to run the old check.

MicahZoltu:
Branch protection rule updated.

PR #5373, opened 2022-07-29
fix: small typo on event name

aramalipoor:


Pandapip1 (approved):
Haha, good catch! 😄

MicahZoltu (changes requested):
This is a normative change to a final EIP so you cannot change the event name, unfortunately.  You could create a new EIP that is exactly like this one but with that text fixed, but you cannot rename the event now that it is final.

Pandapip1:
> This is a normative change to a final EIP

I think it's perfectly obvious what the correct thing is here. I am entirely in favour of this errata.

MicahZoltu:
It doesn't matter what author intent was, what matters is that if someone deployed a contract that was compatible with this standard as finalized, it would no longer be compatible with the standard after this change this we cannot make such a normative change without a new EIP.

If the belief is that no one has implemented this yet, then a new EIP (new number) with the exact same text save that one change would be the right path forward, then people can write contracts compatible with that standard.

The thing to remember is that the EIP process is a standardization process, not a collection of good ideas.  A collection of good ideas can be mutated over time as corrections are made and learning occurs.  A standard is only useful if it is immutable.

Pandapip1:
I think that the number of implementations of this EIP is low. I will submit said PR on their behalf though.

MicahZoltu:
I'm going to close this in favor of #5377.

aramalipoor:
Totally understand the process @MicahZoltu it's fine by me :) and i see @Pandapip1 already created a PR for the new EIP.

Should there be a reference to the new EIP from this one? To encourage people to treat this one as deprecated in favor of new one, or we don't have deprecation concept in EIPs yet (have to educate myself more on the process)?

Pandapip1:
We're working on a superseded header, but it isn't clear how to do it yet.

MicahZoltu:
I'm pretty strongly against superseding because it gives undue power to authors of old EIPs.  For example, the authors of EIP-20 would now get to decide for the entire community which of the dozens of token standards is the blessed one that "officially" supersedes EIP-20.

At the other extreme are EIPs like this and #5377 which are identical but with a minor typo fix and there is no good reason for people to use this one once the bugfixed one is final and it likely is obvious to any onlooker that 5377 supersedes this.

However, I don't like editors having to make subtle nuanced decisions like this because for most cases things aren't as clear cut as EIP-20 or this one, they fall somewhere in that fuzzy middle where it isn't obvious if supersedes is appropriate or not.

PR #5374, opened 2022-07-30
Add updateEIP to exempt labels

Pandapip1:
Rationale: if a `updateEIP` PR doesn't get merged immediately, it's either waiting on editor or author approval and should remain open until the author or editor approves.


MicahZoltu:
It is very common for someone to submit a PR for an EIP they aren't an author of and for the author to never respond/approve the PR.  I believe that before the stale bot, a significant portion of the stale PRs were exactly this scenario, so disabling stale bot on them feels like we would be going the opposite of the desired direction (closing fewer things that are actually stale).

Pandapip1:
Good idea.

PR #5377, opened 2022-07-30
Create EIP-4401 (typographical fix of EIP-4400)

Pandapip1:
Closes https://github.com/ethereum/EIPs/pull/5373

**I have assigned this EIP number `4401`.**

### DIFF
--- a/EIPS/eip-4401.md
+++ b/EIPS/eip-4401.md
@@ -1,0 +1,115 @@
+---
+eip: 4401
+title: EIP-721 Consumable Extension - Errata
+description: Interface extension for EIP-721 consumer role
+author: Daniel Ivanov (@Daniel-K-Ivanov), George Spasov (@Perseverance), Pandapip1 (@Pandapip1)
+discussions-to: https://ethereum-magicians.org/t/EIP-4400-EIP721consumer-extension/7371
+status: Final
+type: Standards Track
+category: ERC
+created: 2021-10-30
+requires: 165, 721, 4400
+---
+
+## Abstract
+
+This standard fixes a small typo in [EIP-4400](./eip-4400.md),
+
+This specification defines standard functions outlining a `consumer` role for instance(s) of [EIP-721](./eip-721.md). An implementation allows reading the current `consumer` for a given NFT (`tokenId`) along with a standardized event for when an `consumer` has changed. The proposal depends on and extends the existing [EIP-721](./eip-721.md).
+
+## Motivation
+
+Many [EIP-721](./eip-721.md) contracts introduce their own custom role that grants permissions for utilising/consuming a given NFT instance. The need for that role stems from the fact that other than owning the NFT instance, there are other actions that can be performed on an NFT. For example, various metaverses use `operator` / `contributor` roles for Land (EIP-721), so that owners of the land can authorise other addresses to deploy scenes to them (f.e. commissioning a service company to develop a scene).
+
+It is common for NFTs to have utility other than ownership. That being said, it requires a separate standardized consumer role, allowing compatibility with user interfaces and contracts, managing those contracts.
+
+Having a `consumer` role will enable protocols to integrate and build on top of dApps that issue EIP-721 tokens. One example is the creation of generic/universal NFT renting marketplaces.
+
+Example of kinds of contracts and applications that can benefit from this standard are:
+- metaverses that have land and other types of digital assets in those metaverses (scene deployment on land, renting land / characters / clothes / passes to events etc.)
+- NFT-based yield-farming. Adopting the standard enables the "staker" (owner of the NFT) to have access to the utility benefits even after transferring his NFT to the staking contract
+
+## Specification
+
+The keywords “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+Every contract compliant to the `EIP721Consumable` extension MUST implement the `IEIP721Consumable` interface. The **consumer extension** is OPTIONAL for EIP-721 contracts.
+
+```solidity
+/// @title EIP-721 Consumer Role extension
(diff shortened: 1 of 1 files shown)


EIPS/eip-4401.md
```diff
@@ -0,0 +1,115 @@
+---
+eip: 4401
+title: EIP-721 Consumable Extension - Errata
+description: Interface extension for EIP-721 consumer role
+author: Daniel Ivanov (@Daniel-K-Ivanov), George Spasov (@Perseverance), Pandapip1 (@Pandapip1)
```
Pandapip1, line 5: @Daniel-K-Ivanov @Perseverance I would like to double check you're okay with me adding myself as an author here.

Pandapip1:
This is a carbon copy of the original, plus the fix, the author change, the title, and the snippet at the beginning of the abstract. This *should* be okay to be merged as final, as it has almost exactly the same spec as an existing proposal.

EIPS/eip-4401.md
```diff
@@ -0,0 +1,115 @@
+---
+eip: 4401
```
MicahZoltu, line 2:
```suggestion
eip: 5377
```
We have a big enough problem with EIP number selection as is (and contention/debate over it), I think following a different selection mechanism here will just make the situation marginally worse.

EIPS/eip-4401.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+Every contract implementing the `EIP721Consumable` extension is free to define the permissions of a `consumer` (e.g. what are consumers allowed to do within their system) with only one exception - consumers MUST NOT be considered owners, authorised operators or approved addresses as per the EIP-721 specification. Thus, they MUST NOT be able to execute transfers & approvals.
+
+The `consumerOf(uint256 _tokenId)` function MAY be implemented as `pure` or `view`.
+
+The `changeConsumer(address _consumer, uint256 _tokenId)` function MAY be implemented as `public` or `external`.
+
+The `ConsumEIPhanged` event MUST be emitted when a consumer is changed.
```
MicahZoltu, line 72:
```suggestion
The `ConsumerChanged` event MUST be emitted when a consumer is changed.
```

EIPS/eip-4401.md
```diff
@@ -0,0 +1,115 @@
...
+status: Final
+type: Standards Track
+category: ERC
+created: 2021-10-30
+requires: 165, 721, 4400
+---
+
+## Abstract
+
+This standard fixes a small typo in [EIP-4400](./eip-4400.md),
```
MicahZoltu, line 16:
```suggestion
Identical to [EIP-4400](./eip-4400.md), but with a typo fixed.
```

EIPS/eip-4401.md
```diff
@@ -0,0 +1,115 @@
...
+
+The other name for the role that was considered was `operator`, however it is already defined and used within the `EIP-721` standard.
+
+### Restriction on the Permissions
+
+There are numerous use-cases where a distinct role for NFTs is required that MUST NOT have owner permissions. A contract that implements the consumer role and grants ownership permissions to the consumer renders this standard pointless.
+
+## Backwards Compatibility
+
+This standard is compatible with current EIP-721 standards. There are no other standards that define a similar role for NFTs and the name (`consumer`) is not used by other EIP-721 related standards.
```
MicahZoltu, line 99: Probably worth mentioning that this isn't backward compatible with EIP-4400, due to the event name change.

EIPS/eip-4401.md
Pandapip1: I mean, technically I can assign my own number here, but fine.

MicahZoltu: *Generally* we try to have editors who are authors not act in an editorial capacity for the EIP they are authors of.  That being said, I'm not an ERC editor so one of the ERC editors could choose a different number for this (though, with the recent events I'm generally pretty apprehensive about any non-random-like number assignments happening).

EIPS/eip-5377.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Specification
+
+The keywords “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+Every contract compliant to the `EIP721Consumable` extension MUST implement the `IEIP721Consumable` interface. The **consumer extension** is OPTIONAL for EIP-721 contracts.
+
+```solidity
+/// @title EIP-721 Consumer Role extension
+///  Note: the EIP-165 identifier for this interface is 0x953c8dfa
```
SamWilsn, line 40: I think this interface id changes when you change an event?

SamWilsn (changes requested):

EIPS/eip-5377.md
Pandapip1: This is correct. Not sure if they used the correct identifier or not, but I have a trick.

Pandapip1: "fixed"

EIPS/eip-5377.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Specification
+
+The keywords “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+Every contract compliant to the `EIP721Consumable` extension MUST implement the `IEIP721Consumable` interface. The **consumer extension** is OPTIONAL for EIP-721 contracts.
+
+```solidity
+/// @title EIP-721 Consumer Role extension
+///  Note: the EIP-165 identifier for this interface is `type(IEIP721Consumable).interfaceId`
```
SamWilsn, line 40: Generally I recommend including the interface id as an additional check for implementers.

Pandapip1: I don't like including it at all since it's explicitly stated in EIP-165, as well as it adds another thing that may become out of date and or might be wrong.

PR #5379, opened 2022-07-30
Update EIP-1: Add License Checker

Pandapip1:
Makes sure all solidity files are CC0.


Pandapip1:
Testing...

Pandapip1:
Well uh, glad I tested 😆

Pandapip1:
Well, I guess I'm about to be adding some identifiers to a few files. This will now need a manual merge.

Pandapip1:
Actually, on second thought, I'll just exempt those files.

Pandapip1:
image
Can confirm it works... I guess I'll wait for @MicahZoltu's judgement.

.licenserc.json
```diff
@@ -0,0 +1,18 @@
...
+    "assets/eip-3525",
+    "assets/eip-4361",
+    "assets/eip-4400",
+    "assets/eip-4519",
+    "assets/eip-4675",
+    "assets/eip-4886",
+    "assets/eip-4907",
+    "assets/eip-4987",
+    "assets/eip-5006",
+    "assets/eip-5007"
```
MicahZoltu, line 16: Why is this specific set of things ignored?  Is this every EIP that has some other license and we are grandfathering them in for now?

MicahZoltu (commented):
Does this *not* error if files are missing a license line at the top of the file entirely?  I have always found license at top of file quite silly and would like to continue supporting people just leaving the line out entirely.

Pandapip1:
> Does this not error if files are missing a license line at the top of the file entirely? I have always found license at top of file quite silly and would like to continue supporting people just leaving the line out entirely.

No, it does, but IIRC the solidity *compiler* warns you if you don't have a license file, as if it's deployed to Ethereum and the contract source verified, the license should be obvious from the file alone.

That's my case for `.sol` files. I am okay with removing the req for `.ts` and `.js` though.

MicahZoltu:
Is it possible to make it so it checks to make sure licenses are CC0 if present, but ignore the file if they aren't present?  That would be my preference if it is possible.

.licenserc.json
Pandapip1: Essentially, yes.

Pandapip1:
> Is it possible to make it so it checks to make sure licenses are CC0 if present, but ignore the file if they aren't present?

Again, that's explicitly not allowed by the solidity style guide. I'd be glad to make an issue if it's a blocker, but I would much rather force people to do the right thing and make it obvious that their code can be used for whatever purpose.

MicahZoltu:
> that's explicitly not allowed by the solidity style guide

We are not beholden to following other people's bad decisions just because they write them down in a style guide.  😄  The argument that the license should be embedded in the contract so that it shows up in various tools that deterministically build based on the source file alone (like Etherscan) is a much more compelling argument that including a license in the header is a good idea, so I recommend you pursue that angle rather than "because other people do it".

> I would much rather force people to do the right thing

Aside from the Ethesrcan argument above, I'm very much unconvinced this practice is "the right thing" and in almost all other contexts I think it is "the wrong thing".

----

All the above being said, I think I'm compelled enough by the Etherscan issue to allow this for .sol files, though I would still like licenses to not be required for other file types as I consider it to be a ridiculous pattern.

Pandapip1:
I can get behind not requiring licenses in other files (after all, most of the issues with using the wrong license are in .sol files due to most developers automatically putting in the MIT license header in their .sol files). I have already made that change.

Pandapip1:
Modifies EIP-1 now, so should be safe to mark as ready for review.

Pandapip1:
@MicahZoltu @axic @lightclient @SamWilsn @gcolvin any objections to this?

- This only forces an SPDX identifier of `CC0-1.0` for `.sol` files. All other file types are unaffected.
- The current system only looks at changed files. No exceptions are needed, nor are any unnecessary file changes.
- Adds a short snippet to EIP-1 reminding EIP authors that assets must be cc0 if they can be.

.github/workflows/ci.yml
```diff
@@ -90,3 +90,36 @@ jobs:
...
+          cd all
+          for FILE in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            [ -f ${FILE} ] && mkdir -p $(dirname "../unchanged/${FILE}")
+            [ -f ${FILE} ] && cp ${FILE} ../unchanged/${FILE}
+          done
+          cd ..
+          rm -rf all
+          cp -r unchanged/* .
+          cd ..
+          rm -rf unchanged
```
MicahZoltu, line 122: While not a hard blocker, I'm not a huge fan of obscure uncommented shell scripts.  Is there any way to *easily* turn this into something that is more readable like a python or TS script (or perhaps a GitHub action for "remove all unchanged files"?

MicahZoltu: Alternatively, some comments may help make this block more understandable... like I'm pretty sure at the moment that `shopt -u dotglob` was a cat walking across your keyboard while you were in the middle of authoring this.

Pandapip1: Yea, I'm planning on making it an action. This is a nice stop-gap.

Pandapip1: And no, `shopt -u dotglob` is not due to an unherded Ethereum cat.

MicahZoltu: I have been thinking about this more, and I think I preferred the file ignore-list approach as it was overall simpler than this one, which requires an extra two steps, one of which is an external dependency and one of which is an obscure script.  In general, I really like simple solutions as our CI already has enough problems as it is due largely (I suspect) to its complexity.  If you were able to create a single action that handled all of this I would be more compelled to keep the pattern as it is, but 4 actions just to check for the presence of first line of files feels like overkill on complexity.

EIPS/eip-1.md
```diff
@@ -195,7 +195,7 @@ References to other EIPs should follow the format `EIP-N` where `N` is the EIP n
 
 ## Auxiliary Files
 
-Images, diagrams and auxiliary files should be included in a subdirectory of the `assets` folder for that EIP as follows: `assets/eip-N` (where **N** is to be replaced with the EIP number). When linking to an image in the EIP, use relative links such as `../assets/eip-1/image.png`.
+Images, diagrams and auxiliary files should be included in a subdirectory of the `assets` folder for that EIP as follows: `assets/eip-N` (where **N** is to be replaced with the EIP number). When linking to an image in the EIP, use relative links such as `../assets/eip-1/image.png`. These files must be in the public domain (`CC0`). Solidity files must use the SPDX Identifier `CC0-1.0`.
```
SamWilsn, line 198:
```suggestion
Images, diagrams and auxiliary files should be included in a subdirectory of the `assets` folder for that EIP as follows: `assets/eip-N` (where **N** is to be replaced with the EIP number). When linking to an image in the EIP, use relative links such as `../assets/eip-1/image.png`. These files must be released into the public domain (or as close as is legally possible) using [CC0](../LICENSE.md). Solidity files must use the SPDX Identifier `CC0-1.0`.
```

SamWilsn (commented):
I still do not agree that we should require assets to be CC0.

Pandapip1:
@SamWilsn I'm pretty sure during meeting 63 we agreed to merge this.

EIPS/eip-1.md
```diff
@@ -195,7 +195,7 @@ References to other EIPs should follow the format `EIP-N` where `N` is the EIP n
 
 ## Auxiliary Files
 
-Images, diagrams and auxiliary files should be included in a subdirectory of the `assets` folder for that EIP as follows: `assets/eip-N` (where **N** is to be replaced with the EIP number). When linking to an image in the EIP, use relative links such as `../assets/eip-1/image.png`.
+Images, diagrams and auxiliary files should be included in a subdirectory of the `assets` folder for that EIP as follows: `assets/eip-N` (where **N** is to be replaced with the EIP number). When linking to an image in the EIP, use relative links such as `../assets/eip-1/image.png`. These files must be released into the public domain (or as close as is legally possible) using [CC0](../LICENSE.md). Solidity files must use the SPDX Identifier `CC0-1.0`.
```
xinbenlv, line 198: I like to agree with the requirment for asset file to be CC-0, but only on condition of that the reference implementation don't have to live in `ethereum/EIPs` repository

Pandapip1: I disagree. We should definitely force authors to use CC0 if they legally can, since it means that users don't have to worry about attribution and all that stuff that just can't apply to blockchains. If they legally can't use CC0, then and **only** then will we allow non-CC0 assets. But we will allow them.

lightclient: I'm afraid that "as close as is legally possible" is opening the door for a fair bit of conflict in the future. Now editors need to verify claims from authors that "this is the best license I can give this!1!!".

Pandapip1: Well... we don't necessarily need to verify those claims. Depending on how restrictive the license is, we can, well, not approve the PR.

Pandapip1:
Somehow the stale bot hasn't been reprompting me.

I still think this is a good idea. This doesn't need to be a required check, but it makes my job of scanning for aberrant licenses easier.

Pandapip1:
Dismissing stale bot.

Pandapip1:
@SamWilsn please rereview

Pandapip1:
Re-requesting @SamWilsn

Pandapip1:
Re-cc @SamWilsn

Pandapip1:
CC @SamWilsn

SamWilsn:
@Pandapip1 I assume you're pinging me just to keep the stale bot away, but I won't approve a PR that only allows `CC0-1.0` assets.

Pandapip1:
> @Pandapip1 I assume you're pinging me just to keep the stale bot away, but I won't approve a PR that only allows CC0-1.0 assets.

This was something I was not aware of. Would you mind telling me which licenses can be allowed?

SamWilsn:
> This was something I was not aware of. Would you mind telling me which licenses can be allowed?

I think we can start with `CC0-1.0`, `Unlicense`, `Apache-2.0`, and `MIT`? Probably best discussed on the FEM thread though.

Pandapip1:
I didn't know there was a FEM thread. I'll add the non-copyleft ones for now.

Pandapip1:
Also, I assume you mean Unilicence as opposed to Unlicenced?

Pandapip1:
The license checker isn't sophisticated enough to do SPDX checking. I'll probably make my own CI tool.

PR #5389, opened 2022-08-02
Promote EIP-5313 to last call

fulldecent:


MicahZoltu:
@Pandapip1 It seems the merge bot doesn't automerge ~100% of the time now (as seen by the fact that your approvals now come with a comment telling people to close and re-open).  I think that perhaps it is time to turn Kodiak back on until a solution can be found.

Pandapip1:
> I know exactly what's causing the issue, and I can work on a solution now.

PR #5360, opened 2022-07-28
Change Stale Thresholds

Pandapip1:
## Abstract

The stale bot thresholds are too high. This PR reduces them.

## Motivation

Reduces EIP Editor work.

## Specification

1. The stale issue threshold was reduced from 6 months to four weeks
2. The stale PR threshold was reduced from two months to two weeks
3. The close PR threshold was *increased* from one week to two weeks
4. The stale PR label was changed to a more descriptive one

## Rationale

1. Encourages people to discuss relevant issues, and reduces the time it takes for irrelevant issues to disappear. Issues may be re-opened at any time if they become relevant again.
2. Encourages authors to update their EIP with editor-provided feedback or remind editors to review. PRs may be re-opened at any time if the author re-gains interest.
3. Since item 2 reduced the threshold so much, the PR close threshold was increased to give authors enough time to ensure that their PRs will not be mistakenly closed.
4. If the editors are actually the ones that need to review, the author will quickly make that clear, which will remove the stale tag.

PR #5392, opened 2022-08-02
Add Waiting on Author Label on CI Failure

Pandapip1:
Pretty self-explanatory. Might conflict with #5360 (the worst that will happen is that the stale bot will remove the labels, so there isn't a major problem).

### DIFF
--- a/.github/workflows/ci-comment.yml
+++ b/.github/workflows/ci-comment.yml
@@ -9,9 +9,8 @@ on:
 jobs:
   on-failure:
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
     steps:
-      - name: 'Download artifact'
+      - name: Download Artifact
         uses: actions/github-script@v6
         with:
           script: |
@@ -32,11 +31,12 @@ jobs:
             let fs = require('fs');
             fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/pr_number.zip`, Buffer.from(download.data));
 
-      - name: 'Unzip artifact'
+      - name: Unzip Artifact
         run: unzip pr_number.zip
 
-      - name: 'Comment on PR'
+      - name: On Failure
         uses: actions/github-script@v6
+        if: ${{ github.event.workflow_run.conclusion == 'failure' }}
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-comment.yml
```diff
@@ -55,4 +55,25 @@ jobs:
               issue_number: issue_number,
               body: `The commit ${pr_sha} (as a parent of ${merge_sha}) contains errors. Please inspect the [Run Summary](${context.payload.workflow_run.html_url}) for details.`
             });
+            await octokit.request(`POST /repos/{owner}/{repo}/issues/{issue_number}/labels`, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              labels: [ 'waiting: response from author' ]
+            })
```
SamWilsn, line 63: Any reason to use the raw POST over [`addLabels`](https://octokit.github.io/rest.js/v18#issues-add-labels) (all PRs are issues)?

Pandapip1: Laziness.

MicahZoltu: I think we should prefer using the SDK over the rest interface when possible.  It is less likely we'll screw something up, and the code becomes more readable.

.github/workflows/ci-comment.yml
```diff
@@ -55,4 +55,25 @@ jobs:
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            let fs = require('fs');
+            let issue_number = Number(fs.readFileSync('./pr_number', 'utf8').trim());
+            await octokit.request(`DELETE /repos/{owner}/{repo}/issues/{issue_number}/labels/{name}`, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              name: 'waiting: response from author'
+            })
```
MicahZoltu, line 78: As mentioned above, we should similarly use https://octokit.github.io/rest.js/v18#issues-remove-label for this.

.github/workflows/ci-comment.yml
```diff
@@ -55,4 +55,25 @@ jobs:
...
+
+      - name: On Success
+        uses: actions/github-script@v6
+        if: ${{ github.event.workflow_run.conclusion != 'failure' }}
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            let fs = require('fs');
+            let issue_number = Number(fs.readFileSync('./pr_number', 'utf8').trim());
```
MicahZoltu, line 72: I know this file uses `readFileSync` elsewhere, but as a general best practice we should be using async methods.  Also, we should be using `const` rather than `let` as another best practice (doesn't matter much for a tiny script like this, but a generally good habit to be in when we are writing TS code).

.github/workflows/ci-comment.yml
Pandapip1: Agreed here. I'd prefer using a pre-existing action, and will likely do so.

.github/workflows/ci-comment.yml
Pandapip1: The problem is fetching the `issue_number`. I'm planning on making a few PRs to clear up some loose ends in the CI.

.github/workflows/ci-comment.yml
Pandapip1: This isn't TS, but your point is valid.

PR #5393, opened 2022-08-02
Update eip-4626.md

pandadefi:
With the recent hack of Byte-Masons/tarot-multi-strategy implementation of 4626, it seems we should make it clear, msg.sender should be allowed to spend owner funds.

### DIFF
--- a/EIPS/eip-4626.md
+++ b/EIPS/eip-4626.md
@@ -383,6 +383,8 @@ MUST support a withdraw flow where the shares are burned from `owner` directly w
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `withdraw` execution, and are accounted for during `withdraw`.
 
+MUST check msg.sender can spend owner funds, assets needs to be converted to shares and shares should be check for allowance.
+
 MUST revert if all of `assets` cannot be withdrawn (due to withdrawal limit being reached, slippage, the owner not having enough shares, etc).
 
 Note that some implementations will require pre-requesting to the Vault before a withdrawal may be performed. Those methods should be performed separately.
@@ -468,6 +470,8 @@ MUST support a redeem flow where the shares are burned from `owner` directly whe
 MUST support a redeem flow where the shares are burned from `owner` directly where `msg.sender` has ERC-20 approval over the shares of `owner`.
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `redeem` execution, and are accounted for during `redeem`.
+
+MUST check msg.sender can spend owner funds using allowance.
 
 MUST revert if all of `shares` cannot be redeemed (due to withdrawal limit being reached, slippage, the owner not having enough shares, etc).
 
--- a/EIPS/eip-4626.md
+++ b/EIPS/eip-4626.md
@@ -471,7 +471,7 @@ MUST support a redeem flow where the shares are burned from `owner` directly whe
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `redeem` execution, and are accounted for during `redeem`.
 
-MUST check msg.sender can spend owner funds using allowance.
+SHOULD check `msg.sender` can spend owner funds using allowance.
 
 MUST revert if all of `shares` cannot be redeemed (due to withdrawal limit being reached, slippage, the owner not having enough shares, etc).
 
--- a/EIPS/eip-4626.md
+++ b/EIPS/eip-4626.md
@@ -383,7 +383,7 @@ MUST support a withdraw flow where the shares are burned from `owner` directly w
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `withdraw` execution, and are accounted for during `withdraw`.
 
-MUST check msg.sender can spend owner funds, assets needs to be converted to shares and shares should be check for allowance.
+MUST check msg.sender can spend owner funds, assets needs to be converted to shares and shares should be checked for allowance.
 
 MUST revert if all of `assets` cannot be withdrawn (due to withdrawal limit being reached, slippage, the owner not having enough shares, etc).
 
--- a/EIPS/eip-4626.md
+++ b/EIPS/eip-4626.md
@@ -1,7 +1,7 @@
 ---
 eip: 4626
 title: Tokenized Vault Standard
-description: A standard for tokenized Vaults with a single underlying ERC-20 token.
+description: A standard for tokenized Vaults with a single underlying EIP-20 token.
 author: Joey Santoro (@joeysantoro), t11s (@transmissions11), Jet Jadeja (@JetJadeja), Alberto Cuesta Cañada (@alcueca), Señor Doggo (@fubuloubu)
 discussions-to: https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900
 status: Final
@@ -14,8 +14,8 @@ requires: 20, 2612
 ## Abstract
 
 The following standard allows for the implementation of a standard API for tokenized Vaults
-representing shares of a single underlying [ERC-20](./eip-20.md) token.
-This standard is an extension on the ERC-20 token that provides basic functionality for depositing
+representing shares of a single underlying [EIP-20](./eip-20.md) token.
+This standard is an extension on the EIP-20 token that provides basic functionality for depositing
 and withdrawing tokens and reading balances.
 
 ## Motivation
@@ -28,20 +28,20 @@ A standard for tokenized Vaults will lower the integration effort for yield-bear
 
 ## Specification
 
-All ERC-4626 tokenized Vaults MUST implement ERC-20 to represent shares.
+All [EIP-4626](./eip-4626.md) tokenized Vaults MUST implement EIP-20 to represent shares.
 If a Vault is to be non-transferrable, it MAY revert on calls to `transfer` or `transferFrom`.
-The ERC-20 operations `balanceOf`, `transfer`, `totalSupply`, etc. operate on the Vault "shares"
+The EIP-20 operations `balanceOf`, `transfer`, `totalSupply`, etc. operate on the Vault "shares"
 which represent a claim to ownership on a fraction of the Vault's underlying holdings.
 
-All ERC-4626 tokenized Vaults MUST implement ERC-20's optional metadata extensions.
+All EIP-4626 tokenized Vaults MUST implement EIP-20's optional metadata extensions.
 The `name` and `symbol` functions SHOULD reflect the underlying token's `name` and `symbol` in some way.
 
-ERC-4626 tokenized Vaults MAY implement [EIP-2612](./eip-2612.md) to improve the UX of approving shares on various integrations.
+EIP-4626 tokenized Vaults MAY implement [EIP-2612](./eip-2612.md) to improve the UX of approving shares on various integrations.
 
 ### Definitions:
 
 - asset: The underlying token managed by the Vault.
-  Has units defined by the corresponding ERC-20 contract.
+  Has units defined by the corresponding EIP-20 contract.
 - share: The token of the Vault. Has a ratio of underlying assets
   exchanged on mint/deposit/withdraw/redeem (as defined by the Vault).
 - fee: An amount of assets or shares charged to the user by the Vault. Fees can exists for
@@ -55,7 +55,7 @@ ERC-4626 tokenized Vaults MAY implement [EIP-2612](./eip-2612.md) to improve the
 
 The address of the underlying token used for the Vault for accounting, depositing, and withdrawing.
 
-MUST be an ERC-20 token contract.
+MUST be an EIP-20 token contract.
 
 MUST _NOT_ revert.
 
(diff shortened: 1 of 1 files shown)


Pandapip1:
This is a final EIP. If you would like to modify it, make a copy of it and submit it as a new EIP.

Pandapip1:
CC @MicahZoltu to close this.

fubuloubu:
> This is a final EIP. If you would like to modify it, make a copy of it and submit it as a new EIP.

This is an update to clarify a security property of an interface, it does not functionally modify it

fubuloubu (approved):
As an author, I approve this update, it is a vital clarification of a security property for this EIP and does not functionally modify it (e.g. implementors should have already been doing this, but this just makes it crystal clear)

EIPS/eip-4626.md
```diff
@@ -383,6 +383,8 @@ MUST support a withdraw flow where the shares are burned from `owner` directly w
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `withdraw` execution, and are accounted for during `withdraw`.
 
+MUST check msg.sender can spend owner funds, assets needs to be converted to shares and shares should be check for allowance.
```
transmissions11, line 386:
```suggestion
MUST check msg.sender can spend owner funds, assets needs to be converted to shares and shares should be checked for allowance.
```

Pandapip1:
The `EIP Walidator` errors need to be fixed. If you can fix them without making any non-normative changes, then I am fine merging this PR if another editor agrees.

fubuloubu:
> The `EIP Walidator` errors need to be fixed. If you can fix them without making any non-normative changes, then I am fine merging this PR if another editor agrees.

@pandadefi please change `ERC` to `EIP` in any links mentioned by the linter, and then `Reference Implementation` has an extra s at the end

pandadefi:
I am still getting the following error:
```
error[markdown-link-first]: the first match of the given pattern must be a link
```
the first reference to EIP-20 is linked https://github.com/ethereum/EIPs/pull/5393/files#diff-027d57a2e6989398a18dfbc27faa37253afbdf7a91b666aa7301261385a20911R17

EIPS/eip-4626.md
```diff
@@ -379,10 +379,12 @@ MUST emit the `Withdraw` event.
 
 MUST support a withdraw flow where the shares are burned from `owner` directly where `owner` is `msg.sender`.
 
-MUST support a withdraw flow where the shares are burned from `owner` directly where `msg.sender` has ERC-20 approval over the shares of `owner`.
+MUST support a withdraw flow where the shares are burned from `owner` directly where `msg.sender` has EIP-20 approval over the shares of `owner`.
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `withdraw` execution, and are accounted for during `withdraw`.
 
+MUST check msg.sender can spend owner funds, assets needs to be converted to shares and shares should be checked for allowance.
+
```
MicahZoltu, line 387: This appears to be a normative change, which cannot be made to a final EIP.  You can create a new EIP that mirrors this one but with this change, it would get a new EIP number but otherwise could be basically the same.

pandadefi: I think that's not a normative change more of a reminder.
@fubuloubu

MicahZoltu: If someone implemented this EIP previously and they did not check that `msg.sender` can send the owner funds, I believe it would have been compliant with this standard.  After this change, that contract would no longer be compliant.  Thus, this is a normative change.

pandadefi: If someone isn't checking that the msg.sender is not checking the funds can be spent anyone will be able to still the funds and that's what happened with the hack.

EIPS/eip-4626.md
```diff
@@ -1,7 +1,7 @@
 ---
 eip: 4626
 title: Tokenized Vault Standard
```
Pandapip1, line 3:
```suggestion
title: Tokenized Vaults
```

EIPS/eip-4626.md
```diff
@@ -1,7 +1,7 @@
...
-description: A standard for tokenized Vaults with a single underlying ERC-20 token.
+description: A standard for tokenized Vaults with a single underlying EIP-20 token.
```
Pandapip1, line 4:
```suggestion
description: Tokenized Vaults with a single underlying EIP-20 token.
```

EIPS/eip-4626.md
```diff
@@ -28,20 +28,20 @@ A standard for tokenized Vaults will lower the integration effort for yield-bear
...
-The ERC-20 operations `balanceOf`, `transfer`, `totalSupply`, etc. operate on the Vault "shares"
+The EIP-20 operations `balanceOf`, `transfer`, `totalSupply`, etc. operate on the Vault "shares"
 which represent a claim to ownership on a fraction of the Vault's underlying holdings.
 
-All ERC-4626 tokenized Vaults MUST implement ERC-20's optional metadata extensions.
+All EIP-4626 tokenized Vaults MUST implement EIP-20's optional metadata extensions.
 The `name` and `symbol` functions SHOULD reflect the underlying token's `name` and `symbol` in some way.
 
-ERC-4626 tokenized Vaults MAY implement [EIP-2612](./eip-2612.md) to improve the UX of approving shares on various integrations.
+EIP-4626 tokenized Vaults MAY implement [EIP-2612](./eip-2612.md) to improve the UX of approving shares on various integrations.
```
Pandapip1, line 39: This is interesting... EIP-2612 isn't final yet. @MicahZoltu?

EIPS/eip-4626.md
```diff
@@ -567,15 +571,15 @@ For applications that need an exact value that attempts to account for fees and
...
-ERC-4626 is fully backward compatible with the ERC-20 standard and has no known compatibility issues with other standards.
-For production implementations of Vaults which do not use ERC-4626, wrapper adapters can be developed and used.
+EIP-4626 is fully backward compatible with the EIP-20 standard and has no known compatibility issues with other standards.
+For production implementations of Vaults which do not use EIP-4626, wrapper adapters can be developed and used.
 
-## Reference Implementations
+## Reference Implementation
 
-See [Solmate ERC4626](https://github.com/Rari-Capital/solmate/blob/main/src/mixins/ERC4626.sol):
+See [Solmate EIP-4626](https://github.com/Rari-Capital/solmate/blob/main/src/mixins/ERC4626.sol):
```
Pandapip1, line 579: Must be moved to the assets folder.

EIPS/eip-4626.md
```diff
@@ -567,15 +571,15 @@ For applications that need an exact value that attempts to account for fees and
...
 a minimal and opinionated implementation of the standard with hooks for developers to easily insert custom logic into deposits and withdrawals.
 
-See [Vyper ERC4626](https://github.com/fubuloubu/ERC4626):
+See [Vyper EIP-4626](https://github.com/fubuloubu/ERC4626):
```
Pandapip1, line 582: Must be moved to the assets folder.

Pandapip1 (changes requested):

EIPS/eip-4626.md
MicahZoltu: I'm not suggesting it is a good idea, but it *is* standards compliant.  Also, there may be other mechanisms that are used internally to determine who is allowed to access the funds, so there may not be a vulnerability.

EIPS/eip-4626.md
MicahZoltu: Huh, I wonder why eipw didn't catch it cc @SamWilsn

EIPS/eip-4626.md
pandadefi: this PR can be closed.

fubuloubu: This is not a normative change because this invariant comes from ERC20, we are simply lifting it up to this ERC as a reminder to follow the same set of rules and avoid security consequences that have led to a serious exploit this week. Implementors should have been following this already, originally we thought it was okay that it was implicit given the rule from ERC20, but turns out we were wrong. 

If the editors still think this is normative, we can move this text to the Security section, which would no longer be a normative change to this ERC.

EIPS/eip-4626.md
Pandapip1: This was moved to final before eipw was added I think.

MicahZoltu: I mean eipw running against this PR should be warning about EIP-2612 not being final, and I'm surprised it isn't (suggests a bug in eipw).

EIPS/eip-4626.md
MicahZoltu: I strongly suspect that having this in the spec wouldn't have avoided the issue, but if this is in fact non-normative because it is already part of the spec then it *can* be added.

That being said, I strongly recommend against redefining all of EIP-20 in this EIP as that is massively redundant.  No matter what you do, people will make mistakes and I think including redundancy is more likely to lead to mistakes than adding it because it results in this specification getting longer and harder to read.  An implementer who is using an off the shelf ERC-20 implementation would now need to parse out the parts of this EIP that are new and which are part of the ERC-20 implementation they are using, and this could cause them to make a mistake.

fubuloubu: I think the best solution here is to add a note instead to Security that states "per ERC-20, an implementator must ensure that if shares are being redeemed by someone who is not the owner, an allowance check must occur first"

EIPS/eip-4626.md
Pandapip1: Ah, it's showing an error now.

EIPS/eip-4626.md
```diff
@@ -379,10 +379,12 @@ MUST emit the `Withdraw` event.
 
 MUST support a withdraw flow where the shares are burned from `owner` directly where `owner` is `msg.sender`.
 
-MUST support a withdraw flow where the shares are burned from `owner` directly where `msg.sender` has ERC-20 approval over the shares of `owner`.
+MUST support a withdraw flow where the shares are burned from `owner` directly where `msg.sender` has EIP-20 approval over the shares of `owner`.
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `withdraw` execution, and are accounted for during `withdraw`.
 
+MUST check msg.sender can spend owner funds, assets needs to be converted to shares and shares should be checked for allowance.
```
Pandapip1, line 386:
```suggestion
SHOULD check `msg.sender` can spend owner funds, assets needs to be converted to shares and shares should be checked for allowance.
```
I would be okay approving this if this uses should, as vulnerable contracts would still be compliant.

EIPS/eip-4626.md
```diff
@@ -465,10 +467,12 @@ MUST emit the `Withdraw` event.
 
 MUST support a redeem flow where the shares are burned from `owner` directly where `owner` is `msg.sender`.
 
-MUST support a redeem flow where the shares are burned from `owner` directly where `msg.sender` has ERC-20 approval over the shares of `owner`.
+MUST support a redeem flow where the shares are burned from `owner` directly where `msg.sender` has EIP-20 approval over the shares of `owner`.
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `redeem` execution, and are accounted for during `redeem`.
 
+MUST check msg.sender can spend owner funds using allowance.
```
Pandapip1, line 474:
```suggestion
SHOULD check `msg.sender` can spend owner funds using allowance.
```

Pandapip1 (changes requested):

SamWilsn (approved):
Since this PR is adding a `SHOULD` clause, I don't think it affects compliance for older implementations. Changing the external links to `assets/` is probably out of scope for a final EIP.

Pandapip1:
Please close and re-open this PR.

Pandapip1 (approved):

Pandapip1:
This needs approval from from one of @Joeysantoro, @transmissions11, @JetJadeja, @alcueca, or @fubuloubu

EIPS/eip-4626.md
```diff
@@ -1,7 +1,7 @@
 ---
 eip: 4626
-title: Tokenized Vault Standard
-description: A standard for tokenized Vaults with a single underlying ERC-20 token.
+title: Tokenized Vaults
+description: Tokenized Vaults with a single underlying EIP-20 token.
```
fubuloubu, line 4: @Pandapip1 @MicahZoltu this is okay to update title like this after Finalization?

EIPS/eip-4626.md
```diff
@@ -379,10 +379,12 @@ MUST emit the `Withdraw` event.
 
 MUST support a withdraw flow where the shares are burned from `owner` directly where `owner` is `msg.sender`.
 
-MUST support a withdraw flow where the shares are burned from `owner` directly where `msg.sender` has ERC-20 approval over the shares of `owner`.
+MUST support a withdraw flow where the shares are burned from `owner` directly where `msg.sender` has EIP-20 approval over the shares of `owner`.
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `withdraw` execution, and are accounted for during `withdraw`.
 
+SHOULD check `msg.sender` can spend owner funds, assets needs to be converted to shares and shares should be checked for allowance.
```
fubuloubu, line 386: Agree to this addition

EIPS/eip-4626.md
```diff
@@ -465,10 +467,12 @@ MUST emit the `Withdraw` event.
 
 MUST support a redeem flow where the shares are burned from `owner` directly where `owner` is `msg.sender`.
 
-MUST support a redeem flow where the shares are burned from `owner` directly where `msg.sender` has ERC-20 approval over the shares of `owner`.
+MUST support a redeem flow where the shares are burned from `owner` directly where `msg.sender` has EIP-20 approval over the shares of `owner`.
 
 MAY support an additional flow in which the shares are transferred to the Vault contract before the `redeem` execution, and are accounted for during `redeem`.
 
+SHOULD check `msg.sender` can spend owner funds using allowance.
```
fubuloubu, line 474: Agree to this addition

fubuloubu (approved):

EIPS/eip-4626.md
Pandapip1: Yes

Pandapip1:
EIPW errors need fixing... oof

PR #5409, opened 2022-08-04
EIP-1155 Non-Fungible Token extension

wighawag:
## Abstract

This standard is an extension of [EIP-1155](./eip-1155.md). It proposes an additional function, `ownerOf`, which allows EIP-1155 tokens to support Non-Fungibility (unique owners). By implementing this extra function, EIP-1155 tokens can benefit from [EIP-721](./eip-721.md)'s core functionality without implementing the (less efficient) EIP-721 specification in the same contract.

## Motivation

Currently, EIP-1155 does not allow an external caller to detect whether a token is truly unique (can have only one owner) or fungible. This is because EIP-1155 do not expose a mechanism to detect whether a token will have its supply remain to be "1". Furthermore, it does not let an external caller retrieve the owner directly on-chain.

The EIP-1155 specification does mention the use of split id to represent non-fungible tokens, but this requires a pre-established convention that is not part of the standard, and is not as simple as EIP-721's `ownerOf`.

The ability to get the owner of a token enables novel use-cases, including the ability for the owner to associate data with it.

## Specification

The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY" and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.

### Contract Interface

```solidity
interface IERC1155OwnerOf {

    /// @notice Find the owner of an NFT
    /// @dev The zero address indicates that there is no owner: either the token does not exist or it is not an NFT (supply potentially bigger than 1)
    /// @param tokenId The identifier for an NFT
    /// @return The address of the owner of the NFT
    function ownerOf(uint256 tokenId) external view returns (address);
}
```

The `ownerOf(uint256 tokenId)` function MAY be implemented as `pure` or `view`.

The `supportsInterface` method MUST return `true` when called with `0x6352211e`.

## Rationale

`ownerOf` does not throw when a token does not exist (or does not have an owner). This simplifies the handling of such a case. Since it would be a security risk to assume all EIP-721 implementation would throw, it should not break compatibility with contract handling EIP-721 when dealing with this EIP-1155 extension.

## Backwards Compatibility

This EIP is fully backward compatible with EIP-1155.


Pandapip1 (commented):
I feel that this EIP is trying to allow EIP-1155 contracts to be somewhat compatible with EIP-721 contracts, but I'm not convinced by the motivation. Somewhere, it states that EIP-1155 is *more* efficient than EIP-721, but that isn't true. In fact, it is very much the case that EIP-721 is *much* more efficient gas-wise than EIP-1155.

burnto:
This raises an intriguing case when`IERC1155OwnerOf` is implemented but the `ownerOf(tokenId)` does *not* hold a balance of `tokenId` and/or other addresses do hold balances of `tokenId`. Decouples ownership vs balance, which I think holds some promise for novel use cases.

wighawag:
Hmm, i am not sure i follow.
Feel like i should indeed precise that the following is always true when `ownerOf(x) != address(0)`
`balanceOf(ownerOf(x), x) == 1`

wighawag:
> I feel that this EIP is trying to allow EIP-1155 contracts to be somewhat compatible with EIP-721 contracts, but I'm not convinced by the motivation. Somewhere, it states that EIP-1155 is _more_ efficient than EIP-721, but that isn't true. In fact, it is very much the case that EIP-721 is _much_ more efficient gas-wise than EIP-1155.

The discussion is supposed to happen in the ethereum magician thread but i could i guess expand on the motivation.

Erc1155 is definitely cheaper. Amulet by @Arachnid chose erc1155 and added ownerOf for that reason

The reason it is cheaper is that it does not need to track global balance so transfer are cheaper

But there are other reason to want ownerOf, again for efficiency: if you wan to support both nft and ft in the same contract. sandbox.game donit by implementing both eip721 and eip1155 spec in one contract. This add complexity.
With ownerOf in erc1155 we can reduce complexity and gas

burnto:
> Hmm, i am not sure i follow.
> 
> Feel like i should indeed precise that the following is always true when `ownerOf(x) != address(0)`
> 
> `balanceOf(ownerOf(x), x) == 1`

If that constraint is not imposed, then we decouple ownership and balance holding. Owners and balance holders could both transfer tokens. Perhaps interesting for derivative or fractional cases.

If there's a better forum for discussing, happy to jump in there. Thanks.

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+}
+```
+
+The `ownerOf(uint256 tokenId)` function MAY be implemented as `pure` or `view`.
+
+The `supportsInterface` method MUST return `true` when called with `0x6352211e`.
+
+## Rationale
+
+Having the ability to fetch the owner for EIP-1155 tokens that have one, allows us to use them in most of the EIP-721 ecosystem, without the need to implement the whole EIP-721 spec which comes with inefficiencies.
```
SamWilsn, line 49: This should be in the Motivation section. The Rationale section is more about describing technical decisions within the EIP itself.

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+
+`ownerOf` do not throw when a token does not exist (or does not have an owner). This simplifies the handling of such a case. And since it would be a security risk to assume all EIP-721 implementation would throw, it should not break compatibility with contract handling EIP-721 when dealing with this EIP-1155 extension.
+
+## Backwards Compatibility
+
+It is fully backwards compatible with EIP-1155.
+
+As for EIP-721, There is no intention to support EIP-721 here, but the use of the same function `ownerOf` should allow many applications tied to EIP-721 to be compatible with EIP-1155. Not all though.
```
SamWilsn, line 57:
```suggestion
There is no intention to support EIP-721 here, but the use of the same function `ownerOf` should allow many applications tied to EIP-721 to be compatible with EIP-1155, but not all.
```

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY" and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+### Contract Interface
+
+```solidity
+interface IERC1155OwnerOf {
+
+    /// @notice Find the owner of an NFT
+    /// @dev The zero address indicates that there is no owner: either the token do not exist or it is not an NFT (supply potentially bigger than 1)
```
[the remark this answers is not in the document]
Pandapip1: This behavior is inconsistent with EIP-721. I recommend that it revert instead.

wighawag:
> If that constraint is not imposed, then we decouple ownership and balance holding. Owners and balance holders could both transfer tokens. Perhaps interesting for derivative or fractional cases.

I see, I think if we do that we would want a different name that `ownerOf` which already have a meaning in ERC721


> If there's a better forum for discussing, happy to jump in there. Thanks.

The link to the discussion is in the EIP : https://ethereum-magicians.org/t/eip-5409-non-fungible-token-extension-for-eip-1155/10240 but apparently there is more activity on github so I am fine to keep discussing it. we can always make a summary or point to the discussion if the community want to move to ethrreum-magicians forum

EIPS/eip-5409.md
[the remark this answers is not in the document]
wighawag: This is what I thought initialy: let's make it exactly like ERC721 but there are few issues with reverting

In ERC721 a revert happen when a token does not exist. If the token exist it never revert.

With EIP-1155 implementation do not track existence on-chain and so if we do not want to force extra gas use, we would have to revert when a token do not have an owner which could means either that the token do not exist or that it is a fungible token.

As such, it is preferable to not follow ERC721 strictly here and instead return the zero address. Of course, as always an implementation can always revert if it wishes so

Pandapip1:
> If there's a better forum for discussing, happy to jump in there. Thanks.

https://ethereum-magicians.org/t/eip-5409-non-fungible-token-extension-for-eip-1155/10240

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+`ownerOf` does not throw when a token does not exist (or does not have an owner). This simplifies the handling of such a case. Since it would be a security risk to assume all EIP-721 implementation would throw, it should not break compatibility with contract handling EIP-721 when dealing with this EIP-1155 extension.
+
+## Backwards Compatibility
+
+It is fully backwards compatible with EIP-1155.
+
+There is no intention to support EIP-721 here, but the use of the same function `ownerOf` should allow many applications tied to EIP-721 to be compatible with EIP-1155, but not all.
+
+## Reference Implementation
+
```
Pandapip1, line 60:
```suggestion
```

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+## Security Considerations
```
Pandapip1, line 61:
```suggestion
## Security Considerations

Needs discussion.
```

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+
+Having the ability to fetch the owner for EIP-1155 tokens that have one, allows us to use them in most of the EIP-721 ecosystem, without the need to implement the whole EIP-721 spec which comes with inefficiencies.
+
+`ownerOf` does not throw when a token does not exist (or does not have an owner). This simplifies the handling of such a case. Since it would be a security risk to assume all EIP-721 implementation would throw, it should not break compatibility with contract handling EIP-721 when dealing with this EIP-1155 extension.
+
+## Backwards Compatibility
+
+It is fully backwards compatible with EIP-1155.
+
+There is no intention to support EIP-721 here, but the use of the same function `ownerOf` should allow many applications tied to EIP-721 to be compatible with EIP-1155, but not all.
```
Pandapip1, line 57:
```suggestion
This EIP is fully backward compatible with EIP-1155.
```

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+
+The `ownerOf(uint256 tokenId)` function MAY be implemented as `pure` or `view`.
+
+The `supportsInterface` method MUST return `true` when called with `0x6352211e`.
+
+## Rationale
+
+Having the ability to fetch the owner for EIP-1155 tokens that have one, allows us to use them in most of the EIP-721 ecosystem, without the need to implement the whole EIP-721 spec which comes with inefficiencies.
+
+`ownerOf` does not throw when a token does not exist (or does not have an owner). This simplifies the handling of such a case. Since it would be a security risk to assume all EIP-721 implementation would throw, it should not break compatibility with contract handling EIP-721 when dealing with this EIP-1155 extension.
```
Pandapip1, line 51: This is incorrect. EIP-721 specifically states that `ownerOf` a `tokenId` assigned to the zero address or a nonexistent NFT throws.

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+The `ownerOf(uint256 tokenId)` function MAY be implemented as `pure` or `view`.
+
+The `supportsInterface` method MUST return `true` when called with `0x6352211e`.
+
+## Rationale
+
+Having the ability to fetch the owner for EIP-1155 tokens that have one, allows us to use them in most of the EIP-721 ecosystem, without the need to implement the whole EIP-721 spec which comes with inefficiencies.
+
```
Pandapip1, line 50:
```suggestion
```
Move to Motivation

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+## Motivation
+
+Currently, EIP-1155 does not allow an external caller to detect whether a token is truly unique (can have only one owner) or fungible. This is because EIP-1155 do not expose a mechanism to detect whether a token will have its supply remain to be "1". Furthermore, it does not let an external caller retrieve the owner directly on-chain.
+
+The EIP-1155 spec does mention the use of split id to represent non-fungible tokens, but this requires a pre-established convention that is not part of the standard. Plus it would not be as fully featured as the `ownerOf` function also present in the EIP-721 standard.
+
+The ability to get the owner of a token offer great benefit, including the ability for the owner to associate data to it.
```
Pandapip1, line 24:
```suggestion
The ability to get the owner of a token enables novel use-cases, including the ability for the owner to associate data with it.
```

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Abstract
+
+This standard is an extension of [EIP-1155](./eip-1155.md). It proposes an additional function, `ownerOf` to allow EIP-1155 to support Non-Fungible Tokens (tokens that have a unique owner). By implementing this extra function, EIP-1155 can benefit from [EIP-721](./eip-721.md) core functionality without implementing the full (and less efficient) EIP-721 spec in the same contract. In particular, this extension allows EIP-1155 tokens of supply one to have an owner and have it exposed.
+
+## Motivation
+
+Currently, EIP-1155 does not allow an external caller to detect whether a token is truly unique (can have only one owner) or fungible. This is because EIP-1155 do not expose a mechanism to detect whether a token will have its supply remain to be "1". Furthermore, it does not let an external caller retrieve the owner directly on-chain.
+
+The EIP-1155 spec does mention the use of split id to represent non-fungible tokens, but this requires a pre-established convention that is not part of the standard. Plus it would not be as fully featured as the `ownerOf` function also present in the EIP-721 standard.
```
Pandapip1, line 22:
```suggestion
The EIP-1155 specification does mention the use of split id to represent non-fungible tokens, but this requires a pre-established convention that is not part of the standard, and is not as simple as EIP-721's `ownerOf`.
```

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+eip: 5409
+title: EIP-1155 Non-Fungible Token extension
+description: Allow EIP-1155 to represent Non-Fungible Tokens (tokens who have a unique owner)
+author: Ronan Sandford (@wighawag)
+discussions-to: https://ethereum-magicians.org/t/eip-5409-non-fungible-token-extension-for-eip-1155/10240
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-07-23
+requires: 165, 1155
```
Pandapip1, line 11:
```suggestion
requires: 165, 721, 1155
```

EIPS/eip-5409.md
```diff
@@ -0,0 +1,65 @@
...
+---
+
+## Abstract
+
+This standard is an extension of [EIP-1155](./eip-1155.md). It proposes an additional function, `ownerOf` to allow EIP-1155 to support Non-Fungible Tokens (tokens that have a unique owner). By implementing this extra function, EIP-1155 can benefit from [EIP-721](./eip-721.md) core functionality without implementing the full (and less efficient) EIP-721 spec in the same contract. In particular, this extension allows EIP-1155 tokens of supply one to have an owner and have it exposed.
```
Pandapip1, line 16:
```suggestion
This standard is an extension of [EIP-1155](./eip-1155.md). It proposes an additional function, `ownerOf`, which allows EIP-1155 tokens to support Non-Fungibility (unique owners). By implementing this extra function, EIP-1155 tokens can benefit from [EIP-721](./eip-721.md)'s core functionality without implementing the (less efficient) EIP-721 specification in the same contract.
```

Pandapip1 (changes requested):

EIPS/eip-5409.md
wighawag: Here I describe the behavior of `ownerOf` in EIP-5409, not the behavior in EIP-721

EIPS/eip-5409.md
wighawag: hmm, not sure it does
EIP-5409 is only softly attached to EIP-721, it does not attempt to be 100% compatible with it, the only function from EIP-721 it implements is `ownerOf` but it does not even attempt to be 100% compatible on this single function either (See discussion around throwing)

wighawag:
Thanks @Pandapip1 for the review, feedback and suggestion!

EIPS/eip-5409.md
wighawag: would this not make the bot complain ?

Pandapip1: Nope. Reference implementation is optional.

EIPS/eip-5409.md
Pandapip1: I know. It references EIP-721, though, so it belongs in the requires header.

EIPS/eip-5409.md
Pandapip1: > Since it would be a security risk to assume all EIP-721 implementation would throw

It is not a security risk, as all EIP-721 implementations MUST throw.

I therefore also propose that `ownerOf` throw too. Small things like these have resulted in major annoyances down the line.

Pandapip1 (approved):
I would rather `ownerOf` throw, but this EIP does meet the requirements for a draft. Merging.

EIPS/eip-5409.md
wighawag: My point is that a smart contract that intends to support ERC721 generically would take a risk when they assume all ERC721 implementation throw when `ownerOf` is called with a non-existent token (There is a chance some invalid implementation exists maliciously or not). 

One reason for not throwing is that it makes checking for existence cumbersome as you have to catch the error.
But to be honest, I am not too attached to not throwing if that helps. Do you have links to examples of the "small things" you mentioned?

Note that If we go we decide to throw, it must throw in both the following situations:
1. the token is a fungible token (it has no owner) 
2. or the token does not exist

Pandapip1: Having to catch it is a *good* thing. Exceptions are *good.* It's when there's not obviously a problem and your program misbehaves in a subtle but extremely problematic way that major issues arise.

wighawag: Well i never said throw is bad as a general statement. It all depends on the context.

In this particular one, i gave my reasoning

PR #5419, opened 2022-08-05
EIP-2200: Change unordered list to ordered

chfast:
In the specification of EIP-2200 change the format of the list from
unordered to order. This way a particular rule can be referenced from
outside by e.g. "4.1.2".

### DIFF
--- a/EIPS/eip-2200.md
+++ b/EIPS/eip-2200.md
@@ -19,7 +19,7 @@ to make it interoperable with other gas changes such as [EIP-1884].
 
 This EIP provides a structured definition of net gas metering changes
 for `SSTORE` opcode, enabling new usages for contract storage, and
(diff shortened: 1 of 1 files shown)


Pandapip1 (changes requested):
Fix the EIPW errors

chfast:
> Fix the EIPW errors

Not my errors.

Pandapip1:
They need to be fixed in order for this PR to be merged.

SamWilsn:
Since this EIP is final, I think we should avoid changes.

chfast:
I think you are right, people should be able to count to 3 in head. https://github.com/ethereum/evmc/pull/661

chfast:
Here is EVMC's copy of the specification with clauses numbered. https://evmc.ethereum.org/storagestatus.html

PR #5439, opened 2022-08-10
Simplify and Fix Auto Review Bot

Pandapip1:
Collapses three jobs into a single job, somewhat simplifying the file.


Pandapip1:
Successfully E2E tested. CC @SamWilsn

SamWilsn (commented):
If the steps depend on different triggers, why put them all in the same file?

Pandapip1:
> If the steps depend on different triggers, why put them all in the same file?

Firstly, because they all do effectively the same thing: fetch the PR number and add it as an asset. Secondly, it would vastly overcomplicate the other half of the bot.

PR #5440, opened 2022-08-10
Add Name to Label Bot

Pandapip1:
All this does is add a name to the only job that for some reason doesn't have one.

### DIFF
--- a/.github/workflows/auto-label-bot.yml
+++ b/.github/workflows/auto-label-bot.yml
@@ -1,9 +1,10 @@
 on:
   pull_request_target:
 
+name: Auto Label Bot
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-label-bot.yml
```diff
@@ -1,9 +1,10 @@
 on:
   pull_request_target:
 
+name: Auto Label Bot
 jobs:
   jekyll-label-action:
-    name: Automatic Label Bot
```
Pnoyako19, line 6: Sha256

Pnoyako19 (approved):

PR #5443, opened 2022-08-10
EIP-4881: DepositTreeSnapshot Refactor

ethDreamer:
Refactor the `DepositTreeSnapshot`:
* rename `deposits` to `deposit_count` to match `Eth1Data`
* include the `execution_block_height` for convenience to consensus client implementors
* include the `deposit_root` for easy self-validation of the snapshot
* specify the maximum length of the `finalized` array
* added in some cheap extra error checking


ethDreamer:
@lightclient @Pandapip1 @SamWilsn
I really would like to keep the links in this EIP. All of them already existed, and they only link to:
 * a specific version of ethereum/consensus-specs so it is guaranteed to always exist
 * the discussion on eth-magicians - the bot doesn't complain about the first link, just the second link to it
 
I'm open to removing the second link to the discussion thread but I think EIPs should absolutely be able to reference the ethereum spec.

skaille:
Чт, 11 авг. 2022 г. в 20:12, ethDreamer ***@***.***>:

lightclient:
I'm okay with allow listing `github.com/ethereum/consensus-specs/*` as an acceptable external link. I'm ambivalent about `ethereum-magicians.org` - I think I lean towards not allowing it in the body of the EIP because it is easy to then lean on it by saying "look at this here" rather that having the EIP explain the situation succinctly.

ethDreamer:
> I'm okay with allow listing `github.com/ethereum/consensus-specs/*` as an acceptable external link. I'm ambivalent about `ethereum-magicians.org` - I think I lean towards not allowing it in the body of the EIP because it is easy to then lean on it by saying "look at this here" rather that having the EIP explain the situation succinctly.

okay I've removed the second link to the discussion :) What is needed now to merge this?

EIPS/eip-4881.md
```diff
@@ -20,15 +20,15 @@ Most client implementations require beacon nodes to download and store every dep
 Consensus clients MAY continue to implement the deposit Merkle tree however they choose. However, when transmitting the tree to newly syncing nodes, clients MUST use the following format:
 
 ```
```
xinbenlv, line 22: please specify programing language

EIPS/eip-4881.md
```diff
@@ -20,15 +20,15 @@ Most client implementations require beacon nodes to download and store every dep
...
+    deposit_root: Hash32
+    deposit_count: uint64
+    execution_block_hash: Hash32
+    execution_block_height: uint64
 ```
 
-Where the hashes in the `finalized` vector are defined in the [Deposit Finalization Flow](#deposit-finalization-flow) section below, `deposits` is the number of deposits stored in the snapshot, and `execution_block_hash` is the hash of the execution block containing the highest index deposit stored in the snapshot. Consensus clients MUST make this structure available via the Beacon Node API endpoint:
-
+Where `finalized` is a variable-length list (of maximum size `DEPOSIT_CONTRACT_DEPTH`) containing the hashes defined in the [Deposit Finalization Flow](#deposit-finalization-flow) section below. The fields `deposit_root`, `deposit_count`, and `execution_block_hash` store the same information as the [`Eth1Data`](https://github.com/ethereum/consensus-specs/blob/v1.1.9/specs/phase0/beacon-chain.md#eth1data) object that corresponds to the snapshot, and `execution_block_height` is the height of the execution block with hash `execution_block_hash`. Consensus clients MUST make this structure available via the Beacon Node API endpoint:
```
xinbenlv, line 31: I think this link shall be allowed.

Pandapip1: CC @SamWilsn

I am mildly against including this at this time, as this contradicts EIP-1. @xinbenlv if you would like to add a change to EIP-1 allowing links to commits of the consensus specs in a new PR, go ahead, as it will likely prompt discussion.

ethDreamer: @Pandapip1 
There are 213 other EIPs that link to `https://github.com/ethereum`, one of which is this one as it currently exists. This PR doesn't modify those links at all.

Is it really necessary that my update to this EIP, which is needed for people who are currently implementing it, should wait for this discussion to play out?

Pandapip1: Yes, it is absolutely necessary. Those EIPs were made when the external links policy was more relaxed.

Sorry for the hassle. The link does need to be removed, though.

lightclient: EIP-1 states that links to external resources **SHOULD NOT** be included. Therefore it is not "absolutely necessary" for the links to be removed. And as I said above, I would like `ethereum/consensus-specs` added to the allow list. It makes no sense to not allow EIP authors to link to `ethereum/consensus-specs`. The beacon chain _is_ Ethereum and we shouldn't be asking authors to needlessly duplicate spec code in their EIP.

poojaranjan: image

Looking into errors, it appears to me that EIPW checked for non-relative links and thus found error with specs link and missed ref of EIP-4444 (because it is added as relative link in the proposal). 

image

IMO, specs link are not "external" as they are part of the same GitHub organization. Probably "external resources" could be defined in EIP-1.

Pandapip1: I am okay with letting this one link slide for now.

Pandapip1:
@SamWilsn please manually merge.

EIPS/eip-4881.md
timbeiko: > And as I said above, I would like ethereum/consensus-specs added to the allow list. It makes no sense to not allow EIP authors to link to ethereum/consensus-specs. The beacon chain is Ethereum and we shouldn't be asking authors to needlessly duplicate spec code in their EIP.

+1 to this.

ethDreamer:
@SamWilsn can you please merge this?

SamWilsn (changes requested):
Tags aren't technically immutable in git. Would you mind switching to commit links?

---

Since we don't have an established precedent for allowing or not allowing links to the consensus spec, I'll open a discussion on the ECH discord to see if there are any objections.

ethDreamer:
> Tags aren't technically immutable in git. Would you mind switching to commit links?
> 

done!

xinbenlv:
@Pandapip1 @SamWilsn it seems the code change has changed to use commit per your suggestion. The lint error is mostly about "external links". per our conversion `ethereum/consensu-spec` will be allowed.  To unblock @ethDreamer ,  anything else do you think it should fix before a manual merge?

PR #5448, opened 2022-08-12
Fix the waiting on EIP number label again

Pandapip1:
@SamWilsn

Yea. Turns out that files that aren't EIPs don't have valid EIP numbers. Oops.


config/.jekyll-labels.yml
```diff
@@ -5,7 +5,7 @@
 "type: Informational": this?.new?.type == "Informational"
 "type: Meta": this?.new?.type == "Meta"
 "type: EIP1 (Process)": this?.old?.status == "Living" || this?.old?.eip == "<to be assigned>"
-"waiting: EIP number": "!(this?.new?.eip > 0)"
+"waiting: EIP number": isNaN(this?.new?.eip)
```
Pandapip1, line 8: Relevant: we *want* null to return false.

SamWilsn (approved):

PR #5456, opened 2022-08-14
EIP-777: pngcrush assets

wchargin:
These simple PNG images were bafflingly big: like, 16.8 MB for a
single-color 2048p image. This patch shrinks the total repository size
from 152 MB to 45 MB.

Generated with:

    for f in assets/eip-777/logo/png/*.png; do pngcrush -ow "$f"; done

using pngcrush 1.8.13.

Passing each file through `convert - rgba:-` before and after this patch
shows that the raw pixel data is unchanged.

(The no-op change EIP-777 itself is to appease the automerge bot.)

wchargin-branch: eip-777-pngcrush


Pandapip1 (approved):

Pandapip1:
@SamWilsn I request a manual merge.

SamWilsn:
Unless you're doing a shallow clone, this won't change the size of the repository, right? What's the point of doing this without re-writing history?

Pandapip1:
> Unless you're doing a shallow clone, this won't change the size of the repository, right? What's the point of doing this without re-writing history?

Oftentimes people do use shallow clones.

SamWilsn (approved):

wchargin:
> Unless you're doing a shallow clone, this won't change the size of the repository, right?

Correct. (I guess to be more precise I might have said that it reduces
the total size of the *tree* rather than the repository.)

> What's the point of doing this without re-writing history?

Aside from shallow clones (which I'll use if cloning a repository is
taking a while and I just need to write a quick patch), the EIP-777 text
directs readers to use the assets in this directory. So, to the extent
that anyone does this, it seems nice to have reasonably sized images.

In any case, thanks for merging!

PR #5457, opened 2022-08-14
Update EIP-712: Fix `eth_signTypedData` definition

wchargin:
The previous EIP text for `eth_signTypedData` seemed to describe the
existing "Ethereum signed message" flows, without any relation to typed
data. The new definition is consistent with what's actually implemented
by (e.g.) [the Ethers signer][1] and [the OpenZeppelin validator][2], as
well as with the start of the "Specification" section of this EIP.

[1]: https://github.com/ethers-io/ethers.js/blob/ce8f1e4015c0f27bf178238770b1325136e3351a/packages/hash/src.ts/typed-data.ts#L392-L398
[2]: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/1b27c13096d6e4389d62e7b0766a1db53fbb3f1b/contracts/utils/cryptography/ECDSA.sol#L216-L218

wchargin-branch: eip-712-eth-signtypeddata


Pandapip1:
This makes a change to a final EIP. The implementations will need to be changed.

wchargin:
> The implementations will need to be changed.

I believe that this patch updates the text to match the implementations
(and the intent), not the other way around. Am I misunderstanding?

In addition to Ethers and OpenZeppelin (as listed above), other
prominent implementations that are consistent with this patch include
OpenSea's [Seaport][1], [Uniswap V3's position NFTs][2], and [MetaMask's
signing libraries][3]. Are there actually implementations that would
need to be updated here?

[1]: https://github.com/ProjectOpenSea/seaport/blob/c30dd90272609677606f03f9c885466f15e891eb/reference/lib/ReferenceGettersAndDerivers.sol#L147-L162
[2]: https://github.com/Uniswap/v3-periphery/blob/9ca9575d09b0b8d985cc4d9a0f689f7a4470ecb7/contracts/base/ERC721Permit.sol#L65-L72
[3]: https://github.com/MetaMask/eth-sig-util/blob/cde38f7aa3f8e8d0a7ee75730f363325b8e1924c/src/sign-typed-data.ts#L375

> This makes a change to a final EIP.

(For what it's worth, [I also did raise this before the EIP was
finalized][em], but received no response. Which is okay, no worries, but
it seems better to fix late than never if the fix is indeed correct.)

[em]: https://ethereum-magicians.org/t/eip-712-eth-signtypeddata-as-a-standard-for-machine-verifiable-and-human-readable-typed-data-signing/397/30

Pandapip1:
> I believe that this patch updates the text to match the implementations

Yes, but that isn't allowed to a final EIP.

xinbenlv:
I was agains the 712 to go final. but since it is now `Final`, unless we make a rule update, this definition change is non-compatible with the final version and hence I don't think it shall be approved. It seems the only way to make this improvement is to start a competing EIP that defines a different sign typed data

Pandapip1:
I feel like I should make a bot that adds a comment on every PR that moves an EIP to final that **the entire EIP *must* be proofread**. It seems obvious to me, but here we are...

xinbenlv:
I think pure editorial change that improves readability shall be allowed, but breaking change of the specificiation are not allowed. if an EIP intends to be a standard/protocol, the behaviors mandated by spec shall not change.

decentralion:
My understanding is that as written, the EIP is literally incorrect and includes inaccurate information, and that this pull request fixes that, so that the EIP will correctly describe the real behavior of all compliant implementations. Can we agree that the top priority should be that the EIP is an accurate reference?

Maybe original authors (@recmo @LogvinovLeon @dekz) can comment?

xinbenlv:
Let's let original authors respond. I think the more accurate way to say is current 712 snapshot have un-eliminated ambiguity and ethers and open-zepplin has adopted non-standard approach.

Pandapip1:
I guess one could also do the "SHOULD" and "SHOULD NOT" trick to correct EIPs.

xinbenlv:
> I guess one could also do the "SHOULD" and "SHOULD NOT" trick to correct EIPs.
> 
> ```diff
> - MUST do X
> + SHOULD do Y
> + SHOULD NOT do X
> ```
> 
> _Technically_ this isn't backwards incompatible.

I believe it would be backward incompatible. 


Suppose the following change was proposal for the *final* EIP-X:

```diff
- MUST implement `iterate(uint)`
+ SHOULD implement `iterate(uint)`
```

A client assuming the smart contract is EIP-X final version will assume `iterate(uint)` interface to exist. But after the change, the `iterate(uint)` may not exist in certain case and therefore client would have to deal with such scenarios .

per https://www.rfc-editor.org/rfc/rfc2119
```
This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to *ignore a
   particular item*, but the full implications must be understood and
   carefully weighed before choosing a different course.
```

Pandapip1:
I guess I didn't think about that particular case. You are correct there.

As written, however, this doesn't make any sense since it references a nonexistent variable. Therefore, since there are no applications that *can* correctly parse this, there is no backward compatibility issue there since there is nothing that was previously compatible.

xinbenlv:
> I guess I didn't think about that particular case. You are correct there.
> 
> As written, however, this doesn't make any sense since it references a nonexistent variable. Therefore, since there are no applications that _can_ correctly parse this, there is no backward compatibility issue there since there is nothing that was previously compatible.

I don't know if we are referring to the same aspect of this PR. I am referring to this part

```diff
- The sign method calculates an Ethereum specific signature with: `sign(keccak256("\x19Ethereum Signed Message:\n" + len(message) + message)))`.
+ The sign method calculates an Ethereum specific signature with: `sign(keccak256("\x19\x01" + domainSeparator + structHash))`, as defined in the "Specification" section above.
```

The particular `"\x19Ethereum Signed Message:\n" + len(message)` was removed and replaced by `domainSeparator`.
I think one could potentially implement the former version that has `Ethereum Signed Message` as prefix.

When you say "nonexistent variable", which nonexistent variable do you refer to?

Pandapip1:
`message` is not defined.

xinbenlv:
Here is what caused me to think `message` is defined:

In the spec section:
- https://github.com/ethereum/EIPs/blob/6fe3596850403b3cc1257522de24fa426c607027/EIPS/eip-712.md?plain=1#L82 
- https://github.com/ethereum/EIPs/blob/6fe3596850403b3cc1257522de24fa426c607027/EIPS/eip-712.md?plain=1#L87
both have mentioned `message`, assuming to be the message being signed. 

If the rationale for accepting this PR is due to `message` not defined, with same argument requires to update L82 and L87 too, I guess?  I'd lean towards leaving the author to decide.

Pandapip1:
I agree that the first line defines `message`. The second line doesn't. That's still a good enough indication that `message` is defined.

As @xinbenlv mentioned in discord, there appear to be *two* specification sections?

EIPS/eip-712.md
```diff
@@ -165,9 +165,7 @@ The method `eth_signTypedData` is added to the Ethereum JSON-RPC. The method par
 
 #### eth_signTypedData
 
-The sign method calculates an Ethereum specific signature with: `sign(keccak256("\x19Ethereum Signed Message:\n" + len(message) + message)))`.
-
-By adding a prefix to the message makes the calculated signature recognisable as an Ethereum specific signature. This prevents misuse where a malicious DApp can sign arbitrary data (e.g. transaction) and use the signature to impersonate the victim.
+The sign method calculates an Ethereum specific signature with: `sign(keccak256("\x19\x01" + domainSeparator + structHash))`, as defined in the "Specification" section above.
```
SamWilsn, line 168: The rest of the document uses `‖` for concatenation:

```suggestion
The sign method calculates an Ethereum specific signature with: `sign(keccak256("\x19\x01" ‖ domainSeparator ‖ structHash))`, as defined above.
```

SamWilsn (commented):
This is pretty clearly a copy-paste error from [`eth_sign`'s API documentation](https://github.com/ethereum/wiki/wiki/JSON-RPC/e8e0771b9f3677693649d945956bc60e886ceb2b#eth_sign). No one has implemented `eth_signTypedData` as an alias for `eth_sign`.

wchargin:
@SamWilsn: Exactly; thank you. Switched to `‖`, and also changed
`structHash` to `hashStruct(message)` for consistency with other parts
of the document.

Pandapip1:
I would prefer if you also removed the first Specification section.

Pandapip1 (dismissed):

SamWilsn (approved):
After [some discussion](https://discord.com/channels/916516304972824576/916713912970412103/1009300460810809416) among editors, we're going to approve this.

If I understand the CI process, and I'm not sure I do, this will still require author approval.

SamWilsn:
> I would prefer if you also removed the first Specification section.

Like the whole section, or just the header?

Pandapip1:
> Like the whole section, or just the header?

The whole first specification section, unless I misread it, is redundant.

xinbenlv (changes requested):
As much as I appreciate your PR - I think it's greatly valueable and critical to fix this incorrect EIP.

I'd object to merging this PR without changing the status from "FINAL" to a non_FINAL

If I may propose, I'd love to suggest you @wchargin, to also consider start a new EIP, duplicating EIP-712, patching your proposed change and pursue that new EIP as an updated EIP to go-in. I'd be in full support for this way of fixing EIP-712.

Note: I am just also a peer of you @wchargin doing editorial helpout on the side. I am not an official Editor so I'm just voicing my concern. I'd respect the editor group's consensus.

wchargin:
No, sorry, I won't be starting a competing EIP with this fix because
that is preposterous. This is obviously a simple copy-and-paste error.
The text is inconsistent with itself, with all implementations, and with
the clear intent of the EIP. The consequences of starting a new
"EIP-713" with this fix and trying to get everyone to "migrate" to that
are so much worse than the alternatives that it makes no sense to
seriously discuss them. The only advantage to that plan is that it
follows the process for process's sake, sticks to consistency for
consistency's sake, without ever taking a step back to consider the
*purposes* of that process and the *impact* of the various options.
I won't be party to that.

@xinbenlv: Thanks for clarifying that you're not an official editor.
If the EIP-712 editors want to merge this patch, great. If the editors
would prefer to abandon it and leave the EIP inconsistent, well, it'll
be a bit confusing, but no one will die. I don't have the energy or
inclination to debate this endlessly, but I'm happy to help shepherd
this through if there's anything needed from me, and happy to close it
if that's the will of the council. Cheers.

Pandapip1:
It's still not clear what the consensus among editors is. CC @gcolvin @axic @lightclient

SamWilsn:
> It's still not clear what the consensus among editors is.

Consensus was reached and documented in [this thread](https://discord.com/channels/916516304972824576/916713912970412103/1009300460810809416).

We aren't waiting on editors to approve, but rather one of @Recmo, @LogvinovLeon, or @dekz.

fulldecent:
This is a very important EIP. I would like to see an erratta section at least recognizing this change, if not explaining it.

minebuu:
I still don't know why these critical errors haven't been fixed. EIP712 is a very important document and can resolve ambiguity with many examples already implemented, but it is very confusing for noobs.

I hope to see that change soon.

Pandapip1:
CC @recmo, @LogvinovLeon, @dekz

Pandapip1:
Waiting on authors

Pandapip1:
Waiting on authors

Pandapip1:
@SamWilsn @lightclient @axic @gcolvin are we okay with manually merging this?

Pandapip1:
Dismiss stale bot.

Pandapip1:
Dismissing stale bot.

Pandapip1:
@eth-bot rerun

PR #5459, opened 2022-08-14
EIP-5192: add event LockingStatusChanged(uint256 tokenId, bool status)

0xanders:
Add `event LockingStatusChanged(uint256 tokenId, bool status)`

So the locking status of the NFT could be updated  off-chain.

### DIFF
--- a/EIPS/eip-5192.md
+++ b/EIPS/eip-5192.md
@@ -2,7 +2,7 @@
 eip: 5192
 title: Minimal Soulbound NFTs
 description: Minimal interface for soulbinding EIP-721 NFTs
-author: Tim Daubenschütz (@TimDaub)
+author: Tim Daubenschütz (@TimDaub), Anders (@0xanders)
 discussions-to: https://ethereum-magicians.org/t/eip-5192-minimal-soulbound-nfts/9814
 status: Review
 type: Standards Track
@@ -34,6 +34,9 @@ A token with a `uint256 tokenId` may be inseparably-nound to a receiving account
 pragma solidity ^0.8.0;
 
 interface IERC5192 {
+  /// @notice Emitted when locking status is changed
+  event UpdateLockingStatus(uint256 tokenId, bool status);
+
   /// @notice Returns the locking status of an Soulbound Token
   /// @dev SBTs assigned to zero address are considered invalid, and queries
   /// about them do throw.
--- a/EIPS/eip-5192.md
+++ b/EIPS/eip-5192.md
@@ -35,7 +35,7 @@ pragma solidity ^0.8.0;
 
 interface IERC5192 {
   /// @notice Emitted when locking status is changed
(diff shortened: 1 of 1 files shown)


EIPS/eip-5192.md
```diff
@@ -34,6 +34,9 @@ A token with a `uint256 tokenId` may be inseparably-nound to a receiving account
 pragma solidity ^0.8.0;
 
 interface IERC5192 {
+  /// @notice Emitted when locking status is changed
+  event UpdateLockingStatus(uint256 tokenId, bool status);
```
xinbenlv, line 38: [Optional] suggest 
```suggestion
  event LockinStatusChanged(uint256 tokenId, bool status);
```

xinbenlv (approved):
Reasonable updates: add an event interface, adding an author

xinbenlv:
Typo :p

EIPS/eip-5192.md
```diff
@@ -34,6 +34,9 @@ A token with a `uint256 tokenId` may be inseparably-nound to a receiving account
 pragma solidity ^0.8.0;
 
 interface IERC5192 {
+  /// @notice Emitted when locking status is changed
```
TimDaub, line 37: I think we should add an `@dev` note on what should happen upon minting. E.g. when minting "a regular NFT", what should `bool status` be?

EIPS/eip-5192.md
```diff
@@ -34,6 +34,9 @@ A token with a `uint256 tokenId` may be inseparably-nound to a receiving account
...
+  event LockingStatusChanged(uint256 tokenId, bool status);
```
TimDaub, line 38: I'm unsure if I like the readability of `LockingStatusChanged(..., bool status)`.

Naturally, what does it mean `Locking Status Changed to false`? If we reflect upon it actively, then we can probably understand that a lock is locked on `true` and open on `false.`

What about:

`event Locked(uint256 tokenId, bool state)` where if `state=true` then the token is `Locked` and if `state=false` the token is NOT `Locked`.

TimDaub (changes requested):
left few more comments

EIPS/eip-5192.md
0xanders: How about these two events:  Locked and Unlocked

/// @notice Emitted when the locking status is changed to locked
/// @param tokenId The identifier for a token.
event Locked(uint256 tokenId);

/// @notice Emitted when the locking status is changed to unlocked
/// @param tokenId The identifier for a token.
event Unlocked(uint256 tokenId);

@TimDaub

TimDaub: ok. But also we must decide what should be done if e.g. a token is minted. Must a `Unlocked(tokenId)` event be emitted? Whatever we decide on, let's please also add it to the natspec using `/// @dev`.

PR #5493, opened 2022-08-18
Rerun CI on rerun

Pandapip1:


.github/workflows/ci-rerun-trigger.yml
```diff
@@ -0,0 +1,33 @@
...
+              repo: context.repo.repo,
+              pull_number: context.payload.issue_comment.issue.number,
+              body: '[RETRIGGER]\n' + pr.data.body
+            });
+            await github.rest.pulls.update({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.issue_comment.issue.number,
+              body: pr.data.body
+            });
```
SamWilsn, line 33: So this re-runs the workflow by editing the body of the PR?

Is there some way to use [`workflow_dispatch`](https://octokit.github.io/rest.js/v18#actions-create-workflow-dispatch) for this? Or perhaps [re-running a check](https://docs.github.com/en/rest/checks/runs#rerequest-a-check-run)?

If you've already investigated these options, sorry!

Pandapip1: Re-running a check and workflow_dispatch requires write permission to use, which you have stated has too many permissions.

SamWilsn: It looks like something like this might work, if you can find the right run id:

```javascript
await github.request('POST /repos/{owner}/{repo}/actions/runs/{run_id}/rerun', {
  owner: context.repo.owner,
  repo: context.repo.repo,
  run_id: '...'
});
```

I tried it with an account without write permissions on my repo, so maybe `issue_comment` gets elevated permissions?

Pandapip1: Sorry, I should have specified: I thought you were talking about the editors clicking the button.

I'll see what I can do there.

Pandapip1: So that would be ideal, but GitHub doesn't have an API for fetching check runs!

SamWilsn: That's ridiculous...

Can we open/close the PR instead of editing it?

Pandapip1: Yes... but that pollutes history.

The reason I suggest editing the body is that it's pretty much invisible.

SamWilsn: Makes sense!

SamWilsn (approved):

PR #5495, opened 2022-08-18
Update EIP-1: Explicitly state that backward compatible errata are allowed for final EIPs

Pandapip1:


EIPS/eip-1.md
```diff
@@ -82,7 +82,7 @@ The following is the standardization process for all EIPs in all tracks:
 
 If this period results in necessary normative changes it will revert the EIP to `Review`.
 
-**Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
+**Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata, add non-normative clarifications, or make backward compatible changes.
```
SamWilsn, line 85: I would prefer that backward compatible changes be made in new EIPs. Authors shouldn't be adding new features (that are backwards compatible) to their Final proposals without going through the review process.

Pandapip1: Fixed.

xinbenlv:
At the very least it's hard to determine whether something is backward compatible.
I've love to voice my opposition to this change without boarder EIP Editor group consensus.
Could I suggest you instead start an `informational EIP` that gives examples of what's considered backward compatible or not?

Pandapip1:
Backward compatibility has a very simple definition: if a particular behavior has a valid unambiguous definition, then it can't be changed.

xinbenlv:
Most often I'd envision a change like "I want to add this method (but optional) to the interface" 

Do we count this type of change as "backward compatible"?

Pandapip1:
> Do we count this type of change as "backward compatible"?

That's a really good point. I would say that it wouldn't be backward compatible in the majority of the cases since it's pretty standard to use EIP-165. Not sure about other cases - I'll have to revise the definition.

Pandapip1:
How about:

If a particular behavior has a satisfiable unambiguous definition, then it can't be changed. If a particular behavior has no definition, then one can't be added. If a particular behavior has a definition that is either unsatisfiable or ambiguous, then it may be changed.

EIPS/eip-1.md
```diff
@@ -82,7 +82,7 @@ The following is the standardization process for all EIPs in all tracks:
 
 If this period results in necessary normative changes it will revert the EIP to `Review`.
 
-**Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
+**Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata in a backward compatible manner, or add non-normative clarifications.
```
SamWilsn, line 85: Personally I prefer the more permissive language we had before. With this new wording, the fix for EIP-712 wouldn't have been allowed.

Perhaps this is something we add to [EIP-5069](https://eips.ethereum.org/EIPS/eip-5069) instead?

xinbenlv: I personally concern this will break "FINAL"-ness. If we want to allow authors of an EIP have options to update, I am personally in favor of we change "LIVE" status to be allowed for ERC instead of changing requiurement of "FINAL"

Pandapip1: > With this new wording, the fix for EIP-712 wouldn't have been allowed.

With my definition, this would have been allowed. The sign typed data RPC was defined in such as to be unsatisfiable.

gcolvin: The only normative changes should be errata, which make the document say what it should have said in the first place. Or in the (we hope) rare case that we need to make it say what was actually implemented, in spite of what it said.

Pandapip1:
Dismissing stale bot.

gcolvin (approved):

Pandapip1:
Dismissing stale bot

PR #5521, opened 2022-08-22
EIP-5521:  Referable NFT (rNFT)

OniReimu:
This standard proposes an extension to ERC-721 Non-Fungible Tokens (NFTs) to construct reference relationships among NFTs.

When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #5523, opened 2022-08-23
CI: Add "shiping" to CodeSpell whitelist

Pandapip1:
See https://github.com/ethereum/EIPs/pull/5521


Pandapip1:
(It's the name of somebody)

.codespell-whitelist
```diff
@@ -12,3 +12,4 @@ acn
 ende
 sting
 complies
+shiping
```
xinbenlv, line 15: I am unsure why `shiping` to whitelist....

abcoathup:
Shiping Chen is an author on EIP5521.
Is there a way to skip the author section rather than having to allow list everywhere?

Pandapip1:
> Is there a way to skip the author section rather than having to allow list everywhere?

Unfortunately not. I wish this was the case though.

xinbenlv:
Without alternatives, I am in favor of checking in this commit to allow list Shiping's name

SamWilsn (approved):

PR #5463, opened 2022-08-15
Add adoptable field

Pandapip1:
See the EIP-1 changes for description.

Changes needed:
- [x] EIP template change
- [x] EIP-1 description
- [ ] EIP-Bot implementation
- [ ] eipw validation

Closes https://github.com/ethereum/EIPs/issues/5245

PR #5533, opened 2022-08-24
Update EIP-1: Add adoptable field

SamWilsn:
Reopening the discussion from #5463.

I merged without waiting for sufficient consensus, so I've reverted that PR, and opened this one.


Pandapip1 (dismissed):

Pandapip1:
@eth-bot rerun

lightclient:
I don't think this provides much value. It should simply be left up to editor discretion if a stagnant EIP can be reassigned. And I generally don't think that should happen unless the original author agrees.

Pandapip1:
> I generally don't think that should happen unless the original author agrees.

This flag is basically the author in advance stating "Yes, I am okay with someone else taking over this proposal."

Pandapip1:
Also - this flag is optional. Anyone can choose to not have their EIP adoptable.

SamWilsn:
I am, as usual, torn on this proposal.

I dislike reassigning proposals to new authors, but I equally dislike abandoning a well-known EIP number just because the authors can't be reached.

I would feel more comfortable if the authors opt-in to the adoption process over just reassigning without their permission.

Pandapip1:
> I would feel more comfortable if the authors opt-in to the adoption process over just reassigning without their permission.

Should the adoptable field be by default set to false then?

Pandapip1:
@SamWilsn @lightclient is this better?

SamWilsn:
> > I would feel more comfortable if the authors opt-in to the adoption process over just reassigning without their permission.
> 
> Should the adoptable field be by default set to false then?

IMO, if the `adoptable` field is missing, the EIP should not be reassigned without permission but the EIP template should have `adoptable: true`.

Pandapip1:
> IMO, if the adoptable field is missing, the EIP should not be reassigned without permission but the EIP template should have adoptable: true.

I agree with this. This was in fact my original idea. I'm not sure how it ended up being an "if it's missing then it defaults to true" situation, but it's fixed now.

eip-template.md
```diff
@@ -9,6 +9,7 @@ type: <Standards Track, Meta, or Informational>
 category (*only required for Standards Track): <Core, Networking, Interface, or ERC>
 created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
 requires (*optional): <EIP number(s)>
+adoptable: true  # It is recommended that you keep this as true. See EIP-1#adoptable-header for more information.
```
xinbenlv, line 12: I am in favor of having an adoptable field, but I suggest keeping the template default as `false` until we see a trend that authors intentionally choose to flip it to `true`.  I worry if we create a field and default to true, the new comers of EIP authors will think it's a community etiquette to sign it off. Just like the "[ ] I agree " box in many terms of services.

xinbenlv, line 12: I am in favor of having an adoptable field, but I suggest keeping the template default as `false` until we see a trend that authors intentionally choose to flip it to `true`.  I worry if we create a field and default to true, the new comers of EIP authors will think it's a community etiquette to sign it off. Just like the "[ ] I agree " box in many terms of services. 
```suggestion
adoptable: false  # It is recommended that you keep this as true. See EIP-1#adoptable-header for more information.
```

Pandapip1: Uh, it should be etiquette. Nobody likes stagnant EIPs.

Pandapip1:
> I prefer to let EIPs be "rehomed" solely at the discretion of the editors.

This enables this :)

Previously, it requires both editor and author approval. If EIPs opt-in to adoption, then only editors are required. But editors are obviously still required.

lightclient:
It doesn't require authors approval. The editors can already unilaterally reassign an EIP. We haven't done this before and we try to get author consent, but we do have the ability to reassign.

xinbenlv:
@lightclient I know that is for the good intention of community. 

but I respectively disagree with the assertion that editors as a group (individually or together) shall have the power to unilaterally reassign EIP. I think editors reserve the power to decline an EIP to be published, but reassign right seems beyond its governing power.

Pandapip1:
I agree with @xinbenlv here. Adoptability should either be opt-in or opt-out, but some control should be given to authors.

Pandapip1:
Also, I'm not sure what's up with the stagnant bot. Why wasn't it reminding us?

Pandapip1:
Would still like this.

lightclient:
As I've said multiple times in this thread and on EIPIP, I don't think this is a good idea. I would rather leave it up to editor discretion.

Pandapip1:
Dismissing stale bot.

lightclient:
I don't understand why you're dismissing the stale bot. I don't think this should be merged. Are we just going to have it sit here open forever?

Pandapip1:
> I don't understand why you're dismissing the stale bot. I don't think this should be merged. Are we just going to have it sit here open forever?

I want to remind everybody that this still exists, and that I am still actively interested in pursuing it.

Pandapip1:
Bump!

Pandapip1:
Bumpity boop!

Pandapip1:
Since enough people objected, I've set the adoptable entry in the template to be default-false, so that new EIPs are opt-in instead of opt-out.

This is a reminder: existing EIPs are, and will always be opt-in.

mspaansen (approved):

Pandapip1:
Still would like this.

Pandapip1:
Would like to reopen this.

Pandapip1:
Bump...

SamWilsn:
I thought we had decided we weren't doing this?

lightclient:
I also thought we agreed not to do this

Pandapip1:
> I thought we had decided we weren't doing this?

I thought I had simply stopped bringing it up. I probably misremembered then. Closing.

Pandapip1:
FWIW I would still like this

PR #5535, opened 2022-08-25
EIP-5450: Add rationale, more clarifications and update authors

gumb0:


EIPS/eip-5450.md
```diff
@@ -13,27 +13,31 @@ requires: 3540, 3670, 4200, 4750
...
+Code section validation rules as defined by [EIP-3670](./eip-3670.md) (which has been extended by [EIP-4200](./eip-4200.md) and [EIP-4750](./eip-4750.md)) are extended again to include the instruction flow traversal procedure, where every possible code path is examined, and data stack height at each instruction is recorded. 
 
-1. For each reachable instruction in the section, data stack height when executing it is the same for all possible code paths going through this instruction.
+*Data stack height* here refers to the number of stack values accessible by this function, i.e. it does not take into account values of caller functions' frames (but does include this function's inputs).  Note that validation procedure does not require actual data stack implementation, but only to keep track of its height. Current height value starts at `types[code_section_index].inputs` (number of inputs of this function) at function entry and is updated at each instruction.
```
xinbenlv, line 32:
```suggestion
*Data stack height* here refers to the number of stack values accessible by this function, i.e. it does not take into account the values of caller functions' frames (but does include this function's inputs).  Note that the validation procedure does not require actual data stack implementation, but only to keep track of its height. Current height value starts at `types[code_section_index].inputs` (number of inputs of this function) at function entry and is updated at each instruction.
```

EIPS/eip-5450.md
```diff
@@ -42,13 +46,30 @@ The complexity of this traversal is linear in the number of instructions, becaus
 
 ### Execution
 
-Given new deploy-time guarantees, EVM implementation is not required anymore to have run-time stack underflow check for each executed instruction.
+Given new deploy-time guarantees, EVM implementation is not required anymore to have run-time stack underflow check for each executed instruction. However implementations can keep it if they choose to do so.
```
xinbenlv, line 49:
```suggestion
Given new deploy-time guarantees, EVM implementations are not required anymore to have a run-time stack underflow check for each executed instruction. However each implementation can keep it if they choose to do so.
```

xinbenlv (approved):
LGTM. I let some editorial suggestions.

PR #5548, opened 2022-08-28
Add EIP-5548: NFT Operator Approval Control

mitchellfchan:
This is a draft EIP proposing:

An extension to ERC721 which standardizes NFT creators' ability to deny specific marketplaces or smart contracts the ability to spend tokens on an owner's behalf. This is achieved by adding a check against an owner-defined denyList in the `Approve` or `SetApprovalForAll` functions.

### DIFF
--- a/EIPS/eip-5548.md
+++ b/EIPS/eip-5548.md
@@ -3,7 +3,7 @@ eip: 5548
 title: NFT Operator Approval Control
 description: An EIP-721 extension to deny specific smart contracts the ability to spend tokens on an owner's behalf
 author: Mitchell F Chan (@mitchellfchan), et al.
-discussions-to: TK
+discussions-to: https://ethereum-magicians.org/t/eip-5548-eip-721-approve-operator-denylist/10549
 status: Draft
 type: Standards Track
 category: ERC
--- a/EIPS/eip-5548.md
+++ b/EIPS/eip-5548.md
@@ -74,7 +74,7 @@ The following functions in the standard OpenZeppelin `ERC721.sol` are slightly m
 ```solidity
 
 function setApprovalForAll(address operator, bool approved) public virtual override {
-  require(denyList[operator] == false, "Operator has been denied by contract owner.";   //ADD THIS LINE ONLY
+        require(denyList[operator] == false, "Operator has been denied by contract owner.");   //ADD THIS LINE ONLY
         _setApprovalForAll(_msgSender(), operator, approved);
     }
     
--- a/EIPS/eip-5548.md
+++ b/EIPS/eip-5548.md
@@ -2,7 +2,7 @@
 eip: 5548
 title: NFT Operator Approval Control
 description: An EIP-721 extension to deny specific smart contracts the ability to spend tokens on an owner's behalf
-author: Mitchell F Chan (@mitchellfchan), Harm van den Dorpel (@harmvandendorpel), et al.
+author: Mitchell F Chan (@mitchellfchan), Harm van den Dorpel (@harmvandendorpel), Billy Rennekamp (@okwme)
 discussions-to: https://ethereum-magicians.org/t/eip-5548-eip-721-approve-operator-denylist/10549
 status: Draft
 type: Standards Track
@@ -41,7 +41,7 @@ Our denyList is a single new mapping of operator addresses, paired with true/fal
 ```solidity
 
 mapping (address => bool) public denyList;
-event OperatorDenied(address badOperator);
+event OperatorFlagged(address indexed flaggedOperator, bool indexed status);
 
 ```
 
@@ -51,19 +51,14 @@ Addresses may be added to the denyList.
 
 ```solidity
 
-function updateDenyList(address badOperator, bool status) public onlyOwner {
+function updateDenyList(address flaggedOperator, bool status) public onlyOwner {
        
-        require(
-            _msgSender() == owner ,
-            "ERC721: approve caller is not token owner nor approved for all"
-        );
-
-        _addToDenyList(badOperator, status);
+        _updateDenyList(flaggedOperator, status);
     }
 
-function _updateDenyList(address badOperator, bool status) internal virtual {
-        denyList[badOperator] = status;
-        emit OperatorDenied(badOperator);
+function _updateDenyList(address flaggedOperator, bool status) internal virtual {
+        denyList[flaggedOperator] = status;
+        emit OperatorFlagged(flaggedOperator, status);
     }
 
 ```
--- a/EIPS/eip-5548.md
+++ b/EIPS/eip-5548.md
@@ -1,5 +1,5 @@
 ---
-eip: 5545
+eip: 5548
 title: NFT Operator Approval Control
 description: An EIP-721 extension to deny specific smart contracts the ability to spend tokens on an owner's behalf
 author: Mitchell F Chan (@mitchellfchan), et al.
@@ -19,7 +19,7 @@ This is intended for NFT creators who wish to exercise some control over where t
 
 There are many reasons why an NFT creator may wish to exercise control over which intermediary smart-contracts may buy and sell their creations on an owner's behalf. An intermediary platform may present artwork in a manner which the creator finds objectionable or dishonest, or the intermediary platform may openly disregard off-chain codes, conventions, or practices which the creator considers to be an essential part of the artwork or their career.
 
-At the time of writing, several intermediary platforms have chosen to disregard the widely agreed-upon convention of remitting resale royalties to creators. It is expected that the extension proposed in this EIP will frequently be paired with [EIP-2981](./eip-2981.md), with EIP-2981 acting as a way of signalling standardized royalty information, and [EIP-5545](./eip-5545.md) acting as a method of denying non-compliant intermediary platforms.
+At the time of writing, several intermediary platforms have chosen to disregard the widely agreed-upon convention of remitting resale royalties to creators. It is expected that the extension proposed in this EIP will frequently be paired with [EIP-2981](./eip-2981.md), with EIP-2981 acting as a way of signalling standardized royalty information, and this EIP acting as a method of denying non-compliant intermediary platforms.
 
 
 ## Motivation
--- a/EIPS/eip-5548.md
+++ b/EIPS/eip-5548.md
@@ -2,7 +2,7 @@
 eip: 5548
 title: NFT Operator Approval Control
 description: An EIP-721 extension to deny specific smart contracts the ability to spend tokens on an owner's behalf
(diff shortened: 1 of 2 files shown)


EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+Addresses may be added to the denyList.
+
+
+
+```solidity
+
+function updateDenyList(address badOperator, bool status) public onlyOwner {
+       
+        require(
+            _msgSender() == owner ,
```
okwme, line 57: this is redundant with the `onlyOwner` modifier

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+Our denyList is a single new mapping of operator addresses, paired with true/false values. An event is emitted when an address is added to the denyList.
+
+```solidity
+
+mapping (address => bool) public denyList;
+event OperatorDenied(address badOperator);
```
okwme, line 44: you need to add the boolean if you want the `updateDenyList` to allow the ability to remove an operator from the deny list as well. otherwise it will emit the `OperatorDenied` event even when they are being "allowed".

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+            _msgSender() == owner ,
+            "ERC721: approve caller is not token owner nor approved for all"
+        );
+
+        _addToDenyList(badOperator, status);
+    }
+
+function _updateDenyList(address badOperator, bool status) internal virtual {
+        denyList[badOperator] = status;
+        emit OperatorDenied(badOperator);
```
okwme, line 66: use the added boolean mentioned on line 44 comment unless you want to only deny

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+
+```
+
+
+The following functions in the standard OpenZeppelin `ERC721.sol` are slightly modified.
+
+```solidity
+
+function setApprovalForAll(address operator, bool approved) public virtual override {
+  require(denyList[operator] == false, "Operator has been denied by contract owner.";   //ADD THIS LINE ONLY
```
okwme, line 77:
```suggestion
  require(denyList[operator] == false, "Operator has been denied by contract owner.");   //ADD THIS LINE ONLY
```

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+    
+function approve(address to, uint256 tokenId) public virtual override {
+      address owner = ERC721.ownerOf(tokenId);
+      require(to != owner, "ERC721: approval to current owner");
+
+      require(
+          _msgSender() == owner || isApprovedForAll(owner, _msgSender()),
+          "ERC721: approve caller is not token owner nor approved for all"
+      );
+      require(denyList[operator] == false, "Operator has been denied by contract owner."; //ADD THIS LINE ONLY
```
okwme, line 89:
```suggestion
      require(denyList[to] == false, "Operator has been denied by contract owner."; //ADD THIS LINE ONLY
```

EIPS/eip-5548.md
mitchellfchan: Right... I think it's good to allow previously denied addresses to be readmitted. Suppose a situation where the creator and the platform were willing to work out an agreement or -- wild thought here -- the offending operator actually changed its policy.

So I guess the `OperatorDenied` event should be renamed `denyListUpdated` or  something like that.

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+
+      require(
+          _msgSender() == owner || isApprovedForAll(owner, _msgSender()),
+          "ERC721: approve caller is not token owner nor approved for all"
+      );
+      require(denyList[to] == false, "Operator has been denied by contract owner."; //ADD THIS LINE ONLY
+      _approve(to, tokenId);
+  }
+    
+ ```
```
harmvandendorpel, line 93: Do we need a getter to ask the contract whether a particular marketplace has been blocked.

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+
+```
+
+
+The following functions in the standard OpenZeppelin `ERC721.sol` are slightly modified.
+
+```solidity
+
+function setApprovalForAll(address operator, bool approved) public virtual override {
+        require(denyList[operator] == false, "Operator has been denied by contract owner.");   //ADD THIS LINE ONLY
```
harmvandendorpel, line 77: could the repetition of line 77 and 89 be DRY-ed up with a `modified`?

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
+---
+eip: 5548
+title: NFT Operator Approval Control
+description: An EIP-721 extension to deny specific smart contracts the ability to spend tokens on an owner's behalf
+author: Mitchell F Chan (@mitchellfchan), Harm van den Dorpel (@harmvandendorpel), et al.
```
xinbenlv, line 5: > et al.
[Editorial] Are there other authors? Usually all authors are listed unless there are particular reasons.?

EIPS/eip-5548.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+Our denyList is a single new mapping of operator addresses, paired with true/false values. An event is emitted when an address is added to the denyList.
+
+```solidity
+
+mapping (address => bool) public denyList;
```
xinbenlv, line 43: [Technical] This is a particular implementation. Are you sure you want to require a public mapping to be the "interface"?

You shall be able to use something like 
```solidity
function setAddressDenyListed(address, bool ) public;
function getAddressDenyListed(address ) public view returns (bool);
```
instead of mapping to make it a lot more generalize and allow other form of implementation. F

EIPS/eip-5548.md
xinbenlv, line 44: [Editorial] Naming it `OperatorDenied` with an `Operator` implies the address being added shall be an operator smart contract, I wonder if there is any thoughts put into checking if it's an operator?

EIPS/eip-5548.md
mitchellfchan: Adding more authors as they contribute. Used "et al." for now as a placeholder, anticipating more coming... which they are, thankfully!

EIPS/eip-5548.md
mitchellfchan: I used `operator` because that's the name of the parameter passed to `Approval` `SetApprovalForAll` etcetera. So I thought it would be clearer to use the name `Operator` because those are the methods being affected.

EIPS/eip-5548.md
mitchellfchan: Thanks! Removed the `require` in latest commit.

EIPS/eip-5548.md
mitchellfchan: Also added in most recent commit.

EIPS/eip-5548.md
mitchellfchan: Added in latest commit.

EIPS/eip-5548.md
mitchellfchan: LIne 77 here is the only line that I actually wrote, the rest is straight from the ERC721. So I wrote Line 77 to be in the same syntax/style so as to avoid confusion.

EIPS/eip-5548.md
```diff
@@ -0,0 +1,121 @@
...
+
+This is intended for NFT creators who wish to exercise some control over where their creations may be bought and sold, without limiting the rights of token holders. 
+
+There are many reasons why an NFT creator may wish to exercise control over which intermediary smart-contracts may buy and sell their creations on an owner's behalf. An intermediary platform may present artwork in a manner which the creator finds objectionable or dishonest, or the intermediary platform may openly disregard off-chain codes, conventions, or practices which the creator considers to be an essential part of the artwork or their career.
+
+At the time of writing, several intermediary platforms have chosen to disregard the widely agreed-upon convention of remitting resale royalties to creators. It is expected that the extension proposed in this EIP will frequently be paired with [EIP-2981](./eip-2981.md), with EIP-2981 acting as a way of signalling standardized royalty information, and this EIP acting as a method of denying non-compliant intermediary platforms.
```
SamWilsn, line 22: These paragraphs are more motivation than abstract.

EIPS/eip-5548.md
```diff
@@ -0,0 +1,121 @@
...
+
+
+## Motivation
+
+The promise of creator empowerment has driven adoption of the NFT standard (and of the Ethereum blockchain in general) for artists and creators. The opportunity to cut out intermediaries, or "middlemen," is a central tenet not just of NFT culture, but of crypto culture in general. Nonetheless, intermediaries persist. In many cases, these intermediaries serve an important function, creating smoother experiences for end users, or presenting NFT metadata in easily consumable formats. 
+
+However, as intermediary platforms willfully circumvent social codes established by NFT creators, it is within the spirit of crypto to provide technological tools to deny allowances to those intermediaries.
+
+It should be noted that this extension will not (nor does it aim to) stamp out any nefarious behaviour (such as royalty-dodging) from token holders. The position of this EIP is that any restriction to owner's ability to `Transfer` their NFTs would be overreaching. And of course, denyListed platforms could deploy new smart contracts, placing creators in the position of playing "whack-a-mole" to add new deployments to the denyList.
```
SamWilsn, line 31: This paragraph belongs in Security Considerations, and should be expanded with more detail.

EIPS/eip-5548.md
```diff
@@ -0,0 +1,121 @@
...
+
+Addresses may be added to the denyList.
+
+
+
+```solidity
+
+function updateDenyList(address flaggedOperator, bool status) public onlyOwner {
+       
+        _updateDenyList(flaggedOperator, status);
```
SamWilsn, line 56: I would remove all implementation details, and instead give a Solidity `interface`.

How the denylist is implemented should be up to the contract developers, since it won't matter at all to user interfaces.

SamWilsn (commented):
Looking pretty good!

I do have to ask, however, what is the benefit of this being an EIP (as opposed to a library)? Do you expect there to be multiple user interfaces that must be compatible with multiple NFT contracts? EIPs are generally used to standardize interfaces where there are multiple implementations and multiple consumers of the interface, and I'm not sure I see too many UIs popping up for this (likely just etherscan?)

PR #5550, opened 2022-08-28
Simple Works & Royalty Interest Representation Standard

royosherove:
A standardized way to represent intellectual works on chain, along with a refined royalty representation mechanism and associated metadata. This standard is not associated with a specific type of work and could represent many types of works such as musical works, videos, books, images and more.
The standard is kept very generic on purpose to allow the industry to evolve new ecosystems that can all rely on the same basic standard at their core


royosherove:
OK. Will fix and do another PR

PR #5572, opened 2022-09-01
Website: fix index.html issues

Pandapip1:


SamWilsn (commented):
Why are we changing the relative links? Aren't both styles equivalent?

Why are we changing the example EIPs?

SamWilsn:
> > Why are we changing the relative links? Aren't both styles equivalent?
> 
> If index.html gets renamed or used as a template, then the links will break.

I believe that unless the first character is a slash, both `./index.html` and `index.html` are resolved relative to the current base URL. I could entirely be mistaken. I think [this](https://url.spec.whatwg.org/#relative-state) is the relevant section of the spec.

Pandapip1:
>  resolved relative to the current base URL

That's the problem. If `index.html` gets copied to `foo.html`, then `foo.html` will need to change all its links, due to the way that jekyll paths work (`foo.html` -> `/foo`).

SamWilsn (approved):
:shrug:

PR #5586, opened 2022-09-03
Configuration: Add "furter" to whitelist

Pandapip1:
Add someone else's name to CodeSpell

### DIFF
--- a/.codespell-whitelist
+++ b/.codespell-whitelist
@@ -13,4 +13,4 @@ ende
 sting
 complies
 shiping
-complies
+furter
--- a/.codespell-whitelist
+++ b/.codespell-whitelist
@@ -13,3 +13,4 @@ ende
 sting
 complies
 shiping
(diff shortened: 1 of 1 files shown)


xinbenlv (approved):
LGTM for this change. 
If I may suggest, consider just exclude the author names instead.

Pandapip1:
> If I may suggest, consider just exclude the author names instead.

This has already been suggested. Sadly, this is not possible.

xinbenlv:
> 

I think it can be archived by markdown linter plus suppress marker `markdownlint-disable` ish, have we considered look into it?

Pandapip1:
Markdownlint doesn't check spelling. This is the current situation.

Admittedly, I have added a bypass so that codespell doesn't actually error right now, so finding a fix is not top priority.

SamWilsn (approved):

PR #5588, opened 2022-09-04
Update EIP-158: Update wording

fulldecent:
Updates wording to remove reference to a deprecated word


Pandapip1 (approved):
👍 from me

xinbenlv (approved):
LGTM

SamWilsn:
While I do agree we should be sensitive to these issues, I think we should keep existing EIPs as-written, and encourage future EIPs to use the correct terminology.

One issue is that if we replaced all "suicide" occurrences with "self-destruct" everywhere, [EIP-6](https://eips.ethereum.org/EIPS/eip-6) would no longer make sense.

SamWilsn:
Perhaps we could come to a solution that doesn't require editing the EIPs? Perhaps have Jekyll put up a warning on pages containing the word?

PR #5593, opened 2022-09-04
Add EIP-5593: Restrict Web3 Provider Object API Injection

kdenhartog:

 - The build passes.
 
cc @lightclient @axic @SamWilsn @MicahZoltu this is my first time submitting an EIP so if there's anything additional that needs to be done here please let me know.


EIPS/eip-5593.md
```diff
@@ -0,0 +1,76 @@
+---
+eip: 5593
+title: Restrict Web3 Provider Object API Injection
+description: Wallet guidance for restricting web3 provider object API access to secure contexts for improved privacy and security for wallet users.
+author: Yan Zhu (@diracdeltas), Brian R. Bondy (@bbondy), Kyle Den Hartog (@kdenhartog)
+discussions-to:  https://ethereum-magicians.org/t/rfc-limiting-provider-object-injection-to-secure-contexts/10670
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): Interface
```
Pandapip1, line 9:
```suggestion
category: Interface
```

EIPS/eip-5593.md
```diff
@@ -0,0 +1,76 @@
...
+created: 2022-09-05
+requires (*optional): <EIP number(s)>
```
Pandapip1, line 11:
```suggestion
```

EIPS/eip-5593.md
```diff
@@ -0,0 +1,76 @@
...
+---
+
+## Abstract
+
+Historically the web platform has had a notion of “powerful” APIs like [geolocation](https://www.w3.org/TR/geolocation/) and [camera/microphone](https://www.w3.org/TR/mediacapture-streams/), which are subject to additional security restrictions such as those defined by [secure contexts](https://www.w3.org/TR/secure-contexts/). Since the web3 provider object APIs allow dApp websites to request access to sensitive user data and to request use of user funds, new web3 provider object APIs generally should align to the security considerations of other powerful APIs like geolocation.
```
Pandapip1, line 16: No external links

EIPS/eip-5593.md
```diff
@@ -0,0 +1,76 @@
...
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Restrictions for providers
+
+The provider objects, e.g. `window.ethereum`, are not provided in all contexts. The following restrictions are REQUIRED for conformant wallets:
+
+- Provider objects SHOULD NOT be accessible in private (incognito) windows.
+- The origin MUST be a [secure origin](https://www.chromium.org/Home/chromium-security/prefer-secure-origins-for-powerful-new-features/#definitions) to have access to `window.ethereum`. This can be checked using `window.isSecureContext`, including inside iframes.
```
Pandapip1, line 31: No external links

EIPS/eip-5593.md
```diff
@@ -0,0 +1,76 @@
...
+    - Secure contexts include sites that are served from HTTPS but also HTTP `localhost`.
+- By default the web3 provider object APIs MUST NOT be exposed to 3p iframes.
+- `window.ethereum` MUST be `undefined` in an iframe where `window.isSecureContext` returns `false` in that iframe.
+- If the iframe is first-party to the top-level origin AND the `sandbox` attribute is set on the iframe, the provider object MUST be left `undefined`. If `sandbox="allow-same-origin"` attribute is set the provider object MAY be injected. 
+    - Note "allow-same-origin"` does nothing if the iframe is third-party.
```
Pandapip1, line 36:
```suggestion
    - Note `"allow-same-origin"` does nothing if the iframe is third-party.
```

EIPS/eip-5593.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Rationale
+
+By limiting the capabilities of where the web3 provider object APIs are being injected we can reduce the surface area of where attacks can be executed.Given the sensitivity of data that's passed to the web3 provider object APIs some basic levels of authentication and confidentiality should be met in order to ensure that request data is not being intercepted or tampered with. While there has been attempts to [limit request access via the wallet](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2255.md) interface itself there's not been limitations that have been set to where these web3 provider object APIs are expected to be or not be injected. Since the secure contexts web platform API is a well developed boundary that's been recommended by W3C and the fact that the web3 provider object APIs are extending the traditional web platform APIs, no other alternative solutions have been considered in order to extend current established prior art.
```
Pandapip1, line 40: No external links

EIPS/eip-5593.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+## Backwards Compatibility
+
+Wallet extensions SHOULD consider adding a "developer mode" toggle via a UX so that dApp developers have the capability to disable the insecure context (http) check for the http://localhost:<any-port> origin only in the event that [localhost does not return `true` for secure context](https://www.w3.org/TR/secure-contexts/#localhost). This will allow dApp developers to be able to continue to host dApps on the localhost origin if a browser environment has chosen to not already consider localhost a secure context. Most major browser providers do consider localhost a secure context already. This toggle MUST be set to disabled by default.
```
Pandapip1, line 45: No external links

EIPS/eip-tbd.md
```diff
@@ -0,0 +1,76 @@
+---
```
Pandapip1, line 1: Delete this duplicate file please

EIPS/eip-5593.md
xinbenlv, line 31: @EIPW please suppress warning

I support adding this link. I think this is informative and non-promotional.
FYI: @gcolvin @SamWilsn @Pandapip1

xinbenlv: @EIPW please suppress warning

I support adding this link. I think this is informative and non-promotional.
FYI: @gcolvin @SamWilsn @Pandapip1

EIPS/eip-5593.md
xinbenlv, line 40:
```suggestion
By limiting the capabilities of where the web3 provider object APIs are being injected we can reduce the surface area of where attacks can be executed.Given the sensitivity of data that's passed to the web3 provider object APIs some basic levels of authentication and confidentiality should be met in order to ensure that request data is not being intercepted or tampered with. While there has been attempts to [limit request access via the wallet](./eip-2255.md) interface itself there's not been limitations that have been set to where these web3 provider object APIs are expected to be or not be injected. Since the secure contexts web platform API is a well developed boundary that's been recommended by W3C and the fact that the web3 provider object APIs are extending the traditional web platform APIs, no other alternative solutions have been considered in order to extend current established prior art.
```

EIPS/eip-5593.md
xinbenlv: @EIPW please suppress warning

@Pandapip1 , I support adding this link. I think this is informative and non-promotional.
FYI: @gcolvin @SamWilsn @Pandapip1

EIPS/eip-tbd.md
```diff
@@ -0,0 +1,76 @@
...
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): Interface
+created: 2022-09-05
+requires (*optional): 1193
+---
+
+## Abstract
+
+Historically the web platform has had a notion of “powerful” APIs like [geolocation](https://www.w3.org/TR/geolocation/) and [camera/microphone](https://www.w3.org/TR/mediacapture-streams/), which are subject to additional security restrictions such as those defined by [secure contexts](https://www.w3.org/TR/secure-contexts/). Since the web3 provider object APIs allow dApp websites to request access to sensitive user data and to request use of user funds, new web3 provider object APIs generally should align to the security considerations of other powerful APIs like geolocation.
```
xinbenlv, line 16: @EIPW please suppress warning

I support adding this link. I think this is informative and non-promotional.
FYI: @gcolvin @SamWilsn @Pandapip1

EIPS/eip-5593.md
kdenhartog: What's the best way to externally reference other normative specs and useful documentation then to enhance context for readers and implementers? Within W3C specs there's something similar to citation bibliography that they use. This would be possible to achieve in a similar method using markdowns [reference style links](https://www.markdownguide.org/basic-syntax/). 

This has the benefit of better readability while also being able to externally link for additional context. I assume there's a reason this rule was put in place though and this benefit was outweighed by the downsides of using them. 

Out of curiosity what's the motivation for this rule as? I assume either security related or to prevent maintaining broken links, but am not certain.

I'd be happy to help define some additional editorial rules in the process document to allow for external links if the editor team is open to revisiting this rule.

xinbenlv: @kdenhartog I think it's complete reasonable to include reference to w3c. The editors are still debating rules about external links. 

FYI: @lightclient @gcolvin for comments

EIPS/eip-5593.md
xinbenlv: This is not an external link...

EIPS/eip-5593.md
xinbenlv: External links are allowed by EIP-1 when necessarily. 
Please see https://github.com/ethereum/EIPs/issues/5597

EIPS/eip-tbd.md
xinbenlv: +1

EIPS/eip-5593.md
Pandapip1: I do not. This is in the specification section, and is not a link that anyone could reasonably say is permanent. Please add a PDF of this to the assets folder.

EIPS/eip-tbd.md
Pandapip1: Those, again, are not the criterion for determining what is or is not an okay link. The only criterion is "is it going to last longer than the EIPs repository." I would say that links to w3.org are likely okay, but this is not the place to debate that. If you feel like links to w3 are justified, the correct place to do so is in a PR making that change to EIP-1.

EIPS/eip-5593.md
Pandapip1: image

It's an external link, admittedly one that has an easier fix than most.

EIPS/eip-5593.md
Pandapip1: Those are not the criteria for determining whether a link is allowed or not.

EIPS/eip-5593.md
kdenhartog: I believe I've updated that one, but I'll need to double check when I'm back on a laptop. That one definitely makes sense to just directory reference.

EIPS/eip-5593.md
kdenhartog:
```suggestion
requires: 1193
```

EIPS/eip-5593.md
kdenhartog: I'd be open to changing this link to a better defined definition. Chromium docs can change so relying on this for a normative definition is a risk. I'll take a look at if I can reference a WHATWG or W3C definition here instead.

kdenhartog: I'd say a PDF to a piece of text isn't the right call here because this is a normative statement that needs to be tightly defined so that implementers don't end up with a different implementations.

rekmarks:
I just want to chime in to thank @kdenhartog for their excellent work here and that we will follow up on [the PR to the MetaMask extension](https://github.com/MetaMask/metamask-extension/pull/15736) to get it merged.

kdenhartog:
Thanks @rekmarks credit definitely goes to @diracdeltas and @bbondy for the idea and implementation. I got the easy job of turning the docs into an EIP. 😄

EIPS/eip-5593.md
kdenhartog: resolving this thread since there's a lot of overlap with the discussion above. Let's consolidate this conversation to the following comment thread: https://github.com/ethereum/EIPs/pull/5593#discussion_r962979310

EIPS/eip-tbd.md
kdenhartog: I'm happy to hold this until we get that resolved and will contribute to the discussion as well based on my experience as an editor in W3C.

Pandapip1:
@kdenhartog mind adding these files to the directory I just created? https://github.com/kdenhartog/EIPs/pull/1

kdenhartog:
> @kdenhartog mind adding these files to the directory I just created? [kdenhartog#1](https://github.com/kdenhartog/EIPs/pull/1)

Thank you for the PR to consolidate this into PDFs. I'll merge that in if we decide that linking isn't going to be allowed in the #5597 discussion. My preference would be to link to these documents at this point because the TR directory at W3C has a very rigorous process for updating it and along with that they've built in good processes for how specs are edited with errata and new versions. If we go towards the PDF route we end up having to maintain these PDFs rather than being able to lean on W3C's processes which handle this. Especially since some of these things like the [permissions API](https://www.w3.org/TR/permissions/) are actively being worked on at the moment and require that I normatively reference them for an additional normative MAY I'm working on adding right now.

As an example, here's a [spec for XML](https://www.w3.org/TR/xml/) which was published in 2008, so going on 14 years now as a stable URL. It includes notes about links being updated and errata which are not available with the PDF approach. In my opinion using links for the W3C spec is actually an advantage here. If we were going to link to Chromium docs, I'd be more open to it, but even in that case we can lean on commits

Pandapip1:
> If we decide that linking isn't going to be allowed in the https://github.com/ethereum/EIPs/issues/5597 discussion then we can update everything accordingly. That work for you?

It would be my preference for you to just link to the PDFs anyway - it's also likely to get your EIP merged faster.

Pandapip1:
Currently, the policy on links is very restrictive - and for a pretty good reason. I'm leaning toward allowing some external specifications (as long as they have a PDF backup), but this would have to be an EIP-1 policy.

EIPS/eip-5593.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Provider objects SHOULD NOT be accessible in private (incognito) windows.
+- The origin MUST be a [secure origin](https://www.chromium.org/Home/chromium-security/prefer-secure-origins-for-powerful-new-features/#definitions) to have access to `window.ethereum`. This can be checked using `window.isSecureContext`, including inside iframes.
+    - Secure contexts include sites that are served from HTTPS but also HTTP `localhost`.
+- By default the web3 provider object APIs MUST NOT be exposed to 3p iframes.
+- `window.ethereum` MUST be `undefined` in an iframe where `window.isSecureContext` returns `false` in that iframe.
+- If the iframe is first-party to the top-level origin AND the `sandbox` attribute is set on the iframe, the provider object MUST be left `undefined`. If `sandbox="allow-same-origin"` attribute is set the provider object MAY be injected. 
+    - Note `"allow-same-origin"` does nothing if the iframe is third-party.
+- If the iframe is a third party to the top-level secure origin, it SHOULD be blocked. It MAY be unblocked if  the iframe uses the `allow=ethereum` attribute. In order to achieve this implementers are REQUIRED to extend the Permissions API.
```
coryhardman, line 37: I think this violates the spirit of why you block access to the wallet to 3p iframes. If the user is on a malicious website and trying to get the vicitm to do a "click jack" attack. 

For example the user is on a copycat domain and the site to keep the user believing that they want to avoid the connect prompt. They get the user to click on a purchase button or transfer button on the real website through opaque iframes.

EIPS/eip-5593.md
```diff
@@ -0,0 +1,83 @@
...
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Restrictions for providers
+
+The provider objects, e.g. `window.ethereum`, are not provided in all contexts. The following restrictions are REQUIRED for conformant wallets:
+
+- Provider objects SHOULD NOT be accessible in private (incognito) windows.
+- The origin MUST be a [secure origin](https://www.chromium.org/Home/chromium-security/prefer-secure-origins-for-powerful-new-features/#definitions) to have access to `window.ethereum`. This can be checked using `window.isSecureContext`, including inside iframes.
```
coryhardman, line 31: Could we also expand EIP-2255 here to say that the permissioning should be scoped to the origin. Origin must be secure is great, but like a permission you should get the users consent that they agree prior to displaying and the scope of the grant should be to the origin.

EIPS/eip-5593.md
```diff
@@ -0,0 +1,83 @@
...
+Wallet extensions SHOULD consider adding a "developer mode" toggle via a UX so that dApp developers have the capability to disable the insecure context (http) check for the http://localhost:<any-port> origin only in the event that [localhost does not return `true` for secure context](https://www.w3.org/TR/secure-contexts/#localhost). This will allow dApp developers to be able to continue to host dApps on the localhost origin if a browser environment has chosen to not already consider localhost a secure context. Most major browser providers do consider localhost a secure context already. This toggle MUST be set to disabled by default.
+
+## Test Cases
+
+### Required Test Cases
+
+- Top level `http://a.com` -> blocked (insecure)
+- Top level `https://a.com` -> allowed
+- Top level `https://a.com` with `<iframe src="http://a.com/">` -> blocked (insecure/3p)
+- Top level `http://a.com` with `<iframe src="https://a.com/">` -> blocked (insecure/3p)
```
coryhardman, line 55: insure first party right?

EIPS/eip-5593.md
coryhardman: So test cases for that would be:

https://example.com is granted, a new grant would be required on
https://www.example.com
https://sub.example.com
https://example2.com
http://example.com -- blocked not secure

EIPS/eip-5593.md
```diff
@@ -0,0 +1,83 @@
...
+- Top level `https://a.com` with `<iframe src="https://a.com">` -> allowed
+- Top level `https://a.com` with `<iframe src="https://b.com">` -> blocked (3p)
```
coryhardman, line 57: we should add a subdomain example:

top level `https://a.com` with `<iframe src="https://sub.a.com">` -> blocked (3p) given that the origin is not the same even if it is a subdomain.

EIPS/eip-5593.md
```diff
@@ -0,0 +1,75 @@
...
+### Restrictions for providers
+
+The provider objects, e.g. `window.ethereum`, are not provided in all contexts. The following restrictions are REQUIRED for conformant wallets:
+
+- Provider objects SHOULD NOT be accessible in private (incognito) windows.
+- The origin MUST be a [secure origin](https://www.chromium.org/Home/chromium-security/prefer-secure-origins-for-powerful-new-features/#definitions) to have access to `window.ethereum`. This can be checked using `window.isSecureContext`, including inside iframes.
+    - Secure contexts include sites that are served from HTTPS but also HTTP `localhost`.
+- By default the web3 provider object APIs MUST NOT be exposed to 3p iframes.
+- `window.ethereum` MUST be `undefined` in an iframe where `window.isSecureContext` returns `false` in that iframe.
+- If the iframe is first-party to the top-level origin AND the `sandbox` attribute is set on the iframe, the provider object MUST be left `undefined`. If `sandbox="allow-same-origin"` attribute is set the provider object MAY be injected.
```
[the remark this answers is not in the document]
kdenhartog: Good point, I think this should become a MUST. I'll double check later to make sure this isn't an issue with the secure contexts implementations that makes it a MAY or if it's just a mistake I made when writing it up.

EIPS/eip-5593.md
kdenhartog: So this would only work if the user grants permission to the ethereum permission via the 3rd party. Most extension based wallets don't get access to this, but Brave and Opera can. We chose to implement this and I've also kicked off an issue in the permissions repo at W3C to see if browsers are open to extending this API to extensions.

EIPS/eip-5593.md
kdenhartog: I quite like this usage of EIP-2255. We've been looking at how to better permission things for privacy and security and this seems like a good candidate to expand the 2255 permissions.

EIPS/eip-5593.md
kdenhartog: Yeah that's insecure top level not 3p. Will get this switched. Good catch!

EIPS/eip-5593.md
kdenhartog: That seems right, but I'll need to double check what window.isSecureContext returns for this case. I agree some subdomain examples should be added here.

EIPS/eip-5593.md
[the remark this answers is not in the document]
kdenhartog: Figured out why I put this as a MAY and updated it since it wasn't well defined. The reasoning for the MAY was that in the third party iframe case it wouldn't make a difference, so rather than leave it as a MAY I updated it to a MUST and specified it to be around first party frames. Third party frames would still be updated based on the `allow` attribute and the permissions API.

EIPS/eip-5593.md
kdenhartog: On re-reading this I think this is better to be discussed under the contents of EIP-2255. I'm generally in support of the idea, but if I'm reading this correctly it sounds like something that needs to be documented in that EIP instead. Given it's a `stagnant` status, I don't see any reason we can't directly update EIP-2255 unlike this which should have gone in EIP-1193, but it is already a `final` status so a secondary EIP was needed. Wdyt of just getting that discussion going in [this thread](https://ethereum-magicians.org/t/web3-login-permissions/3583/29) and then updating based on where discussion lands?

EIPS/eip-5593.md
diracdeltas: FWIW we added this attribute only because we ran into real sites that were using 3p iframes for features that required `window.ethereum` access. IMO solving clickjacking is not in scope here - if a site doesn't want to be framed, it should use CSP or x-frame-options to signal so to prevent clickjacking.

diracdeltas: this follows the same security model as https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-allow

EIPS/eip-5593.md
kdenhartog: @Pandapip1 I updated the link here to now link to a normative definition defined by Secure Contexts called ["Potentially trustworthy origins"](https://www.w3.org/TR/secure-contexts/#potentially-trustworthy-origin). This is effectively the algorithm being tested by `window.isSecureContext`

EIPS/eip-5593.md
SamWilsn: The current policy is to disallow external links in their entirety. If the linked resources allows redistribution, you may embed it in `assets/eip-5593` and link to it from there.

In this case, I don't think the links are necessary to understand the proposal, so I'd just remove them.

EIPS/eip-5593.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Motivation
+
+Wallets are oftentimes maintaining security and safety of users' funds that can be equivalent to large portions of money. For this reason, it's a good idea to restrict access to the APIs that grant this access in certain contexts to align it more closely with the principles followed for web platform APIs given web3 provider object APIs are often extending this platform via extensions. By adding in restrictions to where extensions inject their APIs, we're reducing the surface area that malicious web pages could fingerprint or falsely request access to the web3 provider object APIs to harm users by accessing their data or stealing their web3 assets.
```
SamWilsn, line 20: Just simplifying the phrasing here:

```suggestion
Wallets maintain security and safety of users' funds. For this reason, it's a good idea to restrict access to the Ethereum APIs to align more closely with the principles followed for web platform APIs given web3 provider object APIs are often extending this platform via extensions. By adding in restrictions to where extensions inject their APIs, we're reducing the surface area that malicious web pages could use to fingerprint or falsely request access.
```

EIPS/eip-5593.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Restrictions for providers
+
+The provider objects, e.g. `window.ethereum`, are not provided in all contexts. The following restrictions are REQUIRED for conformant wallets:
```
SamWilsn, line 28: > The provider objects, e.g. `window.ethereum`, are not provided in all contexts.

Are you saying they currently aren't provided in all contexts, or that they shouldn't be provided in all contexts by wallets following this standard?

EIPS/eip-5593.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Provider objects SHOULD NOT be accessible in private (incognito) windows.
+- The origin MUST be a [potentially trustworthy origin](https://www.w3.org/TR/secure-contexts/#potentially-trustworthy-origin) to have access to `window.ethereum`. This can be checked using `window.isSecureContext`, including inside iframes.
+    - Secure contexts include sites that are served from HTTPS but also HTTP `localhost`. 
+    - The browser implementation may optionally also support configured [potentially trustworthy origins] that would normally not be considered trustworthy if the user configures their browser to do so. See the [Development Environments section of Secure Contexts](https://www.w3.org/TR/secure-contexts/#development-environments) for additional details.
```
SamWilsn, line 33:
```suggestion
    - The browser implementation may optionally also support configured "potentially trustworthy origins" that would normally not be considered trustworthy if the user configures their browser to do so.
```

EIPS/eip-5593.md
kdenhartog: > they shouldn't be provided in all contexts by wallets following this standard?

Meaning this one. Good catch this language could use to be updated

Pandapip1:
```suggestion
In some contexts, Ethereum provider objects (such as `window.ethereum`) MUST NOT be provided. Wallets MUST obey the following restrictions:
```
Nit: EIPs are always optional to implement, and it is up to the various implementations to choose to apply or not apply any given EIP. The usage of the word "compliant" is therefore redundant when it refers to the current EIP.

kdenhartog: Two points:

1. The ambiguity of the MUST NOT normative language update here is an untestable statement. 

2. The purpose of this statement is to say that ALL of the following statements MUST be implemented for this to be conformant. It's not conforming with this spec to pick and choose the rules you like and don't like here as that would lead to compatibility issues between wallets.

EIPS/eip-5593.md
kdenhartog: This is less precise as to what needs to be accounted for here. The point of this statement is to call out that the `window.isSecureContext` API can be overridden by the user if they configure they're browser to treat an origin as potentially trustworthy when the standard implementation would not consider the origin as such.

EIPS/eip-5593.md
kdenhartog: Looks like this subdomain example would actually be secure. See [example 5 here](https://www.w3.org/TR/secure-contexts/#examples-framed)

I'll be sure to add this as an example case though.

kdenhartog: Double checked the implementation here in chromium and confirmed this is allowed. Also, found there's a test case for this already. 

https://github.com/ethereum/EIPs/pull/5593/files#diff-988509a7775edd935e28ec73219554c2c3a5713b53029e79b7188eebe93230d4R61

kdenhartog:
@SamWilsn found a way to bypass the relative links linting of the bot 🙃

EIPS/eip-5593.md
```diff
@@ -0,0 +1,84 @@
...
+status: Draft
+type: Standards Track
+category: Interface
+created: 2022-09-05
+requires: 1193
+---
+
+## Abstract
+
+Historically the web platform has had a notion of “powerful” APIs like <a href="https://www.w3.org/TR/geolocation/">geolocation</a> and <a href="https://www.w3.org/TR/mediacapture-streams/">camera/microphone</a>, which are subject to additional security restrictions such as those defined by <a href="https://www.w3.org/TR/secure-contexts/">secure contexts</a>. Since the Ethereum Provider APIs allow dApp websites to request access to sensitive user data and to request use of user funds, new Ethereum Provider APIs generally should align to the security considerations of secure context in order to better protect the users data and users funds managed via the web.
```
Pandapip1, line 16: Please use markdown-based links.

coryhardman (dismissed):

EIPS/eip-5593.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Top level `http://a.com` -> blocked (insecure/top level)
+- Top level `https://a.com` -> allowed
+- Top level `https://a.com` with `<iframe src="http://a.com/">` -> blocked (insecure first party iframe)
+- Top level `http://a.com` with `<iframe src="https://a.com/">` -> blocked (insecure top level window)
+- Top level `https://a.com` with `<iframe src="https://a.com">` -> allowed
+- Top level `https://a.com` with `<iframe src="https://b.com">` -> blocked (3p iframe without sufficient privileges)
+- Top level `https://b.com` with `<iframe src="http://a.com/">` with `<iframe src="https://b.com">` -> blocked (insecure iframe)
+- Top level `https://b.com` with `<iframe src="https://a.com">` with `<iframe src="https://b.com">` -> blocked (3p iframe without sufficient privileges)
+- Top level `https://a.com` with `<iframe src="https://sub.a.com">` -> allowed
```
coryhardman, line 61: Why is this allowed? https://a.com and https://sub.a.com are completely different origins. There should be no difference between this and a.com and b.com. We should require the definer to add the `allow="ethereum"` explicitly here.

diracdeltas: Cory is correct here; I think there was some confusion because Brave considered changing "third party" to "different eTLD+1" at one point because the policy as-written would break ImmutableX (https://github.com/brave/brave-core/pull/13564), but we never actually did that. @kdenhartog do you see any cases where we are considering suborigins as first party for provider injection?

we should probably explicitly state this case in https://github.com/brave/brave-wallet-docs/blob/main/docs/provider-availability.md and check that brave-core behavior matches.

diracdeltas: actually i see it's already stated here in our docs:
 
> Top level https://a.com with <iframe src="https://b.a.com"> -> blocked (3p)

kdenhartog: Good catch, I'll double check my notes and tests to see what's going on here. I think what happened was because [example 5 in the spec](https://www.w3.org/TR/secure-contexts/#examples-framed) shows the suborigin as a secure context I was under the impression that this would be allowed.

kdenhartog:
I've updated this to use markdown reference links and have reduced every link to either W3C or an in repo link (other than the eth magicians link at the header). I'm not certain @SamWilsn and @Pandapip1 are acceptable to this though, but I've not been tracking the latest EIP editors discussions that have been considering this.

An alternative option for this is for me to create a Citations PDF in the assets folder to address this linking concern as an alternative method to not need to directly link from the EIP.

Pandapip1 (commented):
@SamWilsn there are some weird EIPW errors

kdenhartog:
@Pandapip1 I'm betting [my changes to the regex](https://github.com/ethereum/eipw/pull/47) to address HTML links not being caught by the eipw-lint lib is what's causing these. This is assuming those changes have been deployed to the GH action. @SamWilsn it might be worth reverting those changes out for the time being.

EIPS/eip-5593.md
kdenhartog: I've confirmed I can't connect in this test case by using this: https://projectopensea.github.io/test-3p-sandbox-iframe/

So this one just needs to be updated to blocked (3p) again

kdenhartog: This one has been properly updated here now

Pandapip1 (dismissed):
Looks fine by me. Adding to manual merge queue.

coryhardman (approved):

kdenhartog:
Small editorial changes made and also removed the optional test cases

SamWilsn:
Opened https://github.com/kdenhartog/EIPs/pull/2 with my suggestions.

kdenhartog:
@SamWilsn I've reviewed https://github.com/kdenhartog/EIPs/pull/2 and updated to try and and make it work as much as possible. FWIW, I just realized there's no way to link to a reference implementation right now either so I'm probably going to have to drop this section to merge it too.

kdenhartog:
Once again the EIPW bot does not reflect language in the EIP-1 document. It's not a requirement for the sections to exactly match the template according to EIP-1. See here for details: https://eips.ethereum.org/EIPS/eip-1#eip-formats-and-templates. No where does it state that the EIP MUST match the template.

If the bot is going to be the ultimate arbiter of truth then there's no point in having EIP-1 in my opinion and there should be more than @SamWilsn implementing and reviewing changes to the bot if so. Right now governance of this repo is being unilaterally managed by the EIP Walidator bot, not EIP-1.

SamWilsn:
I'm sorry you've been having an unpleasant time with the bot and the process in general. I'm happy to try and help get your PR merged as much as I am able.

> Once again the EIPW bot does not reflect language in the EIP-1 document. It's not a requirement for the sections to exactly match the template according to EIP-1. See here for details: https://eips.ethereum.org/EIPS/eip-1#eip-formats-and-templates. No where does it state that the EIP MUST match the template.

`eipw` enforces the generally agreed upon consensus of editors, and EIP-1 lays out the hard rules of the process. Where EIP-1 does allow leeway `eipw` probably doesn't. That's somewhat intentional. For the majority of cases, we expect that the rules in `eipw` will be followed. In the remaining exceptional cases, EIP editors can override `eipw`.

You'll note that EIP-1 says "Each EIP should have the following parts:", and lists the sections you should include. `eipw` enforces those rules but if an editor decides to allow it, you could theoretically have other sections. In practice, it's very hard to convince editors to override those rules.

> Additionally, it's throwing errors on the recommended template for licensing.

Yes, that's a bug: https://github.com/ethereum/eipw/issues/51

Sorry about that! It looks like there's an `http://localhost` without backticks on line 78.

> If the bot is going to be the ultimate arbiter of truth then there's no point in having EIP-1 in my opinion and there should be more than @SamWilsn implementing and reviewing changes to the bot if so. Right now governance of this repo is being unilaterally managed by the EIP Walidator bot, not EIP-1.

`eipw` is just a tool to catch the most common problems we see. The ultimate arbiter of truth is and will always be the rough consensus of EIP editors.

Cryptonics1:
Thank you Sam Wilson! And the gang, I apologize for my novice
understanding....I do sincerely thank you for helping me, I will implement
the things you know as I clearly do not..my phone was completely hacked and
erased 2 days ago so I've been trying to recover all lost data....these
advasaries are in every app program I have.....I usually have about 20 to
30 k trackers on me at all times....You guys are the best and thank you
again for your gnosis.
JCL

kdenhartog:
I appreciate the thoughtful and level-headed response as you always bring to the table @SamWilsn. I presume in this case the 4 ones about the license can be overridden. Is the one about the privacy considerations one that you're willing to do as well since it's a two hash header rather than 3? If not, I'll change it to a subsection of security for now as you've already made a huge effort on getting the links in place which was far more important to me. We can save the privacy considerations as a top level header for another day and merge this one.

SamWilsn:
> I presume in this case the 4 ones about the license can be overridden.

I _think_ there are still external links? Three in the reference implementation, and the one on 78 I mentioned earlier.

> Is the one about the privacy considerations one that you're willing to do as well since it's a two hash header rather than 3? If not, I'll change it to a subsection of security for now [...]

You should move it to be a subsection of security considerations.

Cryptonics1:
Thank you , You guys are top knotch!

kdenhartog:
> > I presume in this case the 4 ones about the license can be overridden.
> 
> I _think_ there are still external links? Three in the reference implementation, and the one on 78 I mentioned earlier.
> 
> > Is the one about the privacy considerations one that you're willing to do as well since it's a two hash header rather than 3? If not, I'll change it to a subsection of security for now [...]
> 
> You should move it to be a subsection of security considerations.

image


Ok, I'll try to go a few more rounds with this thing and see if I can figure it out. I see a few more links in this EIP (implementation ones for example), so I presume it's just flagging the wrong line then and I miss understood https://github.com/ethereum/eipw/issues/51.

xinbenlv:
@Pandapip1 In the industry a lot of people confused security with privacy, I would suggest as a RFC/XIP type of body, we keep the distinction more obivous?

Shall we add a "Other considerations" to catch all? FYI @kdenhartog

kdenhartog:
That's a good approach that allows the bot to remain more strict to the template while also having a good abstraction point that makes this useful beyond just privacy considerations.

EIPS/eip-5593.md
```diff
@@ -0,0 +1,92 @@
...
+- Top level `https://a.com` with `<iframe src="https://a.com" sandbox="allow-same-origin allow-scripts">` -> allowed (Note this case is discouraged by User Agent implementer's such as Mozilla's MDN documents because it’d allow the iframe to remove its own `sandbox` attribute. See the `sandbox` attribute section of the iframes document in MDN for more details.)
+- Top level `data://foo with <iframe src="data://bar">` -> blocked (insecure top level scheme)
+- Top level `file://foo with <iframe src="file://bar">` -> blocked (third-party iframe)
+- Top level `https://a.com` with `<iframe src="https://b.com" sandbox="allow-same-origin allow-scripts">` -> blocked (third-party iframe without sufficient privileges)
+
+## Reference Implementation
+
+Brave Wallet desktop and Android implementation: https://github.com/brave/brave-core/pull/13739
+Brave Wallet iOS implementation: https://github.com/brave/brave-ios/pull/5522/
+Metamask implementation: https://github.com/MetaMask/metamask-extension/pull/15736
```
Pandapip1, line 76: I would like to get this EIP merged. You can always re-add this if the policy changes:
```suggestion
```

EIPS/eip-5593.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Security Considerations
+
+### User Enables Developer Mode
+
+Oftentimes developers require the ability to develop dApps locally in order to test their website and develop while hosting their dApp on http://localhost. In this case localhost would be blocked and compatibility issues would arise when developing a dApp locally. In order to increase compatibility for dApp developers a toggle to disable the check for the localhost can be considered. If this were to be extended beyond the localhost origin it could be used as a means to convince users to enable developer mode in order to subvert the guards put in place by this EIP. Therefore, implementations should be cautious when extending this developer toggle beyond the scope of the localhost origin.
```
Pandapip1, line 82: False positive:
```suggestion
Oftentimes developers require the ability to develop dApps locally in order to test their website and develop while hosting their dApp on `http://localhost`. In this case localhost would be blocked and compatibility issues would arise when developing a dApp locally. In order to increase compatibility for dApp developers a toggle to disable the check for the localhost can be considered. If this were to be extended beyond the localhost origin it could be used as a means to convince users to enable developer mode in order to subvert the guards put in place by this EIP. Therefore, implementations should be cautious when extending this developer toggle beyond the scope of the localhost origin.
```

Pandapip1 (changes requested):

EIPS/eip-5593.md
kdenhartog: Yeah meant to remove these during the update - going to get this fixed up by EoD today by removing these links for now and adding back in once they're acceptable via EIP-5757 registration

Pandapip1 (approved):
LGTM!

PR #5596, opened 2022-09-05
Website: Add MathJax

Pandapip1:
Allows using LaTeX in EIPs (yay!)

### DIFF
--- a/_layouts/eip.html
+++ b/_layouts/eip.html
@@ -92,7 +92,7 @@ https://schema.org/TechArticle
     "copyrightYear": "{{ page.created | date: "%Y" }}"
   }
 </script>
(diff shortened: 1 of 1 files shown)


_layouts/eip.html
```diff
@@ -92,3 +92,11 @@ <h2>Citation</h2>
     "copyrightYear": "{{ page.created | date: "%Y" }}"
   }
 </script>
+<script type="text/javascript" async src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.3/MathJax.js?config=TeX-MML-AM_CHTML">
```
xinbenlv, line 95: Nice! But I wonder if there is any security concern? e.g. consider add MD5 checksum for loading MathJax

Pandapip1: Good catch!

SamWilsn (approved):
As discussed in EIPIP 66 (ethereum-cat-herders/EIPIP#181), this is pretty low risk and if we don't like it, we can remove it.

PR #5607, opened 2022-09-06
Update EIP-2771: Add myself as an author and bring it up to date

Pandapip1:
CC other authors: Ronan Sandford (@wighawag), Liraz Siri (@lirazsiri), Dror Tirosh (@drortirosh), Yoav Weiss (@yoavw), Alex Forshtat (@forshtat), Hadrien Croubois (@Amxx), Sachin Tomar (@tomarsachin2271), Patrick McCorry (@stonecoldpat), Nicolas Venturo (@nventuro), Fabian Vogelsteller (@frozeman)

### DIFF
--- a/EIPS/eip-2771.md
+++ b/EIPS/eip-2771.md
@@ -12,7 +12,7 @@ created: 2020-07-01
 
 ## Abstract
 
-This EIP defines a contract-level protocol for `Recipient` contracts to accept meta-transactions through trusted `Forwarder` contracts. No protocol changes are made. `Recipient` contracts fetch the effective `msg.sender` (`_msgSender()`) and `msg.data` (`_msgData()`) through custom getter functions. 
+This EIP defines a contract-level protocol for `Recipient` contracts to accept meta-transactions through trusted `Forwarder` contracts. No protocol changes are made. `Recipient` contracts are sent the effective `msg.sender` (referred to as `_msgSender()`) and `msg.data` (referred to as `_msgData()`) by appending additional calldata. 
 
 ## Motivation
 
--- a/EIPS/eip-2771.md
+++ b/EIPS/eip-2771.md
@@ -12,9 +12,7 @@ created: 2020-07-01
 
 ## Abstract
 
-This EIP defines a minimal contract-level protocol that a compliant `Recipient` contract needs to support in order to be capable of accepting a meta transaction through a compliant `Forwarder` contract that it trusts to help it identify the address of the Transaction Signer. 
-
-No EVM-level protocol changes are proposed or required.
+This EIP defines a contract-level protocol for `Recipient` contracts to accept meta-transactions through trusted `Forwarder` contracts. No protocol changes are made. `Recipient` contracts fetch the effective `msg.sender` (`_msgSender()`) and `msg.data` (`_msgData()`) through. 
 
 ## Motivation
 
@@ -146,9 +144,9 @@ contract RecipientExample {
 
 ## Security Considerations
 
-A bad forwarder may allow forgery of the `msg.sender` returned from `_msgSender()` and allow transactions to appear to be coming from any address. 
+A malicious forwarder may forge the value of `_msgSender()` and effectively send transactions from any address. Therefore, `Recipient` contracts must be very careful in trusting forwarders. If a forwarder is upgradeable, then one must also trust that the contract won't perform a malicious upgrade.
 
-This means a recipient contract should be very careful which forwarder it trusts and whether this can be modified. The power to change the forwarder trusted by a recipient is equivalent to giving full control over the contract. If this kind of control over the recipient is acceptable, it is recommended that only the owner of the recipient contract be able to modify which forwarder is trusted. Otherwise best to leave it unmodifiable, as in the example above.
+In addition, modifying which forwarders are trusted must be restricted, since an attacker could "trust" their own address to forward transactions, and therefore be able to forge transactions. It is recommended to have the list of trusted forwarders be immutable, and if this is not feasible, then only trusted contract owners should be able to modify it.
 
 ## Copyright
 
--- a/EIPS/eip-2771.md
+++ b/EIPS/eip-2771.md
@@ -12,20 +12,11 @@ created: 2020-07-01
 
 ## Abstract
 
-This EIP defines a contract-level protocol for `Recipient` contracts to accept meta-transactions through trusted `Forwarder` contracts. No protocol changes are made. `Recipient` contracts fetch the effective `msg.sender` (`_msgSender()`) and `msg.data` (`_msgData()`) through. 
+This EIP defines a contract-level protocol for `Recipient` contracts to accept meta-transactions through trusted `Forwarder` contracts. No protocol changes are made. `Recipient` contracts fetch the effective `msg.sender` (`_msgSender()`) and `msg.data` (`_msgData()`) through custom getter functions. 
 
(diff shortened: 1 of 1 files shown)


Amxx:
I'd love to have the cat herder's opinion on that.

I remember taking care of EIP-1967 finalization (including late updates and the PEEPanEIP call) without being added as an author.

Amxx:
[EIP-1](https://eips.ethereum.org/EIPS/eip-1) says:

>  Transferring EIP Ownership
>
>It occasionally becomes necessary to transfer ownership of EIPs to a new champion. In general, we’d like to retain the original author as a co-author of the transferred EIP, but that’s really up to the original author. A good reason to transfer ownership is because the original author no longer has the time or interest in updating it or following through with the EIP process, or has fallen off the face of the ‘net (i.e. is unreachable or isn’t responding to email). A bad reason to transfer ownership is because you don’t agree with the direction of the EIP. We try to build consensus around an EIP, but if that’s not possible, you can always submit a competing EIP.
>
> If you are interested in assuming ownership of an EIP, send a message asking to take over, addressed to both the original author and the EIP editor. If the original author doesn’t respond to the email in a timely manner, the EIP editor will make a unilateral decision (it’s not like such decisions can’t be reversed :)).

My interpretation of that (other co-authors may disagree) is that championing the EIP doesn't make you an author unless you actually participate in the EIP content (by fixing issues, adding details, ...)

Pandapip1:
> My interpretation of that (other co-authors may disagree) is that championing the EIP doesn't make you an author unless you actually participate in the EIP content (by fixing issues, adding details, ...)

I do believe that my contributions (see the [files tab](https://github.com/ethereum/EIPs/pull/5607/files)) are sufficient for this purpose. In addition, championing an EIP generally has been considered an acceptable condition for coauthorship.

Pandapip1:
> I remember taking care of EIP-1967 finalization (including late updates and the PEEPanEIP call) without being added as an author.

Feel free to retroactively add yourself as an author - I definitely think that qualifies you for it!

Pandapip1:
> Note: no need to worry about 1967.

Too late, you've already been added as a co-author :)

Pandapip1:
@Amxx - any more updates?

forshtat:
> > My interpretation of that (other co-authors may disagree) is that championing the EIP doesn't make you an author unless you actually participate in the EIP content (by fixing issues, adding details, ...)
> 
> I do believe that my contributions (see the [files tab](https://github.com/ethereum/EIPs/pull/5607/files)) are sufficient for this purpose. In addition, championing an EIP generally has been considered an acceptable condition for coauthorship.

Hello @Pandapip1 

Thanks for the PR.

May I ask, what was changed in the content of the EIP itself?

The github diff is showing every line has changed, but that seems to be an issue with newlines, whitespaces and lines moved around. This makes it hard to review the changes, if there are any.

Pandapip1:
Mostly, content was moved around, and some sections were rewritten:

- The simple summary section was removed and shortened to fit in the description preamble entry
- The newlines from all sections were stripped, and there was some minor rephrasing throughout (nothing that changed the actual content)
- The reference implementation was removed, as it is contained in an external repository (which is no longer allowed). This should be re-added later.
- The example flow in the specification was moved into its own subsection
- The definitions were moved into their own subsection
- `isTrustedForwarder` was changed to `view` (this is an actual change to the spec!)
- The behavior of `isTrustedForwarder` was not defined. A definition was added.
- Removed the 50k gas limit, as EIPs cannot put fixed gas limits in place because gas costs can change
- Moved the rationale section (it was in the wrong order)
- Moved the example recipient from the specification to the reference implementation

Pandapip1:
The conditions for authorship are also a lot less strict for EIPs than for academia in general. Helping to move an EIP through the EIP process and making minor changes (which is what this EIP does) does warrant co-authorship (see https://github.com/ethereum/EIPs/pull/5621). Since this EIP has been stagnant for >6 months, the "following through with the EIP process" portion of transferring authorship pretty uncontroversially applies here.

Vladisvob (approved):

xinbenlv:
Here is my general views of _this type of request_:
1. EIP adoption shall have existing author approval
2. The title "author" IMHO are reserved for a role that contribute to the content of a work. Purely editorial change, e.g. styling or trimming, doesn't by default be recognized as author, unless recognized by existing authors. 

For _this particular request_, here is my view:
1. I am in full support of reviving this EIP. 
2. For whether to add @Pandapip1  author, I'd like to respect existing authors' decision.
3. Existing authors shall also note that the content of EIP-2771 _was_ published under CC0 which means if this EIP lacks activity, @Pandapip1  or any other person shall be entitled to a freedom of starting a competing EIP, IMHO even with the exact same content, except that the competing EIP will have a new EIP number.

Pandapip1:
> IMHO even with the exact same content

Actually, legally with the exact same content.

I would probably make a few tweaks, primarily trimming and rewording some of the non-normative sections. Most of those changes are in this PR.

EIPS/eip-2771.md
```diff
@@ -1,147 +1,110 @@
 ---
 eip: 2771
 title: Secure Protocol for Native Meta Transactions
-author: Ronan Sandford (@wighawag), Liraz Siri (@lirazsiri), Dror Tirosh (@drortirosh), Yoav Weiss (@yoavw), Alex Forshtat (@forshtat), Hadrien Croubois (@Amxx), Sachin Tomar (@tomarsachin2271), Patrick McCorry (@stonecoldpat), Nicolas Venturo (@nventuro), Fabian Vogelsteller (@frozeman)
+description: A contract interface for receiving meta transactions through a trusted forwarder
+author: Ronan Sandford (@wighawag), Liraz Siri (@lirazsiri), Dror Tirosh (@drortirosh), Yoav Weiss (@yoavw), Alex Forshtat (@forshtat), Hadrien Croubois (@Amxx), Sachin Tomar (@tomarsachin2271), Patrick McCorry (@stonecoldpat), Nicolas Venturo (@nventuro), Fabian Vogelsteller (@frozeman), Pandapip1 (@Pandapip1)
```
xinbenlv, line 5: Defer to authors to decide.

wighawag: personally I am fine with this change. I am glad someone took the lead to improve the wordings with the goal to make the EIP final

Pandapip1: @wighawag would you mind approving this PR? https://github.com/ethereum/EIPs/pull/5607/files#submit-review

EIPS/eip-2771.md
```diff
@@ -1,147 +1,110 @@
...
-
 ## Abstract
 
-This ERC defines a minimal contract-level protocol that a compliant Recipient
-contract needs to support in order to be capable of accepting a meta
-transaction through a compliant Forwarder contract that it trusts to help it
-identify the address of the Transaction Signer. 
-
-No EVM-level protocol changes are proposed or required.
+This EIP defines a contract-level protocol for `Recipient` contracts to accept meta-transactions through trusted `Forwarder` contracts. No protocol changes are made. `Recipient` contracts fetch the effective `msg.sender` (`_msgSender()`) and `msg.data` (`_msgData()`) through custom getter functions.
```
wighawag, line 15: Getter function is an implementation details, we should mention that the msg sender is encoded at the end of the calldata

wighawag (changes requested):

wighawag (approved):

Issue #1238, opened 2018-07-20
ERC1238: Non-transferrable Non-Fungible Tokens (NTT)


nicola:
```
eip: 1238
title: Non-transferrable Non-Fungible Tokens (badges)
author: Nicola Greco (and future others)
status: WIP
category: ERC
created: 2018-07-20
```

# Non-transferable tokens

## Simple Summary
A *badge* is a token that once assigned it cannot be transferred. Badges can be accumulated through time and put at stake. Simply speaking, badges are statements about a public key, they can be quantitative (e.g. reputation, experience) or qualitative (badges, titles).

## Abstract
The *Non-transferrable token* standard defines a set of standard APIs allowing the identification of statements (called *badges*) attributed to a public key, such that different dapps and smart contract can use to filter users or to provide user with different badges different experiences. More importantly, this standard defines a way for which users can put their badges at stake. Badges cannot be transferred but can be lost (after staking) or can expire.

## Motivation
The inspiration for designing badges comes from a conversation with Andrew Miller. The idea is to have a token that once assigned it cannot be transferred (like reputation) and that it can be used by websites, or contracts to make me perform some actions. For example, if a user accumulates paper submissions at conferences, then they can use their paper badges to request grants. It's important that they can never share these badges.

This is the equivalent of a variety of other use cases
- Accumulating badges: degrees from academic institutions, paper publications for conferences, badges that allow for access in a building or in a special chat.
- Experience points: points accumulated in a game, points accumulated by being honest in a decentralized system for some time
- Statements: more broadly, any statement that is issued or signed by a contract, a dao, a single user that requires to be on-chain
- Subscription: badges can represent the validity of a paid subscription

## Requirements for specification
- **Quantity**: a badge can have the form of 1-many, meaning that a badge can represent a single value (Degree from CS, 1), or can a quantity (Experience in Fortnite, 2000)
- **Stake**: it should be simple to put at stake contingent on a smart contract execution (however, it is understandable that this could be done by a different standard)
- **Backwards compatibility**: This standard should have similar apis to existing standards (e.g. #20, #721), I makes sense for an NFT viewer to also view badges
- **Off-chain tokens**: This specification should work also for tokens that are not issued on chain (say by a third party signer) and can be presented on chain only when necessary (it's understandable that this is an edge case and could fall off the spec, but it's an interesting one!)

## Next steps
There are two main steps here:
- Specification
- Implementation

[...]

PR #4966, opened 2022-04-01
Add EIP-1238.md

TimDaub:
This is a "DRAFT" proposal on how #1238 could be implemented.

PR #5617, opened 2022-09-07
Add EIP-1238: Non-transferable Token Standard

ra-phael:
EIP for https://github.com/ethereum/EIPs/issues/1238


ra-phael:
@Pandapip1 isn't it possible to keep the same EIP number? 🙏 

It's not taken, would be simpler in terms of historical continuity, not to mention all the renaming in:
- [https://github.com/violetprotocol/ERC1238-token](https://github.com/violetprotocol/ERC1238-token)
- [https://github.com/violetprotocol/erc1238-extendable](https://github.com/violetprotocol/erc1238-extendable)
- [https://erc1238.notion.site](https://erc1238.notion.site)/

ra-phael:
@Pandapip1 Why not?
The following comment suggests that this number can be kept with the approval of @nicola, the original author of the issue, which I have:
https://github.com/ethereum/EIPs/pull/4966#discussion_r846704349

nicola:
I approve the re-use of the original EIP number.

Pandapip1:
Again, EIP editors are responsible for assigning numbers. I will not assign EIP-1238, but I'll CC @SamWilsn @gcolvin @lightclient @axic in case they might.

SamWilsn:
I am fine with resurrecting the original EIP number. This PR is taking the old format (issue) and turning it into a markdown file, and adding authors. I'd probably recommend adding @nicola's github username.

You'll still need to fix the `eipw` errors though.

ra-phael:
Thank you @SamWilsn. I added Nicola's github username and fixed the EIP validation errors.

EIPS/eip-1238.md
```diff
@@ -0,0 +1,211 @@
...
+The idea is to have tokens that once assigned cannot be transferred (like reputation) and that can be used by websites, or contracts to make users perform some actions. For example, if a user accumulates paper submissions at conferences, then they can use their paper badges to request grants. It's important that they can never share these badges.
+
+This is the equivalent of a variety of other use cases
+
+- Accumulating badges: degrees from academic institutions, paper publications for conferences, badges that allow for access in a building or in a special chat.
+- Experience points: points accumulated in a game, points accumulated by being honest in a decentralized system for some time or a DAO.
+- Statements: more broadly, any statement that is issued or signed by a contract, a dao, a single user that requires to be on-chain.
+- Subscription: badges can represent the validity of a paid subscription.
+
+Additionally, this standard could be used as a primitive for Soulbound tokens, whereby the recovery of a Soul would be done by obtaining signatures from co-owners of tokens with the same id.
```
xinbenlv, line 31: Since there has been at least 4 attempts to propose a soul-bound / non-transferable NFT, I'd also suggest you add to your motivation and rationale on what this ERC has better merit than other competing ERCs e.g.  https://eips.ethereum.org/EIPS/eip-5114
https://eips.ethereum.org/EIPS/eip-5633

ra-phael: Good point @xinbenlv! I'm familiar with the other EIPs, but they either build on previous standards like EIP-1155 as extensions which means unecessary transfer functions or propose different designs leading to a different set of features. I added more in motivation: https://github.com/ethereum/EIPs/pull/5617/commits/a7f94be9088885831bf267fa4a48e263da684f39

EIPS/eip-1238.md
```diff
@@ -0,0 +1,213 @@
...
+interface IERC1238 is IERC165 {
+    /**
+     * @dev Emitted when `amount` tokens of token type `id` are minted to `to` by `minter`.
+     */
+    event MintSingle(address indexed minter, address indexed to, uint256 indexed id, uint256 amount);
+
+    /**
+     * @dev Equivalent to multiple {MintSingle} events, where `minter` and `to` is the same for all token types.
+     */
+    event MintBatch(address indexed minter, address indexed to, uint256[] ids, uint256[] amounts);
```
SamWilsn, line 48: Why do you have both `MintSingle` and `MintBatch`? Couldn't `MintBatch` handle both cases?

If there's a reason, please add it to the EIP's rationale section.

EIPS/eip-1238.md
```diff
@@ -0,0 +1,213 @@
...
+     * @dev Equivalent to multiple {BurnSingle} events, where `owner` and `burner` is the same for all token types.
+     */
+    event BurnBatch(address indexed burner, address indexed owner, uint256[] ids, uint256[] amounts);
+
+    /**
+     * @dev Returns the amount of tokens of token type `id` owned by `account`.
+     *
+     * Requirements:
+     *
+     * - `account` cannot be the zero address.
```
SamWilsn, line 65: Why can't the zero address hold tokens? Is that a hard requirement for all implementations? Again, probably worth putting in the Rationale.

EIPS/eip-1238.md
```diff
@@ -0,0 +1,213 @@
...
+     * of tokens which gets transferred between their respective balances
+     * of tokens held.
+     */
+    function entrust(
+        address to,
+        uint256 id,
+        uint256 amount
+    ) external;
+}
+```
```
SamWilsn, line 172: This holdable extension needs a lot more explanation. You should define the concepts, the expected flow, etc.

nicola (approved):

EIPS/eip-1238.md
ra-phael: Good point, I like how it makes the interface simpler. I also applied the same reasoning to the events related to burning.

EIPS/eip-1238.md
ra-phael: Added more in the Rationale section.
I'm thinking about the consequences if a balance check is used for gating access and there's a bug and the zero address is passed instead of the user's address...

EIPS/eip-1238.md
ra-phael: I added more explanations in https://github.com/ethereum/EIPs/pull/5617/commits/a17e69dd6a6a2301d53ba01fdcc93f694aaa95c7

ra-phael:
@SamWilsn Thank you for your feedback! And sorry for the delay in addressing it. It's ready for re-review.

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-09-07
+requires: 165, 712
+---
+
+## Abstract
+
+A _badge_ or non-transferable token (_NTT_) is a token that cannot be transferred once assigned. Badges can be accumulated through time and put at stake. Simply speaking, badges are statements about a public key: they can be quantitative (e.g. reputation, experience) or qualitative (badges, titles).
```
SamWilsn, line 16: > Badges can be accumulated through time and put at stake.

I am not sure what "put at stake" means here. Is this some kind of staking mechanism, or are you talking about the badge issuer's reputation, or something else entirely?

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+
+The Non-Transferable Token standard defines a set of standard APIs allowing the identification of statements (called badges) attributed to a public key, such that different dapps and smart contract can use to filter users or to provide user with different badges different experiences. More importantly, this standard defines a way for which users can put their badges at stake. Badges cannot be transferred but can be lost (after staking) or can expire.
```
SamWilsn, line 18: Couple grammar and stylistic changes:

```suggestion
The Non-Transferable Token standard defines a set of APIs that identify statements (called badges) attributed to a public key, such that different dapps and smart contracts can filter users or tailor experiences based on what badges a user has. More importantly, this standard defines a way for users to put their badges at stake. Badges cannot be transferred but can be lost (after staking) or can expire.
```

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Motivation
+
+The idea is to have tokens that once assigned cannot be transferred (like reputation) and that can be used by websites, or contracts to make users perform some actions. For example, if a user accumulates paper submissions at conferences, then they can use their paper badges to request grants. It's important that they can never share these badges.
```
SamWilsn, line 22: I don't think a badge can _force_ a user to do something :P

```suggestion
The idea is to have tokens that once assigned cannot be transferred (like reputation) and that can be used by websites, or contracts to allow users to perform some actions. For example, if a user accumulates paper submissions at conferences, then they can use their paper badges to request grants. It's important that they can never share these badges.
```

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+
+This is the equivalent of a variety of other use cases
```
SamWilsn, line 24:
```suggestion
This is the equivalent of a variety of other use cases:
```

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+     * @dev Emitted when some `amounts` of tokens with their respective `ids` are burned by `burner`.
+     */
+    event Burned(address indexed burner, address indexed owner, uint256[] ids, uint256[] amounts);
+
+    /**
+     * @dev Returns the amount of tokens of token type `id` owned by `account`.
+     *
+     * Requirements:
+     *
+     * - `account` cannot be the zero address.
```
SamWilsn, line 56: Coming back to this, I think you should be more specific about what behaviour you expect. Perhaps:

```suggestion
     * - `balanceOf(address(0x0))` MUST revert.
```

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+     */
+    function balanceOf(address account, uint256 id) external view returns (uint256);
+
+    /**
+     * @dev Returns the balance of `account` for a batch of token `ids`.
+     *
```
SamWilsn, line 62: You might want to repeat the non-zero requirement here.

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+interface IERC1238Holder is IERC1238Receiver {
+    /**
+     * @dev This function is called when tokens with id `id` are burnt.
+     * Returns `true` as a sign that the burn was acknowledged and processed.
+     */
+    function onBurn(uint256 id, uint256 amount) external returns (bool);
+}
+```
+
+When burning tokens, implementers MUST try to call the `onBurn` function if the holder is a smart contract. This allows smart contract token holders to get notified when tokens that they hold are being burnt and gives them a chance to react and handle the situation as they see fit.
```
SamWilsn, line 186: I would suggest expanding on what "try to call" means in this context. If `onBurn` reverts, is the token contract permitted to continue burning? If `onBurn` returns `false`?

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+```solidity
+/**
+ * @dev Interface proposal for contracts that need to hold ERC1238 tokens.
+ */
+interface IERC1238Holder is IERC1238Receiver {
+    /**
+     * @dev This function is called when tokens with id `id` are burnt.
+     * Returns `true` as a sign that the burn was acknowledged and processed.
+     */
+    function onBurn(uint256 id, uint256 amount) external returns (bool);
```
SamWilsn, line 182: Is there an argument for making this `payable`?

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+     */
+    function entrust(
+        address to,
+        uint256 id,
+        uint256 amount
+    ) external;
+}
+```
+
+Smart contracts holding tokens MUST implement the following interface:
```
SamWilsn, line 171: This standard really can't place requirements on smart contracts that don't claim to implement it. I'd reword this to something like:

> Smart contracts that are intended to hold NTTs SHOULD implement the following interface to receive burn notifications:

EIPS/eip-1238.md
```diff
@@ -0,0 +1,231 @@
...
+    function setBatchExpiryDates(uint256[] memory ids, uint256[] memory dates) external;
+}
+```
+
+The **Holdable extension** is OPTIONAL for smart contracts that implement this EIP. This extension enables token owners to let their tokens be held by another address, including "staking" them in a smart contract.
+Since NTTs are non-transferable, the usual way of staking tokens by transferring ownership over them is not possible. In order to make staking possible, this extension introduces the notion of token holders, in addition to token owners:
+
+- When minting tokens to Alice's address A, address A is the token owner but also the initial token holder. Alice can then call the `entrust` function to entrust address B with her tokens. Address B becomes the token holder while Alice remains the token owner.
+- Only one address can hold specific tokens at a time.
+- A token holder can transfer tokens to another holder. Still, the important point is that this does not change ownership of the tokens.
```
SamWilsn, line 138: Should this be:

```suggestion
- A token holder can entrust tokens to another holder. Still, the important point is that this does not change ownership of the tokens.
```

?

SamWilsn (commented):
How would you feel about slimming this EIP down a bit, and moving the extensions into their own proposals?

PR #5620, opened 2022-09-07
Update EIP-2718: Add 155 to requires list

Pandapip1:
CC @MicahZoltu

### DIFF
--- a/EIPS/eip-2718.md
+++ b/EIPS/eip-2718.md
@@ -8,6 +8,7 @@ status: Final
 type: Standards Track
 category: Core
 created: 2020-06-13
+requires: 155
 ---
 
 ## Abstract


MicahZoltu:
I think I would prefer to just delete line 18 instead.  It doesn't add significant value to understanding the EIP.

MicahZoltu:
Alternatively, maybe just rewording it to something like,
> This was seen in previous proposals to create new types of transactions where some new value was bit-packed into one of the encoded fields.

Pandapip1:
Actually, EIP-155 is implicitly mentioned on L35

MicahZoltu:
Assuming we are talking about the same line 35, there is no implicit mention of EIP-155:
```
`LegacyTransaction` is `rlp([nonce, gasPrice, gasLimit, to, value, data, v, r, s])`
```
EIP-155 redefines how to interpret `v`, but it doesn't have any impact on how the transaction is encoded.

Pandapip1:
> EIP-155 redefines how to interpret v, but it doesn't have any impact on how the transaction is encoded.

Fair enough.

PR #5621, opened 2022-09-07
Update EIP-1967: Add @Amxx to author list

Pandapip1:


Pandapip1:
CC @Amxx

spalladino (approved):
I'm fine with adding @Amxx, he's done some key work around this PR. I'm not sure about changing the name of the EIP after being finalized, but I'll leave that to the EIP editors around here!

PR #5649, opened 2022-09-12
Synchronize EIP-4844 cryptography with consensus specs

asn-d6:
So... at some point back in June [we updated](https://github.com/ethereum/consensus-specs/pull/2901) the consensus side of 4844 to be executable. This resulted in various fixes and improvements for the cryptographic parts of the consensus code.

These updates were never ported back to EIP land because the code was quite unstable and we wanted to minimize any duplicate overhead between EIP and consensus review processes. This means that the EIP was out of date for the past months...

These days, the 4844 cryptographic code seems relatively stable, so this commit updates the EIP. Also, with the merge coming up, more devs will start looking at 4844 code, and hence having the code be up to date is imperative.

Note that instead of updating the cryptographic functions we instead link to the relevant parts of the consensus-specs
repository. This is done to avoid code duplication between the two repositories -- an approach that seems prudent in
cross-execution-and-consensus situations like EIP-4844. I imagine that people might not like this approach, but in that case, I would appreciate feedback on alternative approaches we could take.

Cheers!

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -42,9 +42,6 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `BLOB_TX_TYPE` | `Bytes1(0x05)` |
 | `FIELD_ELEMENTS_PER_BLOB` | `4096` |
 | `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` |
-| `KZG_SETUP_G2` | `Vector[G2Point, FIELD_ELEMENTS_PER_BLOB]`, contents TBD |
-| `KZG_SETUP_LAGRANGE` | `Vector[KZGCommitment, FIELD_ELEMENTS_PER_BLOB]`, contents TBD |
-| `ROOTS_OF_UNITY` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` |
 | `BLOB_COMMITMENT_VERSION_KZG` | `Bytes1(0x01)` |
 | `POINT_EVALUATION_PRECOMPILE_ADDRESS` | `Bytes20(0x14)` |
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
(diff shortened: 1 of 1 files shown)


lightclient:
@asn-d6 could you update the `consensus-specs` links so that they link to a specific release/revision of the spec?
@SamWilsn what is the status of allowing `consensus-specs` links? Are we just waiting to update the `eipw` or is it still under discussion?

asn-d6:
> @asn-d6 could you update the `consensus-specs` links so that they link to a specific release/revision of the spec?

I was thinking that part of the appeal of this approach is that the link will always point to the latest version of the functions, and as long as the function names remain the same we will not have to change the EIP-side at all. Do you think that's too flexible?

asn-d6:
> @asn-d6 could you update the `consensus-specs` links so that they link to a specific release/revision of the spec? @SamWilsn what is the status of allowing `consensus-specs` links? Are we just waiting to update the `eipw` or is it still under discussion?

Thinking about this again, I think doing it one step at a time is fine.

I have force-pushed a new commit, that pins down a specific revision for all functions cited. This should prevent any sort of rug pulling from the consensus-specs side.

Please let me know what else I can do on this one :)

lightclient:
Hey @SamWilsn is it okay to manually merge while we wait for the walidator to accept the CL spec links, or are we still trying to come to agreement on that?

Inphi (commented):
I'll defer to maintainers on this whether it's appropriate to link to the consensus specs.
Though my 2c is that EIPs should be as self contained as possible, such that implementers can use as a reference without consulting another documents.

lightclient:
I'm going to override `eipw` here as it's been 3 weeks and have not got an update on when `eipw` will support links to the consensus specs.

The editors are still debating the specifics, but it [appears](https://github.com/ethereum/EIPs/pull/5474#issuecomment-1218139142) that links to the consensus specs which refer to a specific commit will be acceptable. In order to avoid any further confusion among implementation teams currently working on EIP-4844, I'm going to merge these changes as I have verified that each link is to a specific revision of the spec. We can continue debating the exact external link policy and we will review the current state of things as this EIP moves forward to last call.

PR #5650, opened 2022-09-13
Remove one unnecessary sentence.

will-edge:
The meaning of the removed sentence is covered by the segment <Design decision: Notification/acceptance mechanism instead of 'Safe Transfer'>, so that it is redundant and some how misleading.


Pandapip1 (approved):
This looks fine to me.

PR #5654, opened 2022-09-15
Update EIP-5192: Move to Final

TimDaub:


Pandapip1 (approved):
I think this is fine.

PR #5659, opened 2022-09-15
Add EIP-5659: Social Media URI Propagation Event

TimDaub:
Heavily inspired by EIP-3722 of @auryn-macmillan


EIPS/eip-unassigned.md
```diff
@@ -0,0 +1,47 @@
+---
+eip: unassigned
```
xinbenlv, line 2:
```suggestion
eip: 5659
```

xinbenlv: Also don't forget to update the file name.

EIPS/eip-5659.md
```diff
@@ -0,0 +1,47 @@
...
+created: 2022-09-15
+---
+
+## Abstract
+
+We introduce a minimal on-chain primitive to emit URIs using Ethereum's event log infrastructure.
+
+## Motivation
+
+The modern social network's base layer is distributing content behind Universal Resource Identifiers as defined by RFC 3986. Any web resource can be identified by an URI or a URL and so upgrading Ethereum with a standard way of distributing content is useful.
```
SamWilsn, line 19: You haven't convinced me that this EIP needs to exist. I think I'd like to see some potential use cases, both for publishers and for event consumers.

EIPS/eip-5659.md
```diff
@@ -0,0 +1,47 @@
...
+discussions-to: https://ethereum-magicians.org/t/social-media-uri-propagation-event/10893
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-09-15
+---
+
+## Abstract
+
+We introduce a minimal on-chain primitive to emit URIs using Ethereum's event log infrastructure.
```
SamWilsn, line 15: We like terse abstracts, but this is perhaps a touch too short. Perhaps include a brief summary of what kinds of URIs you expect people to publish?

EIPS/eip-5659.md
```diff
@@ -0,0 +1,47 @@
...
+The modern social network's base layer is distributing content behind Universal Resource Identifiers as defined by RFC 3986. Any web resource can be identified by an URI or a URL and so upgrading Ethereum with a standard way of distributing content is useful.
+
+## Specification
+
+```solidity
+contract Propagator {
+  event NewPost(address indexed author, string indexed uri);
+  function submit(string calldata uri) external {
+    emit NewPost(msg.sender, uri);
+  }
```
SamWilsn, line 28: Is this implementation the only valid implementation? If not, you should probably switch this to an interface and let implementers decide how to implement it.

EIPS/eip-5659.md
TimDaub: ok, how would you frame this in the context of an EIP though? E.g. I feel like pointing out the countless problems of web2 content distribution could read too political

SamWilsn: I don't think you need to bash web2 to make a case for this EIP. What URIs do you expect to publish? Who do you expect to consume them? What applications are difficult to implement without this standard? These questions, answered in your Motivation section, would make for a more compelling proposal.

PR #5664, opened 2022-09-16
Update EIP-3670: Use exceptions with error messages in reference code

gumb0:
Also add `EOF1 Forward compatibility` subsection to `Motivation`.


EIPS/eip-3670.md
```diff
@@ -91,49 +133,21 @@ def validate_code(code: bytes):
         # Ensure the opcode is valid
         opcode = code[pos]
         pos += 1
-        assert opcode in valid_opcodes
+        if not opcode in valid_opcodes:
```
xinbenlv, line 136: We should be able to use `assert` with a message directly in python3
```suggestion
        assert (not opcode in valid_opcodes), "undefined instruction"
```

Here and below

PR #5665, opened 2022-09-16
Update EIP-4200: Use exceptions with error messages in reference code

gumb0:


EIPS/eip-4200.md
```diff
@@ -172,30 +179,36 @@ def validate_code(code: bytes):
...
             rjumpdests.add(rjumpdest)
 
-        # Save positions of each byte of immediate value
+        # Save immediate value positions
         immediates.update(range(pos, pos + immediate_sizes[opcode]))
         # Skip immediates
         pos += immediate_sizes[opcode]
 
     # Ensure last opcode's immediate doesn't go over code end
-    assert pos == len(code)
```
xinbenlv, line 192: I think the assert can be used in this way

```python3
assert pos == len(code), "truncated immediate"
```

xinbenlv: Here and below. 
It should work for python3.

PR #5670, opened 2022-09-16
Update EIP-5528: Move to Last Call

StartfundInc:
Can we move EIP-5528 to Last-Call?
We do not have significant changes anymore.

### DIFF
--- a/EIPS/eip-5528.md
+++ b/EIPS/eip-5528.md
@@ -4,7 +4,7 @@ title: Refundable Fungible Token
 description: Allows refunds for EIP-20 tokens by escrow smart contract
 author: StartfundInc (@StartfundInc)
 discussions-to: https://ethereum-magicians.org/t/eip-5528-refundable-token-standard/10494
-status: status: Last Call
+status: Last Call
 type: Standards Track
 category: ERC
 created: 2022-08-16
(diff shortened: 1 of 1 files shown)


xinbenlv:
I think the format of specification section will be much more readable if it's restructured as followed:

--------------------

Any compliant contract will implement the following interface
```solidity
interface ERC5528 {
  // methods as in the "Example of Interface"
}
```

And then layout the requirement for each method.

EIPS/eip-5528.md
```diff
@@ -4,7 +4,8 @@ title: Refundable Fungible Token
 description: Allows refunds for EIP-20 tokens by escrow smart contract
 author: StartfundInc (@StartfundInc)
 discussions-to: https://ethereum-magicians.org/t/eip-5528-refundable-token-standard/10494
-status: Review
+status: Last Call
+last-call-deadline: 2022-10-16
```
xinbenlv, line 8: I left some comments below for editorial suggestion, mainly the specification section can be restructure for better readability.

Pandapip1 (approved):
I think this is fine. **Make sure to quadruple-check this before moving this EIP to final, or you risk doing an EIP-712.**

PR #5671, opened 2022-09-17
Update 4844: use clearer name for max fee value

ralexstokes:
The existing text uses the name `max_basefee_per_gas` which suggests the maximum base fee value a transaction would be willing to pay.

This value actually refers to the maximum fee per gas the transaction would be willing to pay (base fee + priority fee) and this change reflects this with what I consider a clearer name.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -13,7 +13,7 @@ requires: 1559, 2718, 2930
 
 ## Abstract
 
-Introduce a new transaction format for "blob-carrying transactions" which contain a large amount of data that cannot be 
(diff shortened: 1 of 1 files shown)


EIPS/eip-4844.md
```diff
@@ -173,7 +173,7 @@ class BlobTransaction(Container):
     chain_id: uint256
     nonce: uint64
     priority_fee_per_gas: uint256
-    max_basefee_per_gas: uint256
```
xinbenlv, line 176: LGTM, to make it consistent with 1559.

Consider also change Line 175 and else where for `priority_fee_per_gas` -> `max_priority_fee_per_gas` to match with 1559?

lightclient (approved):

PR #5508, opened 2022-08-19
CI: Upgrade EIP-Bot to eip-review-bot

Pandapip1:
Has a few advantages over EIP-Bot:

- Smaller, less systemically complex codebase
  - Maintenance and features can be developed quicker
- Runs much quicker

Progress:

- [x] Prototype
- [x] E2E testing
- [x] Automated testing

PR #5676, opened 2022-09-18
Update EIP-2535: Fix Internal Link to Reference Implementation

mudgen:
Added an internal link to a zip file of the reference implementation.
This pull request deletes unneeded reference implementation files.


mudgen:
@SamWilsn Can you please help get this pull request merged?  It is stuck.

mudgen:
@Pandapip1 do you know what caused this pull request to get stuck?

Pandapip1:
> @Pandapip1 do you know what caused this pull request to get stuck?

EIP-Bot bug. You probably deleted a file.

mudgen:
> > @Pandapip1 do you know what caused this pull request to get stuck?
> 
> EIP-Bot bug. You probably deleted a file.

That is correct.  I deleted EIP2535 reference implementation files that were not needed anymore.  Are you able to merge the pull request?  Or only @SamWilsn?

Pandapip1:
> Or only @SamWilsn?

Only Sam is able to merge this PR. https://github.com/ethereum/EIPs/pull/5508 has the downside of being new, but has the upside of being improved - and fixing that stupid bug!

SamWilsn (changes requested):
May I ask why you switched to a ZIP archive for the reference implementation? We prefer including reference implementations as regular source files.

---

The license file included in your reference implementation doesn't match the `package.json`.

---

I may have missed it, but it seems like this version omits the note I had added about "implementing" meaning either directly implementing the interface, or adding a facet with the correct external functions. As I understand the current version, `IDiamondLoupe`/`IDiamond`/`IDiamondCut` must be directly implemented (in the Solidity sense) by the diamond. Just wanted to make sure that everything is correct as written in this PR.

mudgen:
I switched to the ZIP because otherwise I cannot link to multiple files with one link.

If I link to Diamond.sol then people will download that but that is just one part of the reference implementation. Most of the reference implementation is in other files.


I will get the license fixed in the package.json

IDiamondLoupe/IDiamond/IDiamondCut functions can be implemented in facets and that is how all the reference implementations do it.  But they could also be implemented directly in the diamond. I'll add the section about this back.

Pandapip1:
> I switched to the ZIP because otherwise I cannot link to multiple files with one link.

FYI - this is generally considered bad form. Just create an index.html page that then links to each file individually.

mudgen:
> I may have missed it, but it seems like this version omits the note I had added about "implementing" meaning either directly implementing the interface, or adding a facet with the correct external functions.

@SamWilsn You are right about this.  And **Overview** and **Fallback Function** were missing too. This was a mistake. I added these back and added the *A Note on Implementing Interfaces* section back too.  Thanks for catching this!

mudgen:
> The license file included in your reference implementation doesn't match the package.json.
This is now fixed in this pull request.

mudgen:
> > I switched to the ZIP because otherwise I cannot link to multiple files with one link.
> 
> FYI - this is generally considered bad form. Just create an index.html page that then links to each file individually.

If you go to https://eips.ethereum.org/EIPS/eip-2535#reference-implementation and click on "Diamond.sol" you will see what happens when someone downloads a Solidity directly. That works fine when downloading a single Solidity file.

But the implementation is a set of files in a specific directory structure.  If users download each Solidity file one by one form an HTML file the files won't go together because the directory structure will be wrong. 

However if a user downloads a zip file they can unzip it and all the file will be there in the correct directory structure.  Unless there is another alternative to using a zip file to keep the downloaded files in the right directory structure, I'd like to use the zip file.

mudgen:
@SamWilsn This pull request is ready to be merged now.

SamWilsn:
I do see your point @mudgen, but this would be the first zipped reference implementation, and your arguments haven't convinced me to break precedent.

Three options that have worked for authors in the past are:

 - Condense the reference implementation into a single file (reference implementations should be minimal, so this is a popular option.)
 - Link to `README.md` from the EIP, and link to individual files from `README.md`.
 - Link to each file directly from the EIP.

If you'd like, I'd be happy to author and push a commit implementing one of the above options of your choice.

Not that you've suggested it, but linking to GitHub's directory listings is not allowed.

---

We are exploring options for linking to directories, but it'll be a while before we come to a solution there. You could wait, but I wouldn't recommend it. If you think the EIP Process is tedious and pedantic, you should see the EIP Improvement Process :rofl:

mudgen:
@SamWilsn

I updated this pull request so that the reference implementation section does link to a single Solidity file that has all the Solidity code of the reference implementation. So this requirement is met.

However, as you will see the section also includes a link to the zip file reference implementation.  Please here me out about this. The reference implementation is the same in both. So the same reference implementation is given in two formats:  single file, and multi-file with directories.

Here me out:

The directory structure and organization of the reference implementation is important.  The purpose of this standard is to help people build larger, multi-contract systems that can be extended after deployment, and achieve tooling interoperability with it.  The zip file shows people how to organize and structure their code so the purpose of the standard can be achieved. The zip file also includes a deployment script and a test script so they can see how a diamond can be properly deployed and tested. Without these some people may not be able to use this standard as there will not be enough information about how to do it.

Please allow this zip file for now, as a temporary solution.  The single file Solidity file serves as an alternative in case there is some issue concerning using a zip file, and complies with the existing convention of providing implementations in EIPs.

In the future if there is a new solution for including directory structure in reference implementations then this EIP can be changed to use that instead of the zip file.

Pandapip1:
I am okay with this format. I approve this.

EIPS/eip-2535.md
```diff
@@ -345,7 +391,9 @@ This standard makes upgradeable diamonds compatible with future standards and fu
 
 ## Reference Implementation
 
-See [`Diamond.sol`](../assets/eip-2535/reference/contracts/Diamond.sol).
+All the Solidity code for a complete reference implementation has been put in a single file here: [Diamond.sol](../assets/eip-2535/reference/Diamond.sol)
+
+The same reference implementation has been organized into multiple files and directories and also includes a deployment script and tests. Download it as a zip file: [EIP-2535-Diamonds-Reference-Implementation.zip](../assets/eip-2535/reference/EIP-2535-Diamonds-Reference-Implementation.zip)
```
SamWilsn, line 396:
```suggestion
The same reference implementation has been organized into multiple files and directories and also includes a deployment script and tests. Download it as a zip file: [EIP2535-Diamonds-Reference-Implementation.zip](../assets/eip-2535/reference/EIP2535-Diamonds-Reference-Implementation.zip)
```

SamWilsn (changes requested):
I can't comment on it directly but on line 320/365 you have `Solidity` but it needs to be `solidity` (lower case `s`.)

EIPS/eip-2535.md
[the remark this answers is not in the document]
mudgen: Understand.  Yes

mudgen:
> I can't comment on it directly but on line 320/365 you have `Solidity` but it needs to be `solidity` (lower case `s`.)

Thanks, this is now fixed.

EIPS/eip-2535.md
```diff
@@ -345,7 +391,9 @@ This standard makes upgradeable diamonds compatible with future standards and fu
 
 ## Reference Implementation
 
-See [`Diamond.sol`](../assets/eip-2535/reference/contracts/Diamond.sol).
+All the Solidity code for a complete reference implementation has been put in a single file here: [Diamond.sol](../assets/eip-2535/reference/Diamond.sol)
+
+The same reference implementation has been organized into multiple files and directories and also includes a deployment script and tests. Download it as a zip file: [EIP2535-Diamonds-Reference-Implementation.zip](../assets/eip-2535/reference/EIP2535-Diamonds-Reference-Implementation.zip)
```
SamWilsn, line 396:
```suggestion
The same reference implementation has been organized into multiple files and directories and also includes a deployment script and tests. Download it as a zip file: [`EIP2535-Diamonds-Reference-Implementation.zip`](../assets/eip-2535/reference/EIP2535-Diamonds-Reference-Implementation.zip)
```

SamWilsn (approved):

PR #5680, opened 2022-09-18
Update EIP-1: Fix Backward Compatibility Typo

Pandapip1:
This has been bugging me for a while. Thoughts?


xinbenlv (dismissed):

SamWilsn:
I think they're both acceptable, and prefer the option that doesn't require changing `eipw` :rofl:

Pandapip1:
> I think they're both acceptable, and prefer the option that doesn't require changing eipw 🤣

Should we add a [sic] then?

Pandapip1:
> both are technically correct

Every spell checker I've used disagrees.

I can change the 5000+ files if need be. I think we can also just keep the old files as having the old form until they are changed. I volunteer to try to fix eipw.

xinbenlv:
I am ok ether way. I just hope it be consistent across the board....

Pandapip1:
> Although this is the more accepted form, both are technically correct. I think that it's best to leave as-is so that we don't need to change 5000+ EIPs and eipw.

There are <1000 EIPs, and it's possible to automate this in a few minutes (see my change with the LICENSE file)

PR #5707, opened 2022-09-22
Update EIP-4844: Fee Market Update

adietrichs:
Update to the EIP-4844 fee market design. List of changes:

- [x] introduce data gas
- [x] add `max_fee_per_data_gas` tx field
- [x] change fake exponential function to taylor expansion
- [ ] ~~add min data gasprice of 10,000 Gwei~~ reverted, tbd in separate PR
- [x] remove max blobs per tx restriction
- [x] align tx field naming with EIP-1559
- [x] change `DATA_GAS_PER_BLOB` to make it easier to add calldata pricing later
- [x] specify that data gas fees are burned

Potential remaining todos:

- [ ] ~~make excess blobs calculation slot-based instead of block-based~~ will be a separate PR or EIP
- [ ] ~~mention option for mempool to enforce a max blobs per tx restriction~~ solved by EIP-5793
- [ ] ~~change data gas name to something more unique (oil? mana?)~~ data gas it is
- [ ] ~~move data gas from static to dynamic tracking~~ too opinionated for now
- [ ] ~~update EIP structure and rationale to better explain data gas~~ potentially in a later PR

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -274,12 +274,14 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
 
 ### Gas price of blobs (Simplified version)
 
+***WARNING:` This is only for testing***
+
 For early draft implementations, we simply change `get_blob_gas(parent)` to always return `SIMPLE_GAS_PER_BLOB`.
 
 ### Gas accounting (Full version)
 
 We introduce data gas as a new type of gas. It is independent of normal gas and follows its own targeting rule, similar to EIP-1559.
-We use the `excess_blobs` header field to store persistent data needed to compute the data gas price. For now, only blobs are charged for in data gas. In the future this could be extended to also cover calldata (at a different relative gas cost).
+We use the `excess_blobs` header field to store persistent data needed to compute the data gas price. For now, only blobs are priced in data gas. In the future this could be extended to also cover calldata (at a different relative gas cost).
 
 ```python
 def calc_data_fee(tx: SignedBlobTransaction, parent: Header) -> int:
@@ -435,7 +437,7 @@ where `excess_blobs` is the total "extra" number of blobs that the chain has acc
 Like EIP-1559, it's a self-correcting formula: as the excess goes higher, the `data_gasprice` increases exponentially, reducing usage and eventually forcing the excess back down.
 
 The block-by-block behavior is roughly as follows.
-If in block `N`, `data_gasprice = G1`, and block `N` has `X` blobs, then in block `N+1`, `excess_blobs` increases by `X - TARGET_BLOBS_PER_BLOCK`,
+If block `N` contains `X` blobs, then in block `N+1` `excess_blobs` increases by `X - TARGET_BLOBS_PER_BLOCK`,
 and so the `data_gasprice` of block `N+1` increases by a factor of `e**((X - TARGET_BLOBS_PER_BLOCK) / DATA_GASPRICE_UPDATE_FRACTION)`.
 Hence, it has a similar effect to the existing EIP-1559, but is more "stable" in the sense that it responds in the same way to the same total usage regardless of how it's distributed.
 
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -47,7 +47,7 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
 | `MAX_BLOBS_PER_BLOCK` | `16` |
 | `TARGET_BLOBS_PER_BLOCK` | `8` |
-| `MIN_DATA_GASPRICE` | `1` |
+| `MIN_DATA_GASPRICE` | `10**13` |
 | `DATA_GASPRICE_UPDATE_FRACTION` | `84` |
 | `MAX_VERSIONED_HASHES_LIST_SIZE` | `2**24` |
 | `MAX_CALLDATA_SIZE` | `2**24` |
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -293,7 +293,7 @@ def get_data_gasprice(header: Header) -> int:
     return fake_exponential(
         MIN_GASPRICE_PER_BLOB,
         header.excess_blobs,
-        GASPRICE_UPDATE_FRACTION_PER_BLOB
+        DATA_GASPRICE_UPDATE_FRACTION
     )
 ```
 
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -46,7 +46,7 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `POINT_EVALUATION_PRECOMPILE_ADDRESS` | `Bytes20(0x14)` |
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
 | `MAX_BLOBS_PER_BLOCK` | `16` |
-| `TARGET_BLOBS_PER_BLOCK` | `8` |
+| `TARGET_DATA_GAS_PER_BLOCK` | `2**20` |
 | `MIN_DATA_GASPRICE` | `10**8` |
 | `DATA_GASPRICE_UPDATE_FRACTION` | `68` |
 | `MAX_VERSIONED_HASHES_LIST_SIZE` | `2**24` |
@@ -217,10 +217,10 @@ The value of `excess_data_gas` can be calculated using the parent header and num
 ```python
 def calc_excess_data_gas(parent: Header, new_blobs: int) -> int:
     consumed_data_gas = new_blobs * DATA_GAS_PER_BLOB
-    if parent.excess_data_gas + consumed_data_gas < TARGET_BLOBS_PER_BLOCK * DATA_GAS_PER_BLOB:
+    if parent.excess_data_gas + consumed_data_gas < TARGET_DATA_GAS_PER_BLOCK:
         return 0
     else:
-        return parent.excess_data_gas + consumed_data_gas - TARGET_BLOBS_PER_BLOCK * DATA_GAS_PER_BLOB
+        return parent.excess_data_gas + consumed_data_gas - TARGET_DATA_GAS_PER_BLOCK
 ```
 
 ### Beacon chain validation
@@ -434,12 +434,12 @@ sequencers would simply have to switch over to using a new transaction type at t
 ### Data gasprice update rule
 
 The data gasprice update rule is intended to approximate the formula `data_gasprice = MIN_DATA_GASPRICE * e**(excess_data_gas / DATA_GASPRICE_UPDATE_FRACTION)`,
-where `excess_data_gas` is the total "extra" amount of data gas that the chain has consumed relative to the "targeted" number (`TARGET_BLOBS_PER_BLOCK * DATA_GAS_PER_BLOB` per block).
+where `excess_data_gas` is the total "extra" amount of data gas that the chain has consumed relative to the "targeted" number (`TARGET_DATA_GAS_PER_BLOCK` per block).
 Like EIP-1559, it's a self-correcting formula: as the excess goes higher, the `data_gasprice` increases exponentially, reducing usage and eventually forcing the excess back down.
 
 The block-by-block behavior is roughly as follows.
-If block `N` consumes `X` data gas, then in block `N+1` `excess_data_gas` increases by `X - TARGET_BLOBS_PER_BLOCK * DATA_GAS_PER_BLOB`,
-and so the `data_gasprice` of block `N+1` increases by a factor of `e**((X - TARGET_BLOBS_PER_BLOCK * DATA_GAS_PER_BLOB) / DATA_GASPRICE_UPDATE_FRACTION)`.
+If block `N` consumes `X` data gas, then in block `N+1` `excess_data_gas` increases by `X - TARGET_DATA_GAS_PER_BLOCK`,
+and so the `data_gasprice` of block `N+1` increases by a factor of `e**((X - TARGET_DATA_GAS_PER_BLOCK) / DATA_GASPRICE_UPDATE_FRACTION)`.
 Hence, it has a similar effect to the existing EIP-1559, but is more "stable" in the sense that it responds in the same way to the same total usage regardless of how it's distributed.
 
 The parameter `DATA_GASPRICE_UPDATE_FRACTION` controls the maximum rate of change of the blob gas price. It is chosen to target a maximum change rate of 12.5% per block.
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -274,7 +274,7 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
 
 ### Gas price of blobs (Simplified version)
 
-***WARNING:` This is only for testing***
+***WARNING: This is only for testing***
 
 For early draft implementations, we simply change `get_blob_gas(parent)` to always return `SIMPLE_GAS_PER_BLOB`.
 
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -48,7 +48,7 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `MAX_BLOBS_PER_BLOCK` | `16` |
 | `TARGET_BLOBS_PER_BLOCK` | `8` |
 | `MIN_DATA_GASPRICE` | `10**13` |
-| `DATA_GASPRICE_UPDATE_FRACTION` | `84` |
+| `DATA_GASPRICE_UPDATE_FRACTION` | `68` |
 | `MAX_VERSIONED_HASHES_LIST_SIZE` | `2**24` |
 | `MAX_CALLDATA_SIZE` | `2**24` |
 | `MAX_ACCESS_LIST_SIZE` | `2**24` |
@@ -441,7 +441,7 @@ If block `N` contains `X` blobs, then in block `N+1` `excess_blobs` increases by
 and so the `data_gasprice` of block `N+1` increases by a factor of `e**((X - TARGET_BLOBS_PER_BLOCK) / DATA_GASPRICE_UPDATE_FRACTION)`.
 Hence, it has a similar effect to the existing EIP-1559, but is more "stable" in the sense that it responds in the same way to the same total usage regardless of how it's distributed.
 
-The parameter `DATA_GASPRICE_UPDATE_FRACTION` controls the maximum rate of change of the blob gas price. It is chosen to target a maximum change rate of 10% per block.
+The parameter `DATA_GASPRICE_UPDATE_FRACTION` controls the maximum rate of change of the blob gas price. It is chosen to target a maximum change rate of 12.5% per block.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -45,7 +45,7 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `BLOB_COMMITMENT_VERSION_KZG` | `Bytes1(0x01)` |
 | `POINT_EVALUATION_PRECOMPILE_ADDRESS` | `Bytes20(0x14)` |
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
-| `MAX_BLOBS_PER_BLOCK` | `16` |
+| `MAX_DATA_GAS_PER_BLOCK` | `2**21` |
 | `TARGET_DATA_GAS_PER_BLOCK` | `2**20` |
 | `MIN_DATA_GASPRICE` | `10**8` |
 | `DATA_GASPRICE_UPDATE_FRACTION` | `68` |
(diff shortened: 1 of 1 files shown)


EIPS/eip-4844.md
```diff
@@ -323,33 +327,46 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
...
+        MIN_GASPRICE_PER_BLOB,
         header.excess_blobs,
-        GASPRICE_UPDATE_FRACTION_PER_BLOB
+        DATA_GASPRICE_UPDATE_FRACTION
     )
 ```
 
+The block validity conditions are modified to include data gas checks:
+
+```
```
protolambda, line 354:
```suggestion
```python
```

EIPS/eip-4844.md
```diff
@@ -211,7 +215,7 @@ the `TransactionNetworkPayload` version of the transaction also includes `blobs`
 The execution layer verifies the wrapper validity against the inner `TransactionPayload` after signature verification as:
 
 * All hashes in `blob_versioned_hashes` must start with the byte `BLOB_COMMITMENT_VERSION_KZG`
-* There may be at most `MAX_BLOBS_PER_TX` blob commitments in any single transaction.
+* There may be at most `MAX_BLOBS_PER_BLOCK` blob commitments in any single transaction.
```
ralexstokes, line 218:
```suggestion
* There may be at most `MAX_BLOBS_PER_BLOCK` blob commitments in any single block.
```

adietrichs: The point was to clarify that the per-tx blob number is still limited effectively. Could be reworded though to make that more clear.

ralexstokes: I see this is given on the next line -- I would suggest just deleting this line for now.

adietrichs: done

adietrichs:
I would ideally want to add slot-based excess blob accounting to this PR, but that would mean that `calc_excess_blobs` would have to include some argument like `slots_since_parent`, which would introduce the notion of slots to the EL. Alternatively I could use `time_since_parent` or `target_timestamp` and calculate the number of slots internally, but then the EL would have to know to use fixed slot-based timestamp for those.

EIPS/eip-4844.md
```diff
@@ -211,7 +215,6 @@ the `TransactionNetworkPayload` version of the transaction also includes `blobs`
 The execution layer verifies the wrapper validity against the inner `TransactionPayload` after signature verification as:
 
 * All hashes in `blob_versioned_hashes` must start with the byte `BLOB_COMMITMENT_VERSION_KZG`
-* There may be at most `MAX_BLOBS_PER_TX` blob commitments in any single transaction.
```
gakonst, line 214: What was the motivation behind removing this limitation? (Would also be helpful understanding why it was introduced originally)

EIPS/eip-4844.md
```diff
@@ -323,33 +326,46 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
...
+        ...
+
+        # the signer must be able to afford the transaction
+        assert signer(tx).balance >= tx.message.gas * tx.message.max_fee_per_gas + get_total_data_gas(tx) * tx.message.max_fee_per_data_gas
+
+        # ensure that the user was willing to at least pay the current data gasprice
+        assert tx.message.max_fee_per_data_gas >= get_data_gasprice(parent(block).header)
+```
+
+The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sender balance before transaction execution, and is not refunded in case of transaction failure.
```
gakonst, line 367: Makes sense to not have the data fee be refunded, given that the data persists even on reverted transaction.

EIPS/eip-4844.md
ralexstokes: it was added to bound the size of blob carrying txns in the mempool

but clients have other ways to bound txn size and the overall number of blobs is already limited in a single block

adietrichs: It might be helpful to add a paragraph to the rationale section pointing out that mempools can still come with a max blobs per tx limit, similar to their current max size per tx limit. No need to enforce such a limit at the level of chain validity rules though.

EIPS/eip-4844.md
```diff
@@ -323,33 +326,46 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
...
-def get_blob_gas(header: Header) -> int:
+def calc_data_fee(tx: SignedBlobTransaction, parent: Header) -> int:
+    return get_total_data_gas(tx) * get_data_gasprice(header)
+
+def get_total_data_gas(tx: SignedBlobTransaction) -> int:
+    return DATA_GAS_PER_BLOB * len(tx.message.blob_versioned_hashes)
+
+def get_data_gasprice(header: Header) -> int:
     return fake_exponential(
+        MIN_GASPRICE_PER_BLOB,
```
dankrad, line 345:
```suggestion
        MIN_DATA_GASPRICE,
```

EIPS/eip-4844.md
```diff
@@ -146,16 +147,18 @@ def evaluate_polynomial_in_evaluation_form(poly: List[BLSFieldElement], x: BLSFi
...
 
 ```python
-def fake_exponential(numerator: int, denominator: int) -> int:
-    cofactor = 2 ** (numerator // denominator)
-    fractional = numerator % denominator
-    return cofactor + (
-        fractional * cofactor * 2 +
-        (fractional ** 2 * cofactor) // denominator
-    ) // (denominator * 3)
+def fake_exponential(factor: int, numerator: int, denominator: int) -> int:
```
dankrad, line 153: Is there a rationale for switching to the Taylor expansion?

EIPS/eip-4844.md
```diff
@@ -50,15 +50,16 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
...
+| `MIN_DATA_GASPRICE` | `10**13` |
+| `DATA_GASPRICE_UPDATE_FRACTION` | `84` |
 | `MAX_VERSIONED_HASHES_LIST_SIZE` | `2**24` |
 | `MAX_CALLDATA_SIZE` | `2**24` |
 | `MAX_ACCESS_LIST_SIZE` | `2**24` |
 | `MAX_ACCESS_LIST_STORAGE_KEYS` | `2**24` |
 | `MAX_TX_WRAP_KZG_COMMITMENTS` | `2**24` |
 | `LIMIT_BLOBS_PER_TX` | `2**24` |
-| `GAS_PER_BLOB` | `120000` |
+| `DATA_GAS_PER_BLOB` | `1` |
```
djrtwo, line 61: Are we okay with defining as 1 and closing the door on forward compatibility of cheaper data-gas items? It enshrines blobs the smallest theoretical unit. That is, you can't price something as costing less than a blob without repricing blobs. 

I don't know what that smaller data-gas item might be so I lean toward this being okay and nice for easy accounting but I could have  a blind spot

EIPS/eip-4844.md
djrtwo: And can be operated on in a subsequent normal TX with a sufficient merkle proof into a prior header and TX list...

That said, it might make sense to put some sort of accumulator of KZG commitments in the EL or CL state so that they can be operated on easily in a historic fashion. This would be easy on CL and accessible to EL if 4788 ships

EIPS/eip-4844.md
```diff
@@ -494,17 +510,19 @@ allowing the point verification precompile to work with the new format.
...
+The data gasprice update rule is intended to approximate the formula `data_gasprice = MIN_DATA_GASPRICE * e**(excess_blobs / DATA_GASPRICE_UPDATE_FRACTION)`,
 where `excess_blobs` is the total "extra" number of blobs that the chain has accepted relative to the "targeted" number (`TARGET_BLOBS_PER_BLOCK` per block).
-Like EIP-1559, it's a self-correcting formula: as the excess goes higher, the `blob_gas` increases exponentially, reducing usage and eventually forcing the excess back down.
+Like EIP-1559, it's a self-correcting formula: as the excess goes higher, the `data_gasprice` increases exponentially, reducing usage and eventually forcing the excess back down.
 
 The block-by-block behavior is roughly as follows.
-If in block `N`, `blob_gas = G1`, and block `N` has `X` blobs, then in block `N+1`, `excess_blobs` increases by `X - TARGET_BLOBS_PER_BLOCK`,
-and so the `blob_gas` of block `N+1` increases by a factor of `2**((X - TARGET_BLOBS_PER_BLOCK) / GASPRICE_UPDATE_FRACTION_PER_BLOB)`.
+If in block `N`, `data_gasprice = G1`, and block `N` has `X` blobs, then in block `N+1`, `excess_blobs` increases by `X - TARGET_BLOBS_PER_BLOCK`,
```
djrtwo, line 520: `data_gasprice = G1` is unused subsequently in the explanation. Suggest removing and simplifying as such:
```suggestion
If block `N` contains `X` blobs, then in block `N+1` `excess_blobs` increases by `X - TARGET_BLOBS_PER_BLOCK`,
```

EIPS/eip-4844.md
```diff
@@ -494,17 +510,19 @@ allowing the point verification precompile to work with the new format.
...
+and so the `data_gasprice` of block `N+1` increases by a factor of `e**((X - TARGET_BLOBS_PER_BLOCK) / DATA_GASPRICE_UPDATE_FRACTION)`.
 Hence, it has a similar effect to the existing EIP-1559, but is more "stable" in the sense that it responds in the same way to the same total usage regardless of how it's distributed.
 
+The parameter `DATA_GASPRICE_UPDATE_FRACTION` controls the maximum rate of change of the blob gas price. It is chosen to target a maximum change rate of 10% per block.
```
djrtwo, line 524: Any justification of this 10%? Is this a learning of empirical 1559 data/usage?

djrtwo (commented):
nice!

djrtwo:
oh, and I think "data gas" is sufficient and self explanatory as a name. Coming up with something clever will lead to endless bikeshedding

EIPS/eip-4844.md
timbeiko: +1, why not default to +/-12.5 like EIP-1559?

EIPS/eip-4844.md
adietrichs: This came out of a discussion with @vbuterin. If I recall correctly, reasons were:

- simpler conceptually
- more precise approximation
- only slightly more computationally expensive (and only done once per block, so insignificant)

EIPS/eip-4844.md
```diff
@@ -50,15 +50,16 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
 | `MAX_BLOBS_PER_BLOCK` | `16` |
 | `TARGET_BLOBS_PER_BLOCK` | `8` |
-| `MAX_BLOBS_PER_TX` | `2` |
-| `GASPRICE_UPDATE_FRACTION_PER_BLOB` | `64` |
+| `MIN_DATA_GASPRICE` | `10**13` |
```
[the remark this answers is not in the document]
adietrichs: Avoiding waste (giving out blobs "for free"), but also reducing the ramp-up time when transitioning out of below-target demand.

EIPS/eip-4844.md
```diff
@@ -323,33 +326,46 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
 
 ### Gas price of blobs (Simplified version)
 
-For early draft implementations, we simply change `get_blob_gas(parent)` to always return `GAS_PER_BLOB`.
+For early draft implementations, we simply change `get_blob_gas(parent)` to always return `SIMPLE_GAS_PER_BLOB`.
```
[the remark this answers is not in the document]
adietrichs: I think this has been in the EIP from the beginning. Seems sensible to add that clarification. Now that we will move to a proper fee market implementation for the devnets anyway, maybe we could just remove it entirely?

EIPS/eip-4844.md
```diff
@@ -323,33 +326,46 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
...
 
-### Gas price update rule (Full version)
+### Gas accounting (Full version)
 
-We propose a simple independent EIP-1559-style targeting rule to compute the gas cost of the transaction.
-We use the `excess_blobs` header field to store persistent data needed to compute the cost.
+We introduce data gas as a new type of gas. It is independent of normal gas and follows its own targeting rule, similar to EIP-1559.
+We use the `excess_blobs` header field to store persistent data needed to compute the data gas price. For now, only blobs are charged for in data gas. In the future this could be extended to also cover calldata (at a different relative gas cost).
```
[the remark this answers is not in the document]
adietrichs: I think charging for calldata in data gas is a very interesting question, but agree that it might be premature to discuss here. Will remove.

EIPS/eip-4844.md
adietrichs: Agreed, seems sensible to move to a "1 gas per byte" model, i.e. `DATA_GAS_PER_BLOB = 2 ** 17`. That also makes a potential move to time-based targeting easier.

EIPS/eip-4844.md
adietrichs: The reasoning was that blob demand is expected to be somewhat less volatile than normal block space demand, so we can choose a point slightly more towards the smooth side of the smoth-reactive tradeoff. Also, given the different exponential function used, we could only approximate 12.5% (an update fraction of 68 would result in a max change rate of 12.48%). That said, maybe departing from the main 1559 value is premature and we should just stick with the value closest to 12.5% for conceptual simplicity?

EIPS/eip-4844.md
```diff
@@ -323,33 +326,48 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
 
 ### Gas price of blobs (Simplified version)
 
-For early draft implementations, we simply change `get_blob_gas(parent)` to always return `GAS_PER_BLOB`.
+***WARNING:` This is only for testing***
```
djrtwo, line 329: a back-tick slipped in there
```suggestion
***WARNING: This is only for testing***
```

roberto-bayardo:
FYI I'm starting to implement this to be part of our next EIP-4844 devnet, hopefully going out by end of week. I'll work around the open discussions best I can, though if some can be closed out in the short term please do so!

EIPS/eip-4844.md
adietrichs: I went ahead and changed it to 68.

adietrichs:
@roberto-bayardo I made some more data gas related changes:

- updated `DATA_GASPRICE_UPDATE_FRACTION`
- updated `DATA_GAS_PER_BLOB` and `MIN_DATA_GASPRICE`
- changed `excess_blobs` header field to `excess_data_gas` and adjusted update rules
- changed `TARGET_BLOBS_PER_BLOCK` to `TARGET_DATA_GAS_PER_BLOCK`
- changed `MAX_BLOBS_PER_BLOCK` to `MAX_DATA_GAS_PER_BLOCK`

With the exception of the header field, these should be relatively minor changes. Let me know whether you think it is realistic to include these into the next devnet release.

The only remaining planned update to this PR is optional from an implementation pov, it will move blob gas from static tracking (calculate blob gas usage before tx execution) to dynamic (track blob gas during tx execution). Those behaviors are fully equivalent for 4844 though, so any static tracking implementation should remain spec-compliant.

roberto-bayardo:
We'll work on getting those in @adietrichs I think we can manage.

EIPS/eip-4844.md
```diff
@@ -178,9 +181,9 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
 
 ### Header extension
 
-The current header encoding is extended with a new 256-bit unsigned integer field `excess_blobs`. This is the running
-total of excess blobs included on chain since this EIP was activated. If the total number of blobs is below the
-average, `excess_blobs` is capped at zero.
+The current header encoding is extended with a new 256-bit unsigned integer field `excess_data_gas`. This is the running
```
Inphi, line 184: Will 256-bit be sufficient now that this is counting data gas?

Inphi: ah nvmind. I read this wrong.

Rachi26 (approved):

roberto-bayardo:
@adietrichs don't see it in the spec, but I assume fees due to data gas are burned rather than go to validator?

adietrichs:
> @adietrichs don't see it in the spec, but I assume fees due to data gas are burned rather than go to validator?

Ah, yes they are supposed to be burned. That's a good point though, that should of course be explicitly specified.

Inphi (commented):
Now that data gas is extricated from regular gas and that data gas is fixed for a given transaction, we can remove the [mempool issues section](https://github.com/ethereum/EIPs/blob/638e8aa730c3a5451060f8277db8129cc0e4f5e2/EIPS/eip-4844.md#mempool-issues).

EIPS/eip-4844.md
```diff
@@ -45,17 +45,18 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `BLOB_COMMITMENT_VERSION_KZG` | `Bytes1(0x01)` |
 | `POINT_EVALUATION_PRECOMPILE_ADDRESS` | `Bytes20(0x14)` |
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
-| `MAX_BLOBS_PER_BLOCK` | `16` |
-| `TARGET_BLOBS_PER_BLOCK` | `8` |
-| `MAX_BLOBS_PER_TX` | `2` |
-| `GASPRICE_UPDATE_FRACTION_PER_BLOB` | `64` |
+| `MAX_DATA_GAS_PER_BLOCK` | `2**21` |
```
xinbenlv, line 48: @SamWilsn  and @lightclient  can we quickly allowlist the `https://github.com/ethereum/consensus-specs`which I think was a consensus?

Pandapip1 (approved):

EIPS/eip-4844.md
timbeiko: @Pandapip1 I see you approved the PR but the merge is failing because of the links to the CL spec. Should we force-merge this?

Pandapip1: Yes

timbeiko: Do you have this power :-D ?

MicahZoltu, line 48: All constants, especially unintuitive ones, should include rational for why the specific value has been chosen.

EIPS/eip-4844.md
```diff
@@ -45,17 +45,18 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
...
+| `TARGET_DATA_GAS_PER_BLOCK` | `2**20` |
+| `MIN_DATA_GASPRICE` | `10**8` |
```
MicahZoltu, line 50: This should be `0`, or if there is some mathematical reason it can't be (math often breaks at/near 0) then it should be whatever the smallest number that *doesn't* cause it to break is.

EIPS/eip-4844.md
MicahZoltu: I think this is "conceptually simpler" for a *very* small set of people (math lovers).  I think for the vast majority of the rest of the world (which likely includes most of the people building Ethereum and building stuff on Ethereum) the original was "conceptually simpler".

I don't think the precision here is necessary, given that we are just guessing at capacity targets anyway.

EIPS/eip-4844.md
```diff
@@ -267,33 +271,48 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
 
 ### Gas price of blobs (Simplified version)
 
-For early draft implementations, we simply change `get_blob_gas(parent)` to always return `GAS_PER_BLOB`.
+***WARNING: This is only for testing***
 
-### Gas price update rule (Full version)
+For early draft implementations, we simply change `get_blob_gas(parent)` to always return `SIMPLE_GAS_PER_BLOB`.
```
MicahZoltu, line 276: EIPs should be written toward their final target and not contain historical details (one can look at commit history and discussions for that).  It is OK to not be fully committed to everything in the EIP yet, but the *form* of the EIP should target its final state (which means not saying things like "For early draft implementations do X instead of Y".  Instead, just say "Do X" and then later that may change to "Do Y".

That being said, I'm not an EIP editor anymore so I have no actual authority over this matter.

timbeiko:
@Pandapip1 friendly ping :-) If we could get this merged before tomorrow it would be great, as we have a 4844 implementers' call where we want to discuss the spec https://github.com/ethereum/pm/issues/647

Jiggnubs:
Can I join with faith

Pandapip1:
@Jiggnubs what do you mean by "join with faith?"

EIPS/eip-4844.md
adietrichs: Conceptually simpler not from a "reading the code", but an "understanding where the code came from" perspective. Which I would argue is the main objective.

EIPS/eip-4844.md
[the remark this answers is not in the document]
adietrichs: After some discussion, reverted back to `1` for now. Will be addressed in a separate PR.

EIPS/eip-4844.md
adietrichs: Given that this PR does not introduce either of the [simplified/full] sections, but just modifies the full one, I think removing the simplified one should not be in scope for the PR. It does seem reasonable to me to remove it soon though.

adietrichs:
Set `MIN_DATA_GASPRICE` back to `1` as discussed on https://github.com/ethereum/pm/issues/647 (discussion about final value will be a separate PR).
The PR is ready for merge from my side.

adietrichs:
I also removed most of the old mempool issues section as proposed by @Inphi and replaced it with a short updated version.

protolambda (approved):

jessepollak:
Amazing work @adietrichs - thank you!

EIPS/eip-4844.md
```diff
@@ -271,33 +275,48 @@ def point_evaluation_precompile(input: Bytes) -> Bytes:
...
+___WARNING: This is only for testing___
 
-### Gas price update rule (Full version)
+For early draft implementations, we simply change `get_blob_gas(parent)` to always return `SIMPLE_GAS_PER_BLOB`.
 
-We propose a simple independent EIP-1559-style targeting rule to compute the gas cost of the transaction.
-We use the `excess_blobs` header field to store persistent data needed to compute the cost.
+### Gas accounting (Full version)
+
+We introduce data gas as a new type of gas. It is independent of normal gas and follows its own targeting rule, similar to EIP-1559.
```
SamWilsn, line 284: I _really_ dislike introducing a new resource separate from gas and calling it "data gas". The term "data gas" implies a subtype of gas—this isn't.

As a radical suggestion, I'd suggest calling this resource [teu](https://en.wikipedia.org/wiki/Twenty-foot_equivalent_unit).

I won't let this hold up the PR though.

SamWilsn (approved):
As established in an earlier EIP-4844 PR, we're going to allow the links to the consensus-specs.

PR #5709, opened 2022-09-23
Update EIP-3525

stoneworld:
_getNewTokenId do not use the fromTokenId_  && when you call ERC721._burn(tokenId_) call ownerOf  can't get the valid tokenId Owner.

### DIFF
--- a/EIPS/eip-3525.md
+++ b/EIPS/eip-3525.md
@@ -573,4 +573,3 @@ Since this EIP is EIP-721 compatible, any wallets and smart contracts that can h
 ## Copyright
 
 Copyright and related rights waived via [CC0](../LICENSE.md).
-
--- a/EIPS/eip-3525.md
+++ b/EIPS/eip-3525.md
@@ -573,5 +573,3 @@ Since this EIP is EIP-721 compatible, any wallets and smart contracts that can h
 ## Copyright
 
 Copyright and related rights waived via [CC0](../LICENSE.md).
-
-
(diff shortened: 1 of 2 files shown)


will-edge (changes requested):
The correction for burn() method is good, and thanks for that. It is a typo when creating this this separate demo version for EIP proposal. We will later post a Readme document,  indicating that the code committed here is not a complete implementation for use in production, it is only for the purpose of demonstrating how to implement this EIP.

The parameter of _getNewTokenId is added deliberately, indicating that one should not assume that the generation of a new id only depends on the internal status of the contract, although the implementation in this version did not use this parameter.

will-edge:
> The correction for burn() method is good, and thanks for that. It is a typo when creating this this separate demo version for EIP proposal. We will later post a Readme document, indicating that the code committed here is not a complete implementation for use in production, it is only for the purpose of demonstrating how to implement this EIP.
> 
> The parameter of _getNewTokenId is added deliberately, indicating that one should not assume that the generation of a new id only depends on the internal status of the contract, although the implementation in this version did not use this parameter.

Please remove the change to _getNewTokenId as explained above.

stoneworld:
> 

Thanks for the explanation, it has been revised.

xinbenlv (dismissed):

Pandapip1 (approved):
This is fine by me

will-edge (approved):
Thanks for the fix.

PR #5718, opened 2022-09-26
Update EIP-2315: Example bytecode fix

radeksvarz:

 - The build passes.


EIPS/eip-2315.md
```diff
@@ -317,7 +317,7 @@ Error: at pc=0, op=RETURNSUB: invalid retsub
 
 In this example the RJUMPSUB is on the last byte of code. When the subroutine returns, it should hit the 'virtual stop' _after_ the bytecode, and not exit with error
 
-Bytecode: `0x5c035e5ff` (`RJUMP 3, RETURNSUB, RJUMPSUB -1`)
+Bytecode: `0x5c035e5fff` (`RJUMP 3, RETURNSUB, RJUMPSUB -1`)
```
holiman, line 320:
```suggestion
Bytecode: `0x5c035e5f` (`RJUMP 3, RETURNSUB, RJUMPSUB -1`)
```
Isn't this the intention, rather than adding an `ff`? 

> In this example the RJUMPSUB is on the last byte of code.

holiman (changes requested):

EIPS/eip-2315.md
radeksvarz: I do not think so:
1. RJUMPSUB requires relative offset and it was clearly declared as -1 in the description = 0xff
 -- there was just one f missing - I assume typo
 
2. RETURNSUB returns PC just behind RJUMPSUB - to position 5 - effectively into the STOP code as 0x0 is expected there - so the 'virtual stop' claim is rather optimization for the clients.

We could improve the sentence: 

When the subroutine returns using RETURNSUB, it should hit the 'virtual stop' _after_ the bytecode, and not exit with error

holiman:
Right, my bad.

radeksvarz:
> On another note: the immediate is defined as uint16, so should not be referenced to as '-1'.

I believe that this case demonstrates that it should be rather defined as signed int16 - to be consistent with EIP 4200.
I raised the relevant question here: https://ethereum-magicians.org/t/eip-2315-simple-subroutines-for-the-evm/3941/137


> Also, there needs to be not one but two bytes of immediate data (four nibbles instead of one) . So 0xffff,  65535 rather than -1.

You are right about 0xffff. And similar mistake is for RJUMP as it should be int16. See EIP 4200: `The immediate argument relative_offset is encoded as a 16-bit signed (two’s-complement) big-endian value. `

I guess the resulting bytecode should be: 0x5c00045e5ffff

gcolvin:
I'm getting ready to travel, so am out of time to look at this, but you folks seem to have it under control, thanks.  EIP-4200 came much, much later than this proposal and I'm not surprised that I failed to fully harmonize them.

gcolvin:
> > On another note: the immediate is defined as uint16, so should not be referenced to as '-1'.
> 
> I believe that this case demonstrates that it should be rather defined as signed int16 - to be consistent with EIP 4200. I raised the relevant question here: https://ethereum-magicians.org/t/eip-2315-simple-subroutines-for-the-evm/3941/137

At line 76 the spec already says the immediate is signed:
> The `relative_offset` is relative to the current `PC`. The address is encoded as a two-byte, twos-complement signed integer, stored MSB-first.

gcolvin:
> At line 76 the spec already says the immediate is signed:

And at line 69 it says otherwise.  It should be int16.  Or just removed, as it is redundant.

radeksvarz:
I have updated the PR to reflect the conclusions here.

gcolvin (dismissed):

radeksvarz:
@gcolvin PR merge conflict was corrected, I do not know how to resolve Auto Label server error.

xinbenlv (approved):
LGTM on editorial side

SamWilsn:
This should just need an approval from one of the listed authors.

holiman:
I don't really feel comfortable being listed as one. This eip has changed a lot.

gcolvin (approved):

gcolvin:
> I don't really feel comfortable being listed as one. This eip has changed a lot.

I just made the approval, but it doesn't seem to be merging.  I think another PR to remove you from the author list would be better given the troubles we are having with this merge.

From my point of view the basic facility hasn't really changed: it's one instruction to call a subroutine and one to return, using a return stack.  The changes to the actual opcode specifications are just the move to immediate arguments, (which we always wanted), relative offsets, (to harmonize with RJUMP) and the removal of BEGINSUB, (which was not in the original proposal anyway, and I regretted adding).   The validation I took from EIP-615 and added long ago.  I moved it a separate EIP for a while, but it seemed better to have it all in one place.  The rest is expanded motivation and rationale, not specification.

So I've vacillated on whether to change numbers, as all of the analysis and discussion so far remains relevant, and the number and the concept are pretty well-known by now.  But I'd have no strong objections to renumbering and/or splitting up this proposal if you prefer.

 If it's the proposal itself you don't support I'd like to discuss it with you privately first, if you don't mind.

PR #5743, opened 2022-10-03
Remove @lightclient as ERC editor

lightclient:
Hi all, unfortunately have not been able to contribute much to ERC editing lately and my inbox if overflowing with notifications. I will be more effective if I focus on the other EIP areas and helping with the process.

@Pandapip1 and @SamWilsn have been doing a great job staying on top of ERCs. If we get to a point again where ERCs are not reviewed in a timely manner I can return to editing them.

Thanks for understanding.


SamWilsn (approved):
:crying_cat_face:

PR #5759, opened 2022-10-06
Update EIP-5679: Move to Last Call

xinbenlv:
creating a PR to accept revisions for EIP-5679.
Will update last call deadline before merge


Pandapip1 (changes requested):
There are a few minor things that need fixing. I'll fix them myself.

SamWilsn (commented):
Should EIP-165 support be optional? That would allow contracts with matching mint/burn functions to be compatible.

I'd expand on (6) a bit. I don't find "with respect to the `ERC721TokenReceiver`" clear enough. Maybe something like "`safeMint` MUST implement the same receiver restrictions as `safeTransferFrom` as defined in EIP-721".

`transfer` -> `transferred` on 68/69 and 73/74

You'll need to remove the external links to goerli. If you can, please put the source code in your assets directory.

EIPS/eip-5679.md
```diff
@@ -4,7 +4,8 @@ title: Token Minting and Burning
 description: An extension for minting and burning tokens for EIP-20, EIP-721, EIP-1155.
```
Pandapip1, line 4:
```suggestion
description: An extension for minting and burning EIP-20, EIP-721, and EIP-1155 tokens
```

xinbenlv:
> Should EIP-165 support be optional? That would allow contracts with matching mint/burn functions to be compatible.

Done

> 
> I'd expand on (6) a bit. I don't find "with respect to the `ERC721TokenReceiver`" clear enough. Maybe something like "`safeMint` MUST implement the same receiver restrictions as `safeTransferFrom` as defined in EIP-721".

Done. I think your wording is much clearer.

> `transfer` -> `transferred` on 68/69 and 73/74

Done

> You'll need to remove the external links to goerli. If you can, please put the source code in your assets directory.

Done.  I removed them to comply with your editorial comment. (For the record I oppose to the restriction for adding implementation in the format of deployment address.)

xinbenlv:
@SamWilsn thanks for reviewing, this PR is ready for merge

Pandapip1 (approved):

PR #5761, opened 2022-10-06
Update EIP-2135: Move to Last Call

xinbenlv:
Creating a PR to accept revision suggestions before moving to last call

Will update last call deadline before merge


SamWilsn (changes requested):
- Abstract is a bit light on details. Could maybe define what "consume" means, who is doing the consuming, etc.
 - `consumaed` is not a word.
 - Motivation section doesn't do a good job convincing me that we need an ERC for this (though it does touch on it.) As an example of the content I'd like to see, maybe mention what you'd expect wallets to do with this information? Maybe NFT marketplaces too (say displaying consumed tokens separately from unconsumed ones)?
 - Clarify whether `consume` burns the token. Right now it sounds like a compliant `consume` is allowed to not transfer ownership to the zero address, as long as it emits the event.
 - Please remove the external links to goerli in the Reference Implementation section. If you'd like to include a reference implementation, please put it in this EIP's `assets` directory.
 - Recommending an audit is good advice, but I think it applies to basically every EIP :rofl: You should imagine the Security Considerations section as a list of specific things _that auditors should pay attention to **for this EIP**_ and not as generic advice to smart contract developers.

xinbenlv:
Working on addressing @SamWilsn 's question

PR #5187, opened 2022-06-29
Add EIP-5187: Extend EIP-1155 with rentable usage rights

DerivStudio:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #5771, opened 2022-10-09
Revert "Add EIP-5187: Extend EIP-1155 with rentable usage rights"

DerivStudio:
Reverts ethereum/EIPs#5187


Pandapip1:
Why are you reverting this?

PR #5737, opened 2022-09-30
Update README: Clarify which standards are useful submissions

fulldecent:
Here is the blog post https://blog.phor.net/2022/09/30/What-kinds-of-things-should-be-standardized.html

This PR adds it as recommended reading.

PR #5778, opened 2022-10-11
Update EIP-5615: Move to Review

Pandapip1:
Makes a few clarifications as well


EIPS/eip-5615.md
```diff
@@ -13,33 +13,36 @@ requires: 1155
...
 
-This EIP does not extend [EIP-165](./eip-165.md) as this interface was deemed simple enough that the extra complexity was unnecessary. It would also break backward compatibility with potential existing implementations.
+This EIP does not implement [EIP-165](./eip-165.md), as this interface is simple enough that the extra complexity is unnecessary and would cause incompatibilities with pre-existing implementations.
+
+For extensibility, `totalSupply` is required to call the `exists` function in case a contract extends the base implementation and overrides `exists`.
 
 ## Backwards Compatibility
 
-This EIP is backward compatible with the OpenZeppelin `ERC1155Supply`.
+No backward compatibility issues were found.
```
xinbenlv, line 45: [optional] When an EIP was designed specifically to compatible with some other EIP / protocol  in mind, i think it's great to call it out. I love the old sentence better
```suggestion
This EIP is designed to be backward compatible with the OpenZeppelin `ERC1155Supply`.
```

EIPS/eip-5615.md
```diff
@@ -13,33 +13,36 @@ requires: 1155
...
+  function totalSupply(uint256 id) external view returns (uint256);
 }
 ```
 
 ## Rationale
 
-This EIP does not extend [EIP-165](./eip-165.md) as this interface was deemed simple enough that the extra complexity was unnecessary. It would also break backward compatibility with potential existing implementations.
+This EIP does not implement [EIP-165](./eip-165.md), as this interface is simple enough that the extra complexity is unnecessary and would cause incompatibilities with pre-existing implementations.
+
+For extensibility, `totalSupply` is required to call the `exists` function in case a contract extends the base implementation and overrides `exists`.
```
SamWilsn, line 41: This is an implementation detail.

In a single purpose implementation of this standard (so no inheritance), it might not make sense to ever call `exists`, especially if it always returns true] (for example in a shared 1155 implementation where anyone can create their own tokens.)

The wording you have in the documentation comment for `totalSupply` is fine, however, because it doesn't explicitly say that `totalSupply` must call `exists`, just that if `exists(x)` returns `false` then `totalSupply(x)` must revert.

xinbenlv (commented):
I think it's ok to move to Review. I'd love to hear more rationale of the choice of `exists()` method and alternatives that has been considered.

Also, while I personally think this is an useful standard, and I disagree with the "suggestions" in determining whether something worth a EIP, made in https://github.com/ethereum/EIPs/pull/5737, I like to hear @fulldecent 's thought on how the suggestions will evaluate this particular EIP

fulldecent:
As per #5737, I expect the author here should cite some people that are using this already. At a minimum this should be in the official discussion record.

They started that by saying it is an OZ contract. But it would be nice if they actually did the work of going and finding some live implementations.

In addition to find some people implementing the producer of this function, they should also go find and documentation that is usefully querying this information.

Yes, this might even take a few minutes of time! But the goal here is to publish useful, helpful things, not just find the remaining function names in 4bytes, put your name on it, and "earn" one of the remaining 5,000 EIP numbers.

Pandapip1:
> As per #5737, I expect the author here should cite some people that are using this already. At a minimum this should be in the official discussion record.

I'll add a few relevant links. I don't think this should necessarily be included in the EIP itself, but the linking in the discussion thread is a no-brainer.

OpenSea supports both of these functions. This is undocumented but is clear from their example repository. It's also noted in the first thread I just linked to.

OpenZeppelin, of course, has an implementation of this.

I don't see an easy way to track support for this for deployed contracts.

fulldecent:
@Pandapip1 cool reference. 1155 requires strict token reporting so totalsupply function should be duplicative of the token transfers. If OS does not say they are using this function to collect data for their production website we should assume they are not using it

Pandapip1:
> If OS does not say they are using this function to collect data for their production website we should assume they are not using it

OpenSea doesn't have any documentation about support for EIP-1155 at all, except for a page in the sidebar with a "documentation soon™"

Pandapip1:
It would be nice if someone from @ProjectOpenSea could weigh in.

SamWilsn (dismissed):

Pandapip1:
That did not work :|

Pandapip1:
There we are.

SamWilsn (approved):

PR #5732, opened 2022-09-29
Init a EIP for Commit Interface

xinbenlv:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #5453, opened 2022-08-12
Add EIP-5453: Smart Contract Crypto Endorsement

xinbenlv:
Add an ERC for on-chain endorsement.

PR #5785, opened 2022-10-15
Update EIP-5750: Move to Review

xinbenlv:
A relatively straightforward


EIPS/eip-5750.md
```diff
@@ -1,18 +1,18 @@
 ---
 eip: 5750
-title: Extra Data Parameter in Methods
-description: Defines an extra data parameter in methods.
+title: Extending Method Behavior with Data Field
```
xinbenlv, line 3: Reviewers, I really need some good suggestion for title. :) Suggestions are greatly appreciated.

EIPS/eip-5750.md
```diff
@@ -1,18 +1,18 @@
 ---
 eip: 5750
-title: Extra Data Parameter in Methods
-description: Defines an extra data parameter in methods.
+title: Last Data for Contract Method Extensions
+description: An EIP to recognize the last bytes field to be used for universal extensibility for method behaviors.
```
SamWilsn, line 4:
```suggestion
description: Recommend using the last method parameter for universal extensibility for method behaviors.
```

EIPS/eip-5750.md
```diff
@@ -1,18 +1,18 @@
...
 created: 2022-10-04
 ---
 
 ## Abstract
 
-This EIP defines an extra data parameter in methods, denoted as `methodName(... bytes calldata _data)`. Compliant methods 
+This EIP defines an extra data parameter in methods, denoted as `methodName(... bytes calldata _data)`. Compliant methods
 can use the extra data to introduce extended behaviors. Any method, regardless of name, can use this extra parameter.
 
 ## Motivation
```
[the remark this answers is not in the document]
Pandapip1: > we don't want the EIPs repository to become a catalogue of Solidity best practices

I see no reason why best practices shouldn't be standards. I could see the type being changed to informational, though.

xinbenlv:
Just to be clear, this is not a "recommendation" or "best practice", it's a protocol/standard *specifying& behavior.

xinbenlv:
@SamWilsn thanks for the good question. FYI I put my argument in the motivation section to answer @SamWilsn 's question. PTAL

xinbenlv:
Also @Pandapip1  just to be clear, this EIP is not a "recommendation" or "best practice". EIP intend to serve as a a protocol/standard *specifying* format of method and their behavior. Unlike a lot of other ERCs that specifying the name and all parameters, this EIP specifies only the last data and its format, which makes it maximally compatible, both future and backward.

It could be confusing, so let me make a metaphor her: 
- If someone says: *"I'd recommend you drive your car on the right side of the road because it has benefit of we bump into each other less"* it's an informational EIP.
- If someone says, *"For all standard-compliant road, let's delegate its right side of the road to be reserved for forward traffic, so we can future agree to which side to build traffic lights, ramps, lightings, fire stations..."*, it's an ERC EIP

This EIP serves for the standardizing purpose: There are a thousand ways to extend methods, but this EIP ask to designate certain place and format for extending behaviors. EIPs like EIP-5453(#5453) and EIP-5732(#5732) can all benefit from this EIP if they can rely on that a parameter in a given method is being designated for extending behavior.

Pandapip1 (dismissed):
I think this is okay to go to review. It's a best practices standard, which is a standard nonetheless.

xinbenlv:
Done, this PR is ready for merge

xinbenlv:
@Pandapip1 could you reapprove?

Pandapip1 (approved):

xinbenlv:
It seems still need to re-run CI @Pandapip1 ?

PR #5790, opened 2022-10-17
Update EIP-5528: Move to final

StartfundInc:
Move to final

### DIFF
--- a/EIPS/eip-5528.md
+++ b/EIPS/eip-5528.md
@@ -4,8 +4,7 @@ title: Refundable Fungible Token
 description: Allows refunds for EIP-20 tokens by escrow smart contract
 author: StartfundInc (@StartfundInc)
 discussions-to: https://ethereum-magicians.org/t/eip-5528-refundable-token-standard/10494
-status: Last Call
-last-call-deadline: 2022-10-16
+status: Final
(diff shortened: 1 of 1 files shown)


SamWilsn (changes requested):
I've put up https://github.com/StartfundInc/EIPs-5528-review/pull/1 with some editorial changes.

Besides those changes, I think this EIP needs some more work before moving to the final state:

 - The difference between the *Payable Contract* and the *Escrow Contract* are not clearly explained enough. For example, it seems like they both implement `IERC5528` but behave differently.
 - There is some ambiguity about whether *Escrow Contract*s can implement policies beyond the list enumerated in this proposal. I'd add a line explicitly either allowing or prohibiting extra policies.
 - The instructions for implementing the *Payable Contract*'s functions are a bit confusing. Adding an example Solidity function call to demonstrate what you mean by calling the same function in the *Escrow Contract* with certain parameters would certainly help clear things up.
 - I think I missed replacing `_transfer` with `transfer` in a few places :sweat_smile: 
 - The Rationale section still has some grammar problems.
 - The section header describing policies is called `#### constructor`. I'm not sure if you mean to imply that these policies have to be set in the constructor, or if that was a mistake.

PR #5792, opened 2022-10-17
Add EIP-5792: Wallet Abstract Transaction Send API

moodysalem:


moodysalem:
I would like to get an EIP number assigned so that I can create an Ethereum magicians thread

EIPS/eip-abstract_wallet_transaction_api.md
```diff
@@ -0,0 +1,218 @@
...
+user's wallet, e.g. smart contract wallets utilizing [EIP-4337](./eip-4337.md) or EOA wallets that
+support bundled transactions via [EIP-3074](./eip-3074.md).
+
+## Motivation
+
+The current APIs to send transactions and check their status, `eth_sendTransaction` and `eth_getTransactionReceipt`,
+are not flexible enough for dapps to provide good notifications and user experience for transaction status.
+For example, these methods do not allow sending multiple transactions, or querying the status of a replaced-by-fee (sped
+up)
+transaction.
```
SamWilsn, line 25: Technically you can query the status of a replaced-by-fee transaction by subscribing to pending and mined transactions and comparing nonces, right? It's not exactly trivial though.

EIPS/eip-abstract_wallet_transaction_api.md
```diff
@@ -0,0 +1,218 @@
...
+- `from`: `ADDRESS` - The address of the wallet that should send the messages in the bundle
+- `messages`: `ARRAY` - An array containing 1 or more message objects, where each message has the following keys:
+    - `to`: `ADDRESS` - (optional when creating new contract) The address the message is directed to.
+    - `gas`: `QUANTITY` - (optional) Integer of the gas provided for the message execution. Unused gas MAY be made
+      available to subsequent messages.
+    - `value`: `QUANTITY` - (optional) Integer of the value sent with this transaction.
+    - `data`: `DATA` - The compiled code of a contract OR the message data to include with the call, e.g. encoded
+      function signature and parameters.
+    - `optional`: `BOOLEAN` - (optional, default false) Whether the wallet should continue with subsequent messages if
+      this particular message fails
```
SamWilsn, line 72: Can you specify this in the same way as the [`execution-apis`](https://github.com/ethereum/execution-apis/blob/main/src/eth/submit.yaml)?

EIPS/eip-abstract_wallet_transaction_api.md
```diff
@@ -0,0 +1,218 @@
...
+      }
+    ]
+  }
+]
+```
+
+#### Return value
+
+DATA, 32 Bytes - an identifier that the wallet uses to represent the bundle of messages. This identifier MUST persist
+between connections of the same Dapp and wallet.
```
SamWilsn, line 103: This discussion should happen on your `discussions-to` link, once you create one. In the meantime:

I don't like this design. It doesn't force interoperability between wallets. A bundle created with WalletA may or may not be usable with WalletB. I would suggest either:

```suggestion
DATA, 32 Bytes - an identifier that the wallet uses to represent the bundle of messages. This identifier MUST persist
between connections of the same Dapp and account.
```

Or allowing the dapp to specify the identifier.

EIPS/eip-abstract_wallet_transaction_api.md
moodysalem: That's a ton of work for the dapp to do, essentially you're indexing a firehose just to figure out something the wallet trivially knows because it initiated the change

EIPS/eip-5792.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Backwards Compatibility
+
+Wallets that do not support the following methods should return error responses to the new JSON RPC methods indicating
+as such.
+Dapps MAY attempt to send the same bundle of messages via `eth_sendTransaction` when they receive a failure message.
+
+## Security Considerations
+
+Dapp developers MUST treat each message in a bundle as if it were an independent transactions. In other words, there may
```
Pandapip1, line 250:
```suggestion
Dapp developers MUST treat each message in a bundle as if it were independent transactions. In other words, there may
```

EIPS/eip-5792.md
```diff
@@ -0,0 +1,257 @@
+---
+eip: 5792
+title: Wallet Abstract Transaction Send API
+description: Adds JSON RPC methods for sending multiple function calls from a wallet and checking each bundle's status
```
lightclient, line 4:
```suggestion
description: Adds JSON-RPC methods for sending multiple function calls from a wallet and checking each bundle's status
```

lightclient: Please update throughout

EIPS/eip-5792.md
```diff
@@ -0,0 +1,257 @@
...
+Returns an object describing the status of each message in the bundle. Each message has its own status in the response,
+even if all the messages were submitted as part of a single transaction.
+
+- `id`: `DATA` - The identifier of the transaction returned by `wallet_sendMessageBundle`
+- `messages`: `ARRAY` - An array of objects containing the status of each message from the bundle
+    - `status`: `"PENDING" | "CONFIRMED"` - Either `"PENDING"` indicating the message is not yet included on-chain,
+      or `"CONFIRMED"` indicating the message was included in a block
+    - `receipt`: `OBJECT` - (optional) An object containing the receipt for the given message, _iff_ status
+      is `"CONFIRMED"`
+        - `logs`: `ARRAY` - An array of log objects describing logs that were emitted by the message
```
lightclient, line 159: I think it will be non-trivial to determine this in the case of a batched tx in a wallet since many logs may be emitted during execution and you'd need a trace to determine which subcall it belongs to?

EIPS/eip-5792.md
```diff
@@ -0,0 +1,257 @@
...
+included between each of the transactions.
+The wallet MUST send these messages in the order specified in the request.
+The wallet MUST attempt to deliver all messages if it returns a successful response to this method, and the wallet MUST
+NOT deliver any messages if it returns an error response.
+The wallet MAY reject the request if the chain ID does not match the currently selected chain ID.
+The wallet MUST send the messages on the specified chain ID.
+The wallet MUST reject the request if the `from` address does not match the enabled account.
+The wallet MAY reject the request if one or more messages in the bundle will fail.
+
+#### Parameters
```
lightclient, line 63: Missing gas price parameters?

EIPS/eip-5792.md
moodysalem: It's not difficult, but the implementation differs based on the wallet. You don't have to trace if you just emit a separator log between each call.

EIPS/eip-5792.md
moodysalem: Gas price is deliberately excluded. It's the wallet's responsibility to determine the gas price to get a bundle of messages in, especially because it can differ across multiple transactions and gas abstraction means the parameters may be ignored anyway. Wallets also already allow you to override gas prices, and it's a burden for dapps to set them.

lightclient: sgtm

EIPS/eip-5792.md
```diff
@@ -0,0 +1,278 @@
...
+
+```json
+"0xe670ec64341771606e55d6b4ca35a1a6b75ee3d5145a99d05921026d1527331"
+```
+
+The identifier may be a numeric identifier represented as a string, e.g.:
+
+```json
+"0x01"
+```
```
SamWilsn, line 145: Above, the identifier is specified as having 32-bytes. This example only gives a single byte.

So is `0x01 == 0x0000000000000000000000000000000000000000000000000000000000000001` or is `0x01 == 0x0100000000000000000000000000000000000000000000000000000000000000`?

moodysalem: Changed it to a string

EIPS/eip-abstract_wallet_transaction_api.md
```diff
@@ -0,0 +1,218 @@
+---
+eip: <to be assigned>
+title: Abstract Wallet Transaction API
```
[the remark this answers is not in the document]
moodysalem: think it's less vague now

EIPS/eip-abstract_wallet_transaction_api.md
moodysalem: created a pr documenting these methods https://github.com/ethereum/execution-apis/pull/317

zemse:
How would the dapp know whether the wallet supports batching if it want's to have different UX for both approaches? For e.g. while doing a swap on Uniswap, the frontend shows clickable "Approve" and disabled "Swap" for the guided two-step process for EOA, but for wallets that do support batching the frontend can show just a single "Swap" button that sends approve + swap together on click.

This info (whether the connected wallet supports batching or not) is required while the frontend mounts the components, so one hacky solution is to try `wallet_sendFunctionCallBundle` with an empty call object and see if it gives a method not found error. Otherwise, would it make sense to have something like `wallet_supportsBatching`?

lightclient (approved):

PR #5804, opened 2022-10-19
Update EIP-191: Improve Comprehensibility

YamenMerhi:
- Updating the example to be more comprehensive + more updated. 
- Simplifying the EIP191 Standard, by providing more explanation of the different versions.
     - Re-ordering the text so we start by:
             - The formula for the standard
             - Why we have 0x19
             - What are the different bytes version we have
     - Explicitly mentioning that `E` is version `0x45`and `thereum Signed Message:\n" + len(message)` is version specific data.

### DIFF
--- a/EIPS/eip-191.md
+++ b/EIPS/eip-191.md
@@ -31,7 +31,6 @@ We propose the following format for `signed_data`
 ```
 0x19 <1 byte version> <version specific data> <data to sign>.
 ```
-Version `0` has `<20 byte address>` for the version specific data, and the `address` is the intended validator. In the case of a Multisig wallet, that is the wallet's own address  .
 
 The initial `0x19` byte is intended to ensure that the `signed_data` is not valid [RLP](https://github.com/ethereum/wiki/wiki/RLP)
 
@@ -55,29 +54,58 @@ Using `0x19` thus makes it possible to extend the scheme by defining a version `
 |    `0x01`    | [712][eip-712] | Structured data
 |    `0x45`    | [191][eip-191] | `personal_sign` messages
 
+#### Version `0x00`
+
+```
+0x19 <0x00> <intended validator address> <data to sign>
+```
+
+The version `0x00` has `<intended validator address>` for the version specific data. In the case of a Multisig wallet that perform an execution based on a passed signature, the validator address is the address of the Multisig itself. The data to sign could be any arbitrary data.
+
+#### Version `0x01`
+
+```
+0x19 <0x01> <domain seperator> <structHash>
+```
+
+The version `0x01` has `<domain seperator>` for the version specific data and `<structHash>` for the data to sign as defined in [EIP712][eip-712].
+
+#### Version `0x45` (E)
+
+```
+0x19 <0x45 (E)> <thereum Signed Message:\n" + len(message)> <data to sign>
+```
+
+The version `0x45` (E) has `<thereum Signed Message:\n" + len(message)>` for the version specific data. The data to sign could be any arbitrary data.
+
 [EIP-191]: ./eip-191.md
 [EIP-712]: ./eip-712.md
 
 ### Example
 
-The following snippet has been written in Solidity 0.5.0.
+The following snippets has been written in Solidity 0.8.0.
+
+#### Version `0x00`
 
 ```solidity
-function submitTransactionPreSigned(address destination, uint value, bytes data, uint nonce, uint8 v, bytes32 r, bytes32 s)
-    public
-    returns (bytes32 transactionHash)
-{
+function signatureBasedExecution(address target, uint256 nonce, bytes memory payload, bytes memory signature)
+    public payable {
+        
     // Arguments when calculating hash to validate
     // 1: byte(0x19) - the initial 0x19 byte
     // 2: byte(0) - the version byte
-    // 3: this - the validator address
-    // 4-7 : Application specific data
-    transactionHash = keccak256(abi.encodePacked(byte(0x19),byte(0),address(this),destination, value, data, nonce));
-    sender = ecrecover(transactionHash, v, r, s);
-    // ...
+    // 3: address(this) - the validator address
+    // 4-6 : Application specific data
+
+    bytes32 hash = keccak256(abi.encodePacked(byte(0x19), byte(0), address(this), msg.value, nonce, payload));
+
+    // recovering the signer from the hash and the signature
+    addressRecovered = ECDSA.recover(hash, signature);
+    
+    // logic of the wallet
+    if (addressRecovered == owner) executeOnTarget(target,payload);
 }
 ```
-
 ## Copyright
 
 Copyright and related rights waived via [CC0](../LICENSE.md).
--- a/EIPS/eip-191.md
+++ b/EIPS/eip-191.md
@@ -32,11 +32,11 @@ We propose the following format for `signed_data`
 0x19 <1 byte version> <version specific data> <data to sign>.
 ```
 
-The initial `0x19` byte is intended to ensure that the `signed_data` is not valid [RLP](https://github.com/ethereum/wiki/wiki/RLP)
+The initial `0x19` byte is intended to ensure that the `signed_data` is not valid RLP.
 
 > For a single byte whose value is in the [0x00, 0x7f] range, that byte is its own RLP encoding.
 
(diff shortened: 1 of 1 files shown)


YamenMerhi:
I know this is an old standard and missing some sections, I am happy to add them.
Also was wondering if we should mention in a Security Consideration section that it's advised to use 0x00 for any execution based on signature as people could be easily tricked to sign normal 0x45 messages as it's normally used for SIWE and for offchain verification

EIPS/eip-191.md
```diff
@@ -55,29 +54,58 @@ Using `0x19` thus makes it possible to extend the scheme by defining a version `
...
+
+The version `0x00` has `<intended validator address>` for the version specific data. In the case of a Multisig wallet that perform an execution based on a passed signature, the validator address is the address of the Multisig itself. The data to sign could be any arbitrary data.
+
+#### Version `0x01`
+
+```
+0x19 <0x01> <domain seperator> <structHash>
+```
+
+The version `0x01` has `<domain seperator>` for the version specific data and `<structHash>` for the data to sign as defined in [EIP712][eip-712].
```
holiman, line 71: I'd prefer to leave the definition fully to EIP-712. So simply say

> The version `0x01` is for structured data as defined in [EIP-712](eip-712)

holiman (commented):
Looks good to me, aside from one nitpick

holiman (dismissed):
LGTM

Pandapip1 (approved):

EIPS/eip-191.md
```diff
@@ -55,29 +54,54 @@ Using `0x19` thus makes it possible to extend the scheme by defining a version `
...
+
+The version `0x01` is for structured data as defined in [EIP-712]
+
+#### Version `0x45` (E)
+
+```
+0x19 <0x45 (E)> <thereum Signed Message:\n" + len(message)> <data to sign>
+```
+
+The version `0x45` (E) has `<thereum Signed Message:\n" + len(message)>` for the version specific data. The data to sign could be any arbitrary data.
```
SamWilsn, line 75: So this treats the `E` as the version? I would explicitly mention that the `E` is consumed, and `thereum` is not a typo :P

EIPS/eip-191.md
```diff
@@ -55,29 +54,54 @@ Using `0x19` thus makes it possible to extend the scheme by defining a version `
...
-    transactionHash = keccak256(abi.encodePacked(byte(0x19),byte(0),address(this),destination, value, data, nonce));
-    sender = ecrecover(transactionHash, v, r, s);
-    // ...
+    // 3: address(this) - the validator address
+    // 4-6 : Application specific data
+
+    bytes32 hash = keccak256(abi.encodePacked(byte(0x19), byte(0), address(this), msg.value, nonce, payload));
+
+    // recovering the signer from the hash and the signature
+    addressRecovered = ECDSA.recover(hash, signature);
```
SamWilsn, line 99: Hm, I'm leery about changing this from plain `ecrecover` to what I assume is supposed to be OZ's. Their implementation has additional restrictions that plain `ecrecover` doesn't have. Changes to final EIPs are supposed to be backwards compatible.

EIPS/eip-191.md
```diff
@@ -55,29 +54,54 @@ Using `0x19` thus makes it possible to extend the scheme by defining a version `
...
+    
+    // logic of the wallet
+    if (addressRecovered == owner) executeOnTarget(target,payload);
```
SamWilsn, line 102: This is not the minimal change required to fix errata. Since this is a final EIP, we only allow clarifications and correcting errors. The minimal change would likely be just changing the 7 to a 6, unless there's something else I missed.

SamWilsn (changes requested):

Cryptonics1:
Ok thank you I truly appreciate all the help and wisdom I am learning thank
you for your care...I will send more gracious gifts in the near future.

EIPS/eip-191.md
YamenMerhi: I see your point @SamWilsn , I will revert the OZ integration and keep plain `ecrecover` but what if the example is not clear ? Just for the sake of backwards compatibility we leave it like this ? (Knowing that no one will use this code snippet)

The name of the function IMO is not perfect, the same for the naming of variables, as it's not a `TransactionPreSigned` or a `transactionHash`, it's a signed message and hash of a signed message. This whole standard was created to find a standard way to differentiate between signed transactions and signed messages and here the naming is super confusing.

Also the example show how to sign but doesn't give info about what's the possibilities when signing the message. It's just like showing someone a fish hook but not teaching them how to start fishing 😄 

I would find a balance between your suggestions and mine like that:
- Reverted the OZ integration
- Made the new logic as a comment just to emphasize on how this signed message could be used
- Made change to the naming of the function and variables

```solidity
function signatureBasedExecution(address target, uint256 nonce, bytes memory payload, uint8 v, bytes32 r, bytes32 s) public payable {
        
    // Arguments when calculating hash to validate
    // 1: byte(0x19) - the initial 0x19 byte
    // 2: byte(0) - the version byte
    // 3: address(this) - the validator address
    // 4-6 : Application specific data

    bytes32 hash = keccak256(abi.encodePacked(byte(0x19), byte(0), address(this), msg.value, nonce, payload));

    // recovering the signer from the hash and the signature
    addressRecovered = ecrecover(hash, v, r, s);
   
    // logic of the wallet
    // if (addressRecovered == owner) executeOnTarget(target, payload);
}
```
Please let me know what do you think 😄 🙏

EIPS/eip-191.md
YamenMerhi: > So this treats the E as the version?

Exactly, it was confusing that's why i created this PR :)


I would add on the next line.
> NB: The **E** in `Ethereum Signed Message` refers to the version `0x45`. The character `E` is `0x45` in Hex which make the rest: `thereum Signed Message:\n + len(message)` the version specific data.

Pandapip1:
```suggestion
The version `0x45` (E) has `<thereum Signed Message:\n" + len(message)>` for the version-specific data. The data to sign can be any arbitrary data.

> NB: The `E` in `Ethereum Signed Message` refers to the version byte 0x45. The character `E` is `0x45` in hexadecimal which makes the remainder, `thereum Signed Message:\n + len(message)`, the version-specific data.
```

EIPS/eip-191.md
SamWilsn: I think you can go ahead and make this change to the PR. This version is a smaller change.

Pandapip1 (changes requested):
On line 39, ERC needs to be changed to EIP. On line 35, the external link needs to be removed.

Pandapip1 (approved):
+1 from me. Will need manual merge.

EIPS/eip-191.md
```diff
@@ -55,29 +54,55 @@ Using `0x19` thus makes it possible to extend the scheme by defining a version `
 |    `0x01`    | [712][eip-712] | Structured data
 |    `0x45`    | [191][eip-191] | `personal_sign` messages
 
+#### Version `0x00`
+
+```
+0x19 <0x00> <intended validator address> <data to sign>
```
xinbenlv, line 60: Wait, I must be missing something here, are we adding the `<0x00> <intended validator address>` to the original EIP-191?

EIPS/eip-191.md
```diff
@@ -55,29 +54,55 @@ Using `0x19` thus makes it possible to extend the scheme by defining a version `
...
+```
+
+The version `0x00` has `<intended validator address>` for the version specific data. In the case of a Multisig wallet that perform an execution based on a passed signature, the validator address is the address of the Multisig itself. The data to sign could be any arbitrary data.
+
+#### Version `0x01`
+
+The version `0x01` is for structured data as defined in [EIP-712]
```
xinbenlv, line 67: QQ: Should we mention anything about the EIP-712 "domain" besides the "structured data" here?

EIPS/eip-191.md
YamenMerhi: @xinbenlv  The version `0x00` was always there in the standard from the start, but it was not explained very well, and that's the purpose of the PR

EIPS/eip-191.md
YamenMerhi: @xinbenlv  According to @holiman the co-author of the EIP191 standard:
> I'd prefer to leave the definition fully to EIP-712. So simply say:
The version 0x01 is for structured data as defined in [EIP-712](https://github.com/ethereum/EIPs/pull/eip-712)

PR #5758, opened 2022-10-06
Update EIP-5732: Move to Last Call

xinbenlv:
- [x] Add ENS Ref Impl
- [x] Move to last call
- [ ] Update last call deadline to 2 weeks ahead before merge

PR #5805, opened 2022-10-20
Add EIP-5805: Voting with delegation

Amxx:


EIPS/eip-5805.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Abstract
+
+Many DAOs (decentralized autonomous organizations) rely on tokens to represent one's voting power. In order to perform this task effectively, the token contracts need to include specific mechanisms such as checkpoints and delegation.
+
+The existing implementations are not standardized.
+
+This EIP proposes to standardize the way votes are delegated by an account to another, and the way current and past votes are tracked and queried. The corresponding behavior is compatible with many token types, including but not limited to [EIP-20](./eip-20.md) and [EIP-721](./eip-721.md).
+
+This EIP also considers the diversity of time tracking functions, allowing the voting tokens (and any contract associated with it) to track the votes based on `block.number`, `block.timestamp`, or any other non-decreasing function.
```
Pandapip1, line 22:
```suggestion
Many DAOs (decentralized autonomous organizations) rely on tokens to represent one's voting power. In order to perform this task effectively, the token contracts need to include specific mechanisms such as checkpoints and delegation. The existing implementations are not standardized. This EIP proposes to standardize the way votes are delegated from one account to another, and the way current and past votes are tracked and queried. The corresponding behavior is compatible with many token types, including but not limited to [EIP-20](./eip-20.md) and [EIP-721](./eip-721.md). This EIP also considers the diversity of time tracking functions, allowing the voting tokens (and any contract associated with it) to track the votes based on `block.number`, `block.timestamp`, or any other non-decreasing function.
```

EIPS/eip-5805.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Motivation
+
+Beyond simple monetary transactions, decentralized autonomous organizations are arguably one of the first use cases of blockchain and smart contract technologies. Many communities are today organized around a governance contract that allows users to vote. Among these communities, some represent the voting power using transferable tokens ([EIP-20](./eip-20.md), [EIP-721](./eip-721.md), other). In this context, the more tokens you own, the more voting power you have. Governor contracts, such as Compound's GovernorBravo, read from these "voting token" contracts to get the voting power of the users.
```
Pandapip1, line 26:
```suggestion
Beyond simple monetary transactions, decentralized autonomous organizations are arguably one of the most important use cases of blockchain and smart contract technologies. Today, many communities are organized around a governance contract that allows users to vote. Among these communities, some represent voting power using transferable tokens ([EIP-20](./eip-20.md), [EIP-721](./eip-721.md), other). In this context, the more tokens one owns, the more voting power one has. Governor contracts, such as Compound's `GovernorBravo`, read from these "voting token" contracts to get the voting power of the users.
```

Pandapip1 (changes requested):

Pandapip1 (approved):
There's some formatting stuff that could be improved, but this is fine for a draft.

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
...
+#### delegate
+
+This function changes the caller's delegate, updating the vote delegation in the meantime.
+
+This function MUST be implemented
+
+```yaml
+- name: delegate
+  type: function
+  stateMutability: nonpayable
```
xinbenlv, line 144: I wonder if it's intentional to be `nonpayable`?

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
...
+      type: uint256
+    - name: newBalance
+      indexed: false
+      type: uint256
+```
+
+### Solidity interface
+
+```sol
+interface IERC_XXXX {
```
xinbenlv, line 344: [Optional] Do you want to consider adding EIP-165 identifer?

xinbenlv (commented):
Hi @Amxx , glad to see this EIP. I think it's a super important area to address.

Left some comments of the first round.

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
...
+
+## Specification
+
+Following pre-existing (but not-standardized) implementation, the EIP proposes the following mechanism.
+
+Each user account (address) can delegate to an account of its choice. This can be itself, someone else, or no one (represented by `address(0)`). Assets held by the user cannot express their voting power unless they are delegated.
+
+When a "delegator" delegates its tokens voting power to a "delegatee", its balance is added to the voting power of the delegatee. If the delegator changes its delegation, the voting power is subtracted from the old delegatee's voting power and added to the new delegate's voting power. The voting power of each account is tracked through time so that it is possible to query its value in the past. With tokens being delegated to at most one delegate at a given point in time, double voting is prevented.
+
+Whenever tokens are transferred from one account to another, the associated voting power should be deducted from the sender's delegate and added to the receiver's delegate.
```
xinbenlv, line 43: I wonder if you want to specify whether this is `SHOULD` or "MUST" for handling the transfer behavior

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
...
+  outputs:
+    - name: timepoint
+      type: uint256
+```
+
+#### getVotes
+
+This function returns the current voting weight of an account. This corresponds to all the voting power delegated to it at the moment this function is called.
+
+As tokens delegated to `address(0)` should not be counted/snapshoted, `getVotes(0)` SHOULD always return `0`.
```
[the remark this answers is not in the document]
Amxx: My intention is to say that if an implementer wants to track tokens delegated to `address(0)`, and return this non 0 value in `getVotes(0)` and `getPastVotes(0, x)` ... it would not break the key mechanism of this EIP.

If devs what to track address(0) ... that is ok ... but users should not assume that it will be the case.

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
...
+
+Both timestamps, block numbers, and other possible modes use the same external interfaces. This allows transparent binding of third-party contracts, such as governor systems, to the vote tracking built into the voting contracts. For this to be effective, the voting contracts must, in addition to all the vote-tracking functions, expose the current value used for time-tracking.
+
+### Methods
+
+#### now
+
+This function returns the current timepoint. It could be `block.timestamp`, `block.number` (or any other **non-decreasing** function) depending on the mode the contract is operating on.
+
+- If operating using **block number**, then this function SHOULD be implemented.
```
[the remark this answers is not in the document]
Amxx: Yes, that is documented on line 61

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
...
+This function MUST be implemented
+
+```yaml
+- name: getPastVotes
+  type: function
+  stateMutability: view
+  inputs:
+    - name: account
+      type: address
+    - name: timepoint
```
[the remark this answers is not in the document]
Amxx: It needs to be consistent in the sens that:
- if `now()` returns a block.number (or if `now()` does not exist), then `timepoint` should be a block number.
- if `now()` returns a timestamp, then `timepoint` should be a timestamp.

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
...
+  type: function
+  stateMutability: nonpayable
+  inputs:
+    - name: delegatee
+      type: address
+    - name: nonce
+      type: uint256
+    - name: expiry
+      type: uint256
+    - name: v
```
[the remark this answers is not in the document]
Amxx: > [peer review] The `v` here could either be a boolean for the Y-parity, or consider add chainId to avoid replay attack. (note EIP-2294 is in the proposal to formalize the boundary of chainId to support a clear decision of this)

This is an interresting discussion!

My approach has been to match what currently exists (in particular EIP-2612), and make this EIP backward compatible with existing deployment (that afaik all reused EIP-2612 format).

I'll have a look at 5453

EIPS/eip-5805.md
Amxx: Yes. 

Would you change that?

EIPS/eip-5805.md
Amxx: I'm not sure, that is a good point.

I love EIP-165 but the truth is I don't see it used that often in the field.

EIPS/eip-5805.md
```diff
@@ -0,0 +1,401 @@
+---
+eip: 5805
+title: Voting with delegation
+description: An interface for voting weight tracking, with delegation support
+author: Hadrien Croubois (@Amxx)
+discussions-to: https://ethereum-magicians.org/t/eip-5805-voting-with-delegation/11407
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-07-04
```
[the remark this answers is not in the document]
Amxx: I started writing the draft of this document in july ... but waited until post-devcon to create the PR. I'm not sure what "created" is supposed to represent. I'd don't really care. If you editor thing 2022-10-19 is a better date, then I'm ok with that

EIPS/eip-5805.md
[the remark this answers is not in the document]
xinbenlv: That makes sense! I think I am going to proposal soon in Shanghai fork for the developers to make a decision either the chainId will be a full 256bit space for a likely less than 64bit space. Once that's done, I think some part of the execution layer needs to be updated. If the core developers want to allow a full 256bit space, then we will need to update how the parity bit is calculated

EIPS/eip-5805.md
Amxx: also, with "now()" being optional, I'm not sure what the identifier should be?

EIPS/eip-5805.md
xinbenlv: There are two reasons. For what I know, solidity _add_ restriction when we mark it non-payable. It's more gas efficient if put in a payable rather than non-payable. 


See the solidity source code on how the Type address is implemented: the NonPayable checks callvalue 
https://github.com/ethereum/solidity/blob/5f8b48793cdf9a41d6f16fa601250987bee82e3e/libsolidity/codegen/ContractCompiler.cpp#L429
See conversation here

> Adding a payable keyword also lowers the number of opcodes being executed (no check for msg.value == 0), which lowers the gas consumption. Many production use-cases will include it in the final contract design for gas efficiency reasons even if it doesn't use msg.value, assuming there is no security concern with accepting it.

https://github.com/ethereum/EIPs/pull/5758#issuecomment-1287889709

EIPS/eip-5805.md
xinbenlv: That's totally reasonable. If it's purposefully open to interpretation, I'd suggest
1. change the "should" to "MAY"
2. articulate it's purposefully open to interpretation / implementation choice in the `Rationale` section

EIPS/eip-5805.md
xinbenlv: > also, with "now()" being optional, I'm not sure what the identifier should be?

Good point. It also makes restriction usage of EIP-165. I think the common approach is to make EIP-165 out of all required field, and then any optional one will have an `IERC5805Extension` kind of thing which has its own EIP-165 identifier.

EIPS/eip-5805.md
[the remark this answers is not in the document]
xinbenlv: Sure. I think the nature of this is more of use the date for time it first get drafted, I don't have a preference either. Leaving it for @Pandapip1  and @SamWilsn who have more authoritative answer

EIPS/eip-5805.md
Amxx: I believe case 2 will probably need additional data, or at least additional logic, and it should not use this function signature. This function allows the sender (and only the sender) to change its delegation. If you want another mechanism, that allows a third party to pay something to get the delegation, that should be a completely different function.

For someone to change its own delegation, there is no reason to pay anything. If a payment is involved, then again, its a different mechanism that should use a different function signature.

Also, EIPs are about security, not gas effectiveness. I personally think making function payable to save gas is a very bad thing that will just lead to lost funds.

Just like ERC20's "approve" is not payable, this should also not be payable.

EIPS/eip-5805.md
Amxx: I'll stay with SHOULD .. because that what you should do unless you have a good reason not to, and not what you may do and you are not sure.

I'll add something in the Rationale section

xinbenlv: Sounds good

Amxx:
We might need to rename the `now()` function, because of name conflicts with solidity's now keyword (deprecated since 0.7.0 but still raising issues).

Should we go for `clock()` ?

frangio:
`clock` sounds good to me.

xinbenlv:
> `clock` sounds good to me. Another alternative could be `tick`.

Is it the intention of the function  `now()` to represent the _current_ point of time in _either_ `timestamp` or `blocknum`?

I feel the `tick` or `clock` will all strongly imply `timestamp` and could cause trouble if contract developers choose to implement `blocknum` but client interpreting it as `timestamp`. 

I'd brainstorm something like `currentPointOfTime()`, well that's a bit unfortunately long, please help find better name

xinbenlv:
Also, realizing snapshot doesn't work at blocknum level. I am proposing #5875 to retrieve TX Num. Wondering if for better security, we should support TX level of voting weight snapshot

EIPS/eip-5805.md
xinbenlv: LGTM!. While I reserve my opinion on this particular point but respect this is author's discretion at the technical choice made in this proposal.

That said, this different understanding about ERC20's `approval` whether they mean to be `payable` or `nonpayable` or _unspecified_ seem to hard to be represented in Solidity, and hence a good point to use ABI actually

Amxx:
There is still ongoing discution about this EIP (which IMO should happen on the ethereum magician thread, and not here) ... but I believe the proposal state is sufficient to merge as a draft.

If there is anything preventing that from being merged (as a draft), please let me know what needs to be fixed.

Pandapip1 (approved):

Pandapip1:
The only rule about the creation date is it can't be in the future.

PR #5818, opened 2022-10-23
CI: Fix Markdown Linter

Pandapip1:
Wierd things happen with output variables and multi-line strings. This fixes the Markdownlint check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,10 +103,20 @@ jobs:
     steps:
       - name: Checkout EIP Repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-
-      - name: Delete Unchanged Files
+        
+      - name: Get Changed Files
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -103,16 +103,31 @@ jobs:
...
-      - name: Delete Unchanged Files
+        with:
+          fetch-depth: 0
+        
+      - name: Get Changed Files
+        uses: tj-actions/changed-files@1d6e210c970d01a876fbc6155212d068e79ca584
         id: changed-files
-        uses: Pandapip1/delete-unchanged-files@0937c94da2e9f9f262bbf2035e75065a1fd04bcf
+        with:
+          separator: ","
```
SamWilsn, line 113: I assume the hack is because we can't just use `separator: "\n"` for some reason?

Pandapip1: Yes.

PR #5843, opened 2022-10-27
Add EIP-5843: EVM Modular Arithmetic Extensions

jwasinger:


EIPS/eip-5843.md
```diff
@@ -0,0 +1,448 @@
...
+
+`SETMODMX` should have linear-ish complexity.  Modular inverse appears to be `O(logN)` and my current implementation allocates a lot of memory.
+![](../assets/eip-5843/submodx_low.png)
+![](../assets/eip-5843/submodx_med.png)
+![](../assets/eip-5843/submodx_all.png)
+
+`SETMODX` computes Montgomery constants used by generic `mulmont` and algorithms optimized for smaller limb counts.  As a result `SETMODX` is more expensive than it could be, especially for low limbs.  This choice is made to avoid explicitly including a pivot point in the protocol, as it may vary slightly depending on the implementation.
+
+### Other TODOs to note in the rationale:
+* choice of little-endian limbs made to match the hardware of most (all?) Ethereum nodes, removing the need for endianness-swapping individual limbs.
```
mratsim, line 348: It is unclear what limbs are in the specs.

In arbitrary precision bigints libraries, the usual _internal_ representation is:
- metadata about number of limbs, and whether the number is positive or negative
- limbs, from the least significant to the most significant words (eg. limb-endianness is little-endian)
- words, the endianness is hardware endianness (eg. word-endianness is little-endian)
- the size of the word (32 or 64 bits) is also hardware dependent.

That is common across GMP, Java, Go, LLVM internal bigint lib (in APint.h).
However that internal representation is typically inaccessible by third-parties, and you cannot cast to it to achieve the zero-compute that is aimed there.

The other usual representation is called "octet string" is just big-endian serialization fo the integer, usually called I2OSP (Integer to Octet String) from https://www.rfc-editor.org/rfc/rfc3447#section-4.1
All library supports this with an _external API_.

Furthermore, all CPUs and compilers offer __builtin_bswap64 for endianness swapping. And in any case, endianness swapping is virtually costless:
1. On Haswell CPUs (2013) or later, the BSWAP 64-bit instruction takes 2 cycles (less than 1 nanosecond on a 3GHz CPU) and can be issued twice per cycle (https://www.agner.org/optimize/instruction_tables.pdf)
2. Retrieving data from L2 cache takes 7ns or from RAM takes 100ns (https://gist.github.com/jboner/2841832)
3. data load and swapping have no dependencies so while swapping is taking place, the next limb can be retrieved.
4. so due to endian-swapping "latency hiding", converting endianness has the same cost as copying data + ~1ns

So in conclusion, little-endian limbs require the bigints library to expose some way to cast/copy directly into their internals, which is unlikely due to metadata, input validation, etc AND the difference with using standard octet-string is not even a nanosecond on 10-year old hardware.

Hence I don't think it's worthwhile while octet-string is supported in all bigint tools.

EIPS/eip-5843.md
```diff
@@ -0,0 +1,448 @@
...
+a       = pow(x, -1, y)
+```
+
+```
+def extended_euclidean_algorithm():
+	pass #TODO
+```
+
+TODO:
+* Some derivation of complexity for extended euclidean (https://www.scaler.com/topics/data-structures/extended-euclidean-algorithm/ claims `O(log(min(x, y))`.
```
mratsim, line 439: The rational is that at each iteration you divide by 2 the number of bits left to process, hence logarithmic. And you stop as soon as one input reaches 1.

EIPS/eip-5843.md
```diff
@@ -0,0 +1,448 @@
...
+### Contract Execution Changes
+
+Upon start of execution (TODO define when exactly this happens wrt various gas checks), if the executing contract code has a `EVMMAX_MOD_PARAMS` section, decode it to an `EVMMAXState` object and associate it with `evmmax_state`:
+
+```
+def decode_mod_params(mod_params_bytes) -> EVMMAXState:
+    # first two bytes are the mod limb count
+    # mod = mod_params_bytes[2:2 + mod_limb_count * 8]
+    offset += mod_limb_count * 8
+    # mod_inv = mod_params_bytes[offset:offset + elem_size]
```
mratsim, line 90: This is not needed

EIPS/eip-5843.md
```diff
@@ -0,0 +1,448 @@
...
+```
+
+```
+def extended_euclidean_algorithm():
+	pass #TODO
+```
+
+TODO:
+* Some derivation of complexity for extended euclidean (https://www.scaler.com/topics/data-structures/extended-euclidean-algorithm/ claims `O(log(min(x, y))`.
+* investigate go standard library which uses Lehmer's algorithm as part of calculating inverse.
```
mratsim, line 440: The bigint modular inverse is not needed for addmodx or mulmodx.
The only inverse that is needed fits in a word and it's cheap to compute it.

Now if an EIP needs full blown modular inverse, Bernstein-Yang and Pornin's GCDs from 2019 and 2020 are significantly faster (like 5x to 10x) than any other algorithm like Lehmer's, see https://github.com/mratsim/constantine/pull/172

EIPS/eip-5843.md
jwasinger: The EVMMAX value format is 64bit-aligned, little endian.

I can see how the references to _limbs_ are confusing.  I'm going to revise the document to delineate between `evmmax_val_size` (the size of EVMMAX values in 64bit increments) and _limbs_ (a base-`1<<system_word_size_bits` value):  _limbs_ should only be referred to in the context of the example pseudo-code.

Good to hear that byteswapping overhead should be minimal.  If this is the case, then I think the format should be modified to be 64bit-aligned big-endian.  This also has the benefit of making the EVMMAX opcodes more interoperable with other EVM ops.

EIPS/eip-5843.md
jwasinger: Can you elaborate? Are you saying that it should be represented more concisely in the spec?

EIPS/eip-5843.md
```diff
@@ -0,0 +1,448 @@
...
+mod_inv_small = pow(-mod, -1, aux_mod)
+mod_inv = pow(-mod, -1, r_val)
+r_squared = (r_val ** 2) % mod
+
+evmmax_state = EVMMAXState(mod, r_squared, mod_inv, mod_inv_small, limb_count)
+```
+
+### Arithmetic Opcodes
+
+`ADDMODX`, `SUBMODX`, `MULMONTX` take inputs encoded in a 3-byte immediate value appended to the opcode.  The immediate is interpreted as 3 1-byte values `out_slot`,`x_slot`,`y_slot`.  Each slot corresponds to a range of the memory space:
```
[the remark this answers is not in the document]
jwasinger: Yes

EIPS/eip-5843.md
jwasinger: The "full" `mod_inv` is needed for generic `mulmont` which is used at larger bit-widths.  I think it would be sensible to modify the spec to not compute it if the bitwidth is below a certain cutoff.

jwasinger:
Hi @mratsim, thanks for the feedback on this work-in-progress draft.  I'm working on completing an initial version of the EIP along with a Geth implementation, EVMMAX support in Huff, and a few other resources.

I have some other things on my plate right now that take precedence but hope to get these out very soon :crossed_fingers:

EIPS/eip-5843.md
mratsim: mod_inv is not needed for Montgomery conversion or compute, only mod_inv_small (the 64-bit value is).

Though if it is wanted I forgot that you can compute in O(log(log(n))) as it is a power of 2, see https://github.com/mratsim/constantine/blob/master/constantine/math/arithmetic/limbs_exgcd.nim#L44-L77

Explanation p11 "Dumas iterations" based on Newton-Raphson:

- Cetin Kaya Koc (2017), https://eprint.iacr.org/2017/411
- Jean-Guillaume Dumas (2012), https://arxiv.org/pdf/1209.6626v2.pdf
- Colin Plumb (1994), http://groups.google.com/groups?selm=1994Apr6.093116.27805%40mnemosyne.cs.du.edu
Other sources:

- https://crypto.stackexchange.com/questions/47493/how-to-determine-the-multiplicative-inverse-modulo-64-or-other-power-of-two
- https://mumble.net/~campbell/2015/01/21/inverse-mod-power-of-two
- http://marc-b-reynolds.github.io/math/2017/09/18/ModInverse.html

EIPS/eip-5843.md
mratsim: Here are 2 implementations of mulmont using CIOS and FIPS, mod_inv is not needed:

https://github.com/mratsim/constantine/blob/928f5155827252c01ee0c7a5c58c8b9deecc928f/constantine/math/arithmetic/limbs_montgomery.nim#L217-L307

m0ninv is the `-1/M[0]` the negative modular multiplicative inverse of the first limb of the prime modulus.

The way modular reduction works is by multiplying with it you can zero put the bottom bits and then shift by a whole word, one word at a time, you don't need to multiply by the full inverse.

EIPS/eip-5843.md
```diff
@@ -0,0 +1,437 @@
...
+
+`ADDMOD` and `MULMOD` have a lot of utility.  However, they are restricted to values that fit within 256 bits.  They also cannot take advantage of optimizations that can greatly improve performance in many cases.
+
+If we add the requirement that inputs must be less than the modulus, modular addition and subtraction can be implemented efficiently with each requiring one extended-precision addition and one extended-precision subtraction. Normal modular multiplication is slower and requires an extended-precision multiplication to compute a double-width product, and a modular reduction.
+
+To improve the performance of modular multiplication, values can be expressed in Montgomery form. This allows for the use of Montgomery modular multiplication which replaces the division by the modulus with a bit-shift and single subtraction in the reduction step.
+
+Operating in Montgomery form requires precomputation of a modulus-specific constant as well as the cost of converting values in canonical form to/from Montgomery form. However, these costs are greatly offset by the savings gained when performing many modular multiplications.
```
xinbenlv, line 38: Reading to this, I feel the current title might be too general, consider update title to be more specific: `EVM Montgomery Modulus Instructions`?

EIPS/eip-5843.md
```diff
@@ -0,0 +1,437 @@
+---
+eip: 5843
+title: EVM Modular Arithmetic Extensions
```
xinbenlv, line 3:
```suggestion
title: EVM Instructions for Montgomery Modulus Arithmetic
```

Maybe also tweak if over 44char, but basically consider make it more specific

EIPS/eip-5843.md
jwasinger: Thanks for the suggestion but I'd prefer to keep as-is for now.  The name EVMMAX is used a lot in the related implementation/toolset code that comes along with this EIP.

cowley86-cmd (approved):

cowley86-cmd (dismissed):

jwasinger:
Superseded by https://github.com/ethereum/EIPs/pull/6601

PR #5850, opened 2022-10-29
Add EIP-5850: Complex Numbers stored in Bytes32 types

genkifs:

 - The build passes.

### DIFF
--- a/EIPS/eip-5850.md
+++ b/EIPS/eip-5850.md
@@ -26,7 +26,7 @@ A complex number would be defined as Bytes32 and initalized with the `cnNew` fun
 
 Only the split is defined in this EIP.  Manipulation of complex numbers (addition, multiplication etc.), number of decimal places and other such topics are left for other EIP discussions.
 
-By defining real numbers in the lower 16bytes this allows direct conversion from positive int to bytes32 for positive real values less than 2**127.  
+Defining real numbers in the 16 least-significant bytes allows direct conversion from `uint128` to `bytes32` for positive integers less than 2**127.  
 
 Direct conversion back from bytes32 -> uint -> int is not recommended as the bytes32 may contain imaginary parts and/or the real part may be negative.  Better to always use `RealIm` for seperating the complex part.  
 
--- a/EIPS/eip-5850.md
+++ b/EIPS/eip-5850.md
@@ -1,7 +1,7 @@
 ---
 eip: 5850
 title: Complex Numbers stored in Bytes32 types
-description: Store real and imaginary parts of complex numbers in the least significant 16 bytes and most significant 16 bytes respectively of a bytes32 type.
+description: Store real and imaginary parts of complex numbers in the least significant and most significant 16 bytes respectively of a bytes32 type.
 author: Paul Edge (@genkifs)
 discussions-to: https://ethereum-magicians.org/t/eip-5850-store-real-and-imaginary-parts-of-complex-numbers-in-the-least-significant-and-most-significant-16-bytes-respectively-of-a-bytes32-type/11532
 status: Draft
@@ -22,7 +22,7 @@ Complex numbers can be represented in many different forms (polynomial, cartesia
 
 Packing complex numbers into a single bytes32 data object halves storage costs and creates a more natural code object that can be passed around the solidity ecosystem.  Existing code may not need to be rewritten for complex numbers.  For example, mappings by bytes32 are common and indexing in the 2D complex plane may improve code legibility.  
 
-Decimal numbers, either fix or floating, are not yet fully supported by Solidity so enforcing similar standards for complex versions is premature.  However, it can be suggested that fixed point methods such as [prb-math](https://github.com/paulrberg/prb-math) be used with 18 decimal places (following ERC-20 convention), or floating point methods like [abdk](https://github.com/abdk-consulting/abdk-libraries-solidity/).  
+Decimal numbers, either fix or floating, are not yet fully supported by Solidity so enforcing similar standards for complex versions is premature.  However, it can be suggested that fixed point methods such as prb-math be used with 18 decimal places (following ERC-20 convention), or floating point methods like abdk.  
 
 ## Specification
 
--- a/EIPS/eip-5850.md
+++ b/EIPS/eip-5850.md
@@ -16,6 +16,8 @@ This EIP proposes a natural way for complex numbers to be stored in and retrieve
 
 ## Motivation
 
+Complex numbers are an essential tool for many mathematical and scientific calculations.  [Fourier Transforms](https://en.wikipedia.org/wiki/Fourier_transform), [Characteristic functions](https://en.wikipedia.org/wiki/Characteristic_function_(probability_theory)), [AC Circuits](http://www.mathforengineers.com/AC-circuits/complex-numbers-in-AC-circuits.html) and [Navier-Stokes equations](https://en.wikipedia.org/wiki/Navier%E2%80%93Stokes_equations) all require the concept.   
+
 Complex numbers can be represented in many different forms (polynomial, cartesian, polar, exponential).  The EIP creates a standard that can accomodate cartesian, polar and exponential formats with example code given for the Cartesian representation, where a complex number is just the pair of real numbers which gives the real and imaginary co-ordinates of the complex number. Equal storage capacity is assigned to both components and the order they appear is explicitly defined.  
 
 Packing complex numbers into a single bytes32 data object halves storage costs and creates a more natural code object that can be passed around the solidity ecosystem.  Existing code may not need to be rewritten for complex numbers.  For example, mappings by bytes32 are common and indexing in the 2D complex plane may improve code legibility.  
--- a/EIPS/eip-5850.md
+++ b/EIPS/eip-5850.md
@@ -60,7 +60,7 @@ This EIP proposes to package both the real and imaginary within one existing dat
 Only the split and position of the real & imaginary parts is defined in this EIP.  Manipulation of complex numbers (addition, multiplication etc.), number of decimal places and other such topics are left for other EIP discussions.  This keeps this EIP more focused and therfore more likely to succeed.
 
 Defining real numbers in the 16 least-significant bytes allows direct conversion from `uint128` to `bytes32` for positive integers less than 2**127.  
-Direct conversion back from `bytes32` -> `uint` -> int are not recommended as the complex number may contain imaginary parts and/or the real part may be negative. It is better to always use `RealIm` for separating the complex part.  
+Direct conversion back from `bytes32` -> `uint` -> `int` are not recommended as the complex number may contain imaginary parts and/or the real part may be negative. It is better to always use `RealIm` for separating the complex part.  
 
 Libraries for complex number manipulation can be implemented with the `Using Complex for bytes32` syntax where `Complex` would be the name of the library.  
 
--- a/EIPS/eip-5850.md
+++ b/EIPS/eip-5850.md
@@ -22,7 +22,7 @@ Complex numbers can be represented in many different forms (polynomial, cartesia
 
 Packing complex numbers into a single bytes32 data object halves storage costs and creates a more natural code object that can be passed around the solidity ecosystem.  Existing code may not need to be rewritten for complex numbers.  For example, mappings by bytes32 are common and indexing in the 2D complex plane may improve code legibility.  
 
-Decimal numbers, either fix or floating, are not yet fully supported by Solidity so enforcing similar standards for complex versions is premature.  However, it can be suggested that fixed point methods such as prb-math be used with 18 decimal places (following ERC-20 convention), or floating point methods like abdk.  
+Decimal numbers, either fix or floating, are not yet fully supported by Solidity so enforcing similar standards for complex versions is premature.  However, it can be suggested that fixed point methods such as prb-math be used with 18 decimal places (following EIP-20 convention), or floating point methods like abdk.  
 
 ## Specification
 
--- a/EIPS/eip-5850.md
+++ b/EIPS/eip-5850.md
@@ -1,9 +1,9 @@
 ---
 eip: 5850
 title: Complex Numbers stored in Bytes32 types
-description: Store real and imaginary parts of complex numbers in the lower 16 bytes and upper 16 bytes respectively of a bytes32 type.
+description: Store real and imaginary parts of complex numbers in the least significant 16 bytes and most significant 16 bytes respectively of a bytes32 type.
 author: Paul Edge (@genkifs)
-discussions-to: https://ethereum-magicians.org/t/TBC
+discussions-to: https://ethereum-magicians.org/t/eip-5850-store-real-and-imaginary-parts-of-complex-numbers-in-the-least-significant-and-most-significant-16-bytes-respectively-of-a-bytes32-type/11532
 status: Draft
 type: Standards Track
 category: ERC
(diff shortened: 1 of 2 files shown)


EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+discussions-to: https://ethereum-magicians.org/t/TBC
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-10-29
+---
+ 
+## Abstract
+
+This EIP proposes a natural way for complex numbers to be stored in the bytes32 data-type.
```
Pandapip1, line 15: This needs to be expanded. This is too short.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Motivation
+
+This EIP proposes to package both the real and imaginary within one existing data type, Bytes32.  This allows compact storage without the need for structures and facilitates easy library implementations.  The bytes32 would remain available for existing, non-complex uses.
```
Pandapip1, line 19: This should be moved to the Rationale section.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Specification
+
+A complex number would be defined as Bytes32 and initalized with the `cnNew` function and converted back with `RealIm`, both given below.
```
Pandapip1, line 23: I don't see the functions in this section. Where are they?

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Rationale
+
+Only the split is defined in this EIP.  Manipulation of complex numbers (addition, multiplication etc.), number of decimal places and other such topics are left for other EIP discussions.
```
Pandapip1, line 27: Why?

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+By defining real numbers in the lower 16bytes this allows direct conversion from positive int to bytes32 for positive real values less than 2**127.
```
Pandapip1, line 29:
```suggestion
Defining real numbers in the 16 least-significant bytes allows direct conversion from `uint128` to `bytes32` for positive integers less than 2**127.  
```

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+A complex number would be defined as Bytes32 and initalized with the `cnNew` function and converted back with `RealIm`, both given below.
+
+## Rationale
+
+Only the split is defined in this EIP.  Manipulation of complex numbers (addition, multiplication etc.), number of decimal places and other such topics are left for other EIP discussions.
+
+By defining real numbers in the lower 16bytes this allows direct conversion from positive int to bytes32 for positive real values less than 2**127.  
+
+Direct conversion back from bytes32 -> uint -> int is not recommended as the bytes32 may contain imaginary parts and/or the real part may be negative.  Better to always use `RealIm` for seperating the complex part.
```
Pandapip1, line 31:
```suggestion
Direct conversion back from bytes32 -> uint -> int are not recommended as the complex number may contain imaginary parts and/or the real part may be negative. It is better to always use `RealIm` for separating the complex part.  
```

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+Direct conversion back from bytes32 -> uint -> int is not recommended as the bytes32 may contain imaginary parts and/or the real part may be negative.  Better to always use `RealIm` for seperating the complex part.  
+
+Libraries for complex number manipulation can be implemented with the `Using Complex for bytes32` syntax where `Complex` is the name of the library.  
+
+## Backwards Compatibility
+There is no impact on other uses of the `bytes32` datatype.
+
+## Reference Implementation
+To create a complex number one would use
```
Pandapip1, line 39: This is not the reference implementation. This should be moved to the specification section.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+```JavaScript
```
Pandapip1, line 41: This should be solidity

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+	function cnNew(int128 _Real, int128 _Imag) public pure returns (bytes32){
+        bytes32 Imag32 = bytes16(uint128(_Imag));
+        bytes32 Real32 = bytes16(uint128(_Real));
+        return (Real32>> 128) | Imag32; 
+	}
+
+```
+and to convert back
+
+```JavaScript
```
Pandapip1, line 52: This should also be solidity

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Backwards Compatibility
+There is no impact on other uses of the `bytes32` datatype.
+
+## Reference Implementation
+To create a complex number one would use
+
+```JavaScript
+
+	function cnNew(int128 _Real, int128 _Imag) public pure returns (bytes32){
```
Pandapip1, line 43: The indentation here is incorrect; please fix it.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+        bytes32 Real32 = bytes16(uint128(_Real));
+        return (Real32>> 128) | Imag32; 
+	}
+
+```
+and to convert back
+
+```JavaScript
+
+	function RealIm(bytes32 _cn)  public pure returns (int128 Real, int128 Imag){
```
Pandapip1, line 54: The indentation here is also incorrect.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+type: Standards Track
+category: ERC
+created: 2022-10-29
+---
+ 
+## Abstract
+
+This EIP proposes a natural way for complex numbers to be stored in the bytes32 data-type.
+
+## Motivation
```
Pandapip1, line 17: The motivation for this EIP is unclear/not obvious. Please add a strong motivation section.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
+---
+eip: 5850
+title: Complex Numbers stored in Bytes32 types
+description: Store real and imaginary parts of complex numbers in the lower 16 bytes and upper 16 bytes respectively of a bytes32 type.
```
Pandapip1, line 4: This should use most significant / least significant bytes, as those are the correct terms

EIPS/eip-5850.md
```diff
@@ -0,0 +1,71 @@
...
+ 
+## Abstract
+
+This EIP proposes a natural way for complex numbers to be stored in the bytes32 data-type.
+
+## Motivation
+
+This EIP proposes to package both the real and imaginary within one existing data type, Bytes32.  This allows compact storage without the need for structures and facilitates easy library implementations.  The bytes32 would remain available for existing, non-complex uses.
+
+## Specification
```
Pandapip1, line 21: I don't think this EIP works, as it can only store complex numbers with integer parts.

Pandapip1 (changes requested):

EIPS/eip-5850.md
genkifs: Yep. Changed

EIPS/eip-5850.md
genkifs: Done.   Reference implementation section removed.

EIPS/eip-5850.md
genkifs: Agree

EIPS/eip-5850.md
genkifs: Moved from the, now deleted, reference implementation section

EIPS/eip-5850.md
genkifs: Ok.  Done. 
For a first time submitter, the line between Motivation and Rationale is not totally clear.  Will Improve.

EIPS/eip-5850.md
genkifs: Some other recent submissions have abstracts of similar length, but I'm happy to expand.

EIPS/eip-5850.md
genkifs: Agreed

EIPS/eip-5850.md
genkifs: > it can only store complex numbers with integer parts

Almost, it can only store numbers that have an *int128* representation.  The problem is that decimal numbers, either fix or floating, are not yet fully supported by Solidity so enforcing similar standards for complex versions is premature.  However, I've now modified the EIP to *suggest* that fixed point methods for *int* types (such as [prb-math](https://github.com/paulrberg/prb-math)) be used with 18 decimal places, matching the *erc-20* decimal suggestion.

Your comment made me rethink the wording.  I've made changes so that both polar and exponential storage are valid as well as the more "natural" cartesian approach.  Keeping the wording open allows this EIP to remain flexible for future EIPs to build upon.

EIPS/eip-5850.md
genkifs: Motivation section extended.  I'm still not clear what is considered motivation and what is Rationale, but I'm happy to keep on improving the text.

EIPS/eip-5850.md
genkifs: To keep the EIP focused on a single simple idea.  Manipulation of complex numbers when they are in Byte32 form is undoubtedly a *much* larger topic.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,75 @@
...
+category: ERC
+created: 2022-10-29
+---
+ 
+## Abstract
+
+This EIP proposes a natural way for complex numbers to be stored in and retrieved from the bytes32 data-type.  It splits the storage space exactly in half and, most importantly, assigns the real number part to the least significant 16 bytes and the imaginary number part to the most significant 16 bytes. 
+
+## Motivation
+
```
Pandapip1, line 18: I still don't see a particularly compelling use case that involves manipulating complex numbers on the blockchain.

genkifs: I've added these use cases to the EIP [Fourier Transforms](https://en.wikipedia.org/wiki/Fourier_transform), [Characteristic functions](https://en.wikipedia.org/wiki/Characteristic_function_(probability_theory\)), [AC Circuits](http://www.mathforengineers.com/AC-circuits/complex-numbers-in-AC-circuits.html), [Navier-Stokes equations](https://en.wikipedia.org/wiki/Navier%E2%80%93Stokes_equations). 

Note that this EIP doesn't discuss manipulation of complex numbers, just storage.

EIPS/eip-5850.md
```diff
@@ -0,0 +1,77 @@
...
+Complex numbers are an essential tool for many mathematical and scientific calculations.  For example, Fourier Transforms, Characteristic functions, AC Circuits and Navier-Stokes equations all require the concept.
+
+Complex numbers can be represented in many different forms (polynomial, cartesian, polar, exponential).  The EIP creates a standard that can accomodate cartesian, polar and exponential formats with example code given for the Cartesian representation, where a complex number is just the pair of real numbers which gives the real and imaginary co-ordinates of the complex number. Equal storage capacity is assigned to both components and the order they appear is explicitly defined.  
+
+Packing complex numbers into a single bytes32 data object halves storage costs and creates a more natural code object that can be passed around the solidity ecosystem.  Existing code may not need to be rewritten for complex numbers.  For example, mappings by bytes32 are common and indexing in the 2D complex plane may improve code legibility.
```
xinbenlv, line 23: Hi, thank you for the EIP draft!

My question is regarding movitation: Could you help give more concrete examples of real world usage of on-chain imaginary numbers?

Also, is EIP/ERC the best place or is it better to be proposed as a Solidity/Vyper language feature?

@axic FYI

genkifs: >Could you help give more concrete examples of real world usage of on-chain imaginary numbers?

Fundamentally, this EIP is essential if you want to track/control/record a variable (but more likely an array) that is complex in nature.  

This EIP was focused on defining the storage aspect of the complex number building block, not how or why it would be manipulated on-chain.  However, there is no harm discussing use cases outside of the EIP document.  

- Complex numbers are invaluable for electrical engineering specifications and calculations.  A live certified safety standard defined as a working calculation (using the gasless view functionality) is much more valuable than a dead pdf equation.  

- Time-series analysis use Fourier Transforms for calculations and statistical distributions are defined via their complex number based characteristic function.  

- Linear algebra can easily enter into the realm of complex numbers.  

- Polynomials with non-real roots shouldn't necessarily interrupt your algorithm.  

- Code can be simpler, more readable and reusable if complex numbers are used (e.g. 2D rotations).    

Even if you believe the above sorts of calculations will never be performed on-chain, some intermediate and final results will likely be reported as complex numbers.  

But the simplest (and maybe most useful) use case/hack would be to define a 2d space within a single `bytes32` data type.

 >proposed as a Solidity/Vyper language feature

Even though I gave the specification example in solidity, this EIP is language independent. The **EVM** stores data in 32 byte chunks.  If we standardize that the least significant 16 bytes are Real and the most significant 16 bytes are Imaginary then any higher level language (Solidity/Vyper/etc.) can follow this implementation.  If this were just a Solidity standard then other contracts written in Vyper would have difficulty understanding and interacting with these `bytes32` objects.

Pandapip1: I think @xinbenlv's point was that this should be proposed as a feature either on the solidity or vyper (or both) repos. I think this could be an EIP, but the fact that it doesn't support floating point is a major hurdle to getting this adopted.

genkifs: >should be proposed as a feature either on the solidity or vyper (or both) repos

I believe this is at a language independent level, but as long as the standard is set with real numbers in the lower 16 bytes and imaginary numbers in the upper 16 bytes then I'm not that bothered where it lives.  

>the fact that it doesn't support floating point is a major hurdle

Neither fixed nor floating point are specified by the EIP.  As currently defined, any number representation can be supported, so long as it fits inside the 16 bytes space.

Twitter1268: ARTV

Twitter1268:
我的小姐姐们都是我的小可爱们都是我的小可爱们都是我的小可爱们都是我的小可爱

EIPS/eip-5850.md
Pandapip1: > As currently defined, any number representation can be supported, so long as it fits inside the 16 bytes space.

This was unclear when reading the EIP. I would highly suggest indicating this in the spec.

genkifs: Clarification added to the spec

Pandapip1 (approved):
I really feel this standard is too under-specified to be useful as it stands right now, but this meets the minimum requirements of a draft.

EIPS/eip-5850.md
Erik4569: EIP-5850

EIPS/eip-5850.md
Erik4569: EIP-5850

EIPS/eip-5850.md
Erik4569: Agree EIP-5850

Erik4569 (commented):
Agree

genkifs:
> I really feel this standard is too under-specified to be useful as it stands right now,

If the community evolves this into a stronger, more detailed specification then that is fine by me.

My preference would be for a signed fixed point, 18 decimal place number format but I don't want to force this as other representations may be better suited.

PR #5863, opened 2022-11-01
[EIP-4844] Reduce Throughput

adietrichs:
During discussions at and after Devcon VI, it became clear that there is a desire to lower the initial blob throughput to ensure network stability. This PR reduces the max throughput from 2 MB to 0.5 MB per block, as well as the target from 1 MB to 0.25 MB per block.

The expectation would be to increase the throughput over the next few hardforks back to its original value. In case the EIP is not included into Shanghai, higher initial values might be appropriate, as there will be more lead time for network stress testing and CL client hardening.


djrtwo (commented):
Although I think this is prudent and likely a better value than the current 1MB placeholder.

I'm fine updating the value to this prior to network/simulation analysis, but happy to wait another couple of weeks for results

adietrichs:
Consensus on the last call (https://github.com/ethereum/pm/issues/655) was to move forward with this, with the understanding that it might be updated again in the future once we have throughput metrics.

lightclient (approved):

PR #5867, opened 2022-11-02
Fix typo

thomasWos:
Fix a minor typo


SamWilsn:
iff is shorthand for if-and-only-if

thomasWos:
learned something today, thanks

PR #5874, opened 2022-11-04
Update EIP-1: Add first two external origins

SamWilsn:
If we're okay with this format, I'll update `eipw` and get the csl-json to render into a human readable citation.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -204,7 +204,7 @@ Which renders to:
 
 [Beacon Chain](https://github.com/ethereum/consensus-specs/blob/26695a9fdb747ecbe4f0bb9812fedbc402e5e18c/specs/sharding/beacon-chain.md)
 
-Permitted Consensus Layer Specifications URLs must match this regular expression:
+Permitted Consensus Layer Specifications URLs must anchor to a specific commit, and so must match this regular expression:
 
 ```regex
 ^https://github.com/ethereum/consensus-specs/blob/[0-9a-f]{40}/.*$
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -188,7 +188,97 @@ EIPs may have a `requires` header, indicating the EIP numbers that this EIP depe
 
 ## Linking to External Resources
 
-Links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
+Other than the specific exceptions listed below, links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
+
+The process governing permitted external resources is described in [EIP-5757](./eip-5757.md).
+
+### Consensus Layer Specifications
+
+Links to the Ethereum Consensus Layer Specifications may be included using normal markdown syntax, such as:
+
+```markdown
+[Beacon Chain](https://github.com/ethereum/consensus-specs/blob/26695a9fdb747ecbe4f0bb9812fedbc402e5e18c/specs/sharding/beacon-chain.md)
+```
+
+Which renders to:
+
+[Beacon Chain](https://github.com/ethereum/consensus-specs/blob/26695a9fdb747ecbe4f0bb9812fedbc402e5e18c/specs/sharding/beacon-chain.md)
+
+Permitted Consensus Layer Specifications URLs must match this regular expression:
+
+```regex
+^https://github.com/ethereum/consensus-specs/blob/[0-9a-f]{40}/.*$
+```
+
+### Digital Object Identifier System
+
+Links qualified with a Digital Object Identifier (DOI) may be included using the following syntax:
+
+````markdown
+This is a sentence with a footnote.[^1]
+
+[^1]:
+    ```csl-json
+    {
+      "type": "article",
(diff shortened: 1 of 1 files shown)


EIPS/eip-1.md
```diff
@@ -188,7 +188,97 @@ EIPs may have a `requires` header, indicating the EIP numbers that this EIP depe
 
 ## Linking to External Resources
 
-Links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
+Other than the specific exceptions listed below, links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
+
+The process governing permitted external resources is described in [EIP-5757](./eip-5757.md).
+
+### Consensus Layer Specifications
```
lightclient, line 195: Might be good to explicitly state that CL spec links need to be anchored to a specific commit.

lightclient (commented):
One nit; generally LGTM though.

EIPS/eip-1.md
```diff
@@ -188,7 +188,97 @@ EIPs may have a `requires` header, indicating the EIP numbers that this EIP depe
...
+
+```markdown
+[Beacon Chain](https://github.com/ethereum/consensus-specs/blob/26695a9fdb747ecbe4f0bb9812fedbc402e5e18c/specs/sharding/beacon-chain.md)
+```
+
+Which renders to:
+
+[Beacon Chain](https://github.com/ethereum/consensus-specs/blob/26695a9fdb747ecbe4f0bb9812fedbc402e5e18c/specs/sharding/beacon-chain.md)
+
+Permitted Consensus Layer Specifications URLs must match this regular expression:
```
SamWilsn, line 207:
```suggestion
Permitted Consensus Layer Specifications URLs must anchor to a specific commit, and so must match this regular expression:
```

lightclient (approved):

EIPS/eip-1.md
```diff
@@ -188,7 +188,97 @@ EIPs may have a `requires` header, indicating the EIP numbers that this EIP depe
...
+        ]
+      },
+      "URL": "https://sly-hub.invalid/00.0000/a00000-000-0000-y",
+      "custom": {
+        "additional-urls": [
+          "https://example.com/an-interesting-article.pdf"
+        ]
+      }
+    }
+    ```
```
Pandapip1, line 277: Nit: don't indent the code block.
```suggestion
```csl-json
{
  "type": "article",
  "id": 1,
  "author": [
    {
      "family": "Jameson",
      "given": "Hudson"
    }
  ],
  "DOI": "00.0000/a00000-000-0000-y",
  "title": "An Interesting Article",
  "original-date": {
    "date-parts": [
      [2022, 12, 31]
    ]
  },
  "URL": "https://sly-hub.invalid/00.0000/a00000-000-0000-y",
  "custom": {
    "additional-urls": [
      "https://example.com/an-interesting-article.pdf"
    ]
  }
}
```
```

Comment (SamWilsn) [reply EIPS/eip-1.md:277]: For the code block to be part of the footnote it needs to be indented. It's very non-standard markdown :(

Comment (Pandapip1) [reply EIPS/eip-1.md:277]: Does Commonmark support footnotes? The EIPs website doesn't use GFM.

Comment (SamWilsn) [reply EIPS/eip-1.md:277]: If we use `kramdown`, I think they'll work: https://kramdown.gettalong.org/quickref.html#footnotes

Comment (Pandapip1) [reply EIPS/eip-1.md:277]: We switched to commonmark because it runs 25% quicker, and is run with the longest-running check.

**Comment by kdenhartog** [comment]
Looks like this one is moving pretty fast for the first draft. Anyone mind if I add sections for W3C/IETF as well to this or do people think it's easier to do this in a separate EIP? I'm cautious that we're going to end up dispersing governance across many EIPs and somethings may be lost, so think doing these all in a single EIP would be better. Wdyt?

**Comment by xinbenlv** [comment]
Thank you for awesome effort!

With such complex details about origin, shall it be better to be a seperate json or yaml structura file to be easily integrat with EIPW and website rendering? 

@SamWilsn @Pandapip1

**Comment by Pandapip1** [comment]
> Do people think it's easier to do this in a separate EIP?

Probably - that's what EIP-5757 suggests.

File: EIPS/eip-1.md [hunk 4]
```diff
@@ -188,7 +188,97 @@ EIPs may have a `requires` header, indicating the EIP numbers that this EIP depe
 
 ## Linking to External Resources
```
xinbenlv, line 189: Propose to add [ethereum/devp2p](https://github.com/ethereum/devp2p) as allowable origin

lightclient:
What else are we waiting for here?

Pandapip1 (changes requested):
I still would like confirmation that footnotes work.

SamWilsn:
> I still would like confirmation that footnotes work.

How's this: https://samwilsn.github.io/EIPs/EIPS/eip-1

Pandapip1 (approved):
Footnotes are a bit awkward, but LGTM

lightclient:
Pressing the button.

PR #5877, opened 2022-11-05
Include Ganache private chains (5777) in EIP-155

rickmark:

 - The build passes.

### DIFF
--- a/EIPS/eip-155.md
+++ b/EIPS/eip-155.md
@@ -65,6 +65,7 @@ This would provide a way to send transactions that work on Ethereum without work
 | 5              | Goerli                                     |
 | 42             | Kovan                                      |
 | 1337           | Geth private chains (default)              |
+| 5777           | Ganache GUI private chains (default)       |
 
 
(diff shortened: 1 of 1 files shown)


rickmark:
Not sure this is still valid, Ganache uses 1337/5777 which means the network id is common but not the same as the chain...

EIPS/eip-155.md
```diff
@@ -65,6 +65,7 @@ This would provide a way to send transactions that work on Ethereum without work
 | 5              | Goerli                                     |
 | 42             | Kovan                                      |
 | 1337           | Geth private chains (default)              |
+| 5777           | Ganache GUI private chains (default)       |
```
Pandapip1, line 68: The rationale for this change is unclear. Why is this necessary? Does Gananche not already implement this?

SamWilsn:
We've discussed this change on today's EIPIP, and we will not be modifying EIP-155 to add this chain id.

EIPs are not the best place to maintain tables that are intended to change over time.

PR #5878, opened 2022-11-05
EIP-3540: Clarify contract creation failure

chfast:
Clarify that in case of any of the new contract creation failure conditions all gas provided to the create instruction/transaction is consumed.

Current implementations do not consume all gas for the first condition but do for the second so this is a functional change.

### DIFF
--- a/EIPS/eip-3540.md
+++ b/EIPS/eip-3540.md
@@ -129,8 +129,12 @@ For clarity, the *container* refers to the complete account code, while *code* r
 
 For clarity, the *EOF prefix* together with a version number *n* is denoted as the *EOFn prefix*, e.g. *EOF1 prefix*.
 
-1. If *initcode's container* has EOF1 prefix it must be valid EOF1 code.
-2. If *code's container* has EOF1 prefix it must be valid EOF1 code.
+If any of the following conditions is not fulfilled the *contract creation* fails
+and all gas provided to the create instruction/transaction is consumed
+(behavior is similar to initcode execution failure).
+
+1. If *initcode's container* has EOF1 prefix it MUST be valid EOF1 code.
+2. If *code's container* has EOF1 prefix it MUST be valid EOF1 code.
 
 ## Rationale
 


jochem-brouwer:
What is the motivation that all gas has to be consumed?

chfast:
If you consider the initcode validation is part of initcode execution then this is consistent (because any other failure in initcode execution fails the contract creation and consumes all gas provided).
Moreover, after initcode execution there is runtime code validation and it must consume all gas in case of failure.

chfast:
For the reference, the [EIP-3670: EOF - Code Validation](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3670.md?plain=1#L48) has this situation clarified and it proposes **not** consuming all the gas. I wasn't aware of this when proposing this change.
Also in [EIP-3860: Limit and meter initcode](https://eips.ethereum.org/EIPS/eip-3860#specification) similar check result in "light" failure **not** consuming all the gas.

Nevertheless, we should discuss it and pick better variant.

chfast:
There is also performance aspect of this: for initcode execution we need to validate the code and "load" it into internal representation. It may be more efficient to do both in single pass. This do not prevent implementing it with "light" failure, but is strong indicator that initcode validation is part of execution.

gumb0:
It was originally motivated by suggestions in 3860 discussion, for failures in case initcode larger than limit: https://github.com/ethereum/tests/pull/990#issuecomment-1033181178

So I think arguments in favor not consuming gas for invalid initcode were like:
- no other error checked before initcode execution consumes all gas (maybe not a very big deal still to add one)
- it's arguably better UX - don't burn all gas if we don't have to. (might be especially painful for creation transactions - easy to make mistake in initcode and all your max gas is burnt)

Cryptonics1:
Ok, you guys are the experts....I would help more but I am trying to secure
tokens, Nfts, dapps, wallets...and fend off the advasaries. I trust you
will make the proper decision, I would imput an addendum that it could be
changed back, if the change is not satisfactory and mutually beneficial to
respective users.

chfast:
Abandoned. We decided to do it the old way, so not functional change is needed.

chfast:
See https://github.com/ethereum/EIPs/pull/5907

PR #5888, opened 2022-11-07
eip-4750: note the gas cost of callf and retf

lightclient:
The EIP currently doesn't specify the gas cost of `CALLF` and `RETF`. ~Per the `evmon` [implementation](https://github.com/ethereum/evmone/blob/b81d112aee89d84cfb6f618d4f792cb84c17f003/lib/evmone/instructions_traits.hpp#L169-L170), I've set them both to 8 gas.~ Per the [EOF v1 checklist](https://notes.ethereum.org/@ipsilon/eof1-checklist), I've set them to 5 and 3 respectively.


EIPS/eip-4750.md
```diff
@@ -92,14 +93,15 @@ Under `PC_post_instruction` we mean the PC position after the entire immediate a
...
-6. Sets `current_section_index` to `code_section_index` and `PC` to `0`, and execution continues in the called section.
+7. Sets `current_section_index` to `code_section_index` and `PC` to `0`, and execution continues in the called section.
 
 #### `RETF`
 
 1. Does not have immediate arguments.
 2. If data stack has less than `caller_stack_height + types[code_section_index].outputs`, execution results in exceptional halt.
-3. Pops nothing and pushes nothing to data stack.
-4. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.
+3. Charges 8 gas.
```
xinbenlv, line 102: Good finding. @lightclient , QQ: do you plan to also add justification for the choice of `8` as gas cost? (e.g. comparing it to `JUMP`) or do you intend to let Author add the justification?

lightclient: I'll let the authors add justification.

axic (dismissed):

axic (approved):

xinbenlv (changes requested):
Since this is soon to be implemented, I'd suggest author / contributor to provide gas cost choice plus rationale in the same PR or make it a placeholder signaling gas cost choice was arbitrary and it's pending future change, in chase clients start to implement it

axic:
> Since this is soon to be implemented, I'd suggest author / contributor to provide gas cost choice plus rationale in the same PR or make it a placeholder signaling gas cost choice was arbitrary and it's pending future change, in chase clients start to implement it

You can find the reasoning on the [EOFv1 checklist](https://notes.ethereum.org/@ipsilon/eof1-checklist). It will be added later.

To be honest, most EIPs don't provide a strong reasoning. I appreciate your comment, but you make it seem like we make up numbers randomly, which is definitely not the case.

xinbenlv:
Appreciate the response. 



IMHO: Gas Cost choice affects economy and thus I think from a editorial standpoint, if gas cost is chosen, EIP should document why. Such documentation, or a placeholder of "there will be a documentation", preferably live in the EIP itsel, if the principle is for a EIP should be sufficiently self-contained for important decision and rationale. Just my two cents 🙏

PR #5894, opened 2022-11-08
EIP-3670: Reject CALLCODE and SELFDESTRUCT

chfast:


gumb0 (dismissed):
Looks good to me.

lightclient (dismissed):

EIPS/eip-3670.md
```diff
@@ -44,11 +44,13 @@ The EOF1 format provides following forward compatibility properties:
 
 This feature is introduced on the very same block EIP-3540 is enabled, therefore every EOF1-compatible bytecode MUST be validated according to these rules.
 
+Previously deprecated instructions `CALLCODE` (0xf2) and `SELFDESTRUCT` (0xff) are invalid and their opcodes do not have any instruction assigned.
```
axic, line 47: While I like this, I'm not sure we need to take a stance here before the selfdestruct removal discussion concludes. In case that change goes live at the same time, do we expect it to be allowed within EOF or not?

lightclient: I think there is general consensus that we want less self-destructing regardless though? Of course there is the debate on how to make a backward-incompatible change to the legacy EVM, but it seems like a good thing to just go ahead and get removed in EOF v1. If people decide they really want it, it could be added back without a version bump!

chfast:
Adding `SELFDESTRUCT` back (in any form) is backward compatible.

PR #5895, opened 2022-11-08
CI: Warn on extra sections

Pandapip1:
Uses the new deny/warn/allow system.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           unchecked: 1, 5069, 5757
-          warn-checks: "markdown-rel-links,markdown-order-section"
+          warn-checks: "markdown-order-section"
 
   markdownlint:
     name: Markdown Linter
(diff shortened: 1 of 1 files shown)


SamWilsn:
May I ask why we're switching to warnings?

Pandapip1:
> May I ask why we're switching to warnings?

* Some external links are now allowed
* EIP-1 doesn't say that you can't use sections not defined in EIP-1

SamWilsn:
> * Some external links are now allowed

Those exceptions are already supported:

https://github.com/ethereum/eipw/blob/4e4bc7b50a094ea9b02551113780b380a096120d/eipw-lint/src/lib.rs#L360-L363

> * EIP-1 doesn't say that you can't use sections not defined in EIP-1

While true, I'm not convinced we should allow extra sections without specific editor approval?

Pandapip1:
>  I'm not convinced we should allow extra sections without specific editor approval?

Agreed, but on the other hand I'm not convinced that we should require a manual merge to allow extra sections.

Pandapip1:
@SamWilsn

SamWilsn (changes requested):
If we want to allow extra sections, that's a whole discussion, but I'm firmly against allowing out-of-order sections.

Pandapip1:
I too am firmly against out of order sections. Might it be possible to separate those two lints?

Pandapip1:
@SamWilsn see my previous question

PR #5896, opened 2022-11-08
Add EIP-5896: Rejectable Non-Fungible Token

miquelcabot:
Discussion: https://ethereum-magicians.org/t/eip-5896-rejectable-non-fungible-token/11674

This improvement of the EIP-721 includes the possibility that the receiver could reject a token transfer, allowing the selective reception of NFTs. The proposal depends on and extends the existing [EIP-721](https://eips.ethereum.org/EIPS/eip-721).


EIPS/eip-5896.md
```diff
@@ -0,0 +1,163 @@
+---
+eip: 5896
+title: Rejectable Non-Fungible Token
```
Pandapip1, line 3: Is this not possible by reverting with the EIP-721 receive hook?

miquelcabot: @Pandapip1 
The `beforetokenTransfer()` hook allows modifying the behavior of a token before the token is transferred, for example, stopping the transfer. This can let us restrict transfer to only registered candidates, or only allow tokens that are already accepted by the receiver.
But it doesn't help us to do what we want to do in this proposal, because we propose that the first step must be the call to the `transfer()` function by the sender, and to use this hook, we would need a first step of the receiver calling an `acceptance()` function.

miquelcabot: @Pandapip1

miquelcabot: @Pandapip1

miquelcabot: @Pandapip1 With this proposal there is a “Selective Receipt” of a token. That means that before a token is transferred to the intended recipient, this receipt must explicitly accept the token (i.e. recipient must call the `acceptance()` function). This action represents non-repudiation evidence and it cannot be performed inside the `beforetokenTransfer()` hook.

miquelcabot: @xinbenlv There are situations where the sender must prove that he has tried to send the NFT, regardless receiver accepts the reception. For example, when a sender sends a certified notification. He needs to prove that he attempted to send it on a specific date.
In addition, the receiver must be able to reject or not the NFT transfer because it can represent a negative value.
Therefore, the objective of this proposal cannot be solved with a revert of the sender’s transaction, since a record of the sender’s transaction and a record of the receiver’s transaction are needed.
With this standard proposal, we achieve a non-repudiation of origin and reception proof when we send an NFT.

PR #5757, opened 2022-10-06
Propose process for allowing external links

SamWilsn:

PR #5903, opened 2022-11-09
Update EIP-721: delete broken openzeppelin links

konojunya:
In the NFT Implementation and Other Projects column, the 17th link is to openzeppelin, which is already an invalid link and also a link to SafeERC20. openzeppelin has an ERC721 implementation.
The link should be to ERC721.sol

ref: https://eips.ethereum.org/EIPS/eip-721#references

### DIFF
--- a/EIPS/eip-721.md
+++ b/EIPS/eip-721.md
@@ -440,7 +440,6 @@ XXXXERC721, by William Entriken -- a scalable example implementation
 1. Curio Cards. https://mycuriocards.com
 1. Rare Pepe. https://rarepepewallet.com
 1. Auctionhouse Asset Interface. https://github.com/dob/auctionhouse/blob/master/contracts/Asset.sol
-1. OpenZeppelin ERC721.sol Implementation. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol
 
 ## Copyright
 
--- a/EIPS/eip-721.md
+++ b/EIPS/eip-721.md
@@ -440,7 +440,7 @@ XXXXERC721, by William Entriken -- a scalable example implementation
 1. Curio Cards. https://mycuriocards.com
 1. Rare Pepe. https://rarepepewallet.com
 1. Auctionhouse Asset Interface. https://github.com/dob/auctionhouse/blob/master/contracts/Asset.sol
-1. OpenZeppelin SafeERC20.sol Implementation. https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/token/ERC20/SafeERC20.sol
+1. OpenZeppelin ERC721.sol Implementation. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol
 
 ## Copyright
 


Pandapip1 (dismissed):
+0. I would prefer the links to be removed altogether, but I'd rather the link be fixed than a broken link stay.

fulldecent (dismissed):

EIPS/eip-721.md
```diff
@@ -440,7 +440,7 @@ XXXXERC721, by William Entriken -- a scalable example implementation
 1. Curio Cards. https://mycuriocards.com
 1. Rare Pepe. https://rarepepewallet.com
 1. Auctionhouse Asset Interface. https://github.com/dob/auctionhouse/blob/master/contracts/Asset.sol
-1. OpenZeppelin SafeERC20.sol Implementation. https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/token/ERC20/SafeERC20.sol
```
xinbenlv, line 443: Thank you for the contribution! 
This is a historical doc, to fix it
1. Editors are working on a link policy per EIP-5757 (#5757 ), let's wait for that to be finalized
2. Links are generally suggested to use permlink to reduce chance of such broken links in the future

But still, thank you for your contribution! Greatly appreciated

konojunya: @xinbenlv Yes. Exactly. This EIP is referenced by many people and we should wait for the Editors reply as it is a change to a part of that document.
I too recommend using permlink, but since the source of the change points to the default branch, I made the change using the same policy.

konojunya:
@SamWilsn @axic @gcolvin @lightclient 

How are you?
Let us know what you think!

SamWilsn:
I *think* my preference for this would be to remove the links entirely. I don't want to establish a precedent of updating broken links in old EIPs.

konojunya:
@SamWilsn In my opinion, I leave the update or deletion to the outcome of the discussion, but the current situation with the wrong link should be changed.
Even if it were removed, the openzeppelin implementation of EIP-721 would have little impact because it is widely known, but the EIP-721 page does not have a reference implementation like [EIP-4907](https://eips.ethereum.org/EIPS/eip-4907#reference-implementation), so newcomers could lose the link to a valuable source of information.

Pandapip1:
> I think my preference for this would be to remove the links entirely. I don't want to establish a precedent of updating broken links in old EIPs.

This is my preference too. However, I would like to establish a precedent for updating broken links in old EIPs. I would just like there to *also* be a precedent of deleting those old links too.

EIPS/eip-721.md
gcolvin: > Links are generally suggested to use permlink to reduce chance of such broken links in the future

So, in this case, would the better fix be to replace the broken link with a permlink to  the last SafeERC20, rather than a link to ERC721?

And, given these links are no normative, and this one is not discussed in the proposal, I'd be happy to see it removed.

konojunya:
I understand your basic policy. I will add a commit in the form of a deletion rather than an update to the link.

Pandapip1 (approved):

fulldecent:
Please close this PR in favor of https://github.com/ethereum/EIPs/pull/6012

---

The link to ERC-20 implemented by OpenZeppelin is correct and intentional.

This citation is an important part of the justification of how ERC-721 (and frankly, all) token standards currently work. The text explains the risks of the old ERC-20 behavior and advocates that when transactions fail they should throw, something many now take for granted.

PR #5906, opened 2022-11-09
EIP-4844: declare withdrawals in 4895 as a dependency

ralexstokes:
Revise 4844 to include 4895 as a dependency; including reflecting that EIP's effects on the header format

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -8,7 +8,7 @@ status: Draft
 type: Standards Track
 category: Core
 created: 2022-02-25
-requires: 1559, 2718, 2930
+requires: 1559, 2718, 2930, 4895
 ---
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


Cryptonics1 (approved):

EIPS/eip-4844.md
```diff
@@ -193,21 +193,22 @@ The resulting RLP encoding of the header is therefore:
 ```
 rlp([
     parent_hash,
-    ommers_hash,
+    0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347, # ommers hash
```
protolambda, line 196: If we're going to hardcode post-merge header values in this EIP we should maybe refer to [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) that defines them.

protolambda (commented):
LGTM, one nit about header values.

Inphi (approved):

EIPS/eip-4844.md
ralexstokes: this was downstream of a request in 4895 to hardcode the values

I'm happy to add a note about the EIP, although if we want to standardize on this presentation then it will be a little silly to always reference 3675 in all future EIPs writing the header

protolambda (approved):

lightclient (approved):

PR #5912, opened 2022-11-10
Update EIP-5749: Moving to Review stage

kvhnuke:
Moving [./EIP-5749](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5749.md) to review stage

### DIFF
--- a/EIPS/eip-5749.md
+++ b/EIPS/eip-5749.md
@@ -4,7 +4,7 @@ title: The 'window.evmproviders' object
 description: Add 'window.evmproviders' and suggest the eventual removal of 'window.ethereum'
 author: Kosala Hemachandra (@kvhnuke), Brett Kolodny (@brettkolodny)
 discussions-to: https://ethereum-magicians.org/t/eip-5749-deprecate-window-ethereum/11195
-status: Draft
+status: Review
 type: Standards Track
 category: Interface
 created: 2022-10-04


DONALDSULLENDER (approved):

lightclient (dismissed):
A few small pieces of feedback:
* The title and description should be more descriptive of what is actually happening. It seems like what you're really trying to do is make it possible to have multiple providers. The name of that injected value should just be in the spec.
* You should not use "evm" to refer to Ethereum. The EVM just executes code related to the blockchain. The `ethereum` provider interface gives a much broader scope of information like full blocks, tx status, etc.

Pandapip1 (approved):
Looks fine by me.

kvhnuke:
> A few small pieces of feedback:
> 
> * The title and description should be more descriptive of what is actually happening. It seems like what you're really trying to do is make it possible to have multiple providers. The name of that injected value should just be in the spec.
> * You should not use "evm" to refer to Ethereum. The EVM just executes code related to the blockchain. The `ethereum` provider interface gives a much broader scope of information like full blocks, tx status, etc.

we actually went through multiple different titles, with @Pandapip1's input on the draft merge we thought it'll probably be better to have  `evmproviders` on the title, it might help new comers to easily search it via search engines.

we mainly chose `evmproviders` is due to the fact that evm is the only commonality among ethereum forks and all major rpc calls does require evm execution to return a response. At the same time, some ethereum forks doesnt want to associate themselves with `eth` so a value such as `ethproviders` might not help to get this EIP adopted by those groups.

PR #5917, opened 2022-11-11
Update EIP-5216 with address zero control in _approve function like EIP-20 

ivanmmurciaua:
After discussing _approve function requires here : https://ethereum-magicians.org/t/last-call-eip-5216-erc1155-approval-by-amount/9898/3 , I'm just updating it

### DIFF
--- a/assets/eip-5216/ERC1155ApprovalByAmount.sol
+++ b/assets/eip-5216/ERC1155ApprovalByAmount.sol
@@ -68,8 +68,10 @@ abstract contract ERC1155ApprovalByAmount is ERC1155, IERC1155ApprovalByAmount {
             from == msg.sender || isApprovedForAll(from, msg.sender) || allowance(from, msg.sender, id) >= amount,
             "ERC1155: caller is not owner nor approved nor approved for amount"
         );
-        unchecked {
-            _allowances[from][msg.sender][id] -= amount;
+        if(from != msg.sender && !isApprovedForAll(from, msg.sender)) {
+            unchecked {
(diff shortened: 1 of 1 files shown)


assets/eip-5216/ERC1155ApprovalByAmount.sol
```diff
@@ -108,7 +110,9 @@ abstract contract ERC1155ApprovalByAmount is ERC1155, IERC1155ApprovalByAmount {
 
         require(idsLength == amountsLength, "ERC1155ApprovalByAmount: ids and amounts length mismatch");
         for (uint256 i = 0; i < idsLength;) {
-            require(allowance(from, to, ids[i]) >= amounts[i], "ERC1155ApprovalByAmount: operator is not approved for that id or amount");
+            if(allowance(from, to, ids[i] < amounts[i])) {
```
xinbenlv, line 113: What's the rationale to not use `_allowances` instead of using `allowances()`?

ivanmmurciaua: Mmm that's true, thanks for feedback. Also that `if` won't work fine haha. I'll fix and I'll include new commit!

ivanmmurciaua: Done @xinbenlv . Thank you very much for your feedback!

ivanmmurciaua:
I think that this EIP would be ready to `final` status. Also the deadline is outdated :)

ivanmmurciaua:
Ready for a review!

ivanmmurciaua:
Yes, review it

PR #5920, opened 2022-11-11
Add EIP-5920: PAY opcode

Pandapip1:
Adds a new opcode to do simple ether transfers.

discussions-to: https://ethereum-magicians.org/t/eip-5920-pay-opcode/11717

### DIFF
--- a/EIPS/eip-5920.md
+++ b/EIPS/eip-5920.md
@@ -24,20 +24,26 @@ Currently, to send ether to an address requires you to call a function of that a
 | ------------------- | ------- |
 | `FORK_BLKNUM`       | TBD     |
 | `PAY_OPCODE`        | TBD     |
-| `BASE_GAS_COST`     | `9000`  |
-| `COLD_GAS_COST`     | `2600`  |
-| `CREATION_GAS_COST` | `25000` |
+| `BASE_GAS_COST`     | `8600`  |
+| `COLD_GAS_COST`     | `11100`  |
+| `CREATION_GAS_COST` | `32600` |
 
 A new opcode is introduced: `PAY` (`PAY_OPCODE`), which:
 
 - Pops two values from the stack: `addr` then `val`.
 - Transfers `val` wei to the address `addr`.
 
-The cost of this opcode is `BASE_GAS_COST` if `addr` is a warm account, `BASE_GAS_COST+COLD_GAS_COST` if `addr` is a cold account that is already in the state trie, or `BASE_GAS_COST+CREATION_GAS_COST` if `addr` has not yet been added to the state trie.
+The cost of this opcode is `BASE_GAS_COST` if `addr` is a warm account, `COLD_GAS_COST` if `addr` is a cold account that is already in the state trie, or `BASE_GAS_COST+CREATION_GAS_COST` if `addr` has not yet been added to the state trie.
 
 ## Rationale
 
-The order of arguments is similar to that of the `CALL` family, which have `addr` before `val`. Beyond consistency, though, this ordering aid validators in pattern-matching MEV opportunities, so that the `PAY` always appears immediately after `COINBASE` in the order of execution.
+## Gas pricing
+
+The gas pricing is that of a `CALL` with a positive `msg.value`, but without any memory expansion costs or "gas sent with call" costs, with a gas reduction of `500` to compensate for the reduced amount of computation.
+
+## Argument order
+
+The order of arguments mimicks that of `CALL`, which pops `addr` before `val`. Beyond consistency, though, this ordering aids validators pattern-matching MEV opportunities, so `PAY` always appears immediately after `COINBASE`.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-5920.md
+++ b/EIPS/eip-5920.md
@@ -2,7 +2,7 @@
 eip: 5920
 title: PAY opcode
 description: Introduces a new opcode, PAY, to send ether to an address without calling any of its functions
-author: Pandapip1 (@Pandapip1)
+author: Pandapip1 (@Pandapip1), Zainan Victor Zhou (@xinbenlv)
 discussions-to: https://ethereum-magicians.org/t/eip-5920-pay-opcode/11717
 status: Draft
 type: Standards Track
--- a/EIPS/eip-5920.md
+++ b/EIPS/eip-5920.md
@@ -1,4 +1,5 @@
 ---
+eip: 5920
 title: PAY opcode
 description: Introduces a new opcode, PAY, to send ether to an address without calling it
 author: Pandapip1 (@Pandapip1)
--- a/EIPS/eip-5920.md
+++ b/EIPS/eip-5920.md
@@ -1,7 +1,7 @@
 ---
 eip: 5920
 title: PAY opcode
-description: Introduces a new opcode, PAY, to send ether to an address without calling it
+description: Introduces a new opcode, PAY, to send ether to an address without calling a function
 author: Pandapip1 (@Pandapip1)
 discussions-to: https://ethereum-magicians.org/t/eip-5920-pay-opcode/11717
 status: Draft
--- a/EIPS/eip-5920.md
+++ b/EIPS/eip-5920.md
@@ -20,12 +20,12 @@ Currently, to send ether to an address requires you to call a function of that a
 
 ## Specification
 
-| Parameter | Value |
-| - | - |
-| `FORK_BLKNUM` | TBD |
-| `PAY_OPCODE` | TBD |
-| `BASE_GAS_COST` | `9000` |
-| `COLD_GAS_COST` | `2600` |
+| Parameter           | Value   |
+| ------------------- | ------- |
+| `FORK_BLKNUM`       | TBD     |
+| `PAY_OPCODE`        | TBD     |
+| `BASE_GAS_COST`     | `9000`  |
+| `COLD_GAS_COST`     | `2600`  |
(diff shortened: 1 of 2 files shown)


EIPS/eip-5920.md
```diff
@@ -0,0 +1,51 @@
+---
+eip: 5920
+title: PAY opcode
+description: Introduces a new opcode, PAY, to send ether to an address without calling it
```
xinbenlv, line 4:
```suggestion
description: Introduces a new opcode, PAY, to send ether to an address that does not invoke code call
```

EIPS/eip-5920.md
```diff
@@ -0,0 +1,51 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-5920-pay-opcode/11717
+status: Draft
+type: Standards Track
+category: Core
+created: 2022-03-14
+---
+
+## Abstract
+
+This EIP adds a new opcode, `PAY(addr, val)`, that transfers `val` wei to the address `addr` without calling it.
```
xinbenlv, line 15:
```suggestion
This EIP adds a new opcode at `OPCODE_PlACEHOLDER` as `PAY(addr, val)`, that transfers `val` wei to the address `addr` without calling it.
```

EIPS/eip-5920.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Motivation
+
+Currently, to send ether to an address requires you to call a function of that address, which has a few issues. First of all, it opens a reentrancy attack vector, as the recipient can call back into the sender. Secondly, it opens a DoS vector, so parent functions must be cognizant of the possibility that the recipient will run out of gas or revert. Finally, the `CALL` opcode is needlessly expensive for simple ether transfers, as it requires the memory and stack to be expanded, the recipient's full data including code and memory to be loaded, and finally needs to execute a call, which might do other unintentional operations. Having a dedicated opcode for ether transfers solves all of these issues, and would be a useful addition to the EVM.
```
xinbenlv, line 19: [peer review] Adding a motivation.

```suggestion
Currently, to send ether to an address requires you to call a function of that address, which has a few issues. First of all, it opens a reentrancy attack vector, as the recipient can call back into the sender. Secondly, it opens a DoS vector, so parent functions must be cognizant of the possibility that the recipient will run out of gas or revert. Finally, the `CALL` opcode is needlessly expensive for simple ether transfers, as it requires the memory and stack to be expanded, the recipient's full data including code and memory to be loaded, and finally needs to execute a call, which might do other unintentional operations. Having a dedicated opcode for ether transfers solves all of these issues, and would be a useful addition to the EVM.

Today, to achieve such behavior without core change, solidity introduced lang feature such as nonpayable unless marked as`payable`. This shows the necessity for having such safer measure, but solidity's protection measures unfortunately adds extra gas cost overhead. Having a core-level support reduces gas by potentially significant amount <!-- TODO @Pandapip1 consider add analysis on how much gas it safes -->, which increase global ETH gas efficiency.
```

EIPS/eip-5920.md
```diff
@@ -0,0 +1,51 @@
...
+
```
xinbenlv, line 20: [peer review] Adding a motivation.

```suggestion
Currently, to send ether to an address requires you to call a function of that address, which has a few issues. First of all, it opens a reentrancy attack vector, as the recipient can call back into the sender. Secondly, it opens a DoS vector, so parent functions must be cognizant of the possibility that the recipient will run out of gas or revert. Finally, the `CALL` opcode is needlessly expensive for simple ether transfers, as it requires the memory and stack to be expanded, the recipient's full data including code and memory to be loaded, and finally needs to execute a call, which might do other unintentional operations. Having a dedicated opcode for ether transfers solves all of these issues, and would be a useful addition to the EVM.

Today, to achieve such behavior without core change, solidity introduced lang feature such as nonpayable unless marked as`payable`. This shows the necessity for having such safer measure, but solidity's protection measures unfortunately adds extra gas cost overhead. Having a core-level support reduces gas by potentially significant amount <!-- TODO @Pandapip1 consider add analysis on how much gas it safes -->, which increase global ETH gas efficiency.
```

EIPS/eip-5920.md
```diff
@@ -0,0 +1,51 @@
...
+| - | - |
+| `FORK_BLKNUM` | TBD |
+| `PAY_OPCODE` | TBD |
+| `HOT_GAS_COST` | `500` |
+| `COLD_GAS_COST` | `3000` |
+
+A new opcode is introduced: `PAY` (`PAY_OPCODE`), which:
+
+- Pops two values from the stack: `addr` then `val`.
+- Transfers `val` wei to the address `addr`.
```
xinbenlv, line 33: Since it's replacing calls, do you think it's worthy of clarifiying how `account_nonce`s will be affected?

Pandapip1: This doesn't introduce a new transaction type, only an opcode. Why would account nonces be affected?

EIPS/eip-5920.md
Pandapip1: Why does marking an address as payable increase the gas cost? I thought it was just a compiler protection.

EIPS/eip-5920.md
```diff
@@ -0,0 +1,51 @@
...
+
+The cost of this opcode is `HOT_GAS_COST` if `addr` is a warm account, and `COLD_GAS_COST` if it is a cold account.
```
xinbenlv, line 35: [Peer review]

I think it's worthy of talking about gas costs' economy in the rationale when choosing the gas cost.

Technical context I could think of:

https://eips.ethereum.org/EIPS/eip-2929 which sets 

```
700 gas per BALANCE
```

```
COLD_ACCOUNT_ACCESS_COST 2600
```
as well as the yellow paper mention 

```
G_balance 400 Amount of gas to pay for a BALANCE operation
G_call 700 Paid for a CALL operation
```

EIPS/eip-5920.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Rationale
+
+The order of arguments is similar to that of the `CALL` family, which have `addr` before `val`. Beyond consistency, though, this ordering aid validators in pattern-matching MEV opportunities, so that the `PAY` always appears immediately after `COINBASE` in the order of execution.
+
+## Backwards Compatibility
+
+Needs discussion.
```
xinbenlv, line 43: Another thought came to me: whether we should reject the `PAY` when sending to a non-existing account. e.g. Contract that has not been created, or EOA that has not been identified, such as by checking if `accounce_nonce` is zero.

EIPS/eip-5920.md
Pandapip1: I was under the impression that the BALANCE opcode costs 100 gas. I could be wrong.

EIPS/eip-5920.md
xinbenlv: It's not a compiler protection, there is actually line of code added to do check
Let me find the link.

My point here is that, the fact Solidity does so, demonstrates a strong need

EIPS/eip-5920.md
Pandapip1: I don't think we should reject these payments. There may be good reasons to do so.

Regardless, that cost does need to be reflected.

EIPS/eip-5920.md
xinbenlv: https://github.com/ethereum/solidity/blob/5f8b48793cdf9a41d6f16fa601250987bee82e3e/libsolidity/codegen/ContractCompiler.cpp#L429

Citing my own answer in 
https://github.com/ethereum/EIPs/pull/5805#discussion_r1006934757

EIPS/eip-5920.md
Pandapip1: I've updated the gas cost to reflect uninitialized accounts.

Definitely could use some motivation though. I'll keep this open.

EIPS/eip-5920.md
Pandapip1: I'm pretty sure that line of code refers to adding the check to make sure that non-payable functions aren't called with any `msg.value`, which is used specifically so that wallets can automatically call the correct payable function. That additional bytecode would still be usefully included.

xinbenlv: Yeah, it's still useful to be included. For what I mean is that it shows a `Motivation` to decouple PAY vs CALL with value sometimes.

Pandapip1: I'm still not convinced.

lightclient (approved):

lightclient (changes requested):
Actually sorry -- I see there is already an EIP for this. Have you reached out to the author of EIP-5065? I think it would be good to try and work together on this.

Pandapip1:
> Actually sorry -- I see there is already an EIP for this. Have you reached out to the author of EIP-5065? I think it would be good to try and work together on this.

Oh, I must have missed it. I'll close this and try to push that one forward.

xinbenlv:
@lightclient good point.

Meanwhile, I feel this EIP is much clear in writing, great job. Hope to incorperate a lot of your effort @Pandapip1 done here

Pandapip1:
I'm going to keep this closed in the meantime, but EIP-5065 is nearing on the stagnant deadline, suggesting that the author is not contactable / interested in the EIP anymore. If that happens, I will reopen this.

Pandapip1:
Since there is a PR open to move EIP-5065 to stagnant, I'll open this as a draft.

Pandapip1:
EIP-5065 is now stagnant.

Pandapip1:
Oops, that wasn't meant to happen.

PR #5928, opened 2022-11-14
Update EIP-4750: redefine code section header to be array of code section sizes

lightclient:
I like the idea from discord user `aymen` and @chfast about defining the code section header as an array of code section sizes. Since we're anticipating EIP-4750 to move forward together with EIP-3540, I think it makes sense to do this as it is not necessary to retain a semblance of backwards compatibility.

The new `code_section_header` would look something like this:

```
code_section_header = 1 || (code_section_size)+ || 0
```

Where `||` is a concatenation operator.

### DIFF
--- a/EIPS/eip-4750.md
+++ b/EIPS/eip-4750.md
@@ -30,7 +30,7 @@ Furthermore it aims to improve analysis opportunities by encoding the number of 
 ### EOF container changes
 
 1. The [EIP-3540](./eip-3540.md) code section header is redefined to be an array of code section sizes terminated with a `0x00` byte. E.g. `code_section_header := 1, (code_section_size)+, 0`.
-2. Total number of code sections MUST NOT exceed 1024 and they must all be contiguous.
+2. Total number of code sections MUST NOT exceed 1024 and they MUST all be contiguous.
 3. All code sections MUST precede a data section, if data section is present.
 4. New section with `kind = 3` is introduced called the *type section*.
 5. Exactly one type section MAY be present.
--- a/EIPS/eip-4750.md
+++ b/EIPS/eip-4750.md
@@ -29,12 +29,12 @@ Furthermore it aims to improve analysis opportunities by encoding the number of 
 
 ### EOF container changes
 
-1. The requirement of [EIP-3540](./eip-3540.md) "Exactly one code section MUST be present." is relaxed to "At least one code section MUST be present.", i.e. multiple code sections (`kind = 1`) are allowed.
-2. Total number of code sections MUST NOT exceed 1024.
+1. The [EIP-3540](./eip-3540.md) code section header is redefined to be an array of code section sizes terminated with a `0x00` byte. E.g. `code_section_header := 1, (code_section_size)+, 0`.
+2. Total number of code sections MUST NOT exceed 1024 and they must all be contiguous.
 3. All code sections MUST precede a data section, if data section is present.
 4. New section with `kind = 3` is introduced called the *type section*.
 5. Exactly one type section MAY be present.
-6. The type section, if present, MUST directly precede all code sections.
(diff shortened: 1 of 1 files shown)


EIPS/eip-4750.md
```diff
@@ -29,12 +29,12 @@ Furthermore it aims to improve analysis opportunities by encoding the number of
 
 ### EOF container changes
 
-1. The requirement of [EIP-3540](./eip-3540.md) "Exactly one code section MUST be present." is relaxed to "At least one code section MUST be present.", i.e. multiple code sections (`kind = 1`) are allowed.
-2. Total number of code sections MUST NOT exceed 1024.
+1. The [EIP-3540](./eip-3540.md) code section header is redefined to be an array of code section sizes terminated with a `0x00` byte. E.g. `code_section_header := 1, (code_section_size)+, 0`.
```
axic, line 32: Is this a compression method so that there will be still a single code section header? Why doesn't this rule apply to the data section then? And if so, why not just introduce a simple RLE compression 😅

lightclient: I believe there can only be one data section so this rule wouldn't be able to apply? This is less about compression and more about having a intuitive header format. Practically, the code sections must be sandwiched between the types and data sections. It feels quite natural to just have the code header dynamically list all the sizes. In fact, requiring the type section might not even be such a bad idea. That would simplify the decoding a bit since you can always assume the size of the type section (no longer optional) is `2n` the number of code sections.

EIPS/eip-4750.md
```diff
@@ -29,12 +29,12 @@ Furthermore it aims to improve analysis opportunities by encoding the number of
...
+2. Total number of code sections MUST NOT exceed 1024 and they must all be contiguous.
 3. All code sections MUST precede a data section, if data section is present.
 4. New section with `kind = 3` is introduced called the *type section*.
 5. Exactly one type section MAY be present.
-6. The type section, if present, MUST directly precede all code sections.
+6. The type section, if present, MUST directly precede the code sections.
```
xinbenlv, line 37:
```suggestion
6. The type section, if present, MUST immediately precede the code sections.
```

EIPS/eip-4750.md
```diff
@@ -29,12 +29,12 @@ Furthermore it aims to improve analysis opportunities by encoding the number of
 
 ### EOF container changes
 
-1. The requirement of [EIP-3540](./eip-3540.md) "Exactly one code section MUST be present." is relaxed to "At least one code section MUST be present.", i.e. multiple code sections (`kind = 1`) are allowed.
-2. Total number of code sections MUST NOT exceed 1024.
+1. The [EIP-3540](./eip-3540.md) code section header is redefined to be an array of code section sizes terminated with a `0x00` byte. E.g. `code_section_header := 1, (code_section_size)+, 0`.
+2. Total number of code sections MUST NOT exceed 1024 and they must all be contiguous.
```
xinbenlv, line 33:
```suggestion
2. Total number of code sections MUST NOT exceed 1024 and they MUST all be contiguous.
```

lightclient:
FYI: I wrote up a comparison of a few alternatives to this. https://notes.ethereum.org/@lightclient/rJ0nTvKIj

EIPS/eip-4750.md
gumb0: > In fact, requiring the type section might not even be such a bad idea. That would simplify the decoding a bit since you can always assume the size of the type section (no longer optional) is 2n the number of code sections.

It can stay optional, because omitted type section implies there's just 1 code section.

(but I think it still makes sense to make it required for explicitness and simplifying the parser)

lightclient: Right, so for that reason I think it should be required.

lightclient:
closed as #6156 was merged

PR #5793, opened 2022-10-18
Add EIP-5793: eth/68: Add transaction type to tx announcement

MariusVanDerWijden:
EIP-5793 adds the transaction type and size to the transaction announcement

PR #5930, opened 2022-11-14
Update EIP-4844: Disable Blob Transaction Broadcast

adietrichs:
Specifies that blob transactions should not be broadcast to peers, but only announced (so that they can then be manually retrieved).

Adds a reference to https://github.com/ethereum/EIPs/pull/5793 which will give peers more fine-grained control by making it possible to identify blob transactions (and their size) from announcement messages.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -8,7 +8,7 @@ status: Draft
 type: Standards Track
 category: Core
 created: 2022-02-25
-requires: 1559, 2718, 2930, 4895, 5793
+requires: 1559, 2718, 2930, 4895
 ---
 
 ## Abstract
@@ -322,6 +322,8 @@ The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sen
 
 ### Networking
 
+Blob transactions are not automatically broadcast to peers. Instead, they are only announced using `NewPooledTransactionHashes` messages, and can then be manually requested via `GetPooledTransactions`.
+
 Transactions are presented as `TransactionType || TransactionNetworkPayload` on the execution layer network,
 the payload is a SSZ encoded container:
 
@@ -468,7 +470,9 @@ instead, they go into the `BeaconBlockBody`. This means that there is now a part
 Blob transactions have a large data size at the mempool layer, which poses a mempool DoS risk,
 though not an unprecedented one as this also applies to transactions with large amounts of calldata.
 
-[EIP-5793](./eip-5793.md) addresses this by adding the transaction type and size to eth/68 announcement messages. Nodes are thus expected to no longer broadcast large transactions - including blob transactions - by default. This allows nodes to load balance or throttle peers based on past behavior.
+By only broadcasting announcements for blob transactions, nodes have control over which and how many transactions to receive,
+allowing them to throttle throughput to an acceptable level.
+[EIP-5793](./eip-5793.md) will give further fine-grained control to nodes by extending the `NewPooledTransactionHashes` announcement messages to include the transaction type and size.
 
 In addition, we recommend including a 1.1x data gasprice bump requirement to the mempool transaction replacement rules.
 
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -322,7 +322,8 @@ The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sen
 
 ### Networking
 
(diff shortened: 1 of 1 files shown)


protolambda (approved):

EIPS/eip-4844.md
```diff
@@ -322,6 +322,8 @@ The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sen
 
 ### Networking
 
+Blob transactions are not automatically broadcast to peers. Instead, they are only announced using `NewPooledTransactionHashes` messages, and can then be manually requested via `GetPooledTransactions`.
```
xinbenlv, line 325: I wonder if this shall be moved to a separate EIP addressing Network layer only

xinbenlv, line 325:
```suggestion
Nodes MAY not automatically broadcast Blob transactions to peers. Instead, they are only announced using `NewPooledTransactionHashes` messages, and can then be manually requested via `GetPooledTransactions`.
```

[Editorial] I wouldn't be surprised if some nodes want to choose to broadcast such Blob transactions. Hence, I love to prompt author to clarify if the spec is  mandating(`MUST`) / suggesting(`SHOULD`) /or allowing (`MAY`) this implementation behavior

EIPS/eip-4844.md
```diff
@@ -468,10 +470,11 @@ instead, they go into the `BeaconBlockBody`. This means that there is now a part
 Blob transactions have a large data size at the mempool layer, which poses a mempool DoS risk,
 though not an unprecedented one as this also applies to transactions with large amounts of calldata.
 
-We recommend two solutions:
+By only broadcasting announcements for blob transactions, nodes have control over which and how many transactions to receive,
```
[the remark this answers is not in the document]
xinbenlv: [Technical] But they could also choose to filter and ignore such broadcasting announcements, right?

EIPS/eip-4844.md
```diff
@@ -468,10 +470,11 @@ instead, they go into the `BeaconBlockBody`. This means that there is now a part
...
+allowing them to throttle throughput to an acceptable level.
+[EIP-5793](./eip-5793.md) will give further fine-grained control to nodes by extending the `NewPooledTransactionHashes` announcement messages to include the transaction type and size.
 
-- include a 1.1x (or potentially higher) data gasprice bump requirement to the mempool replacement rules
-- modify the Ethereum Wire Protocol to stop automatically broadcasting large transactions
+In addition, we recommend including a 1.1x data gasprice bump requirement to the mempool transaction replacement rules.
```
xinbenlv, line 477:
```suggestion
In addition, we recommend including a 1.1x data gasprice bump requirement to the mempool transaction replacement rules.
```

Is the "recommendation" added here as a specification? Could you clarify it's a comment or it's a specification?

EIPS/eip-4844.md
adietrichs: This is a good point. The intention here is to explicitly mandate this behavior (so `MUST`), because that allows nodes to drop misbehaving peers.

adietrichs: As this is required behavior for all nodes implementing 4844, it seems appropriate to keep it within the EIP.

xinbenlv: Cool. thanks for clarifying. In that case I'd suggest update to 

```suggestion
Nodes MUST not automatically broadcast Blob transactions to peers. Instead, they are only announced using `NewPooledTransactionHashes` messages, and can then be manually requested via `GetPooledTransactions`.

Any nodes received Blob transactions MUST deemed the sending nodes "misbehaving" and MUST drop them as peer.
```

xinbenlv: While I think there is no EIP editorial policy or guideline strictly prohibiting this, as far as I know, hence I think it's ok to stay same EIP. 

However, proper scoping could help improve readability and technical soundness for an EIP. For large EIPs like 4844, consider discuss among your coauthor to see if it make more sense to put network part of spec into the same core EIP or a separate one.

An example of split scoping: EIP-1559 depends on EIP-2718, EIP-2930 which depends on EIP-2929(Network).

protolambda (approved):

MariusVanDerWijden (approved):
LGTM

protolambda:
@MicahZoltu can you please help merge this? Auto-merge was disabled, but we have reviewed and like to merge it now.

lightclient (approved):

PR #5934, opened 2022-11-15
Update EIP-5298: Move to Review

xinbenlv:


xinbenlv:
WIP

Pandapip1 (approved):
I *think* this is good. I'm a little skeptical that it works for any ENS address. I would say that defining a function to get the balance of an ENS name would be an improvement.

I'd also like this to support EIP-20 and EIP-1155, although this is less trivial to implement and I understand if you would like to limit the scope of this EIP to true NFTs.

PR #5979, opened 2022-11-15
EIP-3651: Last Call

wjmelements:
#### Changes
Status from Review to Last Call
Set last-call-deadline


lightclient (commented):
Hi @wjmelements, normally we don't move core EIPs to last call until their mainnet launch is imminent. Just on the off chance a minor change is needed to be made. Once the EIP is final it cannot be changed so it would be cumbersome/confusing to create an entirely new EIP.

wjmelements:
Thanks @lightclient.

PR #5985, opened 2022-11-16
CI: Grab latest `eipw`, which won't block stagnant bot

SamWilsn:
The latest version of `eipw` will report warnings instead of errors for EIPs with a Stagnant status. This should let the stagnant bot work properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Checkout EIP Repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
 
-      - uses: ethereum/eipw-action@bbf342b62cb8046439c3508f1e98e23a4a9cd27a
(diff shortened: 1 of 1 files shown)


SamWilsn:
> What about htmlproofer and codespell? Codespell was not enforced for a long time, so there are at least 6 EIPs that have codespell errors.

Not sure what to do about those. I just noticed a wave of PRs (eg. #5973) that are only blocked on `eipw` results.

> Now that I think about it, there's a new github feature to add custom roles. Bypassing branch protection is one of the options: https://github.blog/changelog/2022-08-18-bypass-branch-protections-with-a-new-permission/. Could we get this set up?

You'd like to give that permission to editors? I'll look into it.

Pandapip1:
No, I meant give that permission to @eth-bot, which will then be able to bypass the checks

Pandapip1 (approved):
Otherwise, LGTM

Pandapip1:
Actually, I found https://github.blog/changelog/2021-11-19-allow-bypassing-required-pull-requests/. Can we add @eth-bot to this list?

PR #3009, opened 2020-09-30
EIP-3009: transferWithAuthorization - Gas-Abstracted ERC20 transactions

petejkim:
An EIP for `transferWithAuthorization`, as used by the USD Coin (USDC) smart contract, alongside [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612).

[Discussions](https://github.com/ethereum/EIPs/issues/3010)

PR #5987, opened 2022-11-17
Add EIP-5987: Permissionless Transfer With Authorization

grantmike:
An EIP for permissionlessTransferWithAuthorization, extending [EIP-3009](https://github.com/ethereum/EIPs/pull/3009)

### DIFF
--- a/EIPS/eip-5792.md
+++ b/EIPS/eip-5792.md
@@ -387,5 +387,4 @@ it will take a bundle to be included. Dapps must encode deadlines in the smart c
 Dapp developers MUST NOT assume that all calls are sent in a single transaction.
 
 ## Copyright
-
 Copyright and related rights waived via [CC0](../LICENSE.md).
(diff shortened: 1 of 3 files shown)


Pandapip1 (commented):
Please fix the walidator errors

Pandapip1:
FYI - when creating new EIPs, it's highly recommended to use the template, as it follows the most up-to-date sections and their content (minus the external links, which are gradually being re-added).

Pandapip1 (commented):
Here's my partial review

juli:
@grantmike As we discussed before, economic incentives make relayers attacking relayers a more likely scenario than relayers abusing users. Part of a design that protects relayers and incentivizes participation in the system is using unique nonces for each relayer. The user's wallet selects a relayer with a timeout; if the relayer fails, the user tx is delayed, and the relayer's reputation is affected, but the risk is reduced for both.I can share more details by email.

PR #5993, opened 2022-11-17
EIP-5450: Revision 2

chfast:


EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
-Moreover, the `max_stack_height` computed during validation must be stored alongside the code. This can be in the EOF itself or in implementation-defined format.
+Moreover, the `max_stack_height` computed during validation must be stored alongside the code, likely in the type section header.
 
 ### Unreachable code
 
-The current validation algorithm ignores _unreachable_ instructions. The algorithm can be extended to reject any code having any unreachable instructions but additional instructions traversal is needed (or more efficient algorithm must be developed).
+The operand stack validation algorithm ignores *unreachable* instructions. The algorithm can beextended to reject any code having any unreachable instructions but additional instructions traversal is needed to search for unvisited instructions (or more efficient algorithm must be developed).
```
gcolvin, line 107:
```suggestion
The operand stack validation algorithm ignores *unreachable* instructions. The algorithm can be extended to reject any code having any unreachable instructions but additional instructions traversal is needed to search for unvisited instructions (or more efficient algorithm must be developed).
```

gcolvin (changes requested):

EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
 
-Stack overflow check, on the other hand, is still required at run-time, because function execution can start at arbitrary (i.e. known only at run-time) stack height at `CALLF` instruction of a caller (i.e. each execution can be in arbitrary inner call frame). Verification algorithm examines only stack height changes relative to starting stack height of the function.
 
 ## Rationale
 
 ### Stack overflow check only in CALLF
 
 In the current proposal stack overflow checks are unchanged (i.e. are done for every instruction). However, we can provide more efficient variant where stack overflow check is performed only in `CALLF` instruction and uses called function's `max_stack_height` information for this. This decreases flexibility of an EVM program because `max_stack_height` corresponds to the worst-case control-flow path in the function.
```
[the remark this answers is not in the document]
gumb0: It does, because some calls could succeed without this check if they take not-worst paths, but will always be rejected with this check.

EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
 
-*Remark:* We rely on the notions of *data stack* and *type section* as defined by [EIP-4750](./eip-4750.md).
+*Remark:* We rely on the notions of *operand stack* (FIXME) and *type section* as defined by [EIP-4750](./eip-4750.md).
+
+Each code section is validated independently.
+
+#### Instructions validation
+
+In the first validation phase instructions are inspected independently to check if their opcodes and immediate data are valid.
+This effectively replaced code validation in EIP-3670, EIP-4200 and EIP-4750.
```
gumb0, line 36:
```suggestion
This effectively replaces code validation in EIP-3670, EIP-4200 and EIP-4750.
```

EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
 ## Specification
 
 ### Code validation
 
-*Remark:* We rely on the notions of *data stack* and *type section* as defined by [EIP-4750](./eip-4750.md).
+*Remark:* We rely on the notions of *operand stack* (FIXME) and *type section* as defined by [EIP-4750](./eip-4750.md).
+
+Each code section is validated independently.
+
+#### Instructions validation
```
gumb0, line 33: Not sure why we have to repeat this validation in this EIP, it is alreadt specified in others.

EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
+
+In the first validation phase instructions are inspected independently to check if their opcodes and immediate data are valid.
+This effectively replaced code validation in EIP-3670, EIP-4200 and EIP-4750.
+
+A list of code length is created to collect purpose of each byte in the code: *data*, *jump target* or *other*.
```
gumb0, line 38: of length equal to code section length
```suggestion
A list of length equal to code section length is created to collect purpose of each byte in the code: *data*, *jump target* or *other*.
```

EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
+   2. Check if *target function index* does not exceed the number of entries in the *type section*.
+4. If instruction is static jump (`RJUMP`, `RJUMPI`)
+   1. Parse immediate bytes to get *target relative offset*.
+   2. Compute *target offset*.
+   3. Check if *target offset* is within the code.
+   4. Check if *target offset* is not marked as *data*.
+   5. Mark *target offset* as *jump target*.
 
-*Data stack height* here refers to the number of stack values accessible by this function, i.e. it does not take into account values of caller functions' frames (but does include this function's inputs).  Note that validation procedure does not require actual data stack implementation, but only to keep track of its height. Current height value starts at `types[code_section_index].inputs` (number of inputs of this function) at function entry and is updated at each instruction.
+If any of the checks fails the validation fails.
```
gumb0, line 58:
```suggestion
If any of the checks fails, the validation fails.
```
or
```suggestion
If any of the checks fails, then validation fails.
```

EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
 
-To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. See below for reference implementation.
+In the second validation phase data-flow (TODO: or control-flow) analysis is performed on the code.
+For each reachable instruction in the code the operand stack height is recorded.
+The first instruction has the recorded stack height equal to the number of inputs to the function matching the code.
 
-The complexity of this traversal is linear in the number of instructions, because each code path is examined only once, and property 1 guarantees no loops in the validation.
+The FIFO queue *worklist* is used for tracking "to be inspected" instructions. Initially, it contains the first instruction.
+
+While *worklist* is not empty, pop an instruction and:
```
gumb0, line 70: pop => dequeue ?

EIPS/eip-5450.md
```diff
@@ -21,59 +21,102 @@ Currently existing EVM implementations perform a number of validity checks for e
...
+2. Determine the list of successor instructions than can follow the current instructions:
+   1. The next instruction for all instructions other than terminating instructions and unconditional jump.
+   2. The target of a conditional or unconditional jump.
+3. For each successor instruction:
+   1. Check if the instruction is present in the code (i.e. execution must not "fall off" the code).
+   2. If the instruction does not have stack height recorded (visited for the first time):
+      1. Record the instruction stack height as the value computed in 1.2.
+      2. Add the instruction to the *worklist*.
+   3. Otherwise, check if the recorded stack height equals the value computed in 1.2.
+4. Determine the function maximum operand stack height:
```
gumb0, line 85: Maybe better to group this with other stack manipulations of point 1? (make it subpoint of 1?)

EIPS/eip-5450.md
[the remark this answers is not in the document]
gumb0: Right, it's either additional pass or additional O(len(n)) memory

EIPS/eip-5450.md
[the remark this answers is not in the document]
gcolvin: My understanding is that at validation time every path through a function is constrained to have the same stack height, so max_stack_height is just a specification of what that height should be.  I might misunderstand.

EIPS/eip-5450.md
chfast: I wanted a holistic overview and show that 2 passes are needed.

In EIP-4200 it is specified what is valid but not how to validate. 

> verify that each RJUMP/RJUMPI has a relative_offset pointing to an instruction. This means it cannot point to an immediate data of PUSHn/RJUMP/RJUMPI. It cannot point outside of code bounds.

It suggests 2 passes. Here I show how to do it with single pass. This is to visualize the complexity of the algorithm.

EIPS/eip-5450.md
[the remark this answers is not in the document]
chfast: The counterargument is mostly: is it worth the complexity given that the benefits for interpreters are minimal.

EIPS/eip-5450.md
gumb0: I would say it's good for a separate "Implementation recommendations" document, but maybe not for the spec. It's confusing to repeat the same things in several EIPs, they look conflicting.

I don't think EIP-4200 suggests any specific algorithm.

EIPS/eip-5450.md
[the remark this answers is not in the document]
gumb0: I'm generally in favor of rejecting everything that we can prove statically doesn't make sense. The fewer weird things are allowed to deploy, the less is potential for abuse and corner cases at runtime.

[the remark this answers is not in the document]
gcolvin: I agree wholeheartedly.  I think the only current use for unreachable code is to stash data, for which we now have data sections.

gumb0:
Rebased

EIPS/eip-5450.md
gumb0: "Unified spec" document can be a good place for this.

chfast: I agree. Here you can import something from eip3670.

EIPS/eip-5450.md
```diff
@@ -13,67 +13,127 @@ requires: 3540, 3670, 4200, 4750
...
+
+This EIP minimizes the number of such checks required at run-time
+by verifying that no exceptional conditions can happen
+and preventing the execution and deployment of any invalid code.
+
+The operand stack validation provides several benefits:
+
+- removes run-time stack underflow check for all instructions,
+- removes run-time stack overflow check for all instruction except `CALLF`,
+- ensures that an execution terminates with one of the terminating instructions,
```
lightclient, line 31: Is this still relevant?

EIPS/eip-5450.md
```diff
@@ -13,67 +13,127 @@ requires: 3540, 3670, 4200, 4750
...
+- prevents the deployment of code with unreachable instructions, which discourages the use of code sections for data storage (TODO),
+- function type information is only needed for validation.
```
lightclient, line 33: This means only the input, output info? This might be a little confusing if we also add `maxStackHeight` to the type info.

EIPS/eip-5450.md
```diff
@@ -13,67 +13,127 @@ requires: 3540, 3670, 4200, 4750
...
 
-Stack overflow check, on the other hand, is still required at run-time, because function execution can start at arbitrary (i.e. known only at run-time) stack height at `CALLF` instruction of a caller (i.e. each execution can be in arbitrary inner call frame). Verification algorithm examines only stack height changes relative to starting stack height of the function.
 
 ## Rationale
 
 ### Stack overflow check only in CALLF
 
-In the current proposal stack overflow checks are unchanged (i.e. are done for every instruction). However, we can provide more efficient variant where stack overflow check is performed only in `CALLF` instruction and uses called function's `max_stack_height` information for this. This decreases flexibility of an EVM program because `max_stack_height` corresponds to the worst-case control-flow path in the function.
-Moreover, the `max_stack_height` computed during validation must be stored alongside the code. This can be in the EOF itself or in implementation-defined format.
+In the current proposal stack overflow checks are unchanged (i.e. are done for every instruction).
```
lightclient, line 118: I think we decided to incorporate this into the EIP?

EIPS/eip-5450.md
gumb0: I think actually output also will be needed for "stack cleaning" version of RETF

EIPS/eip-5450.md
gumb0: I believe it is, but this doesn't meen final byte is necessary a terminating instruction (can be unconditional jump too)

axic:
@chfast this probably needs to be updated after #6101 was merged

EIPS/eip-5450.md
chfast: This has been removed because no longer true (e.g. `CALLF` requires `max_stack_height`).

EIPS/eip-5450.md
chfast: Yes, this is now included.

EIPS/eip-5450.md
chfast: This is now removed.

EIPS/eip-5450.md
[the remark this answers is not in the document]
chfast: Unreachable instructions are now disallowed. This enables single-pass validation—mentioned in the rationale but not used in the spec.

EIPS/eip-5450.md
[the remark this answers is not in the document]
chfast: Outdated. Now stack overflow is disabled and only `CALLF` checks `max_stack_height`.

EIPS/eip-5450.md
```diff
@@ -13,51 +13,100 @@ requires: 3540, 3670, 4200, 4750
...
+2. Determine the list of successor instructions than can follow the current instructions:
+   1. The next instruction for all instructions other than terminating instructions and unconditional jump.
+   2. All targets of a conditional or unconditional jump.
+3. For each successor instruction:
+   1. **Check** if the instruction is present in the code (i.e. execution must not "fall off" the code).
+   2. If the instruction does not have stack height recorded (visited for the first time):
+      1. Record the instruction stack height as the value computed in 1.2.
+      2. Add the instruction to the *worklist*.
+   3. Otherwise, **check** if the recorded stack height equals the value computed in 1.2.
+4. **Check** if all instruction where reached by the analysis.
```
gumb0, line 78: This looks like it's part if `While *worklist* is not empty` loop. Maybe make it a separate section, like 
```

When *worklist* is empty:

4. **Check** if all instruction where reached by the analysis.
5. ...
```

chfast: Done.

EIPS/eip-5450.md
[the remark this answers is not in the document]
gcolvin: And I am happy about that.  Thanks.

EIPS/eip-5450.md
```diff
@@ -13,51 +13,103 @@ requires: 3540, 3670, 4200, 4750
...
 
-The complexity of this traversal is linear in the number of instructions, because each code path is examined only once, and property 1 guarantees no loops in the validation.
+1. Determine the effect the instruction has on the operand stack:
+   1. **Check** if the recorded stack height satisfies the instruction requirements. Specifically:
+      - for `CALLF` instruction the required stack height is the number of inputs of the called function according to its type defined in the type section.
+      - for `RETF` instruction the recorded stack height must be exactly the number of outputs of the function matching the code.
+   2. Compute new stack height after the instruction execution.
+2. Determine the list of successor instructions than can follow the current instructions:
+   1. The next instruction for all instructions other than terminating instructions and unconditional jump.
+   2. All targets of a conditional or unconditional jump.
```
DemuirGos, line 71: I think we should Add another case clarifying what to do in the case of terminating opcode, and the set of opcodes considered terminating opcode.

EIPS/eip-5450.md
```diff
@@ -13,51 +13,103 @@ requires: 3540, 3670, 4200, 4750
...
-7. No unreachable instructions exist in the code section.
+The FIFO queue *worklist* is used for tracking "to be inspected" instructions. Initially, it contains the first instruction.
 
-To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. After completing this, verify each instruction was visited at least once. Fail if any instructions were not visited as this invalidates 7).
+While *worklist* is not empty, dequeue an instruction and:
 
-The complexity of this traversal is linear in the number of instructions, because each code path is examined only once, and property 1 guarantees no loops in the validation.
+1. Determine the effect the instruction has on the operand stack:
+   1. **Check** if the recorded stack height satisfies the instruction requirements. Specifically:
+      - for `CALLF` instruction the required stack height is the number of inputs of the called function according to its type defined in the type section.
```
DemuirGos, line 66: I guess the wording of the rule ``1.1.CALLF`` should reflect that, instead of saying at ``CALLF`` sight the  ``..required stack height is the number of the inputs of the called function..``, I think it should say ``..the required stack height is at least the number of inputs of the called function..``

smartprogrammer93:
```suggestion
      - for `CALLF` instruction the required stack height is at least the number of inputs of the called function according to its type defined in the type section.
```

EIPS/eip-5450.md
chfast: Added.

EIPS/eip-5450.md
chfast: Changed by avoiding "required stack height" term (I use it internally and it slipped into the spec).

jochem-brouwer (dismissed):
I especially checked the operand stack validation and this looks good to me :smile:

EIPS/eip-5450.md
```diff
@@ -13,51 +13,108 @@ requires: 3540, 3670, 4200, 4750
...
+3. For each successor instruction:
+   1. **Check** if the instruction is present in the code (i.e. execution must not "fall off" the code).
+   2. If the instruction does not have stack height recorded (visited for the first time):
+      1. Record the instruction stack height as the value computed in 1.2.
+      2. Add the instruction to the *worklist*.
+   3. Otherwise, **check** if the recorded stack height equals the value computed in 1.2.
+
+When *worklist* is empty:
+
+1. **Check** if all instruction where reached by the analysis.
```
jochem-brouwer, line 86:
```suggestion
1. **Check** if all instruction were reached by the analysis.
```

jochem-brouwer:
(Correct me if I'm wrong, of course :smile: )

I think there is some part missing here: if there is a `CALLF` then the new stack height will be the current stack height - inputs of the function + number of outputs for the function (function = function called by CALLF, and inputs are read from the type section?)?

chfast:
> (Correct me if I'm wrong, of course smile )
> 
> I think there is some part missing here: if there is a `CALLF` then the new stack height will be the current stack height - inputs of the function + number of outputs for the function (function = function called by CALLF, and inputs are read from the type section?)?

Yes, this is correct. I can add this specific case to the spec.

jochem-brouwer:
I have an untested reference implementation here: https://github.com/ethereumjs/ethereumjs-monorepo/blob/edc7545c2828a11d18609519f62aa176a6e93429/packages/evm/src/eof/eof.ts#L177, this essentially implements all the rules in the EIP (and also in the way it is described)

EIPS/eip-5450.md
```diff
@@ -13,51 +13,108 @@ requires: 3540, 3670, 4200, 4750
...
+*Operand stack height* here refers to the number of stack values accessible by this function, i.e. it does not take into account values of caller functions' frames (but does include this function's inputs). Note that validation procedure does not require actual data stack implementation, but only to keep track of its height.
 
-*Data stack height* here refers to the number of stack values accessible by this function, i.e. it does not take into account values of caller functions' frames (but does include this function's inputs). Note that validation procedure does not require actual data stack implementation, but only to keep track of its height. Current height value starts at `type[code_section_index].inputs` (number of inputs of this function) at function entry and is updated at each instruction.
+*Terminating instructions* refers to the instructions either:
 
-At the same time the following properties are being verified:
+- ending function execution: `RETF`, or
+- ending whole EVM execution: `STOP`, `RETURN`, `REVERT`, `INVALID`.
```
DemuirGos, line 60: What is the expected stack height for these Opcode ? is it 0 ?

chfast: There are no additional expectations besides the regular stack checks: e.g. for `RETURN` the stack height must be at least 2.

EIPS/eip-5450.md
```diff
@@ -13,51 +13,108 @@ requires: 3540, 3670, 4200, 4750
...
+
+- adds constraints to the code structure (similar to JVM, CPython bytecode, WebAssembly and others); however, these constraints can be lifted in a backward-compatible manner if they are shown to be user-unfriendly,
+- adds second validation pass; however, validation's computational and space complexity remains linear.
 
 ## Specification
 
 ### Code validation
 
-*Remark:* We rely on the notions of *data stack* and *type section* as defined by [EIP-4750](./eip-4750.md).
+*Remark:* We rely on the notions of *operand stack* (FIXME) and *type section* as defined by [EIP-4750](./eip-4750.md).
```
lightclient, line 43: still a FIXME here

lightclient (approved):
One nit about the FIXME, otherwise LGTM.

EIPS/eip-5450.md
chfast: This should stay: we want to rename "data stack" to "operand stack" across all EIPs.

PR #6002, opened 2022-11-18
EIP4844: Update crypto API (simplify precompile / sync with consensus-specs)

asn-d6:
This PR does two things (split into corresponding commits):
- Simplifies the implementation of the point evaluation precompile by using the bytes-based `verify_kzg_proof()` from https://github.com/ethereum/consensus-specs/pull/3097 . This way the client developers don't need to understand what field elements are, and can just pass bytes to the KZG library.
- Syncs `validate_blob_transaction_wrapper()` to the latest high-level crypto API of the consensus specs introduced in https://github.com/ethereum/consensus-specs/pull/3038

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -257,22 +257,23 @@ The precompile costs `POINT_EVALUATION_PRECOMPILE_GAS` and executes the followin
 
 ```python
 def point_evaluation_precompile(input: Bytes) -> Bytes:
-    # Verify P(z) = a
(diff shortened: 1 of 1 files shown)


kevaundray (approved):

protolambda (approved):

protolambda:
@MicahZoltu George (asn-d6) disabled auto-merge to get my review after changes in, but now the PR is stuck. Can you help us merge this?

PR #5849, opened 2022-10-28
Added EOF prefix to 4530 and 4200, for ease of searching for EOF-related EIPs.

bobsummerwill:
Minor formatting tweak:
Added EOF prefix to 4530 and 4200, for ease of searching for EOF-related EIPs.

PR #6007, opened 2022-11-19
Add "EOF - " prefix to 3540, 4200

gumb0:
Replacement for #5849


gumb0:
Apparently colons are not allowed in titles :disappointed:

chfast (approved):

PR #6012, opened 2022-11-20
Update EIP-721: Update link

fulldecent:

 - The build passes.

### DIFF
--- a/EIPS/eip-721.md
+++ b/EIPS/eip-721.md
@@ -440,7 +440,7 @@ XXXXERC721, by William Entriken -- a scalable example implementation
 1. Curio Cards. https://mycuriocards.com
 1. Rare Pepe. https://rarepepewallet.com
 1. Auctionhouse Asset Interface. https://github.com/dob/auctionhouse/blob/master/contracts/Asset.sol
-1. OpenZeppelin SafeERC20.sol Implementation [since renamed to just ERC20]. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol
+1. OpenZeppelin SafeERC20.sol Implementation [since renamed to just ERC20]. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5daaf60d11ee2075260d0f3adfb22b1c536db983/contracts/token/ERC20/SafeERC20.sol
 
 ## Copyright
 
--- a/EIPS/eip-721.md
+++ b/EIPS/eip-721.md
@@ -440,7 +440,7 @@ XXXXERC721, by William Entriken -- a scalable example implementation
 1. Curio Cards. https://mycuriocards.com
 1. Rare Pepe. https://rarepepewallet.com
 1. Auctionhouse Asset Interface. https://github.com/dob/auctionhouse/blob/master/contracts/Asset.sol
-1. OpenZeppelin SafeERC20.sol Implementation. https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/token/ERC20/SafeERC20.sol
+1. OpenZeppelin SafeERC20.sol Implementation [since renamed to just ERC20]. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol
 
 ## Copyright
 


EIPS/eip-721.md
```diff
@@ -440,7 +440,7 @@ XXXXERC721, by William Entriken -- a scalable example implementation
 1. Curio Cards. https://mycuriocards.com
 1. Rare Pepe. https://rarepepewallet.com
 1. Auctionhouse Asset Interface. https://github.com/dob/auctionhouse/blob/master/contracts/Asset.sol
-1. OpenZeppelin SafeERC20.sol Implementation. https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/token/ERC20/SafeERC20.sol
```
xinbenlv, line 443: I wonder if we could, instead of link to new link, consider find the permlink to old file and link to that permlink?

@SamWilsn @Pandapip1 for thoughts

fulldecent: That works for me, have a specific version to propose?

xinbenlv: > That works for me, have a specific version to propose?

You are the author, so I'd probably the best version would be the intended version you linkes to when you were authoring this EIP.

Pandapip1: How about we just delete the link altogether? It doesn't add much value.

fulldecent:
Sound good @xinbenlv I'm a fan of permalinks. Just updated it at 85086421c0e64ae87772d50b9f73ca8c93c1ec35 to use the version available at time of publication.

EIPS/eip-721.md
xinbenlv: As a peer author reviewing this EIP, I think such links serve a lot of value collecting landscape of adoptions and make techinical arguments more convincing.

Pandapip1 (approved):
+0 to updating the link, +1 if it's removed altogether.

xinbenlv:
> +0 to updating the link, +1 if it's removed altogether.

I am afraid I can't agree with this argument, @Pandapip1 . IMHO, such links were part of the technical argument demonstrating the adoption landscape back them and made the proposal argument more convincing. I think they play an important role in making ERC721 an model EIP.

xinbenlv (approved):

Pandapip1:
EIP-721 is not a model EIP, as:

* Policies have changed since then
* The template has changed significantly

I agree that the content of EIP-721 is excellent. As such, it would indeed make sense to update EIP-20, EIP-721, and EIP-1155 to reflect current EIP practices. However, current EIP practices include "not including external links to unapproved domains." Respectfully, arbitrary GitHub links and OpenZeppelin link are not approved domains, nor should they be.

fulldecent:
Comments here are discussing two things:

1. Whether EIPs, especially new ones, should link to GitHub and other domains
2. Updating ERC-721

I have updated many EIPs with and without the consent of the original authors. There are limits to the changes that we have pushed through in these situations. Because we are not planning to rewrite all existing published EIPs that means that there is a at least _some_ difference on what is publishable for existing versus new EIPs.

---

I do not have any comments right now about whether linking to GitHub should be acceptable or not.

But I recognize one complaint is that most links to GitHub are rugpullable—a page at some URL can be changed and replaced with some entirely different content. There may be other complaints as well.

This PR changes an existing GitHub link on a named branch to link on a specific hash which cannot be modified. This corrects an error I originally made while publishing (sorry!), always use permalinks.

Whereas this PR addresses one material concern about an existing link to GitHub and it does not introduce any new link to GitHub, I request that this be merged as-is. Also the debate about whether linking to GitHub is good or bad should be closed without prejudice. Such broad discussion can be much better held elsewhere and this PR does not have any net impact on linking to GitHub.

Pandapip1:
I agree with @fulldescent that PRs that fix broken links are good. That's why I've approved this PR, and have added it to the manual merge queue.

I also think that PRs that *remove* links are also good.

SamWilsn:
We've discussed this pull request on EIPIP today. Since this link is not part of the actual specification (aka non-normative), and it breaking doesn't affect the understanding of this EIP, we are going to close this pull request for the time being.

That said, @gcolvin would like to define a more formal process for correcting errata, and we should revisit this change at that time.

PR #6029, opened 2022-11-22
Update EIP-4895: fix small typo

0xGabi:


ralexstokes (approved):
hmm not sure where that came from!

PR #6014, opened 2022-11-21
EIP-4938: Move to back to draft

MariusVanDerWijden:
This PR was falsely moved to stagnant, move to last call instead

PR #6032, opened 2022-11-22
EIP-4938: change EIP to Review status (eth/67)

protolambda:
EIP-4938 (eth/67) is blocking Review status of EIP-5793 (eth/68) and should not be Stagnant. I'm updating this to Review, but the authors may want to change it to Last Call / Final depending on the adoption of the eth/67 protocol.


lightclient:
Duplicate of #6014

PR #6043, opened 2022-11-25
EIP-5007: Change uint64 to int64

emojidao:
EIP-5007: Change uint64 to int64

The max value of `int64` is 9,223,372,036,854,775,807. As a timestamp, 9,223,372,036,854,775,807 is about year 292,471,210,648. `uint256` is too big for C, C++, Java, Go, etc, and `int64` is natively supported by mainstream programming languages.

### DIFF
--- a/EIPS/eip-5007.md
+++ b/EIPS/eip-5007.md
@@ -130,6 +130,7 @@ truffle test
 ```
  
 ## Reference Implementation
+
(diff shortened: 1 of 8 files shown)


0xanders (approved):

0xanders (approved):

Skylerspencer:
So did I do something wrong??

PR #6044, opened 2022-11-25
Update EIP-4750: Disallow dynamic jumps

gumb0:


EIPS/eip-4750.md
```diff
@@ -116,17 +116,19 @@ In addition to container format validation rules above, we extend code section v
     1. Code section is invalid in case offset points to a position outside of section bounds.
     2. Code section is invalid in case offset points to one of two bytes directly following `CALLF` instruction.
 
+### Deprecated instructions
+
+Dynamic jump instructions `JUMP` (`0x56`) and `JUMPI` (`0x57`) are depreceated and invalid and their opcodes do not have any instruction assigned.
```
chfast, line 121:
```suggestion
Dynamic jump instructions `JUMP` (`0x56`) and `JUMPI` (`0x57`) are depreceated and invalid and their opcodes are undefined.
```

Not sure this is better, but I like "undefined opcode" and/or "undefined instruction" term.

chfast (commented):
Looks good to me, but:

1. We need rationale that jumpdest analysis is not needed any more.
2. `NOP` change must be discussed.

EIPS/eip-4750.md
axic: Undefined sounds fine.

EIPS/eip-4750.md
```diff
@@ -116,17 +116,19 @@ In addition to container format validation rules above, we extend code section v
     1. Code section is invalid in case offset points to a position outside of section bounds.
     2. Code section is invalid in case offset points to one of two bytes directly following `CALLF` instruction.
 
+### Deprecated instructions
```
axic, line 119: It uses "deprecated" in abstract, "disallowed" in motivation, "deprecated" here. We should choose one, and "disallowed" sounds closer to what it is.

axic:
@gumb0 rebased to resolve conflicts.

EIPS/eip-4750.md
gumb0: Changed to "disallowed" everywhere

PR #6054, opened 2022-11-28
Small updates to EIP-663

axic:


EIPS/eip-663.md
```diff
@@ -29,8 +29,8 @@ Introducing `SWAPN` and `DUPN` will provide an option to compilers to simplify a
 
 We introduce two new instructions:
 
- 1. `DUPN` (`0xb0`)
- 2. `SWAPN` (`0xb1`)
+ 1. `DUPN` (`0xb3`)
```
xinbenlv, line 32: I am still curious, if these instruction numbers are better provided as a placeholder rather than a given number. Given we only have 2 bytes. With EIP drafts they fill up quickly and have a high chance of conflict

axic: You absolutely need to define *some* number otherwise nobody can implement and test it. It is also absolutely fine to resolve conflicts when things are accepted into a protocol upgrade. This has been the practice.

xinbenlv: @axic the placeholder mark is meant to archive exactly what you mean here. 
When using a placeholder mark, implementers have to choose *some* number, and just like all parameters will be passed in, placeholder mark will also be passed in some number by implementer for sure. In addition to choosing *some* number, using placeholder highlights such number is subject to change, not only when finalized, but also during draft. because:"

1. Having a placeholder name helps people search across different implementations, `geth`, `ethereumjs` with consistent parameter naming and make them configurable.
2. This will be particularly useful say when they want to run cross-implementation consensus testing so a tester can config the placeholder to the same number and start testing.

We could avoid this case: different implementers who implemented EIP-663 before this PR *may* *hardcode* the `0xb0` but implementer after this PR *may hardcode* `0xb3` for `DUPN`, and didn't get reminded for paramatize DUPN as a placeholder for cross-platform testing.

xinbenlv: Another benefit is that, if there is any **implied mathematical relationship** between multiple instructions introduced by an EIP, having placeholder marks for all of them kind of help reminding EIP Author to explain such relationship. For example, when I read 

```
`DUPN` (`0xb3`)
`SWAPN ` (`0xb4`)
```

I interpret that it implies you are suggesting they are better to (1) be allocated starting with `0xb_` section, and are better to (2) be consecutive. 

But when I read

```
`DUPN` (`0xb0`)
`SWAPN ` (`0xb1`)
```

In addition to  (1), (2), there is also a possible interpretation (3) that this EIP Author wants to suggest putting DUPN and SWAPN to a **new** Hex-teen(0x10) sector. Such implied interpretation is now removed because of this PR#6054.

axic: > @axic the placeholder mark is meant to archive exactly what you mean here.
When using a placeholder mark, implementers have to choose some number, and just like all parameters will be passed in, placeholder mark will also be passed in some number by implementer for sure

No idea what your suggestion is, but if it is about introducing yet another location where implementers have to take this number from instead of the EIP, than how is that better?

xinbenlv: I mean instead of say 

```suggestion
--- - `DUPN` (`0xb3`)
--- - `SWAPN ` (`0xb4`)

write something like

``` suggestion
+++ - `DUPN` (`PLACEHOLDER_DUPN`)
+++ - `SWAPN ` (`PLACEHOLDER_SWAPN`)
```

And also write 

```suggestion
+++ The `PLACEHOLDER_DUPN` and `PLACEHOLDER_SWAPN` are ideally consecutive and fall into `0xb_` sector. and the exact number shall be determined upon hard-fork.
```

xinbenlv:
I mean instead of say 

```suggestion
--- - `DUPN` (`0xb3`)
--- - `SWAPN ` (`0xb4`)
```
write something like

``` suggestion
+++ - `DUPN` (`PLACEHOLDER_DUPN`)
+++ - `SWAPN ` (`PLACEHOLDER_SWAPN`)
```

And also write 

```suggestion
+++ The `PLACEHOLDER_DUPN` and `PLACEHOLDER_SWAPN` are ideally consecutive and fall into `0xb_` sector. and the exact number shall be determined upon hard-fork.
```

PR #6056, opened 2022-11-28
EIP-4200: Add RJUMPV

axic:


EIPS/eip-4200.md
```diff
@@ -39,17 +39,21 @@ We introduce two new instructions on the same block number [EIP-3540](./eip-3540
...
-If the code is legacy bytecode, both of these instructions result in an *exceptional halt*. (*Note: This means no change to behaviour.*)
+If the code is legacy bytecode, all of these instructions result in an *exceptional halt*. (*Note: This means no change to behaviour.*)
 
 If the code is valid EOF1:
 
-1. `RJUMP relative_offset` (0x5c), sets the `PC` to `PC_post_instruction + relative_offset`.
-2. `RJUMPI relative_offset` (0x5d), pops a value (`condition`) from the stack, and sets the `PC` to `PC_post_instruction + ((condition == 0) ? 0 : relative_offset)`.
+1. `RJUMP relative_offset` sets the `PC` to `PC_post_instruction + relative_offset`.
+2. `RJUMPI relative_offset` pops a value (`condition`) from the stack, and sets the `PC` to `PC_post_instruction + ((condition == 0) ? 0 : relative_offset)`.
+3. `RJUMPV count relative_offset+` pops a value (`case`) from the stack, and sets the `PC` to `PC_post_instruction + ((case >= count) ? 0 : relative_offset[case])`.
```
axic, line 50: This has some interesting properties:
1. In the current version with fallthrough on out of bounds: `RJUMPV 0 offset` will work like an inverted-`RJUMPI`. It will jump to the offset if and only if the `case` is 0, otherwise fall through.
2. If the we choose to require the default target offset (= instead of fallthrough, the first immediate is the "default target" offset): `RJUMPV 0 offset` would be identical to `RJUMP`.

gumb0:
`Abstract`  needs update

EIPS/eip-4200.md
```diff
@@ -39,17 +39,21 @@ We introduce two new instructions on the same block number [EIP-3540](./eip-3540
...
 
 The immediate argument `relative_offset` is encoded as a 16-bit **signed** (two's-complement) big-endian value. Under `PC_post_instruction` we mean the `PC` position after the entire immediate value.
 
-We also extend the validation algorithm of [EIP-3670](./eip-3670.md) to verify that each `RJUMP`/`RJUMPI` has a `relative_offset` pointing to an instruction. This means it cannot point to an immediate data of `PUSHn`/`RJUMP`/`RJUMPI`. It cannot point outside of code bounds. It is allowed to point to a `JUMPDEST`, but is not required to.
+The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value incremented by 1 determines the number of `relative_offset` values following. Based on this it can be seen that the encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases one would place the *default* path following the `RJUMPV` instruction.
```
gumb0, line 54: why incremented by 1?

EIPS/eip-4200.md
```diff
@@ -39,17 +39,21 @@ We introduce two new instructions on the same block number [EIP-3540](./eip-3540
...
+
+We also extend the validation algorithm of [EIP-3670](./eip-3670.md) to verify that each `RJUMP`/`RJUMPI`/`RJUMPV` has a `relative_offset` pointing to an instruction. This means it cannot point to an immediate data of `PUSHn`/`RJUMP`/`RJUMPI`. It cannot point outside of code bounds. It is allowed to point to a `JUMPDEST`, but is not required to.
```
gumb0, line 56:
```suggestion
We also extend the validation algorithm of [EIP-3670](./eip-3670.md) to verify that each `RJUMP`/`RJUMPI`/`RJUMPV` has a `relative_offset` pointing to an instruction. This means it cannot point to an immediate data of `PUSHn`/`RJUMP`/`RJUMPI`/`RJUMPV`. It cannot point outside of code bounds. It is allowed to point to a `JUMPDEST`, but is not required to.
```

EIPS/eip-4200.md
gumb0: I guess to avoid empty jump table.

`RJUMPV 0` could be a no-op or rejected at validation.

EIPS/eip-4200.md
axic: Fixed.

EIPS/eip-4200.md
axic: > I guess to avoid empty jump table.

Yes.

> RJUMPV 0 could be a no-op or rejected at validation.

Yes that's an option.

EIPS/eip-4200.md
```diff
@@ -13,7 +13,7 @@ requires: 3540, 3670
 
 ## Abstract
 
-Two new EVM jump instructions are introduced (`RJUMP` and `RJUMPI`) which encode the destination as a signed immediate value. These can be useful in the majority of (but not all) use cases and offer a cost reduction.
+Three new EVM jump instructions are introduced (`RJUMP`, `RJUMPI` and `RJUMPV`) which encode the destination as a signed immediate value. These can be useful in the majority of (but not all) use cases and offer a cost reduction.
```
chfast, line 16:
```suggestion
Three new EVM jump instructions are introduced (`RJUMP`, `RJUMPI` and `RJUMPV`) which encode destinations as signed immediate values. These can be useful in the majority of (but not all) use cases and offer a cost reduction.
```

EIPS/eip-4200.md
```diff
@@ -39,17 +39,21 @@ We introduce two new instructions on the same block number [EIP-3540](./eip-3540
...
+3. `RJUMPV` (0x5e)
 
-If the code is legacy bytecode, both of these instructions result in an *exceptional halt*. (*Note: This means no change to behaviour.*)
+If the code is legacy bytecode, all of these instructions result in an *exceptional halt*. (*Note: This means no change to behaviour.*)
 
 If the code is valid EOF1:
 
-1. `RJUMP relative_offset` (0x5c), sets the `PC` to `PC_post_instruction + relative_offset`.
-2. `RJUMPI relative_offset` (0x5d), pops a value (`condition`) from the stack, and sets the `PC` to `PC_post_instruction + ((condition == 0) ? 0 : relative_offset)`.
+1. `RJUMP relative_offset` sets the `PC` to `PC_post_instruction + relative_offset`.
```
chfast, line 48: I cannot find where this `PC_post_instruction` is defined.

EIPS/eip-4200.md
lightclient: (1) seems nice

EIPS/eip-4200.md
lightclient: Is there a reason to not just reject `count=0`?

EIPS/eip-4200.md
axic: On L52.

EIPS/eip-4200.md
```diff
@@ -39,17 +39,21 @@ We introduce two new instructions on the same block number [EIP-3540](./eip-3540
...
+The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value incremented by 1 determines the number of `relative_offset` values following. Based on this it can be seen that the encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases one would place the *default* path following the `RJUMPV` instruction.
+
+We also extend the validation algorithm of [EIP-3670](./eip-3670.md) to verify that each `RJUMP`/`RJUMPI`/`RJUMPV` has a `relative_offset` pointing to an instruction. This means it cannot point to an immediate data of `PUSHn`/`RJUMP`/`RJUMPI`/`RJUMPV`. It cannot point outside of code bounds. It is allowed to point to a `JUMPDEST`, but is not required to.
 
 Because the destinations are validated upfront, the cost of these instructions are less than their dynamic counterparts: `RJUMP` should cost 5, and `RJUMPI` should cost 7. This is a reduction of 2 gas, compared to `JUMP` and `JUMPI`.
```
gumb0, line 58: Need to add cost for `RJUMPV`

axic: I thought I pushed that version.

gumb0 (approved):

EIPS/eip-4200.md
axic: In the current description count is always `+1`d, so this removes the need for a special case of "reject count 0". We can go either way.

axic: We can also keep `count=0` as yet another NOP-style instruction if that's better.

PR #6057, opened 2022-11-28
Update EIP-1153: Move to last call

moodysalem:
EIP-1153 is CFI, so marking it last call


Pandapip1 (commented):
Is this scheduled to be included in an upcoming hard fork?

Pandapip1 (approved):
Never mind, I can't read.

lightclient:
We usually don't move EIPs to last call until they are on a testnet in the unlikely event that something minor needs to change about the EIP before it goes to mainnet.

PR #6077, opened 2022-12-02
Add EIP-6077: Invalidation for Signature-Based Operations

k06a:


frangio:
In the context of permits I think I've seen you suggest that nonces shouldn't be sequential, so that there could be several valid permits created "in parallel". What's your current thinking on that?

k06a:
@frangio maybe we can rename `nonce` to `operationId` and make those EIP to support custom invalidation logic. Implementations could use sequential ids or parallel ids however they want.

frangio:
Can you name use cases that require parallel nonces? Is this common or a good idea? How does a client know which nonce to use? I'm trying to understand what extra complexity it would require and whether it might be worth it.

k06a:
@frangio I discovered these 5 operations in [OpenZeppelin/openzeppelin-contracts](https://github.com/OpenZeppelin/openzeppelin-contracts):
- `ERC20Permit._PERMIT_TYPEHASH` (sequential invalidation)
- `Votes._DELEGATION_TYPEHASH` (sequential invalidation)
- `Governor.BALLOT_TYPEHASH` (parallel invalidation via `proposalId`)
- `Governor.EXTENDED_BALLOT_TYPEHASH` (parallel invalidation via `proposalId`)
- `MinimalForwarder._TYPEHASH` (sequential invalidation)

frangio:
Based on these examples I don't think parallel nonces are needed. `proposalId` is not a nonce.

k06a:
@frangio I see we probably don’t need parallel operation ids. But maybe we don’t have good invalidation technique, which can be invented later. I suspect there could be some cases where operation id contains spender/delegatee address and rest 96 bits are being used as nonce.

frangio:
I'd remove `IParallelOperations` to simplify the EIP.

I think if there is parallelism it will still be sequential but the sequence will depend on more parameters, not just the signer. This could be expressed in the interface by including an additional argument like `bytes params`, but this is useless without a way to specify the parameters that should be included there.

k06a:
@frangio what about having definition for thread-id? Original sequential behaviour have thread-id = signer, but we could have thread-id = hash(signer, spender). Nonce will be sequential but one signer could have multiple threads of ids...

k06a:
@Pandapip1 could you help us here with EIP number?

frangio:
@k06a A thread id would work but it's similar to (though more efficient than) `bytes params`, and I think there would need to be a way to specify to clients how to create the thread id.

IMO, the ideal end goal for full generality would be the ability to create a function like `getNonce(contract: Address, signer: Address, type: EIP712Type, msgWithoutNonce: EIP712Message): Promise<bigint>`. I.e., given a contract and an EIP712 typed message with an empty nonce parameter obtain the nonce that should be included in that nonce parameter. Your proposal here allows a less general version of this function where the nonce does not depend on the message or other parameters, only depends on the signer. I do think this is valuable enough to standardize just that. I'm not sure how to design the more general version or if it even makes sense to think about it...

EIPS/eip-X.md
```diff
@@ -0,0 +1,73 @@
+---
+eip: TBD
```
Pandapip1, line 2:
```suggestion
eip: 6077
```

Pandapip1 (changes requested):
Please fix the walidator errors

EIPS/eip-6077.md
```diff
@@ -0,0 +1,73 @@
+---
+eip: 6077
+title: Invalidation abstraction for signature-based operations powered by EIP-712
```
Pandapip1, line 3:
```suggestion
title: Invalidation Abstraction for Signature-Based Operations
```
Should be title case, and the title was too long

EIPS/eip-6077.md
```diff
@@ -0,0 +1,73 @@
...
+description: Extends EIP-712 and unifies EIP-2612 with many others
+author: Anton Bukov (@k06a), Mikhail Melnik (@zumzoom)
+discussions-to: TBD
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-12-02
+requires: 712
```
Pandapip1, line 11:
```suggestion
requires: 712, 2612
```

EIPS/eip-6077.md
```diff
@@ -0,0 +1,73 @@
+---
+eip: 6077
+title: Invalidation abstraction for signature-based operations powered by EIP-712
+description: Extends EIP-712 and unifies EIP-2612 with many others
+author: Anton Bukov (@k06a), Mikhail Melnik (@zumzoom)
+discussions-to: TBD
```
Pandapip1, line 6: Needs to be a thread on ethereum magicians

EIPS/eip-6077.md
```diff
@@ -0,0 +1,73 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-12-02
+requires: 712
+---
+
+## Abstract
+
+This EIP extends [EIP-712](./eip-712.md) and defines two ways to validate and invalidate signature-based operations. Such operations as [EIP-20](./eip-20.md) permit operation defined in [EIP-2612](./eip-2612.md). This EIP provides two ways to track such operations and invalidate them:
```
Pandapip1, line 16:
```suggestion
This EIP extends [EIP-712](./eip-712.md) and defines two methods to validate and invalidate signature-based operations, such as [EIP-2612](./eip-20.md)'s permit operation. This EIP provides two ways to track such operations and invalidate them:

```
Fixes grammar

EIPS/eip-6077.md
```diff
@@ -0,0 +1,73 @@
...
+Same abstraction could be utilized by mutilple signature-based operations, moreover existing EIPs like [EIP-2612](./eip-2612.md) could be considered as fully compatible with the EIP.
+
+Multiple EIPs define operations that can be executed in behalf of signer and sometime introduce method naming collisions and other. For example, [EIP-2612](./eip-2612.md) defines both `permit` and `nonces` methods, but gives no clue that `nonces` is related to permit operation. In case of multiple same-level EIPs implemented within one smart contract (for example: permit, delegate, vote) it's obvious that they should use different nonces.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+- Smart contract implementing EIP-712 MUST also implement the following interface:
+    ```solidity
```
Pandapip1, line 31:
```suggestion

    ```solidity
```

Pandapip1 (changes requested):

EIPS/eip-6077.md
k06a: Thx, added: https://ethereum-magicians.org/t/eip-6077-invalidation-abstraction-for-signature-based-operations/12162

EIPS/eip-6077.md
```diff
@@ -0,0 +1,75 @@
+---
+eip: 6077
+title: Invalidation Abstraction for Signature-Based Operations
```
SamWilsn, line 3:
```suggestion
title: Invalidation for Signature-Based Operations
```

EIPS/eip-6077.md
```diff
@@ -0,0 +1,75 @@
...
+eip: 6077
+title: Invalidation for Signature-Based Operations
+description: Extends EIP-712 and unifies EIP-2612 with many others
+author: Anton Bukov (@k06a), Mikhail Melnik (@zumzoom)
+discussions-to: https://ethereum-magicians.org/t/eip-6077-invalidation-abstraction-for-signature-based-operations/12162
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-12-02
+requires: 712, 2612
```
Pandapip1, line 11: Is EIP-2612 really *required*?
```suggestion
requires: 712
```

EIPS/eip-6077.md
```diff
@@ -0,0 +1,75 @@
...
+This EIP extends [EIP-712](./eip-712.md) and defines two methods to validate and invalidate signature-based operations, such as [EIP-2612](./eip-20.md)'s permit operation. This EIP provides two ways to track such operations and invalidate them:
+
+- main sequnce of incremental nonces per signer
+- sequence of incremental ids per signer per beneficiary
+
+## Motivation
+
+Same abstraction could be utilized by mutilple signature-based operations, moreover existing EIPs like [EIP-2612](./eip-2612.md) could be considered as fully compatible with the EIP.
+
+Multiple EIPs define operations that can be executed in behalf of signer and sometime introduce method naming collisions and other. For example, [EIP-2612](./eip-2612.md) defines both `permit` and `nonces` methods, but gives no clue that `nonces` is related to permit operation. In case of multiple same-level EIPs implemented within one smart contract (for example: permit, delegate, vote) it's obvious that they should use different nonces.
```
Pandapip1, line 25: I can't understand what you're trying to say here. Mind explaining?

EIPS/eip-6077.md
```diff
@@ -0,0 +1,75 @@
...
+## Abstract
+
+This EIP extends [EIP-712](./eip-712.md) and defines two methods to validate and invalidate signature-based operations, such as [EIP-2612](./eip-20.md)'s permit operation. This EIP provides two ways to track such operations and invalidate them:
+
+- main sequnce of incremental nonces per signer
+- sequence of incremental ids per signer per beneficiary
+
+## Motivation
+
+Same abstraction could be utilized by mutilple signature-based operations, moreover existing EIPs like [EIP-2612](./eip-2612.md) could be considered as fully compatible with the EIP.
```
Pandapip1, line 23: This reads like Rationale, not Motivation. Mind moving this to that section?

EIPS/eip-6077.md
```diff
@@ -0,0 +1,75 @@
...
+created: 2022-12-02
+requires: 712, 2612
+---
+
+## Abstract
+
+This EIP extends [EIP-712](./eip-712.md) and defines two methods to validate and invalidate signature-based operations, such as [EIP-2612](./eip-20.md)'s permit operation. This EIP provides two ways to track such operations and invalidate them:
+
+- main sequnce of incremental nonces per signer
+- sequence of incremental ids per signer per beneficiary
```
Pandapip1, line 19: These should be complete sentences

EIPS/eip-6077.md
```diff
@@ -0,0 +1,75 @@
...
+
+        /// @dev Increments id for the caller in the context of the operation typehash and operation beneficiary
+        /// @param typehash The operation typehash
+        /// @param beneficiary The address of the spender, delegate, or other beneficiary of the transaction
+        /// @param id The operation nonce
+        function useOperationIds(bytes32 typehash, address beneficiary, uint256 id) external;
+    }
+    ```
+
+- Operation EIPs SHOULD use at leat one of the nonces or ids sequences per signer defined by this EIP or both.
```
Pandapip1, line 59: Why not make this a MUST? Also, the phrasing here is a bit odd. Mind explaining?

EIPS/eip-6077.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Rationale
+
+TBD
+
+## Backwards Compatibility
+
+Fully backward compatibile with EIP-712.
```
Pandapip1, line 67:
```suggestion
This EIP is backward compatible with EIP-712.
```

Pandapip1 (changes requested):

EIPS/eip-6077.md
k06a: Trying to create abstraction for signature-based operations, and this abstraction also covers existing EIP-2612 somehow.

k06a:
Don't close it, it's still under discussion and consideration

k06a:
@frangio what do you think? Is it worth fixing? Do you still consider this kind of functionality in OZ library?

frangio:
@k06a We're not quite thinking about this at the moment. Personally I have to look into this topic a bit more and it may need more research and discussion, and I don't think this is the place for that, so I would close this and might revisit later.

PR #6078, opened 2022-12-02
Update-4200: disallow empty jump table

gumb0:
1.  Reject `RJUMPV` with empty jump table at validation.
2. Fix off-by-one bug in reference code when reading jump table size.
3. Update reference code to disallow CALLCODE ad SELFDESTRUCT according to current spec of EIP-3670.

### DIFF
--- a/EIPS/eip-4200.md
+++ b/EIPS/eip-4200.md
@@ -51,7 +51,7 @@ If the code is valid EOF1:
 
 The immediate argument `relative_offset` is encoded as a 16-bit **signed** (two's-complement) big-endian value. Under `PC_post_instruction` we mean the `PC` position after the entire immediate value.
 
-The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value determines the number of `relative_offset` values following. Validation algorithm of [EIP-3670](./eip-3670.md) is extended to verify that `count >= 1`. The encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases one would place the *default* path following the `RJUMPV` instruction.
+The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value determines the number of `relative_offset` values following. Validation algorithm of [EIP-3670](./eip-3670.md) is extended to verify that `count >= 1`. The encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases one would place the *default* path following the `RJUMPV` instruction. An interesting feature is that `RJUMPV 1 relative_offset` is an inverted-`RJUMPI`, which can be used in many cases instead of `ISZERO RJUMPI relative_offset`.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-4200.md
```diff
@@ -47,11 +47,11 @@ If the code is valid EOF1:
 
 1. `RJUMP relative_offset` sets the `PC` to `PC_post_instruction + relative_offset`.
 2. `RJUMPI relative_offset` pops a value (`condition`) from the stack, and sets the `PC` to `PC_post_instruction + ((condition == 0) ? 0 : relative_offset)`.
-3. `RJUMPV count relative_offset+` pops a value (`case`) from the stack, and sets the `PC` to `PC_post_instruction + ((case >= (count + 1)) ? 0 : relative_offset[case])`.
+3. `RJUMPV count relative_offset+` pops a value (`case`) from the stack, and sets the `PC` to `PC_post_instruction + ((case >= count) ? 0 : relative_offset[case])`.
```
axic, line 50:
```suggestion
3. `RJUMPV count relative_offset*` pops a value (`case`) from the stack, and sets the `PC` to `PC_post_instruction + ((case >= count) ? 0 : relative_offset[case])`.
```

EIPS/eip-4200.md
```diff
@@ -47,11 +47,11 @@ If the code is valid EOF1:
...
 
 The immediate argument `relative_offset` is encoded as a 16-bit **signed** (two's-complement) big-endian value. Under `PC_post_instruction` we mean the `PC` position after the entire immediate value.
 
-The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value incremented by 1 determines the number of `relative_offset` values following. Based on this it can be seen that the encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases one would place the *default* path following the `RJUMPV` instruction.
+The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value determines the number of `relative_offset` values following. Validation algorithm of [EIP-3670](./eip-3670.md) is extended to verify that `count >= 1`. The encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases one would place the *default* path following the `RJUMPV` instruction.
```
axic, line 54:
```suggestion
The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value determines the number of `relative_offset` values following. Validation algorithm of [EIP-3670](./eip-3670.md) is extended to verify that `count >= 1`. The encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases one would place the *default* path following the `RJUMPV` instruction. An interesting feature is that `RJUMPV 1 relative_offset` is an inverted-`RJUMPI`, which can be used in many cases instead of `ISZERO RJUMPI relative_offset`.
```

ekpyron:
I'd actually instead of this just make it ``max_case`` instead of ``count`` and adjust the phrasing accordingly (like ``case <= max_case ? relative_offset[case] : 0``, etc.). That still gives you ``max_case + 1`` as the number of relative_offsets, but that's rather natural - and you don't get the noop case.

But also find with the change in the PR :-).

EIPS/eip-4200.md
gumb0: but empty table is not allowed, so there's always at least 1 offset

axic: Yes, but that is not a rule of instruction encoding, but validation :)

axic (approved):

EIPS/eip-4200.md
gumb0: I think `*` would be confusing, merging with `+`

PR #6096, opened 2022-12-07
Update EIP-4200: Add Test Cases for RJUMPV

gumb0:


jochem-brouwer:
I gave this some thought, but at first glance this indeed covers all `RJUMPV` test cases. Thanks!

PR #6107, opened 2022-12-09
Update ci.yml for ubuntu version fixing HTMLProofer

xinbenlv:
Recently, Github updated its `ubuntu-latest` to a new version `ubuntu-22.04` so ruby 2.6 is no longer supported, resulting [this kind of failure](https://github.com/ethereum/EIPs/actions/runs/3654459792/jobs/6174903582)

```
Error: Error: Unknown version 2.6.0 for ruby on ubuntu-22.04
        available versions for ruby on ubuntu-22.04: 3.1.0, 3.1.1, 3.1.2, 3.2.0-preview1, head, debug
        Make sure you use the latest version of the action with - uses: ruby/setup-ruby@v1
```

Here is the fix steps tried:
1. Update the ruby to `3.1.2` which is latest stable release available. result: https://github.com/ethereum/EIPs/actions/runs/3654483141/jobs/6174948461 yields `minitest-5.14.1 requires ruby version ~> 2.2, which is incompatible with the`
2. Revert the `ubuntu-latest` to `ubuntu-20.04`, passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     
     steps:
       - name: Checkout EIP Repository
-        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        uses: actions/checkout@3
 
       - name: Install OpenSSL
         run: sudo apt-get update && sudo apt-get install -y libcurl4-openssl-dev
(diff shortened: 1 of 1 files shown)


SamWilsn (approved):

xinbenlv:
All check passed. Please let me now if there are other approval it needs

PR #6155, opened 2022-12-16
Update EIP-6046: Propose alternative deactivation

Pandapip1:
This changes EIP-6046 such that deactivation effectively clears a contract's memory, at the cost of additional gas costs. Basically, it turns deactivated contracts into symbolic links to other contracts.

This may or may not get used, hence why I am keeping this PR as a draft.

CC @axic


Pandapip1:
Closing and splitting into two EIPs.

PR #6156, opened 2022-12-16
3540,4570,5450: update EOF container format

lightclient:
This PR updates the EOF container format to include the size optimizations we've been discussing. Full change log:

* Require EIP-4750 (due to type section) and EIP-5450 (due to `max_stack_height`)
* Require data section, even if size 0.
* Introduce "list" encoding in section headers. This is defined to be `kind, list_size, size+`.
* Restrict EIP-4750 input and output identifiers to 7 bits. The upper bit is reserved for future uses, such as denoting that a function does not return.
* ~~`max_stack_height` information is compressed from 2 bytes to 1 byte by calculating the true, unbounded max stack height `h` and compressing with the formula `ceil(h / 8)`.~~
* Removed references to `JUMP` and `JUMPI` in EIP-3540.
* Disallow unreachable code.
* Removed reference implementation and some testcases for now. I think the spec is the most important and I would rather keep iterating on there than update these right now.
* Remove stack cleanup logic for `RETF`.
* Remove `JUMPF` instruction - it is complicated to define it in the context of runtime checks in 4750, we'll introduce it separately on top of 5450.


EIPS/eip-3540.md
```diff
@@ -8,7 +8,7 @@ status: Review
 type: Standards Track
 category: Core
 created: 2021-03-16
-requires: 3541, 3860
+requires: 3541, 3860, 4750, 5450
```
chfast, line 11: Isn't this a cyclic dependency?

EIPS/eip-4750.md
```diff
@@ -32,23 +32,14 @@ Furthermore, it aims to improve analysis opportunities by encoding the number of
...
+4. New section with `kind = 1` is introduced called the *type section*.
 5. Exactly one type section MAY be present.
 6. The type section, if present, MUST directly precede all code sections.
-7. The type section, if present, contains a sequence of pairs of bytes: first byte in a pair encodes number of inputs, and second byte encodes number of outputs of the code section with the same index. *Note:* This implies that there is a limit of 256 stack for the input and in the output.
-8. Therefore, type section size MUST be `n * 2` bytes, where `n` is the number of code sections.
+7. The type section, if present, contains a sequence of byte triples: first byte in the triple encodes number of inputs, and second byte encodes number of outputs of the code section with the same index, and third byte encodes `ceil(max_stack_height / 8)` where `max_stack_height` is defined in [EIP-5450](./eip-5450.md). *Note:* This implies that there is a limit of 256 stack for the input and in the output. This is actually further restricted to 127 stack items because the upper bit of both the input and output bytes are reserved for future use.
```
jochem-brouwer, line 38:
```suggestion
7. The type section, if present, contains a sequence of byte triples: first byte in the triple encodes number of inputs, and second byte encodes number of outputs of the code section with the same index, and third byte encodes `ceil(max_stack_height / 8)` where `max_stack_height` is defined in [EIP-5450](./eip-5450.md). *Note:* This implies that there is a limit of 255 stack for the input and in the output. This is actually further restricted to 127 stack items because the upper bit of both the input and output bytes are reserved for future use.
```

jochem-brouwer:
I am sorry I could not make the breakout room.

I have a question about these upper bits for `CALLF`. If we want to change the spec for "future use cases", and code is validated at deploy time, then we could just create EOF2 format, right? What is the motivation to restrict this?

EIPS/eip-5450.md
```diff
@@ -38,15 +38,16 @@ At the same time the following properties are being verified:
...
 5. For `RETF` instruction, data stack before executing it has exactly `n` items to use as output values returned from a function, where `n` is function's number of outputs according to its type defined in the type section.
-6. Maximum data stack height required by a function does not exceed `1024`.
+6. `ceil(maximum data stack height / 8)` matches the corresponding code section's `max_stack_height` within the type section body.
+7. No unreachable instructions exist in the code section.
 
-To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. See below for reference implementation.
+To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. After completing this, verify each instruction was visited at least once. Fail if any instructions were not visited as this invalidates 9). See below for reference implementation.
```
jochem-brouwer, line 44:
```suggestion
To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. After completing this, verify each instruction was visited at least once. Fail if any instructions were not visited as this invalidates 9).
```

EIPS/eip-4750.md
```diff
@@ -32,23 +32,14 @@ Furthermore, it aims to improve analysis opportunities by encoding the number of
 1. The requirement of [EIP-3540](./eip-3540.md) "Exactly one code section MUST be present." is relaxed to "At least one code section MUST be present.", i.e. multiple code sections (`kind = 1`) are allowed.
 2. Total number of code sections MUST NOT exceed 1024.
 3. All code sections MUST precede a data section, if data section is present.
-4. New section with `kind = 3` is introduced called the *type section*.
+4. New section with `kind = 1` is introduced called the *type section*.
```
jochem-brouwer, line 35: Type section is now of same type than code section..?

EIPS/eip-4750.md
```diff
@@ -32,23 +32,14 @@ Furthermore, it aims to improve analysis opportunities by encoding the number of
...
 6. The type section, if present, MUST directly precede all code sections.
-7. The type section, if present, contains a sequence of pairs of bytes: first byte in a pair encodes number of inputs, and second byte encodes number of outputs of the code section with the same index. *Note:* This implies that there is a limit of 256 stack for the input and in the output.
-8. Therefore, type section size MUST be `n * 2` bytes, where `n` is the number of code sections.
+7. The type section, if present, contains a sequence of byte triples: first byte in the triple encodes number of inputs, and second byte encodes number of outputs of the code section with the same index, and third byte encodes `ceil(max_stack_height / 8)` where `max_stack_height` is defined in [EIP-5450](./eip-5450.md). *Note:* This implies that there is a limit of 256 stack for the input and in the output. This is actually further restricted to 127 stack items because the upper bit of both the input and output bytes are reserved for future use.
+8. Therefore, type section size MUST be `n * 3` bytes, where `n` is the number of code sections.
 9. First code section MUST have 0 inputs and 0 outputs.
```
jochem-brouwer, line 40: Could we not default this to omitting this "first code section" entirely and just implicitly make it have 0 inputs and 0 outputs? (This also makes it easier (?) if there is thus only one code section, now the whole type section MUST be ommitted)

EIPS/eip-3540.md
```diff
@@ -79,58 +81,95 @@ The container starts with the EOF header:
...
 
-#### Section kinds
+EOF version 1 is made up of 5 EIPs, including this one: [EIP-3540](./eip-3540.md), [EIP-3670](./eip-3670.md), [EIP-4200](./eip-4200.md), [EIP-4750](./eip-4750.md), and [EIP-5450](./eip-5450.md). Some values in this specification are only discussed briefly. To understand the full scope of EOF, it is necessary to review each EIP in-depth.
+
+#### Container
+
+The EOF version 1 container consists of a `header` and `body`.
+
+```
+magic      := hex"ef00"
```
chfast, line 111: These constants are duplicated in the table below. Can we skip them here?

EIPS/eip-3540.md
```diff
@@ -79,58 +81,95 @@ The container starts with the EOF header:
...
-| 1            | code                                           |
-| 2            | data                                           |
+##### Body
+
+| name             | length   | value         | description |
+|------------------|----------|---------------|-------------|
+| type_section     | variable | n/a       | stores EIP-4750 and EIP-5450 code section metadata  |
+| inputs           | 1 byte   | 0x00-0x7F | number of stack elements the code section consumes  |
+| outputs          | 1 byte   | 0x00-0x7F | number of stack elements the code section returns   |
+| max_stack_height | 1 bytes  | 0x00-0xFF | `ceil(max_stack_height / 8)` during execution       |
```
chfast, line 145: There are still concerns if the value should be stored in the EOF at all. I think for now we should store uncompressed value (i.e. 2 bytes).

EIPS/eip-3540.md
```diff
@@ -79,58 +81,95 @@ The container starts with the EOF header:
...
+| version           | 1 byte   | 0x01          | EOF version                                                                        |
+| kind_type         | 1 byte   | 0x01          | kind marker for EIP-4750 type section header                                       |
+| type_size         | 2 bytes  | 0x0003-0xFFFF | uint16 denoting the length of the type section content                             |
+| kind_code         | 1 byte   | 0x02          | kind marker for code size section                                                  |
+| num_code_sections | 2 bytes  | 0x0001-0xFFFF | uint16 denoting the number of the code sections                                    |
+| code_size         | 2 bytes  | 0x0001-0xFFFF | uint16 denoting the length of the code section content                             |
+| kind_data         | 1 byte   | 0x03          | kind marker for data size section                                                  |
+| data_size         | 2 bytes  | 0x0001-0xFFFF | uint16 integer denoting the length of the data section content                     |
```
chfast, line 135: Because this proposes the data section to be mandatory we should allow a data section of size 0.

EIPS/eip-3540.md
```diff
@@ -79,58 +81,95 @@ The container starts with the EOF header:
...
 
-1. `JUMPDEST`-analysis is only run on the *code*.
-2. Execution starts at the first byte of the *code*, and `PC` is set to 0.
-3. Execution stops if `PC` goes outside the code section bounds.
-4. `PC` returns the current position within the *code*.
-5. `JUMP`/`JUMPI` uses an absolute offset within the *code*.
-6. `CODECOPY`/`CODESIZE`/`EXTCODECOPY`/`EXTCODESIZE`/`EXTCODEHASH` keeps operating on the entire *container*.
-7. The input to `CREATE`/`CREATE2` is still the entire *container*.
-8. The size limit for deployed code as specified in [EIP-170](./eip-170.md) and for init code as specified in [EIP-3860](./eip-3860.md) is applied to the entire *container* size, not to the *code* size.
+1. Execution starts at the first byte of the *code*, and `PC` is set to 0.
```
chfast, line 165: I think we should also mention that the execution starts in the first section. This is defined in 4750, but here it looks like something is missing.

EIPS/eip-3540.md
```diff
@@ -79,58 +81,95 @@ The container starts with the EOF header:
...
 
 1. In addition to general validation rules above, EOF version 1 bytecode conforms to the rules specified below:
-   - Exactly one code section MUST be present.
-   - The code section MUST be the first section.
-   - A single data section MAY follow the code section.
-2. Any other version is invalid.
+   - One type section header MUST be present immediately following the EOF version. Each code section MUST have a specified type signature in the type body.
+   - One code section header MUST be present immediately following the type section. A maxmimum of 1024 individual code sections are allowed.
+   - One data section header MUST be present immediately following the code section.
+2. Any other version than `0x01` is invalid.
```
gumb0, line 157:
```suggestion
2. Any version other than `0x01` is invalid.
```

EIPS/eip-4750.md
gumb0: No code section kind now is `2`, see changes to EIP-3540

EIPS/eip-4750.md
```diff
@@ -32,23 +32,14 @@ Furthermore, it aims to improve analysis opportunities by encoding the number of
 1. The requirement of [EIP-3540](./eip-3540.md) "Exactly one code section MUST be present." is relaxed to "At least one code section MUST be present.", i.e. multiple code sections (`kind = 1`) are allowed.
 2. Total number of code sections MUST NOT exceed 1024.
 3. All code sections MUST precede a data section, if data section is present.
-4. New section with `kind = 3` is introduced called the *type section*.
+4. New section with `kind = 1` is introduced called the *type section*.
 5. Exactly one type section MAY be present.
```
gumb0, line 36:
```suggestion
5. Exactly one type section MUST be present.
```

EIPS/eip-4750.md
```diff
@@ -32,23 +32,14 @@ Furthermore, it aims to improve analysis opportunities by encoding the number of
...
 6. The type section, if present, MUST directly precede all code sections.
```
gumb0, line 37:
```suggestion
6. The type section MUST directly precede all code sections.
```

EIPS/eip-4750.md
gumb0, line 38: 7. The type section ~~, if present,~~ contains a sequence...

EIPS/eip-4750.md
gumb0: We still need `max_stack_height` of it.

gumb0:
> I have a question about these upper bits for `CALLF`. If we want to change the spec for "future use cases", and code is validated at deploy time, then we could just create EOF2 format, right? What is the motivation to restrict this?

We generally want to minimize the need to introduce more EOF versions and be as forward-compatible as possible, because more EOF versions would mean having more EVM implementations in clients.

EIPS/eip-4750.md
gumb0: We still need `max_stack_height` of it.

EIPS/eip-3540.md
lightclient: We can resolve this at a later point. Main focus right now is just getting the spec down.

EIPS/eip-3540.md
```diff
@@ -79,58 +81,95 @@ The container starts with the EOF header:
...
+| section_size_list | dynamic | n/a           | `uint16, uint16+` |
 
-The list of section headers is terminated with the *section headers terminator byte* `0x00`.
+The list of section headers is terminated with the *section headers terminator byte* `0x00`. The body content follows immediately after.
 
 #### Container validation rules
 
 1. `version` MUST NOT be `0`.[^1](#EOF-version-range-start-with-1)
 2. `section_kind` MUST NOT be `0`. The value `0` is reserved for *section headers terminator byte*.
-3. `section_size` MUST NOT be `0`. If a section is empty its section header MUST be omitted.
```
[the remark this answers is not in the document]
lightclient: I don't see how requiring all defined sections today affects our ability to in the future add new section kinds without bumping EOF. We will simply update our parsers to check if the last byte after all the required sections is the terminator byte. If it isn't, we can interpret it as a new, optional kind type.

Having optional sections marginally increases the complexity of parsing the header. I think the cost of 1-2 extra bytes is worth it. I also believe we can in the future relax this requirement and allow certain sections to be optional.

EIPS/eip-3540.md
lightclient: What are the concerns of storing this in EOF? I think not storing it in the type section is basically a non-starter as it makes it much more complicated for bootstrapping nodes to get the type information.

I will update to use the uncompressed value though. We should consider increasing the the max code size limit at some point to make up for this extra data.

EIPS/eip-3540.md
```diff
@@ -79,58 +81,91 @@ The container starts with the EOF header:
...
-| 1            | code                                           |
-| 2            | data                                           |
+##### Body
+
+| name             | length   | value         | description |
+|------------------|----------|---------------|-------------|
+| type_section     | variable | n/a       | stores EIP-4750 and EIP-5450 code section metadata  |
+| inputs           | 1 byte   | 0x00-0x7F | number of stack elements the code section consumes  |
+| outputs          | 1 byte   | 0x00-0x7F | number of stack elements the code section returns   |
+| max_stack_height | 1 bytes  | 0x00-0xFF | max height of data stack during execution           |
```
chfast, line 141:
```suggestion
| max_stack_height | 2 bytes  | 0x00-0x3FF | max height of data stack during execution           |
```

EIPS/eip-4750.md
```diff
@@ -32,23 +32,14 @@ Furthermore, it aims to improve analysis opportunities by encoding the number of
...
-8. Therefore, type section size MUST be `n * 2` bytes, where `n` is the number of code sections.
+4. New section called the *type section* is introduced, identified by `kind_type` (`0x01`).
+5. Exactly one type section MUST be present.
+6. The type section MUST directly precede all code sections.
+7. The type section, if present, contains a sequence of byte triples: first byte in the triple encodes number of inputs, and second byte encodes number of outputs of the code section with the same index, and third byte encodes `max_stack_height`, which is defined in [EIP-5450](./eip-5450.md). *Note:* This implies that there is a limit of 255 stack for the input and in the output. This is actually further restricted to 127 stack items because the upper bit of both the input and output bytes are reserved for future use.
+8. Therefore, type section size MUST be `n * 3` bytes, where `n` is the number of code sections.
```
chfast, line 39:
```suggestion
8. Therefore, type section size MUST be `n * 4` bytes, where `n` is the number of code sections.
```

EIPS/eip-4750.md
```diff
@@ -103,18 +94,16 @@ If the code is valid EOF1, the following execution rules apply:
 #### `JUMPF`
 
 1. Has one immediate argument,`code_section_index`, encoded as a 16-bit unsigned big-endian value.
-2. If data stack has less than `caller_stack_height + type[code_section_index].inputs` items, execution results in exceptional halt.
-3. If data stack has more than `caller_stack_height + type[code_section_index].inputs` items, discards the items between `caller_stack_height` and top `type[code_section_index].inputs` items, so that there are exactly `caller_stack_height +  type[code_section_index].inputs` items left.
-4. Charges 4 gas.
+2. If data stack does not equal `caller_stack_height + type[code_section_index].inputs` items, execution results in exceptional halt.
```
chfast, line 97:
```suggestion
2. If data stack does not equal `caller_stack_height - type[code_section_index].inputs + type[code_section_index].outputs` items, execution results in exceptional halt.
```

Erik4569: JUMPF

EIPS/eip-4750.md
```diff
@@ -103,18 +94,16 @@ If the code is valid EOF1, the following execution rules apply:
...
 5. Pops nothing and pushes nothing to data stack.
 6. Pushes nothing to return stack.
 7. Sets `current_section_index` to `code_section_index` and `PC` to `0`, and execution continues in the called section.
 
 #### `RETF`
 
 1. Does not have immediate arguments.
-2. If data stack has less than `caller_stack_height + type[code_section_index].outputs` items, execution results in exceptional halt.
-3. If data stack has more than `caller_stack_height + type[code_section_index].outputs` items, discards the items between `caller_stack_height` and top `type[code_section_index].outputs` items, so that there are exactly `caller_stack_height +  type[code_section_index].outputs` items left.
+2. If data stack does not equal `caller_stack_height + type[code_section_index].outputs` items, execution results in exceptional halt.
```
chfast, line 106:
```suggestion
2. If data stack does not equal `caller_stack_height + type[current_section_index].outputs` items, execution results in exceptional halt.
```

EIPS/eip-4750.md
gumb0: In this case condition 2 would be redundant.

gumb0: I think I understand now that both are needed, but it looks very confusing because equality requirement is always more strict than "less than" requirement... But here the order matters.

I think this combination would be equivalent:
Your number 3
```
3. If data stack does not equal `caller_stack_height + type[current_section_index].outputs + type[code_section_index].inputs - type[code_section_index].outputs` items, execution results in exceptional halt.
```
plus additional requirement for JUMPF instruction validation:
```
-  `type[current_section_index].outputs >= type[code_section_index].inputs - type[code_section_index].outputs` 
```
^ this should replace current requirement `type[current_section_index].outputs == type[code_section_index].outputs` (equal number of outputs for JUMPF caller and callee)

EIPS/eip-3540.md
```diff
@@ -79,58 +81,91 @@ The container starts with the EOF header:
...
 
-1. `JUMPDEST`-analysis is only run on the *code*.
-2. Execution starts at the first byte of the *code*, and `PC` is set to 0.
-3. Execution stops if `PC` goes outside the code section bounds.
-4. `PC` returns the current position within the *code*.
-5. `JUMP`/`JUMPI` uses an absolute offset within the *code*.
-6. `CODECOPY`/`CODESIZE`/`EXTCODECOPY`/`EXTCODESIZE`/`EXTCODEHASH` keeps operating on the entire *container*.
-7. The input to `CREATE`/`CREATE2` is still the entire *container*.
-8. The size limit for deployed code as specified in [EIP-170](./eip-170.md) and for init code as specified in [EIP-3860](./eip-3860.md) is applied to the entire *container* size, not to the *code* size.
+1. Execution starts at the first byte of the first code section, and PC is set to 0.
```
shemnon, line 161: If we are removing the `PC` operation does it really matter what we call the first byte of the code section?

EIPS/eip-3540.md
```diff
@@ -79,58 +81,91 @@ The container starts with the EOF header:
...
+2. Execution stops if `PC` goes outside the code section bounds.
+3. `PC` returns the current position within the *code*.
```
shemnon, line 163: Wasn't removing `PC` one of the breakout room 4 decisions?

EIPS/eip-4750.md
```diff
@@ -32,23 +32,14 @@ Furthermore, it aims to improve analysis opportunities by encoding the number of
...
-6. The type section, if present, MUST directly precede all code sections.
-7. The type section, if present, contains a sequence of pairs of bytes: first byte in a pair encodes number of inputs, and second byte encodes number of outputs of the code section with the same index. *Note:* This implies that there is a limit of 256 stack for the input and in the output.
-8. Therefore, type section size MUST be `n * 2` bytes, where `n` is the number of code sections.
+4. New section called the *type section* is introduced, identified by `kind_type` (`0x01`).
+5. Exactly one type section MUST be present.
+6. The type section MUST directly precede all code sections.
+7. The type section, if present, contains a sequence of byte triples: first byte in the triple encodes number of inputs, and second byte encodes number of outputs of the code section with the same index, and third byte encodes `max_stack_height`, which is defined in [EIP-5450](./eip-5450.md). *Note:* This implies that there is a limit of 255 stack for the input and in the output. This is actually further restricted to 127 stack items because the upper bit of both the input and output bytes are reserved for future use.
```
shemnon, line 38: ``
and third byte encodes `max_stack_height`, which is defined in [EIP-5450](./eip-5450.md)
```

Is this leftover from the `ceil(stackheight /8)` changes?  Does it need to become a byte quad or is max stack height also limited to 127?

File: EIPS/eip-3540.md (hunk 19 above)
Comment (chfast) [reply EIPS/eip-3540.md:163]: Hm... I thought you said to leave it :)

File: EIPS/eip-4750.md (hunk 20 above)
Comment (lightclient) [reply EIPS/eip-4750.md]: fixed, thx - the stack height is not limited to 127, it is limited to 1024

File: EIPS/eip-4750.md (hunk 16 above)
Comment (lightclient) [reply EIPS/eip-4750.md]: These make sense, but per our conversation offline, I'm going to remove JUMPF from 4750 so that we can introduce it separately on top of 5450.

File: EIPS/eip-4750.md (hunk 17 above)
Comment (lightclient) [reply EIPS/eip-4750.md]: fixed

File: EIPS/eip-3540.md [hunk 21]
```diff
@@ -79,58 +81,91 @@ The container starts with the EOF header:
...
+The EOF header is followed by at least one section header. Each section header contains two fields, `section_kind` and either `section_size` or `section_size_list`, depending on the kind. `section_size_list` is a list of size values when multiple sections of this kind are allowed.
 
-| description  | length  | value         |                                                 |
-|--------------|---------|---------------|-------------------------------------------------|
-| section_kind | 1-byte  | 0x01–0xFF     | Encoded as a 8-bit unsigned number.             |
-| section_size | 2-bytes | 0x0001–0xFFFF | Encoded as a 16-bit unsigned big-endian number. |
+| description       | length  | value         |                   |
+|-------------------|---------|---------------|-------------------|
+| section_kind      | 1-byte  | 0x01–0xFF     | `uint8`           |
+| section_size      | 2-bytes | 0x0001–0xFFFF | `uint16`           |
```
gumb0, line 89: formatting nit
```suggestion
| section_size      | 2-bytes | 0x0001–0xFFFF | `uint16`          |
```

EIPS/eip-5450.md
```diff
@@ -29,150 +29,40 @@ In particular, this extended code validation eliminates the need for EVM stack u
...
-6. Maximum data stack height required by a function does not exceed `1024`.
+4. For `RETF` instruction, data stack before executing it has exactly `n` items to use as output values returned from a function, where `n` is function's number of outputs according to its type defined in the type section.
+5. `max_stack_height` does not exceed 1023.
+6. The maximum data stack height matches the corresponding code section's `max_stack_height` within the type section body.
+7. No unreachable instructions exist in the code section.
 
-To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. See below for reference implementation.
+To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. After completing this, verify each instruction was visited at least once. Fail if any instructions were not visited as this invalidates 9).
```
gumb0, line 44:
```suggestion
To examine every reachable code path, validation needs to traverse every instruction in order, while also following each non-conditional jump, and following both possible branches for each conditional jump. After completing this, verify each instruction was visited at least once. Fail if any instructions were not visited as this invalidates 7).
```

EIPS/eip-3540.md
```diff
@@ -79,58 +81,91 @@ The container starts with the EOF header:
...
+See [EIP-4750](./eip-4750.md) for more information on the type section content.
 
 #### EOF version 1 validation rules
 
 1. In addition to general validation rules above, EOF version 1 bytecode conforms to the rules specified below:
-   - Exactly one code section MUST be present.
-   - The code section MUST be the first section.
-   - A single data section MAY follow the code section.
-2. Any other version is invalid.
+   - One type section header MUST be present immediately following the EOF version. Each code section MUST have a specified type signature in the type body.
```
chfast, line 150:
```suggestion
   - Exactly one type section header MUST be present immediately following the EOF version. Each code section MUST have a specified type signature in the type body.
```

EIPS/eip-3540.md
```diff
@@ -79,58 +81,91 @@ The container starts with the EOF header:
...
+   - One code section header MUST be present immediately following the type section. A maxmimum of 1024 individual code sections are allowed.
```
chfast, line 151:
```suggestion
   - Exactly one code section header MUST be present immediately following the type section. A maxmimum of 1024 individual code sections are allowed.
```

EIPS/eip-3540.md
```diff
@@ -79,58 +81,91 @@ The container starts with the EOF header:
...
+   - One data section header MUST be present immediately following the code section.
```
chfast, line 152:
```suggestion
   - Exactly one data section header MUST be present immediately following the code section.
```

EIPS/eip-4750.md
```diff
@@ -96,25 +82,13 @@ If the code is valid EOF1, the following execution rules apply:
...
-6. Pushes nothing to return stack.
-7. Sets `current_section_index` to `code_section_index` and `PC` to `0`, and execution continues in the called section.
-
 #### `RETF`
 
 1. Does not have immediate arguments.
-2. If data stack has less than `caller_stack_height + type[code_section_index].outputs` items, execution results in exceptional halt.
-3. If data stack has more than `caller_stack_height + type[code_section_index].outputs` items, discards the items between `caller_stack_height` and top `type[code_section_index].outputs` items, so that there are exactly `caller_stack_height +  type[code_section_index].outputs` items left.
+2. If data stack does not equal `caller_stack_height + type[current_section_index].outputs` items, execution results in exceptional halt.
 4. Charges 4 gas.
```
chfast, line 92: As stack cleanup is removed, restore previous gas cost of 3?

EIPS/eip-5450.md
```diff
@@ -29,150 +29,40 @@ In particular, this extended code validation eliminates the need for EVM stack u
...
-
-This can be used for implementing more efficient early exits from a function (e.g. assertion failure).
-
-For "exit" instructions which terminates the whole program execution (`STOP`, `RETURN`, etc) this is no change comparing to pre-EOF EVM. I.e. some *garbage* can be left on the stack. Cleaning the stack does not improve EVM implementation performance but makes the EVM programs potentially cost more (compiler is required to insert additional `POP` instructions).
-
-For `RETF` semantic would be more complicated. For `n` function outputs and `s` the stack height at `RETF` the EVM must erase `s-n` non-top stack items and move the `n` stack items to the place of erased ones. Cost of such operation may be relatively cheap but is not constant.
+In this EIP, we provide a more efficient variant of the EVM where stack overflow check is performed only in `CALLF` instruction using the called function's `max_stack_height` information. This decreases flexibility of an EVM program because `max_stack_height` corresponds to the worst-case control-flow path in the function.
```
chfast, line 58: This is not fully done. E.g. the `CALLF` would need to check `max_stack_height` at runtime.

EIPS/eip-4750.md
lightclient:
```suggestion
4. Charges 3 gas.
```

EIPS/eip-5450.md
lightclient: Does something need to change here? Or you're referring to the outstanding PR to 5450 that defines this behavior?

chfast: How `CALLF` operates, but I will fix it later.

chfast (approved):

EIPS/eip-3540.md
shemnon: I was persuaded by the lack of dynamic jumps making it useless.

lightclient: Re-removed it :) https://github.com/ethereum/EIPs/pull/6209

PR #6161, opened 2022-12-17
Update EIP-1: Use updated list of EL clients

Pandapip1:
This list is very out of date. And includes external links that we can get rid of.

CC @SamWilsn


SamWilsn (approved):

hamzawajahat (commented):
changes

PR #6167, opened 2022-12-17
Render csl-json blocks as citations

SamWilsn:
The commit uses [Citation.js](https://citation.js.org/) to render the citation blobs from EIP-1 in a human-readable format.

**You can see a preview [here](https://samwilsn.github.io/EIPs/EIPS/eip-1#fn:1).**


Pandapip1 (changes requested):
Is it possible to do this server-side? I could maybe write a post-processing step or something. I really would like to limit what needs JS (MathJax is already not ideal IMO).

SamWilsn:
> Is it possible to do this server-side?

That is also my preferred implementation. There's [`citeproc-ruby`](https://github.com/inukshuk/citeproc-ruby) (and I believe they have a Jekyll plugin), but my understanding is that since we use the Pages build (and not an action), we're limited in what we can do.

Pandapip1:
> since we use the Pages build (and not an action), we're limited in what we can do.

It might be time to fix that. Should I make a PR?

SamWilsn:
I'm indifferent. I don't feel strongly enough about server-side vs. client-side rendering to care much. If you want to go ahead, be my guest!

I don't know Ruby as well as I know JS/TS, so I might fumble around getting this particular feature to work though.

Pandapip1 (approved):
I trust you know what you're doing and that this works

PR #6170, opened 2022-12-19
feat: creating the draft for cross-chain messaging

sujithsomraaj:
interface for standardizing multiple cross-chain messaging protocols


sujithsomraaj:
Should I rename it to EIP6170 ?

PR #6185, opened 2022-12-20
Update EIP-5615: Move to Last Call

Pandapip1:


EIPS/eip-5615.md
```diff
@@ -4,7 +4,8 @@ title: EIP-1155 Supply Extension
 description: A simple mechanism to fetch token supply data from EIP-1155 tokens
 author: Pandapip1 (@Pandapip1)
 discussions-to: https://ethereum-magicians.org/t/eip-5615-eip-1155-supply-extension/10732
-status: Review
+status: Last Call
+last-call-deadline: 2023-01-01
```
SamWilsn, line 8:
```suggestion
last-call-deadline: 2023-01-10
```

SamWilsn (commented):
Your security considerations is still "Needs discussion." That'll need to be filled out (even with just "none") before going into last call.

Pandapip1:
Done (this branch isn't updated)

EIPS/eip-5615.md
```diff
@@ -4,10 +4,11 @@ title: EIP-1155 Supply Extension
...
 type: Standards Track
 category: ERC
-created: 2022-09-07
+created: 2023-03-08
```
SamWilsn, line 11: The `created` field should be immutable.

SamWilsn (changes requested):

EIPS/eip-5615.md
Pandapip1: Oops, I think I updated the wrong field lol.

Pandapip1: Also, the created field shouldn't be in the future.

Pandapip1:
@SamWilsn fixed

Pandapip1:
@eth-bot rerun

SamWilsn (dismissed):
> Whether the given token id exists, previously existed, or may exist

This is ambiguous. Should `exists` return true or false in each of these cases:

 - Token cannot exist (unambiguously false?)
 - Token exists (unambiguously true?)
 - Token could exist, but has never been minted.
 - Token has existed, but has been destroyed.

---

I'd reword and move some bits around:

Move to specification:

> Implementations MAY support [ERC-165](./eip-165.md) interface discovery with an id of `0x...`, but consumers MUST NOT rely on it.

Move to rationale:

> This EIP does not mandate [ERC-165](./eip-165.md) support as this interface is simple enough that the extra complexity is unnecessary and would cause incompatibilities with pre-existing implementations.

Pandapip1:
> This is ambiguous. Should exists return true or false in each of these cases:
> 
> Token cannot exist (unambiguously false?)
> Token exists (unambiguously true?)
> Token could exist, but has never been minted.
> Token has existed, but has been destroyed.

False.
True.
True.
True.

It's not ambiguous. It's a binary OR of all three options.

EIPS/eip-5615.md
```diff
@@ -4,7 +4,8 @@ title: EIP-1155 Supply Extension
 description: A simple mechanism to fetch token supply data from EIP-1155 tokens
 author: Pandapip1 (@Pandapip1)
 discussions-to: https://ethereum-magicians.org/t/eip-5615-eip-1155-supply-extension/10732
-status: Review
+status: Last Call
+last-call-deadline: 2023-03-08
```
xinbenlv, line 8:
```suggestion
last-call-deadline: 2023-04-27
```

xinbenlv (approved):
I read through it and believe there is no ambiguity about when to return zero.

Approved and please don't forget to update the deadline

PR #6188, opened 2022-12-20
Add EIP-6188: Nonce Cap

Pandapip1:
The first part of the epic SELFDESTRUCT suite

See also: #6189 , #6190


xinbenlv (commented):
It seems a good idea for this direction of capping nonce. 
I wonder if we shall be even more aggressive by choose capping it at sufficient bit e.g. `nonce < 2^128-1` so that it will have another 128bits large enough for almost any useage

Pandapip1:
> I wonder if we shall be even more aggressive by choose capping it at sufficient bit e.g. nonce < 2^128-1 so that it will have another 128bits large enough for almost any useage

That might be a good idea. I was thinking of the functionality being identified by the contract code (see #6189 for what I mean), but it might make sense to go big.

EIPS/eip-6188.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Motivation
+
+This EIP is not terribly useful on its own, as it adds additional computation without any useful side effects. However, it can be used by other EIPs.
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### `CREATE` and `CREATE2`
```
SamWilsn, line 26: This heading seems to imply that EOA transactions can still use `2^64-1`. Is that correct?

EIPS/eip-6188.md
```diff
@@ -0,0 +1,44 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2022-12-20
+requires: 2929
+---
+
+## Abstract
+
+This EIP reserves a nonce for special contracts. It also caps the nonce at `2^64-2`.
```
SamWilsn, line 16: What is a special contract? Based on the title, it sounds like the primary purpose of the EIP is to cap the nonce (so using "It also" seems weird.)

EIPS/eip-6188.md
Pandapip1: Nope - thanks for pointing that out!

EIPS/eip-6188.md
Pandapip1: I'll rephrase for clarity.

EIPS/eip-6188.md
```diff
@@ -0,0 +1,48 @@
...
+
+This EIP is not terribly useful on its own, as it adds additional computation without any useful side effects. However, it can be used by other EIPs.
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### EOA Transactions
+
+The nonce of a transaction originating from an EOA MUST be less than `2^64-2`. If the nonce is `2^64-1`, the transaction MUST be invalid.
```
SamWilsn, line 28:
```suggestion
The nonce of a transaction originating from an EOA MUST be less than `2^64-2`. If the nonce is either `2^64-1` or `2^64-2`, the transaction MUST be invalid.
```

EIPS/eip-6188.md
```diff
@@ -0,0 +1,48 @@
...
+
+The nonce of a transaction originating from an EOA MUST be less than `2^64-2`. If the nonce is either `2^64-1` or `2^64-2`, the transaction MUST be invalid.
+
+### `CREATE` and `CREATE2`
+
+If a nonce would be incremented to `2^64-1` by `CREATE` or `CREATE2`, it is instead set to `2^64-2`. `2^64-1` is reserved for alias or other special contracts.
+
+## Rationale
+
+Capping a nonce allows for contracts with special properties to be created, with their functionality based on their contract code. This EIP is not terribly useful on its own, but it can be used by other EIPs.
```
SamWilsn, line 36: This reads like motivation, not rationale.

EIPS/eip-6188.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Backwards Compatibility
+
+This EIP requires a protocol upgrade, since it modifies consensus rules. The further restriction of nonce should not have an effect on accounts, as reaching a nonce of `2^64-2` is unfeasible.
+
+## Security Considerations
+
+None.
```
SamWilsn, line 44: If a contract uses `CREATE`, and it somehow was at the nonce cap, and the target address had self-destructed, would further CREATEs put the deployed contract in an unexpected location (i.e. back at the same address instead of at a different address)? Should CREATE fail in that scenario instead? Probably worth mentioning, even if the solution is just "no contract will get to that nonce."

Pandapip1:
@SamWilsn have incorporated your suggestions.

SamWilsn (approved):

PR #6232, opened 2022-12-28
Explained typesize in the v.1 header

qbzzt:
Also, fixed the minimum (0003 -> 0004) and maximum (FFFF -> FFFC) because it has to be divisible by four


chfast (approved):

Jillgrand (commented):
Approve

PR #6237, opened 2022-12-29
Test cases: specify when contract creation fails in Valid initcode

hugo-dc:
This makes explicit when contract creation fails even if initcode is valid.


BigBro17on25tolife:
Thanks for letting me know

gumb0 (approved):

PR #6238, opened 2022-12-29
clarify signed blob transaction hashing

roberto-bayardo:


EIPS/eip-4844.md
```diff
@@ -174,6 +174,13 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
     return ecrecover(tx_hash(tx), int(sig.y_parity)+27, sig.r, sig.s)
 ```
 
+The transaction hash of a signed blob transaction should be computed per [`EIP-2718`](./eip-2718.md):
+
+```python
+def signed_tx_hash(tx: SignedBlobTransaction) -> Bytes32:
+    return keccak256(BLOB_TX_TYPE + ssz.serialize(tx))
+```
```
MicahZoltu, line 182:
````suggestion
```
````

File: EIPS/eip-4844.md [hunk 2]
```diff
@@ -174,6 +174,13 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
     return ecrecover(tx_hash(tx), int(sig.y_parity)+27, sig.r, sig.s)
 ```
 
+The transaction hash of a signed blob transaction should be computed per [`EIP-2718`](./eip-2718.md):
```
MicahZoltu, line 177: EIP-2718 doesn't specify how transaction hashes are calculated.  I *believe* that transaction hashes are the `keccak256` of the bytes in the transaction trie whose root is in the block header, so if the below aligns with that then it is correct.
```suggestion
The transaction hash of a signed blob transaction should be computed as:
```

roberto-bayardo:
See also https://github.com/ethereum/EIPs/pull/6241 which I am keeping separate from this since it's a breaking change rather than a clarification.

protolambda:
> See also https://github.com/ethereum/EIPs/pull/6241 which I am keeping separate from this since it's a breaking change rather than a clarification.

Left a comment there about RLP<>SSZ and hashing behavior which I think is relevant to both PRs.

roberto-bayardo:
It is possible to get these PRs merged by the sync tomorrow @protolambda ?  The code changes to the 4844-geth are already subnmitted.

protolambda (approved):

PR #6241, opened 2022-12-30
change blob tx_hash to use serialize instead of hash_tree_root

roberto-bayardo:
this makes the tx hashing for the purposes of tx signing more consistent with computing the hash of a signed blob transaction, and should be more efficient.  Sister PR to https://github.com/ethereum/EIPs/pull/6238

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -164,17 +164,17 @@ The execution layer verifies the wrapper validity against the inner `Transaction
 The signature is verified and `tx.origin` is calculated as follows:
 
 ```python
-def tx_hash(tx: SignedBlobTransaction) -> Bytes32:
+def unsigned_tx_hash(tx: SignedBlobTransaction) -> Bytes32:
     # The pre-image is prefixed with the transaction-type to avoid hash collisions with other tx hashers and types
-    return keccak256(BLOB_TX_TYPE + ssz.hash_tree_root(tx.message))
+    return keccak256(BLOB_TX_TYPE + ssz.serialize(tx.message))
 
 def get_origin(tx: SignedBlobTransaction) -> Address:
     sig = tx.signature
     # v = int(y_parity) + 27, same as EIP-1559
-    return ecrecover(tx_hash(tx), int(sig.y_parity)+27, sig.r, sig.s)
+    return ecrecover(unsigned_tx_hash(tx), int(sig.y_parity)+27, sig.r, sig.s)
 ```
 
-The transaction hash of a signed blob transaction should be computed as:
+The hash of a signed blob transaction should be computed as:
 
 ```python
 def signed_tx_hash(tx: SignedBlobTransaction) -> Bytes32:


protolambda (commented):
If we're not using the hash-tree-root we may want to consider just changing the EIP to use RLP encoding for the transaction, and updating the consensus function that extracts the data-hashes.

Also, we're not including the signature values in the tx-hash, unlike other tx types, so it's probably good to open some discussion about behavior of encoding & tx hashing & signing-hash, and document our choices in the EIP.

Little of the remaining benefits of using SSZ remain.

Pro-SSZ:
- align with consensus-layer
- hash-tree-root can be used to proof smaller contents of the tx efficiently with a neat binary merkle proof, enabling new things that were too inefficient with RLP hashes.
  - use-cases are unidentified
  - hash-tree-root is not actually used for the full TX hash currently.
  - hash-tree-root is not used for the signing-hash either. 
  - the RLP-encoded tx would already get merkleized as pain bytes as part of the TX list in the ExecutionPayload, which may be good enough for most use-cases once we expose the beacon-block-root to the EVM.
- in the consensus-layer beacon-blocks work like this: these are also identified by their content, and the roots don't include the signature. Gossip uses flat-hashes (with special case to deal with compression variants etc.) of the payload to identify and attribute things when validating for propagation purposes. But signatures will have to be stored and included in sync to validate the chain still, so there's little gain in not including them.
- If we ever lift the TX type into the consensus-layer we prefer SSZ. But current EIP already kind of makes it incompatible.

Pro-RLP:
- align with execution-layer (arguably more relevant here)
- simple: hash-tree-root is more complex and error-prone
- tooling that is built around RLP will more easily adopt the tx type
- more efficient: SSZ offsets, but come with an overhead.
  - fun coincidental fact about SSZ inefficiency here: with the offset to the dynamic part of the tx (4 bytes), and the signature (32+32+1=65 bytes), the first offset is always 69, i.e. all blob transactions start with 5 (type), 69, 0, 0, 0.
- no surprises around tx identificiation:
  - tx-pools that assume the signature validity is attributable to the tx with the specific tx-hash won't break (!!!).
  - tools like explorers will have an easier time to include the tx.
- faster, if we use RLP-serialize + hash with prefix compared to SSZ hash-tree-root + hash with prefix.

SSZ tx typing was included in some original danksharding proposals by @vbuterin and @dankrad so I may be missing some context, but I think it's worth fully switching to RLP encoding, with tx-hashing with signature, and signing-hash with prefixed RLP without signature like the other TX types.

roberto-bayardo:
`Also, we're not including the signature values in the tx-hash, unlike other tx types, so it's probably good to open some discussion about behavior of encoding & tx hashing & signing-hash, and document our choices in the EIP.`

@protolambda re: not including signature values in the tx-hash, I thought that was a bug and it's why I suggested the clarification in  https://github.com/ethereum/EIPs/pull/6238 which specifies including the full blob tx (message+ signature). Wouldn't transaction malleability be a concern otherwise?

roberto-bayardo:
P.S.: I am in favor of keeping the SSZ encoding even though it has proven a bit painful in the EL where RLP assumptions can be baked in pretty deeply. EIP-2718 seemed as if it would drafted to allow exactly this kind of migration to a better encoding scheme.

roberto-bayardo:
As for this PR specifically, I'm inclined to retract given that the signing hash in the consensus layer for SSZ-encoded data always seems to use hash_tree_root.  I'd rather not deviate from that pattern for for the sake of being more consistent with how we compute the regular (non-signing) transaction hash.

protolambda:
I think it makes sense to use either `hash_tree_root` (HTR) for both tx-hash and signing-hash, or for neither.

Keeping SSZ makes sense if we prefer the new encoding, fix the issues with hashing, and clarify in the EIP why we choose HTR or not. And we can open a separate PR to include the signature values as part of the TX hash.

roberto-bayardo:
Let's queue this up for discussion at the next client-dev meeting? cc: @timbeiko

EIPS/eip-4844.md
```diff
@@ -166,7 +166,7 @@ The signature is verified and `tx.origin` is calculated as follows:
 ```python
 def tx_hash(tx: SignedBlobTransaction) -> Bytes32:
```
protolambda, line 167: Change looks good, and don't mean to bike-shed, but I think we should clarify the function naming while we're at it.

We should clarify that this outputs the hash to sign & use for verification, not the full tx hash, which would be easy to mistake when it's called `tx_hash`.

And the other method, "signed_tx_hash", is confusing with CL specs context (we use "signing_root" to describe the hash that is used for signing, and thus doesn't include the signature values yet).

Maybe we can be explicit and say `unsigned_tx_hash` (here) and `signed_tx_hash` (the other PR, already approved)?

roberto-bayardo: Agree .. will clarify.

roberto-bayardo: @protolambda what about "signing_root_hash" for this one and "tx_id_hash" for the other?

protolambda: it's not a root since it's not a tree commitment anymore, and it's not really an ID either (many transaction managers use "ID" to track transactions that may be resubmitted with different fee parameters and thus different hash). I'd just keep it simple and do `{signed, unsigned}_tx_hash`

protolambda: Would also be fine with `signing_hash` and `tx_hash` if you prefer

roberto-bayardo: changed to {signed,unsigned}_tx_hash

protolambda (approved):

PR #6254, opened 2023-01-03
Attempt to fix the page width issue

SamWilsn:
[live](https://samwilsn.github.io/EIPs/)

image


Pandapip1:
Why don't we use 100% content width?

PR #6256, opened 2023-01-04
Website: Change content width to 100%

Pandapip1:
This is an aesthetical choice. I personally would prefer 100% width.

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -2,7 +2,7 @@
 # Only the main Sass file needs front matter (the dashes are enough)
 ---
 
-$content-width: 960px;
+$content-width: 100%;
 
 @import 'minima';
 


SamWilsn:
Hm, I didn't realize my change was contentious :rofl:

In https://github.com/ethereum/EIPs/pull/6231, @aviggiano reported that on ultrawide monitors the text becomes very annoying to read. I'd be fine to change this so that it only limits the width above a certain resolution.

aviggiano:
I agree that I should have limited my change to ultra wide monitors, not to all display sizes. Can we include a media query so that other screen sizes are not affected?

Pandapip1:
I moved this change in the preamble PR (as it causes minor issues w/ formatting), which is why I closed this PR. How about 1152px (1 foot)? This should be enough space that those issues don't crop up.

PR #6260, opened 2023-01-05
Add EIP-6260: Buyable NFT tokens on-Chain and Royalties

lth-elm:
This EIP allows a non-fungible token to become buyable and to specifically enforce royalties as a percentage, directly on-chain, without entrusting a third-party. A decentralized marketplace can be built around this interface.


lth-elm:
Hi @axic, @SamWilsn, @Pandapip1 

I can't seem to understand why EIP walidator fail on the LICENSE url check while it does works perfectly fine when we check [here](https://github.com/ethereum/EIPs/pull/6260/files?short_path=c1223af#diff-c1223af4552387f9ae202241ee29a4afcb6b15d73f5ab60250ba6aa604bf5e25). 

Any idea ?

numtel:
Can you compare this to #6105 ?

lth-elm:
> Can you compare this to #6105 ?

I like this idea, could really change the game if it's adopted by most marketplaces unfortunately some will always continue to do it their way.

The key difference here is that the "marketplace" functions is directly within the nft smart-contract, the whole idea is to be able to sell your nfts without having to rely on any marketplace.

numtel:
> The key difference here is that the "marketplace" functions is directly within the nft smart-contract, the whole idea is to be able to sell your nfts without having to rely on any marketplace.

That's exactly what #6105 does. The extension interface you propose is almost exactly the same:

* `setPrice` -> `listItem`
* `removeTokenSale` -> `delistItem`
* `buyToken` -> `buyItem`

Also, why are you implementing a new royalty calculation format that is less extensible than EIP-2981?

lth-elm:
Indeed, I had come accross one of your response and misinterpret the goal of the EIP. It does actually share the same idea. 

Since nothing has been published on the ethereum-magicians forum I haven't seen any discussions associated with this kind of topic. However, I notice that you've been sharing quite divided thoughts under the pull request. If I can bring some context as to why we thought of this EIP I invite you to read [this message](https://ethereum-magicians.org/t/eip-6260-buyable-nft-tokens-on-chain-and-royalties/12443/2?u=lth-elm) posted on the forum. We've also provided materials under the _assets_ folder.

As for the integration of the EIP-2981, we've thought about it but unfortunately it does not allow to update the % of royalties while we believe that it would be relevant to enable it, but only downwards.

lenifoti:
> As for the integration of the EIP-2981, we've thought about it but unfortunately it does not allow to update the % of royalties while we believe that it would be relevant to enable it, but only downwards.

Eip2981 did not allow the royalty to be changed. That was not considered a good thing to do after an NFT had been minted because it does not seem a fair thing to do to an owner of an NFT. It is effectively unilaterally modifying the contract.
So setting the royalty should happen before the first sale. That is something that could be added at mint time perhaps.
As for having a royalty that changed according to a schedule or value etc. That is precisely why 2981 does not return a percentage. The discussion was long and painful and lasted a year.
I would have less of a concern if the royalty info function in this spec had the same semantics as 2981. At least then it allows the implementation to support both models.
But I do believe that that this is equivalent to the earlier eip and it makes more sense to extend it to support any functions that you feel it is lacking rather than create an incomparable standard.

lenifoti:
> Since nothing has been published on the ethereum-magicians forum I haven't seen any discussions associated with this kind of topic. However, I notice that you've been sharing quite divided thoughts under the pull request. 
I also find the comments process rather confusing. Few proposals are discussed in the wizards forum I find. But watching the eip repo, which is the only sure way of catching everything is also far from ideal as it generates a lot of noise. Not sure what the solution is but this is a real problem.

5660-eth:
> > As for the integration of the EIP-2981, we've thought about it but unfortunately it does not allow to update the % of royalties while we believe that it would be relevant to enable it, but only downwards.
> 
> Eip2981 did not allow the royalty to be changed. That was not considered a good thing to do after an NFT had been minted because it does not seem a fair thing to do to an owner of an NFT. It is effectively unilaterally modifying the contract. So setting the royalty should happen before the first sale. That is something that could be added at mint time perhaps. As for having a royalty that changed according to a schedule or value etc. That is precisely why 2981 does not return a percentage. The discussion was long and painful and lasted a year. I would have less of a concern if the royalty info function in this spec had the same semantics as 2981. At least then it allows the implementation to support both models. But I do believe that that this is equivalent to the earlier eip and it makes more sense to extend it to support any functions that you feel it is lacking rather than create an incomparable standard.

The method to change the royalty is already given in the following reference implementation.This is why we do not give the royalty changing method in [EIP6105](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6105.md).
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/common/ERC2981.sol
 `function _setDefaultRoyalty(address receiver, uint96 feeNumerator)`

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-05-17
+requires: 721
+---
+
+## Abstract
+
+This EIP allows an [EIP-721](./eip-721.md) token to become buyable and to specifically enforce royalties as a percentage, directly on-chain, without entrusting a third-party. A decentralized marketplace can be built around this interface.
```
SamWilsn, line 16: Since collecting royalties on-chain is... difficult, you should include a terse technical summary of _how_ this proposal achieves that goal.

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+     *
+     * @param _newRoyalty uint within the range of `_royaltyDenominator` as new tokens royalties.
+     */
+    function setRoyalty(uint _newRoyalty) external;
+}
+```
+
+To create and deploy this EIP contract one **MAY** only inherit directly from this [EIP-721](./eip-721.md) extension that would directly signal support for this EIP:
+
+```javascript
```
SamWilsn, line 148:
````suggestion
```solidity
````

File: EIPS/eip-6260.md [hunk 3]
```diff
@@ -0,0 +1,290 @@
...
+import "./ERC721Buyable.sol";
+
+contract NFTContract is ERC721Buyable {
+    // ...
+    constructor() ERC721("NFTBuyableContract", "NFTBC") { }
+    // ...
+}
+```
+
+Checking if a contract implement this specification can be done with:
```
SamWilsn, line 159:
```suggestion
Checking if a contract implements this specification can be done with:
```

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+
+```javascript
```
SamWilsn, line 161:
````suggestion
```solidity
````

File: EIPS/eip-6260.md [hunk 5]
```diff
@@ -0,0 +1,290 @@
...
+pragma solidity ^0.8.0;
+import "./ERC721Buyable.sol";
+
+contract NFTContract is ERC721Buyable {
+    // ...
+    constructor() ERC721("NFTBuyableContract", "NFTBC") { }
+    // ...
+}
+```
```
SamWilsn, line 157: This whole block is an implementation detail, and shouldn't be in your proposal. Implementers are free to choose how they make a compatible contract.

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+eip: 6260
+title: Buyable NFT tokens on-Chain & Royalties
+description: Allowing tokens to be buyable and enforce royalties directly on-Chain
+author: Aubay <blockchain-team@aubay.com>, BOCA Jeabby (@bjeabby1507), EL MERSHATI Laith (@lth-elm), KEMP Elia (@eliakemp)
+discussions-to: https://ethereum-magicians.org/t/eip-6260-buyable-nft-tokens-on-chain-and-royalties/12443
+status: Draft
+type: Standards Track
+category: ERC
+created: 2022-05-17
+requires: 721
```
SamWilsn, line 11:
```suggestion
requires: 165, 721
```

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// OR
+
+bytes4 private constant _INTERFACE_ID_ERC721Buyable = 0x8ce7e09d;
+
+function isIDERCBuyable(address _contractAddr) internal view returns(bool) {
+    (bool success) = IERC165(_contractAddr).supportsInterface(_INTERFACE_ID_ERC721Buyable);
+    return success;
+}
```
SamWilsn, line 176: You can just refer people to ERC-165 instead of writing out how to use it here.

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+
+### Functions
+
+#### **supportsInterface**
+
+Returns true if a contract implements the interface defined by `interfaceId`.
+
+| Parameter   | Description                                              |
+| ----------- | -------------------------------------------------------- |
+| interfaceId | bytes4 representing the support of a contract interface. |
```
SamWilsn, line 187: Just refer people to 165.

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+
+#### **setPrice**
```
SamWilsn, line 189: Use backticks for code in headings:

```suggestion
#### `setPrice`
```

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+
+- Royalties matter
+
+Regarding the royalties payment, NFT creators rely on marketplaces like OpenSea & Rarible to collect it when their creations are sold. Over time, we discussed the idea of making this entire process completely decentralised. With such solution, artists would be allowed to 100% enforce their due royalties without taking the risk of trusting a marketplace to pay them directly nor having to specify to each marketplace one by one the % they want to collect while being constrained to be capped. Our solution by being able to directly sell a token onchain makes it easy to do just that.
```
SamWilsn, line 254: These paragraphs are more suited to the Motivation section. The Rationale section should explain individual technical choices made within the proposal itself, while the Motivation section justifies the entire EIP as a whole.

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+
+- Extend compatibility
+
+If this proposal gets approved, we can definitely create an extension for both [EIP-1155](./eip-1155.md) and the Azuki model which is becoming more and more popular among developers as they begin to adopt this new form of standard.
+
+Other aspects could also be integrated such as the support for auctions and/or extension with an operator filter registry model like the Opensea one.
+
+## Backwards Compatibility
+
+There are no backward compatibility issues, this implementation is an extension of the functionality of [EIP-721](./eip-721.md). This EIP is fully backward compatible and introduces new functionality retaining the core interfaces and functionality of the [EIP-721 standard](./eip-721.md).
```
SamWilsn, line 274:
```suggestion
There are no backward compatibility issues, this implementation is an extension of the functionality of [ERC-721](./eip-721.md). This EIP is fully backward compatible and introduces new functionality retaining the core interfaces and functionality of the [ERC-721](./eip-721.md).
```

EIPS/eip-6260.md
```diff
@@ -0,0 +1,290 @@
...
+
+Tests can be found in [buyableTokenTest.js](../assets/eip-6260/test/buyableTokenTest.js).
+
+## Reference Implementation
+
+Contract implementation can be found under [contracts/](../assets/eip-6260/contracts/NFTContract.sol).
+
+## Security Considerations
+
+There are no security considerations related directly to the implementation of this standard. Discussion with reviewers can still be found at https://ethereum-magicians.org/t/eip-6260-buyable-nft-tokens-on-chain-and-royalties/12443/7.
```
SamWilsn, line 286:
```suggestion
There are no security considerations related directly to the implementation of this standard.
```

SamWilsn:
Please move technical discussion to the [`discussions-to`](https://ethereum-magicians.org/t/eip-6260-buyable-nft-tokens-on-chain-and-royalties/12443) thread. Everything here will get hidden when this PR is merged.

PR #6264, opened 2023-01-05
Update EIP-2771: Move to Last Call

Pandapip1:


SamWilsn (approved):
On 77/78, "A Gas limit" should be "A gas limit".

PR #6269, opened 2023-01-06
Add EIP-6269: Full EVM Equivalence

pcaversaccio:
This meta EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is the **complete** compliance with the latest version of the [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/79db92bd4c49bdf5f559fdebea5a9a3a2bf3fb83). This definition applies retroactively since the first release of the [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf) on 2 April 2014 ([first version](https://github.com/ethereum/yellowpaper/blob/0d0d23301d077bbdab5cafae6ab06001e282fae2/Paper.tex)).

### DIFF
--- a/EIPS/eip-6269.md
+++ b/EIPS/eip-6269.md
@@ -26,7 +26,7 @@ Any protocol, network, smart contract system, or similar that claims at time `t`
 
 ## Rationale
 
-A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence. The [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/79db92bd4c49bdf5f559fdebea5a9a3a2bf3fb83) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
+A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence benchmark. The [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/79db92bd4c49bdf5f559fdebea5a9a3a2bf3fb83) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
 
 Prior to the Bellatrix hard fork on 6 September 2022, the [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf) was considered more complete and is therefore chosen as the main reference point.
 
--- a/EIPS/eip-6269.md
+++ b/EIPS/eip-6269.md
@@ -3,7 +3,7 @@ eip: 6269
 title: Full EVM Equivalence
 description: Canonicalise the definition of Full EVM Equivalence
 author: Pascal Caversaccio (@pcaversaccio)
-discussions-to: https://ethereum-magicians.org/t/evm-equivalence-and-ethereum-stack-compatibility-definition-future-informational-eip/10044
+discussions-to: https://ethereum-magicians.org/t/evm-equivalence-eip-6269-and-ethereum-stack-compatibility-definition/10044
 status: Draft
 type: Meta
 created: 2023-01-06
@@ -11,7 +11,7 @@ created: 2023-01-06
 
 ## Abstract
 
-This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Execution Layer Specification](https://github.com/ethereum/execution-specs/tree/24de2192e02bba11e61746aa9dee04d4e7a8b62e). This definition applies retroactively from the first release of the Ethereum Yellow Paper in April 2014.
+This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/master). This definition applies retroactively from the first release of the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) in April 2014.
 
 ## Motivation
 
@@ -19,33 +19,33 @@ In light of the recent zkEVM announcements by various projects and the ongoing d
 
 ## Specification
 
-Any protocol, network, smart contract system, or similar that claims at time `t` full EVM equivalence MUST fully (=100%) comply with the latest [Execution Layer Specification](https://github.com/ethereum/execution-specs/tree/24de2192e02bba11e61746aa9dee04d4e7a8b62e) version at time `t`.
+Any protocol, network, smart contract system, or similar that claims at time `t` full EVM equivalence MUST fully (=100%) comply with the latest [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/master) version at time `t`.
 
 ## Rationale
 
-A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence. The [Execution Layer Specification](https://github.com/ethereum/execution-specs/tree/24de2192e02bba11e61746aa9dee04d4e7a8b62e) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
+A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence. The [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/master) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
 
 ## Backwards Compatibility
 
-This proposal is fully backward compatible until the first release of the Ethereum Yellow Paper in April 2014.
+This proposal is fully backward compatible until the first release of the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) in April 2014.
 
 ## Test Cases
 
 ### Example 1 ✅
 
-_Assumption:_ A protocol `X` follows at time `t` all the specifications stated in the Ethereum Yellow Paper at time `t`.
+_Assumption:_ A protocol `X` follows at time `t` all the specifications stated in the Ethereum Execution Client Specifications at time `t`.
 
 > Protocol `X` is fully EVM equivalent.
 
 ### Example 2 ❌
 
-_Assumption:_ A protocol `X` follows at time `t` all the specifications stated in the Ethereum Yellow Paper at time `t` _except_ using the similar pricing for zero and non-zero byte of data.
+_Assumption:_ A protocol `X` follows at time `t` all the specifications stated in the Ethereum Execution Client Specifications at time `t` _except_ using the similar pricing for zero and non-zero byte of data.
 
 > Protocol `X` is **not** fully EVM equivalent.
 
 ### Example 3 ❌
 
-_Assumption:_ A protocol `X` follows at time `t` all the specifications stated in the Ethereum Yellow Paper at time `t` _except_ that the opcode `SELFDESTRUCT` does not destroy any storage.
+_Assumption:_ A protocol `X` follows at time `t` all the specifications stated in the Ethereum Execution Client Specifications at time `t` _except_ that the opcode `SELFDESTRUCT` does not destroy any storage.
 
 > Protocol `X` is **not** fully EVM equivalent.
 
--- a/EIPS/eip-6269.md
+++ b/EIPS/eip-6269.md
@@ -5,7 +5,7 @@ description: Canonicalise the definition of Full EVM Equivalence
 author: Pascal Caversaccio (@pcaversaccio)
 discussions-to: https://ethereum-magicians.org/t/evm-equivalence-eip-6269-and-ethereum-stack-compatibility-definition/10044
 status: Draft
-type: Meta
+type: Informational
 created: 2023-01-06
 requires: 5757
 ---
--- a/EIPS/eip-6269.md
+++ b/EIPS/eip-6269.md
@@ -11,7 +11,7 @@ created: 2023-01-06
 
 ## Abstract
 
-This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf). This definition applies retroactively from the first release of the Ethereum Yellow Paper in April 2014.
+This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Execution Layer Specification](https://github.com/ethereum/execution-specs/tree/24de2192e02bba11e61746aa9dee04d4e7a8b62e). This definition applies retroactively from the first release of the Ethereum Yellow Paper in April 2014.
 
 ## Motivation
 
@@ -19,11 +19,11 @@ In light of the recent zkEVM announcements by various projects and the ongoing d
 
 ## Specification
 
-Any protocol, network, smart contract system, or similar that claims at time `t` full EVM equivalence MUST fully (=100%) comply with the latest [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf) version at time `t`.
+Any protocol, network, smart contract system, or similar that claims at time `t` full EVM equivalence MUST fully (=100%) comply with the latest [Execution Layer Specification](https://github.com/ethereum/execution-specs/tree/24de2192e02bba11e61746aa9dee04d4e7a8b62e) version at time `t`.
 
 ## Rationale
 
-A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence. The [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
+A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence. The [Execution Layer Specification](https://github.com/ethereum/execution-specs/tree/24de2192e02bba11e61746aa9dee04d4e7a8b62e) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-6269.md
+++ b/EIPS/eip-6269.md
@@ -1,5 +1,5 @@
 ---
-eip:
+eip: 6269
 title: Full EVM Equivalence
 description: Canonicalise the definition of Full EVM Equivalence
 author: Pascal Caversaccio (@pcaversaccio)
@@ -11,7 +11,7 @@ created: 2023-01-06
 
 ## Abstract
 
-This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper). This definition applies retroactively from the first release of the Ethereum Yellow Paper in April 2014.
+This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf). This definition applies retroactively from the first release of the Ethereum Yellow Paper in April 2014.
 
 ## Motivation
 
@@ -19,11 +19,11 @@ In light of the recent zkEVM announcements by various projects and the ongoing d
 
 ## Specification
 
-Any protocol, network, smart contract system, or similar that claims at time `t` full EVM equivalence MUST fully (=100%) comply with the latest [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper) version at time `t`.
+Any protocol, network, smart contract system, or similar that claims at time `t` full EVM equivalence MUST fully (=100%) comply with the latest [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf) version at time `t`.
 
 ## Rationale
 
-A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence. The [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
+A proper equivalence definition requires a common, publicly available reference that specifies the formal specification of the equivalence. The [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf) formulates the formal specification of the Ethereum protocol, which is publicly available and serves as a common reference for all protocols, networks, smart contract systems, or the like that aim to build similar implementations.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-6269.md
+++ b/EIPS/eip-6269.md
@@ -5,7 +5,7 @@ description: Canonicalise the definition of Full EVM Equivalence
 author: Pascal Caversaccio (@pcaversaccio)
 discussions-to: https://ethereum-magicians.org/t/evm-equivalence-and-ethereum-stack-compatibility-definition-future-informational-eip/10044
 status: Draft
-type: Informational
+type: Meta
 created: 2023-01-06
 ---
 
--- a/EIPS/eip-6269.md
+++ b/EIPS/eip-6269.md
@@ -7,11 +7,12 @@ discussions-to: https://ethereum-magicians.org/t/evm-equivalence-eip-6269-and-et
 status: Draft
 type: Meta
 created: 2023-01-06
+requires: 5757
 ---
 
 ## Abstract
 
-This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/master). This definition applies retroactively from the first release of the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) in April 2014.
+This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is the **complete** compliance with the latest version of the [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/79db92bd4c49bdf5f559fdebea5a9a3a2bf3fb83). This definition applies retroactively since the first release of the [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf) on 2 April 2014 ([first version](https://github.com/ethereum/yellowpaper/blob/0d0d23301d077bbdab5cafae6ab06001e282fae2/Paper.tex)).
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


pcaversaccio:
May I ask - what's the preferred way to link the Ethereum Yellow Paper in the EIP? Currently, the CI throws `error[markdown-rel-links]: non-relative link or image`.

pcaversaccio:
> The yellow paper is a deprecated resource, please refer to the [`execution-specs`](https://github.com/ethereum/execution-specs) repository for the latest definitions of the EVM.

thx - for my clarification since this EIP also refers to the past. When was the official switch date? I.e. until what date was the Yellow Paper the official resource to follow?

EIPS/eip-6269.md
```diff
@@ -0,0 +1,58 @@
+---
+eip: 6269
+title: Full EVM Equivalence
+description: Canonicalise the definition of Full EVM Equivalence
+author: Pascal Caversaccio (@pcaversaccio)
+discussions-to: https://ethereum-magicians.org/t/evm-equivalence-and-ethereum-stack-compatibility-definition-future-informational-eip/10044
+status: Draft
+type: Informational
```
Pandapip1, line 8:
```suggestion
type: Meta
```
Since this proposes changes that cannot be ignored, I suggest that this be a Meta-EIP.

pcaversaccio: Makes sense

EIPS/eip-6269.md
```diff
@@ -0,0 +1,58 @@
...
+author: Pascal Caversaccio (@pcaversaccio)
+discussions-to: https://ethereum-magicians.org/t/evm-equivalence-and-ethereum-stack-compatibility-definition-future-informational-eip/10044
+status: Draft
+type: Meta
+created: 2023-01-06
+---
+
+## Abstract
+
+This EIP standardises the definition of "Full Ethereum Virtual Machine (EVM) Equivalence". Full EVM equivalence is **complete** compliance with the latest version of the [Execution Layer Specification](https://github.com/ethereum/execution-specs/tree/24de2192e02bba11e61746aa9dee04d4e7a8b62e). This definition applies retroactively from the first release of the Ethereum Yellow Paper in April 2014.
```
pcaversaccio, line 14: Is there a certain reason we refer to a specific commit @Pandapip1? I would prefer referring to https://github.com/ethereum/execution-specs.

Pandapip1: Yes, as per EIP-5757 external links to any repository must use a specific commit.

pcaversaccio: But there cannot be a unique ID for that link since it should point always to the latest commit. I trust your judgement here how to deal with it but otherwise people will always get a link of an outdated version...

Pandapip1: Yes, that's the point. They get a link to a thing that never changes.

pcaversaccio: ok understood - I fixed all the links accordingly and even added `requires: 5757` since we need it in this context.

pcaversaccio:
I addressed all the link issues by @Pandapip1 and also added more clarity on how this works backwards & added an additional test case.

Pandapip1 (approved):
LGTM. Will need manual merge.

pcaversaccio:
> The execution-specs and yellow paper haven't been added as permitted external resources. We'll have to follow the process in [EIP-5757](https://eips.ethereum.org/EIPS/eip-5757) if we want to allow them.

My understanding is that the editors (not me) will need to approve the execution-specs and yellow paper as a permitted source, am I right? If I misunderstand the process please let me know how I can support this approval process best.

#### Excerpt from [EIP-5757](https://eips.ethereum.org/EIPS/eip-5757)
> Should the editors determine that an origin meets the requirements above, EIP-1 MUST be updated to include:
> 
> The name of the allowed origin;
> The permitted markup and formatting required when referring to resources from the origin; and
> A fully rendered example of what a link should look like.

SamWilsn:
We can discuss adding these sources on the EIPIP meeting tomorrow. If you'd like to create a pull request updating EIP-1 with the sources you'd like to link to, it would certainly help!

pcaversaccio:
@SamWilsn I opened a PR here https://github.com/ethereum/EIPs/pull/6306.

EIPS/eip-6269.md
```diff
@@ -0,0 +1,69 @@
+---
+eip: 6269
+title: Full EVM Equivalence
+description: Canonicalise the definition of Full EVM Equivalence
+author: Pascal Caversaccio (@pcaversaccio)
+discussions-to: https://ethereum-magicians.org/t/evm-equivalence-eip-6269-and-ethereum-stack-compatibility-definition/10044
+status: Draft
+type: Meta
```
lightclient, line 8: Should be type informational

pcaversaccio: My original draft was of type _informational_, @Pandapip1 suggested the _meta_ type https://github.com/ethereum/EIPs/pull/6269#discussion_r1063634390

lightclient: Yeah I think that is the correct type. Meta is used for processes around Ethereum which require consensus like the EIP process, hard forks, testnets, etc. This EIP is unenforceable (but still useful from an informational standpoint) and so I think this is the correct type.

pcaversaccio: I think the question is about whether it can be ignored or not - imho it should not be ignored (otherwise my initiative is somehow useless). And if I understand correctly, an _informational_ EIP can be ignored whilst a _meta_ EIP cannot be ignored. Not sure what's the approach here to find a consensus but I want to clearly define a very important semantics that cannot be simply ignored by the community...

Pandapip1: > Yeah I think that is the correct type. Meta is used for processes around Ethereum which require consensus like the EIP process, hard forks, testnets, etc. This EIP is unenforceable (but still useful from an informational standpoint) and so I think this is the correct type.

It's not enforceable, but it's nonetheless not an EIP that can be ignored. Defining something, IMO, is Meta, while describing an aspect of something is Informational.

SamWilsn:
I'm not 100% sure that defining what "full EVM equivalence" means is a good fit for an EIP.

Someone else could come along and make a completely different EIP that defines full EVM equivalence as being able to simulate **E**xtra-terrestrial **V**ampire **M**achinations and it would be equally valid to progress through the process.

EIP Editors aren't arbiters of The Truth, not do we want to be in that position. We cannot force anyone to use this particular definition, nor will we stop anyone from standardizing an alternate definition.

As much as I hate it when I have to make this recommendation, but this might be better suited as a blog post or article on FEM or ethresear.ch.

pcaversaccio:
To be quite honest, I don't agree with this statement. First, I can understand your arguments, but a proper definition of _Full EVM Equivalence_ (and later _Full Ethereum Equivalence_) helps to get a common understanding of what this term actually means. As with normal ERCs, for example, I cannot propose a new ERC-20 standard and call it _EIP-20: Token Standard_. I would have to use a new name. The same applies to this proposal. No one can come along and simply call a new EIP _Full EVM Equivalence_. Besides, meta-EIPs require community consensus. So if we accept one new definition, we should discard/deprecate another at the same time. Another point I would like to highlight is that you make definitions in many EIPs. The most obvious of these is [EIP-4] (https://eips.ethereum.org/EIPS/eip-4). Technically, I could now come up with the same classification scheme, but with a definition, that hard forks are only called hard forks if they are activated in the months of June and October.

Much like Ethereum requires an official formal specification of the consensus and execution layers, we should have a community-approved official definition of what "Full EVM Equivalence" means. This will not be achieved by a simple blog post.

SamWilsn:
> a proper definition of _Full EVM Equivalence_ (and later _Full Ethereum Equivalence_) helps to get a common understanding of what this term actually means.

I absolutely agree! I never meant to imply that the content of this document is not useful/important.

> As with normal ERCs, for example, I cannot propose a new ERC-20 standard and call it _EIP-20: Token Standard_ (my assumption would be that you would reject it).

No, you couldn't call it "EIP-20", but you could create "EIP-9989: Token Standard" and we wouldn't have a problem with it.

> No one can come along and simply call a new EIP _Full EVM Equivalence_.

For better or worse, they absolutely could. Then if you're writing an article, you need to specify which EIP number your definition is coming from.

> Much like Ethereum requires an official formal specification of the consensus and execution layers, we should have a community-approved official definition of what "Full EVM Equivalence" means. This will not be achieved by a simple blog post or article. Honestly, I wouldn't bother if semantics weren't that important.

Language is a lot trickier than code, so it's much easier to tell what the consensus is on the CL and EL: whatever the client software agrees on. Publishing this as an EIP won't stop an L2 from claiming they have Full EVM Equivalence even if they don't meet this definition, while publishing a Core EIP will stop non-compliant nodes from participating in the same chain.

pcaversaccio:
> Language is a lot trickier than code

Yes absolutely, interpretation is not binary :)

> Publishing this as an EIP won't stop an L2 from claiming they have Full EVM Equivalence even if they don't meet this definition

I agree that they still could do this but if we got an official _meta_ EIP nobody can ignore it (that's the definition of a _meta_ EIP if I understand it correctly). So if an L2 claims _Full EVM Equivalence_ you can always challenge them if this is in accordance with `EIP-6269`. 

The problem what we have right now is the following:
- One L2 claims full EVM Equivalence but has different gas costs
- Another L2 claims full EVM Equivalence but has an additional precompile
- etc.

It's very confusing for many folks to understand what actually _Full EVM Equivalence_ means. What I try to achieve here is that in case anyone claims _Full EVM Equivalence_ you must benchmark it against an official EIP that got community approval. Or in other words, I try to foster a common denominator, and any deviation thereof must be explained.

If I compare it to an `ERC20` token, where you have high-end tokens that do not follow to interface wrt the return values for `transfer`, `transferFrom`, or `approve`, the execution layer doesn't stop working. What will happen is that tokens get stuck or certain application protocols will revert and people will lose money. So a complete on-chain enforcement of a core EIP is also not always possible (not talking about your example with non-compliant nodes where this is possible of course). What will happen is that people realise that they need to build wrapper functions around it and need to justify themselves why they didn't follow the standard. And why does this happen? Because there is a clear formal specification to reference. And the same applies here, where you have a clear specification in form of the EELS.

Pandapip1:
> EIP Editors aren't arbiters of The Truth, not do we want to be in that position. We cannot force anyone to use this particular definition, nor will we stop anyone from standardizing an alternate definition.

@SamWilsn here's my counterargument. EIPs are standards. People are free to use or not use them. This EIP is no exception. By merging this, we aren't saying that this is the truth. All it says is that that this is a subjective, defined interpretation of what "EVM Equivalence" means.

Gesisfm:
pcaversaccio:evm-equivalence

SamWilsn:
Please correct me if I'm wrong, but it sounds like there are two opposite viewpoints here:

> an official _meta_ EIP nobody can ignore it

This implies a change to the EIP process where editors now have to evaluate every EIP that uses the term "Full EVM Equivalence" against the definition in this EIP and make judgment calls on whether to allow that EIP to use the term or not. I am **flat out against** this type of policing. EIP Editors shouldn't have to be technical experts, and I don't want to ever end up caught in an argument over whether or not some L2 is fully equivalent based on some fine technical detail.

Further, if we allow this one "official" definition, that opens the door to other "official" definitions. Any marketing team that wants to monopolize a term will be able to open a Meta EIP to try and it lock down so narrowly that only their product can meet the criteria. To be clear, I'm not saying that's what happening here.

> others have [...] the chance to formalize their own interpretations, not that this is the "correct" one

Then what's the point of making it an EIP? We don't have an EIP that tries to enforce a definition on the community of, for example, "account abstraction" or "soulbound token". Instead those definitions are limited to the EIPs defining relevant technical details.

pcaversaccio:
> Please correct me if I'm wrong, but it sounds like there are two opposite viewpoints here:

Actually the wording adapted is from @Pandapip1 https://github.com/ethereum/EIPs/pull/6269#discussion_r1063634390 and I fully agree on this one. So no, I don't think we have two opposite viewpoints here.

> This implies a change to the EIP process where editors now have to evaluate every EIP that uses the term "Full EVM Equivalence" against the definition in this EIP

Technically, it would be simply `requires: 6269` if you want to use the term _Full EVM Equivalence_ in your EIP. If an EIP defines simply something that is EVM equivalent (not _full_) it does not require this definition. What this EIP achieves IMHO is a canonical definition of what _Full_ EVM Equivalence means. Nothing else.

> Further, if we allow this one "official" definition, that opens the door to other "official" definitions. Any marketing team that wants to monopolize a term will be able to open a Meta EIP to try and it lock down so narrowly that only their product can meet the criteria. To be clear, I'm not saying that's what happening here.

That's exactly why I'm proposing this definition since am not part of any L2 teams etc. - if I understand correctly the EIP process, in order to achieve a final status we need to find a _community-consensus_. No marketing team will ever achieve that if they abuse the EIP process for their own product branding. That's at least how I see it (maybe naively).

> Then what's the point of making it an EIP? We don't have an EIP that tries to enforce a definition on the community of, for example, "account abstraction" or "soulbound token". Instead those definitions are limited to the EIPs defining relevant technical details.

It's the _subjective_ interpretation by the **broad** Ethereum community (and not by one single person or a marketing team) of what "Full EVM Equivalence" means. Anyone can refer to it - if it's a marketing team, a researcher etc. It's the lowest common denominator on which people have agreed on. Like with standards.

SamWilsn:
> > This implies a change to the EIP process where editors now have to evaluate every EIP that uses the term "Full EVM Equivalence" against the definition in this EIP
> 
> Technically, it would be simply `requires: 6269` if you want to use the term _Full EVM Equivalence_ in your EIP. If an EIP defines simply something that is EVM equivalent (not _full_) it does not require this definition. What this EIP achieves IMHO is a canonical definition of what _Full_ EVM Equivalence means. Nothing else.

Ah, okay. I probably misunderstood "if we got an official _meta_ EIP nobody can ignore it". A Meta EIP would be the correct category for a proposal to prevent other proposals from using an incorrect definition.

You'd only need an Informational EIP if it's opt in, like with the `requires` header.

> > Further, if we allow this one "official" definition, that opens the door to other "official" definitions. Any marketing team that wants to monopolize a term will be able to open a Meta EIP to try and it lock down so narrowly that only their product can meet the criteria. To be clear, I'm not saying that's what happening here.
> 
> That's exactly why I'm proposing this definition since am not part of any L2 teams etc.

I'm more concerned with the precedent that would be established if this EIP gets included.

> [...] if I understand correctly the EIP process, in order to achieve a final status we need to find a _community-consensus_.

Final just means that the EIP won't be changed any more, and that the EIP meets a really low bar for completeness and readability. Editors may extend the last call period if there's still active discussion, but that's it really.

> No marketing team will ever achieve that if they abuse the EIP process for their own product branding. That's at least how I see it (maybe naively).

For whatever reason, getting an EIP through the process is seen as a badge of... well... something. Guess it implies the authors have some minimum level of persistence :rofl:

> > Then what's the point of making it an EIP? We don't have an EIP that tries to enforce a definition on the community of, for example, "account abstraction" or "soulbound token". Instead those definitions are limited to the EIPs defining relevant technical details.
> 
> It's the _subjective_ interpretation by the **broad** Ethereum community (and not by one single person or a marketing team) of what "Full EVM Equivalence" means. Anyone can refer to it - if it's a marketing team, a researcher etc. It's the lowest common denominator on which people have agreed on. Like with standards.

I can see some value of having this EIP if used like you describe (in the `requires` header.) I just don't think it's enough to stand on its own. We won't stop anyone from using the term to mean something else, and we don't force consensus before granting a Final status.

The first EIP that would've `require`d this definition can simply include this definition inside itself, and other EIPs can reference it from there.

pcaversaccio:
> A Meta EIP would be the correct category for a proposal to prevent other proposals from using an incorrect definition.

This is actually the case, as previously described above, if someone wants to use the term "Full EVM Equivalence" in their EIP. That's why it _cannot_ be ignored and should be classified as _meta_.

> Guess it implies the authors have some minimum level of persistence
 
Lol, I'm definitely having this persistence hehe

> I can see some value of having this EIP if used like you describe (in the requires header.) I just don't think it's enough to stand on its own. We won't stop anyone from using the term to mean something else, and we don't force consensus before granting a Final status.

That's not fully correct - if someone wants to use the term "Full EVM Equivalence" (see the term _full_) he/she/they need to require this EIP definition. If someone refers simply to "EVM Equivalence" or compatibility etc. it's fine IMHO to not use this term. That's why I think _meta_ is the right classification since you cannot ignore it if you refer to _full_ EVM equivalence.

> The first EIP that would've `require`d this definition can simply include this definition inside itself, and other EIPs can reference it from there.

I don't think this is the correct way of handling such an important definition. Such a definition should not be declared in a context of an EIP that has a further scope. As an example, pivotal mathematical definitions are not declared as part of a corollary or lemma. They are defined as isolated pieces to assure they only focus on the definition itself to prevent misinterpretation. It's actually beneficial since having such pivotal definitions in a modular way, you have a nice composability that fits within your `requires` design.

I feel like we somehow turn in circles here and I want to emphasise that I understand the counterarguments as well. So I'm not fully sure on how to achieve a consensus here. The current important discussion prevents me also currently in drafting the EIP for "Full Ethereum Equivalence", since if my EIP is dismissed here I cannot open a new EIP with a similar logic :)

EIPS/eip-6269.md
lightclient: Please review the EIPs that are classified as "Meta" and then review the EIPs that are classified as "Informational". This EIP is an Informational EIP because it does not affect anything related to Ethereum "process" (again, see definition of meta type in EIP-1). It is providing a definition for a term to the community, not proposing any new feature or process change, and is advisable to follow, but is not enforceable.

EIPS/eip-6269.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+Any protocol, network, smart contract system, or similar that claims at time `t` full EVM equivalence **MUST** fully (=100%) comply with the latest [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/79db92bd4c49bdf5f559fdebea5a9a3a2bf3fb83) version at time `t`. For any time `t` before 6 September 2022 (Bellatrix hard fork date), the **REQUIRED** reference source is the [Ethereum Yellow Paper](https://github.com/ethereum/yellowpaper/blob/1016c0603062b76388e3c3c19786cd5f9ca9ac61/paper.pdf). The first public version of the Ethereum Yellow Paper is available at commit [`0d0d23301d077bbdab5cafae6ab06001e282fae2`](https://github.com/ethereum/yellowpaper/blob/0d0d23301d077bbdab5cafae6ab06001e282fae2/Paper.tex).
```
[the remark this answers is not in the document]
Pandapip1: Do the execution specs specify that behavior not defined in the specs isn't allowed? For example, EIP-20 says that tokens MUST implement an interface, but it doesn't say that other interfaces can't be implemented too.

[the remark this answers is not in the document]
lightclient: In this case, what you're defining to be "EVM Equivalence" is not, as far as I can tell, the same as what the community at large has defined it to be. E.g. see docs from optimism and note the use of "nearly" https://help.optimism.io/hc/en-us/articles/8004922894491-What-is-the-difference-between-EVM-equivalence-and-Ethereum-equivalence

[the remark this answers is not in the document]
pcaversaccio: I want to be impartial here and emphasise that Optimism is _not_ the community. Also, I define here _full_ EVM Equivalence, and not _nearly_. Funny enough, my definition is actually completely in line with Optimism's definition of "Complete EVM Equivalence" (https://medium.com/ethereum-optimism/introducing-evm-equivalence-5c2021deb306):

image

So I want to push back here and argue that my definition actually reflects the community's opinion.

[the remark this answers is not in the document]
lightclient: That post is from late 2021. The optimism docs I sent are newer and there are several other sources about how optimism thinks about it now.

Yes, optimism is not the community. Polygon thinks about somewhat similarly: https://twitter.com/0xPolygon/status/1549716953644044289?s=20&t=E6lJvDYXLDF2QM3Lp4l8bQ

It appears arbitrum doesn't even use the term "EVM equivalence".

--

Ultimately I don't think this should be an EIP, especially if it conflicts with how members of the community are already using this term.

SamWilsn:
> > A Meta EIP would be the correct category for a proposal to prevent other proposals from using an incorrect definition.
> 
> This is actually the case, as previously described above, if someone wants to use the term "Full EVM Equivalence" in their EIP. That's why it _cannot_ be ignored and should be classified as _meta_.

Please forgive my boldness, but **there is absolutely no chance I will allow a `type: Meta` proposal to exclusively define a term** (with, perhaps, the narrow exception of an EIP describing some part of the EIP governance process.) That is _way_ outside the scope of what EIP Editors do, for [the reasons I mentioned above](https://github.com/ethereum/EIPs/pull/6269#issuecomment-1409315925).

This is a hill I will gladly die on.

---

I am marginally less opposed to an opt-in definition (in a `type: Informational` proposal), but I am still opposed to it. The EIPs repository is just not the place to put definitions.

SamWilsn (changes requested):
.

pcaversaccio:
So to summarise, we got the following positions:

@lightclient 
> Ultimately I don't think this should be an EIP, especially if it conflicts with how members of the community are already using this term.

@SamWilsn 
> I am marginally less opposed to an opt-in definition (in a type: Informational proposal), but I am still opposed to it. The EIPs repository is just not the place to put definitions.

I can see the arguments in favour of the _informational_ EIP classification, and I am also willing to adjust it after the extensive back and forth, but I am strongly against discarding this EIP. As the discussion with @lightclient [here](https://github.com/ethereum/EIPs/pull/6269#discussion_r1103436884) nicely highlights, different interpretations were/are used. And that's the **major** problem - it creates confusion! I want to offer a proper definition that does not stem from one of the project teams in order to foster a common understanding, and I am convinced that the EIP is the right place for this, not FEM or ethresear.ch. To put it like @SamWilsn - this is a hill where I will gladly die on.

Is there a way we can still find a consensus here?

EIPS/eip-6269.md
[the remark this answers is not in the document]
pcaversaccio: I replied [here](https://github.com/ethereum/EIPs/pull/6269#issuecomment-1427610712) to also answer to @SamWilsn's answer combined. As mentioned in my reply, the discussion here shows very nicely that different definitions have been/are being used in the community (I can also share further interpretations of zkSync etc. if needed). This creates a lot of confusion and I would like to provide the community with a reference point for guidance.

lightclient:
It seems like there is consensus? Sam and I are both opposed to the EIP.

pcaversaccio:
> It seems like there is consensus? Sam and I are both opposed to the EIP.

Is this how a consensus is reached on whether an EIP should be merged? Two EIP editors disagree and that's it? If this is how the decision-making goes, I will accept the decision with a big disappointment tbh, but I would also like to hear @Pandapip1's final assessment.

Pandapip1:
I see Sam's point. The difference between Meta and Informational is a bit nuanced at the moment, and there doesn't seem to be consensus as to what that difference is. I recommend this EIP to move to Informational, and if we come to a consensus later, and that consensus allows this EIP to be moved to Meta, it can be done.

Pandapip1:
I see Sam's point. The difference between Meta and Informational is a bit nuanced at the moment, and there doesn't seem to be consensus as to what that difference is. I recommend this EIP to move to Informational, and if we come to a consensus later, and that consensus allows this EIP to be moved to Meta, it can be done.

Pandapip1:
I see Sam's point. The difference between Meta and Informational is a bit nuanced at the moment, and there doesn't seem to be consensus as to what that difference is. I recommend this EIP to move to Informational, and if we come to a consensus later, and that consensus allows this EIP to be moved to Meta, it can be done.

Pandapip1:
I see Sam's point. The difference between Meta and Informational is a bit nuanced at the moment, and there doesn't seem to be consensus as to what that difference is. I recommend this EIP to move to Informational, and if we come to a consensus later, and that consensus allows this EIP to be moved to Meta, it can be done.

Pandapip1:
I see Sam's point. The difference between Meta and Informational is a bit nuanced at the moment, and there doesn't seem to be consensus as to what that difference is. I recommend this EIP to move to Informational, and if we come to a consensus later, and that consensus allows this EIP to be moved to Meta, it can be done.

pcaversaccio:
@SamWilsn @lightclient @Pandapip1 I now changed the type to _Informational_ accordingly.

gcolvin:
> Is this how a consensus is reached on whether an EIP should be merged? Two EIP editors disagree and that's it? 

We would need to actually discuss this at a meeting.  If any Editor has a _blocking_ disagreement then we don't have consensus.

gcolvin:
I don't really know what to make of this EIP.  The gist of it seems to be just
> Any protocol, network, smart contract system, or similar that claims at time t full EVM equivalence MUST fully (=100%) comply with the latest Ethereum Execution Client Specifications ...

Anyone is already free to claim that their system "complies with the Ethereum Execution Client Specifications ...", either fully or with qualifications, and it's possible to test such claims.  Trying to further constrain what a marketing department calls things by publishing definitions seems pretty hopeless -- any consensus on that will have to come from the industry itself.  And even then, _caveat emptor_.

Pandapip1 (approved):
I am in favor of this EIP.

pcaversaccio:
@lightclient @SamWilsn @Pandapip1 - you were probably busy with the Denver event, but I am quickly follow-up here. Any update on how we progress from here now?

Pandapip1:
> @lightclient @SamWilsn @Pandapip1 - you were probably busy with the Denver event, but I am quickly follow-up here. Any update on how we progress from here now?

Wait until consensus. I would recommend putting it on the EIPIP agenda.

pcaversaccio:
For obvious reasons, I don't agree with the decision, but I do respect the decision. For that reason, I will also not propose a definition of _Full Ethereum Equivalence_. I would like to quickly highlight (even though it won't change the decision), why such a definition is really important. I break down in my tweet [here](https://twitter.com/pcaversaccio/status/1635346927465406465), why the opcodes `CREATE` and `CREATE2` of zkSync do not have the same behavior as on Ethereum. Any claim who would state that zkSync is _fully_ EVM equivalent, would be wrong.

EIPS/eip-6269.md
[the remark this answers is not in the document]
pcaversaccio: Since this PR got closed, I mark the discussion as resolved.

xinbenlv:
@pcaversaccio that's a good point. While I don't think an EIP maybe best suitable for this use case, I do a agree it's important for people to understand if a VM (regardless of layer or zk), doens't share same behavior defined in Execution spec, shall not be "called" EVM compliant.

pcaversaccio:
Right, but IMHO @xinbenlv having an "official" definition would have helped a lot (that's why I made this proposal). Right now, each project will (continue) define it in a way, that they can market their projects best. Also, if I publish this as a blog on FEM or somewhere else, it will not have the same impact. It is what is and I have to deal with that.

PS: It seems that the EIP title gets removed when a branch is deleted (which is done usually after merging or closing of PR). You should consider fixing this in the CI.

pcaversaccio:
In the meantime (while the other editors hopefully reflect on this EIP proposal again) I keep documenting here why this 'official' definition is needed. Yesterday, almost $2m of Ether got stuck on zkSync Era since their gas pricing is not EVM equivalent, see e.g. here:
- https://twitter.com/zksync/status/1644139364270878720
- https://twitter.com/pcaversaccio/status/1644023695596199945

However, they state [here](https://docs.zksync.io/zkevm/#what-is-zksync-era) that they are an EVM-compatible ZK Rollup. They don't claim equivalence, of course! But people don't understand what equivalence, compatibility etc. Any project that claims full EVM equivalence could simply refer to this EIP. Again, I'm not affiliated with any L1/L2 projects and completely impartial (Disclaimer: I just got a small amount of OP tokens via the recent retroPGF round from Optimism, which is however a recent event that not collide with the date when I opened this PR).

An EIP should also serve the community to clear up ambiguities about core terminology, and this what I want to achieve here.

Sabnock01:
I also think it's worth mentioning re; zkSync that their bytecode specification disallows them from being [EIP-1167](https://eips.ethereum.org/EIPS/eip-1167) equivalent and is also grounds for their classification as non EVM-equivalent. 

* https://twitter.com/pcaversaccio/status/1642448012193546240?s=20

supernovahs:
I also feel there should be an official definition of Full EVM Equivalence.

miohtama:
While it might sound good on paper, I am concerned this will ossify already old EVM technology development even more.

I would say there is 

- No EVM definition
- Projects are discouraged to say they are EVM compatible, by we giving them angry looks
- Always test your projects chain by chain, because there is and never will be a fully agreed definition of EVM compatible as there are always pitfalls and chains evolve

pcaversaccio:
> Em... I have to say, you argument sounds convincing to me to turn netural. Let's see if this change other editors' or authors' minds so we could have a quorum to revisit.

@xinbenlv Any update on whether there is a quorum to revisit this EIP draft? Again, I wouldn't push this hard if I thought it was not a value-add for the entire ecosystem. Official definitions provide the necessary orientation that many people need in our space.

pcaversaccio:
Another situation we are in right now is where such a definition is very helpful. There are numerous chains that _do not_ support currently the latest EVM version `shanghai` with the opcode `PUSH0`. This means that, for example, the use of the latest Solidity version `0.8.20`, which supports `PUSH0`, will likely fail with these chains (if it includes `PUSH0` in the init code) and thus these chains can no longer be considered fully EVM-equivalent.

pcaversaccio:
Let me follow up here on community support on this EIP (by @Sabnock01 and @devanoneth):
- https://twitter.com/devanoneth/status/1661720111827279872
- https://twitter.com/Sabnock66/status/1661465202850689027


@xinbenlv, @lightclient, @SamWilsn, @Pandapip1 I would appreciate if you could re-consider this EIP since it's really really relevant.

xinbenlv:
Yes, given the context of zkEVM, I think it's worthy to reopen this discussion about defining what's considered "EVM compliance"

Also we shall discuss it to be whether an informational or meta EIP?

kdenhartog:
Here's an alternative method you might want to consider @pcaversaccio. Have you thought about getting this term defined within the execution specs rather than as a separate EIP? Also, given the examples you're using here I think what you'd actually want is something more than just a definition on paper. Instead, it would be much more useful for developers to recognize this compatibility based on a test suite. This is[ very common at W3C and WHATWG](https://web-platform-tests.org/) and every web platform API is meant to also have a defined set of tests to be added to the test suite when it gets standardized. This could be very useful as a method to achieving code equivalence or making it much easier to distinguish where implementations actually differ. For example, here's a [giant list of all the Web platform tests](https://wpt.fyi/results/?label=experimental&label=master&aligned) on how browsers differ currently.

pcaversaccio:
> Here's an alternative method you might want to consider @pcaversaccio. Have you thought about getting this term defined within the execution specs rather than as a separate EIP?

Thanks for sharing your thoughts @kdenhartog. The main reason why I think an EIP is the best place for such a definition is its _canonical_ acceptance and the broader reach beyond tech-savvy people. It is very important to eliminate ambiguities across different groups of people (engineering, marketing, etc.) and an EIP is the best way to achieve this IMHO.

> Also, given the examples you're using here I think what you'd actually want is something more than just a definition on paper. Instead, it would be much more useful for developers to recognize this compatibility based on a test suite. This is[ very common at W3C and WHATWG](https://web-platform-tests.org/) and every web platform API is meant to also have a defined set of tests to be added to the test suite when it gets standardized. This could be very useful as a method to achieving code equivalence or making it much easier to distinguish where implementations actually differ. For example, here's a [giant list of all the Web platform tests](https://wpt.fyi/results/?label=experimental&label=master&aligned) on how browsers differ currently.

Good point - there are test suites [here](https://github.com/ethereum/execution-spec-tests) or [here](https://github.com/ethereum/tests) that are also used by other projects (e.g. [Polygon](https://github.com/0xpolygonhermez#ethereum-compatibility)) to test EVM equivalence, BUT, my definition is to be understood as _formal_ in the sense that it does not assume that the test suite is always up to date and that the test suite always covers 100% of the execution specs. There is only **one source of truth** and that is the [Ethereum execution specifications](https://github.com/ethereum/execution-specs/tree/master).

gcolvin:
Late to this party, but this has become relevant to my paid work.  The abstract states that

> Full EVM equivalence is the complete compliance with the latest version of the [Ethereum Execution Client Specifications](https://github.com/ethereum/execution-specs/tree/79db92bd4c49bdf5f559fdebea5a9a3a2bf3fb83).

For better and worse this already _is_ the current definition, and doesn't need an EIP to say so.  And the executable spec already contains code for all previous upgrades, so there is no need for an EIP to discuss that either.

What is more of a problem is less-than-full compliance.  Vendors can already describe this in terms of the upgrade level they are up too, the EIPs they omit, the opcodes they don't implement, the extensions they make, and the versions of Solidity (if any) they can accept code from. If all of this becomes enough of problem for the industry, and people want to use the EIP process to deal with it, then they can form working group to create relevant ERCs.

For our project we would need only the computational subset of the -- no introspection or message calls.  We would extend the EVM with EIP-2315, as we can't wait on EOF to be able to generate machine code.  We would add our own "pre-compiles" for communication with the rest of our system.  And we would extend Vyper (with or without their help) to generate code for our custom EVM.  I don't know if there are enough other projects doing similar things that it would be worth forming a working group to coordinate -- but right now our project has seed capital and deadlines.

PR #6300, opened 2023-01-08
Update EIP-5267: Move to Last Call

frangio:


SamWilsn (approved):

xinbenlv:
Discussion: can you share the rationale of `extensions` being format as `uint256` instead of `bytes32` ?

> Extensions are described by their EIP numbers because EIP-712 states: “Future extensions to this standard can add new fields […] new fields should be proposed through the EIP process.”

Here EIP-712 didn't specify what format of fields will be, so it's open to future designer to design, EIP-5267 is one of such future design.

frangio:
> Is it intentional that in author's perspective the extension shall _at least_ have one element?

No. This is clear in the specification. There are no constraints on the length of the extensions array. Additionally the Solidity reference implementation returns an empty array.

Will reply to the other questions in the discussions thread as I don't think they are editorial.

xinbenlv:
> Will reply to the other questions in the discussions thread as I don't think they are editorial.

Sure. Some questions are in editorial nature e.g. we suggest spell out / clarify some rationale in some design choices that the author made. But those editorial questions are not merge-blocking and can be addressed in the followup PRs.

PR #6303, opened 2023-01-09
Update EIP-5570: Move to review

darcys22:


Pandapip1 (approved):
I remain unconvinced as to the usability of NFTs as receipts, but this is okay to be moved to Review.

PR #6311, opened 2023-01-11
Update EIP-6150: Move to Last Call

fewwwww:
Moving EIP-6150 to last call after adding demo implementation for default and optional interfaces.


keeganlee (approved):

lekapoe:
merged

Pandapip1 (approved):
LGTM. I still don't see this being used a lot though.

PR #6325, opened 2023-01-12
EIP-4895: CL-EL withdrawals harmonization: using units of Gwei

ralexstokes:
Following discussion from here https://github.com/ethereum/pm/issues/702, protocol devs want to explore harmonization in the withdrawals format from CL to EL.

This PR updates EIP-4895 with the first part of this change to use the same units (`Gwei`) for the `amount` in each withdrawal.

### DIFF
--- a/EIPS/eip-4895.md
+++ b/EIPS/eip-4895.md
@@ -44,7 +44,7 @@ Define a new payload-level object called a `withdrawal` that describes withdrawa
 1. a monotonically increasing `index`, starting from 0, as a `uint64` value that increments by 1 per withdrawal to uniquely identify each withdrawal
 2. the `validator_index` of the validator, as a `uint64` value, on the consensus layer the withdrawal corresponds to
 3. a recipient for the withdrawn ether `address` as a 20-byte value
-4. a nonzero `amount` of ether given in wei as a `uint256` value.
+4. a nonzero `amount` of ether given in Gwei (1e9 wei) as a `uint64` value.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-4895.md
```diff
@@ -129,6 +129,8 @@ The `withdrawals` in an execution payload are processed **after** any user-level
 
 For each `withdrawal` in the list of `execution_payload.withdrawals`, the implementation increases the balance of the `address` specified by the `amount` given.
 
+Recall that the `amount` is given in units of Gwei so a conversion of units of wei must be performed when working with account balances in the execution state.
```
timbeiko, line 132: typo?

```suggestion
Recall that the `amount` is given in units of Gwei so a conversion of units to wei must be performed when working with account balances in the execution state.
```

ralexstokes: ah good catch, this is what I meant
```suggestion
Recall that the `amount` is given in units of Gwei so a conversion to units of wei must be performed when working with account balances in the execution state.
```

timbeiko: or: 
```suggestion
Recall that the `amount` is given in units of Gwei so a conversion to units of wei must be performed when working with account balances in the execution state.
```

yperbasis (dismissed):

djrtwo (approved):
looks good other than the lint error

etan-status (dismissed):

g11tech (approved):
lgtm

PR #6330, opened 2023-01-14
Update EIP-5192: Change name to "Lockable tokens"

TimDaub:
When this standard was created, the term "Soulbound tokens" was cool and new. Many people understood that a Soulbound token is "bound to a soul," and so it is a permanently locked token of an account. But since then, it has become clear that a Soulbound token is an umbrella term for all kinds of tokens that break conventional norms around private property assumptions. And besides, EIP-5192 implements dynamic locking of a token. So "Lockable Tokens" is a much better term for this specification and also for developers arriving at it in the future.

I'm hence asking the EIP editors to make an exception regarding changing the contents of this EIP. The interface has not changed. Developers can still rely on using this token standard. The only change is an improvement in terms of accuracy in identifying what it is: A lockable token.

### DIFF
--- a/EIPS/eip-5192.md
+++ b/EIPS/eip-5192.md
@@ -45,9 +45,9 @@ interface IERC5192 {
   event Unlocked(uint256 tokenId);
 
   /// @notice Returns the locking status of a token
-  /// @dev SBTs assigned to zero address are considered invalid, and queries
+  /// @dev Tokens assigned to zero address are considered invalid, and queries
   /// about them do throw.
-  /// @param tokenId The identifier for an SBT.
+  /// @param tokenId The identifier for a token.
   function locked(uint256 tokenId) external view returns (bool);
 }
 ```
--- a/EIPS/eip-5192.md
+++ b/EIPS/eip-5192.md
@@ -1,7 +1,7 @@
 ---
 eip: 5192
-title: Minimal Soulbound NFTs
-description: Minimal interface for soulbinding EIP-721 NFTs
+title: Lockable NFTs
+description: Minimal interface for lockable EIP-721 NFTs
 author: Tim Daubenschütz (@TimDaub), Anders (@0xanders)
 discussions-to: https://ethereum-magicians.org/t/eip-5192-minimal-soulbound-nfts/9814
 status: Final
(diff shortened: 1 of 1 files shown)


sullof:
Good move.

Pandapip1 (commented):
I'm in two minds about this.

1. The correct way to refer to an EIP outside of the EIPs repository is `[EIP-X](https://eips.ethereum.org/EIPS/eip-x)`. As such, changing a name shouldn't cause any issues, and so in this regard I am inclined to approve this.
2. The name is not part of a normative section, and although this does change the Specification section, it is a minor wording change. In this regard, I am again inclined to approve this.
3. I recognize many people don't follow the official recommendation in 1, and that the citation itself will change. In that regard, I am not inclined to approve this.

@SamWilsn what do you think

sullof:
It makes sense. Still, calling the EIP with a strong reference to soulbound tokens, while that is just a specific use case, seems misleading to me.

Pandapip1 (changes requested):
The title shouldn't be changed. I think that the rest is fine.

SamWilsn:
The criteria for updating a final EIP:

> [...] only be updated to correct errata and add non-normative clarifications.

Even though the currently accepted definition of soulbound doesn't match exactly what this proposal provides, I don't think this qualifies as an errata or clarification.

TimDaub:
> The title shouldn't be changed. I think that the rest is fine.

That'd make it quite confusing IMO. If we didn't change the title but referenced SBTs as "lockable tokens" internally. I think it should either be decided that this patch is 100% taken or 100% declined (which is both fine with me).

> 2\. The name is not part of a normative section, and although this does change the Specification section, it is a minor wording change. In this regard, I am again inclined to approve this.

This is my view as well.

> 3\. I recognize many people don't follow the official recommendation in 1, and that the citation itself will change. In that regard, I am not inclined to approve this.

To be fair, a citation always has a date attached to it, so before date X you'd cite it as "Minimal Soulbound tokens" and after the date you cite it as "Lockable tokens." I see no issues with that.

TimDaub:
Btw. I think it's problematic that there is another proposal called "Lockable ...": https://eips.ethereum.org/EIPS/eip-5058

PR #6345, opened 2023-01-17
EIP-4844: Explicitly state sha256 is used

axic:


protolambda (approved):
LGTM. Using `hash` for `sha256` is standard practice everywhere in the consensus specs, but I can see how it is confusing in an execution layer context.

PR #6354, opened 2023-01-18
Update to EIP-5805

Amxx:


Amxx:
Disabled automerge to get a review by @frangio

EIPS/eip-5805.md
```diff
@@ -69,7 +69,29 @@ This function is thus optional, and its absence should be considered as a marker
...
+
+This function returns a string describing the operating mode the contract is operating on.
+
+- If operating using **block number**, then this function SHOULD be implemented and return `mode=blocknumber`.
+- If operating using **timestamp**, then this function SHOULD be implemented and return `mode=timestamp`.
+- If operating using any other mode, then this function MUST be implemented. In particular, if a contract on some L2 uses the blocknumber of the L1 chain it is anchored to (mainnet in our example), the mode should describe that as `mode=blocknumber&fromChainType=L1&fromChainId=1`.
+
+This function is thus optional, and its absence should be considered as a marker of the contract operating using block number or timestamp, which can both be clearly idenfitied from the absence of this function. (This makes this EIP compatible with pre-existing voting contracts).
+
+Full specification of the return string format is beyound the scope of this EIP. Nethertheless, this EIP recommands the strings returned by this function follow the grammar of URI fragments as described in RFC3986.
```
frangio, line 87: This should be a requirement IMO.

```suggestion
The return string MUST be formatted like a URL query string (a.k.a. `application/x-www-form-urlencoded`). This allows easy decoding in standard JavaScript with `new URLSearchParams(CLOCK_MODE)`.
```

EIPS/eip-5805.md
```diff
@@ -69,7 +69,31 @@ This function is thus optional, and its absence should be considered as a marker
...
+      type: uint48
+```
+
+### CLOCK_MODE
+
+This function returns a string describing the operating mode the contract is operating on.
+
+- If operating using **block number**, then this function SHOULD be implemented and return `mode=blocknumber&from=default`.
+- If operating using **timestamp**, then this function SHOULD be implemented and return `mode=timestamp`.
+- If operating using any other mode, then this function MUST be implemented. In particular, if a contract on some L2 uses the blocknumber of the L1 chain it is anchored to (mainnet in our example), the mode should describe that as `mode=blocknumber&from=eip155:1`.
```
frangio, line 81: I would move the L2 note to the block number item above.
```suggestion
- If operating using any other mode, then this function MUST be implemented and return a unique identifier for the encoded `mode` field.
```

EIPS/eip-5805.md
```diff
@@ -69,7 +69,31 @@ This function is thus optional, and its absence should be considered as a marker
...
     - name: timepoint
-      type: uint256
+      type: uint48
+```
+
+### CLOCK_MODE
+
+This function returns a string describing the operating mode the contract is operating on.
+
+- If operating using **block number**, then this function SHOULD be implemented and return `mode=blocknumber&from=default`.
```
frangio, line 79:
```suggestion
- If operating using **block number**:
  - If the block numbers are those of the `NUMBER` opcode (i.e. Solidity's `block.number`), then this function SHOULD be implemented and return `mode=blocknumber&from=default`.
  - If it is any other block number, then this function MUST be implemented and return `mode=blocknumber&from=<CAIP2ID>`, where `<CAIP2ID>` is a CAIP-2 Blockchain ID such as `eip155:1`.
```

EIPS/eip-5805.md
```diff
@@ -69,7 +69,31 @@ This function is thus optional, and its absence should be considered as a marker
...
+- If operating using **timestamp**, then this function SHOULD be implemented and return `mode=timestamp`.
+- If operating using any other mode, then this function MUST be implemented. In particular, if a contract on some L2 uses the blocknumber of the L1 chain it is anchored to (mainnet in our example), the mode should describe that as `mode=blocknumber&from=eip155:1`.
+
+This function is thus optional, and its absence should be considered as a marker of the contract operating using block number or timestamp, which can both be clearly idenfitied from the absence of this function. (This makes this EIP compatible with pre-existing voting contracts).
```
frangio, line 83: As I said above to simplify use I think timestamps should make the interface mandatory.

```suggestion
This function is thus optional, and its absence should be considered as a marker of the contract operating using block numbers, which can be clearly idenfitied from the absence of this function. (This makes this EIP compatible with pre-existing voting contracts).
```

EIPS/eip-5805.md
```diff
@@ -69,7 +69,31 @@ This function is thus optional, and its absence should be considered as a marker
...
+
+Note that when operating using **block number**, the `clock()` is expected to returns the value given by the `number` opcode (`0x43`). In some cases this can be the block number of another chain (in arbitrum, opcode `0x43` returns the block number of the last recorded operation on the parent chain). A contract can use `from=default` to specify that the block number used is the one provided by the `number` opcode (`0x43`). If a more explicit description is needed, CAIP-2 blockchain id should be used, as shown in the example above.
```
frangio, line 85: >as shown in the example above

Just as a note I don't think we should see this as an example, we should be fully specifying it.

EIPS/eip-5805.md
```diff
@@ -94,7 +118,11 @@ This function MUST be implemented
...
+This function returns the historical voting weight of an account. This corresponds to all the voting power delegated to it at a specific timepoint. The timepoint parameter MUST match the operating mode of the contract. This function SHOULD only serve past checkpoints, which SHOULD be immutable.
+
+- Calling this function with a timepoint that is greater or equal to `clock()` SHOULD revert.
+- Calling this function with a timepoint scrictly smaller than `clock()` SHOULD not revert.
+- For any timepoint that is strictly smaller than `clock()`, the value returned by `getPastVotes` SHOULD be constant. This means that for any call to this function that returns a value, re-executing the same call (at any time in the future) SHOULD return the same value.
```
frangio, line 125: "Timepoint" is not precisely defined, in particular nothing says all integers are valid timepoints, so I think it's better to specify "integer" here.

```suggestion
- For any integer that is strictly smaller than `clock()`, the value returned by `getPastVotes` SHOULD be constant. This means that for any call to this function that returns a value, re-executing the same call (at any time in the future) SHOULD return the same value.
```

EIPS/eip-5805.md
```diff
@@ -94,7 +118,11 @@ This function MUST be implemented
 
 #### getPastVotes
 
-This function returns the historical voting weight of an account. This corresponds to all the voting power delegated to it at a specific timepoint. The timepoint parameter should match the operating mode of the contract. This function SHOULD only serve past checkpoints that are immutable. Calling this function with a timepoint that is greater or equal to `clock()` SHOULD revert. For any timepoint that is strictly smaller than `clock()`, the value returned by `getPastVotes` should be constant.
+This function returns the historical voting weight of an account. This corresponds to all the voting power delegated to it at a specific timepoint. The timepoint parameter MUST match the operating mode of the contract. This function SHOULD only serve past checkpoints, which SHOULD be immutable.
+
+- Calling this function with a timepoint that is greater or equal to `clock()` SHOULD revert.
+- Calling this function with a timepoint scrictly smaller than `clock()` SHOULD not revert.
```
frangio, line 124:
```suggestion
- Calling this function with a timepoint scrictly smaller than `clock()` SHOULD NOT revert.
```

EIPS/eip-5805.md
```diff
@@ -375,6 +403,10 @@ Delegation allows token holders to trust a delegate with their vote while keepin
...
 
+`clock` returns `uint48` as it is largelly sufficient for storing realistic values. In timestamp mode, `uint48` will enough until the year 8921556. Even in block number mode, with 10,000 blocks per seconds, it would be enough until the year 2861. Using a type smaller than uint256 allows some storage packing of timepoints with other associated values. Greatly reducing the cost of writting and reading from storage. Depending on evolution of the blockchain (particularly layer twos), `uint32` might cause issues fairly quickly. On the other hand, anything bigger than `uint48` is overkill.
```
frangio, line 406:
```suggestion
`clock` returns `uint48` as it is largelly sufficient for storing realistic values. In timestamp mode, `uint48` will be enough until the year 8921556. Even in block number mode, with 10,000 blocks per seconds, it would be enough until the year 2861. Using a type smaller than uint256 allows some storage packing of timepoints with other associated values. Greatly reducing the cost of writting and reading from storage. Depending on the evolution of the blockchain (particularly layer twos), `uint32` might cause issues fairly quickly. On the other hand, anything bigger than `uint48` is overkill.
```

PR #6358, opened 2023-01-19
Add EIP-6358: Omniverse Token Protocol

xiyu1984:
Hi everyone, 
we are a team focusing on multi-chain interoperability for years, and we have some experience and open-source contributions in this field. We are trying to help smart contracts deployed on different blockchains interact with each other. When we came to our token's technical form, we found some problems.  
In the process of R&D, we found that the fragmentation of tokens is a common and disturbing problem among chains and L2s. Although assets-bridges can more or less make it, 
- It will be better for tokens(no matter fungible or non-fungible) to be treated as a whole instead of being divided into different parts on different blockchains. 
- In addition, we do not want to receive a loss of assets even in some extreme cases, such as when a blockchain we are serving suddenly fails.

Thanks for your time~


xiyu1984:
> #6366

Hey @Gesisfm,
Thanks for your suggestion, but I think I have not found relationships between `Omniverse DLT` and #6366. Could you please give us some more details of your concern? And we will really appreciate it.

By the way, I think the following details may make it clearer and easier to understand:

image

O-DLT could be considered as a **decentralized global ledger** established upon two parts: 
- **Abstract Nodes** are smart contracts or similar mechanisms running at the application level over existing blockchains.  The states recorded by the **Abstract Nodes** that are deployed on different blockchains respectively could be considered as copies of the global state, and they are ultimately consistent. 
- **Synchronizer** is an off-chain procedure responsible for carrying published omniverse transactions from the O-DLT smart contracts on one blockchain to the O-DLT smart contracts of the others. They are absolutely trustless and details could be found [here](https://github.com/ethereum/EIPs/blob/5ac749d918b9d1dbdd0120fd344c11867c321da1/EIPS/eip-6358.md#workflow).

Pandapip1:
> 



> By the way, I think the following details may make it clearer and easier to understand:
> 
> image
> 
> O-DLT could be considered as a **decentralized global ledger** established upon two parts:
> 
>     * **Abstract Nodes** are smart contracts or similar mechanisms running at the application level over existing blockchains.  The states recorded by the **Abstract Nodes** that are deployed on different blockchains respectively could be considered as copies of the global state, and they are ultimately consistent.
> 
>     * **Synchronizer** is an off-chain procedure responsible for carrying published omniverse transactions from the O-DLT smart contracts on one blockchain to the O-DLT smart contracts of the others. They are absolutely trustless and details could be found [here](https://github.com/ethereum/EIPs/blob/5ac749d918b9d1dbdd0120fd344c11867c321da1/EIPS/eip-6358.md#workflow).

If this is not included in the Spec, I highly recommend you include it.

(@eth-bot rerun)

xiyu1984:
Hi @poojaranjan ,

I think I missed the meeting. Is there any chance to have a meeting once more?

xiyu1984:
Hi @Pandapip1 ,

Thanks for your suggestion, and we appreciate it.

I have added the description to the proposal doc.

All the best~

EIPS/eip-6358.md
```diff
@@ -0,0 +1,868 @@
...
+The core meaning of Omniverse is that the ***legitimacy of all on-chain states and operations can be equivalently verified and consistently recorded over different consensus spaces, regardless of where they were initiated.***  
+O-DLT works at an application level, which means everything related is processed in smart contracts or similar mechanisms, just as the [ERC-20](./eip-20.md)/[ERC-721](./eip-721.md) did.  
+
+![img](../assets/eip-6358/img/o-dlt.png)
+
+O-DLT could be considered as a decentralized global ledger established upon two parts:
+
+- **Abstract Nodes** are smart contracts or similar mechanisms running at the application level over existing blockchains. The states recorded by the **Abstract Nodes** that are deployed on different blockchains respectively could be considered as copies of the global state, and they are ultimately consistent. 
+- **Synchronizer** is an off-chain procedure responsible for synchronizing published omniverse transactions from the O-DLT smart contracts on one blockchain to the O-DLT smart contracts of the others.
```
Pandapip1, line 24: Much the same thing. Slim this down.

Pandapip1 (changes requested):
I'm somewhat struggling to understand how this ERC works. Mind breaking it down for me a little more?

xiyu1984:
> I'm somewhat struggling to understand how this ERC works. Mind breaking it down for me a little more?

Hi @Pandapip1 ,

Thanks for your suggestion, and I will explain some details first and then fix the docs. Maybe the fixing work will take me some time and I will contact you as soon as we finish it.  

## Details
Generally, this ERC-6358 targets to improve the interoperability between Ethereum and other blockchains in a brand new way. This is much unlike the current token bridges paradigm, whose drawbacks are as follows:
- Token bridges split, but we want our token to be treated as a whole instead of divided into fragmented parts on different public chains.
- Token bridges may make our token disappear, that is, when a certain chain breaks down, if we have transferred ETHs to it through token bridges, we will lose them.

Simply, we want to solve these problems at the protocol level for tokens. The key point is ***synchronization*** instead of ***bridging***. 

We know that DLT (Decentralized Ledger Technology) is one of the basements of the Web3 world, with which the states of our assets(native tokens and ERC tokens) are recorded. The recorders are off-chain nodes (procedures actually).  
Instead of off-chain procedures, Omniverse-DLT (*Let’s temporarily call it this name, and I’m thinking about its new name according to your suggestions. Will `Omniverse Token Protocol` be better?*) leverages on-chain smart contracts deployed on different blockchains to be the `recorders` (we call them `Abstract Nodes` in the [picture above](https://raw.githubusercontent.com/ethereum/EIPs/40057636cba738a2a3607859a365f7d39e09552e/assets/eip-6358/img/o-dlt.png), which is compared to `Nodes` for normal blockchains). As smart contracts are built over existing consensus spaces, the states recorded on them can be considered trusted.   

Creating a new Omniverse Token means deploying related `ERC-6358` smart contracts on different blockchains. **The core part is that all the transactions of an Omniverse Token are synchronized among all its smart contracts deployed on different blockchains**. Specifically, the [data structure](https://github.com/ethereum/EIPs/blob/40057636cba738a2a3607859a365f7d39e09552e/EIPS/eip-6358.md#data-structure) of an Omniverse transaction (`o-transaction` for short) is defined *at the application level*. I will explain how this works with an intuitive example later.  

One problem is that smart contracts cannot execute all by themselves, and this is why we call them `Abstract Nodes`. And smart contracts deployed on different blockchains cannot interact with each other directly, so the key point is how to make ***synchronization***, and the `synchronizers` are brought in.  
The `synchronizers` are off-chain procedures responsible for carrying `o-transactions` from one chain to another if they found any difference between `Omniverse Token` smart contracts on different chains. Unlike off-chain nodes for normal blockchains, **there’s no need for them to work with a consensus**. This will be explained in the following.     

Next, I will use an intuitive example to explain the details of the principles.  
### Interpretation
- Every user has an [Omniverse Account](https://github.com/ethereum/EIPs/blob/40057636cba738a2a3607859a365f7d39e09552e/EIPS/eip-6358.md#omniverse-account) (`o-account` for short) to operate their assets of an Omniverse Token.
- An `account nonce` is used to record the current `o-transaction` sequence number of an `o-account`.
- A `nonce in o-transaction` is used to describe the sequence number of an `o-transaction`.  

For example, if the `account nonce` of account `A` is $k$, the `nonce in o-transaction` in the next `o-transaction` of `A` will be $k+1$.
### Example Workflow
- For a certain Omniverse Token (`token-OT` for example), suppose a common user `A` and the related operation `account nonce` is $k$.
- `A` initiates an omniverse transfer operation of `token-OT` on Ethereum by calling `omniverse_transfer`. The current `account nonce` of `A` in the `token-OT` smart contracts deployed on Ethereum is $k$ so the valid value of `nonce in o-transaction` needs to be $k+1$.  
- The `token-OT` smart contracts on Ethereum verify the signature of the o-transaction data. If the verification for the signature and data succeeds, the o-transaction data will be published on the `token-OT` smart contracts of the Ethereum side. The verification for the data includes:
    - whether the amount is valid
    - and whether the `nonce in o-transaction` is 1 larger than the `account nonce` maintained by the on-chain `token-OT` smart contracts on Ethereum.
- Now, `A`'s newest submitted `nonce in o-transaction` on Ethereum is $k+1$, but still $k$ on other chains.
- The off-chain synchronizers find the newly published o-transaction, and they will find the `nonce in o-transaction` is larger than the related `account nonce` on other chains.
- The following behavior of the `synchronizers` depends on certain implementations. For example, a reward mechanism could be designed and these synchronizers will rush to deliver this message because whoever submits to the destination chain first will get a reward. No matter how to design the synchronizers, one thing is clear, because the synchronizers only carry `o-transaction` data with other users’ signatures, they cannot cheat.
- Finally, the `token-OT` smart contracts deployed on other chains will all receive the `o-transaction` data, verify the signature and execute it when a *waiting time is up*. After execution, the underlying `account nonce` will add 1. Now all the `account nonce` of account `A` will be $k+1$, and the state of the balances of the related account will be the same too.

### Attack Vector Analysis

We have described the details of the [Attack Vector Analysis](https://github.com/ethereum/EIPs/blob/40057636cba738a2a3607859a365f7d39e09552e/EIPS/eip-6358.md#attack-vector-analysis) in the proposal. In general:  
- It's none of the **synchronizer**'s business as **they cannot create other users' signatures** unless some **common users** tell him, but at this point, we think it's a problem with the role **common user**.
- For **common users**, **maybe they will cheat**, but fortunately, **they can't succeed**.

The key point to prevent **common users**’s cheating is **Verifiable Fraud Verification**, the details of which have [already been presented in the proposal](https://github.com/ethereum/EIPs/blob/40057636cba738a2a3607859a365f7d39e09552e/EIPS/eip-6358.md#will-the-common-user-cheat).  

### Implementation

Lastly, as one `o-transaction` will be submitted onto smart contracts on different blockchains, one important concern might be the gas cost.  

From the implementation side, besides the [implementation example](https://github.com/Omniverse-Web3-Labs/omniverse-evm/tree/main/contracts) submitting and synchronizing `o-transactions` one by one, we are building a new version using ZK-SNARK to make batched submissions, which is similar to the zk-rollup technology but is for the synchronization of multi-chains. Actually, we call this technology **ZK-Synchronization**, and because this is not a short story, I think we can discuss it at another time if you are interested.

Pandapip1:
Bridges refer to anything that can pass a message from one chain to another, iirc. I was less confused by the concept of this EIP (cross-chain state sync) and was more confused about how this is meant to function.

My understanding of this is that any time a user transfers tokens on one chain, tokens are transferred on all other supported chains? If so, does this completely negate any scalability advantages that L2s are meant to provide?

xiyu1984:
Hi @Pandapip1 ,

I agree with your definition of the `Bridges`. And under this definition, maybe what I meant about `bridges` in the last reply is actually `assets bridges`.  

About your questions, 

Firstly, I think it’s just the meaning that “any time a user transfers tokens on one chain, tokens are transferred on all other supported chains”. In [the certain implementation](https://github.com/Omniverse-Web3-Labs/omniverse-evm/tree/main/contracts) that is provided by us, the validity could be guaranteed by a mechanism we called `Verifiable Fraud Proof`. We have analyzed the details in the [Attack Vector Analysis](https://github.com/ethereum/EIPs/blob/40057636cba738a2a3607859a365f7d39e09552e/EIPS/eip-6358.md#attack-vector-analysis) of the proposal.

Secondly, regarding scalability, in all sincerity, thanks for your experience-filled question, and I will use this opportunity to explain some of the details related to it. In fact, this is exactly why I mentioned that batched submissions could be made by ZK-SNARK in the last reply.  
In general, we think this will not only **not** negate any of the scalability advantages that L2s are meant to provide but will also provide similar advantages.  
**But specifically, it depends on certain implementation ways.**  
Recall that a common user can submit an `o-transaction` to `omniverse token` (temporary name) smart contracts deployed on any chain they like, and then the synchronizers make the states synchronization.  
If we implemented it in the way I described in the example workflow of the [last reply](https://github.com/ethereum/EIPs/pull/6358#issuecomment-1445289076), that is, the states are synchronized from one chain to another one by one, then there is indeed no scalability.  
But actually, if we do not synchronize the transactions of an `omniverse token` from one chain to another immediately but synchronize a batch of transactions for a time, and a ZK-Proof is brought in to prove the validity of transactions in the batch, things will become more efficient. The proving process can be implemented similarly to the `ZK-Rollup` of the L2s, and we call it `ZK-Synchronization`. Details of one possible implementation might be as follows:  
- Different users may submit `o-transactions` on different public chains, so a synchronizer collects all the `o-transactions` appearing on any blockchain in a time window. 
- The synchronizer `rollup` a batch of `o-transactions` in a similar way as the `ZK-Rollup`, and then submits the new state with a ZK-Proof.
- On-chain smart contracts of the `omniverse token` on all deployed chains verify the proof and update the states.
So I think the advantages that L2s are meant to provide can still be get by `omniverse token protocol`. Although the underlying technology is similar to the `ZK-Rollup`, there are still some differences between the `omniverse token protocol` and L2s:  
- There’s no need for `omniverse tokens` to be locked on L1s, which prevents the fragment problems of L2s.
- The purposes are different. L2s specializes in serving Ethereum, but the `omniverse token protocol` helps connect Ethereum to other chains.
- They work at different layers. Omniverse Token Protocol is an application-layer protocol that can also work on L2s, breaking down the barrier between L2s.

Pandapip1 (approved):
I'll merge this so that it can be tracked in the repository, but I would like to see at a minimum:

1. The paragraphs edited to remove redundancy
2. A new title (how about "Cross-Chain Balance Synchronization"?)

xiyu1984:
> I'll merge this so that it can be tracked in the repository, but I would like to see at a minimum:
> 
> 1. The paragraphs edited to remove redundancy
> 2. A new title (how about "Cross-Chain Balance Synchronization"?)

Hi @Pandapip1 ,

Thank you very much for your advice.

I will remove the redundancy as soon as possible according to your suggestion. 
I agree with the name you provided, and maybe I will keep thinking about the names. Although I'm not sure I can make out a better name than "Cross-Chain Balance Synchronization".

All the best~

PR #6372, opened 2023-01-25
Add EIP-6372: Contract clock

Amxx:
Following discussion on EIP-5805, we decided to spin of clock part of the interface as its own EIP.

EIP-5805 will add a dependency on this once merged.


Pandapip1 (approved):
Not sure how useful this will actually end up being, but this LGTM for a draft.

illuminated-shade (commented):
EIPS/eip-6372.md

PR #6386, opened 2023-01-26
Update EIP-4844: do not allow creation txs

lightclient:
this builds on https://github.com/ethereum/EIPs/pull/6985, so that PR should be merged first

--

Allowing contract creation from 4844 transactions is unnecessary. If in the unlikely event that there is a use case for creating a contract while also sending blobs to the chain, the transaction should simple do it via `CREATE` or `CREATE2`.


lightclient:
closed via #7038

PR #6389, opened 2023-01-27
Update ERC-5005: Move to Last Call

auryn-macmillan:
This PR changed ERC-5005 to Last Call


SamWilsn:
You'll need to set a `last-call-deadline` date, which should be about two weeks from when we merge the PR.

PR #6404, opened 2023-01-31
Add EIP: SSZ transactions root

etan-status:

 - The build passes.


EIPS/eip-draft_ssz.md
```diff
@@ -0,0 +1,1029 @@
+---
```
Pandapip1, line 1:
```suggestion
---
eip: 6404
```
Make sure to change the filename, too!

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
+---
+eip: 6404
+title: Transition to SSZ
```
lightclient, line 3: I think this title could be more descriptive

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+description: Unifies the CL and EL block representations
```
lightclient, line 4: This doesn't exactly describe what the EIP does, would rather see something like "Migrates to SSZ hash tree root commitment for transactions, receipts, and withdrawals".

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-01-30
+requires: 155, 658, 1559, 2718, 2930, 4844, 4895
+---
+
+## Abstract
+
+This EIP aligns the definition of the EL block with the CL `ExecutionPayload` structure.
```
lightclient, line 16: EL and CL should be spelled out in the first occurrences

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+
+## Motivation
+
+While the CL `ExecutionPayload` and the EL block structure map to each other, they are differently encoded. This EIP aims to align the encoding between them, taking advantage of the more modern SSZ format used by the CL. This brings several advantages:
```
lightclient, line 20:
```suggestion
While the CL `ExecutionPayload` and the EL block structure map to each other, they are encoded differently. This EIP aims to align the encoding between them, taking advantage of the more modern SSZ format used by the CL. This brings several advantages:
```

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+
+1. **Reducing complexity:** Merkle-Patricia Tries (MPT) are hard to work with. Replacing them with SSZ leaves only the state trie in the legacy MPT format.
+
+2. **Better for smart contracts:** The SSZ format is optimized for production and verification of merkle proofs. It allows proving specific fields of containers and allows chunked processing, e.g., to support handling transactions that do not fit into calldata.
+
+3. **Better for light clients:** Light clients with access to the CL `ExecutionPayload` no longer need to obtain the matching EL block header to verify proofs rooted in `transactions_root` or `withdrawals_root`.
```
lightclient, line 26: or receipts_root?

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+    kzg_aggregated_proof: KZGProof
+
+class TransactionPayload(Container):
+    signed_tx: SignedTransaction
+    blob: Optional[BlobTransactionExtension]  # EIP-4844
+```
+
+On the network, `TransactionPayload` is preceded by a single byte, [`BLOB_TX_TYPE`](./eip-4844.md#parameters) to distinguish from other transaction formats according to [EIP-2718](./eip-2718.md), regardless of the inner `signed_tx.hash_version`.
+
+The `blocks.transactions` list remains RLP encoded:
```
lightclient, line 132: Where is `blocks` defined? Also, I don't think this is necessary. How the list of transactions is encoded is not in consensus, only defined in the devp2p spec.

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+```python
+class ExecutionPayload(Container):
+    ...
+    transactions: List[SignedTransaction, MAX_TRANSACTIONS_PER_PAYLOAD]  # Formerly, `Transaction`
+    ...
+```
+
+#### Execution block header changes (Transactions)
+
+The execution block header's `transactions_root` is updated to the `transactions_root` definition of the [consensus-specs `ExecutionPayloadHeader`](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/eip4844/beacon-chain.md#executionpayloadheader).
```
lightclient, line 160: I'm not sure the reason for linking to consensus specs here. This link is also not updated with `transactions_root` as defined in this EIP. I think the link should be omitted and it should simply be stated that the two roots are now in alignment.

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+
+To compute the new `transactions_root`, the list of individual `SignedTransaction` containers is represented as a SSZ `List`.
+
+```python
+block_header.transactions_root == hash_tree_root(List[SignedTransaction, MAX_TRANSACTIONS_PER_PAYLOAD](
+    transaction_0,
+    transaction_1,
+    transaction_2,
+    ...
+)) == hash_tree_root(List[Root, MAX_TRANSACTIONS_PER_PAYLOAD](
```
lightclient, line 170: isn't this equality a given based on the logic of HTR? it seems unnecessary to explicitly call out the recursive merkleization

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+            get_transaction_sighash(signed_tx.tx, signed_tx.hash_version),
+            1 if signed_tx.signature.y_parity else 0,
+            signed_tx.signature.r,
+            signed_tx.signature.s,
+        )
+
+    if hash_version >= TX_HASH_VERSION_EIP155:
+        return ecrecover(
+            get_transaction_sighash(signed_tx.tx, signed_tx.hash_version),
+            (1 if signed_tx.signature.y_parity else 0) + tx.chain_id * 2 + 35,
```
lightclient, line 372: I think ecrecover is supposed to operate on a signature that has the chain id removed?

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+| - | - | - |
+| `TxHashVersion` | `uint8` | Original hashing method used for signing the transaction
+
+| Name | Value | Description |
+| - | - | - |
+| `TX_HASH_VERSION_LEGACY` | `TxHashVersion(0x00)` | Replayable [`LegacyTransaction`](./eip-2718.md#transactions) |
+| `TX_HASH_VERSION_EIP155` | `TxHashVersion(0x01)` | [EIP-155](./eip-155.md) `LegacyTransaction` |
+| `TX_HASH_VERSION_EIP2930` | `TxHashVersion(0x02)` | [EIP-2930](./eip-2930.md#definitions) transaction |
+| `TX_HASH_VERSION_EIP1559` | `TxHashVersion(0x03)` | [EIP-1559](./eip-1559.md#specification) transaction |
+| `TX_HASH_VERSION_EIP4844` | `TxHashVersion(0x04)` | [EIP-4844](./eip-4844.md#parameters) transaction |
```
axic, line 70: 4844 has `TransactionType = 0x05`. Shouldn't the `TxHashVersion`s here mirror the `TransactionType`?

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+| Name | SSZ equivalent | Description |
+| - | - | - |
+| `TxHashVersion` | `uint8` | Original hashing method used for signing the transaction
+
+| Name | Value | Description |
+| - | - | - |
+| `TX_HASH_VERSION_LEGACY` | `TxHashVersion(0x00)` | Replayable [`LegacyTransaction`](./eip-2718.md#transactions) |
+| `TX_HASH_VERSION_EIP155` | `TxHashVersion(0x01)` | [EIP-155](./eip-155.md) `LegacyTransaction` |
+| `TX_HASH_VERSION_EIP2930` | `TxHashVersion(0x02)` | [EIP-2930](./eip-2930.md#definitions) transaction |
+| `TX_HASH_VERSION_EIP1559` | `TxHashVersion(0x03)` | [EIP-1559](./eip-1559.md#specification) transaction |
```
axic, line 69: Similarly, this has `TransactionType = 0x02`.

EIPS/eip-6404.md
```diff
@@ -0,0 +1,1000 @@
...
+
+This EIP aligns the definition of the EL block with the CL `ExecutionPayload` structure.
+
+## Motivation
+
+While the CL `ExecutionPayload` and the EL block structure map to each other, they are differently encoded. This EIP aims to align the encoding between them, taking advantage of the more modern SSZ format used by the CL. This brings several advantages:
+
+1. **Reducing complexity:** Merkle-Patricia Tries (MPT) are hard to work with. Replacing them with SSZ leaves only the state trie in the legacy MPT format.
+
+2. **Better for smart contracts:** The SSZ format is optimized for production and verification of merkle proofs. It allows proving specific fields of containers and allows chunked processing, e.g., to support handling transactions that do not fit into calldata.
```
axic, line 24: Is there any evidence that SSZ is better for proofs **within smart contracts** than RLP?

EIPS/eip-6404.md
axic: Not having it mirror will result in:
- the requirement to specify both `TxHashVersion` and `TransactionType` in new EIP-2718 extensions (like EIP-4844),
- the potential for easy mistakes in implementations.

I understand the reason it can't easily be the same, because we want to separate legacy and EIP-155, but maybe there's a solution somewhere.

EIPS/eip-6404.md
etan-status: As far as I know, RLP does not define a merkle tree, it is only a serialization format.
SSZ allows you to prove individual items such as a transaction's `value`. With RLP, I think that the current approach is to feed the entire serialized transaction into the verifier. Which then brings the next problem, that an RLP transaction may not necessarily fit into calldata.

EIPS/eip-6404.md
etan-status: The `receipts_root` is represented as an MPT across both CL and EL. This discrepancy only exists for `transactions_root` and `withdrawals_root`.

EIPS/eip-6404.md
etan-status: `TxHashVersion` and EIP-2718 transaction types, are, conceptually, different.

`TxHashVersion` solely indicates the hashing method that was originally used to sign the transaction and to derive it's perpetual ID. There is no meaning beyond that, it is not even part of the `Transaction` but instead a property of the `SignedTransaction`.

Multiple EIP-2718 tx types may map to a single `TxHashVersion`:
- EIP-2718 tx type 5 maps to `TxHashVersion` 4
- EIP-2718 tx type 2 maps to `TxHashVersion` 3
- EIP-2718 tx type 1 maps to `TxHashVersion` 2
- EIP-2718 tx types 0xc0 through 0xef map to `TxHashVersion` 1 or 0, depending on the signature's `v` value
- EIP-2718 tx types 0, 3, 4, 0x80 through 0xbf, and 0xff are reserved

I agree that it would be cool to have a 1:1 mapping from EIP-2718 to `TxHashVersion`. Maybe there is a way to at least ensure that the 1/2/5 values match? But in the end, it is something that can be tested very easily.

axic: I know they are different conceptually, the question is if we can avoid making them different to avoid future confusion.

> it is something that can be tested very easily.

Everything in the system can be tested, yet we see numerous consensus bugs popping up in precompiles which have been live for years 🤓

EIPS/eip-6404.md
etan-status: That's right, I'm referring to that `block` from devp2p.
https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-encoding-and-validity

And agree, the `transactions` are already specified to pick up the `TransactionPayload` format based on the EIP-2718 prefix. So, mentioning it here is indeed not necessary.

EIPS/eip-6404.md
etan-status: The definition of `execution_payload_header.transactions_root = hash_tree_root(execution_payload.transactions)` is still the correct definition. But, yes, it can be confusing because under the link, `execution_payload.transactions` is still defined based on a different format than the proposed one. 

Removed the link and updated language. This also makes it clearer that for `withdrawals_root`, the existing format is actually reused (as opposed to `transactions_root` and `receipts_root` that change in both execution and consensus worlds).

EIPS/eip-6404.md
etan-status: Yes, this is just a property of HTR. Removing from the spec section.

EIPS/eip-6404.md
etan-status: Oh, that makes this function obsolete then. The samples I found about `ecrecover` were using the `v` value, so I thought that I need to reconstruct `v` here. Removing `get_transaction_signer`.

The logic inside `get_transaction_id` is still alright, though?

EIPS/eip-6404.md
etan-status: Reduced the scope to not include conversion of block hash, and went with more descriptive title

EIPS/eip-6404.md
etan-status: That's true as well 😅

The only function depending on ordering is `validate_transaction_payload`. If that property should be preserved, we could either change the `BLOB_TX_TYPE` in EIP-4844 to `0x04`, or, alternatively, `TxHashVersion` could start from `1` instead of `0`. This would then at least ensure that going forward, `TxHashVersion` matches EIP-2718 transaction type, essentially ensuring that there is no ongoing maintenance burden from this.

Alternatively, if `TxHashVersion` ordering is not desirable, we could use values 3 and 4 to indicate EIP-155 and legacy, so that 1 and 2 can remain EIP-2930 / EIP-1559.

etan-status: Just checked JSON-RPC. There, it seems that type `0x00` is indicated for both legacy and EIP-155 transactions.
Is there a spare, reserved, `chain_id` value that could be used to distinguish explicit replayability (EIP-155 `chain_id=0`) from implicit replayability (pre-EIP-155 tx)?

Technically, (uint64) `0x7ffffffffffffffc` ends up at 27/28 when applying `v = chain_id * 2 + 35 + {0,1}`.
But seems quite ugly, as well.

In JSON-RPC, the output is the `v` value, not `y_parity`. That could also help, but is kinda ugly as it leaks chain ID specifics into a general-purpose `ECDSASignature` type.


JSON-RPC examples:

```json
[
      {
        "accessList": [],
        "blockHash": "0xeeb1a6c504394b835b06eda9288446d4739d2f70290f52e904be9b550347a2a1",
        "blockNumber": "0x2b0213",
        "chainId": "0xaa36a7",
        "from": "0xd123b238958ea02934e82436c5248dae5501ffd9",
        "gas": "0x5208",
        "gasPrice": "0x59682f07",
        "hash": "0xad0a532e17955b56d42663fab13f90c8310b3003a9ea706387803eba679d2c54",
        "input": "0x",
        "maxFeePerGas": "0x59682f0e",
        "maxPriorityFeePerGas": "0x59682f00",
        "nonce": "0x144e",
        "r": "0xbaba866770cfdb06cd1e912a3947c11e87d5d699489533c14a946515002f0ebc",
        "s": "0x4d9a0246253634bab7dd4663fa2863251603b966d2a05e938b48a067a8071f58",
        "to": "0xd0266a7e4f388c6e6f0cf11d2459624870acc7c2",
        "transactionIndex": "0x4",
        "type": "0x2",
        "v": "0x1",
        "value": "0x16345785d8a0000"
      },
      {
        "blockHash": "0xeeb1a6c504394b835b06eda9288446d4739d2f70290f52e904be9b550347a2a1",
        "blockNumber": "0x2b0213",
        "chainId": "0xaa36a7",
        "from": "0xe2a3422f3168149ad2d11b4de2b97b05f1ebf76e",
        "gas": "0x13880",
        "gasPrice": "0x3b9aca00",
        "hash": "0x1d7e73fc2002a9b02cbab28bd295540f982a16ed10e2a75fc804ff1b59982183",
        "input": "0xa9059cbb000000000000000000000000e7d14b98386cb0752dfca69151f85c91eaa87b540000000000000000000000000000000000000000000000000de0b6b3a7640000",
        "nonce": "0x10409",
        "r": "0x804d47cf511fe2f4c2503e2aeeeab02353cad51566c568d2a3cb97d17cff619e",
        "s": "0x176e47941e0c57d7f0898138b165bf8cf3d29a2b9170b63dc207e0cfef235ea3",
        "to": "0xed43f81c17976372fcb5786dd214572e7dbb92c7",
        "transactionIndex": "0x5",
        "type": "0x0",
        "v": "0x1546d71",
        "value": "0x0"
      }
]
```

EIPS/eip-6404.md
etan-status: Updated the title to be more specific.

EIPS/eip-6404.md
etan-status: I have updated the EIP to use matching 0x01 / 0x02 / 0x05 EIP-2718 tx types, by using 0x03 / 0x04 for the legacy types. I have put a section about reserved types, also mentioning the 0x00 value on JSON-RPC to prevent future accidents.

Adriana1960 (approved):

SamWilsn:
Not-so-subtle reminder to keep technical topics on the `discussions-to` thread and only put blocking comments here on the PR.

EIPS/eip-6404.md
```diff
@@ -0,0 +1,813 @@
...
+
+### SSZ
+
+#### `Optional[T]`
+
+A new [SSZ type](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/ssz/simple-serialize.md) is introduced to represent `Optional[T]` values.
+
+- If value is `None`, serialize as `[]`, and merkleize as `List[T, 1]` with length `0`.
+- If value is not `None`, serialize as `T`, and merkleize as `List[T, 1]` with length `1`.
+- Serialize `Optional[T]` as variable-size object (create offset-table entry in enclosing containers).
```
dapplion, line 42: Should T be allowed to have min_size 0 (for example `Optional[List[uint8, 1]]`) ? If so its deserialization could result in two different values

EIPS/eip-6404.md
```diff
@@ -0,0 +1,813 @@
...
+| `TransactionType` | `uint8` | [EIP-2718](./eip-2718.md) transaction type, range `[0x00, 0x7F]` |
+
+| Name | Value | Description |
+| - | - | - |
+| `TRANSACTION_TYPE_LEGACY` | `TransactionType(0x00)` | [`LegacyTransaction`](./eip-2718.md#transactions) (only used in SSZ) |
+| `TRANSACTION_TYPE_EIP2930` | `TransactionType(0x01)` | [EIP-2930](./eip-2930.md#definitions) transaction |
+| `TRANSACTION_TYPE_EIP1559` | `TransactionType(0x02)` | [EIP-1559](./eip-1559.md#specification) transaction |
+| `TRANSACTION_TYPE_EIP4844` | `TransactionType(0x05)` | [EIP-4844](./eip-4844.md#parameters) transaction |
+
+For transactions with `TRANSACTION_TYPE_LEGACY`, post-[EIP-155](./eip-155.md) legacy transactions with chain ID encoded into the `v` value of the signature need to be distinguishable from `LegacyTransaction` that lack chain ID. For that purpose, a `TransactionSubtype`is defined.
```
dapplion, line 65:
```suggestion
For transactions with `TRANSACTION_TYPE_LEGACY`, post-[EIP-155](./eip-155.md) legacy transactions with chain ID encoded into the `v` value of the signature need to be distinguishable from `LegacyTransaction` that lack chain ID. For that purpose, a `TransactionSubtype` is defined.
```

SamWilsn (approved):

EIPS/eip-6404.md
etan-status: Will follow up about this with @zah, technical discussion should be in discussions-to thread.

etan-status: @dapplion SSZ discussion split off into EIP-6475: https://ethereum-magicians.org/t/eip-6475-ssz-optional/12891

`Optional[List[...]]` is an illegal type. Use an empty list instead to represent that.

PR #6405, opened 2023-02-01
Website: Move last call rss feed to rss folder

Pandapip1:
Pretty self-explanatory.


SamWilsn:
This breaks anything that currently uses the RSS feed?

SamWilsn:
Do you also need to update https://github.com/ethereum/EIPs/blob/611981eb554a9669052cdf08f2cb85afc87a3e94/last-call.xml#L10?

Pandapip1:
> This breaks anything that currently uses the RSS feed?

Yes.

Pandapip1:
Oops didn't mean to close.

Pandapip1:
> Do you also need to update 

Yes

SamWilsn:
> > This breaks anything that currently uses the RSS feed?
> 
> Yes.

@poojaranjan are you aware of anything that depends on the RSS feed?

poojaranjan:
@SamWilsn 
> 
> @poojaranjan are you aware of anything that depends on the RSS feed?

People mentioned subscribing to RSS feeds and we were planning to add that on ECH website.  But, we haven't started working on it. 
Nothing comes to my mind that could be dependent on these RSS feeds.

SamWilsn (approved):

PR #6444, opened 2023-02-05
Update: EIP-6093: improve wording

PaulRBerg:
Several language corrections and improvements for EIP-6093.

Most edits proposed in this PR should be uncontroversial, but I want to share the rationale behind my suggestion to say "greater than or equal to" instead of "equal or greater than". There are two reasons:

1. This is the formal definition ([reference](https://www.cuemath.com/numbers/greater-than-or-equal-to/)).
2. Closer to the actual mathematical sign `>=` (the sign starts with "greater than")

Cc @ernestognw.


ernestognw (commented):
Thanks for taking a look @PaulRBerg!
Agree with the corrections, and the _greater than or equal_ to seems accurate and clearer.

Approving!

ernestognw (approved):

PR #6451, opened 2023-02-06
Update EIP-1328 from Stagnant to Review

pedrouid:
This EIP specifies a URI format for WalletConnect v1.0 and v2.0 and it's widely used in production already


SamWilsn (approved):
If this is ready and in production, feel free to open a PR to move to Last Call (then Final.)

PR #5703, opened 2022-09-21
Adds @xinbenlv as an EIP Editor (#5502)

SamWilsn:
Reopening discussion on #5502

PR #6477, opened 2023-02-09
Add myself to EIP-1 list

xinbenlv:
A follow-up of #5703


gcolvin (commented):
Welcome aboard!

Pandapip1 (approved):

xinbenlv:
@Pandapip1 @SamWilsn thx! By the way, I encountered a mdlint error, but seems special, in the context it's a demonstration of rendering, could you help advice how to fix this mdlint error?

SamWilsn:
> could you help advice how to fix this mdlint error?

We should turn off that error. I think it's the only way to get a code block to appear inside the footnote list.

PR #6484, opened 2023-02-09
Update EIP-5805: Refer to EIP-6372 for clock and move to Review

frangio:


frangio:
@Amxx Please review.

Amxx (dismissed):

EIPS/eip-5805.md
```diff
@@ -52,51 +52,13 @@ Both timestamps, block numbers, and other possible modes use the same external i
 
 ### Methods
 
-#### clock
+#### [ERC-6372](./eip-6372.md): clock and CLOCK_MODE
 
-This function returns the current timepoint. It could be `block.timestamp`, `block.number` (or any other **non-decreasing** function) depending on the mode the contract is operating on.
+A contract implementing this ERC may optionally implement ERC-6372 (Contract clock) to announce the clock that is used for vote tracking.
```
Amxx, line 57:
```suggestion
A contract implementing this ERC SHOULD implement ERC-6372 (Contract clock) to announce the clock that is used for vote tracking.
```

frangio: The rest of the ERC isn't using MUST, SHOULD, etc.

If we want to use those we need a larger review. Let's do that in another PR.

frangio: Oh, my bad I see that it is using those keywords. Just need to add the notice about them.

xinbenlv:
Happy to see clock is split out from this EIP. 

I notice the use of 6372 was "SHOULD" instead of "MUST", in which case when implementor choose to omit 6372, do you want to consider designate a "DEFAULT clock type" so compliants still know what to do with time units?

EIPS/eip-5805.md
xinbenlv: Yeah, basically you can choose to adopt it or not. I just wanna point out it leaves ambiguity in some cases and could cause challenges for adoption (the contract client side)

frangio:
> do you want to consider designate a "DEFAULT clock type" so compliants still know what to do with time units?

Pretty sure this is already specified as you're suggesting, can you check?

> If the contract does not implement ERC-6372, it MUST operate according to a block number clock, exactly as if ERC-6372's `CLOCK_MODE` returned `mode=blocknumber&from=default`.

Amxx (approved):

EIPS/eip-5805.md
```diff
@@ -52,51 +54,13 @@ Both timestamps, block numbers, and other possible modes use the same external i
...
-  type: function
-  stateMutability: view
-  inputs: []
-  outputs:
-    - name: timepoint
-      type: uint48
-```
-
-### CLOCK_MODE
+#### [ERC-6372](./eip-6372.md): clock and CLOCK_MODE
```
SamWilsn, line 57:
```suggestion
#### [ERC-6372](./eip-6372.md): clock and `CLOCK_MODE`
```

SamWilsn (approved):

PR #6487, opened 2023-02-09
Update EIP-1: locally disable MD046 footnote demo

xinbenlv:
A follow up for 

> We should turn off that error. I think it's the only way to get a code block to appear inside the footnote list.

_Originally posted by @SamWilsn in https://github.com/ethereum/EIPs/issues/6477#issuecomment-1424617429_

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -269,6 +269,8 @@ Which renders to:
 
 This is a sentence with a footnote.[^1]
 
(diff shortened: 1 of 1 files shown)


Pandapip1 (approved):
LGTM.

lightclient:
How do we fix the linter here?

SamWilsn:
This looks like it's been fixed in EIP-1? Please reopen if not.

PR #6492, opened 2023-02-10
Add EIP: Signature Validation for Pre-Deploy Contracts

Ivshti:
This PR adds an EIP that extends EIP-1271 with an addition that allows counterfactual contracts to sign messages and get them verified. It was initially inspired by this thread in the EIP-4337 implementation: https://github.com/eth-infinitism/account-abstraction/issues/188

btw sorry for assigning a number, I read that this is the editor responsibility later on but figured I'd leave it in as it should make life easier

### DIFF
--- a/EIPS/eip-6492.md
+++ b/EIPS/eip-6492.md
@@ -1,0 +1,127 @@
+---
+eip: 6492
+title: Standard Signature Validation Method for Counterfactually Deployed Contracts
+description: Standard way to verify a signature when the account is a smart contract that has not been deployed iet
+author: Ivo Georgiev (@Ivshti)
+discussions-to: https://github.com/ethereum/EIPs/issues/6492
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-02-10
+---
+
+## Abstract
+Contracts can sign verifyable messages via [EIP-1271](./eip-1271.md).
+
+However, if the contract is not deployed yet, [EIP-1271](./eip-1271.md) verification is impossible.
+
+We propose a standard way for any contract or off-chain actor to verify whether a signature on behalf of a given not-yet-deployed contract is valid. This standarad way extends [EIP-1271](./eip-1271.md). This is meant to combine with CREATE2-powered counterfactually deployed contracts, as their address can be calculated from the deploy data before they're deployed.
+
+## Motivation
+With the popularity of [EIP-4337](./eip-4337.md) and other account abstraction initiatives, we often find that the best user experience for contract wallets is to defer contract deployment until the first user transaction, therefore not burdening the user with an additional deploy step before they can use their account. However, at the same time, many dApps expect signatures, not only for interactions, but also just for logging in.
+
+As such, contract wallets have been limited in their ability to sign messages before their de-facto deployment, which is often done on the first transaction.
+
+Furthermore, not being able to sign messages from counterfactual contracts has always been a limitation of [EIP-1271](./eip-1271.md).
+
+The initial motivation is discussed in the [eth-infinitism/account-abstraction repo](https://github.com/eth-infinitism/account-abstraction/issues/188).
+
+## Specification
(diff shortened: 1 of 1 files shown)


EIPS/eip-6492.md
```diff
@@ -0,0 +1,143 @@
...
+
+- check if there's contract code at the address. If so perform [ERC-1271](./eip-1271.md) verification as usual by invoking `isValidSignature`
+- check if the signature ends with magic bytes, in which case do an `eth_call` to a multicall contract that will call the factory first with the `factoryCalldata` and deploy the contract; Then, call `contract.isValidSignature` as usual
+- If all this fails, try `ecrecover` verification
+
+## Rationale
+
+We believe that wrapping the signature in a way that allows to pass the deploy data is the only clean way to implement this, as it's completely contract agnostic, but also easy to verify.
+
+The wrapper format ends in `magicBytes`, which ends with a `0x69`, which makes it is impossible for it to collide with a valid `ecrecover` signature if packed in the `r,s,v` format, as `0x69` is not a valid value for `v`.
```
drortirosh, line 72: In general, a prefix is used to mark data, not suffix.
also it is customary to use the EIP number as ID, So I would suggest using "0x64926492" as a magic value.

EIPS/eip-6492.md
```diff
@@ -0,0 +1,143 @@
...
+    address _signer,
+    bytes32 _hash,
+    bytes calldata _signature
+  ) external override returns (bool) {
+    uint32 contractSize;
+    assembly {
+      size := extcodesize(_signer)
+    }
+    if (contractSize > 0) {
+      return IERC1271Wallet(_signer).isValidSignature(_hash, _signature) == 0x1626ba7e;
```
drortirosh, line 103: this logic would fail if the contract gets deployed between the time the signature was created, and the time it was checked.
The magic check should be first, and the deploy code skipped, in case the contract is already deployed.

EIPS/eip-6492.md
```diff
@@ -0,0 +1,143 @@
...
+      (create2Factory, factoryCalldata, sig) = abi.decode(sig, (address, bytes, bytes));
+      (bool success, ) = create2Factory.call(factoryCalldata);
+      if (!success) return false;
+      return IERC1271Wallet(_signer).isValidSignature(_hash, sig) == 0x1626ba7e;
+    }
+    
+    // ecrecover verification
+    require(_signature.length == 65, "SignatureValidator#recoverSigner: invalid signature length");
+    uint8 v = uint8(_signature[64]);
+    bytes32 r = _signature.readBytes32(0);
```
drortirosh, line 119: If you want to reduce your dependency on external libraries, you could use bytes32(signature[0:32])
If you don't care about reducing library dependencies and would like to shorten the sample code then you could use OpenZeppelin.ECDSA library (or OpenZeppelin.ERC1271)

EIPS/eip-6492.md
Ivshti: the suffix is used to make it impossible for the detection to collide with a valid r,s,v signature, with the presumption that the de-facto standard is that v is packed last

As for the magic value, that makes sense

EIPS/eip-6492.md
Ivshti: This is a matter of compromise as this would allow collisions, but it's a good point. If addressed, the magic value will probably need to be bytes32 though to reduce the chance of collisions.

EIPS/eip-6492.md
Ivshti: thanks, IMO the example should have as few dependencies as possible

Ivshti:
hey @drortirosh, would appreciate a follow-up review here :pray:

Pandapip1 (approved):
I think this is fine for a draft. I'll take a closer look when it goes to Review.

Aaronmcrae718:
I WANT ALL ACCESS TO ALL MY WALLETS 
#DATAKEOVA

Aaronmcrae718:
I WANT THE KEY TO UNLOCK ALL THE BLOCKS 
#datakeova

PRADACANDI18 (approved):

PR #6493, opened 2023-02-10
Add EIP-6493: SSZ Transaction Signature Scheme

etan-status:
This PR builds on top of prior work from:
- @lightclient at https://github.com/ethereum/EIPs/pull/6385

The signature malleability issue in the original PR is addressed by reusing the consensus `compute_signing_root` mechanism to link each hash with the transaction's underlying `chain_id` and `tx_type`.

Note that this makes the transaction hashes different from the plain `hash_tree_root` values. This means that if the `transactions_root` MPT is replaced with SSZ (EIP-6404), that the `transaction_hash` would need to be tracked separately, same as for legacy RLP-based transactions. This is mainly a cosmetic issue, not a practical one. In an SSZ tx tree, we could simply include both the HTR as well as the perpetual tx hash.

Cryptographic analysis may be necessary to determine the amount by which the hash collision probability is increased, if we use different hashing algorithms for transactions. On the other hand, using different algo for SSZ transactions reduces the impact of a custom network defining 0x05 as a RLP transaction that might serialize same as the blob SSZ transaction.


etan-status:
Overall, these items need to be mixed into any SHA256 based signature:
1. The fact that this signature is for a transaction
2. Schema version describing how the tx was serialized
3. Spec namespace (to distinguish from Ethereum based L2 / Sidechains)
All these items must be mixed in using static information based on the node configuration, not based on the dynamic network data.

Currently, this PR uses an approach as close as possible to the consensus specs:
1. `DOMAIN_EXECUTION_TRANSACTION` (in consensus: `DomainType`)
2. `tx_type` (in consensus: `Version`)
3. `chain_id` (in consensus: `genesis_validators_root`)

It doesn't necessarily have to be these constants, and it doesn't necessarily have to be the same schema from consensus. While both these tx signatures and the consensus signatures are SSZ based, they are not signed using the same key. For example, we could also just run `uuidgen` for each new TX type, instead of deriving a `Domain` from (1), (2), and (3).

For the perpetual TX hash, it could be relaxed, but the same information needs to be mixed into the hash if the goal is to create globally unique TX hashes (across chains).

etan-status:
There is also https://eips.ethereum.org/EIPS/eip-2124, maybe the "genesis hash" would work to link the signature to the underlying spec document defining the tx type, as the consensus genesis_validators_root equivalent.

EIPS/eip-4844.md
```diff
@@ -177,10 +194,13 @@ def get_origin(tx: SignedBlobTransaction) -> Address:
 The hash of a signed blob transaction should be computed as:
 
 ```python
-def signed_tx_hash(tx: SignedBlobTransaction) -> Bytes32:
-    return keccak256(BLOB_TX_TYPE + ssz.serialize(tx))
+def signed_tx_hash(tx: SignedBlobTransaction) -> Root:
+    domain = compute_transaction_domain(BLOB_TX_TYPE, node.cfg.chain_id)
+    return compute_signing_root(tx, domain)
```
etan-status, line 199: For the `signed_tx_hash`, as it already includes the signature (which already signs over `compute_transaction_domain`), I think it may be possible to just use plain `hash_tree_root`. TBD.

etan-status:
@lightclient 

I moved the scheme for SSZ transaction signatures to a separate EIP (6493).

Note that this representation is for the way how signatures are signed, identified, and represented before inclusion in a block. For post-block-inclusion, there is EIP-6404.

The new signature scheme proposed here is close to what the "SSZ Union" approach in EIP-6404's test section is referring to.

etan-status:
Let's move discussion to here: 
https://ethereum-magicians.org/t/eip-6493-ssz-transaction-signature-scheme/13050

EIPS/eip-6493.md
```diff
@@ -0,0 +1,226 @@
...
+requires: 155, 2124, 2718
+---
+
+## Abstract
+
+This EIP defines a signature scheme for [Simple Serialize (SSZ)](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/ssz/simple-serialize.md) encoded transactions.
+
+## Motivation
+
+Existing [EIP-2718](./eip-2718.md) transaction types are encoded in the RLP format, and then hashed using keccak256 for signing. For new transaction types that are encoded in the SSZ format, it is ideomatic to base their signature hash and their public identifier on `hash_tree_root` instead.
```
g11tech, line 20:
```suggestion
Existing [EIP-2718](./eip-2718.md) transaction types are encoded in the RLP format, and then hashed using keccak256 for signing and to generate a unique transaction reference/public identifier. 
For new transaction types that are encoded in the SSZ format, it is ideomatic to base their signature hash and their public identifier on `hash_tree_root` instead.
```

EIPS/eip-6493.md
```diff
@@ -0,0 +1,226 @@
...
+### SSZ transaction schema
+
+For each SSZ transaction type, its specification defines an SSZ type that contains all transaction data to be signed.
+
+```python
+class XyzTransaction(View):
+    ...
+```
+
+Each specification also defines an SSZ type that represents the transaction's signature.
```
g11tech, line 35:
```suggestion
The specification also defines an SSZ type that represents the transaction's signature.
```

EIPS/eip-6493.md
```diff
@@ -0,0 +1,226 @@
...
+
+```python
+class XyzSignedTransaction(Container):
+    message: XyzTransaction
+    signature: XyzSignature
+```
+
+The unique transaction identifier is defined as this signed container's SSZ `hash_tree_root`.
+
+When representing the transaction as part of the [execution block header's `txs-root`](https://github.com/ethereum/devp2p/blob/bd17dac4228c69b6379644355f373669f74952cd/caps/eth.md#block-encoding-and-validity) Merkle-Patricia Trie, that unique transaction identifier is embedded instead of the transaction's raw network representation.
```
g11tech, line 52:
```suggestion
When representing such a  transaction as part of the [execution block header's `txs-root`](https://github.com/ethereum/devp2p/blob/bd17dac4228c69b6379644355f373669f74952cd/caps/eth.md#block-encoding-and-validity) Merkle-Patricia Trie, this unique transaction identifier is embedded instead of the transaction's raw network representation.
```

EIPS/eip-6493.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Rationale
+
+### Why not keccak256?
+
+SSZ and RLP objects encode differently. Namely, in an encoded SSZ transaction, it is not guaranteed that the `chain_id` and the signature are at the same location as in the RLP transaction. This could be problematic if two different networks accidentally use the same [EIP-2718](./eip-2718.md) transaction type number to define an RLP encoded transaction type on one network, but an SSZ encoded transaction type on the other. A signed transaction on one network could suddenly become a valid transaction on the other network.
+
+### Why the specific domain type value?
+
+[`DomainType`](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/phase0/beacon-chain.md#custom-types) is used in consensus to isolate signing domains for validator BLS signatures. So far, execution uses secp256k1 ECDSA signatures instead, so it is not strictly necessary to isolate consensus and execution domains from each other. However, with 4 bytes, avoiding collisions across layers is trivially possible and might be useful in future use cases.
```
g11tech, line 143:
```suggestion
[`DomainType`](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/phase0/beacon-chain.md#custom-types) is used in consensus to isolate signing domains for validating BLS signatures. So far, execution uses secp256k1 ECDSA signatures instead, so it is not strictly necessary to isolate consensus and execution domains from each other. However, with 4 bytes, avoiding collisions across layers is trivially possible and might be useful in future use cases.
```

EIPS/eip-6493.md
```diff
@@ -0,0 +1,226 @@
...
+# Example transaction
+message = ExampleTransaction(
+    chain_id=CHAIN_ID,
+    nonce=42,
+    max_fee_per_gas=69123456789,
+    gas=21000,
+    tx_to=ExecutionAddress(bytes.fromhex('d8da6bf26964af9d7eed9e03e53415d37aa96045')),
+    tx_value=3_141_592_653,
+)
+sig_hash = compute_example_sig_hash(message)
```
g11tech, line 200:
```suggestion
sig_root = compute_example_sig_root(message)
```
would this be clearer to convey that this goes into signature calc

EIPS/eip-6493.md
Pandapip1: Typo fix:
```
When representing such a transaction as part of the [execution block header's `txs-root`](https://github.com/ethereum/devp2p/blob/bd17dac4228c69b6379644355f373669f74952cd/caps/eth.md#block-encoding-and-validity) Merkle-Patricia Trie, this unique transaction identifier is embedded instead of the transaction's raw network representation.
```

Pandapip1 (approved):
LGTM now (for a draft).

PR #6495, opened 2023-02-11
Update EIP-4844: Use SSZ `Optional` for `Address`

etan-status:


etan-status:
Note that this sheds the union's `selector` byte from the previous serialization, so should be adopted in a coordinated way at devnet X. Optionals infer presence of field based on data length != 0 and do not serialize the selector.

dankrad:
This introduces a new type into SSZ to shed a single byte from the transaction? Sounds very excessive to me.

etan-status:
> This introduces a new type into SSZ to shed a single byte from the transaction? Sounds very excessive to me.

Discussion moved to Magicians forum: https://ethereum-magicians.org/t/eip-6475-ssz-optional/12891/5

lightclient:
closing now as we are no longer pursuing SSZ in 4844

PR #6513, opened 2023-02-16
Website: Add automatic citation data

Pandapip1:
Depends on https://github.com/ethereum/EIPs/pull/6514

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -35,7 +35,7 @@
       <meta name="citation_publication_date" content="{% for part in pubdate %}{{ part }}{% if forloop.last == false %}/{% endif %}/{% endfor %}">
     {% endif %}
     {% assign createdate = page.created | split: '-' %}
-    <meta name="citation_online_date" content="{% for part in createdate %}{{ part }}{% if forloop.last == false %}/{% endif %}/{% endfor %}">
+    <meta name="citation_online_date" content="{{ createdate | join: '/' }}">
     <meta name="citation_technical_report_institution" content="Ethereum Improvement Proposals">
     <meta name="citation_technical_report_number" content="{{ page.eip }}">
   {% else %}
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -32,7 +32,7 @@
     {% endfor %}
     {% if page.status == "Final" %}
       {% assign pubdate = page.finalized | split: '-' %}
-      <meta name="citation_publication_date" content="{% for part in pubdate %}{{ part }}{% if forloop.last == false %}/{% endif %}/{% endfor %}">
+      <meta name="citation_publication_date" content="{{ pubdate | join: '/' }}">
     {% endif %}
     {% assign createdate = page.created | split: '-' %}
     <meta name="citation_online_date" content="{{ createdate | join: '/' }}">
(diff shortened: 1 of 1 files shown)


_includes/head.html
```diff
@@ -13,6 +13,31 @@
...
+      {%- else -%}
+        <meta name="citation_author" content="{{ author }}">
+      {%- endif -%}
+    {% endfor %}
+    {% if page.status == "Final" %}
+      {% assign pubdate = page.finalized | split: '-' %}
+      <meta name="citation_publication_date" content="{% for part in pubdate %}{{ part }}{% if forloop.last == false %}/{% endif %}/{% endfor %}">
+    {% endif %}
+    {% assign createdate = page.created | split: '-' %}
+    <meta name="citation_online_date" content="{% for part in createdate %}{{ part }}{% if forloop.last == false %}/{% endif %}/{% endfor %}">
```
SamWilsn, line 38: Would something like this work?

```suggestion
    <meta name="citation_online_date" content="{{ createdate | join: '/' }}">
```

_includes/head.html
```diff
@@ -13,6 +13,31 @@
...
+      {%- elsif author contains "(@" -%}
+        {%- assign authorparts=author|split:"(@" -%}
+        <meta name="citation_author" content="{{ authorparts[0]|strip }}">
+      {%- else -%}
+        <meta name="citation_author" content="{{ author }}">
+      {%- endif -%}
+    {% endfor %}
+    {% if page.status == "Final" %}
+      {% assign pubdate = page.finalized | split: '-' %}
+      <meta name="citation_publication_date" content="{% for part in pubdate %}{{ part }}{% if forloop.last == false %}/{% endif %}/{% endfor %}">
```
SamWilsn, line 35: Maybe:

```suggestion
      <meta name="citation_publication_date" content="{{ pubdate | join: '/' }}">
```

_includes/head.html
```diff
@@ -13,6 +13,31 @@
     <meta name="description" content="{{page.description | escape }}" />
     <meta property="og:description" content="{{page.description | escape }}" />
     <meta name="twitter:description" content="{{page.description | escape }}" />
+    {% if page.category == "ERC" %}
```
SamWilsn, line 16: I kinda feel like we're at the point where we should do something like:

```liquid
{% if page.category == "ERC" %}
    {% assign prefix = "ERC" %}
{% else %}
    {% assign prefix = "EIP" %}
{% endif %}
```

_includes/head.html
```diff
@@ -13,6 +13,31 @@
...
+      <meta name="citation_title" content="ERC-{{ page.eip }}: {{ page.title | escape }}">
+    {% else %}
+      <meta name="citation_title" content="EIP-{{ page.eip }}: {{ page.title | escape }}">
+    {% endif %}
+    {%- assign authors=include.authors|split:"," -%}
+    {% for author in authors %}
+      {%- if author contains "<" -%}
+        {%- assign authorparts=author|split:"<" -%}
+        <meta name="citation_author" content="{{ authorparts[0]|strip }}">
```
SamWilsn, line 25: Do all of these need `| escape`?

_includes/head.html
Pandapip1: If you're sure this works, okay.

_includes/head.html
Pandapip1: Yes.

_includes/head.html
Pandapip1: Yea, but this isn't the place to do that.

Pandapip1:
Superseded by rewrite

PR #6522, opened 2023-02-18
Improvement proposal to 5192

sullof:
I suggest to add a function `defaultLocked` to the interface, which becomes
```
// SPDX-License-Identifier: CC0-1.0
pragma solidity ^0.8.0;

interface IERC5192 {
  /// @notice Emitted when the locking status is changed to locked.
  /// @param tokenId The identifier for a token.
  event Locked(uint256 tokenId);

  /// @notice Emitted when the locking status is changed to unlocked.
  /// @param tokenId The identifier for a token.
  event Unlocked(uint256 tokenId);

  /// @notice Returns the locking status of an Soulbound Token
  /// @dev SBTs assigned to zero address are considered invalid, and queries
  /// about them do throw.
  /// @param tokenId The identifier for an SBT.
  function locked(uint256 tokenId) external view returns (bool);
  
  /// @notice Returns the default locking status of an Soulbound Token
  /// @dev it avoid enforcing the emission of a Locked or Unlocked event at minting
  ///      It may be a `pure` function in most cases, but a view is more flexible
  function defaultLocked() external view returns (bool);
}
```

**Why?**

I am currently working on a security-oriented project and developing a similar interface (https://github.com/ndujaLabs/protector-protected-protocol/blob/main/contracts/interfaces/IERC721Approvable.sol). In our case, we need to control the approvability of a token. While investigating, I realized a significant problem with emitting events (such as `Locked` and `Unlocked`) is that you have to emit them each time there's a change. However, when you mint a token, it typically starts with a default value for every tokenId, so emitting events at that point would be a waste of gas. For instance, with a typical soulbound token that is not transferable and never will be, it does not make any sense to emit the event. This would cause that whoever is deploying such a token would not use ERC5192.

In this proposal, the emission of the event at minting is unnecessary because an external actor can check if the token supports ERC5192 and query a function that returns the default status if no events have been emitted.

### DIFF
--- a/EIPS/eip-5192.md
+++ b/EIPS/eip-5192.md
@@ -35,15 +35,15 @@ pragma solidity ^0.8.0;
 
 interface IERC5192 {
   /// @notice Emitted when the locking status is changed.
-  /// @dev    It is not emitted when the token is minted because the initial
-  ///         status is returned by defaultLocked().
-  /// @param  tokenId The identifier for a token.
+  /// @dev It is not emitted when the token is minted because that default
+  /// transferability is returned by defaultLocked().
+  /// @param tokenId The identifier for a token.
   event Locked(uint256 indexed tokenId, bool status);
 
   /// @notice Returns the locking status of an Soulbound Token
-  /// @dev    SBTs assigned to zero address are considered invalid, and queries
-  ///         about them do throw.
-  /// @param  tokenId The identifier for an SBT.
+  /// @dev SBTs assigned to zero address are considered invalid, and queries
+  /// about them do throw.
+  /// @param tokenId The identifier for an SBT.
   function locked(uint256 tokenId) external view returns (bool);
   
   /// @notice Returns the default locking status of an Soulbound Token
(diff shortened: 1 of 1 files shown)


sullof:
I changed the function `defaultLocked` from `view` to `pure` because it makes more sense. There may be cases where, however, more flexibility is needed, so maybe `view` is better.

sullof:
I apologize for the many commits. I had some second thoughts. Initially, I was also suggesting to move from `Locked` and `Unlocked` events to a more flexible `Locked(uint,bool)` event, but since I have already implemented ERC5192 in production in a couple of projects, I realized that that would be a self-inflict damage and I removed the change from this PR. In any case, it made more sense to make two separate PRs.

PS > I assume that it is too late to make radical changes, but releasing an inefficient proposal looks worse to me.

Pandapip1:
Please create a new EIP that extends EIP-5192. Final EIPs cannot be modified.

sullof:
I thought so. The problem is that a new EIP with my proposed change cannot extend EIP5192 because the latter requires the events to be emitted all the time, while my change requires them only after the minting. So, the conflict make impossible for someone to implement both at the same time, i.e., it cannot be an extension but has to be a brand new one, which is what I was trying to avoid. Anyway, I knew that it was too late. Thanks for taking a look.

Pandapip1:
You can still override behaviors of other EIPs in EIPs you make (see EIP-777's comments modifying EIP-20, for example).

PR #6531, opened 2023-02-21
ERC-5114: Some wording changes regarding badge vs. token

SamWilsn:
Mostly just switching "token" for "badge", except in the description (to make it more... descriptive.)

### DIFF
--- a/EIPS/eip-5114.md
+++ b/EIPS/eip-5114.md
@@ -1,7 +1,7 @@
 ---
 eip: 5114
 title: Soulbound Badge
-description: A badge that is attached to a "soul" at mint time and cannot be transferred after that.
+description: A token that is attached to a "soul" at mint time and cannot be transferred after that.
 author: Micah Zoltu (@MicahZoltu)
 discussions-to: https://ethereum-magicians.org/t/eip-5114-soulbound-token/9417
 status: Review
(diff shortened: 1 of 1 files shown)


EIPS/eip-5114.md
```diff
@@ -79,16 +79,16 @@ This is a new token type and is not meant to be backward compatible with any exi
...
 
-Users of tokens that claim to implement this EIP must be diligent in verifying they actually do.
-A token author can create a token that, upon initial probing of the API surface, may appear to follow the rules when in reality it doesn't.
+Users of badges that claim to implement this EIP must be diligent in verifying they actually do.
+A badge author can create a badge that, upon initial probing of the API surface, may appear to follow the rules when in reality it doesn't.
 For example, the contract could allow transfers via some mechanism and simply not utilize them initially.
 
-It should also be made clear that soulbound tokens are not bound to a human, they are bound to a persona.
-A persona is any actor (which could be a group of humans) that collects multiple soulbound tokens over time to build up a collection of badges.
+It should also be made clear that soulbound badges are not bound to a human, they are bound to a persona.
```
g11tech, line 86:
```suggestion
It should also be made clear that soulbound badges are not bound to a human, they are bound to an entity.
```

is it better to use entity than persona as entity is a better understood word

MicahZoltu: You may get better engagement on this question over on the discussions-to URL as PRs are transient and this is out of scope of this PR.

TL;DR of my position is that persona is a more correct word here than entity, as one entity may have multiple personas, and each persona may own a set of badges.

MicahZoltu (approved):

PR #6538, opened 2023-02-22
Add Stealth Meta-Address Registry 

nerolation:
The stealth meta-address registry will provide a central place for users to store their stealth-meta address as defined in [EIP-5564](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5564.md).

Its a light-weight contract that can be called to (programmatically) retrieve the stealth meta-address of a users to then engage in more private interactions through a stealth address.

### DIFF
--- a/EIPS/eip-6467.md
+++ b/EIPS/eip-6467.md
@@ -1,0 +1,117 @@
+---
+eip: 6467
+title: Stealth Meta-Address Registry
+description: A registry to map addresses to stealth meta-addresses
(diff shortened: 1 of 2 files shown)


abcoathup:
Was ERC6467 assigned?  EIP editors assign EIP #'s and are generally the PR #.

nerolation:
No, was not assigned. Upated it.

Pandapip1 (approved):
I think this is fine for a draft.

PR #6546, opened 2023-02-22
Update EIP-4200: Allow `RJUMPV` table sizes of up to 256

Philogy:
Since `RJUMPV` table sizes of `0` are disallowed I'm suggesting a change that would make table sizes start at 1, meaning a `count` byte of `0` would indicate a table of size 1 and a `count` byte `0xff` would indicate a table size of 256. This would remove an unnecessary piece of validation logic and allow tables that can map the full range of a 4-bit `case` value without `0xff` resulting in the default path.


gumb0:
Originally we argued against this idea to avoid "number `count` really means `count + 1`" which seems quirky.

But we could also reframe it to mean max index of the table.

chfast:
I think this is a good change. So far I was mostly focused on the encoding but realized having 256 element jump table should be very useful for implementing any bytecode interpreter.

Philogy:
> Originally we argued against this idea to avoid the spec stating "number `count` really means `count + 1`", which seemed quirky.
> 
> But we could also reframe it to mean max index of the table.

I agree that this definition can be confusing although I don't think that's a good enough reason not to include this change. I also think putting it in terms of "the byte represents the max index"  is a good idea.

axic:
I agree with the change.

EIPS/eip-4200.md
```diff
@@ -51,7 +51,7 @@ If the code is valid EOF1:
...
+The immediate encoding of `RJUMPV` is more special: the 8-bit `count` value determines the number of `relative_offset` values, starting at 1. The number of offsets starts at 1 because a table of size `0` wouldn't be useful and it allows table sizes up to 256, meaning it could support the full range of a 4-bit `case` value. The encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case >= count` condition falling through means that in many use cases, one would place the *default* path following the `RJUMPV` instruction. An interesting feature is that `RJUMPV 1 relative_offset` is an inverted-`RJUMPI`, which can be used in many cases instead of `ISZERO RJUMPI relative_offset`.
```
gumb0, line 54: I suggest the rewording with reframing of `count` as `max_index`.
```suggestion
The immediate encoding of `RJUMPV` is more special: the unsigned 8-bit `max_index` value determines the maximum index in the jump table. The number of `relative_offset` values following is `max_index+1`. This allows table sizes up to 256. The encoding of `RJUMPV` must have at least one `relative_offset` and thus it will take at minimum 4 bytes. Furthermore, the `case > max_index` condition falling through means that in many use cases, one would place the *default* path following the `RJUMPV` instruction. An interesting feature is that `RJUMPV 0 relative_offset` is an inverted-`RJUMPI`, which can be used in many cases instead of `ISZERO RJUMPI relative_offset`.
```
I removed `it could support the full range of a 4-bit `case` value`, because it's rather a rationale, if you want you could add it as a separate section in Rationale.

gumb0: Also please change `count` to `max_index` in point 3 on line 50 above.

EIPS/eip-4200.md
```diff
@@ -146,13 +145,13 @@ This change poses no risk to backwards compatibility, as it is introduced at the
...
         - `condition` does not equal `0` 
-- `RJUMPV 1 relative_offset`
+- `RJUMPV 0 relative_offset`
     - `case` equals `0`
     - `case` does not equal `0` 
 - `RJUMPV` with table containing positive, negative, `0` offsets
     - `case` equals `0`
     - `case` does not equal `0` 
-    - `case` outside of table bounds (`case >= count`, fallback case)
+    - `case` outside of table bounds (`case > count`, fallback case)
```
gumb0, line 154:
```suggestion
    - `case` outside of table bounds (`case > max_index`, fallback case)
```

lightclient:
This makes sense to me.

lightclient:
bumping @gumb0 @axic @chfast

chfast (approved):

gumb0 (approved):

PR #6560, opened 2023-02-23
Update EIP-5219: Move to Last Call

Pandapip1:


Pandapip1:
@SamWilsn fixed

Pandapip1:
@eth-bot rerun

Pandapip1:
@eth-bot rerun

Pandapip1:
> How do you encode query parameters in the resource? You don't specify. Would it be something like ["foo", "bar", "?hello=world"] or ["foo", "bar?hello=world"]? Is this something that would be left up to the contract?

I forgot that stuff existed... I'll add a parameter for that.

SamWilsn (approved):

PR #6483, opened 2023-02-09
Website: Rewrite index with filters

fulldecent:
New features:

- Clear presentation
- Focus on the content
- Filters

out

https://youtu.be/gzZqgACuULQ

PR #6579, opened 2023-02-24
Feature: more rss feeds

fulldecent:
This extracts a feature from: https://github.com/ethereum/EIPs/pull/6483

Now added is multiple RSS feeds. And features RSS feeds on the homepage for:

- All changes
- Last Call
- Everything except ERC

This should be of particular interest to people that think eips.ethereum.org has too many notifications about ERCs.


Pandapip1 (changes requested):
Why is EIP-1077 modified?

Pandapip1 (changes requested):
Also, there are merge conflicts. Please fix them.

fulldecent:
@Pandapip1 sorry about that, dumb git mistake, should be fixed now

Pandapip1 (dismissed):

index.html
```diff
@@ -7,7 +7,10 @@ <h1 class="page-heading">EIPs
   <a href="https://discord.io/EthCatHerders"><img src="https://dcbadge.vercel.app/api/server/Nz6rtfJ8Cu?style=flat" alt="Discord channel for ECH eip-editer"></a>
   <a href="https://discord.gg/EVTQ9crVgQ"><img src="https://dcbadge.vercel.app/api/server/EVTQ9crVgQ?style=flat" alt="Discord channel for Eth R&D eip-editing"></a>
   <a href="https://discord.gg/mRzPXmmYEA"><img src="https://dcbadge.vercel.app/api/server/mRzPXmmYEA?style=flat" alt="Discord server for discussions about proposals that impact Ethereum wallets"></a>
+  <a href="rss/all.xml"><img src="https://img.shields.io/badge/rss-Everything-red.svg" alt="RSS"></a>
   <a href="rss/last-call.xml"><img src="https://img.shields.io/badge/rss-Last Calls-red.svg" alt="RSS"></a>
+  <a href="rss/nonerc.xml"><img src="https://img.shields.io/badge/rss-All except ERC-red.svg" alt="RSS"></a>
+  <a href="https://eepurl.com/ikqNIP"><img src="https://img.shields.io/badge/-email%20alerts-red.svg" alt="RSS"></a>
```
SamWilsn, line 13: I'm hesitant to add a link to a third-party service; especially one that handles emails. Is this just a free service mailchimp provides for RSS feeds?

```suggestion
```

rss/erc-last-call.xml
```diff
@@ -0,0 +1,35 @@
+---
+layout: null
+---
+<?xml version="1.0" encoding="UTF-8"?>
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
+  <channel>
+    <title>Ethereum EIPs - Last Call Review</title>
+    <description>All EIPs which are in the two-week "last call" status, please help review these and provide your feedback!</description>
```
SamWilsn, line 8:
```suggestion
    <description>All EIPs which are in the "last call" status, please help review these and provide your feedback!</description>
```

Technically not always two weeks.

rss/erc.xml
```diff
@@ -0,0 +1,33 @@
...
+    <title>Ethereum EIPs - Last Call Review</title>
+    <description>All EIPs which are in the two-week "last call" status, please help review these and provide your feedback!</description>
+    <link>{{ site.url }}</link>
+    <atom:link href="{{ site.url }}/rss/last-call.xml" rel="self" type="application/rss+xml" />
+    <lastBuildDate>{{ site.time | date_to_rfc822 }}</lastBuildDate>
+    {% assign eips = site.pages | sort: 'eip' %}
+    {% for eip in eips %}
+      {% if eip.category == "ERC" %}
+      {% capture description %}
+        <p><strong>EIP #{{ eip.eip }} - {{eip.title }}</strong> is in Last Call status. It is authored by {{ eip.author }} and was originally created {{ eip.created }}. It is in the {{ eip.category }} category of type {{ eip.type }}. Please review and note any changes that should block acceptance.</p>
```
SamWilsn, line 16: This feed still seems to mention Last Call, but doesn't filter for Last Call proposals. Did you mean to update this file?

rss/erc.xml
```diff
@@ -0,0 +1,33 @@
...
+        {% if eip.discussions-to %}
+          <p>The author has requested that discussions happen at the following URL: {{ eip.discussions-to }}</p>
```
SamWilsn, line 18: Can the discussion-to link be in an anchor tag?

index.html
fulldecent: Yes, it is a free service for RSS feeds. If we get past 1,000 or 5,000 people then then I'll need to make a grant or something to fund it or I'll go build something we can host ourself. Or I'll just pay it myself.

fulldecent: Personally I use RSS all day. But probably the rest of the world isn't like me.

SamWilsn: @poojaranjan is going to see what service we already use for the Ethereum blog, and we'll go from there.

timbeiko: We've just (read: literally for the first time today) started using SendPulse. Not sure I can report much so far beyond "it worked!". It's not integrated with RSS, and not sure it can be (if so, we might copy what gets done here for the blog!). 

Worth noting that MailChimp has historically been hostile to crypto projects and has shut down accounts without notice.

rss/erc-last-call.xml
fulldecent: Addressed at a2534bfb77d80b6e4755f5ec0b8ba04b5d921cfb

index.html
fulldecent: Here is a way to look at this point:

1. Emails are good and will probably 10x increase the number of people that are seeing updates to EIPs (source: 2 billion people use Gmail, zero people use Google Reader).
2. Right now we don't see a great email provider we wholeheartedly recommend.
3. The people visiting our website do or do not share our values and will probably recognize a Mailchimp signup form. They are able to make their own informed decision about proceeding.
4. Merging this PR today helps some people, and going forward we can still switch to a new email host.

rss/erc.xml
fulldecent: Thank you, fixed at a2534bfb77d80b6e4755f5ec0b8ba04b5d921cfb

rss/erc.xml
fulldecent: Thank you, fixed at a2534bfb77d80b6e4755f5ec0b8ba04b5d921cfb

index.html
SamWilsn: Unfortunately cryptocurrencies are listed on mailchimp's [prohibited content](https://mailchimp.com/en-ca/legal/acceptable_use/) list as a "may not allow", whatever that means.

fulldecent: Right, so not disallowed. Also reading that full sentence:

> Cryptocurrencies, virtual currencies, and any digital assets related to an Initial Coin Offering

depending on how you read it may not even apply since there is no offering here.

And reading in the context immediately above that of "things that cause spam complaints" very likely does not apply to EIPs.

We are good to proceed here at a minimum for the short term.

Pandapip1: IANAL, but when there's ambiguity, I believe it has to be read in the most permissible sense, so we'd be fine.

Pandapip1 (approved):

lightclient (approved):

fulldecent:
Thank you for the merge!

Now that these RSS feeds are live I have intergrated them into the mailchimp campaigns. Here are some field notes in case we have to do this again some day:

Create an email address and sign up for the signup form using each combination of {Everything, Last call only} x {Everything, EIPonly, ERConly}. This is a workaround for a bug in Mailchimp. This requires having VPN access or friends with other IP addresses.

Create a "campaign" connecting each RSS feed to the appropriate "segment" as described above. Disable link tracking. Fill in required information. Save templates. Use the same template for each. Do not use the "saved segments" feature it is broken if you use adblock, use adhoc segments.

Easy, done, hope it works.

PR #6602, opened 2023-03-03
Update EIP-6150: Add Co-author

fewwwww:


SamWilsn:
Oh, I didn't realize this was a Final proposal. Unfortunately Final proposals cannot be edited outside of extraordinary circumstances.

PR #6603, opened 2023-03-03
Update EIP-1: Follow the style guide rules

Amxx:
As described in the style guide:

> When referring to an EIP with a category of ERC, it must be written in the hyphenated form ERC-X where X is that EIP’s assigned number.


Pandapip1 (approved):
Thanks for catching this!

SamWilsn (approved):

PR #6607, opened 2023-03-03
Add EIP: A parallel system that runs next to the EVM. 

JKincorperated:
This EIP proposes a gasless system for that runs next to the EVM. This system has no native currency so all currencies should be made using smart contracts. The proposed system aims to prevent gas fees using a reputation based system.

As an added bonus smart contracts in this system are in python.

This is still a work in progress and no where near close enough for production.


Pandapip1 (changes requested):
Python itself has no formal specification. Additionally, this appears to be trying to do two things: add python contracts and add reputation-based gas payments. These two components should be split into two separate EIPs.

PR #6610, opened 2023-03-04
Update EIP-4844: Decouple blobs

g11tech:
Decouple the blobs as per 
- https://github.com/ethereum/consensus-specs/pull/3244

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -54,6 +54,7 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `MAX_ACCESS_LIST_SIZE` | `2**24` |
 | `MAX_ACCESS_LIST_STORAGE_KEYS` | `2**24` |
 | `MAX_TX_WRAP_KZG_COMMITMENTS` | `2**12` |
+| `MAX_TX_WRAP_KZG_PROOFS` | `2**12` |
 | `LIMIT_BLOBS_PER_TX` | `2**12` |
 | `DATA_GAS_PER_BLOB` | `2**17` |
 | `HASH_OPCODE_BYTE` | `Bytes1(0x49)` |
@@ -71,12 +72,12 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 
 ### Cryptographic Helpers
 
-Throughout this proposal we use cryptographic methods and classes defined in the corresponding [consensus 4844 specs](https://github.com/ethereum/consensus-specs/blob/23d3aeebba3b5da0df4bd25108461b442199f406/specs/eip4844).
+Throughout this proposal we use cryptographic methods and classes defined in the corresponding [consensus 4844 specs](https://github.com/ethereum/consensus-specs/tree/86fb82b221474cc89387fa6436806507b3849d88/specs/deneb).
 
-Specifically, we use the following methods from [`polynomial-commitments.md`](https://github.com/ethereum/consensus-specs/blob/23d3aeebba3b5da0df4bd25108461b442199f406/specs/eip4844/polynomial-commitments.md):
+Specifically, we use the following methods from [`polynomial-commitments.md`](https://github.com/ethereum/consensus-specs/blob/86fb82b221474cc89387fa6436806507b3849d88/specs/deneb/polynomial-commitments.md):
 
-- [`verify_kzg_proof()`](https://github.com/ethereum/consensus-specs/blob/23d3aeebba3b5da0df4bd25108461b442199f406/specs/eip4844/polynomial-commitments.md#verify_kzg_proof)
-- [`verify_aggregate_kzg_proof()`](https://github.com/ethereum/consensus-specs/blob/23d3aeebba3b5da0df4bd25108461b442199f406/specs/eip4844/polynomial-commitments.md#verify_aggregate_kzg_proof)
+- [`verify_kzg_proof()`](https://github.com/ethereum/consensus-specs/blob/86fb82b221474cc89387fa6436806507b3849d88/specs/deneb/polynomial-commitments.md#verify_kzg_proof)
+- [`verify_kzg_proof_batch()`](https://github.com/ethereum/consensus-specs/blob/86fb82b221474cc89387fa6436806507b3849d88/specs/deneb/polynomial-commitments.md#verify_kzg_proof_batch)
 
 ### Helpers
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-4844.md
```diff
@@ -54,6 +54,7 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `MAX_ACCESS_LIST_SIZE` | `2**24` |
 | `MAX_ACCESS_LIST_STORAGE_KEYS` | `2**24` |
 | `MAX_TX_WRAP_KZG_COMMITMENTS` | `2**12` |
+| `MAX_TX_WRAP_KZG_PROOFS` | `2**12` |
```
dankrad, line 57: This extra constant is not necessary, just use `MAX_TX_WRAP_KZG_COMMITMENTS`.

g11tech: updated :+1:

EIPS/eip-4844.md
```diff
@@ -344,11 +344,12 @@ def validate_blob_transaction_wrapper(wrapper: BlobTransactionNetworkWrapper):
     versioned_hashes = wrapper.tx.message.blob_versioned_hashes
     commitments = wrapper.blob_kzgs
     blobs = wrapper.blobs
+    proofs = wrapper.blob_proofs
     # note: assert blobs are not malformatted
     assert len(versioned_hashes) == len(commitments) == len(blobs)
```
dankrad, line 349:
```suggestion
    assert len(versioned_hashes) == len(commitments) == len(blobs) == len(proofs)
```

dankrad: (`verify_blob_kzg_proof_batch` will also perform this check so it's not essential here, but nice to make it very explicit. If we say that we do the verification in that function, then `assert len(versioned_hashes) == len(commitments)` would be enough because it covers both `blobs` and `proofs`)

EIPS/eip-4844.md
```diff
@@ -334,7 +334,7 @@ class BlobTransactionNetworkWrapper(Container):
     # BLSFieldElement = uint256
     blobs: List[Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB], LIMIT_BLOBS_PER_TX]
     # KZGProof = Bytes48
-    kzg_aggregated_proof: KZGProof
+    blob_proofs: List[KZGProof, MAX_TX_WRAP_KZG_COMMITMENTS]
```
Inphi, line 337: A bit redundant to call it `blob_proofs`. Let's call it `proofs`.

g11tech: added it on the lines of `blob_kzgs`, should we also rename it to just `kzgs`?

asn-d6: IMO `kzgs` is ambiguous because it's not clear if it refers to proofs or commitments.

asn-d6: I think it is, but probably in a separate PR.

I was originally mistaken that you were gonna introduce another `kzgs` naming with this PR, but since it refers to something that already exists, please disregard my previous comment.

I don't have a strong opinion about `blob_kzgs` vs `kzgs`. But I guess if you go with `proofs` it also makes sense to go with `kzgs`.

dankrad: Prefer `commitments` to `kzgs`.

g11tech: updated, can rename kzgs to commitments in a followup PR

g11tech: Here is the followup PR build on top of this current branch: https://github.com/g11tech/EIPs/pull/1  (can update target to this repo once this PR merges)

asn-d6 (dismissed):
LGTM

Inphi (approved):

EIPS/eip-4844.md
g11tech: I have merged the followup changes to this PR on the request from @Inphi

g11tech: should i also rename `blob_versioned_hashes` to `versioned_hashes` in `BlobTransactionType`?

EIPS/eip-4844.md
```diff
@@ -71,18 +71,18 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
...
+- [`verify_kzg_proof()`](https://github.com/ethereum/consensus-specs/blob/86fb82b221474cc89387fa6436806507b3849d88/specs/deneb/polynomial-commitments.md#verify_kzg_proof)
+- [`verify_blob_kzg_proof_batch()`](https://github.com/ethereum/consensus-specs/blob/86fb82b221474cc89387fa6436806507b3849d88/specs/deneb/polynomial-commitments.md#verify_blob_kzg_proof_batch)
 
 ### Helpers
 
 ```python
-def kzg_to_versioned_hash(kzg: KZGCommitment) -> VersionedHash:
-    return BLOB_COMMITMENT_VERSION_KZG + sha256(kzg)[1:]
+def kzg_to_versioned_hash(commitment: KZGCommitment) -> VersionedHash:
+    return BLOB_COMMITMENT_VERSION_KZG + sha256(commitment)[1:]
```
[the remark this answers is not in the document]
g11tech: umm ... `BLOB_COMMITMENT_VERSION_KZG` makes more sense to me as its the version of the commitment scheme

g11tech:
rebase via UI @Inphi could you take a look/re-approve now that you are in authors list :laughing:

EIPS/eip-4844.md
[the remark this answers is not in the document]
hwwhww: I think the rationale was the version of `VersionedHash`. I'm not sure what's the scenario of changing `VERSIONED_HASH_VERSION_*` but not just changing the commitment scheme though.

It should be unified to one of the names anyway.

/cc @asn-d6 @dankrad any thoughts?

---
And `kzg_to_versioned_hash` was also renamed to `kzg_commitment_to_versioned_hash`.

ekc9399 (approved):

ekc9399 (approved):

hwwhww:
It might be off the scope of this PR, but there are some more CL specs changes that are not synced.
e.g., the type of `Blob` was changed to `ByteVector[BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB]`.

EIPS/eip-4844.md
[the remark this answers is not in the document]
asn-d6: No strong opinion here. I think both variants are reasonable.

I would go with @hwwhww suggestions to make it the same with consensus-specs.

asn-d6 (approved):
Let's get this merged finally! Sorry for the delays.

@g11tech is there a chance you can also open another PR to further synchronize the EIP with the consensus-specs in terms of function names and type changes. I think consensus-specs is a bit ahead on this (due to the unittests) and we should get the EIP to catch up. Thanks!

g11tech:
sounds good, will open an update for bringing naming convention uptodate

PR #6647, opened 2023-03-08
Update EIP-6492: apply feedback

Ivshti:
Applying feedback gathered on ETH magicians to EIP-6492


JackSparO3 (commented):
https://github.com/ethereum/EIPs.git

PR #6668, opened 2023-03-09
Update EIP-6492: move to Review

Ivshti:
Now with the reference implementation, and easy to use off-chain and on-chain validation methods, this is ready to move to Review.


SamWilsn (commented):
I can't comment directly on some of the changes, so instead here's a list:

---

Remove this line:

> The initial motivation is discussed in the eth-infinitism/account-abstraction repo.

You can summarize the discussion here if need be, but having a pointer like this isn't extremely useful. You could also leave a comment on the EIP's `discussions-to` thread linking directly to the conversation.

---

This bit is confusing me:

> The signer will normally be a contract wallet, but it could be any other counterfactual contract as well.

Perhaps I'm confusing signer and the contract implementing `isValidSignature`? In the following bulleted list, you describe the steps the off-chain software (what I would call the signer) should take to create the signature.

Is the signer a contract, or some off-chain software?

---

Line 220: `s/it's/its/`

Ivshti:
@SamWilsn 

> Perhaps I'm confusing signer and the contract implementing `isValidSignature`? In the following bulleted list, you describe the steps the off-chain software (what I would call the signer) should take to create the signature.
I changed this to "The signing contract will normally be a contract wallet, but it could be any contract that implements [ERC-1271](./eip-1271.md) and is deployed counterfactually."

> Line 220: `s/it's/its/`
Fixed

SamWilsn (approved):

Issue #6680, opened 2023-03-14
Spam issue


Anastasia1510:
There has been no activity on this issue for 1 week. It will be closed after 3 months of inactivity.

_Originally posted by @github-actions[bot] in https://github.com/ethereum/EIPs/issues/5641#issuecomment-1467156241_

PR #6682, opened 2023-03-14
Add EIP: NFT Flashloans

outdoteth:
This standard is an extension of the existing flashloan standard ([EIP-3156](./eip-3156.md)) to support ERC-721 NFT flashloans. It proposes a way for flashloan providers to lend NFTs to contracts, with the condition that the loan is repaid in the same transaction along with some fee.

### DIFF
--- a/EIPS/eip-6680.md
+++ b/EIPS/eip-6680.md
@@ -120,6 +120,9 @@ contract ExampleFlashLender is IERC6680, IERC3156FlashLender {
         bool success =
             receiver.onFlashLoan(msg.sender, token, tokenId, fee, data) == keccak256("ERC3156FlashBorrower.onFlashLoan");
 
+        // check that flashloan was successful
+        require(success, "IERC6680: Flash loan failed");
+        
         // check that the NFT was returned by the borrower
         require(IERC721(token).ownerOf(tokenId) == address(this), "IERC6680: NFT not returned by borrower");
 
--- a/EIPS/eip-6680.md
+++ b/EIPS/eip-6680.md
@@ -8,7 +8,7 @@ status: Draft
 type: Standards Track
 category: ERC
 created: 2023-03-12
-requires: 3156, 721
+requires: 721, 3156
 ---
 
 ## Abstract
--- a/EIPS/eip-6682.md
+++ b/EIPS/eip-6682.md
@@ -1,9 +1,9 @@
 ---
-eip: 6680
+eip: 6682
 title: NFT Flashloans
 description: Minimal interface for ERC-721 NFT flashloans
 author: out.eth (@outdoteth)
-discussions-to: https://ethereum-magicians.org/t/eip-6680-nft-flashloans/13294
+discussions-to: https://ethereum-magicians.org/t/eip-6682-nft-flashloans/13294
 status: Draft
 type: Standards Track
 category: ERC
@@ -30,7 +30,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 ```solidity
 pragma solidity ^0.8.19;
 
-interface IERC6680 {
+interface IERC6682 {
     /// @dev The address of the token used to pay flash loan fees.
     function flashFeeToken() external view returns (address);
 
@@ -75,9 +75,9 @@ import "../interfaces/IERC20.sol";
 import "../interfaces/IERC721.sol";
 import "../interfaces/IERC3156FlashBorrower.sol";
 import "../interfaces/IERC3156FlashLender.sol";
-import "../interfaces/IERC6680.sol";
+import "../interfaces/IERC6682.sol";
 
-contract ExampleFlashLender is IERC6680, IERC3156FlashLender {
+contract ExampleFlashLender is IERC6682, IERC3156FlashLender {
     uint256 internal _feePerNFT;
     address internal _flashFeeToken;
 
@@ -108,7 +108,7 @@ contract ExampleFlashLender is IERC6680, IERC3156FlashLender {
         returns (bool)
     {
         // check that the NFT is available for a flash loan
-        require(availableForFlashLoan(token, tokenId), "IERC6680: NFT not available for flash loan");
+        require(availableForFlashLoan(token, tokenId), "IERC6682: NFT not available for flash loan");
 
         // transfer the NFT to the borrower
         IERC721(token).safeTransferFrom(address(this), address(receiver), tokenId);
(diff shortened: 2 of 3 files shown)


jenroos78 (dismissed):

abcoathup:
EIP editors assign EIP numbers (as per EIP-1), which is generally the PR number.  #6680 is a spam issue, whilst this PR is #6682
I couldn't see where an editor has assigned the number.

EIPS/eip-6680.md
```diff
@@ -0,0 +1,161 @@
+---
+eip: 6680
```
SamWilsn, line 2:
```suggestion
eip: 6682
```

SamWilsn (changes requested):

EIPS/eip-6680.md
outdoteth: Updated

outdoteth:
@SamWilsn are the rest of the changes ok?

RongxinZhang:
@outdoteth This is very interesting! 
What do you foresee as example use cases?
Who do you think will likely integrate with this?

I would think
- Access to events, e.g BAYC events, where the borrower, only needs to prove ownership of the asset at the gate. They can take a flash loan a return the NFT as soon as they are inside the event.

Questions:
- Would people be able to "counter" this by randomly checking the ownership of an NFT to make sure someone is actually the holder?

jenroos78:
Fraud

jenroos78:
Lock my account!!!

outdoteth:
@RongxinZhang I answered a similar question to yours about potential use cases in the eth magicians thread

EIPS/eip-6682.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Abstract
+
+This standard is an extension of the existing flashloan standard ([ERC-3156](./eip-3156.md)) to support [ERC-721](./eip-721.md) NFT flashloans. It proposes a way for flashloan providers to lend NFTs to contracts, with the condition that the loan is repaid in the same transaction along with some fee.
+
+## Motivation
+
+The current flashloan standard, [ERC-3156](./eip-3156.md), only supports [ERC-20](./eip-20.md) tokens. ERC-721 tokens are sufficiently different from ERC-20 tokens that they require an extension of this existing standard to support them. 
+
+In most cases, the handling of fee payments will be desired to be paid in a seperate currency to the loaned NFTs because NFTs themselves cannot always be fractionalized. Consider the following example where the flashloan provider charges a 0.1 ETH fee on each NFT that is flashloaned; The interface must provide methods that allow the borrower to determine the fee rate on each NFT and also the currency that the fee should be paid in.
```
SamWilsn, line 22: This paragraph belongs more in the Rationale section. The Motivation section should justify the existence of the proposal as a whole, while the Rationale section should explain technical choices made within the standard. In this case, supporting a separate currency for fee payment sounds like a technical choice within this standard.

SamWilsn, line 22:
```suggestion
In most cases, the handling of fee payments will be desired to be paid in a separate currency to the loaned NFTs because NFTs themselves cannot always be fractionalized. Consider the following example where the flashloan provider charges a 0.1 ETH fee on each NFT that is flashloaned; The interface must provide methods that allow the borrower to determine the fee rate on each NFT and also the currency that the fee should be paid in.
```

EIPS/eip-6682.md
```diff
@@ -0,0 +1,161 @@
...
+requires: 721, 3156
+---
+
+## Abstract
+
+This standard is an extension of the existing flashloan standard ([ERC-3156](./eip-3156.md)) to support [ERC-721](./eip-721.md) NFT flashloans. It proposes a way for flashloan providers to lend NFTs to contracts, with the condition that the loan is repaid in the same transaction along with some fee.
+
+## Motivation
+
+The current flashloan standard, [ERC-3156](./eip-3156.md), only supports [ERC-20](./eip-20.md) tokens. ERC-721 tokens are sufficiently different from ERC-20 tokens that they require an extension of this existing standard to support them.
```
SamWilsn, line 20: I'd like to see at least one use case for a flash loaned NFT in this section. As someone who mostly knows NFTs as either art pieces or ENS names, I'm not sure why flashloaning them would be useful.

EIPS/eip-6682.md
```diff
@@ -0,0 +1,161 @@
...
+    // if a contract also supports flash loans for ERC20 tokens then it can
+    // return some value here instead of 0
+    return 0;
+}
+```
+
+## Reference Implementation
+
+```solidity
+// SPDX-License-Identifier: MIT
```
SamWilsn, line 71: Code included inline in the document must be licensed under `CC0-1.0`. You can remove the comment, put the reference implementation in `../assets/eip-6682/`, or:

```suggestion
// SPDX-License-Identifier: CC0-1.0
```

EIPS/eip-6682.md
```diff
@@ -0,0 +1,161 @@
...
+    function availableForFlashLoan(address token, uint256 tokenId) external view returns (bool);
+}
+```
+
+The `flashFeeToken` function MUST return the address of the token used to pay flash loan fees.
+
+If the token used to pay the flash loan fees is ETH then `flashFeeToken` MUST return `address(0)`.
+
+The `availableForFlashLoan` function MUST return whether or not the `tokenId` of `token` is available for a flashloan. If the `tokenId` is not currently available for a flashloan `availableForFlashLoan` MUST return `false` instead of reverting.
+
```
SamWilsn, line 49: I'd like to see a little note saying implementers are supposed to implement `IERC3156FlashLender`. It isn't obvious from the specification text how you're actually supposed to borrow the token.

SamWilsn (commented):
You'll also need to address the `eipw` errors. You can see the annotations on the "Files changed" tab of this PR, or the full output in the logs on the "Checks" tab. Each annotation should have a link explaining the rule and why we enforce them.

We generally don't make exceptions to these rules.

EIPS/eip-6682.md
```diff
@@ -0,0 +1,161 @@
...
+eip: 6682
+title: NFT Flashloans
+description: Minimal interface for ERC-721 NFT flashloans
+author: out.eth (@outdoteth)
+discussions-to: https://ethereum-magicians.org/t/eip-6682-nft-flashloans/13294
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-03-12
+requires: 721, 3156
```
SamWilsn, line 11:
```suggestion
requires: 20, 721, 3156
```

The fees are paid in ERC-20 tokens, so it's a hard requirement.

EIPS/eip-6682.md
outdoteth: Updated

EIPS/eip-6682.md
outdoteth: added some use cases

EIPS/eip-6682.md
outdoteth: updated

EIPS/eip-6682.md
outdoteth: updated

EIPS/eip-6682.md
outdoteth: updated

SamWilsn (approved):

PR #6685, opened 2023-03-14
Option for ERC20 reflecting common no calldata approach

radeksvarz:
PR to demonstrate the update reflecting the discussion:  https://ethereum-magicians.org/t/erc-5679-mint-and-burn-tokens/10913/17


SamWilsn:
This EIP is in the Final state, and may only be modified to correct errata. I would recommend standardizing this interface in its own EIP if you feel it can stand alone.

PR #6689, opened 2023-03-15
normative language adjustments and additional privacy notes

kdenhartog:
This modifies normative language to align it better with the expected outcomes defined. Additionally, it defines how pairwise SIWE accounts MAY be used (there's a tradeoff here that asset accounts will probably still act as global cross site tracking identifiers).

One other notable aspect is that "domain" was changed to "origin" in order to make sure that the scheme is also included. This is important to prevent attacks where the same site is presented over multiple protocols such as wss:// and https:// which would potentially allow for a replay attack to occur to the same backend service. If these services weren't treated as independent and shared a backend service, this could potentially allow for multiple authenticated sessions to be created even if the nonce was implemented properly.

other just general clean up was also done in this EIP in order to prepare it for final call.

### DIFF
--- a/EIPS/eip-4361.md
+++ b/EIPS/eip-4361.md
@@ -25,16 +25,16 @@ Already, many services support workflows to authenticate Ethereum accounts using
 
 Sign-In with Ethereum works as follows:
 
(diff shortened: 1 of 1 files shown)


kdenhartog:
The spec authors have already incorporated the majority of the feedback. So, I'm not going to work on pushing this one through so closing.

PR #6693, opened 2023-03-15
  missing type update.

drortirosh:

 - The build passes.

### DIFF
--- a/EIPS/eip-4337.md
+++ b/EIPS/eip-4337.md
@@ -95,8 +95,8 @@ struct ReturnInfo {
   uint256 preOpGas;
   uint256 prefund;
   bool sigFailed;
-  uint64 validAfter;
-  uint64 validUntil;
+  uint48 validAfter;
+  uint48 validUntil;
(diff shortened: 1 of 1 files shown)


Mister-EA:
I have a question about this EIP: In a decentralized bundler network, what would be the incentive for bundlers to propagate UserOps to the rest of the bundlers in the UserOp mempool? If a bundler can collect fees for being the first to bundle UserOps and send them to the network then I would presume it would become advantageous for a bundler to listen for incoming UserOps but not propagate them further since they can be the first to bundle the UserOps they receive.

In the normal Ethereum mempool this is not an issue because full nodes gossip amongst themselves and with block producers/validators. The full nodes are not in any competition with each other so they have no incentive to censor transactions. 

But without a concept of a full-node for the proposed UserOp mempool, with bundlers in competition with each other to be the first to bundle UserOps, and with no consensus mechanism to regulate which bundler gets to create the next bundle, I would imagine that there wouldn't be any reason or incentive for bundlers to communicate with each other.

PR #6700, opened 2023-03-17
Update EIP-747: Move to Final

Pandapip1:
Waiting on PR-[XXXX]()


SamWilsn:
Perhaps I'm having a stroke, but this is confusing:

> If possible, it is RECOMMENDED not to instead use the `ERC1046` type, which supports images and custom metadata.

Pandapip1:
> Perhaps I'm having a stroke, but this is confusing:

There appears to be an extraneous `not`. I'll fix that.

SamWilsn (approved):

PR #6703, opened 2023-03-17
Update EIP-5570: Move to Last Call

darcys22:


Pandapip1 (changes requested):
I wouldn't usually block an EIP for content-related reasons, but this has some pretty significant privacy concerns, namely, doxxing the Ethereum address used to submit a transaction, their IRL name, telephone number, email address, and **physical address** *and* revealing the contents of everything the person has bought.

The above should be addressed before this moves to last call.

darcys22:
> I wouldn't usually block an EIP for content-related reasons, but this has some pretty significant privacy concerns, namely, doxxing the Ethereum address used to submit a transaction, their IRL name, telephone number, email address, and **physical address** _and_ revealing the contents of everything the person has bought.
> 
> The above should be addressed before this moves to last call.

Hi @Pandapip1 thanks for raising this. I have made a response in the ethereum cat herders thead because its such an important topic. https://ethereum-magicians.org/t/idea-standard-digital-receipts-using-erc-721/9908/40

Do you think this information should be conveyed directly in the ERC? I can update it to include a note recommending that the metadata be encrypted, but the description of how to encrypt it is out of scope for this ERC

SamWilsn:
Is this ready to go to last call, or are you still sorting out the privacy issues?

darcys22:
Hi @SamWilsn i believe it is ready for last call (but definitely needing a privacy NFT ERC to be made in the future to work with this ERC).

i didnt get a response to my comment, i was thinking about adding a line to the ERC saying the privacy issues would be out of scope and recommending the data be encrypted. But that was just going to be a stab in the dark to try resolve pandapips concerns

PR #6773, opened 2023-03-24
Config: Disable maximum concurrent dependency PRs

Pandapip1:
You better merge #6772 first, or about 30 bogus PRs will be opened.


.github/renovate.json
```diff
@@ -3,5 +3,6 @@
   "extends": [
     "config:base",
     ":disableDependencyDashboard"
-  ]
+  ],
+  "prConcurrentLimit": 0
```
SamWilsn, line 7: Can we set this to something like 100? I'd still like a limit, even if it's a high one.

Pandapip1: Sure

SamWilsn:
Just marking this as related to #6906.

PR #6366, opened 2023-01-22
Add EIP-6366: Permission Token

chiro-hiro:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #6775, opened 2023-03-24
Update EIP-6617: Move to Review

chiro-hiro:


EIPS/eip-6617.md
```diff
@@ -14,12 +14,11 @@ created: 2023-02-27
 
 This EIP offers a standard for building a bit-based permission and role system. Each permission is represented by a single bit. By using an `uint256`, up to $256$ permissions and $2^{256}$ roles can be defined. We are able to specify the importance of each permission based on the order of the bits.
 
-
 ## Specification
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
-*Note* The following specifications use syntax from Solidity `0.8.7` (or above)
+_Note_ The following specifications use syntax from Solidity `0.8.7` (or above)
```
SamWilsn, line 22: So it looks like this standard is just a bitfield? Like there's no interface to read or interact with the `uint256`, so there's no way to check if a particular user has or doesn't have a permission?

Because of that, I think this proposal is too incomplete to move into the Review status.

SamWilsn (changes requested):
The security considerations section is just "need more discussion", which needs to be looked at before the document can be considered complete enough for review.

You can just put "No security considerations" for now though.

EIPS/eip-6617.md
chiro-hiro: @SamWilsn We tried to separate this proposal from #6366 following the suggestion from @Pandapip1. This proposal is describing how could we use the `uint256` to store permissions and roles instead of `string` or `bytes32`.

> So it looks like this standard is just a bitfield? Like there's no interface to read or interact with the `uint256`, so there's no way to check if a particular user has or doesn't have a permission?

We implemented this EIP as a library, with several methods to manipulate a subset of bits which were stored in `uint256`

https://github.com/ethereum/EIPs/blob/301d3223bd0aa38836bb73509adaae949db5c23e/EIPS/eip-6617.md?plain=1#L107-L113

This method is used to identify a certain bits in `_permission` is a superset of bits in `_requiredPermission` by which we can tell the user has the permissions or not.

cc @vdusart

vdusart: Ok thanks for the feedbacks @SamWilsn, is it better like that ?

EIPS/eip-6617.md
```diff
@@ -14,18 +14,69 @@ created: 2023-02-27
...
 - `0` MUST be used for none permission
 
+```solidity
+pragma solidity ^0.8.7;
+
+/**
+    @title EIP-6617 Bit Based Permission
+    @dev See https://eips.ethereum.org/EIPS/eip-6617
+*/
+library EIP6617 {
```
SamWilsn, line 35: Hm, so this poses another problem: an EIP should define a standard that needs to work between multiple different implementations. Generally a Solidity library is _not_ such a standard (because it is a single implementation.)

For this particular proposal, if you want to standardize bitfields for permissions, you might want to do something like:

```solidity
interface IPermissions {
    /// Returns a URL (preferably IPFS or data:...) with JSON metadata describing the permission.
    function permissionMetadata(uint8 perm) external returns (string);

    /// Returns `true` if `actor` has the given permission.
    function hasPermission(address actor, uint8 perm) external returns (bool);
}
```

I'm no Solidity expert though, and haven't compiled the above, so YMMV.

This way the contract is free to implement permissions however they like (likely a bitfield), and there's a reason to standardize: human-readable metadata.

chiro-hiro: @SamWilsn I think, it's a part of its sister proposal EIP-6366 https://github.com/orochi-network/EIPs/blob/master/EIPS/eip-6366.md?plain=1#L144-L166

chiro-hiro: @vdusart Do you think we should move the metadata here?

vdusart: Hi @SamWilsn,
I don't really see why metadata should be in this lib. 😢
Here the library is not used to save permissions (or informations about them) but only to perform actions on them.
I think like @chiro-hiro that the metadatas should be in the eip 6366.😁

SamWilsn: > ```solidity
> function permissionsOf(address user) external returns (uint256);
> ```

This function alone isn't enough for an EIP. Just... do that in your contract if you want to? Let's break it down a bit, and explore.

Who are the consumers of this proposal? Generally there are three families of consumers of an interface: other contracts, dapps, and wallets.

I'm imagining one contract wants to know about the permissions of some third address. I'm assuming this is the only use case for contracts because if a contract wants to know its own permissions, it can generally just call the function and get rejected or not. I can see this is a valid use case. We're talking about a similar thing over in [ERC-6454's discussion thread](https://ethereum-magicians.org/t/minimalistic-transferable-interface/12517).

While I believe this is a valid use case, I don't think this interface solves it adequately. With just a bitfield, a contract designed to work with multiple ERC-6617 contracts can't know what those permission bits do. You can't be generic over a permissions bitfield without application specific knowledge. If you need application specific knowledge, it isn't a good fit for an EIP.

Similarly here, your dapp/wallet needs to be built to know what the permission bits mean. What good is a user interface that has 256 checks with no labels?

For a rather contrived example, let's say I create a contract and use `1` as the admin privilege, and Panda creates a contract using `2^255`. Both are sane choices. How will any generic interface display useful information about both of these contracts? They'll need application specific information, which again, isn't great for an EIP.

If this proposal created a list of well-known permissions and their bits, or a way to discover information about those permission bits, then I'd be much more amenable to allowing this proposal. As it stands, you've got a named 256-bit integer and I don't believe that's enough to stand on its own.

vdusart: Hey @SamWilsn, @Pandapip1 
To be honest, I'm a little confused and having trouble understanding what you want us to do. To provide some background, earlier this year @chiro-hiro and I proposed an EIP to handle permissions in a different way than the current methods ([173](https://eips.ethereum.org/EIPS/eip-173), [5982](https://eips.ethereum.org/EIPS/eip-5982)).

@Pandapip1 suggested that we split our proposal into two separate EIPs, as this would be more appropriate (see post [here](https://github.com/ethereum/EIPs/pull/6366#pullrequestreview-1294760359)). So we decided to create a first EIP that focuses on bitwise operations (this one), while the second EIP ([6366](https://eips.ethereum.org/EIPS/eip-6366)) handles permissions (transfer, hasPermission, etc.). These two EIPs, 6617 and 6366, work in tandem and the function you are looking for is included in the second EIP (in the core interface [here](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6366.md#core-interface)).

To answer your comment, @SamWilsn, our goal with these 2 EIPs was to create a solution to manage permissions and roles in an efficient manner. We also wanted to enable adding metadata to give names and descriptions to permissions. All this is possible with EIP 6366.

However, we decided to separate the complex management of bitwise operations into a separate EIP (the 6617).

SamWilsn: @vdusart You're pretty much getting caught up in a disagreement between @Pandapip1 and myself. ERC-6366 certainly meets all of my requirements to be an EIP!

I sincerely hope I'm not coming off too negatively. I don't mean any of this in a harsh way; I very much appreciate your group's contribution here!

That said, you have two options to satisfy my concerns as an EIP editor: withdraw ERC-6617, or expand ERC-6617.

Withdrawing ERC-6117 is easy, but possibly not the right answer.

Expanding ERC-6617 is harder, but I don't think it'd be prohibitively so. You really just need to add enough "meat" to the specification for it to be useful without ERC-6366. There are plenty of opportunities to do so. The biggest one, to me, is defining some way to figure out what each permission bit means so that they can be displayed in a dapp or wallet. You have `IEIP6366Meta` in ERC-6366; perhaps that belongs here instead? Or maybe something like EIP-712's [`domainSeparator`](https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator)? That way you can tell if the permissions bits on one contract have the same meaning on another contract.

I'm not an expert on this permissions stuff. I'm completely open to being convinced that we absolutely need a specification for a bitfield. I just haven't been convinced of that yet.

vdusart: No worries @SamWilsn , this is the kind of situation that can happen when working in a group. Anyway, I didn't take your comment the wrong way, I was just a bit confused, but thanks for helping me improving it !

Let's analyze the two options you gave me:

### Option 1: Withdraw EIP-6617
In this case, we will have to move its content to EIP-6636. Indeed, the contents of EIP-6617 must find a place somewhere, and this will result in a larger EIP-6636 (which did not please @Pandapip1 too much).

### Option 2: Add content to EIP-6617
Actually, in my view, the EIP-6617 is only used to "hide" the bitwise operations required for the EIP-6636 to function. These operations may not be obvious to everyone, and having a library to handle them without worrying about the details seemed like a good idea to me.
Therefore, adding `IEIP6366Meta` here doesn't seem consistent to me with keeping this library only for handling binary operations, and having "actual" permissions handling directly in EIP-6636.

### Option 3
I have a proposal for an option 3, maybe renaming this EIP could reduce confusion about its usefulness. What do you think?

PS: I apologize for the delay in responding. It must be painful to have to dive back into a topic several weeks later. I will be more reactive!

chiro-hiro: @vdusart I think, I'm prefer option 2 here is why:

- EIP-6617 is about "Bit based permission", the bit itself doesn't have any mean. So `IEIP6366Meta` can help to describe the permission better.
- 
We should consider `IEIP6366Meta` is a must or an optional interface.

EIPS/eip-6617.md
```diff
@@ -14,18 +14,122 @@ created: 2023-02-27
...
```
SamWilsn, line 35: I think my last complaint before merging as draft is that this should be an interface.

chiro-hiro: This part is library that focused on bit manipulation that allow us to compose/discompose an `uint256` permission. In my opinion, it shouldn't be an interface we have no benefit by implement it as an interface.

SamWilsn: A library that implements bit manipulation doesn't need to be an EIP. It can live on GitHub, as an npm package, or in OpenZeppelin. You can have multiple different bit manipulation libraries that make different choices in implementation.

The EIP should only describe externally visible functionality, and in the case of Solidity that usually means just interfaces.

xinbenlv:
The provided Reference Implementation doesn't match the spec, I don't think at this version it satisfied the editorial requirement to merge for Review status.

assets/eip-6617/contracts/EIP6617Meta.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: CC0-1.0
+pragma solidity ^0.8.7;
+
+import "./interfaces/IEIP6617Meta.sol";
+
+/**
+ * @dev Implement the metadata of EIP-6617
+ */
+contract EIP6617Meta is IEIP6617Meta {
```
xinbenlv, line 9: This implementation doesn't implement what is being specified.

chiro-hiro: Hi @xinbenlv, we will review then update the EIP to match the requirements. Could you be more specific so we can update all at once?.

EIPS/eip-6617.md
```diff
@@ -14,18 +14,122 @@ created: 2023-02-27
...
 
-
 ## Specification
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
-*Note* The following specifications use syntax from Solidity `0.8.7` (or above)
+_Note_ The following specifications use syntax from Solidity `0.8.7` (or above)
 
 - Permission and role MUST be defined as an `uint256`
```
xinbenlv, line 23: I appreciate author's effort to propose and author this EIP, folks. Hope this help explain it better about what kind of improvement I personally would like to see in this EIP before it can be merged as a Review

The specification of EIP at this current snapshot is far from editorially ready.

For example, It begins with  "- Permission MUST be defined as a power of two"

This is confusing. Is it a requirement for a complying contract to "have the methods of Permission" ? If so, it shall establish a requirement for certain methods. Fo example

A compliant contract MUST have the following interface...

And then, it could talk about the requirement of each method, such as "Permission MUST be defined as a power of two."

Also, what does "Permission MUST be defined as power of two" means, exactly?
Maybe I think you mean that Permission is bit-wise so each Permission is defined as one bit in a bytes32/uint256.
etc. 

Then the implementation shall refer to these requirements shows how a typical example of permission is implemented and how each methods will interact with it.


@SamWilsn FYI: "I think my last complaint before merging as draft is that this should be an interface.".

xinbenlv:
I shared it here:
https://github.com/ethereum/EIPs/pull/6775/files#r1287470281

EIPS/eip-6617.md
vdusart: Thanks for the comments @xinbenlv,
I tried to apply all the requests everyone made, tell me if it's better like that :smile:

SamWilsn (approved):

PR #6733, opened 2023-03-20
Add EIP: Revert creation in case of collision

RenanSouza2:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #6784, opened 2023-03-27
Add EIP: Prevent overwriting contracts

RenanSouza2:
This is an existing EIP that was not properly docummented.

Points I was not sure when creating this:

- The original EIP was not in the current standard so I inserted my own texts from the EIP-6733 (not accepted)
- Date of creation as of today instead of the original post
- My name as co-author


Pandapip1:
Sorry, already approved https://github.com/ethereum/EIPs/pull/6733

PR #6787, opened 2023-03-27
Add EIP: Order Book DEX with Two Phase Withdrawal

qizheng09:
The Order Book DEX Standard is a proposed set of interface specifications that define a decentralized exchange protocol for trading assets using order books. This standard provides a set of functions that allow users to deposit, withdraw, and trade assets on a decentralized exchange. Additionally, it innovatively proposes a two-phase withdrawal scheme to ensure the asset security of both users and the exchange, addressing users' trust issues with the exchange.


qizheng09:
Blockchain-driven finance requires us to continuously iterate protocols and solidify good financial protocols. The 1:1 reserve ratio and two-phase withdrawal protocol can ensure the security of assets for both users and project parties. It is a good solution for order book-based DEX built on Layer 2 technology.

Pandapip1:
@eth-bot rerun

Pandapip1:
@eth-bot rerun

EIPS/eip-6787.md
```diff
@@ -0,0 +1,93 @@
+---
+eip: 6787
+title: Order Book DEX with Two Phase Withdrawal
+description: A order book-based DEX Interfaces that ensures the asset security of both users and the exchange
```
SamWilsn, line 4:
```suggestion
description: An order book-based DEX interface that ensures the asset security of both users and the exchange
```

EIPS/eip-6787.md
```diff
@@ -0,0 +1,93 @@
...
+`function commitForceWithdraw(uint256 requestID) external;`
+
+The **commitForceWithdraw** function is used to execute a forced withdrawal operation after the conditions are met. The function takes a *requestID* parameter, which specifies the ID of the forced withdrawal request to be executed. The request must have been previously initiated using the prepareForceWithdraw function.
+
+The flow charts for **prepareForceWithdraw** and **commitForceWithdraw** are shown below：
+
+![](../assets/eip-6787/image1.png)
+
+## Rationale
+
```
SamWilsn, line 75: You'll need to include some explanation for any technical choices made within this proposal.

EIPS/eip-6787.md
```diff
@@ -0,0 +1,93 @@
...
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Test Cases
+
+
+## Reference Implementation
+
+
```
SamWilsn, line 86:
```suggestion
```

These sections can be omitted if empty.

EIPS/eip-6787.md
```diff
@@ -0,0 +1,93 @@
...
+`function settle(address userA, address tokenA, uint256 amountA, address userB, address tokenB, uint256 amountB) external;`
+
+The **settle** function is used to settle transactions between two parties, involving the exchange of assets represented by tokenA and tokenB. The function takes the following parameters:
+
+1. *userA*: the address of the first user involved in the transaction
+2. *tokenA*: the address of the token being traded by user A
+3. *amountA*: the amount of token A being traded by user A
+4. *userB*: the address of the second user involved in the transaction
+5. *tokenB*: the address of the token being traded by user B
+6. *amountB*: the amount of token B being traded by user B
```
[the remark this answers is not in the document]
qizheng09: Thank for your comment! The settle interface has different implementations for different order book exchanges, and here is an example. I just provided a more detailed explanation in the proposal.

EIPS/eip-6787.md
```diff
@@ -0,0 +1,102 @@
...
+
+### `withdraw`
+
+`function withdraw(address token, uint256 amount) external;`
+
+The **withdraw** function allows a user to withdraw a specified amount of a particular token from the exchange. The *token* parameter specifies the address of the token contract, and the *amount* parameter specifies the amount of the token to be withdrawn.
+
+### `settle`
+
+The settle interface has different implementations for different order book exchanges, and here is an example. When inheriting the business, you can also perform settle operations for batch users or multiple currencies.
```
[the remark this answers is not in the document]
qizheng09: Yes, `settle` can have different signatures. This is the example of two people and two currency. But this also the recommend way for implementer.

EIPS/eip-6787.md
```diff
@@ -0,0 +1,102 @@
...
+`function settle(address userA, address tokenA, uint256 amountA, address userB, address tokenB, uint256 amountB) external;`
+
+The example **settle** function is used to settle transactions between two parties, involving the exchange of assets represented by tokenA and tokenB. The function takes the following parameters:
+
+1. *userA*: the address of the first user involved in the transaction
+2. *tokenA*: the address of the token being traded by user A
+3. *amountA*: the amount of token A being traded by user A
+4. *userB*: the address of the second user involved in the transaction
+5. *tokenB*: the address of the token being traded by user B
+6. *amountB*: the amount of token B being traded by user B
```
SamWilsn, line 59: Since `settle` is supposed to be called by the exchange, and not by users or other contracts, it shouldn't be part of the standard. This is very similar to how [ERC-20](https://eips.ethereum.org/EIPS/eip-20) and [ERC-721](https://eips.ethereum.org/EIPS/eip-721) don't specify a `mint` function. You can (and should) still explain how settling works, and if you add a reference implementation, `settle` should probably be there.

```suggestion
```

EIPS/eip-6787.md
```diff
@@ -0,0 +1,102 @@
...
+
+By introducing the two-phase forced withdrawal function, we can protect users' financial security while ensuring the security of the exchange's assets. In the first phase, the system will conduct a preliminary review of the user's withdrawal request to confirm the user's account status. In the second phase, after the forced withdrawal inspection period, users can directly submit the forced withdrawal request to complete the forced withdrawal process. In this way, we can not only prevent users from exploiting the forced withdrawal function to evade losses but also ensure the asset security for both the exchange and the users.
+
+In conclusion, by adopting the two-phase commit protocol and the two-phase forced withdrawal function, we can effectively guard against malicious behaviors and ensure data consistency and security in distributed systems while ensuring user experience and financial security.
+
+The flow charts for two-phase withdrawal are shown below：
```
SamWilsn, line 88:
```suggestion
The flow charts for two-phase withdrawal are shown below:
```

EIPS/eip-6787.md
```diff
@@ -0,0 +1,102 @@
...
+
+### `withdraw`
+
+`function withdraw(address token, uint256 amount) external;`
+
+The **withdraw** function allows a user to withdraw a specified amount of a particular token from the exchange. The *token* parameter specifies the address of the token contract, and the *amount* parameter specifies the amount of the token to be withdrawn.
+
+### `settle`
+
+The settle interface has different implementations for different order book exchanges, and here is an example. When implements the business, you can also perform settle operations for batch users or multiple currencies.
```
SamWilsn, line 42: > When implements the business

I think this is a mis-translation? Perhaps:

```suggestion
The settle interface has different implementations for different order book exchanges, and here is an example. When implemented in your application, you can also perform settle operations for batch users or multiple currencies.
```

SamWilsn (commented):
Looking good!

As more of a suggestion than a requirement, I would add a two new headings under Specification: `### Interface` and `### Settling`. Move all the functions (eg. `deposit`, `withdraw`, etc) under `### Interface`, and move the settling requirements under `### Settling`.

The example should eventually be removed for now, and re-added to your reference implementation if you add one.

EIPS/eip-6787.md
qizheng09: I see. Thank you for your guidance. I will make some adjustments.

qizheng09:
> 

Thank for your reply. I remove the `settle` and will add it to reference implementation in future.

SniperUsopp:
This is a great idea and I hope this proposal will become a standard agreement soon. This has significant implications for the security of users' assets.

EIPS/eip-6787.md
```diff
@@ -0,0 +1,91 @@
...
+
+By introducing the two-phase forced withdrawal function, we can protect users' financial security while ensuring the security of the exchange's assets. In the first phase, the system will conduct a preliminary review of the user's withdrawal request to confirm the user's account status. In the second phase, after the forced withdrawal inspection period, users can directly submit the forced withdrawal request to complete the forced withdrawal process. In this way, we can not only prevent users from exploiting the forced withdrawal function to evade losses but also ensure the asset security for both the exchange and the users.
+
+In conclusion, by adopting the two-phase commit protocol and the two-phase forced withdrawal function, we can effectively guard against malicious behaviors and ensure data consistency and security in distributed systems while ensuring user experience and financial security.
```
SamWilsn, line 75: So this looks like it's justifying why we need two-phase withdrawals, which is the point of the whole EIP. This kind of information belongs in the motivation section.

The rationale should describe individual technical choices made within the EIP itself.

qizheng09: Yes, an account may initiate multiple forced withdrawal requests in parallel, so requestid is required.

qizheng09: Yes, an account may initiate multiple forced withdrawal requests in parallel, so requestid is required.

SamWilsn: Perfect. If you move the current rationale into the motivation section and add that note about `requestId` in the rationale, that should be it before merging as draft.

qizheng09: Thank you for your reply, I have fix it.

SamWilsn (approved):

PR #6790, opened 2023-03-27
Update EIP-1: Remove `eip` preamble entry

Pandapip1:
Nothing really depends on the EIP number being in the preamble anymore, so we can safely remove it.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -116,8 +116,6 @@ EIPs should be written in [markdown](https://github.com/adam-p/markdown-here/wik
 
 Each EIP must begin with an [RFC 822](https://www.ietf.org/rfc/rfc822.txt) style header preamble, preceded and followed by three hyphens (`---`). This header is also termed ["front matter" by Jekyll](https://jekyllrb.com/docs/front-matter/). The headers must appear in the following order.
 
-`eip`: *EIP number* (this is determined by the EIP editor)
-
 `title`: *The EIP title is a few words, not a complete sentence*
 
 `description`: *Description is one full (short) sentence*


Pandapip1:
@eth-bot rerun

abcoathup:
> Nothing really depends on the EIP number being in the preamble anymore, so we can safely remove it.

Where is the EIP number specified if not in the preamble?  Is it in the filename?

Pandapip1:
> Is it in the filename?

Yes.

lightclient:
There is no need to remove the eip from the preamble.

Pandapip1:
> There is no need to remove the eip from the preamble.

Actually, it makes eip-review-bot slightly simpler, and also removes redundant information. So let me correct your statement:

> There is no need to keep the eip in the preamble.

SamWilsn:
I'm generally in favour of single-sourcing information as much as possible, so in this case I'd lean towards removing `eip:` from the preamble.

@lightclient are there any scenarios where removing the header is problematic?

lightclient:
It's a long standing piece of the preamble. I think it is a unique piece of the EIP process and I would be sad to see it go. I would rather increase the complexity of the bots to deal with it than remove it. However, it's not something I feel terribly strongly about. What I do feel strongly about is the _constant_ change in the repository. Things like this are marginal improvements that will likely cause more confusion to authors than do good in the short term.

I think it is time to start make changes at more spaced intervals and batching many of these changes together at once.

lightclient:
The vast majority of contributors to this repository only do so occasionally. I am messaged frequently about "why is this like this now" and "when did this requirement get added / go away". If software is expected to batch breaking changes into major release, it should go doubly so for processes IMO.

SamWilsn:
I don't think it significantly increases the technical complexity of the bots to include the EIP number in the preamble. We'd just need to figure out exactly how it should work.

On the `eipw` side:

1. If the filename matches `eip-[0-9]+\.md`:
   * The `eip:` header must exist.
   * The `eip:` header must match the filename.
2. If the filename doesn't match `eip-[0-9]+\.md`:
   * The filename must match the `title` header, with underscores.
   * The `status` must be `Draft`.
   * The `eip` header must not exist.

Sound right?

Pandapip1:
Nope, I think that the `eip` header shouldn't exist, ever.

Pandapip1:
> If software is expected to batch breaking changes into major release, it should go doubly so for processes IMO.

How about versioned processes? So that older EIPs can get "grandfathered" and use older formatting/rules?

xinbenlv:
My position is _against_ of removing EIP number in preamble in the content. 

Reason: the piece of information "number" is important enough to be replicated in the content and hence could be double checked, carried in publication, spread etc.

lightclient:
Closing as there wasn't sufficient support in the EIPIP meeting.

PR #6799, opened 2023-03-29
Website: remove dependency on `eips` preamble entry

Pandapip1:


_includes/head.html
```diff
@@ -3,12 +3,13 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   {% if page.layout == "eip" %}
+    {% assign eip = page.url | split: "eip-" | last | xml_escape %}
```
SamWilsn, line 6: This may not be robust enough. What happens with files named like `eip-an_extension_to_eip-1559.md`? Similarly, what about `assets/eip-1234/eip-20-diagram.png`?

Pandapip1: Such a file cannot exist. Only underscores are allowed.

SamWilsn: Oh, interesting. Where's that enforced? This bot?

Pandapip1: Yup! The bot overrides any EIP value that the user provides.

PR #6804, opened 2023-03-30
update 4972

dongshu2013:
This diff is to simplify the proposal and update EIP-4972 from stagnant to draft status:

1. In the previous version, we are inventing a new account type besides CA/EOA. In this latest version, we assume NOA is CA.
2. We simplified the interface to reflect the account type change. Since it's simple CA, we don't need to extend ERC20/721 interface anymore.
3. The new interface is coupled with a name service so we extends the name service registry interface as well.


xinbenlv (approved):
Good enough as the draft

PR #6807, opened 2023-03-30
Add EIP: Domain-contracts two-way binding

VenkatTeja:
Draft PR. Looking for feedback on the scope of idea. Will provide detailed approach soon.


VenkatTeja:
Hi @SamWilsn, can you review this please?

Skylerspencer:
H I have no idea what this is I don't understand the language I've tired I do know that I have receipts and time stamps and work recording of ever and minute as well as all the transaction done in my wallet as well as the keystones to them also . So thanks I guess what ever this . I'll keep it in a safe also
.

Pandapip1:
@eth-bot rerun

Pandapip1 (changes requested):
Please create a new EIP file, instead of modifying the template.

VenkatTeja:
> H I have no idea what this is I don't understand the language I've tired I do know that I have receipts and time stamps and work recording of ever and minute as well as all the transaction done in my wallet as well as the keystones to them also . So thanks I guess what ever this . I'll keep it in a safe also .

Hey, thanks for responding. This EIP has nothing to do with protecting your wallet or funds. Its a way to allow projects to link their official website and official contracts. The intention is to standardise this and prevent DNS and phishing attacks.

VenkatTeja:
> Please create a new EIP file, instead of modifying the template.

I added template file back. This leads to some additional commits. Let me know if this works. If you don't want that, I think I will have to raise a new PR.

VenkatTeja:
Hi @Pandapip1 Can you please review this again? Thanks

YaroslavPysanskyi:
I need the help of a specialist, I will be in contact with a specialist before my further actions.

Pandapip1 (changes requested):
This doesn't solve the problem discussed in the motivation. In the event that the DNS is compromised, the attacker's server can easily serve a malicious registry.

VenkatTeja:
> This doesn't solve the problem discussed in the motivation. In the event that the DNS is compromised, the attacker's server can easily serve a malicious registry.

Correct. But this will allow wallets like Metamask by reading on-chain info about the official contracts of the domain. And instantly warn users that the contract user is trying to call may not be official. 

Example:
example.xyz has official contract as ABC. This information is available both on on-chain registry contract and as an offchain API from example.xyz. 

In case of a DNS compromise, the attacker cannot change both on-chain and offchain registry. They can change offchain registry shown from example.xyz. to change on-chain registry, a waiting period of minimum 7 days will be applicable to prove they actually hold the domain. 

So, if an attacker just compromises DNS info, wallets like metamask can read the official contracts from on-chain info and warn users.

VenkatTeja:
Hey @Pandapip1,
I read through my [readme ](https://github.com/ethereum/EIPs/blob/ba5287bdbfcc7eba1b26f2a6b0b68b89eda62a88/eip-domain-contracts-binding.md#rationale) again and I feel its informative enough to point the solution that uses two-way binding to verify official contracts using both on-chain and off-chain methods (like a 2FA) to ensure the state of official contracts is correct and in-sync. 

Let me know if its still unclear and how I can improve. Thanks

VenkatTeja:
Hi @YaroslavPysanskyi, were you able to contact a specialist?

VenkatTeja:
Hi @Pandapip1

eip-domain-contracts-binding.md
```diff
@@ -0,0 +1,189 @@
...
+description: A standward way to enable a two-way binding between domain and its official contracts to prevent DNS attacks
+author: Venkat Kunisetty (@VenkatTeja)
+discussions-to: https://ethereum-magicians.org/t/eip-domain-contracts-two-way-binding/13209
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-03-08
+---
+
+<!--
```
lightclient, line 12: Please remove all the comments.

EIPS/eip-6807.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Test Cases
+
+<!-- TODO -->
+
+<!--
+
+## Reference Implementation
+
+-->
```
SamWilsn, line 108: We don't allow external links. If you would like to include a reference implementation, please do so in the `../assets/eip-6807` folder with a CC0-1.0 license.

SamWilsn (approved):

SamWilsn (changes requested):
Actually, after reading @Pandapip1's review, I agree with him.

I think you need to make it more clear how the smart contract side defends against DNS changes. For example, if the wallet visits the site for the first time, and the attacker has already poisoned the `contracts.json` file, how will the wallet know where the registry is?

If you had a single registry that all domains use, that requires a signed transaction/message to change, and maybe some DNSSEC validation... you could maybe put something together there, but it would still be pretty difficult.

SamWilsn:

VenkatTeja:
> Actually, after reading @Pandapip1's review, I agree with him.
> 
> I think you need to make it more clear how the smart contract side defends against DNS changes. For example, if the wallet visits the site for the first time, and the attacker has already poisoned the `contracts.json` file, how will the wallet know where the registry is?
> 
> If you had a single registry that all domains use, that requires a signed transaction/message to change, and maybe some DNSSEC validation... you could maybe put something together there, but it would still be pretty difficult.

Hi @SamWilsn, Wallets would use on-chain registry to validate if it matches with the one on `contracts.json`. Its a dual check standard. Is there a call I can present this? Like a community call? I have written decent amount of explanations in comments here but I don't understand how its still unclear.

SamWilsn:
Replying on the discussion thread.

PR #6813, opened 2023-04-01
CI: Bump rakefile deps

fulldecent:
This bumps dependencies a little bit (just latest GitHub Pages gem) and fixes the build on macOS


fulldecent:
Hi @Pandapip1 I re-ran and the build passes.

The newer version of HTML-proofer has different options, so I matched to those. It is now more explicit about its options and in the future we can un-disable some of the checks to improve the website and SEO.

However, one of the checks was not disableable so I need to touch EIP-2535 to add image alt tags. This means the HTML-proofer check passes in CI just like it did yesterday.

If there's EIP editor approval or no complaints from @mudgen, hoping that is allowed.

Also because I touched 2535, the EIP validator attacked me about lots of modernization it wants for that PR. 

Since this PR is an editorial change, not a content change, and because the scope of the PR is to fix build dependencies, I request that test failure can be ignored and EIP editors can override and merge this PR as-is.

fulldecent:
Thank you @Pandapip1, are other reviews required to push this through?

Pandapip1:
> Thank you @Pandapip1, are other reviews required to push this through?

Just any other editor. CC @SamWilsn

xinbenlv (approved):
LGTM to updating ruby version, and LGTM if that requires new htmlproofer and also require image to be updated to support accessibility.

SamWilsn (approved):

PR #6822, opened 2023-04-03
Website: 💄 hide header/footer on print

qd-qd:
Add a media query that automatically hides the header and the footer when printing an EIP.


Pandapip1 (approved):
LGTM. CC @SamWilsn

PR #6823, opened 2023-04-03
Add EIP: Token Mapping Slot Retrieval Extension

qd-qd:
This proposal suggests an approach to improving the accuracy of off-chain transaction simulations that involve contracts that comply with the `ERC-20`/`ERC-721`/`ERC-1155` standards. The proposal provides a standardized entry point to obtain the reserved storage slot of a mapping responsible for tracking ownership of tokens in a contract. This approach not only helps capture state changes more precisely but also enables external tools and services to do so without requiring expertise in specific implementation details.

Due to the unique storage layout for different contracts, it is challenging to simulate transactions that involve smart contracts, specifically due to the use of mappings. The storage location of a value in a mapping depends on a specific storage slot, which can only be determined through knowledge of the contract’s implementation. This prevents external platforms and tools from capturing/validating changes made to a contract’s state with certainty.

This proposal introduces a function named getTokenLocationRoot as an extension to `ERC-20`/`ERC-721`/`ERC-1155` contracts, allowing off-chain callers to retrieve the reserved storage slot for the mapping type. This approach eliminates the reliance on events, enhances the precision of data access from storage, and improves the accuracy of off-chain simulations.

### DIFF
--- a/EIPS/eip-6823.md
+++ b/EIPS/eip-6823.md
@@ -1,0 +1,115 @@
+---
+eip: 6823
+title: Token Mapping Slot Retrieval Extension
+description: Approach to enhance precision of off-chain transaction simulations by accessing mapping storage slot in ERC-20/721/1155 contracts.
+author: qdqd (@qd-qd) <qdqdqdqdqd@protonmail.com>
+discussions-to: https://ethereum-magicians.org/t/eip-6823-token-mapping-slot-retrieval-extension/13666
+status: Draft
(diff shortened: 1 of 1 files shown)


Pandapip1:
@eth-bot rerun

Pandapip1:
@eth-bot rerun

Pandapip1:
@qd-qd sorry for the issues with the bot. Feel free to create a new PR.

PR #6864, opened 2023-04-11
Add EIP: Upgradable Fungible Token

jeffishjeff:
This standard outlines a smart contract interface for upgrading/downgrading existing ERC-20 smart contracts while maintaining user balances. The interface itself is an extension of the ERC-20 standard so that other contracts can continue to interact with the upgraded token without any change other than updating contract address.


cynic-1 (approved):
LGTM

EIPS/eip-6864.md
```diff
@@ -0,0 +1,211 @@
...
+
+- the smart contracts must support proxy pattern from the get go, i.e. it cannot be used on contracts that were not deployed with proxies
+- upgrades are silent and irreversible, i.e. users do not have the option to opt-out
+
+In contrast, by reducing the scope to specifically ERC-20 tokens, this proposal standardizes an ERC-20 extension that works with any existing or future ERC-20 smart contracts, is much simpler to implement and to maintain, can be reversed or nested, and offers a double confirmation opportunity for any and all users to explicitly opt-in on the upgrade.
+
+[ERC-4931](./eip-4931.md) attepts to address the same problem by introducing a third "bridge" contract to help facilitate the upgrade/downgrade operations. While this design decouples upgrade/downgrade logic from token logic, ERC-4931 would require tokens to be pre-minted at the destination smart contract and owned by the bridge contrtact rather then just-in-time when upgrade is invoked. It also would not be able to support upgrade-while-transfer and see-through functions as described below.
```
SamWilsn, line 33: Friendly warning that mentioning another EIP creates an implied dependency. Your proposal won't be able to advance past the status of ERC-4931(so this proposal will have to stay draft until 4931 advances.)

You could do something like:

```suggestion
Other proposals attempt to address the same problem by introducing a third "bridge" contract to help facilitate the upgrade/downgrade operations. While this design decouples upgrade/downgrade logic from token logic, these proposals would require tokens to be pre-minted at the destination smart contract and owned by the bridge contract rather then just-in-time when upgrade is invoked. It also would not be able to support upgrade-while-transfer and see-through functions as described below.
```

SamWilsn (approved):

PR #6886, opened 2023-04-16
Update EIP-1185: Move to Draft

mcdee:
Move 1185 back to draft, this has been used in real-world implementations so should be moved through the process to final.

### DIFF
--- a/EIPS/eip-1185.md
+++ b/EIPS/eip-1185.md
@@ -11,9 +11,11 @@ discussions-to: https://ethereum-magicians.org/t/eip1185-dns-resolver-profile-fo
 ---
   
 ## Abstract
+
 This EIP defines a resolver profile for ENS that provides features for storage and lookup of DNS records. This allows ENS to be used as a store of authoritative DNS information.
 
 ## Motivation
+
 ENS is a highly desirable store for DNS information.  It provides the distributed authority of DNS without conflating ownership and authoritative serving of information.  With ENS, the owner of a domain has full control over their own DNS records.  Also, ENS has the ability (through smart contracts) for a domain's subdomains to be irrevocably assigned to another entity.
 
 ## Specification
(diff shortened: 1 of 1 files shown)


mcdee:
The linter is throwing some errors now that I need help with:

- complaining that `eip` isn't a header but according to EIP-1 it is
- complaining that the link to /LICENSE is absolute (which it is), but this is as per the text in EIP-1

In addition, EIP-1 states that the copyright header should be "Copyright Waiver" but the linter requires it to be "Copyright", not sure which one you want.

YaroslavPysanskyi:
I need the help of a specialist, I will be in contact with a specialist before my further actions.

EIPS/eip-1185.md
```diff
@@ -31,50 +34,239 @@ The DNS resolver interface consists of two functions to set DNS information and
...
+```
+
+## Security Considerations
+
+Security of this solution would be dependent on security of the records within the ENS domain.  This degenenrates to the security of the key(s) which have authority over that domain.
 
 ## Copyright
-Copyright and related rights waived via [CC0](../LICENSE.md).
+
+Copyright and related rights waived via [CC0](/LICENSE)
```
SamWilsn, line 272: To make links work consistently between eips.ethereum.org and GitHub, you need to use a URL relative to the current file (so not starting with a `/`) and include the `.md` extension.

```suggestion
Copyright and related rights waived via [CC0](../LICENSE.md).
```

SamWilsn (approved):

mcdee:
> To make links work consistently between eips.ethereum.org and GitHub, you need to use a URL relative to the current file (so not starting with a `/`) and include the `.md` extension

You may want to update EIP-1 then, as it has incorrect information in it at https://eips.ethereum.org/EIPS/eip-1#what-belongs-in-a-successful-eip

SamWilsn:
It's an oddity with Jekyll, our markdown renderer, that pops up from time to time. It is correct in [the source file](https://github.com/ethereum/EIPs/blob/02e909894525413f42e2651c0880179590a0633a/EIPS/eip-1.md?plain=1#L111).

PR #6894, opened 2023-04-17
Update EIP-1: Strengthen wording of status update PRs

Pandapip1:
Why stick to final EIPs? It makes sense to require this for all status changes.

Why use SHOULD when we can use MUST?

Requiring the latest version of an EIP to review seems like a pretty obvious request.


EIPS/eip-1.md
```diff
@@ -95,6 +93,8 @@ A PR moving an EIP from Last Call to Final SHOULD contain no changes other than
 
 **Living** - A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1.
 
+A PR moving an EIP from one status to another MUST contain no changes other than the status update. Any other changes MUST be submitted in one or more separate PRs. When the content of an EIP changes, the git branch of any PR that changes the status of that EIP MUST be updated to reflect the current version.
```
g11tech, line 96:
```suggestion
A PR moving an EIP from one status to another MUST contain no changes other than the status update. Any other changes MUST be submitted in one or more separate PRs. When the content of an EIP changes, the git branch of any PR that changes the status of that EIP MUST be rebased to reflect the current version.
```
updating changes via merge messes up the commit list on the PR, would prefer pure rebasing and hence the tighter language

SamWilsn:
I am rather torn about this. I like being able to see my comments and suggestions applied to the same PR that I made them on. Makes it easier to track progress.

If we can enforce that the PR's "Files Changed" tab shows the most recent version in `master`, with the status change diff I'd be less opposed. I think that's the intent of:

> When the content of an EIP changes, the git branch of any PR that changes the status of that EIP MUST be updated to reflect the current version.

Is it easy to automatically enforce this rule? I think there's an option in GitHub to require PRs be up-to-date, but that would require _all_ PRs follow that rule.

g11tech:
> I'd be less opposed if we can enforce that the PR's "Files Changed" tab shows the most recent version in `master` with the status change diff. I think that's the intent of:
> 

I think thats how it  normally works, the target/master branch gets auto updated if something is merged (and if there are conflicts they show up)

> Is it easy to automatically enforce this rule? I think there's an option in GitHub to require PRs be up-to-date, but that would require _all_ PRs follow that rule.

i think yes, for e.g. in ethereumjs repo we follow similar rule, however github also shows up with a button to rebase the PR with latest target/master at which then the PR becomes mergable

g11tech:
> 

i think auto magic could be very non intutive to the author who might want to pull/push exactly for the reasons that you mentioned.
However a force push and update by author(s) of PR is still ok and is not really a big ask.

We just need to block the PRs saying they need rebase as the PR is out of touch, this should also provide a signal to the author to re-review (or may be we can explicitly ask the author to rebase and review)

PR #6901, opened 2023-04-19
Update EIP-6454: Move to Final

ThunderDeliverer:
As all the discussions have been resolved and the Last call deadline has been reached, we propose moving EIP-6454 to Final.


SamWilsn:
@ethereum/eip-editors as of right now, I think the discussion thread is still pretty active, so I am hesitant to move to final.

EIPS/eip-6454.md
```diff
@@ -4,8 +4,7 @@ title: Minimal Transferable NFT detection interface
 description: A minimal extension to identify the transferability of Non-Fungible Tokens.
 author: Bruno Škvorc (@Swader), Francesco Sullo (@sullof), Steven Pineda (@steven2308), Stevan Bogosavljevic (@stevyhacker), Jan Turk (@ThunderDeliverer)
 discussions-to: https://ethereum-magicians.org/t/minimalistic-transferable-interface/12517
-status: Last Call
-last-call-deadline: 2023-04-07
+status: Final
```
SamWilsn, line 7: The last commit changing content was May 13th. You'll need to restart the last call period.

```suggestion
status: Last Call
last-call-deadline: 2023-05-30
```

SamWilsn (changes requested):

EIPS/eip-6454.md
steven2308: I've set it to the 27th.

SamWilsn (approved):

PR #6920, opened 2023-04-21
WIP: Eip physical transferable nft ibex addons

ibex-technology:
Proposed Addons in this PR:

- Allow transfer of attested NFTs by everyone.
- Don't burn the NFT once minted, instead transfer it to the contract as "floating" or similar.
- This eliminates the need for the Checksum URI as well as some parts of spec.

Also a point, but possibly not in this PR as it needs to be processed further:
- Draft out a "Transfer of Allowance" or similar to prevent misuse of floating Anchor in DeFi.


ibex-technology:
This was mistakingly opened by me. Sorry. Supposed to merge internally first.

PR #6925, opened 2023-04-24
Update EIP-5656: Move to Review

axic:


xinbenlv (approved):
LGTM! 
Don't forget to clean up the linter error

xinbenlv:
> To be fair this linter error makes little sense, as markdown renders properly on the website, and requiring extra empty lines feel weird. Will do nevertheless.

I agree very much with you that it doesn't make much sense to require extra empty lines,

Also I personally think it doesn't make too much sense to require committer to fix linter error (1) when the part of not touched is not even being changed in this commit, or (2) every-time the linter error comes from a change of editing policy(e.g. EIP->ERC), it could totally be batched before moving to last call.

axic:
@g11tech @lightclient @Pandapip1 @SamWilsn can we move this to review?

g11tech (approved):
:+1:

PR #6949, opened 2023-04-28
Update EIP-6672: Move to Last Call

Randyanto:
Changes list
- update implementation with Redeem and Cancel events
- change status to Last Call


SamWilsn (approved):
A friendly reminder than once a proposal goes into the Final status, it can no longer be edited, so please ensure that you have given your proposal an in-depth review before opening that pull request.

PR #6963, opened 2023-05-01
Add EIP: Multi Injected Provider Discovery

pedrouid:
Introducing a new pattern to discover multiple injected providers using window events to announce EIP-1193 providers in a two-communication standard between Ethereum libraries and Wallet peoviders.

**NOTE:** This EIP replaces the existing [EIP5749](https://eips.ethereum.org/EIPS/eip-5749)

------------

**What changes from EIP-5749?**
- EIP-6963 uses an event-based approach to discover multiple providers
- EIP-6963 does not extend EIP-1193 with info but instead exposes it in parallel
- EIP-6963 includes a local unique identifier and a global unique identifier for each provider
- EIP-6963 does not include provider description to stay agnostic to localization
- EIP-6963 supports any URI encoded image for icons (data, ipfs, http, etc)


EIPS/eip-6963.md
```diff
@@ -0,0 +1,121 @@
...
+
+Browser extensions are loaded in the web page and do not have a predictable pattern resulting in a race condition where the user is not in control to choose which wallet provider must be selected for exposing Ehtereum interface under the `window.ethereum` object.
+
+This results not only in a degraded user experience but also increases the barrier to entry for new browser extensions as users are forced to only install one browser extension at a time.
+
+Some browser extensions attempt to counteract this problem by delaying their injection to overwrite the same `window.ethereum` object which creates an unfair competition for wallet providers and lack of interoperability.
+
+In this proposal, we provide a solution that optimitizes for interoperability and enables fairer competition by lowering the barrier to entry for new wallet providers to improve the user experience on Ethereum.
+
+This is achieved by replacing the current window object `window.ethereum` with a new window object `window.evmproviders` that enables multiple browser extensions to be exposed for user selection in the web page.
```
jxom, line 31: I’m not too sure if we should be “replacing” `window.ethereum` as this will break every dapp and I’d say we want to avoid telling consumers to update their dapp by X date… 😅

IMO, `window.ethereum` has served us quite well and I don’t believe we should be replacing it entirely to solve an issue that affects a niche audience (pro-users & developers). Most “mainstream” users wouldn’t be using multiple browser extension wallets.

EIPS/eip-6963.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+The values in the `ProviderInfo` MUST be used as follows:
+
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- walletId - globally unique identifier of the walelt provider (eg. `my-wallet-extension` or `awesomewallet`)
+- name - human-readable name of the wallet provider (e.g. `MyWalletExtension` or `AwesomeWallet`)
+- icon - base64 encoded image (RFC-4648 complaint)
+- description - short description of the wallet provider (optional)
```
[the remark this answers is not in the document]
pedrouid: I would say localizations are out-of-scope and one of the reasons why I made it optional

EIPS/eip-6963.md
```diff
@@ -0,0 +1,121 @@
...
+  description?: string;
+}
+```
+
+The values in the `ProviderInfo` MUST be used as follows:
+
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- walletId - globally unique identifier of the walelt provider (eg. `my-wallet-extension` or `awesomewallet`)
+- name - human-readable name of the wallet provider (e.g. `MyWalletExtension` or `AwesomeWallet`)
+- icon - base64 encoded image (RFC-4648 complaint)
```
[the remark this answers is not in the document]
pedrouid: It's definitely a must... otherwise we are going to incentivize centralization of some from of wallet registry

WalletConnect protocol also requires wallets and dapps to include an icon to avoid exactly this

It might seem irrelevant for low-level libraries like web3.js, ethers.js and viem.js but the end goal of this standard is to display options to the user which is going to be important for libraries like Web3Modal and others

EIPS/eip-6963.md
pedrouid: Perhaps replacing is the wrong terminology here...

I meant more like prioritizing `window.evmproviders` usage over `window.ethereum`

You will notice a few lines after I mention how an Ethereum library can overwrite `window.ethereum` based on the user selection of a wallet provider

PS - I would avoid changing existing interfaces like `window.ethereum` since that would impact backwards compatiblity

EIPS/eip-6963.md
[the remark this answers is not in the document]
pedrouid: Actually let's just remove this one to avoid this issue 😅

EIPS/eip-6963.md
```diff
@@ -0,0 +1,118 @@
...
+  info: ProviderInfo;
+  ethereum: EIP1193Provider;
+}
+```
+
+The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and can be used to override the `window.ethereum` object once the user as explicitly selected it.
+
+### window.evmproviders
+
+A web app will be able to display multiple wallet providers that were injected by browser extensions by parsing the window object `window.evmproviders` which will include an array of wallet providers that follow the `EVMProvider` described above
```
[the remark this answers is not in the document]
ahollenbach: From my understanding, that falls under a phase two of this proposal (if at all):

> This EIP doesn't require supplanting window.ethereum, so it doesn't directly break existing applications. However, the recommended behavior of eventually supplanting window.ethereum would break existing applications that rely on it.

Even with a Proxy, the developer would need a list of injected wallets to make a decision on which they would want to load. Once this decision is made, a Proxy object wrapping `window.ethereum` could route requests to the intended provider. This feels like a nicer solution vs just overwriting `window.ethereum` once an EVM provider has been selected, but is addressing a slightly different problem in the provider selection chain.

EIPS/eip-6963.md
```diff
@@ -0,0 +1,118 @@
...
+status: Draft
+type: Standards Track
+category: Interface
+created: 2023-05-01
+requires: 1193
+---
+
+## Abstract
+
+A Javascript interface injected by browser extensions installed by the user that enables user selection from multiple wallet providers using a window object labelled `window.evemproviders` instead of overwriting `window.ethereum` object.
```
ahollenbach, line 16: typo: `window.evmproviders`

EIPS/eip-6963.md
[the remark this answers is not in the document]
pedrouid: That's fair enough 😆 it should be any uri path

EIPS/eip-6963.md
```diff
@@ -0,0 +1,118 @@
...
+
+Each wallet provider will be exposed with the following interface `ProviderInfo` that will be used to display to the user:
+
+```typescript
+/**
+ * Represents the assets needed to display a wallet
+ */
+interface ProviderInfo {
+  uuid: string;
+  walletId: string;
```
ahollenbach, line 46: Given the widespread misuse of `window.ethereum.isMetaMask = true`, are there any potential mitigations for wallets that just use the MetaMask `walletId` in their `ProviderInfo` and hope to inject first to be picked up by the wallet switcher?

It's not novel to this proposal - EIP-5749 would have been _more_ sensitive to this, but given that it's an array, the wallet switcher will somehow need to choose (for example, when the user clicks a "Connect with MetaMask" button in the UI  and `window.evmproviders.find(provider => provider.info.walletId === 'metamask')` feels like a default solution in the absence of mitigation.

There's a lot of incentive for a wallet provider to continue to duck-type as MetaMask in order to be prioritized (vs buried deep in a wallet selection screen) and with this EIP formalizing the notion of wallet providers whereas the previous standard did not, I'd be curious to mitigate in some way in a way that wasn't necessary previously.

EIPS/eip-6963.md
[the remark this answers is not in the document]
kvhnuke: I don't think it should be any uri, due to privacy concerns, we don't want users to be tracked by an image url. Image should be locally stored and best way to make sure that's the case if by making it base64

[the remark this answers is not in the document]
Adamj1232: Could we just standardize on an svg for icons? No tracking, no hosting, no url handling, just code.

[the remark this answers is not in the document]
jxom: > Could we just standardize on an svg for icons? No tracking, no hosting, no url handling, just svg code.

What if the svg has an `<image>` element? 😉

This is why I believe this should be part of another EIP. There are so many nuances with icons. We would probably need to standardize this attribute further to avoid misuse, and might also be reasonable to add a standard to define what constitutes of a “default” design spec for the icon (ie. it would be weird for one icon to have rounded corners, and the others do not).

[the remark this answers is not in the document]
hesterbruikman: > This is why I believe this should be part of another EIP. 
Can another EIP still be referenced in this one (6963)? To @pedrouid 's point, not having it will incentivise a centralised registry.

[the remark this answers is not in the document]
jxom: I wouldn't say a this centralized registry of _just_ icons would necessarily be a bad thing until we have a proper standard for icons? Some form of centralization will probably exist regardless as RainbowKit and ConnectKit have [differing](https://github.com/rainbow-me/rainbowkit/pull/1166#discussion_r1169440416) [requirements](https://github.com/rainbow-me/rainbowkit/pull/1041#discussion_r1164381765) for their icons.

cc @DanielSinclair @lochie @0xAsimetriq

EIPS/eip-6963.md
glitch-txs: I'm not sure if I understand, do you say wallets would still be trying to impersonate MetaMask?

I think unless you're trying to onboard users to install a specific wallet you can simply map through the whole array and display what the user has installed, there's no need to know what wallets your Dapp is displaying, as long as you can get a provider following 1193

Maybe this issue would depend on how Dapps would implement this

EIPS/eip-6963.md
kdenhartog: Making this an array rather than an object won't resolve the issue of prototype pollution here. 

For example, a wallet could intentionally `pop()` all other indexes off the array in order to make it so that only their wallet is found. If a display module read all indexes in the array to allow the user to select their wallet then this adversarial behavior would be encouraged by a wallet attempting to growth hack.

In my opinion we shouldn't be making the assumption that all wallets will behave benevolently here because we've seen that's not how it's currently playing out. Instead, we should build a solution that can withstand adversarial attempts to manipulate other wallets and other results either for benign growth reasons or for more malicious reasons such as to steal user funds.

kdenhartog (changes requested):
Left a few concerns here, most notably we're not addressing the concern of prototype pollution still and assuming that all wallets will behave benevolently which seems like a flawed assumption to make.

EIPS/eip-6963.md
```diff
@@ -0,0 +1,118 @@
...
+  name: string;
+  icon: string;
+}
+```
+
+The values in the `ProviderInfo` MUST be used as follows:
+
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
```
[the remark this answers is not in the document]
jxom: I was also going to ask the same question.

EIPS/eip-draft_multi_injected_provider_interf.md
```diff
@@ -0,0 +1,118 @@
+---
+eip: .nan
```
abcoathup, line 2:
```suggestion
eip: 6963
```
Assigning PR number as EIP number

EIPS/eip-6963.md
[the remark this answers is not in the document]
lochie: I have no strong opinions either way. ConnectKit and other libraries could always just ignore these icons or just use them as fallbacks. But logos/icons can get pretty complex especially when considering width/height/ratio/contrast/rounded/etc. as context is important when using branding assets.
If this were to include icons/logos then we'd probably also need to include branding colors, which leads down a rabbit hole...

Correct me if I'm wrong, but image-tracking is probably a non-issue since if the provider exists in your `window.ethereum` object (or in this case `window.evmproviders`) then they'd be able to track in many other ways, so the svg/image uri would be the least of your worries.

Overall I think it's a safer bet to _not_ include icons, but also is there any harm to include them?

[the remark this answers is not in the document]
0xAsimetriq: I similarly have no strong opinion on icons, as lochie mentioned, we'd likely be overriding these in Web3Modal anyway. I do appreciate simplicity though, so I don't see icons as necessary. What's important imho is:

1. Structure that allows to access multiple providers
2. UUID's for each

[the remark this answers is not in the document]
pedrouid: The point of standards to bring consensus over multiple parties

Honestly I think we are nitpicking too much on this topic

We should just agree that icons should be URIs, examples:

1.   data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==
2. ipfs://QmQqzMTavQgT4f4T5v6PWBp7XNKtoPmC9jvn12WPT3gkSE
3. https://ethereum.org/static/f541df14fca86543040c113725b5bd1a/99bcf/metamask.webp

URIs should be pretty much non-controversial... each wallet can decide which type of URI to use

EIPS/eip-6963.md
pedrouid: Actually I was thinking we could do differently and just use reversed domains here instead

com.metamask
com.coinbase
io.rabby
xyz.taho
me.rainbow
com.spot-wallet

pedrouid:
Thank you @glitch-txs @SamWilsn @MicahZoltu for the idea to add event listeners to track whenever a new EVM Provider is added to the window object 💪

EIPS/eip-6963.md
pedrouid: I've added Event Listeners which would allow wallet providers to essentially make themselves observables just in case they are erased from the window object

EIPS/eip-6963.md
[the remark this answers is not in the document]
pedrouid: This was feedback from the TG group that we had created for EIP-5749

The feedback was that `uuid` were insufficient to identify wallets thus we needed `walletId` 

However using just `walletId` would not allow us to prevent duplicate or wallet providers overriding each other

Having two identifiers prevents these conflicts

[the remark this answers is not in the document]
ahollenbach: I missed that earlier conversation - is this `uuid` a permanently fixed value or set on page load? i.e. is the guidance here that a wallet extension should ship with a uuid that won't change across versions?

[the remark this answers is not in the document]
jxom: How does a local id prevent conflicts? How would you distinguish between two arbitrary local ids? 

Expanding on @ahollenbach's question, if the `uuid` is a deterministic fixed value, how is that different from the `walletId`? Another extension could still disguise themselves as that `uuid` if it is fixed. But, if the `uuid` is non-deterministic & set on page load, how do we distinguish between them (as I mentioned above)?

Feel like a global `walletId` is enough.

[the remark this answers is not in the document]
jxom: I'm not really understanding how both a `uuid` and `walletId` prevents conflicts.

Expanding on @ahollenbach's question, if the `uuid` is a deterministic fixed value, how is that different from the `walletId`? Another extension could still disguise themselves as that `uuid` if so; which means that `walletId` and `uuid` **are not** mutually exclusive so a global `walletId` is probably enough. 

On the other hand, if the `uuid` is arbitrary/non-deterministic & set on page load, how do we distinguish between them (as I mentioned above)?

EIPS/eip-draft_multi_injected_provider_interf.md
```diff
@@ -0,0 +1,199 @@
+---
+eip: draft_multi_injected_provider_interf
```
abcoathup, line 2:
```suggestion
eip: 6963
```
Assigning PR number as EIP number

EIPS/eip-draft_multi_injected_provider_interf.md
```diff
@@ -0,0 +1,199 @@
...
+  info: ProviderInfo;
+  ethereum: EIP1193Provider;
+}
+```
+
+The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and can be used to override the `window.ethereum` object once the user as explicitly selected it.
+
+### window.evmproviders
+
+A web app will be able to display multiple wallet providers that were injected by browser extensions by parsing the window object `window.evmproviders` which will include an array of wallet providers that follow the `EVMProvider` described above
```
kdenhartog, line 74: I don't believe we need to inject this if we're using event listeners now. Should we remove this?

GregTheGreek:
LGTM

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+}
+```
+
+The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and can be used to override the `window.ethereum` object once the user as explicitly selected it.
+
+### Window Event
+
+Different wallet providers might inject their scripts at different timeframes plus there is no guarantees that the Ethereum library in the web page will be loaded before all injected scripts are present in the page.
+
+To emit events it must use `window.dispatchEvent` and to observe events it must use `window.addEventListener`. There will be two event interfaces:
```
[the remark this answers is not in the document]
pedrouid: The advantage of this proposal being EVM specific is that EIP-1193 is a significantly dominant and it will take a while until we see a large adoption of something more agnostic like CAIP-25

My suggestion is that we create a parallel CAIP-XXX that will also enable this exact same event-based approach for future CAIP-25 providers

I've been personally pushing CAIP-25 adoption for 2 years and I suspect it will take another 3 years to see it heavily used in production

So EIP-6963 will be useful for at least 3 years if not more since it's unlikely that EIP-1193 will be deprecated anytime soon

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+
+Therefore both Wallet providers and the Ethereum library must emit events when they initialize.
+
+```typescript
+// Ethereum library initializes
+window.dispatchEvent(new Event("evmProviderRequested"));
+
+// Wallet provider initializes
+window.dispatchEvent(
+  new CustomEvent("evmProviderAnnounced", { detail: provider })
```
darrylyeo, line 99: Thoughts on "namespacing" these event names? Something like `"evm:announceProvider"` or `"eip6963:announceProvider"`.

Verbs in JavaScript event names are also traditionally written in present tense. I think we should match that convention.

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+  type: "evmProviderAnnounced";
+  detail: EVMProvider;
+}
+```
+
+Therefore both Wallet providers and the Ethereum library must emit events when they initialize.
+
+```typescript
+// Ethereum library initializes
+window.dispatchEvent(new Event("evmProviderRequested"));
```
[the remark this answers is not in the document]
pedrouid: There is no data currently required for this event

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+  icon: string;
+}
+```
+
+The values in the `ProviderInfo` MUST be used as follows:
+
+- walletId - globally unique identifier of the wallet provider (eg. `io.dopewallet.extension` or `awesomewallet`)
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- name - human-readable name of the wallet provider (e.g. `DopeWalletExtension` or `AwesomeWallet`)
+- icon - uri encoded image (RFC-3986 complaint)
```
bpierre, line 57: What would you think of suggesting some image formats and dimensions here (e.g. lossless format, square, 96x96 or above), so that consumers can rely on a common base?

Also I think this field accepts any valid URI, including HTTP ones, as I understand it from the examples and the reference to RFC-3986. Should it say “image URI (RFC-3986 compliant)” to remove any ambiguity?

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+  const providers: EVMProvider[] = [];
+
+  window.addEventListener(
+    "evmProviderAnnounced",
+    (event: EVMProviderAnnouncedEvent) => {
+      providers.push(event.detail);
+    }
+  );
+
+  window.dispatchEvent(new Event("evmProviderRequested"));
```
jxom, line 199: Can we provide guidance on what to do if a browser extension injects itself after this has been invoked?

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+}
+```
```
tmm, line 201: what happens if a provider:
- hasn't added a listener before "request" is dispatched?
- is removed after it "announced"?
- tries to look like another provider by announcing as metamask?

EIPS/eip-6963.md
jxom: Related: https://github.com/ethereum/EIPs/pull/6963#discussion_r1181838380

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+- icon - uri encoded image (RFC-3986 complaint)
+
+### EVMProvider
+
+The wallet provider will also expose their own [EIP-1193](./eip-1193.md) provider interface in parallel with the provider info provided above in the following interface `EVMProvider`
+
+```typescript
+interface EVMProvider {
+  info: ProviderInfo;
+  ethereum: EIP1193Provider;
```
jxom, line 66: Should this be attributed as `provider`? Kinda contradicts itself being `ethereum` when it represents an "evm provider", and feels like it's a provider within a provider lol.

Thought: align the type of `EVMProvider` closer to event terminology. If we are rolling with `MessageEvent`, this could be `EVMEventData`:

```ts
interface EVMEventData {
  info: ProviderInfo;
  provider: EIP1193Provider;
}
```

This represents itself much more clearer in implementation:

```ts
window.postMessage(
  new MessageEvent<EVMEventData>('evmProviderAnnounced', {
    data: {
      info: {
        walletId: 'org.example.wallet',
        uuid: '350670db-19fa-4704-a166-e52e178b59d2',
        name: 'Example Wallet',
        icon: 'https://wallet.example.org/icon.png',
      },
      provider: ethereum,
    },
  })
);
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,211 @@
...
+}
+```
```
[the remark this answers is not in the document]
jxom: This terminology is more inferred, which is also a good thing, because I don't want to explain to consumers what an EVM Provider and an EIP1193 Provider is.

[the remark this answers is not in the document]
glitch-txs: the EIP-5749 one was called `ProviderWithInfo`

pedrouid:
This PR is pretty much ready to merge but I've collected some feedback to include my next revision:
- Use `postMessage` instead of `dispatchEvent`
- Recommend image formats and size for icons
- Rename events to `eip6963:requestProvider` and `eip693:announceProvider`
- Rename `EVMProvider` interface to `AnnounceProviderData`
- Rename `ethereum` param to `provider` (inside `EVMProvider`)

Targeting next PR revision to be included by May 19

ekc9399:
> This PR is pretty much ready to merge but I've collected some feedback to include my next revision:
> 
> * Use `postMessage` instead of `dispatchEvent`
> * Recommend image formats and size for icons (e.g. lossless, square, 96x96 min)
> * Rename events to `eip6963:requestProvider` and `eip693:announceProvider`
> * Rename `EVMProvider` interface to `AnnounceProviderData`
> * Rename `ethereum` param to `provider` (inside `EVMProvider`)
> 
> Targeting next PR revision to be included by May 19

ghost (dismissed):

pedrouid:
100% agree @MicahZoltu 👍  You can see it's already listed on my list of changes for the next revision

> This PR is pretty much ready to merge but I've collected some feedback to include my next revision:
> 
> * Use `postMessage` instead of `dispatchEvent`
> * Recommend image formats and size for icons (e.g. lossless, square, 96x96 min)
> * Rename events to `eip6963:requestProvider` and `eip693:announceProvider`
> * Rename `EVMProvider` interface to `AnnounceProviderData`
> * Rename `ethereum` param to `provider` (inside `EVMProvider`)
> 
> Targeting next PR revision to be included by May 19

I will have an updated PR soon 💪

glitch-txs (dismissed):

EIPS/eip-6963.md
```diff
@@ -0,0 +1,234 @@
...
+A globally unique identifier is used to for machine-readable detection of which wallet is being used and it can take different formats, for example:
+
+```sh
+# lowercase name
+awesomewallet
+
+# legacy JS variable
+isCoolWallet
+
+# reserved domain
```
pedrouid, line 136:
```suggestion
# reversed domain
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,234 @@
...
+    announcedProvider: EIP6963AnnouncedProvider;
+  };
+}
+```
+
+Therefore both Wallet providers and the Ethereum library must emit events when they initialize.
+
+```typescript
+// Ethereum library initializes
+window.postMessage({ eventName: "eip6963:requestProvider" }));
```
```suggestion
window.postMessage({ eventName: "eip6963:requestProvider" });
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,234 @@
...
+  icon: string;
+}
+```
+
+The values in the `EIP6963ProviderInfo` MUST be used as follows:
+
+- walletId - globally unique identifier of the wallet provider (eg. `io.dopewallet.extension` or `awesomewallet`)
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- name - human-readable name of the wallet provider (e.g. `DopeWalletExtension` or `AwesomeWallet`)
+- icon - uri encoded image (RFC-3986 complaint / image must be squared with 96x96px minimum resolution)
```
bpierre, line 57:
```suggestion
- icon - uri encoded image (RFC-3986 compliant, image must be squared with 96x96px minimum resolution)
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,234 @@
...
+  data: {
+    eventName: "eip6963:requestProvider";
+  };
+}
+
+// Annoucing an EVM provider
+interface EIP6963AnnounceProviderEvent extends MessageEvent {
+  data: {
+    eventName: "eip6963:announceProvider";
+    announcedProvider: EIP6963AnnouncedProvider;
```
bpierre, line 90: Unfortunately I don’t think `postMessage()` can be used to pass an EIP-1193 object as it contains functions, [which will throw a serialization error](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#things_that_dont_work_with_structured_clone) if we attempt to pass it to `postMessage()`. /cc @MicahZoltu

MicahZoltu: It depends on the browser, but generally you should stick to anything that can JSON serialize (no functions, no bigints, no Uint8Array) as that is the lowest common denominator (Chromium).

bpierre: `BigInt` and `Uint8Array` (or any typed array with certain limitations) must be supported by browsers (see the [relevant part](https://html.spec.whatwg.org/multipage/structured-data.html#structuredserializeinternal) of the HTML spec) to serialize objects passed to `postMessage()`. But yes the issue is that functions are not allowed, making EIP-1193 providers unserializable with this algorithm, and incompatible with `postMessage()`.

MicahZoltu: Hmmm, I must be thinking of some other communication channel then, sorry!

pedrouid: Personally I also find dispatchEvent to be a cleaner approach but it does reduce the interoperability for this standard

Using postMessage not only makes it compatible with browser extensions but also iframes

bpierre: It would be much easier if EIP-1193 was using a message-based API rather than shipping an object with its own methods… maybe one day :sweat_smile:

As things are, I only see one way to keep using `postMessage`, which would be to communicate the name of the `ethereum` object attached to `window` rather than the object itself… or we could reuse the uuid for this, e.g.:

```ts
function onPageLoad() {
  let provider: EIP1193Provider;
  const uuid = "350670db-19fa-4704-a166-e52e178b59d2";

  window.ethereum = provider;

  // we attach the provider using the uuid
  window[`ethereum_eip6963_${uuid}`] = provider;

  function announceProvider() {
    const info: EIP6963ProviderInfo = {
      walletId: "org.example.wallet",
      uuid,
      name: "Example Wallet",
      icon: "https://wallet.example.org/icon.png",
    };
    const announcedProvider = { info, provider };
    window.postMessage({
      eventName: "eip6963:announceProvider",

      // we only communicate the uuid here
      announcedProviderUuid: uuid,
    });
  }

  window.addEventListener("message", (event: MessageEvent) => {
    if (event.data.eventName === "eip6963:requestProvider") {
      announceProvider();
    }
  });

  announceProvider();
}
```

But it would require to have an access to `window` to inject the provider no matter what, which kind of defeats the purpose of using `postMessage`.

So yeah reverting back to `dispatchEvent()` is probably the best option we have, since implementing the EIP-1193 methods require to rely on the current context no matter what, in order to communicate with the wallet.

bpierre:
Should wallets have a way to communicate that they are getting removed? It would for example allow users to disable a wallet extension and seeing it reflected in their opened pages, without having to refresh the opened pages to clear the MIPI states.

We could also recommend to listen to the [`disconnect`](https://github.com/ethereum/EIPs/blob/d0d97017e01d6d767d574d8e4a37a5f82649daac/EIPS/eip-1193.md#disconnect-1) event for each announced provider, but I think a provider could be disconnected temporarily (and should probably still appear in current the app UI), while being removed is more permanent (and it should disappear from the current app UI).

EIPS/eip-6963.md
MicahZoltu: Requiring the app to inject into the page's context defeats the purpose of using `window.postMessage`.  The primary argument in favor of using `window.postMessage` is that it is the only communication channel that works for both extensions and iframes and any other kind of sandboxing or containment.  As long as the app has to inject content into the page's context, then extensions will have to do the hacky thing where they inject a script into the DOM which then executes in the page's context to write into the page (something that should generally not be done).

darrylyeo: +1 for switching to a message-based system. If we're already going with `postMessage` for announcing the provider itself, might as well go all the way and replace all the EIP-1193 functions with standardized messages too.

EIPS/eip-6963.md
```diff
@@ -0,0 +1,234 @@
...
+  window.addEventListener(
+    "message",
+    (event: MessageEvent) => {
+      if (event.data.eventName === "eip6963:announceProvider") {
+        providers.push(event.data.announcedProvider);
+      }
+    }
+  );
+
+  window.postMessage({ eventName: "eip6963:requestProvider" }));
```
glitch-txs, line 222:
```suggestion
  window.postMessage({ eventName: "eip6963:requestProvider" });
```

EIPS/eip-6963.md
pedrouid: After doing some testing with both `postMessage` and `dispatchEvent`... I think the best solution is to revert to `dispatchEvent`

Using postMessage with the current spec will break as flagged by @bpierre (screenshot below)
Screenshot 2023-05-14 at 10 40 38

The alternatives are the following:
1. use postMessage with a window map using uuid
2. use postMessage exclusively without EIP-1193 interface
3. revert to dispatchEvent instead of postMessage

Option 1 with the window map using uuid is a simple solution to this problem but defies the whole purpose of postMessage since it will require wallet providers to have window access. If they have window access then they can also use dispatchEvent.

Option 2 would require us to remove the dependency on EIP-1193 spec and instead make all providers use window.postMessage exclusively. IMHO this adds too much friction from both Ethereum libraries and Wallet providers to adapt

Option 3 would simply replace postMessage with dispatchEvent which would immediately make this EIP feasible to be implemented with little to no breaking changes in both Ethereum libraries and Wallet providers (screenshot below)

Screenshot 2023-05-14 at 10 50 18

The impact that Option 3 has completely outweights the very low cost it has in terms of developer time in the community to adopt across multiple Ethereum libraries and Wallet providers

pedrouid:
**TLDR'** After doing some testing with both `postMessage` and `dispatchEvent`... I think the best solution is to revert to `dispatchEvent`

_________

**PROBLEM**

Using postMessage with the current spec will break as flagged by @bpierre (screenshot below)
Screenshot 2023-05-14 at 10 40 38

___________

**SOLUTION**

The alternatives are the following:
1. use postMessage with a window map using uuid
2. use postMessage exclusively without EIP-1193 interface
3. revert to dispatchEvent instead of postMessage

Option 1 with the window map using uuid is a simple solution to this problem but defies the whole purpose of postMessage since it will require wallet providers to have window access. If they have window access then they can also use dispatchEvent.

Option 2 would require us to remove the dependency on EIP-1193 spec and instead make all providers use window.postMessage exclusively. IMHO this adds too much friction from both Ethereum libraries and Wallet providers to adapt

Option 3 would simply replace postMessage with dispatchEvent which would immediately make this EIP feasible to be implemented with little to no breaking changes in both Ethereum libraries and Wallet providers (screenshot below)

Screenshot 2023-05-14 at 10 50 18

_______

**CONCLUSION**

The impact that Option 3 has completely outweights the very low cost it has in terms of developer time in the community to adopt across multiple Ethereum libraries and Wallet providers

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+The values in the `EIP6963ProviderInfo` MUST be used as follows:
+
+- walletId - globally unique identifier of the wallet provider (eg. `io.dopewallet.extension` or `awesomewallet`)
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- name - human-readable name of the wallet provider (e.g. `DopeWalletExtension` or `AwesomeWallet`)
+- icon - uri encoded image (RFC-3986 compliant, image must be squared with 96x96px minimum resolution)
+
+### Announced Provider
```
pedrouid, line 59:
```suggestion
### Provider Detail
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+status: Draft
+type: Standards Track
+category: Interface
+created: 2023-05-01
+requires: 1193
+---
+
+## Abstract
+
+A Javascript communication protocol that enables multiple wallet providers to be discovered from an EVM-based application using a browser extension injected script through window events.
```
pedrouid, line 16:
```suggestion
A Javascript communication standard that enables multiple wallet providers to be discovered from an EVM-based application using a browser extension injected script through window events.
```

glitch-txs (dismissed):

kdenhartog (dismissed):
This is an excellent starting point to move forward with

pedrouid:
Ready to merge 👍 cc @axic @lightclient @Pandapip1 @SamWilsn

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+The wallet provider will also expose their own [EIP-1193](./eip-1193.md) provider interface in parallel with the provider info provided above in the following interface `EIP6963ProviderDetail`
+
+```typescript
+interface EIP6963ProviderDetail {
+  info: EIP6963ProviderInfo;
+  provider: EIP1193Provider;
+}
+```
+
+The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and can be used to override the `window.ethereum` object once the user as explicitly selected it.
```
SamWilsn, line 70:
```suggestion
The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and can be used to override the `window.ethereum` object once the user has explicitly selected it.
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+The values in the `EIP6963ProviderInfo` MUST be used as follows:
+
+- walletId - globally unique identifier of the wallet provider (eg. `io.dopewallet.extension` or `awesomewallet`)
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- name - human-readable name of the wallet provider (e.g. `DopeWalletExtension` or `AwesomeWallet`)
+- icon - uri encoded image (RFC-3986 compliant, image must be squared with 96x96px minimum resolution)
+
+### Provider Detail
+
+The wallet provider will also expose their own [EIP-1193](./eip-1193.md) provider interface in parallel with the provider info provided above in the following interface `EIP6963ProviderDetail`
```
SamWilsn, line 61: What does "expose" mean here?

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+
+This is achieved by introducing a set of window events to provide a two-way communication protocol between Ethereum libraries and injected scripts provided by browser extensions thus enabling users to select their wallet of choice.
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Provider Info
+
+Each wallet provider will be exposed with the following interface `EIP6963ProviderInfo` that will be used to display to the user:
```
SamWilsn, line 38: What does "exposed" mean here?

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+  info: EIP6963ProviderInfo;
+  provider: EIP1193Provider;
+}
+```
+
+The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and can be used to override the `window.ethereum` object once the user as explicitly selected it.
+
+### Window Event
+
+Different wallet providers might inject their scripts at different timeframes plus there is no guarantees that the Ethereum library in the web page will be loaded before all injected scripts are present in the page.
```
SamWilsn, line 74:
```suggestion
Different wallet providers might inject their scripts at different times, plus there is no guarantees that the Ethereum library in the web page will be loaded before all injected scripts are present in the page.
```

SamWilsn, line 74: This blurb probably belongs in the Rationale section, and not the specification.

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+
+To emit events it must use `window.dispatchEvent` and to observe events it must use `window.addEventListener`. There will be two event interfaces:
```
SamWilsn, line 76: What does "it" refer to here?

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// Annoucing an EVM provider
+interface EIP6963AnnounceProviderEvent extends CustomEvent {
+  type: "eip6963:announceProvider";
+  detail: EIP6963ProviderDetail;
+}
+```
+
+Therefore both Wallet providers and the Ethereum library must emit events when they initialize.
```
SamWilsn, line 91:
```suggestion
Therefore both wallet providers and the Ethereum library must emit events when they initialize.
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+To emit events it must use `window.dispatchEvent` and to observe events it must use `window.addEventListener`. There will be two event interfaces:
+
+```typescript
+// Requesting an EVM provider
+interface EIP6963RequestProviderEvent extends Event {
+  type: "eip6963:requestProvider";
+}
+
+// Annoucing an EVM provider
+interface EIP6963AnnounceProviderEvent extends CustomEvent {
```
SamWilsn, line 85: This interface uses `CustomEvent` and the preceding one uses `Event`. Is that intentional?

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+
+Regarding the announced provider interface (`EIP6963ProviderDetail`) it was important to leave the [EIP-1193](./eip-1193.md) provider interface untouched for backwards-compatiblity therefore it's exposed in parallel.
+
+### Identifiers
+
+A locally unique identifier prevents from conflicts using the same globally unique identifier by using UUID v4.0
+
+A globally unique identifier is used to for machine-readable detection of which wallet is being used and it can take different formats, for example:
+
+```sh
```
SamWilsn, line 126:
````suggestion
```text
````

**Comment by asciiman** [comment]
What does "Ethereum library" mean in this context? Is it talking about a wallet connection library like Web3 Onboard or Rainbowkit?

File: EIPS/eip-6963.md [hunk 34]
```diff
@@ -0,0 +1,230 @@
+---
+eip: 6963
+title: Multi Injected Provider Discovery
```
pedrouid, line 3:
```suggestion
title: Multi Injected EIP-1193 Provider Discovery
```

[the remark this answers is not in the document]
pedrouid: I feel like it doesn't roll out of the tongue so much if we make it longer so if you are ok I will leave it as is

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+status: Draft
+type: Standards Track
+category: Interface
+created: 2023-05-01
+requires: 1193
+---
+
+## Abstract
+
+A Javascript communication standard that enables multiple wallet providers to be discovered from an EVM-based application using a browser extension injected script through window events.
```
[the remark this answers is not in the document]
pedrouid: This suggestion is actually not accurate since this EIP is NOT an alternative to EIP-1193 since `window.ethereum` is not part of the EIP-1193 spec

In fact this EIP depends on EIP-1193 but instead provides an alternative discovery mechanism to `window.ethereum` which is unwritten industry standard right now

Here is my counter-suggestion:

```suggestion
An alternative discovery mechanism to `window.ethereum` for EIP-1193 providers which supports discovering multiple injected wallet providers in a web page using Javascript's `window` events.
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+
+This is achieved by introducing a set of window events to provide a two-way communication protocol between Ethereum libraries and injected scripts provided by browser extensions thus enabling users to select their wallet of choice.
+
+## Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Provider Info
+
+Each wallet provider will be exposed with the following interface `EIP6963ProviderInfo` that will be used to display to the user:
```
pedrouid, line 38: Announced is a better term

```suggestion
Each wallet provider will be announced with the following interface `EIP6963ProviderInfo` that will be used to display to the user:
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+The values in the `EIP6963ProviderInfo` MUST be used as follows:
+
+- walletId - globally unique identifier of the wallet provider (eg. `io.dopewallet.extension` or `awesomewallet`)
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- name - human-readable name of the wallet provider (e.g. `DopeWalletExtension` or `AwesomeWallet`)
+- icon - uri encoded image (RFC-3986 compliant, image must be squared with 96x96px minimum resolution)
+
+### Provider Detail
+
+The wallet provider will also expose their own [EIP-1193](./eip-1193.md) provider interface in parallel with the provider info provided above in the following interface `EIP6963ProviderDetail`
```
pedrouid, line 61:
```suggestion
The wallet provider will also announce their own [EIP-1193](./eip-1193.md) provider interface in parallel with the provider info provided above in the following interface `EIP6963ProviderDetail`
```

EIPS/eip-6963.md
pedrouid: Use `announced` instead

EIPS/eip-6963.md
pedrouid: Yes. `CustomEvent` is only needed if data or detail is included

EIPS/eip-6963.md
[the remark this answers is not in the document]
danfinlay: How about "Wallet detection without injection"?

pedrouid:
> What does "Ethereum library" mean in this context? Is it talking about a wallet connection library like Web3 Onboard or Rainbowkit?

That's correct @asciiman! This could include Web3Modal, Web3React, Web3Onboard, RainbowKit, ConnectKit, web3.js, wagmi.js, ethers.js, viem.js, etc

danfinlay:
This is looking good to me. I'm glad to see an approach that doesn't involve more globals. Have we validated that this approach can work under manifest v3 without the need of the contentscript permission?

One motivation to move off in-page injection is to reduce the amount of permissions required to install a wallet WebExtension (reducing the harm one can do!) This current approach still involves returning a JS object as the provider, and I _think_ that will still require a contentscript. I think it would likely be a long-term improvement to establish a bidirectional message channel using PostMessage with the new provider, rather than returning a JS object (as convenient as that is).

Maybe that's better left for a future improvement. I'll try to get more info on that topic here soon.

kvhnuke (dismissed):

EIPS/eip-6963.md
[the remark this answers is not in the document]
Adamj1232: I do agree here somewhat - although it doesnt roll off the tounge as @pedrouid noted it does provide clarity on what kind of "provider" 
```suggestion
title: Multi Injected Wallet Provider Discovery
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+
+Browser extensions are loaded in the web page in an unpredictable and unstable order, resulting in a race condition where the user does not have control over which wallet provider is selected to expose the Ethereum interface under the `window.ethereum` object. Instead, the last wallet to load usually wins.
+
+This results not only in a degraded user experience but also increases the barrier to entry for new browser extensions as users are forced to only install one browser extension at a time.
+
+Some browser extensions attempt to counteract this problem by delaying their injection to overwrite the same `window.ethereum` object which creates an unfair competition for wallet providers and lack of interoperability.
+
+In this proposal, we provide a solution that optimitizes for interoperability of multiple wallet providers and enables fairer competition by lowering the barrier to entry for new wallet providers to improve the user experience on Ethereum.
```
Adamj1232, line 28: This line reads a little off, maybe something like:
```suggestion
In this proposal, we present a solution that focuses on optimizing the interoperability of multiple wallet providers. This solution aims to foster fairer competition by reducing the barriers to entry for new wallet providers, along with enhancing the user experience on the Ethereum network.
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+        "number-of-pages": "61",
+        "publisher": "RFC Editor",
+        "title": "Uniform Resource Identifier (URI): Generic Syntax",
+        "URL": "https://www.rfc-editor.org/info/rfc3986"
+    }
+    ```
+
+### Provider Detail
+
+The wallet provider will also announce their own [EIP-1193](./eip-1193.md) provider interface in parallel with the provider info provided above in the following interface `EIP6963ProviderDetail`
```
Adamj1232, line 129: To not get lost in repetition of the word "provide(r)"
```suggestion
The wallet provider will also announce their own [EIP-1193](./eip-1193.md) provider interface in parallel with the provider info defined above in the following interface `EIP6963ProviderDetail`
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+    announceProvider();
+  }
+);
+```
+
+## Rationale
+
+### Interfaces
+
+Standardizing a provider info interface (`EIP6963ProviderInfo`) allows determining the necessary information to populate a wallet selection popup. This is particularly useful for web3 onboarding libraries such as Web3Modal, RainbowKit, Web3Onboard, ConnectKit, etc.
```
Adamj1232, line 184:
```suggestion
Standardizing a provider info interface (`EIP6963ProviderInfo`) allows determining the necessary information to populate a wallet selection popup. This is particularly useful for web3 onboarding libraries such as Web3Modal, RainbowKit, Web3-Onboard, ConnectKit, etc.
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+
+Regarding the announced provider interface (`EIP6963ProviderDetail`) it was important to leave the [EIP-1193](./eip-1193.md) provider interface untouched for backwards-compatiblity therefore it's exposed in parallel.
+
+### Identifiers
+
+A locally unique identifier prevents from conflicts using the same globally unique identifier by using UUID v4.0
+
+A globally unique identifier is used to for machine-readable detection of which wallet is being used and it can take different formats, for example:
```
Adamj1232, line 192:
```suggestion
A globally unique identifier is used for machine-readable detection of the currently active wallet and it can take different formats, for example:
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+data:image/svg+xml,<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="32px" height="32px" viewBox="0 0 32 32"><circle fill="red" cx="16" cy="16" r="12"/></svg>
+# png (data uri)
+data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==
+# png (ipfs uri)
+ipfs://QmZo7gsostaUdrV1peFz4cX6Z2TgriFJJP4VahG5knVm29
+# webp (http uri)
+https://ethereum.org/static/f541df14fca86543040c113725b5bd1a/99bcf/metamask.webp
+```
+
+Additionally the image must be squared with 96x96px minimum resolution. Image format is recommended to be lossless like PNG or WebP or alternatively vectorial like SVG. We strongly disencourage lossy formats like JPG/JPEG.
```
Adamj1232, line 220:
```suggestion
Additionally the image must be squared with 96x96px minimum resolution. Image format is recommended to be lossless like PNG and WebP or alternatively vectorial like SVG. We strongly discourage lossy formats like JPG/JPEG.
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+
+### Events
+
+Previous proposal introduced mechanisms that relied on a single window object that could be overriden by multiple parties.
+
+Therefore using an event-based approach we avoid these race conditions and potential attacks by making the communication available both ways.
+
+To follow Javascript event name conventions the names are written in present tense and are prefixed with the standard number (`eip6963`).
```
Adamj1232, line 228:
```suggestion
To follow the Javascript event name conventions, the names are written in present tense and are prefixed with the standard number (`eip6963`).
```

EIPS/eip-6963.md
tmm: Curious why we want the namespace? For example, common browser event types (e.g. `"mousedown"`) aren't namespaced, like `"ui-event-20040104:mousedown"`.

`"announceProvider"` is easier to remember and isn't tightly coupled to this specific snapshot in time (e.g. random EIP number). If a namespace is important, seems like `"evm"` would be better suited than `"eip6963"`.

EIPS/eip-6963.md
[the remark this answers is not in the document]
pedrouid: @danfinlay EIP-6963 still requires injection but just uses events instead of globals

@Adamj1232 yeah I guess we could "Wallet" if everyone finds it helpful

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+    announceProvider();
+  }
+);
+```
+
+## Rationale
+
+### Interfaces
+
+Standardizing a provider info interface (`EIP6963ProviderInfo`) allows determining the necessary information to populate a wallet selection popup. This is particularly useful for web3 onboarding libraries such as Web3Modal, RainbowKit, Web3-Onboard, ConnectKit, etc.
```
pedrouid, line 184:
```suggestion
Standardizing a provider info interface (`EIP6963ProviderInfo`) allows determining the necessary information to populate a wallet selection popup. This is particularly useful for Ethereum libraries such as Web3Modal, RainbowKit, Web3-Onboard, ConnectKit, etc.
```

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+
+Browser extensions are loaded in the web page in an unpredictable and unstable order, resulting in a race condition where the user does not have control over which wallet provider is selected to expose the Ethereum interface under the `window.ethereum` object. Instead, the last wallet to load usually wins.
+
+This results not only in a degraded user experience but also increases the barrier to entry for new browser extensions as users are forced to only install one browser extension at a time.
+
+Some browser extensions attempt to counteract this problem by delaying their injection to overwrite the same `window.ethereum` object which creates an unfair competition for wallet providers and lack of interoperability.
+
+In this proposal, we present a solution that focuses on optimizing the interoperability of multiple wallet providers. This solution aims to foster fairer competition by reducing the barriers to entry for new wallet providers, along with enhancing the user experience on the Ethereum network.
```
pedrouid, line 28:
```suggestion
In this proposal, we present a solution that focuses on optimizing the interoperability of multiple wallet providers. This solution aims to foster fairer competition by reducing the barriers to entry for new wallet providers, along with enhancing the user experience on Ethereum networks.
```

EIPS/eip-6963.md
tmm: cc @pedrouid  @darrylyeo

EIPS/eip-6963.md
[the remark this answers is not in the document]
darrylyeo: "Event-Based Wallet Discovery" perhaps

EIPS/eip-6963.md
darrylyeo: I think `CustomEvent` is still better and more semantic. The `Event` type usually refers to a built-in event in my experience.

darrylyeo: It would also make introspection and TypeScript type narrowing slightly easier for patterns like isomorphic callback handlers:

```ts
const onCallback = (e: Event) => {
  if (e instanceof CustomEvent) {
    ...
  }
}
```

danfinlay:
> One motivation to move off in-page injection is to reduce the amount of permissions required to install a wallet WebExtension (reducing the harm one can do!) This current approach still involves returning a JS object as the provider, and I _think_ that will still require a contentscript. I think it would likely be a long-term improvement to establish a bidirectional message channel using PostMessage with the new provider, rather than returning a JS object (as convenient as that is).
> 
> Maybe that's better left for a future improvement. I'll try to get more info on that topic here soon.

That future browser API is [externally_connectable](https://developer.chrome.com/docs/extensions/mv3/manifest/externally_connectable/). It does seem a bit further out, but just worth noting that we should also rally around migrating towards it.

EIPS/eip-6963.md
```diff
@@ -0,0 +1,298 @@
...
+    (event: EIP6963RequestProviderEvent) => {
+      announceProvider();
+    }
+  );
+
+  announceProvider();
+}
+```
+
+### Ethereum library
```
asciiman, line 271: >>What does "Ethereum library" mean in this context? Is it talking about a wallet connection library like Web3 Onboard or Rainbowkit?

>That's correct @asciiman! This could include Web3Modal, Web3React, Web3Onboard, RainbowKit, ConnectKit, web3.js, wagmi.js, ethers.js, viem.js, etc

Maybe "Wallet Consumer" would be better here. It could be any of the libraries above, and it could also conceivably be a dApp handling the selection on it's own.

kdenhartog (dismissed):

glitch-txs (dismissed):

EIPS/eip-6963.md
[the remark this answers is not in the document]
SamWilsn: Works for me!

EIPS/eip-6963.md
SamWilsn: That's great!

EIPS/eip-6963.md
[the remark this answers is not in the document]
danfinlay: Wallets with no bull don’t use globals

EIPS/eip-6963.md
```diff
@@ -0,0 +1,230 @@
...
+  icon: string;
+}
+```
+
+The values in the `EIP6963ProviderInfo` MUST be used as follows:
+
+- walletId - globally unique identifier of the wallet provider (eg. `io.dopewallet.extension` or `awesomewallet`)
+- uuid - locally unique of the wallet provider (UUIDv4 compliant)
+- name - human-readable name of the wallet provider (e.g. `DopeWalletExtension` or `AwesomeWallet`)
+- icon - uri encoded image (RFC-3986 compliant, image must be squared with 96x96px minimum resolution)
```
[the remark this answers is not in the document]
SamWilsn: Give me a minute, apparently there is a bug in our validator bot.

EIPS/eip-6963.md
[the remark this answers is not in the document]
darrylyeo: `Javascript` → `JavaScript`

SamWilsn:
@eth-bot rerun

SamWilsn:
@eth-bot rerun

SamWilsn (dismissed):

SamWilsn (approved):

EIPS/eip-6963.md
pedrouid: Using `evm` as a prefix IMO is going to create more problems than it solves

We can all agree on EIP-1193 and EIP-6963 specifications

But it's hard to find consensus over terms like Ethereum and EVM

It's ok to use them as descriptions but I would like to avoid them in naming conventions that can cause future issues with versioning or namespacing

PR #6966, opened 2023-05-01
Update EIP-5570: Move to Last Call

darcys22:


SamWilsn (approved):
A friendly reminder that once a proposal goes to the Final status, it can no longer be edited, so please make sure to give your proposal an in-depth review before opening that pull request!

PR #6970, opened 2023-05-01
Update EIP-1185: Move to Review

mcdee:


SamWilsn (commented):
Some changes I can't comment directly on:

 - Lines 26-28 (from "Traditionally, DNS is a zone-based system..." to "... considered further for this EIP") belong in the Rationale section since they explain choices made within the proposal.
 - Please surround code snippets in headings with backticks (\`).
 - You can (and should) add links to relevant RFCs using their DOIs (like 10.17487/RFC1034). You can see an example [here](https://github.com/ethereum/EIPs/blob/ed61569182aa4de0ab72d8a70db80116ab557684/EIPS/eip-6963.md?plain=1#L59).
 - The current content of the rationale section justifies the EIP as a whole, so it should be moved to the motivation section.

xinbenlv:
Nothing block advancing to review, hence the approval.
Agree with most comments of @SamWilsn, please consider addressing them in the next status

xinbenlv (approved):

PR #6979, opened 2023-05-03
Update EIP-6493: Drop `DomainType`, `fork_version`, `genesis_hash`

etan-status:
Simplify the SSZ transaction signature scheme by removing some elements:

- `tx_type_fork_version` is not needed, as transaction signatures do not expire. On consensus messages, certain messages are constrained in validity to a single fork, which we won't need here.

- `DomainType` is not needed, as execution signatures use ECDSA and consensus signatures use BLS, so there is no harm done by reusing domain types across layers.

- `genesis_hash` is not needed, as the `chain_id` already uniquely identifies an individual chain. Malicious actors setting up custom networks can deliberately introduce both `genesis_hash` and `chain_id` conflicts with the same effort, so there is no security benefit in including both of these values.

- The EIP-2718 transaction type already encodes the type of message being signed, with 0x19 being for message signatures, and others either reserved or used for transaction signatures. This means that no extra info is needed to constrain the namespace of a signature. The EIP-2718 transaction type is sufficient for that.

- `chain_id` could be dropped from SSZ transaction payloads and always mixed into the signature, reducing transaction size.

### DIFF
--- a/EIPS/eip-6493.md
+++ b/EIPS/eip-6493.md
@@ -55,96 +55,55 @@ When representing such a transaction as part of the [execution block header's `t
 
 ### Network configuration
 
-Each SSZ transaction type is introduced to a network during a fork transition. For the new fork, the network-specific [EIP-2124](./eip-2124.md#specification) `FORK_HASH` is recorded. Furthermore, an [EIP-2718](./eip-2718.md) transaction type is assigned.
-
-### Domain types
-
-A [`DomainType`](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/phase0/beacon-chain.md#custom-types) range is defined for signing SSZ transactions.
-
-| Name | SSZ equivalent |
-| - | - |
-| [`DomainType`](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/phase0/beacon-chain.md#custom-types) | `Bytes4` |
-| `TransactionType` | `uint8` |
-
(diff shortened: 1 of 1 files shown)


etan-status:
@eth-bot rerun

g11tech:
looks good to me for 4844 sig hash computes :+1:

EIPS/eip-6493.md
```diff
@@ -55,96 +55,55 @@ When representing such a transaction as part of the [execution block header's `t
...
+        tx_type=tx_type,
+        chain_id=chain_id,
+    ).hash_tree_root())
 ```
 
 The [`Root`](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/phase0/beacon-chain.md#custom-types) to sign is computed using [`compute_signing_root`](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/phase0/beacon-chain.md#compute_signing_root) based on the unsigned transaction's `hash_tree_root` and the additional information about the transaction type.
 
+```python
+def compute_signing_root(ssz_object: SSZObject, domain: Domain) -> Root:
+    return hash_tree_root(SigningData(
```
Inphi, line 90: @etan-status Can we clarify the definition of `SigningData`?

etan-status: https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/specs/phase0/beacon-chain.md#signingdata

It's just the pre-existing type. I can copy it here for clarity as well.

PR #6982, opened 2023-05-03
Add EIP: Efficient Default Lockable Tokens

sullof:
We suggest a solution that is applicable in most scenarios, saving gas.

```solidity
pragma solidity ^0.8.9;

// ERC165 interfaceId 0x6b61a747
interface IERC6982 {
  // This event MUST be emitted upon deployment of the contract, establishing 
  // the default lock status for any tokens that will be minted in the future.
  // If the default lock status changes for any reason, this event 
  // MUST be re-emitted to update the default status for all tokens.
  // Note that emitting a new DefaultLocked event does not affect the lock 
  // status of any tokens for which a Locked event has previously been emitted.
  event DefaultLocked(bool locked);

  // This event MUST be emitted whenever the lock status of a specific token 
  // changes, effectively overriding the default lock status for this token.
  event Locked(uint256 indexed tokenId, bool locked);

  // This function returns the current default lock status for tokens.
  // It reflects the value set by the latest DefaultLocked event.
  function defaultLocked() external view returns (bool);

  // This function returns the lock status of a specific token.
  // If no Locked event has been emitted for a given tokenId, it MUST return 
  // the value that defaultLocked() returns, which represents the default 
  // lock status.
  // This function MUST revert if the token does not exist.
  function locked(uint256 tokenId) external view returns (bool);
}
```

The primary limit in existing proposal like EIP-5192 is that

1. it has 2 events for Locked and Unlocked, which is not optimal.
To make a comparison, it's like in the ERC721 instead of Transfer(from, to, id) used for mints, transfers and burns, there were Transfer(from, to, id), Mint(to, id), Burn(from, id), etc.

2. it forces you to emit an event even when the token is minted, causing a waste of gas when a token borns with a status and dies with it.

Take for example most soulbounds and non-transferable badges. They will be locked forever and it does not make sense to emit an extra event for all the tokens.


sullof:
I didn't add an implementation example because the interface is very simple.
However, I can easily add one taking code from https://github.com/cruna-cc/cruna-protocol if necessary.

sullof:
@poojaranjan I am sorry I missed the meeting (it was full night here).
How does it work for ERCs not discussed last night?

poojaranjan:
@sullof 
With the allocated time of EIP Editing Office Hour, the editor could review 14/19 PRs added to the agenda. Most likely, the rest of the proposals will be reviewed and merged before the next meeting by one of the editors. If not, we will add them for review to the next agenda.

EIPS/eip-6982.md
```diff
@@ -0,0 +1,80 @@
+---
+eip: 6982
+title: Efficient Default Lockable Tokens
+description: A gas-efficient approach to lockable ERC-721 tokens
+author: Francesco Sullo (@sullof), Alexe Spataru (@urataps)
+discussions-to: https://ethereum-magicians.org/t/erc721-default-lockable-proposal/13366
+status: Review
```
SamWilsn, line 7:
```suggestion
status: Draft
```

Unless this was previously a draft?

EIPS/eip-6982.md
```diff
@@ -0,0 +1,80 @@
...
+  // Returns the status of the token.
+  // If no special event occurred, it MUST return what defaultLocked() returns.
+  // It MUST revert if the token does not exist.
+  function locked(uint256 tokenId) external view returns (bool);
+}
+```
+
+The [ERC-165](./eip-165.md) interfaceId is `0xb45a3c0e`.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
SamWilsn, line 48: Please move this to the top of the specification section.

EIPS/eip-6982.md
```diff
@@ -0,0 +1,80 @@
...
+  function defaultLocked() external view returns (bool);
+
+  // Returns the status of the token.
+  // If no special event occurred, it MUST return what defaultLocked() returns.
+  // It MUST revert if the token does not exist.
+  function locked(uint256 tokenId) external view returns (bool);
+}
+```
+
+The [ERC-165](./eip-165.md) interfaceId is `0xb45a3c0e`.
```
SamWilsn, line 46: This interface id doesn't match the one in the comment.

assets/eip-6982/README.md
```diff
@@ -0,0 +1,14 @@
...
+To run the tests, run the following commands:
+
+```shell
+npm i -g pnpm
+pnpm i
+pnpm test
+```
+
+This implementation is used in many projects.  
+The package is on NPM at https://www.npmjs.com/package/@ndujalabs/erc721lockable
```
SamWilsn, line 14: Please remove external links.

```suggestion
```

assets/eip-6982/contracts/ERC721LockableUpgradeable.sol
```diff
@@ -0,0 +1,167 @@
...
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+import "./IERC721Lockable.sol";
+
+contract ERC721LockableUpgradeable is
+  IERC721Lockable,
+  Initializable,
+  OwnableUpgradeable,
+  ERC721Upgradeable,
+  ERC721EnumerableUpgradeable
```
SamWilsn, line 19: Is the upgradeable part necessary for this reference implementation? We generally prefer minimal implementations for simplicity.

assets/eip-6982/hardhat.config.js
SamWilsn, line 1: Please remove build system files.

assets/eip-6982/pnpm-lock.yaml
SamWilsn, line 1: Definitely remove this one.

assets/eip-6982/README.md
sullof: Done

EIPS/eip-6982.md
sullof: Done.

EIPS/eip-6982.md
sullof: I made the change in latest push

EIPS/eip-6982.md
```diff
@@ -0,0 +1,80 @@
...
+  // It MUST revert if the token does not exist.
+  function locked(uint256 tokenId) external view returns (bool);
+}
+```
+
+The [ERC-165](./eip-165.md) interfaceId is `0x6b61a747`.
+
+## Rationale
+
+This standard aims to minimize gas usage by reducing the emission of events. The **DefaultLocked** event sets the initial lock status for all tokens, thus eliminating the need for an event upon the minting of each token. Some tokens may change their behavior (for example, after a reveal), in that case, the **DefaultLocked** event may be emitted again to set up the new default behavior when no token-specific event has been emitted yet.
```
SamWilsn, line 52: Looks like you're defining a requirement here:

> the **DefaultLocked** event may

Requirements should go in the specification section.

Plus I believe this is directly contradicting the comment above the `DefaultLocked` event definition.

sullof: Thanks, I will move that requirements  it the specification section. 

About the apparent contradiction, there are tokens that are locked during the pre-reveal phase and so would emit a DefaultLocked(true) event, that will be unlocked after the reveal. In that case, forcing the contract to emit a Locked(tokenId, false) for every tokenId would go against the goal behind this proposal, saving gas when a locked/unlocked status is applicable to all the tokens.

PR #6985, opened 2023-05-03
Update EIP-4844: de-sszify spec

lightclient:
I'd like to propose removing SSZ entirely from the EL side of 4844.

The only real value in blob transactions being SSZ encoded (as specified) is that the CL is able to parse the transaction and easily access `blob_versioned_hashes`. I don't have a great solution to this if they aren't SSZ encoded. Ultimately it seems that either i) one of the layers implements the other layer's serialization scheme or ii) the engine api is extended to resolve blob hashes given a blob tx (may not even be efficient enough).

I think adding RLP into the CL requires much less decision making than supporting SSZ in the EL. It's already clear how to lay transactions out, calculate their sig hashes and standard hashes, etc. CLs simply need to search through the opaque blobs, find the blob txs, then RLP decode them.

If we do end up going forward with SSZ in 4844, it is almost a certainty that either i) we have to change something about it in the future (defeating the purpose a bit) or ii) we are unnecessarily constrained when we do decide make the full transition to SSZ.

The least risky option appears to be making 4844 transactions RLP. I would love if we were to spend time and figure out how to apply SSZ properly (à la EIP-6404), but it doesn't seem there is interest in that route at the moment. So given all that, I believe RLP is the best path forward.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -274,12 +274,13 @@ rlp([blob_tx_payload, blob_kzgs, blobs, kzg_aggregated_proof])
 
 Each of these elements are defined as follows:
 
-* `blob_tx_payload` - standard EIP-2718 blob transaction `TransactionPayload`
-* `blob_kzgs` - list of `KZGCommitment`
-* `blobs` - list of `blob` where `blob` is a list of `BLSFieldElement`
-* `kzg_aggregated_proof` - `KZGProof`
+- `blob_tx_payload` - standard EIP-2718 blob transaction `TransactionPayload`
+- `blob_kzgs` - list of `KZGCommitment`
+- `blobs` - list of `blob` where `blob` is a list of `BLSFieldElement`
+- `kzg_aggregated_proof` - `KZGProof`
 
 The node MUST validate `blob_tx_payload` and verify the wrapped data against it. To do so, ensure that:
+
 - `blob_tx_payload.blob_versioned_hashes` must not be empty
 - All hashes in `blob_tx_payload.blob_versioned_hashes` must start with the byte `BLOB_COMMITMENT_VERSION_KZG`
 - There must be at most `MAX_DATA_GAS_PER_BLOCK // DATA_GAS_PER_BLOB` total blob commitments in the transaction.
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -269,7 +269,7 @@ The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sen
 Blob transactions have two network representations. During transaction gossip responses (`PooledTransactions`), the EIP-2718 `TransactionPayload` of the blob transaction is wrapped to become:
 
 ```
-rlp([blob_tx_payload, blob_kzgs, blobs, kzg_aggregated_proof])
+rlp([blob_tx_payload, blob_kzgs, blobs, blob_kzg_proofs])
 ```
 
 Each of these elements are defined as follows:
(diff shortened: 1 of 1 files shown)


EIPS/eip-4844.md
```diff
@@ -101,93 +101,28 @@ def fake_exponential(factor: int, numerator: int, denominator: int) -> int:
...
+#### Signature
 
-```python
-def signed_tx_hash(tx: SignedBlobTransaction) -> Bytes32:
-    return keccak256(BLOB_TX_TYPE + ssz.serialize(tx))
-```
+The signature values `y_parity`, `r`, and `s` are calculated by constructing a secp256k1 signature over the following digest:
 
-Blob transactions with empty `blob_versioned_hashes` are also considered invalid during execution block verification, and must not be included in a valid block.
+`keccak256(BLOB_TX_TYPE || rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, amount, data, blob_versioned_hashes, access_list]))`.
```
hwwhww, line 121: Any reason why `max_fee_per_data_gas` and `blob_versioned_hashes` are not appended at the end?

lightclient: Really no reason to not append at end, not sure why I didn't add it originally there. Also was missing the the `max_fee_per_data_gas` field, so I added that back.

g11tech (approved):
looks good to me

EIPS/eip-4844.md
```diff
@@ -331,41 +266,33 @@ The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sen
...
-class BlobTransactionNetworkWrapper(Container):
-    tx: SignedBlobTransaction
-    # KZGCommitment = Bytes48
-    blob_kzgs: List[KZGCommitment, MAX_TX_WRAP_KZG_COMMITMENTS]
-    # BLSFieldElement = uint256
-    blobs: List[Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB], LIMIT_BLOBS_PER_TX]
-    # KZGProof = Bytes48
-    kzg_aggregated_proof: KZGProof
+```
+rlp([blob_tx_payload, blob_kzgs, blobs, kzg_aggregated_proof])
```
dankrad, line 272:
```suggestion
rlp([blob_tx_payload, blob_kzgs, blobs, blob_kzg_proofs])
```
We are now using one KZG proof per blob (no more aggregated proofs)

EIPS/eip-4844.md
```diff
@@ -331,41 +266,33 @@ The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sen
...
+
+The node MUST validate `blob_tx_payload` and verify the wrapped data against it. To do so, ensure that:
+
+- `blob_tx_payload.blob_versioned_hashes` must not be empty
+- All hashes in `blob_tx_payload.blob_versioned_hashes` must start with the byte `BLOB_COMMITMENT_VERSION_KZG`
+- There must be at most `MAX_DATA_GAS_PER_BLOCK // DATA_GAS_PER_BLOB` total blob commitments in the transaction.
+- There are an equal number of versioned hashes, kzg commitments, and blobs.
+- The KZG commitments hash to the versioned hashes, i.e. `kzg_to_versioned_hash(kzg[i]) == versioned_hash[i]`
+- The KZG commitments match the blob contents. (Note: this can be optimized with additional data, using a proof for a
+  random evaluation at two points derived from the commitment and blob data)
```
dankrad, line 290:
```suggestion
- The KZG commitments match the blob contents. (Note: this can be optimized using `blob_kzg_proofs`, using a proof for a
  random evaluation at a point derived from the commitment and blob data for each blob)
```

EIPS/eip-4844.md
```diff
@@ -331,41 +266,33 @@ The actual `data_fee` as calculated via `calc_data_fee` is deducted from the sen
...
+
+For body retreival responses (`BlockBodies`), the standard EIP-2718 blob transaction `TransactionPayload` is used.
```
dankrad, line 292:
```suggestion
For body retrieval responses (`BlockBodies`), the standard EIP-2718 blob transaction `TransactionPayload` is used.
```

dankrad (changes requested):

dankrad (dismissed):

lightclient:
@dankrad previously approved, but I didn't realize I needed to rebase. I will go ahead with merge since the rebase was rather straightforward.

PR #6991, opened 2023-05-05
Update EIP-6953: Move to Review

timbeiko:
Status change


SamWilsn (commented):
I'd be careful about this EIP. If it goes to final, you won't be able to update it any further, so it won't be an "exhaustive" list.

If you'd like living status, you'd want to bring it up at an EIPIP meeting.

Generally, we try to avoid EIPs that are tables of things because they don't fit well into our immutability flow.

g11tech:
In the Abstract, we mentioned the current network upgrades, may be we can forgo that and specify a link to here: https://github.com/ethereum/execution-specs/tree/master/network-upgrades shouldn't be a problem since the link should be allowed

timbeiko:
Replaced by https://github.com/ethereum/EIPs/pull/7073

PR #6992, opened 2023-05-05
Add EIP: ECC DSS signatures

JKincorperated:
As the transition from web2 to web3 some API's and services are not compatible with smart contract. This EIP allows for smart contracts to make HTTP requests without "making the worlds larges DDoS machine"


sullof:
@JKincorperated  
You can fix the failing check adding 
```
eip: 6992
```
as first premble.

JKincorperated:
> @JKincorperated You can fix the failing check adding
> 
> ```
> eip: 6992
> ```
> 
> as first premble.

Thanks, should be fixed now.

Pandapip1 (changes requested):
Why not just take advantage of the fact that servers sign the body with their certificate when using HTTPS?

JKincorperated:
> Why not just take advantage of the fact that servers sign the body with their certificate when using HTTPS?

It is not possible. TLS is engineered is for encryption not signing and uses symmetric cryptography, verifying the certificate might overload smaller web servers whereas DNS servers can handle the request easily, it would be unnecessary bloat to implement a custom system to make web requests instead of using the many modules available and webservers don't even sign the body parameter.

Pandapip1:
> It is not possible. TLS is engineered is for encryption not signing and uses symmetric cryptography, verifying the certificate might overload smaller web servers whereas DNS servers can handle the request easily, it would be unnecessary bloat to implement a custom system to make web requests instead of using the many modules available and webservers don't even sign the body parameter.

I'm afraid you're wrong here. Servers using TLS **do** provide a cryptographic signature. Here's an implementation: https://old.tlsnotary.org/

JKincorperated:
> 
> I'm afraid you're wrong here. Servers using TLS **do** provide a cryptographic signature. Here's an implementation: https://old.tlsnotary.org/

I never knew that someone has implemented this, however it would still be easier to implement this as it only uses common algorithms that should already have widespread coverage across languages. Also, using DNS would stop all of the nodes basically DDoSing that single web server for verification.

JKincorperated:
> It's easy to avoid this DDoS effect. Simply add data to proposed blocks that include HTTP request and response data.

That's the issue, to verify that that domain has responded every node must have the key and no one client could **ever** send it with the request to maintain security.

lightclient (changes requested):
It's not possible to validate this regardless because you can get different responses for the http request. It would be trivial to split the network this way.

JKincorperated:
> It's not possible to validate this regardless because you can get different responses for the http request. It would be trivial to split the network this way.

The request is passed via call data, along with a signature of the body which has been signed by a key hosted via DNS. The data inside the request can vary, however the signature must remain and be verifiable via querying the key at the domain the request originated from. This should not split the network as the request is passed through calldata.

Pandapip1:
I think this would be better as an ERC or precompile. Probably the former.

JKincorperated:
> I think this would be better as an ERC or precompile. Probably the former.

Would changing it to an ERC effect how the opcode is implemented?

lightclient:
@JKincorperated you say
> the server must make a DNS query to a TXT located in the domain field at the subdomain

it isn't clear if you mean the node here during the operation of the opcode - if so, then the dns server may send different reponses to the TXT query how ever it feels.

JKincorperated:
I see what you mean now, I have adjusted it so the opcode just verifies an ECC DSS signature.

lightclient:
Now that this is a pure cryptographic function, it should be defined as a precompile. We don't really implement cryptographic primitives as opcodes. Check out the format from EIP-152 for inspiration: https://eips.ethereum.org/EIPS/eip-152

JKincorperated:
I just finished changing the specs so it is a precompile, as before I started this I had no idea what a precompiled contract was, after 30 minutes of intense googling, I have adjusted the EIP accordingly. 

Also the EIP Walidator is being annoying and does not like my Ethereum magicians link.

EIPS/eip-6992.md
```diff
@@ -0,0 +1,89 @@
...
+- PublicKey (string): An ECC P-256 public key in PEM format
+
+Output:
+
+- Verified (bool): True if the data is correctly signed.
+
+Description:
+This contract must encode the data to bytes using utf-8 and then must hash the Body field with the SHA256. Then, the contract should verify against the ECC P-256 public key which should be used to verify the signature against the data using the fips-186-3 DSS signing algorithm.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
SamWilsn, line 40: Please move this paragraph to the top of the specification section.

EIPS/eip-6992.md
```diff
@@ -0,0 +1,89 @@
...
+
+TBD
+
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Test Cases
+
+Opcode : 0xa5
```
SamWilsn, line 52: Is this proposal creating a precompile or an opcode?

EIPS/eip-6992.md
```diff
@@ -0,0 +1,89 @@
...
+discussions-to: https://ethereum-magicians.org/t/draft-eip-6992-ecc-dss-signatures/14153
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-04-25
+---
+
+## Abstract
+
+This EIP allows for smart contracts to verify DSS signatures.
```
SamWilsn, line 15: Is this not possible today with `ecrecover`?

JKincorperated:
When I saw "ecrecover" I misread and did not think is was for verifying ECC signatures. As the EIP does exactly that, I am closing it.

PR #7001, opened 2023-05-09
Update EIP-4844: change op to `BLOBHASH`

lightclient:
Right now `DATAHASH` sort of conflicts with the EOF data section concept (which will want to have things like `DATALOAD`, `DATACOPY`, etc for). So I propose we change the mnemonic for blob hashes to `BLOBHASH` instead.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -181,7 +181,7 @@ The `ethereum/consensus-specs` repository defines the following beacon-node chan
 
 ### Opcode to get versioned hashes
 
-We add an opcode `BLOBHASH` (with byte value `HASH_OPCODE_BYTE`) which reads `index` from the top of the stack
+We add an instruction `BLOBHASH` (with opcode `HASH_OPCODE_BYTE`) which reads `index` from the top of the stack
 as big-endian `uint256`, and replaces it on the stack with `tx.blob_versioned_hashes[index]`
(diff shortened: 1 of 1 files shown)


EIPS/eip-4844.md
```diff
@@ -246,7 +246,7 @@ The `ethereum/consensus-specs` repository defines the following beacon-node chan
 
 ### Opcode to get versioned hashes
 
-We add an opcode `DATAHASH` (with byte value `HASH_OPCODE_BYTE`) which reads `index` from the top of the stack
+We add an opcode `BLOBHASH` (with byte value `HASH_OPCODE_BYTE`) which reads `index` from the top of the stack
```
chfast, line 249:
```suggestion
We add an instruction `BLOBHASH` (with opcode `HASH_OPCODE_BYTE`) which reads `index` from the top of the stack
```

chfast: Some other name suggestions were:

- `TXBLOBHASH`
- `TXDATAHASH`

Inphi:
Makes sense as we already call the hashes in the blob tx `blob_versioned_hashes`. Though I don't think we need to replace "data" with "blob" everywhere else. The context is clear enough.

g11tech (approved):
looks good to me

dankrad (dismissed):

lightclient:
@dankrad also already approved this but the review was dismissed due to a rebase. The change is still the exact same, so I will go ahead and merge.

PR #7006, opened 2023-05-10
Update EIP-3026: Move to Draft

hujw77:
## Gas estimation repo
https://github.com/hujw77/eip-test-gen/tree/main/eip-3026
## Gas estimation strategy
Gas cost is derived by taking the average timing of the same operations over different implementations and assuming a constant `51 MGas/second`. Since the execution time is machine-specific, this constant is determined based on execution times of [ECRECOVER](https://github.com/hujw77/eip-test-gen/blob/main/eip-3026/benches/ec_recover.rs) and [BNPAIR](https://github.com/hujw77/eip-test-gen/blob/main/eip-3026/benches/bn_pairing.rs) precompiles on my machine and their proposed gas price (`63.8 MGas/s` for ECRECOVER and `37.8 MGas/s` for BNPAIR). Following are the proposed methods to time the precompile operations:

- G1 addition: Average timing of 1000 random samples.
- G1 multiplication: Average timing of 1000 samples of random worst-case of double-and-add algorithm (scalar of max bit length and max hamming weight and random base points in G1)
- G2 addition: Average timing of 1000 random samples
- G2 multiplication: Average timing of 1000 samples of radnom worst-case of double-and-add algorithm (scalar of max bit length and max hamming weight and random base points in G2)
- G1 and G2 multiexponentiations: Expected to be performed by the Peppinger algorithm, with a table prepared for discount in case of `k <= 128` points in the multiexponentiation with a discount cup `max_discount` for `k > 128`. To avoid non-integer arithmetic call cost is calculated as `k * multiplication_cost * discount / multiplier` where `multiplier = 1000`, `k` is a number of (scalar, point) pairs for the call, `multiplication_cost` is a corresponding single multiplication call cost for G1/G2.
- Pairing: Average timing of 1000 random samples (random points in G1 and G2) for different number of pairs with linear lifting.
## Gas estimation result
https://github.com/hujw77/eip-test-gen/blob/main/eip-3026/benches/BENCHMARKS.md
https://github.com/hujw77/eip-test-gen/blob/main/eip-3026/gas-meter/benches.md


EIPS/eip-3026.md
```diff
@@ -1,7 +1,7 @@
 ---
 eip: 3026
 title: BW6-761 curve operations
-author: Youssef El Housni (@yelhousni), Michael Connor (@iAmMichaelConnor), Aurore Guillevic <aurore.guillevic@inria.fr>
+author: Youssef El Housni (@yelhousni), Michael Connor (@iAmMichaelConnor), Aurore Guillevic <aurore.guillevic@inria.fr>, hujw77 <hujw77@gmail.com>
```
lightclient, line 4:
```suggestion
author: Youssef El Housni (@yelhousni), Michael Connor (@iAmMichaelConnor), Aurore Guillevic <aurore.guillevic@inria.fr>, hujw77 (@hujw77)
```
you probably want this as it will allow you to merge changes to the EIP w/o waiting for approval from the other authors.

EIPS/eip-3026.md
```diff
@@ -1,7 +1,7 @@
...
 discussions-to: https://ethereum-magicians.org/t/eip-3026-bw6-761-curve-operations/4790
 status: Stagnant
```
lightclient, line 6: this needs to be set to the last non-stagnant value.

iAmMichaelConnor:
> 



> @hujw77 you're adding yourself as an author to an EIP, can you please have one of the current authors approve this PR so we know it is acceptable to authorize you to make changes to the EIP?

No objection from me! This has fallen far down my priority list since we began the EIP, so it's nice to see someone else take it up. @yelhousni any objections?

EIPS/eip-3026.md
hujw77: Yep, I want to become one of the author in EIP-3026.

PR #7008, opened 2023-05-10
Update EIP-4844: amend inconsistency in calc_data_fee

yperbasis:
The second argument of `calc_data_fee` is `parent` rather than `header`.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -291,7 +291,7 @@ We use the `excess_data_gas` header field to store persistent data needed to com
 
 ```python
 def calc_data_fee(tx: SignedBlobTransaction, parent: Header) -> int:
-    return get_total_data_gas(tx) * get_data_gasprice(header)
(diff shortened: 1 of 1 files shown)


Inphi (approved):

EIPS/eip-4844.md
```diff
@@ -291,7 +291,7 @@ We use the `excess_data_gas` header field to store persistent data needed to com
 
 ```python
 def calc_data_fee(tx: SignedBlobTransaction, parent: Header) -> int:
-    return get_total_data_gas(tx) * get_data_gasprice(header)
+    return get_total_data_gas(tx) * get_data_gasprice(parent)
```
roberto-bayardo, line 294: I don't think this is correct. We compute the fee from the current block's "excess data gas" value, not the parent's.  

I think calc_data_gas here should be updated to take in "excess data gas" rather than the header of a block that doesn't exist yet.

geenutts (approved):

PR #7021, opened 2023-05-12
Update EIP-5507: Move to Final

Pandapip1:


SamWilsn (approved):
I'd like to register my misgivings for this EIP one last time:

> I don’t believe this proposal is effective:
>
>  1.  Tokens can pretend to be refundable, but not actually allow refunds.
>  1.  If refunds only belong to the first buyer, then it encourages reselling, and provides little  benefit to “actual users”.
> 1.    An escrow contract could accomplish the same thing, for any transaction, not just for the first buyers. This would work for older NFTs too!

Pandapip1:
@SamWilsn just a reminder - this doesn't apply to only 'first buyers'. It's configurable.

PR #7030, opened 2023-05-15
Update EIP-107: Use standardized markdown formatting

Pandapip1:
I was wondering why my vitepress builds were failing...

### DIFF
--- a/EIPS/eip-107.md
+++ b/EIPS/eip-107.md
@@ -8,14 +8,14 @@ type: Standards Track
 category: Interface
 ---
 
-Abstract
-========
+## Abstract
+
 This draft EIP describes the details of an authorization method that if provided by rpc enabled ethereum nodes would allow regular websites to send transactions (via ```eth_sendTransaction```) without the need to enable CORS. Instead, user would be asked to confirm the transaction via an html popup.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-107.md
```diff
@@ -90,22 +90,22 @@ the error object cannot be a javascript Error object due to postMessage limitati
...
 The current design, instead has a very simple implementation (self contained html file that can be shared across node's implementation) and its safeness is guarantess by browsers' cross domain policies.
 
 The use of iframe/ window was required to have both security and user friendliness. The invisible iframe allows the dapp to execute read only calls without the need for user input, and the window ensures user approval before making a call. While we could have made it without the window mode by making the iframe confirmation use the native browser ```window.confirm``` dialog, this would have prevented the use of a more elegant confirmation popup that the current design allows. It also happens to be that the ```window.confirm``` is not safe in some browsers, as it gives focus to the accept option and can be triggered automatically (https://bugs.chromium.org/p/chromium/issues/detail?id=260653).
 
 
-Implementations
-===============
+### Implementations
```
SamWilsn, line 101:
```suggestion
## Implementations
```

SamWilsn (approved):

PR #7039, opened 2023-05-17
Add EIP: Scheme-Handler Discovery Option for Wallets

SamWilsn:


EIPS/eip-7039.md
```diff
@@ -0,0 +1,103 @@
...
+category: Interface
+created: 2023-05-15
+requires: 1193
+---
+
+## Abstract
+
+This proposal (affectionately known as SHADOW) is an alternative to [EIP-1193](./eip-1193.md) for wallet discovery in web browsers that requires no special permissions. Web pages intending to open a connection to a wallet inject an `iframe` tag pointing at a well-known scheme. Communication between the page and the wallet uses the `postMessage` API.
+
+## Motivation
```
jxom, line 18: I may be missing something, but can’t we utilise 6963 to solve the motivation for this? I mean, if a user has already requested perms for a wallet, can’t we just switch the current active provider from the built up `providers` array?

jxom: We can even request perms by switching the active provider too*

SamWilsn: I've updated the motivation to try and distinguish this proposal from 6963. Using a protocol handler reduces the required extension permissions significantly, and lets non-extension wallets connect.

lightclient (approved):
LGTM

PR #7044, opened 2023-05-18
Add EIP: Perpetually Valid Signed Voluntary Exits

dapplion:
todo:

- [x] assign number

### DIFF
--- a/EIPS/eip-7044.md
+++ b/EIPS/eip-7044.md
@@ -3,7 +3,7 @@ eip: 7044
 title: Perpetually Valid Signed Voluntary Exits
 description: Lock voluntary exit signature domain on capella for perpetual validity
 author: Lion (@dapplion)
-discussions-to: https://github.com/ethereum/consensus-specs/pull/3288
+discussions-to: https://ethereum-magicians.org/t/eip-7044-perpetually-valid-signed-voluntary-exits/14348
 status: Draft
 type: Standards Track
 category: Core
@@ -16,9 +16,9 @@ Lock validator voluntary exit signature domain on Capella for perpetual validity
 
 ## Motivation
 
-Currently, signed voluntary exits are valid for only two forks, due to the Beacon Chain state holding only the current and previous fork. This limitation affects increases the complexity of some staking operations. Specifically those that require the exchange of signed exits ahead of time valid for an unbounded number of forks.
+Currently, signed voluntary exits are valid up-to only two forks for block inclusion due to the Beacon Chain state considering only the current and previous fork. This limitation increases the complexity of some staking operations, specifically those that require the exchange of signed exits ahead of time for an unbounded number of forks.
 
-The limited validity of volutary exits was originally motivated to isolate them in the event of a perpetual fork. If fork A and B exist and a validator operators on both, if they send an exit, it will be replayable on both. However, this positive is not sufficient to justify the UX degradation exposed above.
+The limited validity of voluntary exits was originally motivated to isolate them in the event of a perpetual fork. If fork A and B exist and a validator operates on both, if they send an exit, it will be replayable on both. However, this possibility is not sufficient to justify the UX degradation exposed above.
 
 ## Specification
 
@@ -50,7 +50,7 @@ Test cases are work-in-progress within the standard Consensus Layer tests.
 
 ## Security Considerations
 
-The divergent signature domains across forked networks would previously have prevented the replay of VoluntaryExits after two hard forks. This specification change causes this replay protection to no longer exist. These potential replays could impact individual stakers on both sides of a fork, but does not impact the security of the chain.
+The divergent signature domains across forked networks would previously have prevented the replay of VoluntaryExits after two hard forks. This specification change causes the replay protection to no longer exist. These potential replays could impact individual stakers on both sides of a fork, but does not impact the security of the chain.
 
 ## Copyright
 
--- a/EIPS/eip-7044.md
+++ b/EIPS/eip-7044.md
@@ -24,7 +24,13 @@ The limited validity of voluntary exits was originally motivated to isolate them
 
 ### Consensus Layer
 
-The mechanics of the specification can be found in the [Consensus Layer specs](https://github.com/ethereum/consensus-specs/commit/14212958d3605c5dd4f8ab617f157328f53ce559).
+Specification changes are built into the Consensus Specs Deneb upgrade.
+
+The specific makes one change to the state transition function:
+
+- Modify [`process_voluntary_exit`](https://github.com/ethereum/consensus-specs/blob/75971a8c218b1d76d605dd8b88a08d39c42de221/specs/deneb/beacon-chain.md#modified-process_voluntary_exit) to compute the signing domain and root fixed on `CAPELLA_FORK_VERSION`.
+
+Additionally, the `voluntary_exit` gossip conditions are implicitly modified to support this change.
 
 To make the change backwards compatible the signature domain is locked on the Capella fork
 
--- a/EIPS/eip-7044.md
+++ b/EIPS/eip-7044.md
@@ -18,7 +18,7 @@ Lock validator voluntary exit signature domain on Capella for perpetual validity
 
 Currently, signed voluntary exits are valid up-to only two upgrades for block inclusion due to the Beacon Chain state considering only the current and previous fork version. This limitation increases the complexity of some staking operations, specifically those in which the staking operator (holder of active key) is distinct from the owner of the funds (holder of the withdrawal credential). Because voluntary exits can only be signed by the active key, such a relationship requires the exchange of signed exits ahead of time for an unbounded number of forks.
 
-The limited validity of voluntary exits was originally motivated to isolate them in the event of a perpetual fork. If fork A and B exist and a validator operates on both, if they send an exit, it will be replayable on both. However, this possibility is not sufficient to justify the UX degradation exposed above.
+The limited validity of voluntary exits was originally motivated to isolate them in the event of a hard fork that results in two maintained chains. If fork A and B exist and a validator operates on both, if they send an exit, it will be replayable on both. However, this possibility is not sufficient to justify the UX degradation exposed above, as no funds are at risk and the staker can re-stake on one or both of the chains.
 
 ## Specification
 
--- a/EIPS/eip-7044.md
+++ b/EIPS/eip-7044.md
@@ -34,7 +34,7 @@ This specification does not require any changes to the Execution Layer.
 
 ## Rationale
 
-Perpetually valid signed exits allow simpler staking operation designs. It also aligns the UX of such objects to BLSToExecutionChanges and deposits, such that downstream tooling do not need to be updated with fork future information.
+Perpetually valid signed voluntary exits allow simpler staking operation designs. It also aligns the UX of such objects to `BLSToExecutionChanges` and deposits, such that downstream tooling does not need to be updated with fork version information.
 
 ## Backwards Compatibility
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
+---
+eip: 7044
+title: Perpetually Valid Signed Voluntary Exits
+description: Lock voluntary exit signature domain on capella for perpetual validity
+author: Lion (@dapplion)
+discussions-to: https://github.com/ethereum/consensus-specs/pull/3288
```
g11tech, line 6: you might need to open a discussion or ethereum-magicians.org or https://ethresear.ch

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+created: 2023-05-18
+---
+
+## Abstract
+
+Lock validator voluntary exit signature domain on Capella for perpetual validity. Currently, signed voluntary exits are only valid for two forks.
+
+## Motivation
+
+Currently, signed voluntary exits are valid for only two forks, due to the Beacon Chain state holding only the current and previous fork. This limitation affects increases the complexity of some staking operations. Specifically those that require the exchange of signed exits ahead of time valid for an unbounded number of forks.
```
g11tech, line 19:
```suggestion
Currently, signed voluntary exits are valid up-to only two forks for block inclusion due to the Beacon Chain state considering only the current and previous fork. This limitation increases the complexity of some staking operations, specifically those that require the exchange of signed exits ahead of time for an unbounded number of forks.
```

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+The limited validity of volutary exits was originally motivated to isolate them in the event of a perpetual fork. If fork A and B exist and a validator operators on both, if they send an exit, it will be replayable on both. However, this positive is not sufficient to justify the UX degradation exposed above.
```
g11tech, line 21:
```suggestion
The limited validity of voluntary exits was originally motivated to isolate them in the event of a perpetual fork. If fork A and B exist and a validator operates on both, if they send an exit, it will be replayable on both. However, this possibility is not sufficient to justify the UX degradation exposed above.
```

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Specification
+
+### Consensus Layer
+
+The mechanics of the specification can be found in the [Consensus Layer specs](https://github.com/ethereum/consensus-specs/pull/3288).
```
g11tech, line 27: you will need to refer to a `blob` commit of the relevant content rather than a PR

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+There are no forwards/backwards compatibility issues with the Execution Layer
+
+## Test Cases
+
+Test cases are work-in-progress within the standard Consensus Layer tests.
+
+## Security Considerations
+
+The divergent signature domains across forked networks would previously have prevented the replay of VoluntaryExits after two hard forks. This specification change causes this replay protection to no longer exist. These potential replays could impact individual stakers on both sides of a fork, but does not impact the security of the chain.
```
g11tech, line 53:
```suggestion
The divergent signature domains across forked networks would previously have prevented the replay of VoluntaryExits after two hard forks. This specification change causes the replay protection to no longer exist. These potential replays could impact individual stakers on both sides of a fork, but does not impact the security of the chain.
```

EIPS/eip-7044.md
abcoathup: You need to create a discussion thread on ethereum-magicians.org
See Danny's as an example:
https://ethereum-magicians.org/t/eip-7045-increase-max-attestation-inclusion-slot/14342

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+The limited validity of voluntary exits was originally motivated to isolate them in the event of a perpetual fork. If fork A and B exist and a validator operates on both, if they send an exit, it will be replayable on both. However, this possibility is not sufficient to justify the UX degradation exposed above.
+
+## Specification
+
+### Consensus Layer
+
+The mechanics of the specification can be found in the [Consensus Layer specs](https://github.com/ethereum/consensus-specs/pull/3288).
+
+To make the change backwards compatible the signature domain is locked on the capella fork
```
abcoathup, line 29:
```suggestion
To make the change backwards compatible the signature domain is locked on the Capella fork
```

zah:
I suspect this EIP is trying to address a common problem faced by staking pool operators: How can their customers be sure that the operator won't keep the created validators active forever (this allows the operator to profit from transaction fees and MEV)?

To solve this problem in a more comprehensive way, it may be beneficial to also introduce a new type of exit message that identifies the exited validator by their public key. This is an improvement because it will allow the staking pool to share the signed exit message with the staker before the staker has committed to fund the created validator. This is not easy to do otherwise, because the final validator index is now known until the deposit has been processed by the beacon chain. 

I've written more about this here:
https://github.com/ethereum/consensus-specs/issues/2666

One possible work-around with significant downsides is to require the operator to sign a very wide range of exit messages that target all possible validator indices.

EIPS/eip-7044.md
g11tech: @dapplion we are modifying the regex to support the `compare` diffs in drafts instead of PRs as the diffs are static: so basically this for your PR: `https://github.com/ethereum/consensus-specs/compare/53a9324cf0afd90ea06c5ebb0bfef8fc95a01fd1...1c35eb1c3303fe1e0b101323106d766d8f848cd6` which would is [this link](https://github.com/ethereum/consensus-specs/compare/53a9324cf0afd90ea06c5ebb0bfef8fc95a01fd1...1c35eb1c3303fe1e0b101323106d766d8f848cd6) which seems that the entire PR diff would need to be squashed in a commit for diff to be genrated correctly

g11tech:
@eth-bot rerun

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-7044-perpetually-valid-signed-voluntary-exits/14348
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-05-18
+---
+
+## Abstract
+
+Lock validator voluntary exit signature domain on Capella for perpetual validity. Currently, signed voluntary exits are only valid for two forks.
```
djrtwo, line 15: I think "upgrades" or "hardforks" is more specific and valuable here. as "forks" can mean manythings including short term fork-choice disagreements

```suggestion
Lock validator voluntary exit signature domain on Capella for perpetual validity. Currently, signed voluntary exits are only valid for two upgrades.
```

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Motivation
+
+Currently, signed voluntary exits are valid up-to only two forks for block inclusion due to the Beacon Chain state considering only the current and previous fork. This limitation increases the complexity of some staking operations, specifically those that require the exchange of signed exits ahead of time for an unbounded number of forks.
```
djrtwo, line 19: ditto

```suggestion
Currently, signed voluntary exits are valid up-to only two upgrades for block inclusion due to the Beacon Chain state considering only the current and previous fork version. This limitation increases the complexity of some staking operations, specifically those in which the staking operator (holder of active key) is distinct from the owner of the funds (holder of the withdrawal credential). Because voluntary exits can only be signed by the active key, such a relationship requires the exchange of signed exits ahead of time for an unbounded number of forks.
```

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+The limited validity of voluntary exits was originally motivated to isolate them in the event of a perpetual fork. If fork A and B exist and a validator operates on both, if they send an exit, it will be replayable on both. However, this possibility is not sufficient to justify the UX degradation exposed above.
```
djrtwo, line 21:
```suggestion
The limited validity of voluntary exits was originally motivated to isolate them in the event of a hard fork that results in two maintained chains. If fork A and B exist and a validator operates on both, if they send an exit, it will be replayable on both. However, this possibility is not sufficient to justify the UX degradation exposed above, as no funds are at risk and the staker can re-stake on one or both of the chains.
```

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Specification
+
+### Consensus Layer
+
+The mechanics of the specification can be found in the [Consensus Layer specs](https://github.com/ethereum/consensus-specs/commit/14212958d3605c5dd4f8ab617f157328f53ce559).
```
djrtwo, line 27: Specification changes are built into the Consensus Specs Deneb upgrade.

The specific makes one change to the state transition function:
* Modify `process_voluntary_exit`](https://github.com/ethereum/consensus-specs/blob/75971a8c218b1d76d605dd8b88a08d39c42de221/specs/deneb/beacon-chain.md#modified-process_voluntary_exit) to compute the signing domain and root fixed on `CAPELLA_FORK_VERSION`.

Additionally, the `voluntary_exit` gossip conditions are implicitly modified to support this change.

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+To make the change backwards compatible the signature domain is locked on the Capella fork
+
+### Execution Layer
+
+This specification does not require any changes to the Execution Layer.
+
+## Rationale
+
+Perpetually valid signed exits allow simpler staking operation designs. It also aligns the UX of such objects to BLSToExecutionChanges and deposits, such that downstream tooling do not need to be updated with fork future information.
```
djrtwo, line 37:
```suggestion
Perpetually valid signed voluntary exits allow simpler staking operation designs. It also aligns the UX of such objects to `BLSToExecutionChanges` and deposits, such that downstream tooling does not need to be updated with fork version information.
```

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Backwards Compatibility
+
+This change is backwards compatible to the Consensus Layer of Ethereum block processing logic.
+
+The expectation of future validity of exits is not forward compatible. Specifically, users who have already pre-signed exits utilizing the Deneb fork domain with an expectation of their validity should be aware that these pre-signed exits will no longer be recognized as valid. Consequently, users should adjust their approach moving forward. For continued validity across forks, including Deneb and subsequent forks, users should ensure that their exits are signed using the Capella fork domain.
+
+There are no forwards/backwards compatibility issues with the Execution Layer
```
djrtwo, line 45:
```suggestion
There are no forwards/backwards compatibility issues with the Execution Layer.
```

EIPS/eip-7044.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Test Cases
+
+Test cases are work-in-progress within the standard Consensus Layer tests.
+
+## Security Considerations
+
+The divergent signature domains across forked networks would previously have prevented the replay of VoluntaryExits after two hard forks. This specification change causes the replay protection to no longer exist. These potential replays could impact individual stakers on both sides of a fork, but does not impact the security of the chain.
```
djrtwo, line 53:
```suggestion
The divergent signature domains across forked networks would previously have prevented the replay of VoluntaryExits after two hard forks. This specification change causes the replay protection to no longer exist. These potential replays could impact individual stakers on both sides of a fork, but does not put funds at risk and does not impact the security of the chain.
```

djrtwo (commented):
some clean-ups in language and links to get this thing mergeddd

g11tech:
rebased via UI  to check the link validation, but it still fails cc @SamWilsn

g11tech (approved):
lgtm

PR #7047, opened 2023-05-19
Update EIP-1: Modernize section list and requirements

Pandapip1:
The section list was rather dense and hard to read. It was also particularly focused on Core EIPs, as evidenced by the wording not being significantly changed in 5 years.

There are a few normative changes to EIP-1:

- A limit of one section, and a requirement as to the order is added. This was always the case by editor consensus, but was never explicitly stated.
- The abstract now allows for a short discussion of the motivation, and also should define the scope of an EIP.
- The motivation section now explicitly states that readers should know that lack of interoperability is a problem implied when creating an EIP, and doesn't need to be included.
- The motivation section was made required for Core EIPs, since the former text stated that "A motivation section is critical for EIPs that want to change the Ethereum protocol."
- Guidelines are established for when another section conflicts with the Specification section, to avoid a repeat of EIP-712.
- The rationale section now requires authors to list any tradeoffs made, and allows authors to assume that the reader has read the motivation section.
- The backwards compatibility section was made non-optional, and placeholder text is suggested
- The backwards compatibility section has guidelines for post-final changes
- The test cases section now recommends inline key/value pairs over unit tests
- The reference implementation is now required to be minimal, and suggests authors submit more complete implementations to ercref and openzeppelin (mostly to make authors that like writing lots of code happy)
- The security considerations section now explicitly requires privacy considerations to be included, as well as providing a list of common security considerations to include.
- The security considerations section has guidelines for post-final changes
- Assets must be CC0 or CC0 multi-licensed, unless the author can't legally do so.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -97,7 +97,7 @@ A PR moving an EIP from Last Call to Final SHOULD contain no changes other than 
 
 ## What's in an EIP?
 
-An EIP should must no more than one of each section, in the following order:
+An EIP must contain no more than one of each section, in the following order:
 
 ### Preamble/Front Matter
 
@@ -184,6 +184,8 @@ The Security Considerations section discusses the security and privacy implicati
 ```md
 No security considerations found. Needs discussion.
 ```
+
+EIPs cannot be moved to Last Call or Final with an insufficient security considerations.
 
 This is a non-normative section. Additions of newly-identified security considerations should be proposed, even after an EIP reaches final.
 
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -95,20 +95,109 @@ A PR moving an EIP from Last Call to Final SHOULD contain no changes other than 
 
 **Living** - A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1.
 
-## What belongs in a successful EIP?
+## What's in an EIP?
 
-Each EIP should have the following parts:
+An EIP should must no more than one of each section, in the following order:
 
-- Preamble - RFC 822 style headers containing metadata about the EIP, including the EIP number, a short descriptive title (limited to a maximum of 44 characters), a description (limited to a maximum of 140 characters), and the author details. Irrespective of the category, the title and description should not include EIP number. See [below](./eip-1.md#eip-header-preamble) for details.
-- Abstract - Abstract is a multi-sentence (short paragraph) technical summary. This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does.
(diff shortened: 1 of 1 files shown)


EIPS/eip-1.md
```diff
@@ -95,20 +95,111 @@ A PR moving an EIP from Last Call to Final SHOULD contain no changes other than
...
+
+An EIP's preamble consists of RFC 822-style headers with EIP metadata. See [EIP Header/Preamble](#eip-header-preamble) for details.
+
+### Abstract
+
+**Required for all EIPs**
+
+The abstract is a short paragraph (multi-sentence) technical summary. This should be a very terse yet human-readable version of the specification and motivation sections, and should define the narrow scope of the EIP. Reading an abstract should be sufficient to generally understand the EIP.
+
+This is a normative section.
```
SamWilsn, line 114: The abstract is normative? I'm not sure I agree 100% on that, but I think I can see your reasoning.

SamWilsn (changes requested):
Mixing content changes with formatting changes makes it hard to compare.

I love the new formatting, but I think each content change needs to be made in its own PR so we can discuss properly.

EIPS/eip-1.md
Pandapip1: The abstract shouldn't change when the EIP turns final.

EIPS/eip-1.md
```diff
@@ -95,20 +95,111 @@ A PR moving an EIP from Last Call to Final SHOULD contain no changes other than
...
+This is a normative section, but in the case of a conflict, the Specification section is considered more normative, and the Test Cases should be amended.
+
+### Reference Implementation
+
+**Optional for all EIPs**
+
+The Reference Implementation provides a minimal non-production-ready implementation that people can use to assist in understanding or implementing the specification. More complex, production-ready implementations can be submitted to [ERCRef](https://github.com/ercref/ercref-contracts) and/or [OpenZeppelin Contracts](https://github.com/OpenZeppelin/openzeppelin-contracts).
+
+This is a non-normative section. In the case of a conflict with the Specification section, the Reference Implementation should be amended.
+
```
gcolvin, line 177: In #7230 I've suggested that for Core EIPs this section requires at lease a link to the relevant commits to the Executable Specifications.

gcolvin (approved):

EIPS/eip-1.md
gcolvin: @SamWilsn could you resolve this?  Git thinks you are requesting a change.  Or, @Pandapip1, could you make the chanbe.

I can see both arguments.  It does summarize the motivation, which is not normative.

Scamreno (commented):
Pandapip1-modernize-section-list #7047

PR #7062, opened 2023-05-24
Update EIP-4844: add `data_gas_used` to header

lightclient:
At the moment, EIP-4844 uses the `excess_data_gas` from a block's parent header in order to determine the data gas price of transactions in the current block. This is a bit different than EIP-1559 which first ensures the header's `base_fee` is correctly derived from the parent header, then proceeds to use that value for further gas price calculation.

tldr; `excess_data_gas` is pricing for _next_ block, while `base_fee` is pricing for _current_ block -- this PR aligns them so that both represent pricing information for the current block

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -250,10 +250,12 @@ On the execution layer, the block validity conditions are extended as follows:
 ```python
 def validate_block(block: Block) -> None:
     ...
+    
     # check that the excess data gas was updated correctly
     assert block.header.excess_data_gas == calc_excess_data_gas(block.parent.header)
 
     data_gas_used = 0
+    
     for tx in block.transactions:
         ...
 
@@ -288,7 +290,6 @@ def validate_block(block: Block) -> None:
 
     # ensure data_gas_used matches header
     assert block.header.data_gas_used == data_gas_used
-
 
 ```
 
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -122,7 +122,7 @@ The signature values `y_parity`, `r`, and `s` are calculated by constructing a s
 
 ### Header extension
 
-The current header encoding is extended with a new 64-bit unsigned integer field `data_gas_used` and a 256-bit unsigned integer field `excess_data_gas` . This is the running total of excess data gas consumed on chain since this EIP was activated. If the total amount of data gas is below the
+The current header encoding is extended with a new 64-bit unsigned integer field `data_gas_used` and a 256-bit unsigned integer field `excess_data_gas`. This is the running total of excess data gas consumed on chain since this EIP was activated. If the total amount of data gas is below the
 target, `excess_data_gas` is capped at zero.
 
 The resulting RLP encoding of the header is therefore:
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -122,7 +122,7 @@ The signature values `y_parity`, `r`, and `s` are calculated by constructing a s
 
 ### Header extension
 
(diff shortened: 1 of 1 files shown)


lightclient:
note, just an idea, wanted to see what it actually looked like to do this

EIPS/eip-4844.md
```diff
@@ -146,22 +146,22 @@ rlp([
     0x0000000000000000, # nonce
     base_fee_per_gas,
     withdrawals_root,
-    excess_data_gas
+    excess_data_gas,
+    data_gas_used
 ])
 ```
 
 The value of `excess_data_gas` can be calculated using the parent header and number of blobs in the block.
```
roberto-bayardo, line 154: remove "...and number of blobs in the block"?

roberto-bayardo:
I find this easier to understand. (...I've lost track of how many times I've gotten confused over parent vs current block for gathering appropriate data for blob pricing in the current spec.)  Not sure how we trade that off against requiring another header field though.

lightclient:
> Not sure how we trade that off against requiring another header field though.

IMO there is very little difference between adding one additional header field vs. two.

yperbasis (approved):
Yes, this is less confusing and error-prone.

lightclient:
Had to rebase - changes should be the same though.

g11tech (dismissed):

EIPS/eip-4844.md
```diff
@@ -122,7 +122,7 @@ The signature values `y_parity`, `r`, and `s` are calculated by constructing a s
 
 ### Header extension
 
-The current header encoding is extended with a new 256-bit unsigned integer field `excess_data_gas`. This is the running total of excess data gas consumed on chain since this EIP was activated. If the total amount of data gas is below the
+The current header encoding is extended with a new 64-bit unsigned integer field `data_gas_used` and a 256-bit unsigned integer field `excess_data_gas` . This is the running total of excess data gas consumed on chain since this EIP was activated. If the total amount of data gas is below the
```
adietrichs:
```suggestion
The current header encoding is extended with a new 64-bit unsigned integer field `data_gas_used` and a 256-bit unsigned integer field `excess_data_gas`. This is the running total of excess data gas consumed on chain since this EIP was activated. If the total amount of data gas is below the
```

EIPS/eip-4844.md
```diff
@@ -276,16 +278,18 @@ def validate_block(block: Block) -> None:
...
-            block_data_gas += get_total_data_gas(tx)
+            data_gas_used += get_total_data_gas(tx)
 
     # ensure the total data gas spent is at most equal to the limit
-    assert block_data_gas <= MAX_DATA_GAS_PER_BLOCK
+    assert data_gas_used <= MAX_DATA_GAS_PER_BLOCK
+
+    # ensure data_gas_used matches header
+    assert block.header.data_gas_used == data_gas_used
+
```
adietrichs, line 291:
```suggestion
```

EIPS/eip-4844.md
```diff
@@ -249,9 +250,10 @@ On the execution layer, the block validity conditions are extended as follows:
 ```python
 def validate_block(block: Block) -> None:
     ...
+    # check that the excess data gas was updated correctly
+    assert block.header.excess_data_gas == calc_excess_data_gas(block.parent.header)
 
-    block_data_gas = 0
-
+    data_gas_used = 0
```
adietrichs, line 256:
```suggestion
    data_gas_used = 0
    
```

adietrichs (dismissed):
left some whitespace nitpicks, but LGTM.

lightclient:
applied suggestions from @adietrichs, going to undraft now

PR #7069, opened 2023-05-25
Add EIP: Revamped CALL instructions

axic:


EIPS/eip-draft_calls2.md
```diff
@@ -0,0 +1,142 @@
+---
+eip: nnnn
```
lightclient, line 2:
```suggestion
eip: 7069
```

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+|------|-------|---------|
+| WARM_STORAGE_READ_COST | 100 | From [EIP-2929](./eip-2929.md) |
+| COLD_ACCOUNT_ACCESS | 2600 | From [EIP-2929](./eip-2929.md) |
+| ACCOUNT_CREATION_COST | 25000 | |
+| MIN_RETAINED_GAS | 5000 | |
+| MIN_CALLEE_GAS | 2300 | |
+
+We introduce three new instructions:
+
+- `CALL2` (`0xf8`) with arguments `(target_address, input_offset, input_size, value)`
```
wjmelements, line 52: should preserve the original arg order from CALL:
`(target_address, value, input_offset, input_size)`

axic: The motivation for changing the order is that both compiler and EVM implementations can be simplified as the difference is a stack value at the top or bottom.

EIPS/eip-draft_calls2.md
axic: Added.

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+
+Removing gas selectability also introduces a valuable property that future revisions to the gas schedule will benefit from: you can always overcome Out of Gas errors by sending more gas as part of the transaction (subject to the block gas limit). Previously when raising storage costs ([EIP-1884](./eip-1884.md)) some contracts that sent only a limited amount of gas to their calls were broken by the new costing.
+
+Hence some contracts had a gas ceiling they were sending to their next call, permanently limiting the amount of gas they could spend. No amount of extra gas could fix the issue as the call would limit the amount sent.  The notion of a stipend floor is retained in this spec. This floor can be changed independent of the smart contracts and still preserve the feature that OOG halts can be fixed by sending more gas as part of the transaction.
+
+### Stipend and 63/64th rule
+
+The purpose of the stipend is that a callee in case of being a "contract wallet" to have enough gas to emit logs (i.e. perform non-state-changing operations). The stipend is only added when the target has code and the call value is non-zero.
```
delaaxe, line 96:
```suggestion
The purpose of the stipend is to have enough gas to emit logs (i.e. perform non-state-changing operations) when calling from a "smart contract account". The stipend is only added when the target has code and the call value is non-zero.
```

delaaxe (approved):

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+    11c. `2` if the call has failed.
+11. Gas not used by the callee is returned to the caller.
+
+Note: Unlike `CALL` there is no extra charge for value bearing calls.
+
+*TODO:* Clarify which side (caller/callee) is gas deducted from and where an error originates from.
+
+*TODO:* Mention gas refunds?
+
+*TODO:* Consider option where non-calldata value transfer is not allowed, but there's a specific `TRANSFER`/`PAY` function for that. Would simplify the logic greatly.
```
SamWilsn, line 82: If you put these in comments, the linter will catch them for you before moving to review:

```suggestion
<!-- *TODO:* Clarify which side (caller/callee) is gas deducted from and where an error originates from. -->

<!-- *TODO:* Mention gas refunds? -->

<!-- *TODO:* Consider option where non-calldata value transfer is not allowed, but there's a specific `TRANSFER`/`PAY` function for that. Would simplify the logic greatly. -->
```

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Rationale
+
+### Removing gas selectability
+
+One major change from the original `CALL` series of instructions is that the caller has no conrol over the amount of gas passed in as part of the call. The number of cases where such a feature is essential are probably better served by direct protocol integration.
```
SamWilsn, line 88:
```suggestion
One major change from the original `CALL` series of instructions is that the caller has no control over the amount of gas passed in as part of the call. The number of cases where such a feature is essential are probably better served by direct protocol integration.
```

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+
+Removing gas selectability also introduces a valuable property that future revisions to the gas schedule will benefit from: you can always overcome Out of Gas errors by sending more gas as part of the transaction (subject to the block gas limit). Previously when raising storage costs ([EIP-1884](./eip-1884.md)) some contracts that sent only a limited amount of gas to their calls were broken by the new costing.
```
SamWilsn, line 90: Since you use OOG as an abbreviation later, you should expand it at least once beforehand:

```suggestion
Removing gas selectability also introduces a valuable property that future revisions to the gas schedule will benefit from: you can always overcome Out of Gas (OOG) errors by sending more gas as part of the transaction (subject to the block gas limit). Previously when raising storage costs ([EIP-1884](./eip-1884.md)) some contracts that sent only a limited amount of gas to their calls were broken by the new costing.
```

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+requires: 150, 211, 214, 2929
+---
+
+## Abstract
+
+Introduce three new call instructions, `CALL2`, `DELEGATECALL2` and `STATICCALL2`, with simplified semantics. The existing call instructions remain unchanged.
+
+The new instructions do not allow specifying a gas limit, but rather rely on the "63/64th rule" ([EIP-150](./eip-150.md)) to limit gas. An important improvement is the rules around the "stipend" are simplified, and callers do not need to perform special calculation whether the value is sent or not. 
+
+Furthermore, the obsolete functionality of specifying output buffer address is removed, because it is predominantly unused and implementers prefer to use `RETURNDATACOPY` instead. One exception is the case when the expected return size is known (i.e. non-dynamic return values), Solidity still uses the output buffer.
```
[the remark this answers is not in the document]
g11tech: but it isn't Rationale, but what the EIP does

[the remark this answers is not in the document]
g11tech: This would be better i think:

Furthermore, the obsolete functionality of specifying output buffer address is removed in favor of using `RETURNDATACOPY` instead.

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+
+Lastly, instead of returning a boolean for execution status, an extensible list of status codes is returned: `0` for success, `1` for revert, `2` for failure.
+
+We expect most new contracts to rely on the new instructions (for simplicity and in order to save gas), and some specific contracts where gas limiting is required to keep using the old instructions (e.g. [ERC-4337](./eip-4337.md)).
```
g11tech, line 24:
```suggestion
```
should also be moved to motivation/rationale

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Motivation
+
+Observability of gas has been a problem for very long. The system of gas has been (and likely must be) flexible in adapting to changes to both how Ethereum is used as well as changes in underlying hardware.
```
g11tech, line 28:
```suggestion
Predictability of gas has been a problem for very long. The system of gas has been (and likely must be) flexible in adapting to changes to both how Ethereum is used as well as changes in underlying hardware.
```

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+
+Unfortunately, in many cases compromises or workarounds had to be made to avoid affecting call instructions negatively, mostly due to the complex semantics and expectations of them.
+
+This change aims to remove gas observability from the new instructions and opening the door for new classes of contracts that are not affected by repricings. Furthermore, once the EVM Object Format (EOF) is introduced, the legacy call instructions can be rejected within EOF contracts, making sure they are mostly unaffected by changes in gas prices. Because these operations are required for removing gas observability they will be required for EOF in lieu of the existing instructions.
```
g11tech, line 32:
```suggestion
This change aims to remove gas observability from the new instructions and opening the door for new classes of contracts that are not affected by repricings. Furthermore, once the EVM Object Format (EOF) is introduced, the legacy call instructions can be rejected within EOF contracts, making sure they are mostly unaffected by changes in gas fees. Because these operations are required for removing gas observability they will be required for EOF in lieu of the existing instructions.
```

EIPS/eip-7069.md
g11tech: may be this should be mentioned and explained in rationale for better clarity

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
...
+
+Additionally there is a call depth counter, and calls fail if the depth would exceed 1024.
+
+Before the 63/64th rule was introduced, it was required to calculate available gas semi-accurately on caller side. Solidity has a complicated ruleset where it tries to estimate how much it will cost on the caller side to perform the call itself, in order to set a reasonable gas value.
+
+We have changed the ruleset:
+
+1. Removed the call depth check.
+2. Use the 63/64th rule, but
+    2a. ensure that at least 5000 gas is retainted prior to executing the callee,
```
g11tech, line 111:
```suggestion
    2a. ensure that at least 5000 gas is retained prior to executing the callee,
```

EIPS/eip-7069.md
axic: Added.

EIPS/eip-7069.md
axic: Ok perhaps fine to move to motivation, but rationale?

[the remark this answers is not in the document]
axic: I assumed 4337 is final given all the fanfare around it 😅 

It is fine, there's a long way for EIP and this here helps explanation.

EIPS/eip-7069.md
```diff
@@ -0,0 +1,145 @@
+---
+eip: 7069
+title: Revamped CALL instructions
```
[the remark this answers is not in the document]
axic: I see most EIP titles are not following this scheme. Is this a required change?

EIPS/eip-7069.md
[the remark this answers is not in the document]
axic: This was moved to rationale and included Vyper's case.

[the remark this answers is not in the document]
axic: Moved.

axic:
@g11tech @SamWilsn @lightclient is this closer to merging now?

g11tech (commented):
looks good from my side, leaving to @SamWilsn for final review

EIPS/eip-7069.md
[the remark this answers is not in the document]
SamWilsn: Nope, feel free to ignore.

SamWilsn (approved):

PR #7095, opened 2023-05-30
Update EIP-4844: reduce size of `excess_data_gas` to 64 bit

lightclient:
It seems like an oversight on my part in #5353 to make `excess_blobs` a 256 bit unsigned integer and it has propagated through the various changes until now. @karalabe noticed that `excess_data_gas` will never even need more than a 64 bit unsigned integer. After 1500 blocks, the data fee ends up being around `2**256`, which is obviously much larger than could possibly be paid. `log2(1500*(MAX_DATA_GAS_PER_BLOCK - TARGET_DATA_GAS_PER_BLOCK)) = 28.5`, so 64 bit is fine.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -123,7 +123,7 @@ The signature values `y_parity`, `r`, and `s` are calculated by constructing a s
 
 ### Header extension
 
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden (approved):

yperbasis (approved):

lightclient:
Approved in CL call 110

PR #7100, opened 2023-05-31
Update EIP-4844: clarify to must be non-nil

lightclient:
@karalabe raised a concern that the current transaction type spec in 4844 is not clear on whether a `nil` `to` field is allowed. This PR aims at clarifying any confusion around the correct representation of blob txs.


lightclient:
nice number

g11tech (approved):

adietrichs (approved):
LGTM

PR #6691, opened 2023-03-15
Update EIP-1: add support for linking to other well known SDO specifications

kdenhartog:
This defines how EIP authors can link to RFCs, W3C specs, and WHATWG specs. It additionally defines regexes that comply with the stable links that are used by each SDO so that the eipw bot can be updated easily as well.

PR #7111, opened 2023-06-01
Update EIP-1: Add W3C as a permissible origin

SamWilsn:
Originally from #6691 by @kdenhartog.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -252,6 +252,24 @@ Permitted Networking Specifications URLs must anchor to a specific commit, and s
 ^https://github.com/ethereum/devp2p/blob/[0-9a-f]{40}/.*$
 ```
 
+### World Wide Web Consortium (W3C)
+
+Links to a W3C "Recommendation" status specification may be included using normal markdown syntax. For example, the following link would be allowed:
+
+```markdown
+[Secure Contexts](https://www.w3.org/TR/2021/CRD-secure-contexts-20210918/)
+```
+
+Which renders as:
+
+[Secure Contexts](https://www.w3.org/TR/2021/CRD-secure-contexts-20210918/)
+
+Permitted W3C recommendation URLs MUST anchor to a specification in the technical reports namespace with a date, and so MUST match this regular expression:
+
+```regex
+^https:\/\/www\.w3\.org\/TR\/[0-9][0-9][0-9][0-9]/.*$
+```
+
 ### Digital Object Identifier System
 
 Links qualified with a Digital Object Identifier (DOI) may be included using the following syntax:
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -267,7 +267,7 @@ Which renders as:
 Permitted W3C recommendation URLs MUST anchor to a specification in the technical reports namespace with a date, and so MUST match this regular expression:
 
 ```regex
-^https:\/\/www\.w3\.org\/TR\/[0-9][0-9][0-9][0-9]/.*$
+^https://www\.w3\.org/TR/[0-9][0-9][0-9][0-9]/.*$
 ```
 
 ### Digital Object Identifier System


xinbenlv (approved):
Great!

kdenhartog (approved):
I personally prefer mutable references, but understand the purpose in making these immutable via dates and do not object to it. LGTM

g11tech:
gtg from my side :+1:

Pandapip1 (commented):
Would still like the rules regarding the anchor text.

EIPS/eip-1.md
```diff
@@ -252,6 +252,24 @@ Permitted Networking Specifications URLs must anchor to a specific commit, and s
...
+```
+
+Which renders as:
+
+[Secure Contexts](https://www.w3.org/TR/2021/CRD-secure-contexts-20210918/)
+
+Permitted W3C recommendation URLs MUST anchor to a specification in the technical reports namespace with a date, and so MUST match this regular expression:
+
+```regex
+^https:\/\/www\.w3\.org\/TR\/[0-9][0-9][0-9][0-9]/.*$
```
kdenhartog, line 270:
```suggestion
^https:\/\/www\.w3\.org\/TR\/[0-9][0-9][0-9][0-9]\/.*$
```

need to escape this slash it looks like

EIPS/eip-1.md
```diff
@@ -252,6 +252,24 @@ Permitted Networking Specifications URLs must anchor to a specific commit, and s
 ^https://github.com/ethereum/devp2p/blob/[0-9a-f]{40}/.*$
 ```
 
+### World Wide Web Consortium (W3C)
+
+Links to a W3C "Recommendation" status specification may be included using normal markdown syntax. For example, the following link would be allowed:
+
+```markdown
+[Secure Contexts](https://www.w3.org/TR/2021/CRD-secure-contexts-20210918/)
```
chaals, line 260: Except it isn't 'Recommendation' status ...

bumblefudge: I agree that being more explicit/precise about which stages in the Recommendation track can be linked would be good! "Proposed Recommendation" seems a stable-enough referent for my tastes (given what I've seen of W3C timelines) but maybe the group prefers a hard line?

image

Also, how do we feel about W3C **Registries**?  I've seen some suspicion of registries per se in other threads but in this case (as upstream specification) I think it would be worth being able to link to them as authoritative...

chaals: >  being more explicit/precise about which stages in the Recommendation track can be linked would be good!

It would indeed.

Having lived through most of the development of W3C process (e.g. I had to produce the first "Candidate Recommencation", which at that time didn't even have a name and wasn't yet a formal part of the Process), I think we could learn some stuff.

Rather than trying to set bright lines about what is an acceptable status it would be better to be clear about what the different things mean, and to require more explanation / justification of a given reference. 

That is in fact roughly what W3C's reference policy for Recommendation is now. They tried really hard over many years to maintain something that had stricter rules until it was obvious that it was just a lot of bureaucracy and entanglement holding up work for the sake of not having to think hard about where the links went.

kdenhartog: I never realized this didn't make it to Rec status. Right now the main requirement of a link is that it has immutable content and is a stable URL since there's no official way to update links once an EIP has made it to final status. With that in mind, I think it's less about being rec status for W3C and more about having an official date to recognize the version being referenced and being published in TR since once it's hit there it appears to become a stable link to reference.

EIPS/eip-1.md
SamWilsn: Rust regexes (which is where these eventually end up) don't escape forward slashes.

EIPS/eip-1.md
SamWilsn: First and foremost, we want the link to exist forever. Secondly, we want the content of that link to never change.

Ideally, we'd prefer linking to stuff that's ready for production use, but that isn't technically a requirement.

I'm not familiar enough with W3C to know if the regex as written meets those requirements. I _think_ it's good enough? Would appreciate an expert confirming that though.

chaals: I believe the answer is "Yes". The regex matches documents for which W3C provides a pretty strong guarantee that they won't make changes beyond fixing markup errors, broken links, and definite typographical errors.

PR #7113, opened 2023-06-01
Update EIP-1: Add IETF as a permissible origin

SamWilsn:
Originally from https://github.com/ethereum/EIPs/pull/6691 by @kdenhartog.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -290,6 +290,24 @@ Permitted WHATWG specification URLs must anchor to a specification defined in th
 
 Although not recommended by WHATWG, EIPs must anchor to a particular commit so that future readers can refer to the exact version of the living standard that existed at the time the EIP was finalized. This gives readers sufficient information to maintain compatibility, if they so choose, with the version referenced by the EIP and the current living standard.
 
+### Internet Engineering Task Force (IETF)
+
+Links to an IETF Request For Comment (RFC) specification may be included using normal markdown syntax, such as:
+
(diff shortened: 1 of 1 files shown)


kdenhartog (dismissed):
LGTM thanks for following up on this since I never got back to it. I'll convert mine to focus just on WHATWG since I'm going to what to reference it in EIP-6963

EIPS/eip-1.md
```diff
@@ -252,6 +252,24 @@ Permitted Networking Specifications URLs must anchor to a specific commit, and s
 ^https://github.com/ethereum/devp2p/blob/[0-9a-f]{40}/.*$
 ```
 
+### Internet Engineering Task Force (IETF)
```
Pandapip1, line 255: Again, would mildly prefer thr following:


```suggestion
### IETF Request for Comments (RFC)
```

EIPS/eip-1.md
```diff
@@ -252,6 +252,24 @@ Permitted Networking Specifications URLs must anchor to a specific commit, and s
...
+
+Links to a Request For Comment (RFC) specification may be included using normal markdown syntax, such as:
```
Pandapip1, line 257:
```suggestion
Links to an IETF Request For Comment (RFC) specification may be included using normal markdown syntax, such as:
```

Pandapip1 (changes requested):
Again, I think we should ensure that EIP authors use the right reference (`RFC X`)

PR #7117, opened 2023-06-01
Update EIP-1: Add WHATWG as a permissible origin

SamWilsn:
Originally from https://github.com/ethereum/EIPs/pull/6691 by @kdenhartog.

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -270,6 +270,26 @@ Permitted W3C recommendation URLs MUST anchor to a specification in the technica
 ^https://www\.w3\.org/TR/[0-9][0-9][0-9][0-9]/.*$
 ```
 
+### Web Hypertext Application Technology Working Group (WHATWG)
+
+Links to WHATWG specifications may be included using normal markdown syntax, such as:
+
+```markdown
+[HTML](https://html.spec.whatwg.org/commit-snapshots/578def68a9735a1e36610a6789245ddfc13d24e0/)
+```
+
+Which renders as:
+
+[HTML](https://html.spec.whatwg.org/commit-snapshots/578def68a9735a1e36610a6789245ddfc13d24e0/)
+
+Permitted WHATWG specification URLs must anchor to a specification defined in the `spec` subdomain (idea specifications are not allowed) and to a commit snapshot, and so must match this regular expression:
+
+```regex
+^https:\/\/[a-z]*\.spec\.whatwg\.org/commit-snapshots/[0-9a-f]{40}/$
+```
+
+Although not recommended by WHATWG, EIPs must anchor to a particular commit so that future readers can refer to the exact version of the living standard that existed at the time the EIP was finalized. This gives readers sufficient information to maintain compatibility, if they so choose, with the version referenced by the EIP and the current living standard.
+
 ### Digital Object Identifier System
 
 Links qualified with a Digital Object Identifier (DOI) may be included using the following syntax:


kdenhartog (approved):
Huge thanks @SamWilsn for fixing these up! LGTM

g11tech:
fine by me :+1:

xinbenlv (approved):

EIPS/eip-1.md
```diff
@@ -252,6 +252,24 @@ Permitted Networking Specifications URLs must anchor to a specific commit, and s
 ^https://github.com/ethereum/devp2p/blob/[0-9a-f]{40}/.*$
 ```
 
+### Web Hypertext Application Technology Working Group (WHATWG)
```
Pandapip1, line 255:
```suggestion
### WHATWG
```
I weakly prefer the abbreviation.

SamWilsn: We require authors to expand abbreviations, so I figured we should probably too :rofl:

SamWilsn:
> Are we not going to add rules re: acceptable titles?

Can we do that in a separate PR? I think it'll require some thought (eg. subsections & anchor links, colons/dashes/spaces, just the identifier vs. the whole title, etc.)

kdenhartog:
Should we be adding a comment regarding this? 
Screenshot 2023-06-06 at 12 01 51 PM

It seems like a bad idea to do exactly what the authoritative body on the standard is telling us not to do.

SamWilsn:
> Should we be adding a comment regarding this?

They do an excellent job explaining in their banner! We have [`markdown-rel-links`](https://ethereum.github.io/eipw/markdown-rel-links/) that _sort of_ explains our reasoning behind our general external links policy. Would you prefer something in EIP-1?

> It seems like a bad idea to do exactly what the authoritative body on the standard is telling us not to do.

I think we fit into the "historical reference" exception?

Imagine some EIP gets finalized, and on-chain implementations written. Some years go by, and the WHATWG standard has evolved, but the contracts are frozen on-chain.

A new implementer comes along. If we link to the living standard, then important context will be lost, but if we link to the commit, the implementer can make an informed decision about what to do: use the old version or the current living edition.

kdenhartog:
Yeah I think a little note in here to clarify this in EIP-1 would be good to at least document our rationale for using commitments. It also just occurred to me that if they rewrite their commit history (I'd venture to guess they won't do this) then some commit links could be broken here. So I don't think we'll find a perfect solution here anyways and the next best thing is probably the commit and document why were ignoring their suggestions approach. We may find this wasn't the right approach later and have to update to the living standards link in the future, but we can cross that bridge when we need to.

EIPS/eip-1.md
```diff
@@ -252,6 +252,24 @@ Permitted Networking Specifications URLs must anchor to a specific commit, and s
...
+
+```markdown
+[HTML](https://html.spec.whatwg.org/commit-snapshots/578def68a9735a1e36610a6789245ddfc13d24e0/)
+```
+
+Which renders as:
+
+[HTML](https://html.spec.whatwg.org/commit-snapshots/578def68a9735a1e36610a6789245ddfc13d24e0/)
+
+Permitted WHATWG specification URLs must anchor to a specification defined in the `spec` subdomain (idea specifications are not allowed) and to a commit snapshot, and so must match this regular expression:
```
kdenhartog, line 267:
```suggestion
Permitted WHATWG specification URLs must anchor to a specification defined in the `spec` subdomain (idea specifications are not allowed) and to a commit snapshot. While anchoring to a particular snapshot is not recommended by WHATWG, this has been defined to do so still so that readers of specifications can be aware of the exact version of the WHATWG living standard being referenced by the author at the time they linked to it in an immutable way. This way readers can choose to implement against the version referenced by the author or the updated living standard if they choose. As such, the link must match this regular expression:
```

kdenhartog: @SamWilsn this suggestion should address my last comment

SamWilsn: Updated, with different wording. What do you think of:

> Although not recommended by WHATWG, EIPs must anchor to a particular commit so that future readers can refer to the exact version of the living standard that existed at the time the EIP was finalized. This gives readers sufficient information to maintain compatibility, if they so choose, with the version referenced by the EIP and the current living standard.

kdenhartog: LGTM for that text

PR #7122, opened 2023-06-01
Update EIP-155: Added LUKSO for 42, with note to Kovan

frozeman:
This adds LUKSO as 42 to the EIP155, with reference to the Kovan testnet


lightclient:
This EIP is already final and so we cannot proceed with this change. You should probably look to something like chainlist: https://github.com/DefiLlama/chainlist

PR #7132, opened 2023-06-02
Update EIP-6963: Update Window Events section

everdimension:
* Provide motivation for the design of Events Interfaces
* Better showcase symmetry of events on both sides

### DIFF
--- a/EIPS/eip-6963.md
+++ b/EIPS/eip-6963.md
@@ -90,65 +90,76 @@ interface EIP6963ProviderDetail {
 }
 ```
 
-### Window Event
+### Window Events
 
-In order to prevent provider collisions, the DApp and the Wallet are expected to emit an event and instantiate an eventListener to discover the various Wallet's. This forms an Event concurrency loop. 
+In order to prevent provider collisions, the DApp and the Wallet are expected to emit an event and instantiate an eventListener to discover the various Wallets. This forms an Event concurrency loop. 
 
-To emit events both DApps and Wallets MUST use the `window.dispatchEvent` function to emit events and MUST use the `window.addEventListener` function to observe events. There will be two Event interfaces used for the DApp and Wallet to discover each other.
+Since the DApp code and Wallet code aren't guaranteed to run in a particular order, the events are designed to handle such race conditions.
 
-#### Wallet Events
+To emit events both DApps and Wallets MUST use the `window.dispatchEvent` function to emit events and MUST use the `window.addEventListener` function to observe events. There are two Event interfaces used for the DApp and Wallet to discover each other.
+
+#### Announce and Request Events
 
 The `EIP6963AnnounceProviderEvent` interface MUST be a `CustomEvent` object with a `type` property containing a string value of `eip6963:announceProvider` and a `detail` property with an object as it's value of type `EIP6963ProviderDetail`. The `EIP6963ProviderDetail` object MUST be frozen by calling `Object.freeze()` on the  value of the `detail` property.
 
 ```typescript
-// A Wallet announcing the EIP6963ProviderDetail object
+// Announce Event dispatched by a Wallet
 interface EIP6963AnnounceProviderEvent extends CustomEvent {
   type: "eip6963:announceProvider";
   detail: EIP6963ProviderDetail;
 }
 ```
 
-The Wallet must announce to the DApp this `CustomEvent` via a `window.dispatchEvent()` function call. In order for the Wallet to recognize that the DApp is Web3 enabled and can support interactions with the Web3 Platform APIs the Wallet MUST add an EventListener to catch an Event dispatched from the DApp. This EventListener MUST have a `type` property with the string `eip6963:requestProvider` and MUST use a handler that will re-dispatch a `CustomEvent` object of type `EIP6963AnnounceProviderEvent`. This re-announcement by the Wallet is useful for when a Wallet's initial Event announcement may have been delayed or fired before the DApp had initialized it's EventListener. This allows the various Wallet Providers to react to the DApp without the need to pollute the `window.ethereum` namespace which can produce non-deterministic wallet behavior such as different wallets connecting each time.
+The `EIP6963RequestProviderEvent` interface MUST be an `Event` object with a `type` property containing a string value of `eip6963:requestProvider`.
 
 ```typescript
- function announceProvider() {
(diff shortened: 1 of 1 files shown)


GregTheGreek (dismissed):
Lgtm

EIPS/eip-6963.md
```diff
@@ -158,25 +158,40 @@ interface EIP6963AnnounceProviderEvent extends CustomEvent {
...
-Additionally the Wallet providers will react to the event emitted by the Ethereum library.
+Content script listens to the `"eip6963:requestProvider"` event and dispatches the `"eip6963:announceProvider"`:
 
 ```typescript
-window.addEventListener(
-  "eip6963:requestProvider",
-  (event: EIP6963RequestProviderEvent) => {
-    announceProvider();
-  }
+const announceEvent: EIP6963AnnounceProviderEvent = new Event(
```
darrylyeo, line 177: This should be a `CustomEvent` as there is a `detail` field.

EIPS/eip-6963.md
```diff
@@ -158,25 +158,40 @@ interface EIP6963AnnounceProviderEvent extends CustomEvent {
...
+window.addEventListener(
+  "eip6963:announceProvider",
+  (event: EIP6963AnnounceProviderEvent) => {}
+);
 
-// Wallet provider initializes
-window.dispatchEvent(new CustomEvent("eip6963:announceProvider", { detail }));
+// Page script dispatches a request event which will be heard by 
+// content scripts injected earlier
+window.dispatchEvent(new Event("eip6963:requestProvider"));
```
darrylyeo, line 171: I think this should also stay a `CustomEvent` for consistency.

EIPS/eip-6963.md
everdimension: Right, good catch! Thanks

EIPS/eip-6963.md
everdimension: Well, the `EIP6963RequestProviderEvent` extends a regular `Event` (which is correct, imo), so I think it should be used here

EIPS/eip-6963.md
```diff
@@ -139,17 +139,17 @@ The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and c
 
 ### Window Event
 
-Different wallet providers might inject their scripts at different times, plus there is no guarantees that the Ethereum library in the web page will be loaded before all injected scripts are present in the page.
+Since page scripts (Ethereum library code) and content scripts (Wallet provider code) aren't guaranteed to run in a particular order (the [official recommendation](https://developer.chrome.com/docs/extensions/mv3/content_scripts/#run_time) being for content scripts to run at "document idle" phase), the events are designed to handle such race conditions.
```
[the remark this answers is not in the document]
everdimension: I think the original concern about wallet providers fighting for the global is mentioned in a different section: https://github.com/ethereum/EIPs/blob/444313be872f5a11bfbd26aa27b752603672a62d/EIPS/eip-6963.md#L22

If you want, we can reiterate on it here, too. So after the last sentence, I'd add:

```
By incorporating an event system, wallet providers are relieved from both accidental and intentional conflicts that arise from having to interact with a single global.
```

EIPS/eip-6963.md
glitch-txs: there's no need to use CustomEvent since there's no data being transfered

everdimension:
Two of the validator error are from the lines unrelated to this diff.

Is passing these validators optional?

EIPS/eip-6963.md
kdenhartog: Does it make a difference in terms of interop whether a CustomEvent interface or Event interface is used? Given this is non-normative I think implementers could do either or and it wouldn't impact interoperability here.

kdenhartog:
How do you want to roll these changes into this one https://github.com/ethereum/EIPs/pull/7134/files?

everdimension:
I have updated the PR to include editorial changes made in the now merged https://github.com/ethereum/EIPs/pull/7134

In addition to providing more context for motivation behind the design of the events, these changes also fix some of the issues of https://github.com/ethereum/EIPs/pull/7134:

* The "Wallet Events" and "DApp Events" sections are merged into one section. This solves the problem of having to mention the interface from the latter section in the former one, which created a kind of "used before defined" problem.
* Fixes a semantical error which prescribed the dapp to send a wrong event
* Fixes confusing terminology about EventListeners needing to have types. Since event listeners in JavaScript are functions, they do not have a type. Instead, they are called with an Event object (in fact, event listeners _can_ can be objects with a `handleEvent` method, but their type is still irrelevant)
* Some minor grammatical fixes

EIPS/eip-6963.md
darrylyeo: I like `CustomEvent` better because it better signifies that it's a non-native browser event, plus it's simpler for introspection and TypeScript type narrowing.

darrylyeo: Also, this would leave open the possibility of dapps exposing metadata about themselves to the wallet providers. You never know 😉

everdimension: @darrylyeo I initially preferred the use of `Event` to follow the rule of least power (i.e. you shouldn't choose `CustomEvent` if a regular `Event` suffices)

But at the same time it does seem like a good idea to pave the road towards dapp being able to send some request data along with the event.

I had an idea of making an additional proposal (or perhaps even updating this one) where a dapp can pass a `locale` in the `eip6963:requestProvider` request, and the wallet would attach a localized `name` field in the `EIP6963ProviderInfo` if it supports that language.

everdimension: Yeah I'd expect nothing to break irregardless of whether `Event` or `CustomEvent` is dispatched from the DApp
Wallets would be expected to check for the `detail` field in case they want to handle it

kdenhartog (approved):
Changes LGTM. For the most part these are just re-ordering to make this a bit easier to read with some additional context being added.

kdenhartog:
This should be done by the wallet not by the DApp since the wallets are the ones meant to be enforcing this security on behalf of their users. I likely made a mistake when typing this up and this should get fixed.

glitch-txs:
Could you guys check #https://github.com/ethereum/EIPs/pull/7175 also not sure why the linting failed

PR #7134, opened 2023-06-03
Update EIP-6963: Editorial changes

kdenhartog:
Just some editorial updates mainly so far, still making changes as I go. Wanted to get this published so others can provide feedback as I'm finishing up.

### DIFF
--- a/EIPS/eip-6963.md
+++ b/EIPS/eip-6963.md
@@ -33,6 +33,13 @@ This is achieved by introducing a set of window events to provide a two-way comm
 
 The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
 
+### Definitions
+
+Wallet: A user agent that manages keys and facilitates transactions with Ethereum.
+
+Decentralized Application (DApp): A web page that relies upon one or many Web3 platform APIs which are exposed to the web page via the Wallet.
+
+
 ### Provider Info
 
 Each wallet provider will be announced with the following interface `EIP6963ProviderInfo` that will be used to display to the user:
--- a/EIPS/eip-6963.md
+++ b/EIPS/eip-6963.md
@@ -51,8 +51,8 @@ interface EIP6963ProviderInfo {
 
 The values in the `EIP6963ProviderInfo` MUST be used as follows:
 
-- **`uuid`** - a globally unique of the wallet provider that MUST be (UUIDv4[^rfc4122] compliant) to uniquely distinguish 
-- **`name`** - A human-readable name of the wallet provider to be displayed to the user on the DApp. (e.g. `DopeWalletExtension` or `AwesomeWallet`)
+- **`uuid`** - a globally unique of the wallet provider that MUST be (UUIDv4[^rfc4122] compliant) to uniquely distinguish different [EIP-1193](./eip-1193.md) provider sessions. The cryptographic uniqueness provided by UUIDv4[^rfc4122] guarantees that two independent `EIP6963ProviderInfo` objects can be separately identified.
+- **`name`** - A human-readable local alias of the wallet provider to be displayed to the user on the DApp. (e.g. `DopeWalletExtension` or `AwesomeWallet`)
 - **`icon`** - A URI[^rfc3986] pointing to an image. Icon images MUST be square with 96x96px minimum resolution
 
 [^rfc4122]:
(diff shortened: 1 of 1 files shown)


EIPS/eip-6963.md
```diff
@@ -33,9 +33,21 @@ This is achieved by introducing a set of window events to provide a two-way comm
...
+
+Decentralized Application (DApp): A web page that relies upon one or many Web3 platform APIs which are exposed to the web page via the Wallet.
+
+Provider Discovery Library: A library or piece of software that assists a DApp to interact with the Wallet. 
+
 ### Provider Info
 
-Each wallet provider will be announced with the following interface `EIP6963ProviderInfo` that will be used to display to the user:
+Each wallet provider will be announced with the following interface `EIP6963ProviderInfo`. The values in the `EIP6963ProviderInfo` MUST be used as follows:
+
```
kdenhartog, line 47: Note, I dropped the `walletId` property here because UUIDv4 + Wallet Display Name effectively achieves the same goal to uniquely identify the providers I believe. Curious to get some feedback on this change which is the only technical change in this PR at this point. I can split it out into a separate PR as well if it's a bit more controversial.

EIPS/eip-6963.md
```diff
@@ -296,3 +229,100 @@ The use of SVG images introduces a cross-site scripting risk as they can include
 ## Copyright
 
 Copyright and related rights waived via [CC0](../LICENSE.md).
+
+
+[^rfc4122]:
```
kdenhartog, line 234: @pedrouid do you care how we end up citing these RFCs? With #7113 incoming I think we can do this more easily now using the usual markdown linking normally used.

EIPS/eip-6963.md
```diff
@@ -268,22 +195,28 @@ function onPageLoad() {
...
-### Ethereum library
+### DApp implementation
 
-Here is a reference implementation for a Ethereum library to display and track multiple wallet providers that are injected by browser extensions.
+Here is a reference implementation for a DApp to display and track multiple wallet providers that are injected by browser extensions. 
 
 ```typescript
-function onPageLoad() {
-  const providers: EIP6963ProviderDetail[] = [];
+function onPageLoad(): EIP6963ProviderDetail[] => {
```
jxom, line 203:
```suggestion
function onPageLoad() => {
```

Return types are inferred in TypeScript.

jxom: Nit: Shouldn’t it ideally return a cleanup function if anything though? To clean up the `addEventListener`

jxom:
Looking good!

EIPS/eip-6963.md
kdenhartog: Potentially, but it's not required so that would be up to an implementer to decide if they want to take that approach. My goal with these changes is to rely upon the text to decide what should be done not the examples. The examples, are there more to make the text a bit more clear.

jxom: Makes sense!

EIPS/eip-6963.md
```diff
@@ -135,101 +91,83 @@ interface EIP6963ProviderDetail {
 }
 ```
 
-The `EIP1193Provider` interface is documented at [EIP-1193](./eip-1193.md) and can be used to override the `window.ethereum` object once the user has explicitly selected it.
-
 ### Window Event
 
-Different wallet providers might inject their scripts at different times, plus there is no guarantees that the Ethereum library in the web page will be loaded before all injected scripts are present in the page.
+In order to prevent provider collisions, the DApp and the Wallet are expected to emit an event and instantiate an eventListener to discover the various Wallet's. This forms an Event concurrency loop.
```
everdimension, line 96: While "Wallet" is used as a term, I don't think an apostrophe is needed for pluralizing it

EIPS/eip-6963.md
```diff
@@ -135,101 +91,83 @@ interface EIP6963ProviderDetail {
...
 );
 ```
+#### DApp Events
 
-## Rationale
-
-### Interfaces
-
-Standardizing a provider info interface (`EIP6963ProviderInfo`) allows determining the necessary information to populate a wallet selection popup. This is particularly useful for Ethereum libraries such as Web3Modal, RainbowKit, Web3-Onboard, ConnectKit, etc.
+The DApp MUST listen for the `CustomEvent` dispatched by the Wallet via a `window.dispatchEvent()` function call. In order for the DApp to discover the `EIP6963AnnounceProviderEvent` this EventListener MUST have a `type` property with the string `eip6963:requestProvider`.
```
everdimension, line 138: The DApp actually listens to the "announce" event:
```suggestion
The DApp MUST listen for the `CustomEvent` dispatched by the Wallet via a `window.dispatchEvent()` function call. In order for the DApp to discover the `EIP6963AnnounceProviderEvent` this EventListener MUST have a `type` property with the string `eip6963:announceProvider`.
```

flexibrah (approved):

PR #7154, opened 2023-06-08
Update EIP-4844: Increase Blob Throughput

adietrichs:
Increases the throughput to a target of 3 blobs (0.375MB) and limit of 6 blobs (0.75MB) as decided on https://github.com/ethereum/pm/issues/786.

Also adjusts `DATA_GASPRICE_UPDATE_FRACTION` to keep targeting a max change rate for the blob gas price of 12.5% per block (see EIP section for math).


EIPS/eip-4844.md
```diff
@@ -427,9 +427,9 @@ TBD
 
 ## Security Considerations
 
-This EIP increases the storage requirements per Beacon block by a maximum of ~0.5 MB.
-This is 4x larger than the theoretical maximum size of a block today (30M gas / 16 gas per calldata byte = 1.875M bytes), and so it will not greatly increase worst-case bandwidth.
-Post-merge, block times are expected to be static rather than an unpredictable Poisson distribution, giving a guaranteed period of time for large blocks to propagate.
+This EIP increases the bandwidth requirements per Beacon block by a maximum of ~0.75 MB.
```
ralexstokes, line 430: unrelated but nitpick
```suggestion
This EIP increases the bandwidth requirements per beacon block by a maximum of ~0.75 MB.
```

EIPS/eip-4844.md
```diff
@@ -45,10 +45,10 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `BLOB_COMMITMENT_VERSION_KZG` | `Bytes1(0x01)` |
 | `POINT_EVALUATION_PRECOMPILE_ADDRESS` | `Bytes20(0x14)` |
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
-| `MAX_DATA_GAS_PER_BLOCK` | `2**19` |
-| `TARGET_DATA_GAS_PER_BLOCK` | `2**18` |
+| `MAX_DATA_GAS_PER_BLOCK` | `786432` |
+| `TARGET_DATA_GAS_PER_BLOCK` | `393216` |
```
ralexstokes, line 49: checking math: 
2**18 works out to 2 blobs (4096 * 32 bytes per blob given trusted setup size)
2**19 works out to 4 blobs (max is double target)

if we increase target by 50% (to move from 2 to 3), then we in fact get the numbers proposed

EIPS/eip-4844.md
```diff
@@ -45,10 +45,10 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
...
 | `MIN_DATA_GASPRICE` | `1` |
-| `DATA_GASPRICE_UPDATE_FRACTION` | `2225652` |
+| `DATA_GASPRICE_UPDATE_FRACTION` | `3338477` |
```
ralexstokes, line 51: why did this change?

ralexstokes (approved):
looks good, I think we should move ahead w/ this change on 4844 devnets and can always dial it back down if needed

EIPS/eip-4844.md
ralexstokes: ah I see below this follows from the gas price sensitivity calculation in the "Data gasprice update rule" section below

adietrichs: Exactly. I added a sentence pointing that out to the PR description.

EIPS/eip-4844.md
dankrad: consider `3 * 2**18` and `3 * 2**17` for legibility?

ralexstokes: could add in the third column (so after the decimal representations), I do like having the multiplier * base factor representation

adietrichs: the problem is that this depends on the specific context, e.g. not obvious why it always should be base 2. or also, might be better in this specific case to always use factor * 2**17 (so that factor would be the number of blobs)

PR #7165, opened 2023-06-11
Update EIP-6963: Add walletId back into the design

jxom:
The [wagmi](https://github.com/wagmi-dev) core team proposes to add `walletId` back into the design of EIP-6963.

Decided not to use an `rdns` (Reverse DNS) for the said deterministic identifier to keep the objective clear, focused and simple, without having to deal with the complexities of potential side-effects of an rDNS in the future.

### DIFF
--- a/EIPS/eip-6963.md
+++ b/EIPS/eip-6963.md
@@ -2,7 +2,7 @@
 eip: 6963
 title: Multi Injected Provider Discovery
 description: Using window events to announce injected wallet providers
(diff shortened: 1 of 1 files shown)


EIPS/eip-6963.md
```diff
@@ -39,27 +39,72 @@ Wallet: A user agent that manages keys and facilitates transactions with Ethereu
 
 Decentralized Application (DApp): A web page that relies upon one or many Web3 platform APIs which are exposed to the web page via the Wallet.
 
-Provider Discovery Library: A library or piece of software that assists a DApp to interact with the Wallet. 
+Provider Discovery Library: A library or piece of software that assists a DApp to interact with the Wallet (e.g. wagmi, Web3Modal, RainbowKit, Web3-Onboard, ConnectKit, etc).
```
jxom, line 42: Moved from L172 to here to make it clear on examples of Provider Discovery Libraries. Readers might need an example to properly piece it together before moving on.

PR #7172, opened 2023-06-12
Update EIP-4844: Update precompile address for 4844

ralexstokes:
It looks like the next available precompile address is `0xA` so let's use this one in 4844

(cf https://github.com/ethereum/execution-specs/blob/master/src/ethereum/shanghai/vm/precompiled_contracts/__init__.py#L38 )


Kowtko21:
[$nicholaskowtko]()

g11tech (approved):
lgtm

lightclient (approved):
LGTM, discussed on ACD 164 and everyone was okay with making this change

PR #7178, opened 2023-06-13
Update EIP-4788: Bound precompile storage

ralexstokes:


EIPS/eip-4788.md
```diff
@@ -52,37 +53,63 @@ Validity is guaranteed from the consensus layer, much like how withdrawals are h
...
 
 #### New stateful precompile
 
-Beginning at the execution timestamp `FORK_TIMESTAMP`, the code and storage at `HISTORY_STORAGE_ADDRESS` constitute a "stateful" precompile.
+Beginning at the execution timestamp `FORK_TIMESTAMP`, a "stateful" precompile is deployed at `HISTORY_STORAGE_ADDRESS`.
+
+Callers of the precompile should provide the `timestamp` they are querying encoded as 32 bytes in big-endian format.
 
-Callers of the precompile should provide the `root` they are querying encoded as 32 bytes.
+Given this input, the precompile reduces the `timestamp` in the same way during the write routine and first checks if
```
djrtwo, line 87: Given the timestamp is checked before the root is read, it seems a bit more natural to have it ordered timestamp buffer followed by the root buffer. But just a stylistic suggestion

i.e.
1. Check something
2. if check passes,  move on the next thing to read

EIPS/eip-4788.md
```diff
@@ -52,37 +53,63 @@ Validity is guaranteed from the consensus layer, much like how withdrawals are h
...
-timestamp = sload(HISTORY_STORAGE_ADDRESS, root)
-evm.returndata[:32].set(timestamp)
+timestamp = evm.calldata[:32]
+timestamp_reduced = to_uint64_be(timestamp) % HISTORICAL_ROOTS_LENGTH
+timestamp_extended = timestamp_reduced + HISTORICAL_ROOTS_LENGTH
+timestamp_index = to_uint256_be(timestamp_extended)
+
+recorded_timestamp = sload(HISTORY_STORAGE_ADDRESS, timestamp_index)
+if recorded_timestamp != timestamp:
+    evm.returndata[:32].set(0x0000000000000000000000000000000000000000000000000000000000000000)
```
djrtwo, line 104: Probably define zero-word as a constant so it can be referenced here and in the text above

EIPS/eip-4788.md
```diff
@@ -52,37 +53,63 @@ Validity is guaranteed from the consensus layer, much like how withdrawals are h
...
+else:
+    root_index = to_uint256_be(timestamp_reduced)
+    root = sload(HISTORY_STORAGE_ADDRESS, root_index)
+    evm.returndata[:32].set(root)
 ```
 
-If there is no timestamp stored at the given root, the opcode follows the existing EVM semantics of `sload` returning `0`.
+Alongside the existing gas for calling the precompile, there is an additional gas cost of `G_beacon_root` to reflect the two (2) implicit `SLOAD`s from
```
djrtwo, line 111: What do you mean "existing gas"? Do you mean `CALL` costs? I would be a bit more specific.

Also, is the cost of the SLOAD sufficient because it's dominant or should we estimate the other operations (adds, assignments, etc)

Also, this could be priced cheaper than SLOAD if we just assume these two buffers are held in memory. That said, diverging from treating it just like native code/storage is probably not great for stateless and other things

EIPS/eip-4788.md
```diff
@@ -104,6 +131,16 @@ be nonfavorable conditions.
 Use of block root over state root does mean proofs will require a few additional nodes but this cost is negligible (and could be amortized across all consumers,
 e.g. with a singleton state root contract that caches the proof per slot).
 
+### Why two ring buffers?
+
+The first ring buffer only tracks `HISTORICAL_ROOTS_LENGTH` worth of roots and so for all possible timestamp values would consume a constant amount of storage.
+However, this design opens the precompile to an attack where a skipped slot that has the same value modulo the ring buffer length would return an old root value,
+rather than the most recent one.
+
+To nullify this attack, this EIP keeps track of the pair of data `(parent_beacon_block_root, timestamp)` for each index into the
```
djrtwo, line 140:
```suggestion
To nullify this attack while retaining a fixed memory footprint, this EIP keeps track of the pair of data `(parent_beacon_block_root, timestamp)` for each index into the
```

djrtwo (commented):
nice! a bit strange, but I like it

EIPS/eip-4788.md
ralexstokes: sure I can update

EIPS/eip-4788.md
ralexstokes: yeah I may just write `uint256(0)` instead

I was explicit here just to belabor the point but I think most readers know what is meant here

EIPS/eip-4788.md
ralexstokes: > Do you mean CALL costs? I would be a bit more specific.

yes, the idea is that as a user I need to `CALL` the precompile and I want to say on top of whatever existing costs to get to this point in the execution, let's add in some extra gas to reflect the `SLOAD`s, ill make another pass on the language

EIPS/eip-4788.md
```diff
@@ -52,43 +53,68 @@ Validity is guaranteed from the consensus layer, much like how withdrawals are h
 At the start of processing any execution block where `block.timestamp >= FORK_TIMESTAMP` (i.e. before processing any transactions),
 write the parent beacon root provided in the block header into the storage of the contract at `HISTORY_STORAGE_ADDRESS`.
 
-The root itself is used as a key into the contract's storage and the timestamp of the header is written as the key's value.
-The timestamp (a 64-bit unsigned integer value) is encoded as 32 bytes in big-endian format.
+In order to bound the storage used by this precompile, two ring buffers are used: one to track the latest timestamp at a given index and another to track
```
ralexstokes, line 56:
```suggestion
In order to bound the storage used by this precompile, two ring buffers are used: one to track the latest timestamp at a given index in the ring buffer and another to track
```

EIPS/eip-4788.md
```diff
@@ -29,7 +29,8 @@ restaking constructions, smart contract bridges, MEV mitigations and more.
 |---                           |---                                           |---
 | `FORK_TIMESTAMP`             | TBD                                          |
 | `HISTORY_STORAGE_ADDRESS`    | `Bytes20(0xB)`                               |
-| `G_beacon_root`              | 2100                                         | gas
+| `G_beacon_root`              | 4200                                         | gas
+| `HISTORICAL_ROOTS_LENGTH`    | 8192                                         |
```
ralexstokes, line 33: note: this provides just a bit over 1 day of history, the only downside I can see here is if a user needed to verify something against a root and was facing censorship although I think we generally consider it too expensive to censor for an entire day

jochem-brouwer: I think currently this is not correct. If we have this as key for timestamps: `timestamp_reduced = block_header.timestamp % HISTORICAL_ROOTS_LENGTH`, then this would actually generate 2048 unique keys to store this timestamp on, rather than 8192 (which means only ~6.82 hours are saved). I think what is meant here is that 8192 roots are saved, so the calculation of the timestamp key should adapt here to ensure that 8192 keys can fit here.

ralexstokes: yep good catch, this commit should actually just be rolled back; I was talking to someone about this and convinced myself I had made an error but the original value is correct, thanks!

EIPS/eip-4788.md
```diff
@@ -43,7 +44,9 @@ To bound the amount of storage this construction consumes, a ring buffer is used
 Beginning at the execution timestamp `FORK_TIMESTAMP`, execution clients **MUST** extend the header schema with an additional field: the `parent_beacon_block_root`.
 This root consumes 32 bytes and is exactly the [hash tree root](https://github.com/ethereum/consensus-specs/blob/fa09d896484bbe240334fa21ffaa454bafe5842e/ssz/simple-serialize.md#merkleization) of the parent beacon block for the given execution block.
 
-Validity is guaranteed from the consensus layer, much like how withdrawals are handled.
+Validity of the parent beacon block root is guaranteed from the consensus layer, much like how withdrawals are handled.
+
+When verifying a block, execution clients should ensure the root value in the block header matches the one provided by the consensus client.
```
ralexstokes, line 49:
```suggestion
When verifying a block, execution clients **MUST** ensure the root value in the block header matches the one provided by the consensus client.
```

EIPS/eip-4788.md
```diff
@@ -52,43 +55,75 @@ Validity is guaranteed from the consensus layer, much like how withdrawals are h
...
 
 #### New stateful precompile
 
-Beginning at the execution timestamp `FORK_TIMESTAMP`, the code and storage at `HISTORY_STORAGE_ADDRESS` constitute a "stateful" precompile.
+Beginning at the execution timestamp `FORK_TIMESTAMP`, a "stateful" precompile is deployed at `HISTORY_STORAGE_ADDRESS`.
+
+Callers of the precompile should provide the `timestamp` they are querying encoded as 32 bytes in big-endian format.
+Clients **MUST** sanitize this input call data to the precompile.
+If the input is _more_ than 32 bytes, the precompile only takes the first 32 bytes of the input buffer and ignores the rest.
+If the input is _less_ than 32 bytes, the precompile should revert.
```
jochem-brouwer, line 90: To clarify: if it reverts, would it just put 0 on the stack (signalling that the CALL has failed), instead of 1? Or, does it consume all provided gas? (In the earlier case the precompile always consumes 4200 gas)

jochem-brouwer: Regarding the reference implementation: seems like a 'normal' evm revert.

EIPS/eip-4788.md
```diff
@@ -29,7 +29,8 @@ restaking constructions, smart contract bridges, MEV mitigations and more.
 |---                           |---                                           |---
 | `FORK_TIMESTAMP`             | TBD                                          |
 | `HISTORY_STORAGE_ADDRESS`    | `Bytes20(0xB)`                               |
-| `G_beacon_root`              | 2100                                         | gas
+| `G_beacon_root`              | 4200                                         | gas
+| `HISTORICAL_ROOTS_LENGTH`    | 98304                                        |
```
jochem-brouwer, line 33: I think this name is now a bit misleading:

```suggestion
| `HISTORICAL_ROOTS_LENGTH`    | 8192                                        |
| `HISTORICAL_ROOTS_MODULO`    | `HISTORICAL_ROOTS_LENGTH * 12` (98304)                                        |
```

(Then change elsewhere in EIP as well).

Otherwise people think 98304 roots are saved, which is not the case. (Also note that this "only" works if blocks have a constant drift time of `12 * x` (`x` integer) - what happens if we at some point change this drift time..? If we change it to 6 seconds we save twice as much. Could also note that this is thus the max roots ever saved on the chain, which is nice since it upper bounds it)

jochem-brouwer:
Also I just realized this (regarding testnets):

There is a super ugly edge case here, in which you do not pre-fund the 0x0B address in genesis. In that case, the balance is 0, the nonce is zero, and the code is the empty code hash. The storage however, will be non-empty starting at the fork block. However, this still means that the account is empty as per https://eips.ethereum.org/EIPS/eip-158. This means that if someone calls into 0x0B on the testnet (but does not fund it), then this should cleanup the account after the tx, thus *wiping the storage*. Not sure if it should be mentioned anywhere in the EIP, since I think most (all?) testnets pre-fund the precompile addresses with 1 wei, but I could imagine this gets messed up in testing setups if they set it up privately :thinking: (It is super weird if you call with tx1 into the contract to get timestamp X, and then call in the same block again into the contract again with timestamp X, and now get a zero result while it was nonzero before :thinking: )

EIPS/eip-4788.md
ralexstokes: the intention is a "normal" EVM revert

which I believe just puts 0 on the stack and consumes the 4200 gas

I don't see a reason for the precompile to consume all provided gas

If anything we could try to refine where execution fails and change how much gas is consumed but it is much simpler to just change the same amount no matter what

jochem-brouwer: Yup, I agree :+1:

EIPS/eip-4788.md
ralexstokes: I'd rather just call the 98304 number the `HISTORICAL_ROOTS_MODULUS`

there is a bit of a hidden dependency on the 12 number which comes from the `SECONDS_PER_SLOT` constant on the consensus layer but part of the design conversation around this EIP wants to avoid mixing the CL constant into the EL

we could change this slot time and it would have implications for this EIP in terms of the number of roots stored but (1) this does set an upper bound and (2) I don't think we will change the slot time so dramatically that it drastically changes the effective window a user of this precompile has to use

worst case we can always adjust how this precompile works if we do something like adjusting the slot time

ralexstokes:
yeah it sounds like the working knowledge is to pre-fund precompiles to avoid this behavior (cc @MariusVanDerWijden )

I can add a note to the EIP

PR #7194, opened 2023-06-20
Update EIP-3076: Update eip-3076.md

nalepae:
Fix ordered list rendering issue

Adding blank line after `4. Refuse to sign...` to comply with the markdown linter seems to reset the HTML ordered list...
image

This PR removes the blank line. (The markdown linter will complain, but at least the HTML rendering will be OK.)

### DIFF
--- a/EIPS/eip-3076.md
+++ b/EIPS/eip-3076.md
@@ -173,14 +173,15 @@ After importing an interchange file with data field `data`, a signer must respec
 2. Refuse to sign any block with `slot <= min(b.slot for b in data.signed_blocks if b.pubkey == proposer_pubkey)`, except if it is a repeat signing as determined by the `signing_root`.
 
 3. Refuse to sign any attestation that is slashable with respect to the attestations contained in `data.signed_attestations`. For details of what constitutes a slashable attestation, see `is_slashable_attestation_data`.
+
 4. Refuse to sign any attestation with source epoch less than the minimum source epoch present in that signer's attestations (as seen in `data.signed_attestations`). In pseudocode:
 
-```python3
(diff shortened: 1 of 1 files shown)


SamWilsn:
If you indent the fenced code block, I think you can get it to be a part of the `<li>`. There's an [example in EIP-1](https://github.com/ethereum/EIPs/blob/082834fed2556b1eb83729d35c2ecb3c7926f243/EIPS/eip-1.md?plain=1#L371-L396).

michaelsproul (approved):
Let's try it

michaelsproul:
Dang, it didn't work. Sorry, I didn't actually try compiling locally because I have 0 in the way of Ruby stuff installed.

Will re-visit in a bit unless you feel like tackling it @nalepae

nalepae:
Actually I tried to build it locally but did not achieve to do it because Ruby stuffs seem to have trouble with Mac M1.
(Tried on docker as well but did not achieve as well.)

==> I'll build it on a linux machine today and will try to both have ordered list AND markdown linter to be happy.

nalepae:
To reproduce the bug in local:

```
docker run -it ruby:2.6 bash
git clone https://github.com/ethereum/EIPs.git
cd EIPs
bundle install
bundle exec jekyll doctor
bundle exec jekyll build
bundle exec htmlproofer ./_site --check-html --check-opengraph --report-missing-names --log-level=:debug --assume-extension --empty-alt-ignore --timeframe=6w --disable-external
```

==> HTML file is here: `_site/EIPS/eip-3076.html`

Bug reproduced, let's fix it :)

Issue #7198, opened 2023-06-20
Proposal: Make EIPW only check changed lines except for changing status


xinbenlv:
This was shared on: https://ethereum-magicians.org/t/proposal-eipw-should-only-complain-about-changing-lines/14762

### Proposed Change

Currently some Author Pain comes from the EIPW complains about **lint errors that were not introduced in that specific pull request they are working on**, such as 
1. links that were not allowed
2. mandates to call ERC "ERC" or call ERC "EIP"
3. spaces between section
4. order of section

In particular, sometimes new policy will be imposed on EIP, and it means many existing EIP are in the state of inconsistency.

This is particular confusing and painful which requires Authors to get up to date to EIP editorial policy.

I hereby propose a change to EIPW:

1. For any PR that's not changing state or Draft/Review, only lint the line of code being changed.
2. Only lint full EIP when moving into Last Call or Final

This was inspired by https://twitter.com/sproulM_/status/1671096282482622464?s=20

xinbenlv:
Created #7199

Pandapip1:
I would be okay with the following:

- eipw will always output an error for last call EIPs
- eipw will always output an error for status change PRs
- Otherwise, eipw will output warnings (non-blocking annotations)

xinbenlv:
That sounds great

Pandapip1:
No. An additional requirement is that all status changes (not just review -> last call or last call -> final) also proc eipw.

xinbenlv:
Oh, I see. I am ok all status change commits require error-clean. Let me update this proposal

xinbenlv:
@Pandapip1 agree with you. This update is now reflected in #7199 , can you see if it's good to merge as a draft?

gcolvin:
Yes, please. I've been bitten by this myself more than once.

PR #5270, opened 2022-07-15
Update EIP-5000: Renumber to EIP-5159

SamWilsn:
The EIP number was sniped using a bot. In order to discourage sniping,
we're renumbering the EIP to a less attractive number.

PR #7199, opened 2023-06-20
Add EIP: Linter Scope

xinbenlv:
Reflecting #7198


EIPS/eip-7198.md
```diff
@@ -0,0 +1,33 @@
+---
+eip: 7198
```
abcoathup, line 2:
```suggestion
eip: 7199
```
Assigning the EIP number as PR number as standard practice rather than using the issue number #7198.  
I recommend following standard practice for EIP editors so there isn't any perception of special treatment.

xinbenlv:
Thank you @abcoathup, same as you, I am a big fan to avoid special treatment being an Editor, just like I said in [this post](https://github.com/ethereum/EIPs/pull/5270#issuecomment-1471384844).

I have no objection to use 7198 or 7199 or any number. Since you prefer 7199 I am going with what you want.

gcolvin (approved):

Andredreyer1 (approved):

Andredreyer1 (approved):

PR #7201, opened 2023-06-21
Add EIP: Namespaced Storage Layout

frangio:


EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
+---
+eip: <to be assigned>
```
Amxx, line 2:
```suggestion
eip: 7201
```

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
...
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-06-20
+---
+
+## Abstract
+
+We define a formula to compute from an identifier a location in storage that is suitable for structs of arbitrary size. The formula is chosen to be safe against collisions with the standard Solidity storage layout. We define a convention to document this location in Solidity source code.
```
Amxx, line 15:
```suggestion
We define a formula to derivate a location in storage, that is suitable for structs of arbitrary size, from an identifier. The formula is chosen to be safe against collisions with the standard Solidity storage layout. We define a convention to document this location in Solidity source code.
```

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Specification
+
+A _namespace_ consists of a set of variables that are placed contiguous in the storage layout of a contract. It should be implemented as a struct.
+
+A _namespace id_ is a string that uniquely identifies a namespace in a contract. It should not contain any whitespace characters.
+
+The storage location for a namespace is defined as `ns_loc(id: string) = keccak256(uint256(keccak256(id)) - 1)`.
+
+A Solidity contract using namespaced storage can annotate a struct with the NatSpec tag `@custom:storage-location erc1234:<NAMESPACE_ID>` to identify it as a namespace with id `<NAMESPACE_ID>`.
```
Amxx, line 33:
```suggestion
A Solidity contract using namespaced storage can annotate a struct with the NatSpec tag `@custom:storage-location erc7201:<NAMESPACE_ID>` to identify it as a namespace with id `<NAMESPACE_ID>`.
```

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
...
+
+No backward compatibility issues found.
+
+## Reference Implementation
+
+```solidity
+pragma solidity ^0.8.19;
+
+contract Example {
+    /// @custom:storage-location erc1234:example.main
```
Amxx, line 60:
```suggestion
    /// @custom:storage-location erc7201:example.main
```

Amxx (commented):
Update EIP number

EIPS/eip-xxxx.md
pcaversaccio, line 1: Since @Amxx asked me about feedback, here some inputs:

- As a general note: the storage slot location of a `HashMap` (a `mapping` in Solidity) in Vyper is `keccak256(slot_hashmap, key)` and not like in Solidity `keccak256(key, slotMapping)`. So the order is simply reversed. That should not have any security implications, but it's good to know.

- Also, the storage slot ordering in Vyper is sequential. Let's make an example:

```vyper
# @version ^0.3.9
# filename: example.vy
owner: public(address)
minter: public(address)
balanceOf: public(HashMap[address, uint256])
```

To verify this, you can run the following command to get the storage layout of any contract:

```bash
vyper -f layout .\example.vy
```

This will result in the following output for the example above:

```vyper
{"storage_layout": {"owner": {"type": "address", "slot": 0}, "minter": {"type": "address", "slot": 1}, "balanceOf": {"type": "HashMap[address, uint256]", "slot": 2}}, "code_layout": {}}
```

Please note that Vyper does currently not support tight packing, but will do in the near future.

frangio: >One issue, I generally see with this EIP, is that it increases the code size.

Is that with this EIP in particular or with the namespaced/diamond storage pattern in general?

I assume you say this because storage locations are now 32 bytes instead of 1-2 bytes.

pcaversaccio: Right, it's more of a general statement, since whenever you access some variable `x` it adds 32 bytes to the bytecode. That was also one of the arguments why [hash-based storage slots](https://github.com/vyperlang/vyper/issues/1733) have not been implemented in Vyper (so far).

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
...
+
+The storage location for a namespace is defined as `ns_loc(id: string) = keccak256(uint256(keccak256(id)) - 1)`.
+
+A Solidity contract using namespaced storage can annotate a struct with the NatSpec tag `@custom:storage-location erc7201:<NAMESPACE_ID>` to identify it as a namespace with id `<NAMESPACE_ID>`.
+
+## Rationale
+
+A requirement for the location is that it shouldn't overlap with any storage location that would be used by the standard Solidity layout. This is in case namespaced storage is used alongside standard linear storage, either deliberately or accidentally.
+
+First, note that a namespace may be larger than a single storage slot, so a variable in a namespace will be placed in a slot `ns_loc(id) + k`. If we assume collision resistance of Keccak-256, the chosen `ns_loc` function has the desired property with very high probability, because the cases in which a Solidity variable receives a location like it, that is of the form `keccak256(X)`, are:
```
ernestognw, line 39:
```suggestion
First, note that a namespace may be larger than a single storage slot, so a variable in a namespace will be placed in a slot `ns_loc(id) + k`. If we assume collision resistance of Keccak-256, the chosen `ns_loc` function has the desired property with very high probability, because the cases in which a Solidity variable receives a location of the form `keccak256(X)` are:
```

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. Arrays:
+    1. If the array is at the top level and is variable number `N` in the layout, the location of the `k`th item in the array will be `keccak256(N) + k`. If `N` will be a number much smaller than `uint256(keccak256(id)) - 1`.
```
ernestognw, line 42: This section is confusing:

> If `N` will be a number much smaller than `uint256(keccak256(id)) - 1`.

From what I read in the other bullet points, this is justifying why there won't be a collision. If so, I'd suggest:

```suggestion
    1. If the array is at the top level and is variable number `N` in the layout, the location of the `k`th item in the array will be `keccak256(N) + k`, but `N` will be a number much smaller than `uint256(keccak256(id)) - 1`.
```

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
...
+    2. If the array is within another array or mapping, it will be in some location `keccak256(X) + j`, and the `k`th item will be at `keccak256(keccak256(X) + j) + k`. For this to equal `ns_loc(X) + k` we would need `j = -1`, but `j` will always be a positive number in standard Solidity layout.
+2. Mappings: The value for key `q` in a mapping will be at location `keccak256(H(q) .. X)` where `X` is the location of the mapping itself. Note that `H(q)` can be any number of bytes. If it is a non-zero number of bytes, it is distinct from any `ns_loc(id) + k`. If it is zero bytes, it can be that `ns_loc(id) = keccak256(X)` if `X = keccak256(id) - 1`. But we know that `X` is the location of the mapping and as mentioned for arrays above a variable can be at `keccak256(Y) + j` for a positive number `j`.
```
ernestognw, line 44: I think we might want to define `H()` or reference the [Solidity docs](https://solidity-fr.readthedocs.io/fr/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays) because it might not be immediately clear.

ernestognw, line 44: Following the notation used in Solidity docs

```suggestion
2. Mappings: The value for key `q` in a mapping will be at location `keccak256(H(q) . X)` where `X` is the location of the mapping itself. Note that `H(q)` can be any number of bytes. If it is a non-zero number of bytes, it is distinct from any `ns_loc(id) + k`. If it is zero bytes, it can be that `ns_loc(id) = keccak256(X)` if `X = keccak256(id) - 1`. But we know that `X` is the location of the mapping and as mentioned for arrays above a variable can be at `keccak256(Y) + j` for a positive number `j`.
```

ernestognw, line 44:
```suggestion
2. Mappings: The value for key `q` in a mapping will be at location `keccak256(H(q) .. X)` where `X` is the location of the mapping itself. Note that `H(q)` can be any number of bytes. If it is a non-zero number of bytes, it is distinct from any `ns_loc(id) + k`. If it is zero bytes, it can be that `ns_loc(id) = keccak256(X)` if `X = keccak256(id) - 1`, but we know that `X` is the location of the mapping and (as mentioned for arrays above) a variable will be at `keccak256(Y) + j` for a positive number `j`.
```

ernestognw (commented):
Adding comments on readability mostly, I also want to ask, what's the rationale for picking variable names? For example, we have `q`, `k` and `j` in lowercase but `X` and `Y` in uppercase. Similarly we use `H()` uppercase where the Solidity docs use `h()`.

EIPS/eip-xxxx.md
frangio: I would favor referencing the Solidity docs, though EIP editors will reject the external link so we may have to simply say "as defined in the Solidity documentation".

@Amxx wants to rewrite this proof so I will leave this unresolved until then.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,89 @@
+---
+eip: 7201
```
abcoathup, line 2: Confirming the assignment of the EIP number as the pull request number.  Please rename the filename to match.

xinbenlv (approved):

EIPS/eip-xxxx.md
Amxx: > * As a general note: the storage slot location of a `HashMap` (a `mapping` in Solidity) in Vyper is `keccak256(slot_hashmap, key)` and not like in Solidity `keccak256(key, slotMapping)`. So the order is simply reversed. That should not have any security implications, but it's good to know.

Do you have a link to documentation for that ?

pcaversaccio: I don't think it's in the docs, but iirc it was specced out here: https://github.com/vyperlang/vyper/issues/769, and in the compiler code you see it here: https://github.com/vyperlang/titanoboa/blob/e0852fe309315172517675dca8910eee66a5ead3/boa/vyper/contract.py#L318-L333, where the decoder uses `slot, k`.

PR #7212, opened 2023-06-22
Add EIP: Precompiled for secp256r1 Curve Support

ulerdogan:
Add EIP: Precompiled for secp256r1 Curve Support

The post is about a new EIP which proposes the addition of a new precompiled contract to the EVM that allows signature verifications in the “secp256r1” elliptic curve by given parameters of message hash, r - s components of the signature, and x - y coordinates of the public key.


xinbenlv (approved):
LGTM, 

The proposal provided reasonable motivation and clear specs about the input and output and tests to validate, gave gas cost and other spec decision in the rationale. Editorially, I think this PR is good enough to merge as a Draft. 

Leaving technical feedback, such as the design choice of verifying only `v` component, to the peer reviews by the client developers.

EIPS/eip-7212.md
```diff
@@ -0,0 +1,180 @@
...
+
+```
+
+### Required Checks in the Verification
+
+Some requirements have to be checked to understand if the given signature components are valid:
+
+- Verify that both `r` - `s` values are greater than 0 and less than the curve order.
+- Verify that s is equal to or less than half of the order of the subgroup to prevent signature malleability.
+- Verify that the point formed by (`x`, `y`) values are on the curve.
```
jwasinger, line 90: Also should check that `x`/`y` are less than `p`.

jwasinger (commented):
I left some comments.

One other comment:  I don't fully understand the rationale behind not compressing the public key.  It would save a substantial amount of calldata (31 bytes) and there's already precedence for it with existing secp256k1 ecrecover precompile.

EIPS/eip-7212.md
```diff
@@ -0,0 +1,180 @@
...
+    - If the signature verification process succeeds, it returns 1 in 32 bytes format, if not it returns 0 in 32 bytes format for the result with a null error.
+    - If the given `x` and `y` coordinates are not valid, it returns null for the result with the given error.
+
+### Precompiled Contract Gas Usage
+
+The use of signature verification cost by `P256VERIFY` is `3450` gas. Following reasons and calculations are provided in the [Rationale](#rationale) and [Test Cases](#test-cases) sections.
+
+## Rationale
+
+The "secp256r1" elliptic curve signatures consists of `v`, `r`, `s` components. Even if recovering the public key on the curve is possible, most of the applications are not generating `v` component of the signature and it causes an uncertainty of the result values. However, the signatures can be verified with only `r` - `s` values. In order to provide an exact and more compatible method, verification is preferred over recovery to propose in a precompiled.
```
jwasinger, line 112: It would be nice to clarify exactly what we gain from requiring `x`/`y` as inputs.  It's not clear to me from the explanation here.

And requiring them as inputs requires substantial added call data.

dcposch: If I understand correctly,

- `v` is a sign bit
- the P256 precompile is intended to support hardware signatures
- most of the existing hardware signature mechanisms (Secure Enclave, Webauthn, etc) return only the (r,s) signature

easy to recover v -- you try the two possible values, and see which one returns your public key.

so the comparison is something like

### `v, r, s` style

Pros:
- **Matches ecrecover.**
- **Less calldata.** On the surface, 1 word = 32 bytes less, but actual savings should be a bit more, since the `v` parameter is mostly 0 bytes and therefore cheaper.

Cons:
- **Users / libraries have to recover `v`.** This is a bit awkward. Instead of just calling (hardware signing module), it has to call (hardware signing module) > (attempt verification with + v) > (attempt verification with - v). I am not a cryptographer obv, I am not sure if there's a better way.

### `v, r, x, y` style

Pros:
- **Better matches the hardware interface.** Users/libraries don't have to recover `v`, they just pass a pubkey (x,y) directly from the hardware enclave.

EIPS/eip-7212.md
```diff
@@ -0,0 +1,180 @@
...
+3. **Android Keystore:** Android Keystore is an API that manages the private keys and signing methods. The private keys are not processed while using Keystore as the applications’ signing method. Also, it can be done in the “Trusted Execution Environment” in the microchip.
+4. **Passkeys:** Passkeys is utilizing FIDO Alliance and W3C standards. It replaces passwords with cryptographic key-pairs which is also can be used for the elliptic curve cryptography.
+
+Modern devices have these signing mechanisms that are designed to be more secure and they are able to sign transaction data, but none of the current wallets are utilizing these signing mechanisms. So, these secure signing methods can be enabled by the proposed precompiled contract to initiate the transactions natively from the devices and also, can be used for the key management. This proposal aims to reach maximum security and convenience for the key management.
+
+## Specification
+
+As of `FORK_TIMESTAMP` in the integrated EVM chain, add precompiled contract `P256VERIFY` for signature verifications in the “secp256r1” elliptic curve at address `PRECOMPILED_ADDRESS`.
```
dcposch, line 31: **Curious if people are interested in `CREATE2` shadowing.**

In short, if we make `PRECOMPILED_ADDRESS` the CREATE2 address of a contract that implements identical functionality--instead of a special low address like `0x00..0019`--then it allows a smoother deployment.

Chains will adopt this EIP at different times; maybe certain L2s are first, others later, and mainnet last. With CREATE2 shadowing, contracts can work on all chains immediately. They just become more gas-efficient once a given chain adds the precompile. A related benefit is that it reduces the need for upgradeability in user contracts.

This is especially salient for this particular precompile, since a primary use-case for it is hardware backed wallets. A 4337 wallet is itself a CREATE2 contract works on all chains with identical code.

EIPS/eip-7212.md
```diff
@@ -0,0 +1,180 @@
...
+No backward compatibility issues found as the precompiled contract will be added to `PRECOMPILED_ADDRESS`, one of the next address in the precompiled address set.
+
+## Test Cases
+
+Functional tests are applied for multiple cases in the [reference implementation](#reference-implementation) of `P256VERIFY` precompiled contract and they succeed. Benchmark tests are also applied for both `P256VERIFY` and `ECRECOVER` with some pre-calculated data and signatures in the “go-ethereum”s precompile testing structure to propose a meaningful gas cost for the “secp256r1” signature verifications by the precompiled contract implemented in the [reference implementation](#reference-implementation). The benchmark test results by example data in the assets can be checked:
+
+- [P256Verify Benchmark Test Results](../assets/eip-7212/p256Verify_benchmark_test)
+- [Ecrecover Benchmark Test Results](../assets/eip-7212/ecrecover_benchmark_test)
+
+```
```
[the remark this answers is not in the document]
ulerdogan: Hi, thanks for your comments! How do the benchmarks be displayed? I think that it's important to propose gas costs for the precompiled contract. I thought if I am comparing two results to propose gas cost in the same system would be beneficial. Should I include more systems benchmark results or completely remove it?

EIPS/eip-7212.md
```diff
@@ -0,0 +1,180 @@
...
+
+- **Input data:** 160 bytes of data including:
+    - 32 bytes of the signed data `hash`
+    - 32 bytes of the `r` component of the signature
+    - 32 bytes of the `s` component of the signature
+    - 32 bytes of the `x` coordinate of the public key
+    - 32 bytes of the `y` coordinate of the public key
+- **Output data:** 32 bytes of result data and error
+    - If the signature verification process succeeds, it returns 1 in 32 bytes format, if not it returns 0 in 32 bytes format for the result with a null error.
+    - If the given `x` and `y` coordinates are not valid, it returns null for the result with the given error.
```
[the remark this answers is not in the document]
ulerdogan: Thanks! I will update with this information. The implementation uses big-endian numbers.

EIPS/eip-7212.md
ulerdogan: I've tried to mean this statement with the below information and the reference implementations `elliptic.P256().IsOnCurve(x, y)` check verifies that the coordinate values are less than `p`. I will point out better in the EIP specification.

> Verify that the point formed by (x, y) values are on the curve.

EIPS/eip-7212.md
xinbenlv: I like your idea. I see the biggest merit of what you describe being a route reduce barrier for adding precompile contracts, mitigating the chicken-and-egg problem of "needs to demonstrate usage and interest to convince adding precompile" vs "needs adding the precompile to get more usage".

EIPS/eip-7212.md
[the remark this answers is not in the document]
xinbenlv: IMHO The benchmark is being supplied as assets, I personally think it's a good supporting evidence for rationale like the gas cost choice. Hence I think it's good to be there. My two cents. Others might disagree.

EIPS/eip-7212.md
xinbenlv: FYI https://eips.ethereum.org/EIPS/eip-2098 for  might be relevant in the same spirit where user / library recovers `v` for ECDSA. 

This is provided to you as a reference and you can gauge the preference by usage / adoption for designing the precompiles

ulerdogan: Thanks @dcposch for the explanation of the logic.

Yes, the mechanisms using the curve are returning only `r` and `s`, so the verification is possible with inputting public key coordinates `x` and `y`.

We need `v` value to recover public key without the `x` and `y` coordinates. The `v` value can be found by above-mentioned methods, but let me explain my design choice:

- While we are making recovery with the `ecrecover`, we can reach the public address of the EOA accounts, so it can be directly used in the smart contracts. Unlikely, recovering the `secp256r1` public key does not match any default stored types, and we still need to store the account public key.
- Having to find the `v` value in the implementation part of the signature creates complexity on the application side. So, I didn't want to bring this complexity for the applications. Still, I would love to reassess and edit the EIP to implement recovery after discussion.

EIPS/eip-7212.md
ulerdogan: The idea looks very interesting. I also would love to follow future plans and integrations about it.

dcposch: @xinbenlv happy to, but does it need an EIP? this is more a general pattern that can be applied to all precompiles that are pure functions

it raises the question of "why haven't we done it so far", but i think there were good reasons:
- early precompiles like [eip 152](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md) didn't use `CREATE2` shadowing because CREATE2 didn't exist yet, and also because cross-chain compatibility wasn't a concern yet
- many L2 precompiles like Optimism's [`0x42....` series](https://www.evmdiff.com/diff?base=1&target=10) can't use it because they're not pure functions (instead, it's a built-in L2 specific oracle or bridge)

for future pure functions precompiles, CREATE2 shadowing seems like a win/win

xinbenlv: I agree with you having a CREATE2 will be a wthat it's win/win

To answer your question "does it need an EIP"? I'd assume the question being: Do we need to standardize it?

Here is what I think:

When there are many ways and variables ("moving pieces") to do the same thing, standardization helps people coordinate. An example would be screw driver and screw, the size of screw bing standardize so people can manufacture screw and drivers independently and they will work together.

In the CREATE2 shadowing, I presume there are a few "moving pieces." that if people make different choices, they will end up with the different sizes of screw, hence there are value to standardize.

1. For the same functionality, ideally everyone use the same contact address generated by CREATE, so that ultimately, with a hard-fork we can change them to precompile and with lower gas cost. We don't want people generating different contacts deployed at different addresses.

2. How to determine the input of CREATE2? Which `sender`, which `nonce`, which `bytecode`?  
 - 2.1 Are there going to be a sender that everyone want to use so if nonce and bytecode are also determined it we can all agree upon the contract address generated by CREATE2?
 - 2.2 How do we generate `nonce`? Shall we do something like ERC-600 / ERC-601 or BIP-32?
 - 2.3 Should there be [any magic](https://twitter.com/_SamWilsn_/status/1666793849585369088?s=20) in bytecode? E.g. Should bytecode be consistent should there be any additional behavior that caller can expect?
 
 These are a few things I can think of.

To this point, I think the discussion of CREAT2 shadwing seems a bit off topic from this EIP. I'd suggest we make a ethereum-magician.org post to discuss, and leave this PR for the Precomiple of secp256r1 Curve support.

dcposch: Yeah, fair points. The CREATE2 address would eventually have to be part of the proposed precompile regardless. I agree re: standardizing how it's created.

Happy to make an ethmagicians post

ulerdogan: Considering that the solutions that are expected to be presented as precompiled contracts cannot always be implemented as a Solidity smart contract, can we say that this is a downside for the standardization purpose?

paulmillr:
-1

1. 256r1 is usually more vulnerable to timing attacks than stuff like 25519.
2. 256r1 is not even recommended by nist at this point, 384r1 is.
3. There are some rumors with regards to general security of r1 curves, it's unclear.
4. Adding a new elliptic curve impl into ALL execution layer clients is not a trivial task. I don't think the feature is too useful for this.

EIPS/eip-7212.md
dcposch: Thread moved to ethmagicians: https://ethereum-magicians.org/t/create2-shadowing-for-precompiles/14821/3

jskelly84 (approved):

xinbenlv:
@paulmillr thanks for the feedback. 
Editor here: for any technical feedback like those you gave, they are recommended to be put in https://ethereum-magicians.org/t/eip-7212-precompiled-for-secp256r1-curve-support/14789 so when this PR is merged, the discussion can continue and also has a higher chance to be seen by other peer reviewers.

ulerdogan:
> **[paulmillr](/paulmillr) ** commented

Hi, thanks for you comment! 

Moved it to the EM forum for discussion: https://ethereum-magicians.org/t/eip-7212-precompiled-for-secp256r1-curve-support/14789/5?u=ulerdogan

SamWilsn (dismissed):

SamWilsn (approved):

jskelly84 (approved):

PR #7230, opened 2023-06-25
Update EIP-1: EIP pain relief

gcolvin:
This draft PR is very much a work in progress.  So far it tries to alleviate some known pain points in the EIP process.  These include
* getting a Draft into the repo in the first place
* editing a document while complying with changing rules
* referencing relevant resources that aren't specifically allowed
* ensuring that all referenced resources remain accessible
* assuring that the many Drafts we edit are technically sound and of high quality
* integrating the EIP process with the Reference Implementations and Core Devs workflows.

**Note: This is independent of whether to split the repo.**

*Note: The intent is that once an EIP is a Draft the Core Developers will have full control of their workflow until the EIP reaches FInal.  The stages of their workflow are not defined by the Editors.  The Editors should never be in the position of blocking an upgrade.*

To begin with the last:

The Editors and the Developers are fairly independent organizations, with the Editors trying provide the services needed to publish a consistent series of high-quality EIPs.  We had envisioned (years back) that most proposals would arrive via independent Working Groups.  For the most part working groups didn't happen except, de-facto, the Core Developers.

We propose to clarify the status of Working Groups, and to treat the Core Developers as an independent Working Group, responsible for the specification of Core EIPs and _in control of their own workflow._

The Editorial stages for WG EIPs are reduced to three: Draft, Final, or Withdrawn.  Anything in between is part of the WG workflow, and should be tracked as such.  The formatting and notational requirements for a Specification, its relationship to other specifications, none of these are editorial concerns.  They belong to the Working Group.

The Editors retain responsibility for publishing a high-quality series of standards.  To that end we maintain the overall requirements for spelling, style, headers, citations, overall format (Abstract ... References ...) and the like.  (Some core developers are already serving as editors, so we have a start on coordination.)

For the rest:

We propose to relax the enforcement of EIPW rules for Drafts, enforcing tighter rules only on changes of Status.  This should reduce a lot of the friction in the workflow.

Our policy of accepting only a limited range of URLs has caused much pain.  We propose to relax that policy.   We allow wide editorial _and_ working group discretion on the careful use of external resources, in line with IETF practice.  _All_ references must include _full citations_ with authors, title, and publication information, including available DOIs.  This helps to ensure that over time external resources can almost always be found _somewhere_.

We add a few optional headers for use by Working Groups to track their workflow.

We propose to introduce the role of Technical Peers --  volunteers with relevant expertise who can help Authors review their work at the Idea stage.  Editors may ask for them when a proposal needs more technical review than the Editors are able to give it.  This should help to reduce the strain on the Editor's and increase the quality and general usefulness of proposals, especially ERCs.

EDITED 2023-08-01


EIPS/eip-1.md
```diff
@@ -67,54 +67,51 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
-
-**Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
-
-A PR moving an EIP from Last Call to Final SHOULD contain no changes other than the status update. Any content or editorial proposed change SHOULD be separate from this status-updating PR and committed prior to it.
+- **Idea** - An idea that is pre-draft. This is not tracked within the EIP Repository.
+- **Draft** - The first formally tracked stage of an EIP in development. An EIP is merged by an EIP Editor into the EIP repository when it meets some initial requirements: It should be technically sound, the Title and Preamble must be correct, and all included Sections should be readable.  Problems with language like spelling, grammar, markup, etc. are only considered warnings.
+- **Review** - An EIP Author marks an EIP as ready for and requesting Peer Review.  An Editor will merge the EIP when it meets all EIP requirements: it should be a technically sound, complete specification, properly formatted, with correct language, markup and external references.
```
Pandapip1, line 79: The wording here feels a bit off.

EIPS/eip-1.md
```diff
@@ -67,54 +67,51 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+- **Last Call** - This is the final review window for an EIP before moving to `Final`. An EIP editor will assign `Last Call` status and set a review end date (`last-call-deadline`), typically 14 days later.
+  - If this period results in necessary normative changes it will revert the EIP to `Review`.
+- **Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
+  - A PR moving an EIP from Last Call to Final SHOULD contain no changes other than the status update. Any content or editorial proposed change SHOULD be separate from this status-updating PR and committed prior to it.
```
Pandapip1, line 83: How about any status to any status?

EIPS/eip-1.md
```diff
@@ -67,54 +67,51 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
-- Copyright Waiver - All EIPs must be in the public domain. The copyright waiver MUST link to the license file and use the following wording: `Copyright and related rights waived via [CC0](../LICENSE.md).`
-
-## EIP Formats and Templates
+- **Preamble** - RFC 822 style headers containing metadata about the EIP, including the EIP number, a short descriptive title (limited to a maximum of 44 characters), a description (limited to a maximum of 140 characters), and the author details. Irrespective of the category, the title and description should not include EIP number. See [below](./eip-1.md#eip-header-preamble) for details.
```
Pandapip1, line 97: I feel like these should be formatted as Nth level headings instead of bullets.

EIPS/eip-1.md
gcolvin: Agreed.  Just a sketch to get discussion started.

EIPS/eip-1.md
gcolvin: Not sure, but I don't think I changed this.

EIPS/eip-1.md
gcolvin: I'll try it.

EIPS/eip-1.md
```diff
@@ -19,17 +19,37 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
...
-- A **Meta EIP** describes a process surrounding Ethereum or proposes a change to (or an event in) a process. Process EIPs are like Standards Track EIPs but apply to areas other than the Ethereum protocol itself. They may propose an implementation, but not to Ethereum's codebase; they often require community consensus; unlike Informational EIPs, they are more than recommendations, and users are typically not free to ignore them. Examples include procedures, guidelines, changes to the decision-making process, and changes to the tools or environment used in Ethereum development. Any meta-EIP is also considered a Process EIP.
+Describes any change that affects most or all Ethereum implementations, such as—a change to the network protocol, a change in block or transaction validity rules, proposed application standards/conventions, or any change or addition that affects the interoperability of applications using Ethereum. Standards Track EIPs consist of three parts—a design document, an implementation, and (if warranted) an update to the [formal specification](https://github.com/ethereum/yellowpaper). Furthermore, Standards Track EIPs can be broken down into the following categories:
```
lightclient, line 26:
```suggestion
Describes any change that affects most or all Ethereum implementations, such as—a change to the network protocol, a change in block or transaction validity rules, proposed application standards/conventions, or any change or addition that affects the interoperability of applications using Ethereum. Standards Track EIPs consist of three parts—a design document, an implementation, and (if warranted) an update to the [formal specification](https://github.com/ethereum/execution-specs). Furthermore, Standards Track EIPs can be broken down into the following categories:
```
😉

EIPS/eip-1.md
```diff
@@ -19,17 +19,37 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
...
 
-- An **Informational EIP** describes an Ethereum design issue, or provides general guidelines or information to the Ethereum community, but does not propose a new feature. Informational EIPs do not necessarily represent Ethereum community consensus or a recommendation, so users and implementers are free to ignore Informational EIPs or follow their advice.
+#### Core
+
+Improvements requiring a consensus fork (e.g. [EIP-5](./eip-5.md), [EIP-101](./eip-101.md)), as well as changes that are not necessarily consensus critical but may be relevant to [“core dev” discussions](https://github.com/ethereum/pm) (for example, [EIP-90], and the miner/node strategy changes 2, 3, and 4 of [EIP-86](./eip-86.md)).
+]
+#### Networking
```
lightclient, line 32: Didn't you delete this and interface eips above?

EIPS/eip-1.md
```diff
@@ -67,54 +87,88 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+The EIP Author(s) have withdrawn the proposed EIP. This state has finality and can no longer be resurrected using this EIP number. If the idea is pursued at later date it is considered a new proposal.
 
-**Living** - A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1.
+#### Living
+
+A special status for EIPs that are designed to be continually updated and not reach a state of finality. This includes most notably EIP-1.
+
+### Technical Assistants
+
+Core EIPs are intensely scrutinized by the client teams and other developers for technical soundness, but the EIP Editors may still lack relevant expertise. And the numerous ERCs cover a wide and growing range of applications that are even more likely to be outside the expertise of any Editor.  Historically, this has put a strain on the Editors' responsibility to maintain quality.  Therefore, the Editors *at their discretion* may require Technical Assistants with the relevant expertise who can review the Author's Idea and work with the Author to ensure a technically sound proposal.
```
lightclient, line 136: This is a good idea, but I think we've tried on a few occasions to activate these people, but have failed.

EIPS/eip-1.md
```diff
@@ -67,54 +87,88 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
 
 ## What belongs in a successful EIP?
 
-Each EIP should have the following parts:
+### EIP Sections
+
+Each EIP should have the following sectio:
```
lightclient, line 142:
```suggestion
Each EIP should have the following sections:
```

EIPS/eip-1.md
```diff
@@ -492,3 +554,4 @@ This document was derived heavily from [Bitcoin's BIP-0001](https://github.com/b
 ## Copyright
 
 Copyright and related rights waived via [CC0](../LICENSE.md).
+](https://github.com/ethereum/EIPs/pull/7230)
```
lightclient, line 557:
```suggestion
```

EIPS/eip-1.md
gcolvin: @xinbenlv think we need make it a solid policy and try again. We've also failed to get enough new Editors to handle all of the proposals we get.  I think part of the difficulty is that proposals span too wide of a range of technical expertise for a small group of Editors.  Many publishers handle this by weeding out articles via peer review before editing even starts.

EIPS/eip-1.md
xinbenlv: I am ok that any changing status shall be a separate PR @Pandapip1

gcolvin: Actually, this might conflict with enforcing EIPW only at status changes.

xinbenlv: I think there is value running EIPW on non-status PRs for the lines being changed.

gcolvin: What has bitten me in the past is the rules changing in the course of drafting an EIP.  I'd much rather see warnings that I can plan to deal with later rather than hard errors that force me to deal with them right away.  Once it comes time to move to the next phase then indeed it's time to get everything up to the latest standards.

EIPS/eip-1.md
```diff
@@ -67,54 +87,112 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+#### Idea
 
-**Draft** - The first formally tracked stage of an EIP in development. An EIP is merged by an EIP Editor into the EIP repository when properly formatted.
+An idea that is pre-draft. This is not tracked within the EIP Repository.
 
-**Review** - An EIP Author marks an EIP as ready for and requesting Peer Review.
+#### Draft
 
-**Last Call** - This is the final review window for an EIP before moving to `Final`. An EIP editor will assign `Last Call` status and set a review end date (`last-call-deadline`), typically 14 days later.
+The first formally tracked stage of an EIP in development. An EIP is merged by an EIP Editor into the EIP repository when it meets some initial requirements: It should be technically sound, the Title and Preamble must be correct, and all included Sections should be readable.  Problems with language like spelling, grammar, markup, etc. are only considered warnings.
```
SamWilsn, line 103: I would drop the "technically sound" requirement, and expand a bit:

```suggestion
Draft is the first formally tracked stage of an EIP in development. An EIP is merged by an EIP Editor into the EIP repository when it meets some initial requirements: the title and preamble must be complete and descriptive, and required sections must be present and readable (but not necessarily complete.)  Problems with language like spelling, grammar, etc. are only considered warnings.

The specification itself does not need to be completely fleshed out, but should contain enough content to give a good idea of the intent of the proposal. Importantly, the proposal must be a good fit for the EIP process, as defined by the EIP Editors.
```

EIPS/eip-1.md
```diff
@@ -67,54 +87,112 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
 
-If this period results in necessary normative changes it will revert the EIP to `Review`.
+#### Review
 
-**Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
+An EIP Author marks an EIP as ready for and requesting Peer Review.  An Editor will merge the EIP when it meets all EIP requirements: it should be a technically sound, complete specification, properly formatted, with correct language, markup and external references.
```
SamWilsn, line 107:
```suggestion
An EIP Author marks an EIP as ready for and requesting Peer Review.  An Editor will merge the status change when the proposal meets all EIP requirements: it should be a complete specification, properly formatted, with correct language, markup and external references.
```

EIPS/eip-1.md
```diff
@@ -67,54 +87,112 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+#### Last Call
+
+This is the final review window for an EIP before moving to `Final`. An EIP editor will assign `Last Call` status and set a review end date (`last-call-deadline`), typically 14 days later.
+
+*If this period results in necessary normative changes it will revert the EIP to `Review`.*
```
SamWilsn, line 113:
```suggestion
This is the final review window for an EIP before moving to `Final`. An EIP Author may request `Last Call` status and set a review end date (`last-call-deadline`), typically 14 days later.

*If this period results in necessary normative changes it will revert the EIP to `Review`.*
```

EIPS/eip-1.md
```diff
@@ -67,54 +87,112 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+#### Final
+
+This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
 
 A PR moving an EIP from Last Call to Final SHOULD contain no changes other than the status update. Any content or editorial proposed change SHOULD be separate from this status-updating PR and committed prior to it.
 
-**Stagnant** - Any EIP in `Draft` or `Review` or `Last Call` if inactive for a period of 6 months or greater is moved to `Stagnant`. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to `Draft` or it's earlier status. If not resurrected, a proposal may stay forever in this status.
+#### Stagnant
+
+Any EIP in `Draft` or `Review` or `Last Call` if inactive for a period of 6 months or greater is moved to `Stagnant`. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to `Draft` or it's earlier status. If not resurrected, a proposal may stay forever in this status.
```
SamWilsn, line 123:
```suggestion
Any EIP in `Draft` or `Review` or `Last Call` if inactive for a period of 6 months or greater is moved to `Stagnant`. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to `Draft` or its earlier status. If not resurrected, a proposal may stay forever in this status.
```

EIPS/eip-1.md
```diff
@@ -67,54 +87,112 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+#### Security Considerations
+
+All EIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life-cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. EIP submissions missing the "Security Considerations" section will be rejected. An EIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.
+
+#### Normative References
+
+**Informative References**
+
+Complete references for each external resource must be included in one of these two sections, as specified below.
```
SamWilsn, line 185: Perhaps something like this:

```suggestion
#### References

**Normative References**

**Informative References**

Complete references for each external resource must be included in one of these two sections, as specified below.  
```

EIPS/eip-1.md
```diff
@@ -168,37 +246,45 @@ if neither the email address nor the GitHub username are given.
...
-## Linking to External Resources
+### Linking to External Resources
 
 Other than the specific exceptions listed below, links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
 
 The process governing permitted external resources is described in [EIP-5757](./eip-5757.md).
 
-### Execution Client Specifications
+External resources not so permitted **MAY** be included *at the Editors' discretion*.
+
```
SamWilsn, line 280: This is already permitted by the "**SHOULD NOT**" above.

```suggestion
```

EIPS/eip-1.md
```diff
@@ -168,37 +246,45 @@ if neither the email address nor the GitHub username are given.
...
+A complete reference to every external resource **MUST** be included in a Normative or Informative References section.
+
+We take several precautions to ensure the continued accessibility of all external resources.  First, we follow the requirements of the [RFC 7322 RFC Style Guide](https://www.rfc-editor.org/rfc/rfc7322.html): references **MUST** be full citations, including author(s), title, and publication information -- URLs are acceptable, but **MUST NOT** be the sole information provided for a reference.  In addition, Links **SHOULD** meet the Requirements for Origins in [EIP-5757](./eip-5757.md), and a Digital Object Identifier (DOI), where available, **MUST** be included. Within "fair use", short quotations from referenced works are encouraged, as embedded quotations will not change.
```
SamWilsn, line 283: I'd much prefer we use a computer tractable citation format.

There is an example of how `csl-json` looks when rendered in [EIP-6963](https://eips.ethereum.org/EIPS/eip-6963#fn:rfc4122), and [the relevant source](https://github.com/ethereum/EIPs/blob/e4cb6fda2825de5e8afbe27d33789f9558faaf8d/EIPS/eip-6963.md?plain=1#L248-L280).

You can convert between several well known formats (like BibTeX) here: https://citation.js.org/demo/ (choose CSL-JSON as output format.)

> URLs are acceptable, but **MUST NOT** be the sole information provided for a reference

Why not? If the URL contains a git commit hash or points to an exact document (like links to RFCs), I don't see why the URL alone is insufficient. We're an internet-first standards body. If anyone prints EIPs, I'd be quite surprised.

EIPS/eip-1.md
```diff
@@ -168,37 +246,45 @@ if neither the email address nor the GitHub username are given.
...
+
+According to the IESG Statement on [Normative and Informative References](https://www.ietf.org/about/groups/iesg/statements/normative-informative-references/), Normative references "specify documents that must be read to understand or implement the technology" whereas an Informative reference "only provides additional information".  The Specification section of an EIP **MUST NOT** contain Informative references, and **SHOULD** contain all necessary Normative references. Motivation and Rationale sections **MAY** include Informative references.
```
SamWilsn, line 285: We do already have this distinction for EIPs (the `requires` preamble field.)

If the Specification section can only contain normative references, then I think we can infer that references in that section are normative. Similarly, if normative references are not permitted in non-Specification sections, we can infer that all references in those section are informative.

My point being: why have specific sections for normative and informative references? We can let authors use normal markdown links (and `csl-json` footnote blobs) to create references, and their location within the document decides whether they're normative or informative.

SamWilsn (commented):
Overall I think this is a generally positive change to EIP-1, and alongside #7047, makes some good changes to the EIPs repository.

That said, I'd like to know what _specific_ changes to `eipw` and `eip-review-bot` you'd expect to fall out of the changes in this document?

EIPS/eip-1.md
gcolvin: I didn't change the "technically sound" part, and am inclined to keep it.  Otherwise I'm fine with your suggesting.

EIPS/eip-1.md
gcolvin: Still wanting to keep "technically sound".

EIPS/eip-1.md
gcolvin: We don't need them, you are right.  I took the one part from IETF before I wrote this part and didn't notice that I'd made it obsolete.   (And thanks for all your reviews here!)

EIPS/eip-1.md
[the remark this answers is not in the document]
gcolvin: Yeah, but too late now.  My apologies.

EIPS/eip-1.md
gcolvin: I'm not sure why you want to change this.  The Author may request Last Call, but it seems like an Editors job to make the move.  But I'm not  that concerned either way.

EIPS/eip-1.md
gcolvin: I'd also prefer a computer-tractable citation format, but didn't try to specify one here.  For the URLs which we have "blessed" we can probably generate citations automatically.
 
 I took need for a non-URL reference from the IETF, which is the original internet-first standards body, and they apply the rule to their own standards.  As ever, the Editors can make exceptions if there is no other choice, but most documents at least have authors and titles.   Given that (and any other publication info) it is much easier to find a copy if the link goes dead.  And I don't see requiring an URL where a document is published but doesn't have a stable URL.  But I'm not a stickler on details here -- just the policy of "as complete a citation as possible."

EIPS/eip-1.md
gcolvin: Also, I'm not sure what "a good fit for the EIP process" means.  I wrote "must actually be a proposal to improve Ethereum", but I'm not sure if that's better.

SamWilsn: I'd like to include the rule somewhere that editors can reject things that have little business being standards.

For example, we might reject:

 * `EIP-XXXX: Solidity Style Guide` because it isn't a technical specification, 
 * `EIP-XXXX: An implementation of secp256r1` because it doesn't define how two or more parties interoperate, or
 * `EIP-XXXX: A comprehensive table of precompile addresses` because it can never go to final.

I'm not sure how to write that concisely without unduly limiting editor discretion though.

EIPS/eip-1.md
SamWilsn: I'm not qualified to judge technical soundness for all EIPs. Maybe we go with a lower bar of plausibly sound?

EIPS/eip-1.md
SamWilsn: I've always operated under the assumption that authors request a status change, and editors grant it.

I'm concerned that the wording "[a]n EIP editor will assign `Last Call` status" makes it sound like authors don't need to open a pull request. Without a pull request, I don't see how a proposal can move to last call.

EIPS/eip-1.md
SamWilsn: I feel pretty strongly that someone with only an internet connection and a web browser should be able to view an EIP and all documents referenced therein.

EIPS/eip-1.md
```diff
@@ -67,54 +87,114 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+RFC 822 style headers containing metadata about the EIP, including the EIP number, a short descriptive title (limited to a maximum of 44 characters), a description (limited to a maximum of 140 characters), and the author details. Irrespective of the category, the title and description should not include EIP number. See [below](./eip-1.md#eip-header-preamble) for details.
 
-## EIP Formats and Templates
+#### Abstract
+
+A multi-sentence (short paragraph) technical summary. This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does.
+
+#### Motivation *(optional)*
+
+A motivation section is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. This section may be omitted if the motivation is evident.
```
kdenhartog, line 157: Would it be useful for these two to be merged? In the EIPs I've read through often times they re-iterate the same information for the most part.

EIPS/eip-1.md
```diff
@@ -168,37 +248,45 @@ if neither the email address nor the GitHub username are given.
 
 At least one author must use a GitHub username, in order to get notified on change requests and have the capability to approve or reject them.
 
-### `discussions-to` header
+#### `discussions-to` header
```
kdenhartog, line 251: I've seen a few RFCs that have been moved through the process utilizing alternative channels of communication such as Telegram and Discord essentially subverting the intent of this header. Would it be useful to allow this to link to where the group is actually communicating to be more conducive for open communication? For example, one I'd really like to see done is to be able to link to an alternative GH repository where a spec is actually iterated on. The value of this is that it allows for separate issue tracking to make sure changes are getting implemented in the spec and Async debate to occur. Furthermore, it can link to authoritative places of discussion such as a Discord server or a Zoom link where synchronous discussions regularly occur via the Readme so that the discussions can be more transparent.

If we move in this direction we can also iterate on this process further so that the concept of Rings can be more formalized (e.g. form a ETH-Core-Devs org to iterate on Core specs) such that it makes it more known where the peer review and the experts are actually communicating.

EIPS/eip-1.md
kdenhartog: For reference, CSL did create some confusion so I moved the linking to just the RFC link here. I like the way the CSL is getting rendered, but the way the IETF handles normative referencing seems to also work. If there's a way to easily generate a CSL-JSON object from a URL that would be ideal. I couldn't figure this out with the demo link provided above @SamWilsn.

kdenhartog:
I think this is moving in the right direction, but I still think we should push for more formal structuring of WG's. Whether that is done in this iteration of updates to EIP-1 or in a subsequent one doesn't much matter to me.

EIPS/eip-1.md
bumblefudge: huge +1 to "warning not error/merge-blocker"

EIPS/eip-1.md
bumblefudge: some kind of nominal honorarium or [on-chain?] badge of honor for participating never hurts to make folks feel appreciated for this "unpaid overtime".  i also think there's a cultural change where having engineers from your *competitors*  reviewing and approving your EIPs could really be a prestige/maturity symbol! in any case, I am a huge +1 to formalizing recognition of review and getting external/horizontal/non-editor review.

EIPS/eip-1.md
bumblefudge: ^ I tend to agree with Sam's point, that:
- never being finalizable 
- not improving interop at any level of the EVM stack (incl wallet-to-wallet, for ex)
and
- not improving the functioning of clients
should all be valid reasons for saying no.  i defer to the group where/how best to insert these restrictions, but that certainly seems a good starting point for acceptance criteria.

Sidenote: by both *my*  version of Sam's three rules and his, an INFORMATIONAL EIP defining EVM "compatibility" could be debatably in or out of scope.  That implies to me that the scope of "interop" should perhaps be more explicit, so it's less subjective whether defining EVM (at least for 2023) is a good EIP document or not :D

EIPS/eip-1.md
bumblefudge: Yeah, Technical Assistants could help editors feel qualified/informed enough to make this assessment, but delegating that assesment to the consensus of a WG (however tightly or loosely defined) of, say, 2 or more competing implementers and 1 or more editors-or-technical-assistants sounds like a great option if it's on the table!

EIPS/eip-1.md
bumblefudge: if part of the problem is public perception of editorial omnipotence, it never hurts to *phrase* things less editor-centrically, as an author-push not an editor-pull :D

EIPS/eip-1.md
SamWilsn: https://citation.js.org/demo/ is how I format it. Doesn't automatically support pulling from a URL, but you can get, for example, bibtex and convert it to CSL-JSON.

shemnon:
While this EIP addresses many unrelated pain points, it does not address the core problem that Ethereum Protocol upgrades and application layer ERC specifications are serving two increasingly divergent and differentiated communities.  A lowest common denominator one-size-fits-all process is leading to dissatisfaction, disenfranchisement, and departures from people and projects who should be using a separated EIP and ERC process.  

Forcing two different processes to use a single process prevents high-signal process changes that can serve only one community.  Core Protocol changes have only a limited number of software codebases that can deonstrate a proof of concept, but ERC changes have no such restriction, so a "working prototype" requirement would only serve to hamper team alignment in the ERC process because a team could simply abandon the ERC process and ship their prototype.  Such an option does not exist in the consensus heavy core protocol changes.

The core protocol has also adopted a parallel "viability" status for EIPs to reflect their inclusion in an upcoming fork.  Hence a review prior to CFI status would react differently that a call for review once it's considered for a fork.  The former review step would ask questions about the utility and appropriateness for protocol, whereas the latter review step would focus almost entirely on hammering out unclear portions of the specification.  [EIP-6780](https://eips.ethereum.org/EIPS/eip-6780) is a good example of how this two step review was beneficial.  A process that forces two communities that have almost completely diverged prevents any serious discussion of changing the lifecycle because questions of "how would ERCs do this" will stall the progress.

The proposed creation of working groups underscores this fundamental difference, and moves the EIP process further down the path of serving neither community well.  For Protocol changes targeting a hard fork, the WG is AllCoreDevs, and viability for review, last call, and accepted will be reflected almost entirely by the actions of that group.  ERC change, instead, need to gather together a working group whose credibility may or may not be accepted by the rest of the ERC community.  This working group step will work radically different between the two groups, further demonstrating their divergent needs.

This PR is being presented as an alternative to splitting the repos, but it is mostly addressing tangential issues.  The linking policy in particular is 100% disconnected from the split community issue, and feels like it is being offered as a compensation for splitting.  

What should have been an amicable split has now become a contested divorce.

kdenhartog:
> For Protocol changes targeting a hard fork, the WG is AllCoreDevs, and viability for review, last call, and accepted will be reflected almost entirely by the actions of that group.

Would it not be useful to also have a broad community review beyond this WG? It seems under the proposed solution for completely separating the work is to propose that problems get solved in a vacuum to a degree. Changes to the consensus layer are bound to impact others within the community beyond the core developers so should they also not have a say in objecting or appealing the decisions that are proposed by the WG?

The biggest problem today is that we've created a governance pattern of vetocracy, where if an editor opposes a change within the EIP-1 process a single veto can nix forward progress. What it sounds like you're suggesting is closer in line with what many other SDOs have done which is more in line with a technocracy. The one key difference that I don't think we're appreciating the value of within this community is a proper set of checks and balances with hard power. Put another way, if we're not an editor there's no way for someone to stick their hand up and say "hey I don't think that should be a standard and here's why" today. We need that checks and balances via an appeals process where anyone can present and argument and the merits of the argument will be heard. Without that, and by extension the idea of completely splitting, will produce more conflict during the times when there's tradeoffs made across these clean separated lines. This will naturally lead to factions without a common goal which is antithetical to the goals of this community in my opinion.

bumblefudge:
I think Kyle is pointing out

> This working group step will work radically different between the two groups, further demonstrating their divergent needs.

I think Kyle makes a good point-- WGs are one possible set of solutions to the general problem of how to get substantial and wide review (wider and more substantial than the editors can be expected to bring).  If the processes are responding to diff community needs, it make sense that different kinds of review would be facilitated differently.  Perhaps the AllCoreDevs already works well for garnering horizontal (internal) review, and there the pain point (partic if a different set of editors is looking at proposals) is more along the lines Kyle was hinting at-- people downstream of core changes with app-level or DeFi/economic input, to identify tradeoffs that core devs don't see.  It would be a win if this PR helped identify divergent requirements for two distinct processes!

EIPS/eip-1.md
gcolvin: I'm resolving this unchanged for now.  Outside review is helpful and often necessary, but in the end the Editors have to make the call.

EIPS/eip-1.md
gcolvin: Hmm.  I think the "MAY" mostly eases the tone of the "SHOULD NOT", so not completely redundant.

EIPS/eip-1.md
gcolvin: > I feel pretty strongly that someone with only an internet connection and a web browser should be able to view an EIP and all documents referenced therein.

I agree, they should, especially for normative links.  They should have enough information  to find the document in a library, and for a successful internet search if the link goes dead.

EIPS/eip-1.md
```diff
@@ -19,17 +19,37 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
...
+
+#### Networking
+
+Includes improvements around [devp2p](https://github.com/ethereum/devp2p/blob/readme-spec-links/rlpx.md) ([EIP-8](./eip-8.md)) and [Light Ethereum Subprotocol](https://ethereum.org/en/developers/docs/nodes-and-clients/#light-node), as well as proposed improvements to network protocol specifications of [whisper](https://github.com/ethereum/go-ethereum/issues/16013#issuecomment-364639309) and [swarm](https://github.com/ethereum/go-ethereum/pull/2959).
+
+#### Interface
+
+Includes improvements around client [API/RPC](https://github.com/ethereum/execution-apis#README) specifications and standards, and also certain language-level standards like method names ([EIP-6](./eip-6.md)) and [contract ABIs](https://docs.soliditylang.org/en/develop/abi-spec.html). The label “interface” aligns with the [interfaces repo] and discussion should primarily occur in that repository before an EIP is submitted to the EIPs repository.
```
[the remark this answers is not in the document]
gcolvin: I'm not sure, but I'm not proposing any changes here.

EIPS/eip-1.md
```diff
@@ -67,54 +87,114 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
 ![EIP Status Diagram](../assets/eip-1/EIP-process-update.jpg)
 
-**Idea** - An idea that is pre-draft. This is not tracked within the EIP Repository.
+#### Idea
 
-**Draft** - The first formally tracked stage of an EIP in development. An EIP is merged by an EIP Editor into the EIP repository when properly formatted.
+An idea that is pre-draft. This is not tracked within the EIP Repository.
 
-**Review** - An EIP Author marks an EIP as ready for and requesting Peer Review.
+#### Draft
```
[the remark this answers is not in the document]
gcolvin: It makes sense, but outside of Core EIPs that hasn't really happened.  I'm proposing Technical Assistants (or the like) as a workable alternative.

EIPS/eip-1.md
```diff
@@ -67,54 +87,114 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+#### Last Call
+
+This is the final review window for an EIP before moving to `Final`. An EIP editor will assign `Last Call` status and set a review end date (`last-call-deadline`), typically 14 days later.
+
+*If this period results in necessary normative changes it will revert the EIP to `Review`.*
+
+#### Final
+
+This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
```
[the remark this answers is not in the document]
gcolvin: Yeah, we've been missing an Errata section for a long time.

EIPS/eip-1.md
gcolvin: It might sometimes, but I'm not proposing a change here.  (The Abstract really should be brief, whereas the Motivation can take as much space as it needs.)

EIPS/eip-1.md
gcolvin: I've also seen cases where the Magician's link points off to other places where discussion is happening.  Does that suffice?

gcolvin:
> The biggest problem today is that we've created a governance pattern of vetocracy, where if an editor opposes a change within the EIP-1 process a single veto can nix forward progress.

Yes, that is a problem with any consensus-based approach.  What you really want is unanimity, when you can't get it you have a problem.  If something needs doing and no compromise can be found then the group may need move on despite the dissent, and the dissenter will have to decide whether it's a hill worth dying on.  I'm currently creating such a problem myself, and am trying to do so in good faith.

gcolvin:
> 

> I'd like to know what _specific_ changes to `eipw` and `eip-review-bot` you'd expect to fall out of the changes in this document?

I sorry, but I honestly don't know enough about them to tell you.

gcolvin:
> While this EIP addresses many unrelated pain points, it does not address the core problem that Ethereum Protocol upgrades and application layer ERC specifications are serving two increasingly divergent and differentiated communities.

This is true.  As you've noted elsewhere it's more of a start at clearing out pain points that _don't _ have to do with that problem and apply to all kinds of EIPs.

I've decided not to engage right here and now with the bulk of your thoughtful comments, as we are mostly discussing them on your own https://github.com/ethereum/EIPs/pull/7329.  I'll keep them in mind as a work on the next draft of this, and will come back to them here if I'm not sure I'm addressing them.  Thanks.

shemnon:
On "my own" #7329?  I only wrote it because you demanded it, to be clear.

EIPS/eip-1.md
gcolvin: Still, I agree that I'd rather see the bulk of the discussion happen here or on the Magicians where the history can be maintained.

gcolvin:
> On "my own" #7329? I only wrote it because you demanded it, to be clear.

And I very much appreciate that you did.  Thank you.

EIPS/eip-1.md
gcolvin: Expecting laurels for service to Ethereum?!

gcolvin:
> I think Kyle is pointing out
> 
> > This working group step will work radically different between the two groups, further demonstrating their divergent needs.

I'm suggesting that, in effect, the Core Developers already are a Working Group, and we would do well to treat them like one.

EIPS/eip-1.md
bumblefudge: > This is a good idea, but I think we've tried on a few occasions to activate these people, but have failed.

I was not paying attention when these occasions came and went, does anyone have links to catch me up on these attempts?

EIPS/eip-1.md
```diff
@@ -19,29 +20,59 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
...
 
-If a **Core** EIP mentions or proposes changes to the EVM (Ethereum Virtual Machine), it should refer to the instructions by their mnemonics and define the opcodes of those mnemonics at least once. A preferred way is the following:
+Many proposals will come into EIP process via working groups who take responsibility for their specifications.  Working Group concerns include the technical quality, formatting and notations for their specifications, coordination with other specifications, the workflow for their tasks and documents, and addditional repos for tracking their work.
 
-```
-REVERT (0xfe)
-```
+* Working group proposals first enter the EIP process (see below) as **Drafts**.
+* While WG proposals are **Drafts** their Specification and workflow belongs to the WG.
+* WG EIPs reenter the EIP process to become **Final**.
```
Pandapip1, line 65:
```suggestion
* Working group proposals are merged into the EIP repository with a status of `Draft`.
* As long as a working group EIP has a status of `Draft`, every member of the working group should have the same permissions as a co-author.
* With approval from an EIP editor, an EIP's working group can move that EIP's status from `Draft` to `Final`.
```
Is this wording better?

EIPS/eip-1.md
```diff
@@ -67,54 +98,120 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
-Each EIP should have the following parts:
+### EIP Sections
+
+*Note that Working Group EIPs may have additional sections at the WG's discretion, as may any EIP at the Editors' discretion.*
+
+Each EIP should have the following sections:
+
+#### Preamble
+
+RFC 822 style headers containing metadata about the EIP, including the EIP number, a short descriptive title (limited to a maximum of 44 characters), a description (limited to a maximum of 140 characters), and the author details. Irrespective of the category, the title and description should not include EIP number. See [below](./eip-1.md#eip-header-preamble) for details.
```
Pandapip1, line 166:
```suggestion
RFC 822 style headers containing metadata about the EIP, including the EIP number, a short descriptive title (limited to a maximum of 44 characters), a description (limited to a maximum of 140 characters), and the author details. Irrespective of the category, the title and description should not include the EIP number. See [below](#eip-header-preamble) for details.
```

EIPS/eip-1.md
```diff
@@ -67,54 +98,120 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+#### Abstract
+
+A multi-sentence (short paragraph) technical summary. This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does.
+
+#### Motivation *(optional)*
```
Pandapip1, line 172:
```suggestion
#### Motivation

This section is optional.
```

EIPS/eip-1.md
```diff
@@ -67,54 +98,120 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+A motivation section is critical for EIPs that want to change the Ethereum protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the EIP solves. This section may be omitted if the motivation is evident.
+
+#### Specification
+
+The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Ethereum platforms (besu, erigon, ethereumjs, go-ethereum, nethermind, or others).
+
+Working groups MAY specify additional requirements, including on the form of the Specification and the implementation of the technology.
+
+#### Rationale *(optional)*
```
Pandapip1, line 182:
```suggestion
#### Rationale

This section is optional.
```

EIPS/eip-1.md
```diff
@@ -67,54 +98,120 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale should discuss important objections or concerns raised during discussion around the EIP.
+
+#### Backwards Compatibility *(optional)*
```
Pandapip1, line 186:
```suggestion
#### Backwards Compatibility

This section is optional for all EIPs except those with a status of `Last Call` or `Final`.
```

EIPS/eip-1.md
```diff
@@ -67,54 +98,120 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
-- Copyright Waiver - All EIPs must be in the public domain. The copyright waiver MUST link to the license file and use the following wording: `Copyright and related rights waived via [CC0](../LICENSE.md).`
+All EIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their consequences. The EIP must explain how the author proposes to deal with these incompatibilities. This section may be omitted if the proposal does not introduce any backwards incompatibilities, but this section must be included if backward incompatibilities exist.
 
-## EIP Formats and Templates
+#### Test Cases *(optional)*
```
Pandapip1, line 190:
```suggestion
#### Test Cases

This section is optional for all non-Core EIPs.
```

EIPS/eip-1.md
```diff
@@ -67,54 +98,120 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+Test cases for an implementation are mandatory for EIPs that are affecting consensus changes. Tests should either be inlined in the EIP as data (such as input/expected output pairs, or included in `../assets/eip-###/<filename>`. This section may be omitted for non-Core proposals.
+
+#### Reference Implementation *(optional)*
```
Pandapip1, line 194:
```suggestion
#### Reference Implementation

This section is optional.
```

EIPS/eip-1.md
```diff
@@ -67,54 +98,120 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+An optional section that contains a reference/example implementation that people can use to assist in understanding or implementing this specification. This section may be omitted for many EIPs.  Most ERCs should include an implementation. Final Core EIPs should link to the relevant Execution Client or Consensus Client commits that implement the EIP.
+
+#### Security Considerations
+
+All EIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life-cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. EIP submissions missing the "Security Considerations" section will be rejected. An EIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.
+
+#### References *(if any)*
```
Pandapip1, line 202:
```suggestion
#### References

This section is optional if there are no references.
```

EIPS/eip-1.md
```diff
@@ -168,37 +271,45 @@ if neither the email address nor the GitHub username are given.
...
 
-## Linking to External Resources
+### Linking to External Resources
 
 Other than the specific exceptions listed below, links to external resources **SHOULD NOT** be included. External resources may disappear, move, or change unexpectedly.
 
 The process governing permitted external resources is described in [EIP-5757](./eip-5757.md).
 
-### Execution Client Specifications
+External resources not so permitted **MAY** be included *at the Editors' discretion*.
```
Pandapip1, line 304:
```suggestion
Other external resources may be allowed at Editor discretion.
```

EIPS/eip-1.md
```diff
@@ -447,7 +558,7 @@ For each new EIP that comes in, an editor does the following:
 - The title should accurately describe the content.
 - Check the EIP for language (spelling, grammar, sentence structure, etc.), markup (GitHub flavored Markdown), code style
 
-If the EIP isn't ready, the editor will send it back to the author for revision, with specific instructions.
+If the EIP isn't ready, the editor will send it back to the author for revision, with specific instructions and possibly a request for review by Technical Peers.
```
Pandapip1, line 561:
```suggestion
If the EIP isn't ready, the editor will send it back to the author for revision, with specific instructions.
```
This is already encompassed by "with specific instructions."

Pandapip1 (changes requested):

EIPS/eip-1.md
```diff
@@ -19,29 +20,59 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
 
 ## EIP Types
 
-There are three types of EIP:
+There are three types of EIP: Standards Track, Meta and Informational.
```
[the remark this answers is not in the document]
gcolvin: It might. It might also be worth pulling all of the EIPs together into one working group-- there aren't that many Networking and Interface proposals, they tend to be more closely related to Core work than ERCs, (some require client changes?) and the core devs are probably more competent than most to review them.

EIPS/eip-1.md
```diff
@@ -19,29 +20,59 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
...
+Many proposals will come into EIP process via working groups who take responsibility for their specifications.  Working Group concerns include the technical quality, formatting and notations for their specifications, coordination with other specifications, the workflow for their tasks and documents, and addditional repos for tracking their work.
 
-```
-REVERT (0xfe)
-```
+* Working group proposals first enter the EIP process (see below) as **Drafts**.
+* While WG proposals are **Drafts** their Specification and workflow belongs to the WG.
+* WG EIPs reenter the EIP process to become **Final**.
+
+Currently, only the Core Developers are organized enough to already be doing this job.
```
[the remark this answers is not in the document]
gcolvin: I'm willing to kick that can down the road until we have other people who want to be a working group -- by then we will have more experience.

As laid out here any group of people can self-organize as a working group and start vetting proposals according to whatever process and standards they like.  If other people prefer to submit proposals directly to the Editors that is fine too, but if the Editors judge that a proposal needs more technical review a relevant working group would be an obvious choice.  So a working group would succeed to the extent that people found them a useful means of creating EIPs.

The Core Devs are special in that nothing goes onto the mainnet unless they put it there, so there wouldn't be much point in trying to go around them.

EIPS/eip-1.md
```diff
@@ -19,29 +20,59 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
...
+
+### Core EIPs
+
+***The Specification of Core EIPs is the responsibility of the Core Developers***, who the Editors treat as an independent Working Group.
```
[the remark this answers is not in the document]
gcolvin: That would be fine.  Main issue is who would be the members of the groups.  The Core Devs already exist as a cohesive organization that wants more say over the Core EIP process, so I'm proposing to recognize that fact and give them more discretion.

EIPS/eip-1.md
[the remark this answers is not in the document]
gcolvin: If we do that, Victor and I have considered creating an ERC working group, if he can find a few other people to help.  If we can work out how to set up repos for each group something close to a split could be realized without an actual bifurcation.

EIPS/eip-1.md
gcolvin: Well, formally it is, but a reminder seems in order.  I'll reconsider the wording.

EIPS/eip-1.md
gcolvin: A little better, but I really meant that "While WG proposals are **Drafts** their Specification and workflow belongs to the WG."  So it's up to the WG as to whether "every member of the working group should have the same permissions as a co-author."  Also up to the WG what it takes for them approve a PR from Draft to Final.  From there it's up to the Editors what our process is for accepting the PR, and I haven't changed that here.  I'd expect only minor editorial stuff would be appropriate at that point.

EIPS/eip-1.md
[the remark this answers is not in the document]
Pandapip1: I would be happy to join an ERC WG.

EIPS/eip-1.md
Pandapip1: How about this then:

```
As long as a working group EIP has a status of `Draft`, the working group itself should be treated as if it were the sole author of the EIP for any process described in EIP-1 that requires author approval.
```

Pandapip1 (approved):
+1. I would slightly prefer for WGs to be able to define their own statuses.

EIPS/eip-1.md
gcolvin: Hmm.  I think if a WG wants to be a group author then they can without our help.  I can imagine scenarios where an author of an EIP makes a PR of their proposal to Final over the objections of the WG, and we merge it before becoming aware of the controversy, but don't see a need to spell that out here.  If we were to get into the nitty-gritty I'd suggest that at least one WG member should volunteer as an Editor who needs to approve PRs on that WG's EIPs.

gcolvin:
> +1. I would slightly prefer for WGs to be able to define their own statuses.

My intent in adding these fields to the Preamble was to support that:

> `working-group`: The name of the Working Group responsible for this EIP (If any.)
> `working-group-draft`: A link to the Working Group draft of this EIP if not in etherem/EIP/EIPs. (Optional, if any.)
> `working-group-status`: The status of this EIP in the Working Group workflow. (Optional, if any.)
>
> Working Groups can introduce their own headers, which should come last.

gcolvin:
@Pandapip1  I'm almost certain I resolved all of your suggestions on Optional sections by accepting them, but Github disagrees.  Not normative, but I'll get back to doing it the hard way at some point.

Pandapip1:
> @Pandapip1 I'm almost certain I resolved all of your suggestions on Optional sections by accepting them, but Github disagrees. Not normative, but I'll get back to doing it the hard way at some point.

Github doesn't disagree on my end. They appear as resolved.

gcolvin:
> > @Pandapip1 I'm almost certain I resolved all of your suggestions on Optional sections by accepting them, but Github disagrees. Not normative, but I'll get back to doing it the hard way at some point.
> 
> Github doesn't disagree on my end. They appear as resolved.

Yes, they are resolved, but they don't contain your changes.

Pandapip1:
For me they do appear to have committed just fine

gcolvin:
> For me they do appear to have committed just fine
I don't understand at all then.  But I just went to this PR on the Web and used Github's own editor and cleaned up all the old-style Optional annotations I still found there. (And then had to undo the damage of an ill-advised search-and-replace.)

EIPS/eip-1.md
```diff
@@ -363,8 +490,8 @@ This is a sentence with a footnote.[^1]
 
 Which renders to:
 
-<!-- markdownlint-capture -->
-<!-- markdownlint-disable code-block-style -->
+<!--- markdownlint-capture --->
+<!--- markdownlint-disable code-block-style --->
```
SamWilsn, line 494: Why the extra dash?

gcolvin: The dashes are from an ill-advised global search-and-replace.  I've fixed them, but the bot, in it's infinite wisdom, suddenly decided that the DOI example -- which has been broken since 2022-- must now be fixed before I can check in my changes.  And since I am a second-class editor I can't override it.

gcolvin: Following is the error that just popped up.  I cannot find any difference to explain why it suddenly complained.

gcolvin:
```
This is a sentence with a footnote.[^1]

[^1]:
    ```csl-json
    {
 Check failure on line 500 in EIPS/eip-1.md
GitHub
Actions
/ Markdown Linter

Code block style [Expected: fenced; Actual: indented]

EIPS/eip-1.md:500 MD046/code-block-style Code block style [Expected: fenced; Actual: indented]
      "type": "article",
      "id": 1,
      "author": [
```

EIPS/eip-1.md
```diff
@@ -395,17 +522,17 @@ This is a sentence with a footnote.[^1]
     }
     ```
 
-<!-- markdownlint-restore -->
+<!--- markdownlint-restore --->
```
Pandapip1, line 525: Is there a reason you changed this?

EIPS/eip-1.md
Pandapip1: It's always complained about this.

Pandapip1: Also, the markdownlint bot isn't a required check.

gcolvin: OK, thanks.  It does seem to be generating bad output at the end of the document, and I don't know how I broke it.

EIPS/eip-1.md
gcolvin: I agree with where this is going, but would rather not resolve it just yet.

EIPS/eip-1.md
```diff
@@ -4,6 +4,7 @@ title: EIP Purpose and Guidelines
 status: Living
 type: Meta
 author: Martin Becze <mb@ethereum.org>, Hudson Jameson <hudson@ethereum.org>, et al.
+discussions-to: https://ethereum-magicians.org/t/update-eip-1-eip-pain-relief-7230/15082
```
Pandapip1, line 7: Is this meant to be here?

EIPS/eip-1.md
```diff
@@ -19,29 +20,61 @@ For Ethereum implementers, EIPs are a convenient way to track the progress of th
...
+
+Currently, only the Core Developers are organized enough to do this job.
+
+### Core EIPs
+
+***The Specification of Core EIPs is the responsibility of the Core Developers***, who the Editors treat as an independent Working Group.
+
+### Editorial Services
+
+The Editors' job, for individual authors and for working groups, is to work with the authors to ensure that proposals merged to EIPs/EIPS meet standards of quality, format, style, and usage, and to provide administrative and publication services for EIPs and related digital assets.
```
Pandapip1, line 77:
```suggestion
The editors' job, for individual authors and for working groups, is to work with the authors to ensure that proposals merged to EIPs/EIPS meet standards of quality, format, style, and usage, and to provide administrative and publication services for EIPs and related digital assets.
```
Editor here is not a proper noun

EIPS/eip-1.md
```diff
@@ -67,54 +100,132 @@ These calls generally result in a "rough consensus" around what EIPs should be i
 
 *In short, your role as the champion is to write the EIP using the style and format described below, shepherd the discussions in the appropriate forums, and build community consensus around the idea.*
 
+
 ### EIP Process
 
-The following is the standardization process for all EIPs in all tracks:
+The following is the full, default standardization process for all EIPs in all tracks.  The diagram shows all of the  stages a document can be in and the transitions beteen those states.
```
Pandapip1, line 106:
```suggestion
The following is the full, default standardization process for all EIPs in all tracks. The diagram shows all of the  stages a document can be in and the transitions between those states. 
```
Typos

EIPS/eip-1.md
```diff
@@ -67,54 +100,132 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
 ### EIP Process
 
-The following is the standardization process for all EIPs in all tracks:
+The following is the full, default standardization process for all EIPs in all tracks.  The diagram shows all of the  stages a document can be in and the transitions beteen those states. 
 
 ![EIP Status Diagram](../assets/eip-1/EIP-process-update.jpg)
 
-**Idea** - An idea that is pre-draft. This is not tracked within the EIP Repository.
+***For Working Groups only Draft, Final, and Withdrawn are used in the Preamble as a `status`.  While a document is a Draft the working group is responsible for its process, which the group can track as `working-group-status`.***
```
Pandapip1, line 110:
```suggestion
***For Working Groups only Draft, Final, and Withdrawn are used in the Preamble as a `status`. While a document is a Draft the working group is responsible for its process, which the group can track as `working-group-status`.***
```

EIPS/eip-1.md
```diff
@@ -67,54 +100,132 @@ These calls generally result in a "rough consensus" around what EIPs should be i
...
+
+All other proposals will move through all stages of this process.
```
Pandapip1, line 112:
```suggestion
All other proposals can move through all stages of this process.
```

Pandapip1 (changes requested):

EIPS/eip-1.md
```diff
@@ -419,41 +546,41 @@ If you are interested in assuming ownership of an EIP, send a message asking to
...
 
-- Read the EIP to check if it is ready: sound and complete. The ideas must make technical sense, even if they don't seem likely to get to final status.
-- The title should accurately describe the content.
-- Check the EIP for language (spelling, grammar, sentence structure, etc.), markup (GitHub flavored Markdown), code style
+* Read the EIP to check if it is ready: sound and complete. The ideas must make technical sense, even if they don't seem likely to get to final status.
+* The title should accurately describe the content.
+* Check the EIP for language (spelling, grammar, sentence structure, etc.), markup (GitHub flavored Markdown), code style
 
-If the EIP isn't ready, the editor will send it back to the author for revision, with specific instructions.
+If the EIP isn't ready, the editor will send it back to the author with instructions for revision and possibly a request for review by Technical Peers.
```
minimalsm, line 577:
```suggestion
If the EIP isn't ready, the editor will send it back to the author with instructions for revision and possibly a request for review by Technical Peers.
```

kdenhartog: +1 to the suggestion. Would it make sense to get more specific and recommend a particular WG that they get review from as well?

Karlitos88 (changes requested):
TheNewFactory@Art&Contents.Red 
Jcarlossantosferreiros77@gmail.com 
JUANCARLOSS1977 
Solicitar información adicional

EIPS/eip-1.md
Karlitos88: genial laborioso

PR #7233, opened 2023-06-26
Update EIP-5793: change size type to prevent leading zeros

MariusVanDerWijden:
changes the type of the size to P which implies non-leading zeros as discussed with @tbenr


fab-10 (approved):
LGTM

tbenr:
@MariusVanDerWijden you might have referenced the wrong handle, I doubt you discussed this with me

g11tech:
Should we also include the change in how the types are serialized in flat format as per the current implementations ?

PR #7241, opened 2023-06-27
Update EIP-663: Replace runtime stack checks with EOF validation rules

gumb0:


EIPS/eip-663.md
```diff
@@ -74,10 +71,10 @@ This has no effect on backwards compatibility because the opcodes were not previ
 
 For `0 <= n <= 255`:
 
-    - `DUPN n` to fail if `stack_height < n`.
-    - `SWAPN n` to fail if `stack_height < (n + 1)`.
-    - `DUPN n` to fail if `stack_height + 1 > 1024`.
-    - `DUPN n` and `SWAPN n` to fail if gas available is less than 3.
+    - `DUPN n` to fail validation if `stack_height < n`.
```
hugo-dc, line 74: It's a bit confusing using variable `n` here.

For example, if `n=0`:

- `DUPN 0` to fail validation if `stack_height < 0`.

It is defined above that `n` is `imm+1`, but in the same line `n` looks like it should be the same value. Maybe it could be changed to something like:

For `0 <= i<= 255`:

- `DUPN i` to fail validation if `stack_height < i+1`.

Where `i` represents the immediate.

gumb0: It was a mistake not to change the line with `0 <= n <= 255`, changed now, please check.

axic (approved):

PR #7256, opened 2023-06-29
Update EIP-6963: Update rDNS and icon properties and improve security considerations

kdenhartog:
This PR makes the following updates:

1. Add the optional rDNS property in order for DApps to optimistically detect specific wallets
2. Updates the normative requirements around the `icon` property
3. Further fleshes out the details of the security and privacy considerations


jxom (approved):

EIPS/eip-6963.md
```diff
@@ -77,7 +77,7 @@ ipfs://QmZo7gsostaUdrV1peFz4cX6Z2TgriFJJP4VahG5knVm29
 https://ethereum.org/static/f541df14fca86543040c113725b5bd1a/99bcf/metamask.webp
 ```
 
-Additionally the image must be squared with 96x96px minimum resolution. Image format is recommended to be lossless like PNG and WebP or alternatively vectorial like SVG. We strongly discourage lossy formats like JPG/JPEG.
+The `icon` string MUST be a data URI as defined in [^RFC 2397]. The image SHOULD be squared with 96x96px minimum resolution. The image format is RECOMMENDED to be either lossless or vector based such as PNG, WebP or SVG to make the image easy to render on the DApp. Since SVG images can execute Javascript, applications and libraries MUST render SVG images using the `<img>` tag to ensure no untrusted Javascript execution can occur. The DApp SHOULD reject providers from being an available option with an icon using a lossy formats such as JPG or JPEG in order to prevent pixelation of images.
```
everdimension, line 80: Are we forbidding src URLs?

Rejecting based on lossy format seems a weird recommendation. While pixelation is mostly probably undesired, one could argue it could even be a purposeful artistic element.
And also we can't really say anything about image quality based only on its format.

EIPS/eip-6963.md
```diff
@@ -234,11 +235,21 @@ function onPageLoad(): EIP6963ProviderDetail[] => {
...
 ## Security Considerations
 
-The security considerations of [EIP-1193](./eip-1193.md) apply to this EIP.
+### EIP-1193 Security considerations
+
+The security considerations of [EIP-1193](./eip-1193.md) apply to this EIP. Implementers are expected to consider and follow the guidance of the providers they're utilizing as well.
+
+### Prototype Pollution of Wallet Provider objects
+
+Browser extensions, and therefore Wallet extensions, are able to modify the contents of the page and the Provider object by design. Given the provider objects of various Wallets are considered high trusted interface to communicate transaction data. In order to prevent the page or various other extensions from modifying the interaction between the DApp and the Wallet in an unexpected way. The expected way to handle this is to utilize `object.freeze()` on the  `EIP1193Provider` object before the wallet dispatches the `eip6963:announceProvider` Event.
```
everdimension, line 244: Might be a wording issue, the second and the third sentences seem unfinished

EIPS/eip-6963.md
```diff
@@ -234,11 +235,21 @@ function onPageLoad(): EIP6963ProviderDetail[] => {
...
+One advantage to the concurrency Event loop utilized by this design is that it operates in a manner where either the DApp or the Wallet can initiate the flow to announce a provider. For this reason, Wallet implementers can now consider whether or not they wish to announce themselves to all pages or attempt alternative means in order to reduce the ability for a user to be fingerprinted by the injection of the `window.ethereum` object. Some examples, of alternative flows to consider would be to wait to inject the provider object until the DApp has announced the `eip6963:requestProvider`. At that point, the wallet can initiate a UI consent flow to ask the user if they would like to share their wallet address. This allows for the Wallet to enable the option of a "private connect" feature. However, if this approach is taken Wallet's must also consider how they intend to support backwards compatability with a DApp that does not support this EIP.
```
everdimension, line 252: This is an awesome side-feature, but I'm still concerned as to how we educate dapps to make sure they react to the announcements that are emitted later than on the initial page load phase.

We can imagine a wallet exposing itself not after the `eip6963:requestProvider` event, but after a deliberate user action provided by wallet UI, e.g. a dialog confirming with the user whether they want to announce themselves to the current website.

kdenhartog: > This is an awesome side-feature, but I'm still concerned as to how we educate dapps to make sure they react to the announcements that are emitted later than on the initial page load phase. 

With the new concurrent model, DApps will automatically behave in this way by registering their event handler on page load and then firing the `eip6963:requestProvider` event afterwards. As long as that event handler is registered by the DApp then it is able to respond accordingly. This means that in some cases DApp's may still prompt the user to click the `connect` button after already providing consent and intiating the flow in the wallet, which will feel a bit clunky but it will work. Alternatively, the DApp's could optimistically fire the `requestAccounts` call right as soon as a provider is chosen in which case because the user already granted the consent it would skip prompting the user and return the wallet address.

This would also work in the same way with the legacy flow. For example, under the current presumed flow here's how it works:

1. User clicks connect button which calls `async window.ethereum.request({ method: 'eth_requestAccounts' })`
2. User consents to the account being provided
3. Wallet supplies the address back in a promise structure during success or promise reject when no consent is given

Under the new model, here's how it would work:

1. User clicks "inject" button (not the best name for UI) in wallet to inject it and selects the address
2. User clicks "connect" button on the DApp
3. DApp responds with the wallet address in a promise immediately because the User has already consented

In the case where step 1 is skipped the wallet would not connect in this new "Private" mode. That's what makes this an implementation considerations is because both flows can be executed and supported using the same interface, but the implementation logic differs, defaults can differ, and configuration logic can differ but we can support the new and the old flow using this pattern AND potentially address fingerprinting concerns if implemented this way.

Does that make sense?

EIPS/eip-6963.md
```diff
@@ -46,6 +46,7 @@ Provider Discovery Library: A library or piece of software that assists a DApp t
 Each wallet provider will be announced with the following interface `EIP6963ProviderInfo`. The values in the `EIP6963ProviderInfo` MUST be used as follows:
 
 - **`uuid`** - a globally unique of the wallet provider that MUST be (UUIDv4[^rfc4122] compliant) to uniquely distinguish different [EIP-1193](./eip-1193.md) provider sessions. The cryptographic uniqueness provided by UUIDv4[^rfc4122] guarantees that two independent `EIP6963ProviderInfo` objects can be separately identified.
+- **`rdns`** - an optional identifier that is used to optimistically identify the wallet by the Decentralized Application. If supplied by the wallet it MUST utilize an reverse DNS notation identifier that matches the domain name of the wallet provider.
```
pedrouid, line 49: Move this below "name" and "icon" since they are required and this one is optional

EIPS/eip-6963.md
```diff
@@ -55,15 +56,14 @@ Each wallet provider will be announced with the following interface `EIP6963Prov
  */
 interface EIP6963ProviderInfo {
   uuid: string;
+  rdns: string;
```
pedrouid, line 59: This makes it optional for TS interfaces

```suggestion
  rdns?: string;
```

And we should also move this one below "name" and "icon" in the interface

EIPS/eip-6963.md
pedrouid: Yeah I agree. We should make it REQUIRED that a Data URI is used and also make it RECOMMENDED that SVG or lossless format is used

But rejecting a provider for using a lossy format like JPEG seems to extreme

EIPS/eip-6963.md
```diff
@@ -193,6 +193,7 @@ function onPageLoad() {
   function announceProvider() {
     const info: EIP6963ProviderInfo = {
       uuid: "350670db-19fa-4704-a166-e52e178b59d2",
+      rdns: "com.example.wallet",
```
pedrouid, line 196: Again 😅 can we put it below "name" and "icon"

pedrouid (changes requested):
This is looking really good 👌 just made my final request for changes

EIPS/eip-6963.md
jxom: Wonder if this should be worded as RECOMMENDED (but optional) to make my life easier so us library authors don’t have to follow up with wallets because their wallet doesn’t work with RainbowKit, ConnectKit, etc…

jxom: Won’t die on the hill though, but it’s just a caution. I’m grateful for this being here in the first place 😂

EIPS/eip-6963.md
```diff
@@ -46,6 +46,7 @@ Provider Discovery Library: A library or piece of software that assists a DApp t
 Each wallet provider will be announced with the following interface `EIP6963ProviderInfo`. The values in the `EIP6963ProviderInfo` MUST be used as follows:
 
 - **`uuid`** - a globally unique of the wallet provider that MUST be (UUIDv4[^rfc4122] compliant) to uniquely distinguish different [EIP-1193](./eip-1193.md) provider sessions. The cryptographic uniqueness provided by UUIDv4[^rfc4122] guarantees that two independent `EIP6963ProviderInfo` objects can be separately identified.
```
bumblefudge, line 48:
```suggestion
- **`uuid`** - a globally unique identifier for the wallet provider that MUST be (UUIDv4[^rfc4122] compliant) to uniquely distinguish different [EIP-1193](./eip-1193.md) provider sessions. The cryptographic uniqueness provided by UUIDv4[^rfc4122] guarantees that two independent `EIP6963ProviderInfo` objects can be separately identified.
```

EIPS/eip-6963.md
bumblefudge: Review closely, I added a lot of detail here resolving as many requests/comments as possible!
```suggestion
- **`rdns`** - a RECOMMENDED optional string that is used by the Decentralized Application to optimistically identify the kind of wallet being injected. If supplied by the wallet, it MUST contain a valid lowercase domain name in reverse order (e.g. `tld.domain.subdomain`) that is under the control of the developer of that wallet. It SHOULD NOT be supplied by any other party.
```

EIPS/eip-6963.md
bumblefudge: agree moving the optional property to the end would be clearer, would do in both places

EIPS/eip-6963.md
```diff
@@ -55,15 +56,14 @@ Each wallet provider will be announced with the following interface `EIP6963Prov
...
   name: string;
   icon: string;
 }
 ```
 
 #### Images/Icons
 
-// TODO for @kdenhartog: This section should be updated to require support of SVGs instead of arbitrary data URIs or URLs that point to images.
-
 A uri encoded image was chosen to enable flexibility for multiple protocols for fetching and rendering icons, for example:
```
bumblefudge, line 67:
```suggestion
A URI-encoded image was chosen to enable flexibility for multiple protocols for fetching and rendering icons, for example:
```

EIPS/eip-6963.md
bumblefudge:
```suggestion
The `icon` string MUST be a data URI as defined in [^RFC 2397]. The image SHOULD be squared with 96x96px minimum resolution. The image format is RECOMMENDED to be either lossless or vector based such as PNG, WebP or SVG to make the image easy to render on the DApp. Since SVG images can execute Javascript, applications and libraries MUST render SVG images using the `<img>` tag to ensure no untrusted Javascript execution can occur. The DApp MAY reject providers from being an available option with an icon using a lossy formats such as JPG or JPEG in order to prevent pixelation of images.
```

EIPS/eip-6963.md
pedrouid: As long as it's defined as OPTIONAL then I don't see an issue to also mention that it's RECOMMENDED

bumblefudge (commented):
LGTM

EIPS/eip-6963.md
bumblefudge: reworded in my other PR

bumblefudge: rewording suggestion in my other PR

EIPS/eip-6963.md
kdenhartog: RECOMMENDED is equivalent to SHOULD in RFC 2119. OPTIONAL is equivalent to MAY. My opinion is that we should stick with this being a MAY. For more rationale why I'm of this opinion see the final two paragraphs of this https://github.com/ethereum/EIPs/pull/7167#issuecomment-1619313354

EIPS/eip-6963.md
kdenhartog: Approving, but I'm hesitant that we should actually be defining the interface with a specific programming language. Unfortunately, it's too much of a headache to get the rendering working properly with something like [WebIDL](https://developer.mozilla.org/en-US/docs/Glossary/WebIDL) so this will have to be "good enough".

EIPS/eip-6963.md
kdenhartog: > Review closely, I added a lot of detail here resolving as many requests/comments as possibleI

I'm a +1 to this overall with the exception of using RECOMMENDED. Can you edit it to say `a string that MAY be used...`

kdenhartog:
In reference to this comment from @pedrouid https://github.com/ethereum/EIPs/pull/7256#discussion_r1248755024

Yeah, I'm good supporting JPEG and keeping it as a SHOULD NOT. Main reason I did that is because using too much non-deterministic normative language like SHOULD/SHOULD NOT etc can lead to interop bugs very easily so I prefer going with MUST/MUST NOT whenever possible.

EIPS/eip-6963.md
jxom: I agree with "MAY" implying that `rdns` may be used by the DApp, but I definitely think that we should include a "SHOULD" for Wallet Providers implementing `rdns`: "a RECOMMENDED string that MAY be used by the Decentralized Application".

If one of the top Wallets do not implement `rdns`, this will seriously jeopardize the connection UX of the DApp if they are using a wallet connection library. 

To put as an analogy, WalletConnect v2 has the concept of "optional chains", which only just a month before v1 cutoff, only one wallet had implemented the "optional chains" implementation. Optional chains are important, as libraries need it to create a working multichain experience. As a result, WalletConnect is advocating for libraries to lean more on the "optional chains" implementation. Can get into more detail about that, but just want to get to the point: if we are labelling this property with an "OPTIONAL" term, then I can see myself right now scrambling to get all these wallets to implement `rdns` (just like optional chains in WCv2) so that they can work with wagmi, RainbowKit, ConnectKit, Web3Modal, etc...

I am okay with this being an optional property (still feel strongly about required but if the consensus thinks otherwise, i concur), but at least, we need to label it as "a string that SHOULD be provided..."

bumblefudge: I think Kyle is right there that we SHOULD use all these capitalized verbs in the RFC2119 sense, not in the contemporary colloquial English usage of the words :D 

Here's the RFC2119 version of the edit:

```suggestion
- **`rdns`** - an OPTIONAL string that is used by the Decentralized Application to optimistically identify the kind of wallet being injected. It MAY be supplied by the wallet to facilitate smoother user experience but SHOULD NOT be supplied by any other party. If supplied by the wallet, it MUST contain only a valid lowercase domain name in reverse order (e.g. `com.walletbrand.extension`) that is under the control of the developer of that wallet.
```

bumblefudge: Or, if we wanted to signal to the wallet developer that this OPTIONAL isn't really realistically optional, we can sneak in a little rhetorical oomph in modifying clauses:

```suggestion
- **`rdns`** - an OPTIONAL string that is used by the Decentralized Application to optimistically identify the kind of wallet being injected to enable better end-user experience. It MAY be supplied by the wallet to opt in to this mechanism but SHOULD NOT be supplied by any other party. If supplied by the wallet, it MUST contain only a valid lowercase domain name in reverse order (e.g. `com.walletbrand.extension`) that is under the control of the developer of that wallet.
```

everdimension, line 49: > _to optimistically identify the kind of wallet being injected_

I'm totally confused by the concept here. Can you please explain what info can be _optimistically_ obtained from an RDN property?

The announced wallet provider already contains _name_, _icon_ and an Ethereum provider object, which is everything that is needed for a dapp to connect to this wallet and interact with it.
More so, this data is available immediately as the event is received by the dapp. I do not understand what "optimistic" mechanism is being made available here.

EIPS/eip-6963.md
everdimension: The event system makes perfect sense, I'm just beware of cases where dapps "bail out" of listening after hearing first wallet announcement or once hearing a "desired" wallet announced.

EIPS/eip-6963.md
everdimension, line 49: Should we explicitly disambiguate `rdns`?
Funnily, it's not that obvious

Possible disambiguations:
* RDN — Reverse Domain Notation
* RDN — Reverse Domain Name
* RDNS — Reverse DNS (as in Reverse Domain Name System)
* RDNS — Reverse Domain Name Scheme (as referred to in this document: https://gcc.gnu.org/legacy-ml/java/2000-04/msg00095.html)

jxom: They are needed for “connect wallet” libraries like Web3Modal, RainbowKit and ConnectKit to achieve a robust and non-buggy experience, something that has been very hard to manage with flags on the EIP-1193 provider.

https://github.com/ethereum/EIPs/pull/7167/files#diff-927530d6d340560a972fe4fc1b91d7833e7a959089cc9c9d2baee54dd128c410R237

bumblefudge: 'Optimistically' wasn't my word, but I interpreted it as a reference to the leap of faith that the wallet features are [still] the features the dapp thinks that vendor provides? In the sense that the legacy, unverifiable `isMetamask` flags can only be trusted by an optimist on a happy path😅.

Would it still be an "optimistic" identification process if the DNS string were a valid one under the control of the wallet vendor? In the telegram group there was some speculation that this could be an extension point for verifiable identification, publishing proof-of-control of, say, a package manager or app-store identifier at that address...

bumblefudge: I tried doing that above:
https://github.com/ethereum/EIPs/pull/7256#discussion_r1251624311
the RFC for domain names actually includes a section on reverse-ordered domains, which I think makes more sense than referencing the application-manifest tradition from Java onward (which, though popularized by Apple, was never actually *standardized* as such... making it more of a convention uniting various context-specific syntaxes than a universal, IETF-style syntax!)

kdenhartog: > In the sense that the legacy, unverifiable isMetamask flags can only be trusted by an optimist on a happy path

Yes, this is the intention behind that word when I added it in there.

The `providerId` and `rdns` both act as identifiers for previously interacted wallets, but only in the case where wallets behave as expected. In the following events the expected behavior is undefined:

1. When the user deletes their wallet and switches so the wallet is no longer available
2. When the wallet changes the identifier used
3. When the wallet doesn't supply a compliant identifier or the identifier is unknown
4. When a wallet overlaps it's identifier (either intentionally or not) with another

In all of these cases, the `providerId` and `rdns` properties will potentially behave in undefined ways according to the spec. These are just the ones that I've come up and there's likely to be more implementation specific ones that arise overtime. Essentially we're leaving this part of the UX up to the DApp connection library implementers (or independent implementers) so if the implementers opt to not handle one of these cases or new implementation specific ones then the UX will behave in undefined ways and produce bugs.

EIPS/eip-6963.md
glitch-txs: TBH I don't think any dapp would reject the provider for this reason. I don't think this would affect the functionality of the provider as well. Is there any other reason for not accepting JPG or JPEG format besides the pixelation issue?

EIPS/eip-6963.md
```diff
@@ -77,7 +78,7 @@ ipfs://QmZo7gsostaUdrV1peFz4cX6Z2TgriFJJP4VahG5knVm29
 https://ethereum.org/static/f541df14fca86543040c113725b5bd1a/99bcf/metamask.webp
 ```
 
-Additionally the image must be squared with 96x96px minimum resolution. Image format is recommended to be lossless like PNG and WebP or alternatively vectorial like SVG. We strongly discourage lossy formats like JPG/JPEG.
+The `icon` string MUST be a data URI as defined in [RFC 2397]. The image SHOULD be squared with 96x96px minimum resolution. The image format is RECOMMENDED to be either lossless or vector based such as PNG, WebP or SVG to make the image easy to render on the DApp. Since SVG images can execute Javascript, applications and libraries MUST render SVG images using the `<img>` tag to ensure no untrusted Javascript execution can occur. The DApp SHOULD reject providers from being an available option with an icon using a lossy formats such as JPG or JPEG in order to prevent pixelation of images.
```
glitch-txs, line 81: Is there any other reason besides pixelation to reject a provider for presenting a lossy format?

```suggestion
The `icon` string MUST be a data URI as defined in [RFC 2397]. The image SHOULD be squared with 96x96px minimum resolution. The image format is RECOMMENDED to be either lossless or vector based such as PNG, WebP or SVG to make the image easy to render on the DApp. Since SVG images can execute Javascript, applications and libraries MUST render SVG images using the `<img>` tag to ensure no untrusted Javascript execution can occur.
```

EIPS/eip-6963.md
kdenhartog: > If one of the top Wallets do not implement rdns, this will seriously jeopardize the connection UX of the DApp if they are using a wallet connection library.

Can you specify how we'll jeopardize the connection UX with greater detail here so I can understand your concerns better? From what I can tell today, eip6963.org is working fine for discovery when I render non-existent wallets without the providerId property. Also, how is rendering useful if the provider object isn't available yet from the wallet? For example, in the video below how would it be useful for the Enkrypt wallet to appear if it's not actually available?

As you can see in this video, even with Enkrypt providing a `providerId` and the other 2 not, we're unable to disambiguate between the two as the spec currently because the implementation isn't error handling for the backwards compatibility aspect. This goes to show this is more about implementations handling different cases properly than it is about interfaces including properties. The issue here is no longer discovery, it's purely about implementation branching logic on the DApp side from what I can tell.

https://github.com/ethereum/EIPs/assets/23125059/2c77007e-605a-48b7-aff2-9d004c764184

jxom: For all the reasons in my original PR. On top of that, without a deterministic identifier, it will not be trivially possible to sync up injected wallets with their corresponding WalletConnect wallet, which would cause duplication there, and is probably one of the more important cases I should have listed in the original PR.

> Also, how is rendering useful if the provider object isn't available yet from the wallet?

Not all wallets inject themselves immediately. Some could inject themselves a few seconds after. A DApp could provide a better experience here by displaying a loading indicator rather than nothing, and confusing the end-user. I thought we discussed this property being an extension point so that DApp libraries can do this? There is an error scenario here regarding wallet removal (an edge case), and that can be addressed accordingly.

kdenhartog:
@bumblefudge I merged in your editorial branch, so I think everything you addressed should be in here now via [`64fcd1b` (#7256)](https://github.com/ethereum/EIPs/pull/7256/commits/64fcd1b137e8c0733ab510ecc340a402bf805393)

kdenhartog:
Given the `rdns` property is the most controversial so far, I'm thinking it would it make sense to split it out from the rest of the changes in this PR so we can continue discussing it. Anyone have concerns with that?

EIPS/eip-6963.md
kdenhartog: Ahh, that's a good point we should probably add something to the text to state that must not be done.

EIPS/eip-6963.md
kdenhartog: > For all the reasons in my original PR.

I believe you're referring to these [here](https://github.com/ethereum/EIPs/pull/7167/files#diff-927530d6d340560a972fe4fc1b91d7833e7a959089cc9c9d2baee54dd128c410R239-R241) is that correct. Can you show me a demo of these issues so we can properly assess the tradeoffs here between including and not including this? I'm hoping we can move past using hypotheticals and let the code determine the direction here. So far, what I'm seeing is that the UX is determined based on implementation choices on the DApp side rather than the inclusion of the `providerId` as I showed in the video above. 

Right now we're talking past each other because I'm coming at this from the perspective of knowing how text can be misinterpreted 4 years down the road and knowing that the spec text can't be updated after this occurs. I think (correct me if I'm wrong) you're coming at this from the perspective that if a bug occurs then people file an issue and all can be fixed. Implementations can be updated easily but specs cannot. Once a spec is finalized it can't be undone or modified. This is the difference between producing an open standard and producing an open source implementation. When you get a spec wrong it means all the legacy stuff has to stay in place which is the reason `eth_signTypedData` has 4 different versions. When an implementation is wrong, it get's fixed and versions get bumped and all is fine. That's why I'm trying to force us to identify and handle all of the bugs now before we finalize the spec.

> On top of that, without a deterministic identifier, it will not be trivially possible to sync up injected wallets with their corresponding WalletConnect wallet

I think that's a feature not a bug since are still two separate Wallets with two independent sets of state (e.g. loading a private key in my Desktop browser does not automatically load it into my mobile wallet). My mobile browser is not the same as my desktop browser either and I need to configure each of them independently. A worse scenario in my eyes is if two independent wallets get mismatched because of one of the errors I highlighted in this [comment](https://github.com/ethereum/EIPs/pull/7256#discussion_r1252359386).

> Not all wallets inject themselves immediately. Some could inject themselves a few seconds after. A DApp could provide a better experience here by displaying a loading indicator (before the provider hydrates) rather than nothing, and confusing the end-user.

Would it not confuse the end user to have a wallet appear that's been deleted as well? For example, today if that happens we automatically render the page to download the wallet, but the code path you're suggesting would assume the wallet is already available so is unlikely to render that page unless we define it to (which I don't think should be the case - that's an implementation detail in my eyes).

everdimension: I guess my issue is with term itself. "Optimistic" conventionally means a situation where a program assumes some state which will be verified or discarded later (usually this means a response from the server which is considered to be source of truth)

In our case, I don't see any of such mechanics.
* If we're talking about feature detection based on rdns string, then this is a terrible idea (see https://webaim.org/blog/user-agent-string-history/) (but I'm not sure we're talking about that)
* If the dapp is going to _verify_ the rdns value by making an async rpc request to the wallet, then the term "optimistic" does make sense. But I've never heard or such rpc api.

kdenhartog: > If we're talking about feature detection based on rdns string, then this is a terrible idea (see [webaim.org/blog/user-agent-string-history](https://webaim.org/blog/user-agent-string-history/)) (but I'm not sure we're talking about that)

Great point here! We are litterally reinventing the UA string and will almost certainly re-encounter all it's problems along the way. My understanding is this is the primary reason that others have been wanting this and this example really exemplifies why this is a bad idea.

glitch-txs: Maybe we could find a better replacement of rdns, but I think making it optional would remove the value it's meant to have

bumblefudge: Well, I think the `isMetamask` flag is a better analogy to UA string, in that it's self-attested and people are highly incentivized to use it. thus my idea of making an optional verification step when it mattered that relied on DNS control...

glitch-txs: We also should consider that this new EIP will offer a new environment when isMetaMask flag won't be as useful as before, why would a wallet impersonate another one when it's going to be recognized by the dapp and be shown as wanted?

bumblefudge: well, today there are good and bad reasons why a non-metamask wallet might set that flag:
- metamask fork or clone has all the same interfaces and UX flows, makes it easy for dapps to integrate by announcing equivalent features
- phishing wallet presents user and dapp alike metamask interface but does something else under the hood (usually bad)

i'm presuming there would be both good and bad use-cases might still apply for `rdns`, but in the latter case some kind of verification extension would prevent it, and in the former case, (which I consider a feature and not a bug for wallet competition, evolution, and adoption), subdomains could work wonders here, as in:
`xyz.upandcoming.compatibilityMode.metamask` . this way, a dapp that already implemented a Metamask-style flow and doesn't want to support upandcoming.xyz's bespoke flow could just match on `compatibilityMode.metamask` and assume the metamask integration will work for that wallet 😎

kdenhartog: > why would a wallet impersonate another one when it's going to be recognized by the dapp and be shown as wanted

That is highly dependent on whether the web page will choose to discriminate based on wallets. One reasoning could be because the wallet doesn't support an RPC call they need as an example and get's published as an IPFS Static DApp so rather than update the DApp, future wallets that add support for the RPC call may be forced to imitate others.

EIPS/eip-6963.md
kdenhartog: Not that I could think of. This is more about just eliminating unnecessary code branches, but I don't think this is likely to cause issues for us anyways.

EIPS/eip-6963.md
kdenhartog: Resolving this, we've opted to go with a REQUIRED usage of this property

EIPS/eip-6963.md
kdenhartog: I'm thinking it's probably fine to accept and discourage, going to modify this text and link to commit.

EIPS/eip-6963.md
kdenhartog: addressed in latest changes, so will be resolving this one.

EIPS/eip-6963.md
kdenhartog: addressed in latest round of changes 👍🏻

EIPS/eip-6963.md
kdenhartog: I did my best to define this a bit further in the latest round of changes without getting to specific here so that we can make this more extensible/specific in the future if necessary. Marking this one as resolved.

EIPS/eip-6963.md
```diff
@@ -13,41 +13,42 @@ requires: 1193
...
 
-- **`uuid`** - a globally unique of the wallet provider that MUST be (UUIDv4[^rfc4122] compliant) to uniquely distinguish different [EIP-1193](./eip-1193.md) provider sessions. The cryptographic uniqueness provided by UUIDv4[^rfc4122] guarantees that two independent `EIP6963ProviderInfo` objects can be separately identified.
-- **`name`** - A human-readable local alias of the wallet provider to be displayed to the user on the DApp. (e.g. `DopeWalletExtension` or `AwesomeWallet`)
-- **`icon`** - A URI[^rfc3986] pointing to an image. Icon images MUST be square with 96x96px minimum resolution
+- **`uuid`** - a globally unique identifier the Wallet Provider that MUST be ([UUIDv4][RFC-4122] compliant) to uniquely distinguish different [EIP-1193](./eip-1193.md) provider sessions that have matching properties defined below during the liftetime of the page. The cryptographic uniqueness provided by [UUIDv4][RFC-4122] guarantees that two independent `EIP6963ProviderInfo` objects can be separately identified.
+- **`name`** - a human-readable local alias of the Wallet Provider to be displayed to the user on the DApp. (e.g. `ExampleWalletExtension` or `AwesomeExampleWallet`)
```
glitch-txs, line 49: Not sure if it's that relevant*
```suggestion
- **`name`** - a human-readable local alias of the Wallet Provider to be displayed to the user on the DApp. (e.g. `Example Wallet Extension` or `Awesome Example Wallet`)
```

EIPS/eip-6963.md
```diff
@@ -100,7 +96,7 @@ To emit events both DApps and Wallets MUST use the `window.dispatchEvent` functi
 
 #### Announce and Request Events
 
-The `EIP6963AnnounceProviderEvent` interface MUST be a `CustomEvent` object with a `type` property containing a string value of `eip6963:announceProvider` and a `detail` property with an object as it's value of type `EIP6963ProviderDetail`. The `EIP6963ProviderDetail` object MUST be frozen by calling `Object.freeze()` on the  value of the `detail` property.
+The `EIP6963AnnounceProviderEvent` interface MUST be a `CustomEvent` object with a `type` property containing a string value of `eip6963:announceProvider` and a `detail` property with an object as it's value of type `EIP6963ProviderDetail`. The `EIP6963ProviderDetail` object MUST be frozen by calling `Object.freeze()` on the value of the `detail` property.
```
```suggestion
The `EIP6963AnnounceProviderEvent` interface MUST be a `CustomEvent` object with a `type` property containing a string value of `eip6963:announceProvider` and a `detail` property with an object value of type `EIP6963ProviderDetail`. The `EIP6963ProviderDetail` object MUST be frozen by calling `Object.freeze()` on the value of the `detail` property.
```

EIPS/eip-6963.md
```diff
@@ -216,7 +212,7 @@ function onPageLoad() {
 
 ### DApp implementation
 
-Here is a reference implementation for a DApp to display and track multiple wallet providers that are injected by browser extensions. 
+Here is a reference implementation for a DApp to display and track multiple Wallet Providers that are injected by browser extensions. 
 
 ```typescript
 function onPageLoad(): EIP6963ProviderDetail[] => {
```
glitch-txs, line 218: There's no actual return here right?

```suggestion
const providers: EIP6963ProviderDetail[];

function onPageLoad() {
```

EIPS/eip-6963.md
```diff
@@ -234,109 +230,32 @@ function onPageLoad(): EIP6963ProviderDetail[] => {
...
+
+### Prevent Wallet Fingerprinting
+
+One advantage to the concurrency Event loop utilized by this design is that it operates in a manner where either the DApp or the Wallet can initiate the flow to announce a provider. For this reason, Wallet implementers can now consider whether or not they wish to announce themselves to all pages or attempt alternative means in order to reduce the ability for a user to be fingerprinted by the injection of the `window.ethereum` object. Some examples, of alternative flows to consider would be to wait to inject the provider object until the DApp has announced the `eip6963:requestProvider`. At that point, the wallet can initiate a UI consent flow to ask the user if they would like to share their wallet address. This allows for the Wallet to enable the option of a "private connect" feature. However, if this approach is taken Wallet's must also consider how they intend to support backwards compatibility with a DApp that does not support this EIP.
```
glitch-txs, line 251: Sharing the *address* would in some cases imply a request of `eth_requestAccounts` method from the DApp right? In this case we are refering to the provider announcement.

I would think this more of a whitelist of trusted sites, so the Wallet doesn't need to ask everytime to announce its provider, although I know this is just an example n.n
```suggestion
One advantage to the concurrency Event loop utilized by this design is that it operates in a manner where either the DApp or the Wallet can initiate the flow to announce a provider. For this reason, Wallet implementers can now consider whether or not they wish to announce themselves to all pages or attempt alternative means in order to reduce the ability for a user to be fingerprinted by the injection of the `window.ethereum` object. Some examples of alternative flows to consider would be to wait to inject the provider object until the DApp has announced the `eip6963:requestProvider`. At that point, the wallet can initiate a UI consent flow to ask the user if they would like to share their wallet provider. This allows for the Wallet to enable the option of a "private connect" feature. However, if this approach is taken Wallet's must also consider how they intend to support backwards compatibility with a DApp that does not support this EIP.
```

pedrouid (approved):
👍

EIPS/eip-6963.md
kdenhartog: Not quite, by only sharing the wallet provider you'd double up the consent flows in an unnecessary way. Essentially what the UI is doing is eliminating the separate consent of injection and wallet address by presuming that if a user is willing to consent to sharing the address that they're fine to also inject the provider. Therefore, the only UI that needs to be initiated is to ask the user which wallet address they'd want to use, when the wallet detects the `requestProvider` Event, and at that point the wallet can both respond with the `announceProvider` and prepare itself to automatically respond to the `eth_requestAccounts` RPC call all in one go without needing a secondary UI consent flow.

kdenhartog: Maybe, this should be updated to say "... share their wallet address which implicitly grants consent to share the Wallet Provider."

glitch-txs: could be! I understand now! thanks for the explanation :)

kvhnuke (approved):

kdenhartog:
merging this as-is. We have a few additional updates being discussed that should be made, but this doesn't need to hold this PR up any longer.

PR #7276, opened 2023-07-04
Update EIP-4337: align `eth_estimateUserOperationGas` result with the bundler spec

kaiix:
According to the definition of [bundler-spec](https://github.com/eth-infinitism/bundler-spec/blob/9cafd4be1d9b21221a57bc06c219a73cbd6cdebd/src/schemas/transaction.yaml#L365), the validation gas result of the `eth_estimateUserOperationGas` should be called `verificationGas` without suffix `Limit`


SamWilsn:

PR #7284, opened 2023-07-06
Update EIP-6492: Handle non-ready deployed wallet

Agusx1211:
Extends EIP-6492 to also handle the case of a wallet being deployed but not ready to sign using EIP-1271.

See: https://ethereum-magicians.org/t/eip-6492-signature-validation-for-pre-deploy-contracts/12903/44?u=agusx1211

### DIFF
--- a/EIPS/eip-6492.md
+++ b/EIPS/eip-6492.md
@@ -63,7 +63,7 @@ Full signature verification MUST be performed in the following order:
 
 - check if the signature ends with magic bytes, in which case do an `eth_call` to a multicall contract that will call the factory first with the `factoryCalldata` and deploy the contract if it isn't already deployed; Then, call `contract.isValidSignature` as usual with the unwrapped signature
 - check if there's contract code at the address. If so perform [ERC-1271](./eip-1271.md) verification as usual by invoking `isValidSignature`
-- if the [ERC-1271](./eip-1271.md) verification fails, and `factory` was skipped because the wallet already had code, execute the `factoryCalldata` transaction and try `isValidSignature` again
+- if the [ERC-1271](./eip-1271.md) verification fails, and the deploy call to the `factory` was skipped due to the wallet already having code, execute the `factoryCalldata` transaction and try `isValidSignature` again
 - if there is no contract code at the address, try `ecrecover` verification
 
 ## Rationale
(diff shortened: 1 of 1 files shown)


EIPS/eip-6492.md
```diff
@@ -135,6 +140,11 @@ contract UniversalSigValidator {
       try IERC1271Wallet(_signer).isValidSignature(_hash, sigToValidate) returns (bytes4 magicValue) {
         bool isValid = magicValue == ERC1271_SUCCESS;
 
+        // retry, but this time assume the prefix is a prepare call
+        if (!isValid && !tryPrepare && contractCodeLen > 0) {
```
Ivshti, line 144: shouldn't this be `!isValid && tryPrepare && contractCodeLen > 0`

EIPS/eip-6492.md
```diff
@@ -62,6 +63,7 @@ Full signature verification MUST be performed in the following order:
 
 - check if the signature ends with magic bytes, in which case do an `eth_call` to a multicall contract that will call the factory first with the `factoryCalldata` and deploy the contract if it isn't already deployed; Then, call `contract.isValidSignature` as usual with the unwrapped signature
 - check if there's contract code at the address. If so perform [ERC-1271](./eip-1271.md) verification as usual by invoking `isValidSignature`
+- if the [ERC-1271](./eip-1271.md) verification fails, and `factory` was skipped because the wallet already had code, execute the `factoryCalldata` transaction and try `isValidSignature` again
```
Ivshti, line 66: yoyu mean "the deploy call to the factory was skipped"?

Ivshti (changes requested):
it's simple enough as implementation, I'm not sure how common/realistic this case is but it doesn't hurt to have the extra flexibility

we need to think of extra security implications, because I think this may increase the attack surface a bit more when using on-chain verification. Keep in mind that `prepareTo` and `prepareCalldata` are malleable and anyone can change them. Same goes for the factory but there's only a single scenario in which the arbitrary call is applicable. Although I think the revert that we do negates that concern?

EIPS/eip-6492.md
Agusx1211: No, because if we didn't `tryPrepare` then we should, but if we did then there isn't much more we could do and the verification is in fact failing

mandervrik (commented):
O

PR #7287, opened 2023-07-06
Update EIP-7201: Make Solidity example compile

frangio:


Ramarti (commented):
There is no need for this line:

`bytes32 location = MAIN_STORAGE_LOCATION;`

Since MAIN_STORAGE_LOCATION is now a hardcoded constant, this should compile

`
assembly {
            $.slot := MAIN_STORAGE_LOCATION
}
`

frangio:
Ah, that's correct, I'll make another PR. Thanks!

PR #7303, opened 2023-07-09
Add EIP: Token-Controlled Token Circulation

kofujimura:

 - The build passes.

### DIFF
--- a/EIPS/eip-7303.md
+++ b/EIPS/eip-7303.md
@@ -8,11 +8,11 @@ status: Draft
 type: Standards Track
 category: ERC
 created: 2023-07-09
-requires: 721
+requires: 721, 1155, 5679
 ---
 ## Abstract
 
-This ERC introduces an access control scheme termed Token-Controlled Token Circulation (TCTC). By representing the privileges associated with a role as an [ERC-721](./eip-721.md) token (referred to as a `control token`), the processes of granting or revoking a role can be facilitated through the minting or burning of the corresponding `control token`. 
+This ERC introduces an access control scheme termed Token-Controlled Token Circulation (TCTC). By representing the privileges associated with a role as an [ERC-721](./eip-721.md) or [ERC-1155](./eip-1155.md) token (referred to as a `control token`), the processes of granting or revoking a role can be facilitated through the minting or burning of the corresponding `control token`. 
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


mandervrik (approved):

EIPS/eip-7303.md
```diff
@@ -0,0 +1,132 @@
...
+     * specified token.
+     * @param r byte32 The role which you want to grant.
+     * @param c address The address of contract ID of which token the user required to own.
+     */
+    function _grantRoleByToken (bytes32 r, address c) internal {
+        _controlTokens[r].push(c);
+    }
+
+    function _checkHasToken (bytes32 r, address u) internal view returns (bool) {
+        for (uint i = 0; i < _controlTokens[r].length; i++) {
```
aaldemir, line 84: It may be more gas efficient to store the array of addresses from `_controlTokens[r]` in memory of the functions scope as opposed to accessing state on each iteration

EIPS/eip-7303.md
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * @notice Grant a role to user who owns a control token specified by the contract ID. 
+     * Multiple calls are allowed, in this case the user must own at least one of the 
+     * specified token.
+     * @param r byte32 The role which you want to grant.
+     * @param c address The address of contract ID of which token the user required to own.
+     */
+    function _grantRoleByToken (bytes32 r, address c) internal {
```
aaldemir, line 79: Should this function call `supportsInterface` on `c` via [EIP-165](https://eips.ethereum.org/EIPS/eip-165) to assert that a non-ERC721 token is not being added to the list?

EIPS/eip-7303.md
```diff
@@ -0,0 +1,132 @@
...
+        _controlTokens[r].push(c);
+    }
+
+    function _checkHasToken (bytes32 r, address u) internal view returns (bool) {
+        for (uint i = 0; i < _controlTokens[r].length; i++) {
+            if (IERC721(_controlTokens[r][i]).balanceOf(u) > 0) return true;
```
aaldemir, line 85: note that ERC20 tokens have the same `balanceOf(address)` interface

EIPS/eip-7303.md
kofujimura: Yes! Thank you for your valuable comments! I've incorporated it into the revised version.

EIPS/eip-7303.md
kofujimura: Yes! Thank you again for your valuable comments! I've incorporated it into the revised version. Is this modification okay?

aaldemir: Yes that change lgtm given it would result in `true` on an OpenZeppelin ERC721 ([code ref](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L53))

EIPS/eip-7303.md
```diff
@@ -0,0 +1,141 @@
...
+eip: 7303
+title: Token-Controlled Token Circulation
+description: Access control scheme based on token ownership.
+author: Ko Fujimura (@kofujimura)
+discussions-to: https://ethereum-magicians.org/t/erc-7303-token-controlled-token-circulation/15020
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-07-09
+requires: 721
```
xinbenlv, line 11:
```suggestion
requires: 721, 5679
```
Since 5679 is being specified in the specification, it will needs to be mentioned in the "requires" field

EIPS/eip-7303.md
```diff
@@ -0,0 +1,141 @@
...
+
+The choice to utilize ERC-721 as the control token for privileges enhances visibility of such privileges within wallets, thus simplifying privilege management for users.
+
+Generally, when realizing privileges as tokens, specifications like Soulbound Token (e.g., [ERC-5192](./eip-5192.md)) are used. Given that ERC-5192 inherits from ERC-721, this ERC has choiced ERC-721 as the requirement for the control token.
+
+Employing a transferable control token can cater to scenarios where role delegation is necessary. For example, when an authority within an organization is replaced or on vacation, the ability to transfer their privileges to another member becomes possible. The decision to designate the control token as transferable will depend on the specific needs of the application.
+
+## Backwards Compatibility
+
+This ERC is designed to be compatible for [ERC-721](./eip-721), [ERC-1155](./eip-1155), and [ERC-5679](./eip-5679) respectively.
```
xinbenlv, line 55: It seems you didn't mention how it can be compatible with ERC-1155. Can you clarify?

EIPS/eip-7303.md
kofujimura: Since the requirement level of ERC-5679 is "SHOULD" in this specification, I did not specify it in this field. But, your suggest is that this is needed even "SHOULD" level, am I right? If so, I will add it.

EIPS/eip-7303.md
kofujimura: Thank you.  I just refer ERC-1155 only because tokens minted using ERC-1155 can be used as control tokens same as ERC-721.  Should I write it explicitly or drop the reference?

EIPS/eip-7303.md
kofujimura: I added 1155 and 5679 in the new version. Thank you.

EIPS/eip-7303.md
kofujimura: I updated the spec to support ERC-1155 as a control token. Thank you.

xinbenlv (approved):

PR #7308, opened 2023-07-11
Update EIP-6780: Add clarifications to EIP-6780

shemnon:
* Contract creation and account creation can occur in separate transactions
* If a contract is not destroyed, any SELFDESTRUCT to self burning will not occur.
* SELFDESTRUCT always halts the current frame, wether in "sweep" mode or legacy mode


EIPS/eip-6780.md
```diff
@@ -27,37 +27,45 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
+Removing "hollow" accounts at addresses that are not externally owned and that only held a balance prior to contract creation provides a better user experience than leaving an account with a non-zero nonce, possible storage, and no code to access that storage.
+
 ## Backwards Compatibility
 
 This EIP requires a hard fork, since it modifies consensus rules.
 
-The only breaking change occurs when a contract is re-created at the same address using `CREATE2` (after a `SELFDESTRUCT`), where the `SELFDESTRUCT` is executed in a transaction that's different from the one that originally creates the contract.
+Contracts that depended on re-deploying contracts at the same address using `CREATE2` (after a `SELFDESTRUCT`) will no longer function properly if the created contract does not call `SELFDESTRUCT` within the current transaction.
+
+Previously it was possible to burn ether by calling `SELFDESTRUCT` targeting the executing contract as the beneficiary.  If the contract existed prior to the transaction the ether will not be burned.  If the contract was newly created in the transaction the ether will be burned, as before.
```
smartprogrammer93, line 60: In the case of the contract existing prior to the transaction. what about gas cost of accessing accounts. if we know that the contract address and the refund address are the same. should we just not touch those addresses and not charge for gas?

shemnon: That's adding in more special cases, for added value I don't see.

We do have to load the account prior to initcode running.  For the case where initcode depends on value to transfer in a call it may not exist in the prior frame's call to the init code, but in the account.

There is also the case where value could be passed in as part of the call as well as being in an existing hollow account.  We would have to touch the account to send the value.

IMHO less rules and letting the interactions shake out where they may land is a better design.  Willing to be overruled.

EIPS/eip-6780.md
```diff
@@ -27,37 +27,46 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
```
maurelian, line 61: Hi from Optimism.
FWIW this change does address our needs. 

The way we burn ETH is to deploy and fund [a contract that instantly SELFDESTRUCTs](https://github.com/ethereum-optimism/optimism/blob/a6e6988cbd89985a72c1fa2fcb9549e6a6ddfbba/packages/contracts-bedrock/contracts/libraries/Burn.sol#L28-L31).

smartprogrammer93: @maurelian, would it be hard to change the mechanism if we decide to disable the burn mechanism?

maurelian: AFAIK there is no other mechanism in the EVM for removing ETH from the state. Yes, Optimism could do something Optimism specific, but that would break our EVM equivalence.

IMO the ability to remove ETH from state is generally valuable, and has been [in use for a long time now](https://twitter.com/jeffehh/status/872871251366334464?s=20).

EIPS/eip-6780.md
```diff
@@ -27,37 +27,46 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
 
- - `SELFDESTRUCT` continues to behave as originally, i.e. deletes all storage keys and the account itself.
- - Subsequently, the account will behave like exactly like an empty account, both in the same transaction and in all later ones
- - Transfer the account balance to the target **and** set account balance to `0.`
- - Note that no refund is given since [EIP-3529](./eip-3529.md).
- - Note that the rules of [EIP-2929](./eip-2929.md) regarding `SELFDESTRUCT` remain unchanged.
- - Note that when verkle tries are implemented on Ethereum, the cleared storage will be marked as having been written before but empty. This leads to no observable differences in EVM execution, but a contract having been created and deleted will lead to different state roots compared to the action not happening.
+   - `SELFDESTRUCT` continues to behave as originally, i.e. deletes all storage keys and the account itself, and execution halts.
+   - Subsequently, the account will behave exactly like an empty account, both in the same transaction and in all later ones.
+   - Transfer the account balance to the target **and** set account balance to `0`.
```
invocamanman, line 40: I have a take about this:
Currently the self destruct does not deletes all storage and account instantly, but the account is registered to be destroyed, and will be at the end of the current transaction.
The described behavior seems that is intended to be instant. 
I would not recommend changing this particular behavior as it enables new ways of using it ( like deploy 2 different contracts in the same address on the same transaction, for example).

As i understand, the rationale is just enabling the current behavior if happens in the same transaction, but this also adds a little tweak to the current behavior.

EIPS/eip-6780.md
shemnon: I think the future will have no burning mechanism, but that depends on what OP advocates for.

EIPS/eip-6780.md
invocamanman: in that case i would add a clarification there about the new instant destruction^^

EIPS/eip-6780.md
smartprogrammer93: @shemnon 
if the core devs agree on disabling the burn mechanism in the future, then we should probably give a signal to the community in this EIP

EIPS/eip-6780.md
smartprogrammer93: I am not entirely sure about the intended behavior. the we implemented it is as described by @invocamanman that it will be registered for destruction at the end of the transaction. if this is not the intended behavior and instant destruction is required. then a clarification is definitely needed.

invocamanman:
So, IMH think that after this EIP self destruct basically will be useful for:
- Burning ether
- Creating contracts on different transactions on the same address that finally get destroyed.

As per the second point, if an account wants to execute certain logic, isn't cheaper to just create a contract and then halt the execution with an STOP for example?  It won't be on the same address every time, but it's way cheaper than Selfdestruct, anyway there are no refunds attached.
My point is that i think there are a very few use cases except for burning ether currently.

EIPS/eip-6780.md
```diff
@@ -27,37 +27,50 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
 1. When `SELFDESTRUCT` is executed in a transaction that is not the same as the contract calling `SELFDESTRUCT` was created:
 
- - `SELFDESTRUCT` does not delete any storage keys or code
- - `SELFDESTRUCT` transfers the entire account balance to the target.
- - No refund is given, as per [EIP-3529](./eip-3529.md).
- - [EIP-2929](./eip-2929.md)'s rules regarding `SELFDESTRUCT` remain unchanged.
+   - The current execution frame halts.
+   - `SELFDESTRUCT` does not delete any data (including storage keys, code, or the account itself).
+   - `SELFDESTRUCT` transfers the entire account balance to the target.
+   - Note that If the target is the same as the contract calling `SELFDESTRUCT` there is no net change in balances.  Unlike the prior specification ethere will not be burnt in this case.
```
dankrad, line 33:
```suggestion
   - Note that if the target is the same as the contract calling `SELFDESTRUCT` there is no net change in balances.  Unlike the prior specification, Ether will not be burnt in this case.
```

EIPS/eip-6780.md
```diff
@@ -27,37 +27,50 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
- - Transfer the account balance to the target **and** set account balance to `0.`
- - Note that no refund is given since [EIP-3529](./eip-3529.md).
- - Note that the rules of [EIP-2929](./eip-2929.md) regarding `SELFDESTRUCT` remain unchanged.
- - Note that when verkle tries are implemented on Ethereum, the cleared storage will be marked as having been written before but empty. This leads to no observable differences in EVM execution, but a contract having been created and deleted will lead to different state roots compared to the action not happening.
+   - `SELFDESTRUCT` continues to behave as previously specified, this includes the following actions
+     - The current execution frame halts.
+     - `SELFDESTRUCT` deletes data as previously specified.
+     - `SELFDESTRUCT` transfers the entire account balance to the target
+     - The account balance of the contact calling `SELFDESTRUCT` is set to `0`.
+   - Note that if the target is the same as the contract calling `SELFDESTRUCT` that ether will be burnt.
```
dankrad, line 44:
```suggestion
   - Note that if the target is the same as the contract calling `SELFDESTRUCT` that Ether will be burnt.
```

EIPS/eip-6780.md
```diff
@@ -27,37 +27,50 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
 
 The `SELFDESTRUCT` opcode remains deprecated as specified in [EIP-6049](./eip-6049.md). Any use in newly deployed contracts is strongly discouraged even if this new behaviour is taken into account, and future changes to the EVM might further reduce the functionality of the opcode.
 
 ## Rationale
 
 Getting rid of the `SELFDESTRUCT` opcode has been considered in the past, and there are currently no strong reasons to use it. This EIP implements a behavior that will attempt to leave some common uses of `SELFDESTRUCT` working, while reducing the complexity of the change on EVM implementations that would come from contract versioning. A further option that was considered was to just remove storage clearing from `SELFDESTRUCT` in [EIP-6046](./eip-6046.md) while preserving removal of contracts, however this is not safe because existing contracts rely on storage being empty when they are deployed.
 
+Removing "hollow" accounts at addresses that are not externally owned and that only held a balance prior to contract creation provides a better user experience than leaving an account with a non-zero nonce, possible storage, and no code to access that storage.
```
dankrad, line 57: I don't understand this sentence. Is it necessary?

EIPS/eip-6780.md
```diff
@@ -27,37 +27,50 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
+
 ## Backwards Compatibility
 
 This EIP requires a hard fork, since it modifies consensus rules.
 
-The only breaking change occurs when a contract is re-created at the same address using `CREATE2` (after a `SELFDESTRUCT`), where the `SELFDESTRUCT` is executed in a transaction that's different from the one that originally creates the contract.
+Contracts that depended on re-deploying contracts at the same address using `CREATE2` (after a `SELFDESTRUCT`) will no longer function properly if the created contract does not call `SELFDESTRUCT` within the current transaction.
```
dankrad, line 63:
```suggestion
Contracts that depended on re-deploying contracts at the same address using `CREATE2` (after a `SELFDESTRUCT`) will no longer function properly if the created contract does not call `SELFDESTRUCT` within the same transaction.
```

EIPS/eip-6780.md
```diff
@@ -27,37 +27,50 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
+Previously it was possible to burn ether by calling `SELFDESTRUCT` targeting the executing contract as the beneficiary.  If the contract existed prior to the transaction the ether will not be burned.  If the contract was newly created in the transaction the ether will be burned, as before.
 
 ## Security Considerations
 
 The following applications of `SELFDESTRUCT` will be broken and applications that use it in this way are not safe anymore:
 
-2. Where `CREATE2` is used to redeploy a contract in the same place in order to make a contract upgradable. This is not supported anymore and [ERC-2535](./eip-2535.md) or other types of proxy contracts should be used instead.
+1. Where `CREATE2` is used to redeploy a contract in the same place in order to make a contract upgradable. This is not supported anymore and [ERC-2535](./eip-2535.md) or other types of proxy contracts should be used instead.
+
+2. Where a contract depended on burning ether via a `SELFDESTRUCT` with the contract as beneficiary, in a contract not created within the transaction.
```
dankrad, line 73:
```suggestion
2. Where a contract depended on burning Ether via a `SELFDESTRUCT` with the contract as beneficiary, in a contract not created within the same transaction.
```

smartprogrammer93:
Why are we not merging this already? clients' implementations match this now

EIPS/eip-6780.md
shemnon: It covers a specific use case, so it is needed.

shemnon:
suggestions merged.  @dankrad @gballet @vbuterin please merge pr into master.

jochem-brouwer (commented):
Two style comments. 

Need to await thoughts about destroying pre-existing accounts (i.e. with balance) regarding Verkle.

EIPS/eip-6780.md
dankrad: I was saying it is quite unintelligible. What's a "hollow account"? Do you mean counterfactual accounts?

Can you maybe try rewriting it because it makes no sense to me at all. Can you explain the use case?

EIPS/eip-6780.md
```diff
@@ -27,37 +27,50 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
 1. When `SELFDESTRUCT` is executed in a transaction that is not the same as the contract calling `SELFDESTRUCT` was created:
 
- - `SELFDESTRUCT` does not delete any storage keys or code
- - `SELFDESTRUCT` transfers the entire account balance to the target.
- - No refund is given, as per [EIP-3529](./eip-3529.md).
- - [EIP-2929](./eip-2929.md)'s rules regarding `SELFDESTRUCT` remain unchanged.
+   - The current execution frame halts.
+   - `SELFDESTRUCT` does not delete any data (including storage keys, code, or the account itself).
+   - `SELFDESTRUCT` transfers the entire account balance to the target.
+   - Note that if the target is the same as the contract calling `SELFDESTRUCT` there is no net change in balances.  Unlike the prior specification, Ether will not be burnt in this case.
```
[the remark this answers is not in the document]
shemnon: Given the confusion we've had in reading it, I think it needs to be called out that this is a change from prior behavior.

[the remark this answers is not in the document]
jochem-brouwer: I misinterpreted and agree that a clarification here is very helpful :smile: :+1:

EIPS/eip-6780.md
shemnon: Deciding wether nor not to completely remove the burn I view as out of scope for the EIP and orthogonal to the issue at hand, so I don't think giving any indication one way or another in the text of the EIP is warranted.

EIPS/eip-6780.md
shemnon: Instant destruction is not required.  I updated the spec to make it clear with "`SELFDESTRUCT` continues to behave as it did prior to this EIP"

EIPS/eip-6780.md
```diff
@@ -27,37 +27,50 @@ The behaviour of `SELFDESTRUCT` is changed in the following way:
...
+   - Note that no refund is given since [EIP-3529](./eip-3529.md).
+   - Note that the rules of [EIP-2929](./eip-2929.md) regarding `SELFDESTRUCT` remain unchanged.
   
-2. When `SELFDESTRUCT` is executed in the same transaction as the contract was created:
+2. When `SELFDESTRUCT` is executed in the same transaction as the contract was created
```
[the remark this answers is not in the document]
shemnon: done

EIPS/eip-6780.md
shemnon: Counterfactual state channels is one use case. It could just as easily be used with smart contract accounts for stealth addresses.  I'll re-write it using counterfactual accounts as the sample case.

dankrad (approved):

PR #7348, opened 2023-07-17
Update EIP-3102: Fix duplicate status

Pandapip1:


Pandapip1:
It's invalid markdown and might cause some of my tools to break (you can see the label bot and eip-review-bot already broke)

PR #7354, opened 2023-07-18
Update EIP-4844: Rename "data gas" to "blob gas"

lightclient:
I think "data gas" is problematic because we already have the concept of "data" in transaction -> calldata. To uninitiated, things like `excess_data_gas` seems to refer to calldata. Of course this isn't the case though. We should improve the naming to be clear what "data gas" is and that it actually refers to "blob data". 

Ultimately I decided to drop "data" almost entirely and refer to gas for blobs as "blob gas". The idea here is that it is gas for paying for blobs. One argument might be that this conflates the idea that a blob might cost a lot (2**17) "blob gas". I think this isn't the case. The "blob" adjective is just a discriminator to note that this gas is typed for blob payment. `GAS_PER_BLOB` further clarifies the unit of gas being a fraction of the total cost of a blob. This is the cleanest approach IMO.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -45,17 +45,17 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `BLOB_COMMITMENT_VERSION_KZG` | `Bytes1(0x01)` |
 | `POINT_EVALUATION_PRECOMPILE_ADDRESS` | `Bytes20(0x0A)` |
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
-| `MAX_DATA_GAS_PER_BLOCK` | `786432` |
-| `TARGET_DATA_GAS_PER_BLOCK` | `393216` |
-| `MIN_DATA_GASPRICE` | `1` |
-| `DATA_GASPRICE_UPDATE_FRACTION` | `3338477` |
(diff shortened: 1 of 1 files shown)


djrtwo:
@adietrichs wanted this term to be more general to the abstract notion of transient data -- in whatever form it might come. Thus the name and how it has it's own min/max and independently a price.

I sympathize with this generality -- blobs are a very specific construct in this pricing regime and not necessarily all types of transient data constructs. If we need to change to disambiguate, I might be in favor of something like DA_GAS

fjl:
So you are arguing for `dataAvailabilityGas`?

lightclient:
I don't think we should be worried too much about a post-blob world with the naming if it actively harms the understandability of the protocol. Blobs are designed specifically to be our "end game" scaling form factor (thus all the work on getting the crypto right). If something else with a slightly different form factor shows up, we should tackle the disambiguation then IMO.

hwwhww:
Note: if we accept it, it requires a corresponding CL specs update.

g11tech (commented):
imo its good since we also renamed opcode from data hash => blob hash :+1:

jflo (approved):

lightclient:
Approved in 4844 breakout call 27.

PR #7366, opened 2023-07-19
Update EIP-6059: Add the missing parameter to the specification

ThunderDeliverer:
During the proposal lifecycle from Draft to Final stage, an additional parameter was added to the `nestTransferFrom` method. This parameter is already taken into account within the `interfaceId`. This means that this PR fixes the inconsistency.

Additionally, the `ChildTransferred` event was missing an important quality of life parameter which doesn’t impact the `interfaceId`, but it does provide better usability of the proposal.

We also expanded the _Rationale_ and _Security consideration_ to be more illustrative of the proposal.

While we understand that finalized proposals are not meant to be updated, we feel like this one might benefit from the correction proposed here. Anyone that uses the proposal in their project either used our open-source library (which already included the proposed changes before the finalization) or the interface provided within the assets (which also includes the proposed changes). This means that modifying the proposal will align it with current implementations.

### DIFF
--- a/EIPS/eip-6059.md
+++ b/EIPS/eip-6059.md
@@ -355,7 +355,7 @@ interface IERC6059 /* is ERC165 */ {
      * @param to Address of the receiving token's collection smart contract
      * @param tokenId ID of the token being transferred
      * @param destinationId ID of the token to receive the token being transferred
-     * @param data Additional data with no specified format, sent in the addChild call
+     * @param data Additional data with no specified format, which SHOULD be sent in the addChild call
      */
     function nestTransferFrom(
         address from,
@@ -455,6 +455,10 @@ This state change places the token in the pending array of the new parent token.
 
 The Nestable token standard has been made compatible with [ERC-721](./eip-721.md) in order to take advantage of the robust tooling available for implementations of ERC-721 and to ensure compatibility with existing ERC-721 infrastructure.
 
+The only incompatibility with ERC-721 is that Nestable tokens cannot use a token ID of 0.
+
+There is some differentiation of how the `ownerOf` method behaves, compared to ERC-721. The `ownerOf` method will now recursively look up through parent tokens until it finds an address which is not an NFT, this is referred to as the *root owner*. Additionally we provide the `directOwnerOf` which returns the most immediate owner of a token using 3 values: the owner address, the tokenId which MUST be 0 if the direct owner is not an NFT, and a flag indicating whether or not the parent is an NFT. In case the token is owned by an EoA or an ERC721Receiver, the `ownerOf` method will behave the same as in ERC-721.
+
 ## Test Cases
 
 Tests are included in [`nestable.ts`](../assets/eip-6059/test/nestable.ts).
@@ -477,6 +481,8 @@ See [`NestableToken.sol`](../assets/eip-6059/contracts/NestableToken.sol).
 The same security considerations as with [ERC-721](./eip-721.md) apply: hidden logic may be present in any of the functions, including burn, add child, accept child, and more.
 
 Since the current owner of the token is allowed to manage the token, there is a possibility that after the parent token is listed for sale, the seller might remove a child token just before before the sale and thus the buyer would not receive the expected child token. This is a risk that is inherent to the design of this standard. Marketplaces should take this into account and provide a way to verify the expected child tokens are present when the parent token is being sold or to guard against such a malicious behaviour in another way.
+
+It is worth noting, that `balanceOf` method only accounts for immediate tokens owned by the address; the tokens that are nested into a token owned by this address will not be reflected in this value as the recursive lookup needed in order to calculate this value is potentially too deep and might break the method.
 
 Caution is advised when dealing with non-audited contracts.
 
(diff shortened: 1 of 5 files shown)


Pandapip1 (dismissed):
+1 because it does seem like this erratas a real mistake
-1 because it modifies a normative section of a Final EIP

Overall, I'm weakly in favor of correcting the mistake.

EIPS/eip-6059.md
```diff
@@ -144,13 +144,16 @@ interface IERC6059 /* is ERC165 */ {
...
+     *  not (`false`)
      */
     event ChildTransferred(
         uint256 indexed tokenId,
         uint256 childIndex,
         address indexed childAddress,
         uint256 indexed childId,
-        bool fromPending
+        bool fromPending,
+        bool toZero
```
lightclient, line 156: this is a breaking change to the api

EIPS/eip-6059.md
```diff
@@ -352,12 +355,14 @@ interface IERC6059 /* is ERC165 */ {
...
      * @param destinationId ID of the token to receive the token being transferred
+     * @param data Additional data with no specified format, which SHOULD be sent in the addChild call
      */
     function nestTransferFrom(
         address from,
         address to,
         uint256 tokenId,
-        uint256 destinationId
+        uint256 destinationId,
+        bytes memory data
```
lightclient, line 365: breaking change

EIPS/eip-6059.md
```diff
@@ -450,6 +455,10 @@ This state change places the token in the pending array of the new parent token.
 
 The Nestable token standard has been made compatible with [ERC-721](./eip-721.md) in order to take advantage of the robust tooling available for implementations of ERC-721 and to ensure compatibility with existing ERC-721 infrastructure.
 
+The only incompatibility with ERC-721 is that Nestable tokens cannot use a token ID of 0.
```
lightclient, line 458: remove

EIPS/eip-6059.md
```diff
@@ -473,6 +482,8 @@ The same security considerations as with [ERC-721](./eip-721.md) apply: hidden l
 
 Since the current owner of the token is allowed to manage the token, there is a possibility that after the parent token is listed for sale, the seller might remove a child token just before before the sale and thus the buyer would not receive the expected child token. This is a risk that is inherent to the design of this standard. Marketplaces should take this into account and provide a way to verify the expected child tokens are present when the parent token is being sold or to guard against such a malicious behaviour in another way.
 
+It is worth noting that `balanceOf` method only accounts for immediate tokens owned by the address; the tokens that are nested into a token owned by this address will not be reflected in this value as the recursive lookup needed in order to calculate this value is potentially too deep and might break the method.
```
lightclient, line 485: remove

lightclient (changes requested):
One serious breaking change to the API and one change to update the consistency of the spec to the reference implementation.

I'm definitely opposed to merging with the change to `ChildTransferred`. The standard is final and you shouldn't be modifying it any longer.

I might be able to be convinced to support changing `nestTransferFrom` since it does have a `data` field in the reference implementation. But it's still not great and definitely a spec change.

Please remove the elaborations in general.

EIPS/eip-6059.md
steven2308, line 458:
```suggestion
```

steven2308:
> I'm definitely opposed to merging with the change to `ChildTransferred`. The standard is final and you shouldn't be modifying it any longer.
We are well aware of this, however we contacted other implementers and they agreed this new parameter makes it easier since you can know if an NFT was soft locked (by being sent to address zero). All of them are using the open source package we provide, which includes this parameter on the event.

> I might be able to be convinced to support changing `nestTransferFrom` since it does have a `data` field in the reference implementation. But it's still not great and definitely a spec change.
This is the most critical part for us, thanks.

> Please remove the elaborations in general.
Will remove.

Yuripetusko:
> > I'm definitely opposed to merging with the change to `ChildTransferred`. The standard is final and you shouldn't be modifying it any longer.
> 
> We are well aware of this, however we contacted other implementers and they agreed this new parameter makes it easier since you can know if an NFT was soft locked (by being sent to address zero). All of them are using the open source package we provide, which includes this parameter on the event.
> 
> > I might be able to be convinced to support changing `nestTransferFrom` since it does have a `data` field in the reference implementation. But it's still not great and definitely a spec change.

Just to elaborate on this, While I fully agree that this is a no-no, and allowing a change to an ERC can open a pandora box. But in this particular case we have handful of devs looking into using this ERC right now and they are all in our workgroup, so we have direct contact with them. There are no projects that implemented this ERC in this form, with incorrect event and missing `data` argument in nestTransferFrom, as all devs use our implementation contract that have this event with missing field. And if you to allow us to add this change, this is the best time to do it, before anyone actually used it in production as per ERC rather than using our provided implementation. We will make sure to announce this change within our small workgroup too if the change makes it in.

SamWilsn:
> We are well aware of this, however we contacted other implementers and they agreed this new parameter makes it easier since you can know if an NFT was soft locked (by being sent to address zero).

This really should've happened before this ERC went to Final. That's the whole point of the review/last call statuses: give implementers time to try it out and provide feedback.

---

Final means, well... final. I'd recommend opening a new EIP, with a new number, if you'd like to make this change.

ThunderDeliverer:
Could we perhaps at least update the `nestTransferFrom` so that the specified interface matches the interface ID as well as the example implementation?

lightclient:
I recommend opening a new EIP to completely replace this one.

PR #7370, opened 2023-07-21
Update EIP-210: Make it not crash my build system

Pandapip1:
Sorry Vitalik, but that's not actually valid markdown.


lightclient (commented):
Why do you only fix line 36 and not 32?

SamWilsn (approved):

PR #7377, opened 2023-07-22
Add EIP: Migration Transaction

lightclient:
This EIP proposes a new transaction type that allows EOAs to submit a one-time upgrade to a smart contract.


EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Motivation
+
+For as long as Ethereum has existed, so have complaints about the user experience of the platform.
+
+Ethereum was designed to be an abstract platform, limited only by the mind of the developers who deploy to it. Smart contract wallets were expected to placate unhappy users. Unfortunately, many years into this journey, the tension is still here.
+
+The ecosystem looks different than it did when when the first smart contract wallets were written. Several application-level standards have mitigated some of the issues. There are large standardization efforts around smart contract wallets, and a big push to make them the default on layer two chains. Everything seems to be moving in the right direction.
+
+One significant problems that hasn't been resolved though is what to do about the EOA users of *today*. Manually migrating assets from an EOA is not a tenable path to converting Ethereum's user base to smart contract wallets. We must provide a mechanism, embedded in the protocol, to migrate. This EIP proposes such mechanism.
```
Pandapip1, line 26:
```suggestion
For as long as Ethereum has existed, so have complaints about its user experience. Ethereum was designed to be an abstract platform capable of running any program. Smart contract wallets were expected to bridge the gap between users and developers. Unfortunately, many years into this journey, the tension is still here.

The ecosystem looks different than it did when when the first smart contract wallets were written. Several application-level standards have mitigated some of the issues. There are large standardization efforts around smart contract wallets, and a big push to make them the default on layer two chains. Everything seems to be moving in the right direction.

However, one significant problem that hasn't been resolved is how to migrate exising EOA users. Manually migrating assets from an EOA is not a tenable path to converting Ethereum's user base to smart contract wallets. We must provide a mechanism, embedded in the protocol, to migrate. This EIP proposes such mechanism.
```
Some minor rephrasing. How does this sound?

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Specification
+
+At the fork block `X`, introduce the MigrationTransaction transaction type.
```
Pandapip1, line 30: Might be worth including a parameters table for the block number and transaction type, in case somehow a new Core EIP comes in the interim and takes up `0x04`.

Pandapip1, line 30:
```suggestion
At the fork block `X`, introduce the `MigrationTransaction` transaction type.
```
Nit: should be in code blocks.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+| `v`                    | `uint256` |
+| `r`                    | `uint256` |
+
+The EIP-2718 `TransactionType` is `0x04` and the `TransactionPayload` is `rlp([chainId, nonce, maxFeePerGas, maxPriorityFeePerGas, gasLimit, codePtr, storageTuples, data, value, accessList, yParity, r, s])`.
+
+The transaction's signature hash is `keccak256(0x04 || rlp([chainId, nonce, maxFeePerGas, maxPriorityFeePerGas, gasLimit, codePtr, storageTuples, data, value, accessList])`
+
+#### Validation
+
+A MigrationTransaction is considered valid if the follow properties hold:
```
Pandapip1, line 58:
```suggestion
A `MigrationTransaction` is considered valid if the follow properties hold:
```
Another nit: should be in code blocks.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+* all [EIP-1559](./eip-1559.md) properties, unless specified otherwise
+* the code at `codePtr` is less than the [EIP-170](./eip-170) limit of `24576`
```
Pandapip1, line 61:
```suggestion
* the code at `codePtr` is less than the [EIP-170](./eip-170.md) limit of `24576`
```
Nit: the extension is not optional.

CC @SamWilsn - this should be caught.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+* the code at `codePtr` must not have size `0`
+
+The intrinsic gas calculation modified from [EIP-1559](./eip-1559) to be `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count + 15000 * length of storage`.
```
Pandapip1, line 64:
```suggestion
The intrinsic gas calculation modified from EIP-1559 to be `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count + 15000 * length of storage`.
```
```suggestion
The intrinsic gas calculation modified from [EIP-1559](./eip-1559.md) to be `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count + 15000 * length of storage`.
```
You don't need to re-link to EIP-1559 (and you don't re-link anywhere else in the EIP), but if you do, you must include the .md at the end.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Processing
+
+Execution a MigrationTransaction transaction has two parts.
```
Pandapip1, line 68:
```suggestion
Executing a `MigrationTransaction` transaction has two steps.
```
```suggestion
Execution of a `MigrationTransaction` transaction has two parts.
```
This is currently grammatically incorrect, and there is a missing code tag. I don't know which fixed version you would prefer, so here they both are.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+##### Contract Deployment
+
+Unlike standard contract deployment, the MigrationTransaction directly specifies what `code` value of the sender's account should be set to.
```
Pandapip1, line 72:
```suggestion
Unlike standard contract deployment, a `MigrationTransaction` directly specifies what `code` value of the sender's account should be set to.
```
I imagine that there will be more than one `MigrationTransaction` ever. Additionally, missing code tag.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+As the first step of processing the transaction, set the sender's `code` value equal to `state[tx.codePtr].code`. Next, iterate through each tuple in `tx.storage` and in the sender's storage trie, set `storage[t.first] = t.second`.
```
Pandapip1, line 74:
```suggestion
As the first step of processing the transaction, set the sender's `code` to `state[tx.codePtr].code`. Next, for each tuple in `tx.storage` and the sender's storage trie, set `storage[t.first] = t.second`.
```
Rephrase to be a bit more trite.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+##### Transaction Execution
+
+Now instantiate an EVM call into the sender's account using the same rules as [EIP-1559](./eip-1559) and set the transaction's origin to be `keccak256(sender)[0..20]`.
```
Pandapip1, line 78:
```suggestion
Now instantiate an EVM call into the sender's account using the same rules as EIP-1559 and set the transaction's origin to be `keccak256(sender)[0..20]`.
```
```suggestion
Now instantiate an EVM call into the sender's account using the same rules as [EIP-1559](./eip-1559.md) and set the transaction's origin to be `keccak256(sender)[0..20]`.
```
I don't know why this is being done with EIP-1559 specifically. Pick one, but if there is a re-link, then the .md extension is not optional.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+No backward compatibility issues found.
+
+## Test Cases
+
+TBD
+
+## Reference Implementation
+
+TBD
```
Pandapip1, line 118:
```suggestion
```
No placeholders as optional sections.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Security Considerations
+
+As with all sufficiently sophisticated account designs, if a user can be convinced to sign an arbitrary message, that message could be MigrationTransaction which is owned by a malicious actor instead of the user. This can generally be avoided if wallets treat these transactions with *extreme* care and create as much friction and verification as possible before completing the signature.
```
Pandapip1, line 122:
```suggestion
If a user can be convinced to sign an arbitrary message by a malicious actor, that message could now set the account code of the victim. This can generally be avoided if wallets treat these transactions with *extreme* care and create a reasonable amount of friction to sign such a signature.
```

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+Additionally, the EOA already exists because it has enough balance for the MigrationTransaction to be considered valid. Therefore, we don't need to pay a premium for adding a new account into the state trie.
+
+### Manipulating transaction origin
+
+Many applications have a security check `caller == origin` to verify the caller is an EOA. This is done to "protect" assets. While it is usually more of a bandage than an actual fix, we attempt to placate these projects by modifying the origin of the transaction so the check will continue performing it's duty.
+
+### One-time migration
+
+There is no technical reason we couldn't allow EOAs to change their code at anytime with this transaction type. The only inhibitor at the moment is [EIP-3607](./eip-3607) which will cause migration transactions to be considered invalid if they come from an account with code already deployed. A functional reason for retaining this behavior though is that it makes it simpler to reason about contracts and their upgradability.
```
Pandapip1, line 106:
```suggestion
While there is no technical reason we couldn't allow EOAs to submit more than one `MigrationTransaction`, [EIP-3607](./eip-3607.md) causes subsequent migration transactions to be invalid. Retaining this behavior makes it simpler to reason about contracts and their upgrade-ability.
```
```suggestion
There is no technical reason we couldn't allow EOAs to change their code at any time with this transaction type. The only inhibitor at the moment is [EIP-3607](./eip-3607.md) which will cause migration transactions to be considered invalid if they come from an account with code already deployed. A functional reason for retaining this behavior though is that it makes it simpler to reason about contracts and their upgrade-ability.
```
Fix three typos (any time; upgrade-ability; missing .md). Optionally rephrase this paragraph.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+eip: 7377
+title: Migration Transaction
+description: Allow EOAs to send a one-time transaction which deploys code at their account.
+author: lightclient (@lightclient), Sam Wilson (@samwilsn), Ansgar Dietrichs (@adietrichs)
+discussions-to: https://ethereum-magicians.org/t/eip-xxxx-migration-transaction/15144
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-07-21
+requires: 170, 1559, 2718
```
Pandapip1, line 11:
```suggestion
requires: 170, 1559, 2718, 3607
```
You mentioned EIP-3607 should be used to avoid multiple migration transactions (which would incidentally not work as intended as all the memory is incremented). EIP-3607 should be added to the requires list.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+##### Transaction Execution
+
+Now instantiate an EVM call into the sender's account using the same rules as [EIP-1559](./eip-1559) and set the transaction's origin to be `keccak256(sender)[0..20]`.
+
+## Rationale
+
+### No `to` address field
+
+This transaction is only good for one time use to migrate an EOA to a smart contract. It is designed to immediately call the deployed contract, which is at the sender's address, after deployment to allow the sender to do any kind of further processing.
```
Pandapip1, line 84:
```suggestion
This transaction is designed to perform a one-time migration from an EOA to a smart contract. After deployment, it immediately calls the deployed contract at the sender's address to allow for further processing.
```
How does this re-phrasing sound?

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Code pointer for deployment
+
+Naively, one could design the migration transaction to have a field `code` that is just bytes. The outcome of this is that there would be substantial duplication of code calldata, since many users will want to deploy the exact same thing (a wallet). Using a pointer instead acknowledges the overwhelming use case for the transaction type, and exploits it as an optimization.
```
Pandapip1, line 88:
```suggestion
Naively, one could design the migration transaction to have a field `code` of type `bytes`. However, there would be substantial duplication of code calldata, since many users will want to deploy the same wallet. Using a pointer acknowledges the overwhelming use case for the transaction type while reducing its cost to the blockchain and the user.
```

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Cheaper storage
+
+Sometimes it's better to lead with a carrot instead of a stick.
```
Pandapip1, line 92:
```suggestion
Cheaper storage incentivizes users to migrate, while maintaining the storage cost means that there is more friction to the migration process. Additionally, since mass adoption of contract wallets is a positive externality, a storage subsidy makes sense.
```
Should probably use more formal language here.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Intrinsic does not account for contract deployment
+
+This is takes advantage of the fact that clients tend to single, unique copies of code. Therefore, the only operation here is changing a pointer in the state trie to desired code's pointer.
+
+Additionally, the EOA already exists because it has enough balance for the MigrationTransaction to be considered valid. Therefore, we don't need to pay a premium for adding a new account into the state trie.
```
Pandapip1, line 98:
```suggestion
Additionally, the EOA already exists because it has enough balance for the `MigrationTransaction` to be considered valid. Therefore, we don't need to pay a premium for adding a new account into the state trie.
```
Nit.

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+Naively, one could design the migration transaction to have a field `code` that is just bytes. The outcome of this is that there would be substantial duplication of code calldata, since many users will want to deploy the exact same thing (a wallet). Using a pointer instead acknowledges the overwhelming use case for the transaction type, and exploits it as an optimization.
+
+### Cheaper storage
+
+Sometimes it's better to lead with a carrot instead of a stick.
+
+### Intrinsic does not account for contract deployment
+
+This is takes advantage of the fact that clients tend to single, unique copies of code. Therefore, the only operation here is changing a pointer in the state trie to desired code's pointer.
```
Pandapip1, line 96:
```suggestion
Clients tend to single, unique copies of code. Therefore, the only operation clients have to perform is changing the account's code pointer in the state trie.
```
How about this? Also, there's a typo in your original phrasing ("This is takes...")

EIPS/eip-7377.md
```diff
@@ -0,0 +1,127 @@
...
+
+Additionally, the EOA already exists because it has enough balance for the MigrationTransaction to be considered valid. Therefore, we don't need to pay a premium for adding a new account into the state trie.
+
+### Manipulating transaction origin
+
+Many applications have a security check `caller == origin` to verify the caller is an EOA. This is done to "protect" assets. While it is usually more of a bandage than an actual fix, we attempt to placate these projects by modifying the origin of the transaction so the check will continue performing it's duty.
```
Pandapip1, line 102:
```suggestion
Many applications have a "security" check that `caller == origin` to test to see if an account is an EOA. This is done to "protect" assets, specifically where randomness is involved that a contract could subsequently revert on. While it has always been more of a bandage than an actual fix, we attempt to placate these projects by modifying the origin of the transaction so the check will continue performing it's duty.
```

EIPS/eip-7377.md
Pandapip1, line 1: General question: how will newly-migrated contract accounts be called considering EIP-3607? Gas needs to be paid!

Pandapip1 (changes requested):

EIPS/eip-7377.md
lightclient: Prefer the original, but updated some typos.

EIPS/eip-7377.md
lightclient: The newly-migrated contract accounts will need to be called by a transaction from an EOA. This is essentially what 4337 bundlers do.

EIPS/eip-7377.md
lightclient: meh I can update it

YaroslavPysanskyi:
I can't access my funds.

EIPS/eip-7377.md
lightclient: changed to migration transaction

EIPS/eip-7377.md
lightclient: gg

EIPS/eip-7377.md
lightclient: correct

EIPS/eip-7377.md
lightclient:
```suggestion
Naively, one could design the migration transaction to have a field `code` of type `bytes`. However, there would be substantial duplication of code calldata, since many users will want to deploy the exact same thing (often a wallet). Using a pointer instead acknowledges this overwhelming use case for the transaction type, and exploits it as an optimization.
```

EIPS/eip-7377.md
lightclient: That's no fun though

EIPS/eip-7377.md
lightclient:
```suggestion
This take advantage of the fact that clients tend to store a single, unique copy of code; no matter the number of deployments. Therefore, the only operation here is changing a pointer in the state trie to the desired code.
```

EIPS/eip-7377.md
lightclient: Not really true. This check is done because it "protects" the protocol from flash loans, programmable ownership, KYC requirements, and some other things.

EIPS/eip-7377.md
lightclient:
```suggestion
There is no technical reason we couldn't allow EOAs to change their code at any time with this transaction type. The only inhibitor at the moment is [EIP-3607](./eip-3607) which will cause migration transactions to be considered invalid if they come from an account with code already deployed. A functional reason for retaining this behavior though is that it makes it simpler to reason about contracts and their upgradability.
```

EIPS/eip-7377.md
lightclient: this is important

lightclient:
@YaroslavPysanskyi did u try turning it off and on again?

lightclient:
thanks for the feedback @Pandapip1, ready for review again

EIPS/eip-7377.md
lightclient: It isn't required to implement the EIP though. I just wanted to bring it to the reader's attention that this will be the behavior given 3607 exists on mainnet and there are some useful properties of it.

Pandapip1: Yes, but this EIP depends on that specific EIP-3607 behavior. I would argue that means this EIP requires the other EIP, since otherwise there are unintended consequences.

EIPS/eip-7377.md
Pandapip1: Ah, fair enough. I'll fix:
```suggestion
Many applications evaluate `msg.caller == tx.origin` to test if the caller is an EOA. This is done to "protect" assets from certain attacks that require contract wallets. While it has always been more of a bandage than an actual fix, we attempt to placate these projects by modifying the origin of the transaction so the check will continue performing it's duty.
```
The old wording just felt off.

EIPS/eip-7377.md
Pandapip1: Okay. You can have your fun. As long as you follow it up with a formal explanation.

```suggestion
Sometimes it's better to lead with a carrot instead of a stick.

Cheaper storage incentivizes users to migrate, while maintaining the storage cost means that there is more friction to the migration process. Additionally, since mass adoption of contract wallets is a positive externality, a storage subsidy makes sense.
```

Pandapip1 (changes requested):

lightclient:
@Pandapip1 the PR meets the criteria for draft so I would like to ask that you approve it.

EIPS/eip-7377.md
lightclient: EIP-1 states: 

> "A requires dependency is created when the current EIP cannot be understood or implemented without a concept or technical element from another EIP. Merely mentioning another EIP does not necessarily create such a dependency."

This EIP can both be understood and implemented without the concept / technical element from EIP-3607, therefore it is not a requirement.

EIPS/eip-7377.md
lightclient: `msg.caller` doesn't exist. `caller` and `origin` are EVM instructions, `msg` is a solidity concept.

EIPS/eip-7377.md
lightclient: I am alluding to the fact that all complex account designs have this issue. That is an important part of the statement and must not be omitted.

EIPS/eip-7377.md
Pandapip1: The first sentence is grammatically incorrect and the phrasing is wrong. I think that what you mean is that "This transaction is designed to perform a one-time migration from an EOA to a smart contrct." It's not "only good" for migration - it's its explicit goal.

Additionally, the second sentence has misplaced commas and the structure makes it hard to read.

EIPS/eip-7377.md
Pandapip1: I will be blocking this EIP until a formal explanation for this is added.

EIPS/eip-7377.md
Pandapip1: Why is that an important part of the statement? Isn't a "sufficiently sophisticated account design" implied by a "sign an arbitrary message"?

Pandapip1 (changes requested):

EIPS/eip-7377.md
```diff
@@ -0,0 +1,119 @@
...
+requires: 170, 1559, 2718
+---
+
+## Abstract
+
+Introduce a new [EIP-2718](./eip-2718.md) transaction type with the format `0x04 || rlp([chainId, nonce, maxFeePerGas, maxPriorityFeePerGas, gasLimit, codePtr, storage, data, value, accessList, yParity, r, s])` which sets the sending account's `code` field in the state trie to the `code` value at `codePtr` and applies the storage tuples to the sender's storage trie.
+
+## Motivation
+
+For as long as Ethereum has existed, so have complaints about the user experience of the platform.
```
g11tech, line 20: Would be better to not make generic blanket statements

EIPS/eip-7377.md
```diff
@@ -0,0 +1,119 @@
...
+
+Ethereum was designed to be an abstract platform, limited only by the mind of the developers who deploy to it. Smart contract wallets were expected to placate unhappy users. Unfortunately, many years into this journey, the tension is still here.
```
g11tech, line 22: Similarly here, don't see how text is adding value to eip

EIPS/eip-7377.md
```diff
@@ -0,0 +1,119 @@
...
+
+The ecosystem looks different than it did when when the first smart contract wallets were written. Several application-level standards have mitigated some of the issues. There are large standardization efforts around smart contract wallets, and a big push to make them the default on layer two chains. Everything seems to be moving in the right direction.
```
g11tech, line 24: May be we can compress this all in just one short line

EIPS/eip-7377.md
```diff
@@ -0,0 +1,119 @@
...
+
+One significant problem that hasn't been resolved though is what to do about the EOA users of *today*. Manually migrating assets from an EOA is not a tenable path to converting Ethereum's user base to smart contract wallets. We must provide a mechanism, embedded in the protocol, to migrate. This EIP proposes such mechanism.
```
g11tech, line 26: A short reason for why migration is important

EIPS/eip-7377.md
```diff
@@ -0,0 +1,119 @@
...
+#### Definition
+
+| field                  | type      |
+|------------------------|-----------|
+| `chainId`              | `int256`  |
+| `nonce`                | `uint64`  |
+| `maxFeePerGas`         | `int256`  |
+| `maxPriorityFeePerGas` | `int256`  |
+| `gasLimit`             | `uint64`  |
+| `codePtr`              | `address` |
```
g11tech, line 43: Better to name it codeAddress which is more clear than ptr

lightclient:
@Pandapip1 the formatting is correct. There is nothing in EIP-1 that says you can block merging a well-formatted draft because an explanation isn't formal enough. I again ask that you merge this to draft. I want to share this with the community, but you are delaying it.

EIPS/eip-7377.md
lightclient:
```suggestion
| `codeAddr`              | `address` |
```

lightclient: updated

lightclient:
@g11tech, wrote a new motivation, PTAL

Pandapip1:
> @Pandapip1 the formatting is correct. There is nothing in EIP-1 that says you can block merging a well-formatted draft because an explanation isn't formal enough. I again ask that you merge this to draft. I want to share this with the community, but you are delaying it.

Quoting EIP-1:

> ## EIP Editor Responsibilities
> 
> For each new EIP that comes in, an editor does the following:
> 
> * Read the EIP to check if it is ready: sound and complete. The ideas must make technical sense, even if they don’t seem likely to get to final status.
> * The title should accurately describe the content.
> * Check the EIP for language (spelling, grammar, sentence structure, etc.), markup (GitHub flavored Markdown), code style
>
> If the EIP isn’t ready, the editor will send it back to the author for revision, with specific instructions.

The majority of my complaints fall under the third option. In fact, sentence structure is **explicitly** called out as something that I can block an EIP for.

Additionally, my requiring of adding a formal explanation for your "lead with a carrot, not a stick" falls squarely under the first option, since I am blocking the EIP for not containing a sound and complete explanation for that particular section.

lightclient:
@Pandapip1 I've added more formal language. I don't agree with your other change requests and they are not reasonable grounds to block on.

PR #7389, opened 2023-07-24
Config: Yeet eip_validator

Pandapip1:
At this point, it is completely unused (superseded by eipw) and can be safely removed.


Pandapip1:
I give up. I can't figure out how the heck to change ruby deps.

PR #7395, opened 2023-07-25
Update EIP-6059: Align specification with interfaceID and example implementation

ThunderDeliverer:
During the proposal lifecycle from `Draft` to `Final` stage, an additional parameter was added to the `nestTransferFrom` method. This parameter is already taken into account within the `interfaceId`. This means that this PR fixes the inconsistency.

Additionally, there are other proposals (namely ERC-6220) that rely on the specification of this proposal that already implements the missing parameter.

While we understand that finalized proposals are not meant to be updated, we feel like updating the parameter that is already included in the example implementation as well as live implementations, will benefit the ecosystem. The current mismatch of `interfaceId` and the interface itself makes implementation according to the proposal impossible. Updating it will have a positive impact on the implementers as well as the ecosystem, as they are already relying on the complete interface (with the addition of the missing field) and have it integrated into their smart contracts.

### DIFF
--- a/EIPS/eip-6059.md
+++ b/EIPS/eip-6059.md
@@ -352,12 +352,14 @@ interface IERC6059 /* is ERC165 */ {
      * @param to Address of the receiving token's collection smart contract
      * @param tokenId ID of the token being transferred
      * @param destinationId ID of the token to receive the token being transferred
+     * @param data Additional data with no specified format, which SHOULD be sent in the `nestTransferFrom` call
      */
     function nestTransferFrom(
         address from,
         address to,
         uint256 tokenId,
-        uint256 destinationId
+        uint256 destinationId,
+        bytes memory data
     ) external;
 }
 ```


lightclient:
We already noted in https://github.com/ethereum/EIPs/pull/7366 that this ERC final and unfortunately cannot have any changes to the specification.

lightclient:
This is technically a normative change because it is modifying the specification. The reference implementation does not supersede the specification section.

steven2308:
> This is technically a normative change because it is modifying the specification. The reference implementation does not supersede the specification section.

I understand the possibility to update applies to two cases:
1. To correct errata.
2. To add non-normative clarifications
It does not say that both must apply, in this case we are asking to consider it under the first case.

If the `interfaceId` on the specification matched the methods we would not be asking for this change, but there is an inconsistency due this errata that makes the implementation non viable and affects over 60 projects already deployed with the right interface, which includes the data parameter. 

I don't think we are being irrational with this request and it is really for the benefit of the ecosystem, creating a duplicate of this with a tiny fix will only clutter the amount EIPs and affect several projects.

lightclient:
We typically only considered things like misspelled words, unclosed parentheses, and the like to be modified. Your proposal adds an entire new parameter, completely changing the function signature and abi selector. For this reason, I would not consider it non-normative. The reference implementation does not define the interface, the spec does.

I hear you, this is frustrating and seemingly a very silly distinction. Unfortunately we will not make an exception for this case and set the precedent that authors can just appeal to an editor to fix problems with their final EIP. Authors must be very confident in the spec of their standard and have it in production before moving to final as to avoid situations such as this.

PR #7397, opened 2023-07-25
Update EIP-820: Fix accidental tag

Pandapip1:
The creative commons CC0 link, since it was contained in an HTML code block instead of a markdown one, was interpreted as an HTML tag instead of raw text.

### DIFF
--- a/EIPS/eip-820.md
+++ b/EIPS/eip-820.md
@@ -331,7 +331,9 @@ The contract has the address above for every chain on which it is deployed.
 <details>
 <summary>Raw metadata of <code>./contracts/ERC820Registry.sol</code></summary>
 <pre>
-<code>{
+
+```json
+{
   "compiler": {
     "version": "0.4.24+commit.e67f0147"
   },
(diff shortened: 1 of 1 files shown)


EIPS/eip-820.md
```diff
@@ -331,7 +331,9 @@ The contract has the address above for every chain on which it is deployed.
 <details>
 <summary>Raw metadata of <code>./contracts/ERC820Registry.sol</code></summary>
 <pre>
```
SamWilsn, line 333: Still has an extra `<pre>` tag, no?

Pandapip1: Nope. L653/L657.

SamWilsn: Right, the `<pre>` tag is closed, but is it necessary?

The rendered HTML (at least on GitHub) already inserts a `<pre>` tag for fenced code blocks:

image

Pandapip1: I will not be removing the `<pre>` tag, as it is an unrelated modification to a Final EIP

SamWilsn (approved):

PR #7400, opened 2023-07-25
Add EIP: Flash Loans

alcueca:
Draft for a new Flash Loans ERC. Based on 3156, it offers more flexibility by segregating the asset and control flows. The `pull` architecture has also been changes to a `push` architecture.


EIPS/eip-7400.md
```diff
@@ -0,0 +1,211 @@
...
+        bytes calldata data,
+        /// @dev callback. This is a concatenation of (address, bytes4), where the address is the callback receiver, and the bytes4 is the signature of callback function.
+        /// @param initiator The address that called this function
+        /// @param paymentReceiver The address that needs to receive the amount plus fee at the end of the callback
+        /// @param asset The asset to be loaned
+        /// @param amount The amount to loaned
+        /// @param fee The fee to be paid
+        /// @param data The ABI encoded data to be passed to the callback
+        /// @return result ABI encoded result of the callback
+        function(address, address, IERC20, uint256, uint256, bytes memory) external returns (bytes memory) callback
```
fubuloubu, line 77: function callbacks are not an officially defined type in the ABIv2 spec, so this behavior is potentially undefined and specific to how Solidity handles it internally

alcueca: Interesting, I wanted to ask you specifically about this.

I would prefer to do the specification in YAML, and I assume in that case the callback would be split into a callback selector and a callback address.

Then, I assume that in the solidity reference implementation I could use a callback type, while in a potential vyper implementation of would be to separate variables.

Would that be acceptable?

fubuloubu: Currently, I believe this code produces a ABI tuple `(address, bytes4)` which would be more ideal for an ERC. You could make a note that Solidity will interpret a function pointer as a tuple and vice versa, and also mention the constraints that the function should have (i.e. it returns `bytes`, it's `nonpayable`, etc.)

alcueca: I'm thinking that the function call type in solidity is `abi.encodePacked(address, selector)`, meaning that maybe vyper would need to implement the same type 🤔

fubuloubu: nope, nevermind it gets encoded as `bytes24` according to docs:
image

fubuloubu: https://docs.soliditylang.org/en/latest/types.html#function-types

fubuloubu: So yeah, I would just put this as a `callback: bytes24` type (Vyper can handle that) and make a note that `function ....` in Solidity is converted to `bytes24` packed encoding of `address` + `bytes4` selector which can be used for processing the callback.

alcueca: Sounds good to me.

However, in putting the callback as a `bytes24` I think I'll have to specify that the 4 bytes are a function selector, and that the function selector MUST include the specific list of arguments described.

A bit cumbersome, but doable.

It would be easier if vyper would include a function call type the same as solidity :D

fubuloubu: Perhaps, but remember this isn't standardized behavior even in solidity, and could change any time

Also, remember that the ABI spec does not specify return values in computing a selector, nor does it specify mutability

What's wrong with just defining a specific callback function name that targets should implement? There's already more specific args in the function pointer, so it's not really a generic callback, might as well just make it simple and define the interface type and make this arg a simple `ERCXXXCallbackInterface` type

EIPS/eip-7400.md
```diff
@@ -0,0 +1,211 @@
+---
+eip: 7400
```
abcoathup, line 2:
```suggestion
eip: 7399
```
Manually assigning first PR number #7399.  Number sniping is not allowed.  PR 7400 was made one minute after a one line change PR 7399 by the same author.

EIPS/eip-7400.md
alcueca: Often there are several possible flows in the flash loan callback (leverage/deleverage, borrow/repay, etc) which the user needs to define an enum for, then encode in `data`, then decode in the callback, to finally route the flow to the right spot. [Example](https://github.com/yieldprotocol/yvarb/blob/7ac3648142ff48bbd0c12e429885697001007993/src/YieldNotionalLever.sol#L274).

Allowing the callback signature to be user defined removes this, allowing the user to specify that he wants the callback to be `Foo.bar` or `Foo.baz`.

It looks complicated in the spec, but not so much in the [implementation] (https://github.com/alcueca/erc3156pp/blob/56ba9dfdee8ad34150bc77de2bb67c950a1bb415/src/FlashLender.sol#L78).

fubuloubu: It also allows arbitrary callback target, is that intentional?

alcueca: Yes, this standard aims to segregate the asset and control flows on flash loans.

EIPS/eip-7400.md
abcoathup: Please use the assigned number 7399 for the EIP.  

Given the PRs could have been done in any order, to avoid any perception of number sniping, the first PR number should be used.

EIPS/eip-7400.md
```diff
@@ -0,0 +1,211 @@
...
+title: Flash Loans
+author: Alberto Cuesta Cañada (@alcueca), (@ultrasecreth)
```
abcoathup, line 4:
```suggestion
author: Alberto Cuesta Cañada (@alcueca)
```
Please only use a valid GitHub username.  You can list an author by name without contact information.

alcueca: That's a valid github username, do you mean I should do it like this?

`author: Alberto Cuesta Cañada (@alcueca), ultrasecreth (@ultrasecreth)`

EIPS/eip-7400.md
alcueca: Some encoding and decoding of function types to bytes24 [here](https://github.com/alcueca/erc7399-wrappers/blob/ca5cad7c5387b7aa582760f1d0aa24c28771357a/src/dydx/DYDXWrapper.sol#L120). It does produce much cleaner code than using `.call`.

EIPS/eip-7399.md
```diff
@@ -0,0 +1,211 @@
+---
+eip: 7399
+title: Flash Loans
+author: Alberto Cuesta Cañada (@alcueca), (@ultrasecr.eth)
```
abcoathup, line 4:
```suggestion
author: Alberto Cuesta Cañada (@alcueca), Ultrasecr.eth (@ultrasecreth)
```

EIPS/eip-7400.md
alcueca: Please let's continue this conversation [here](https://ethereum-magicians.org/t/erc7400-flash-loans/15211), I'd like to merge this PR and link the .md as soon as possible to avoid confusion between the PR number and the EIP number.

alcueca:
@fubuloubu, is there a way to give a kick to @eth-bot to review this?

fubuloubu:
> @fubuloubu, is there a way to give a kick to @eth-bot to review this?

No idea

alcueca:
@abcoathup?

EIPS/eip-7399.md
```diff
@@ -0,0 +1,217 @@
...
+        bytes calldata data,
+        /// @dev callback. This is a concatenation of (address, bytes4), where the address is the callback receiver, and the bytes4 is the signature of callback function.
+        /// @param initiator The address that called this function
+        /// @param paymentReceiver The address that needs to receive the amount plus fee at the end of the callback
+        /// @param asset The asset to be loaned
+        /// @param amount The amount to loaned
+        /// @param fee The fee to be paid
+        /// @param data The ABI encoded data to be passed to the callback
+        /// @return result ABI encoded result of the callback
+        function(address, address, IERC20, uint256, uint256, bytes memory) external returns (bytes memory) callback
```
Pandapip1, line 76: Is this valid syntax? I've never seen it before.

EIPS/eip-7399.md
```diff
@@ -0,0 +1,217 @@
...
+    /// @dev Use the aggregator to serve an ERC3156++ flash loan.
+    /// @param loanReceiver The address receiving the assets in the flash loan
+    /// @param asset The asset to be loaned
+    /// @param amount The amount to loaned
+    /// @param data The ABI encoded user data
+    /// @param callback The address and signature of the callback function
+    /// @return result ABI encoded result of the callback
+    function flash(
+        address loanReceiver,
+        IERC20 asset,
```
Pandapip1, line 65:
```suggestion
        ERC20 asset,
```
The interface name is `ERC20`, not `IERC20`

EIPS/eip-7399.md
```diff
@@ -0,0 +1,217 @@
...
+import { IERC20 } from "./IERC20.sol";
+
+interface IERC3156PPFlashLender {
+
+    /// @dev The fee to be charged for a given loan.
+    /// @param asset The loan currency.
+    /// @param amount The amount of assets lent.
+    /// @return The amount of `asset` to be charged for the loan, on top of the returned principal. Returns type(uint256).max if the loan is not possible.
+    function flashFee(
+        IERC20 asset,
```
Pandapip1, line 52:
```suggestion
        ERC20 asset,
```
The interface name is `ERC20`, not `IERC20`

EIPS/eip-7399.md
```diff
@@ -0,0 +1,217 @@
...
+        function(address, address, IERC20, uint256, uint256, bytes memory) external returns (bytes memory) callback
+    ) external returns (bytes memory);
+}
+```
+
+The `flashFee` function MUST return the fee charged for a loan of `amount` `asset`. If the asset is not supported `flashFee` MUST revert. If the lender doesn't have enough liquidity to loan `amount` the fee returned must be `type(uint256).max`.
+
+The `flash` function MUST execute the callback passed on as an argument.
+
+```
```
Pandapip1, line 85:
```suggestion
```solidity
```

Pandapip1 (changes requested):
Partial review

EIPS/eip-7399.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Specification
+
+A flash loan integrates two smart contracts using a callback pattern. These are called the `lender` and the `callback receiver` in this EIP.
+
+In the same flash loan, assets are transferred to the `loan receiver` before the callback and expected in the `payment receiver` at the end of the callback. An additional amount of assets, called a `fee`, MIGHT be also required to the present in the `payment receiver` at the end of the callback.
+
+### Lender Specification
+
+A `lender` MUST implement the IERC7399FlashLender interface.
```
Pandapip1, line 39:
```suggestion
A `lender` MUST implement the `IERC7399FlashLender` interface.
```

EIPS/eip-7399.md
```diff
@@ -0,0 +1,210 @@
...
+Alternatively, the callback receiver can verify that `amount` was received by the `loanReceiver` and use its own heuristics to determine if a `fee` is fair and the loan repaid, or the transaction reverted.
+
+### Flash minting external security considerations
+
+The typical quantum of assets involved in flash mint transactions will give rise to new innovative attack vectors.
+
+#### Interest Rate Attack
+
+If there exists a lending protocol that offers stable interests rates, but it does not have floor/ceiling rate limits and it does not rebalance the fixed rate based on flash-induced liquidity changes, then it could be susceptible to the following scenario:
+
```
Pandapip1, line 178:
```suggestion
```

EIPS/eip-7399.md
devtooligan: yup
https://docs.soliditylang.org/en/v0.8.20/types.html#function-types
https://twitter.com/PaulRBerg/status/1609917508223475712?s=20

EIPS/eip-7399.md
```diff
@@ -0,0 +1,217 @@
...
+A flash loan integrates two smart contracts using a callback pattern. These are called the `lender` and the `callback receiver` in this EIP.
+
+In the same flash loan, assets are transferred to the `loan receiver` before the callback and expected in the `payment receiver` at the end of the callback. An additional amount of assets, called a `fee`, MIGHT be also required to the present in the `payment receiver` at the end of the callback.
+
+### Lender Specification
+
+A `lender` MUST implement the IERCXXXXFlashLender interface.
+
+```
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
alcueca: I didn't know that the license for the specification needed to be CC0, I thought that was just for the EIP. In any case, I changed it to CC0.

EIPS/eip-7399.md
alcueca: I've added some extra comments in the EIP explaining how it works.

EIPS/eip-7399.md
```diff
@@ -0,0 +1,217 @@
...
+## Backwards Compatibility
+
+This EIP is a successor of [ERC-3156](./eip-3156.md). While not directly backwards compatible, a wrapper contract offering a [ERC-7399](./eip-7399.md) external interface with liquidity obtained from an ERC-3156 flash `lender` is trivial to implement.
+
+## Reference Implementation
+
+TBA
+
+
+## Security Considerations
```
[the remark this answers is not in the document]
alcueca: Thank you, but they are a reviewed copy of the ERC3156 considerations, since they are equally valid for ERC7399. We are working on possible issues that could arise from the increased complexity of ERC7399 in comparison to ERC3156, or from the switch to a push repayment process.

EIPS/eip-7399.md
```diff
@@ -0,0 +1,222 @@
...
+        /// @return result ABI encoded result of the callback
+        function(address, address, IERC20, uint256, uint256, bytes memory) external returns (bytes memory) callback
+    )
+        external
+        returns (bytes memory);
+}
+
+```
+
+The `flashFee` function MUST return the fee charged for a loan of `amount` `asset`. If the asset is not supported `flashFee` MUST revert. If the lender doesn't have enough liquidity to loan `amount` the fee returned must be `type(uint256).max`.
```
MathisGD, line 105: For some contracts, the difference between a "not supported" asset and an asset for which the lender has zero liquidity can be unclear. For example UniV4 singleton contract: it technically supports all assets, but it can have zero liquidity for some of them. So should it reverts or return `type(uint).max` in this case ?

EIPS/eip-7399.md
```diff
@@ -0,0 +1,222 @@
...
+
+A `fee` will often be calculated in the callback function, which the callback receiver must be aware of for repayment. Passing the `fee` as a parameter instead of appended to `data` is simple and effective.
+
+Arbitrary callback functions on callback receivers allows to implement different behaviours to flash loans on callback receivers without the need for encoding a function router using the `data` argument. A function call type is 24 bytes of which the first 20 bytes are the target address and the last 4 bytes are the function signature.
+
+The `amount + fee` are pushed to the `payment receiver` to allow for the segregation of asset and control flows. While a "pull" architecture is more prevalent, "push" architectures are also common. For those cases where the `lender` can't implement a "push" architecture, a simple wrapper contract can offer an ERC-7399 external interface, while using liquidity from the `lender` using a "pull" architecture.
```
MathisGD, line 167: I think that push architectures are compatible with less contracts. Let's say that you have a lending pool which exposes an EIP-7399 `flash`, a `supply` and a `withdraw` entry-point. One could drain the contract by doing: flash(amount: x) then in the callback call supply(amount: x), which returns the assets to the lending pool, so the balance check passes and the flashLoan can end. Then the user is able call withdraw(amount: x) and get back all the supplied funds.

There are some ways to prevent this, but those are much less direct than implementing a pull flashLoan.

EIPS/eip-7399.md
alcueca: In that case I would return `type(uint).max`, since the asset is technically supported but has no liquidity.

The rationale is that borrowers will usually know which assets they can get from each lender, but call `maxFlashLoan` anyway to find out if they can borrow a particular amount at a particular time. Reverting on no liquidity would be bad UX.

Returning `type(uint).max` on `flashFee` follows the same logic. Borrowers are expected to know what they are doing and where to get each asset from. They can bypass the `maxFlashLoan` check and use `flashFee` instead to get both data points (enough liquidity and fee to pay) in a single call. A revert on `flashFee` is unexpected and notifies the borrower that they are doing something wrong.

alcueca: That said, this is a great point, and probably needs to be added to the EIP. In the [wrappers we have just released](https://github.com/alcueca/erc7399-wrappers) I'm not following a consistent approach on `flashFee` for Aave (probably needs to revert more), Balancer (shouldn't revert) and Uniswap (shouldn't revert).

alcueca:
@Pandapip1, could I get a hand here, please?

EIPS/eip-7399.md
alcueca: Fixed the EIP, thanks for your feedback 👍

SamWilsn:

PR #7412, opened 2023-07-26
Add EIP: On-Demand Off-Chain Data Retrieval

noahlitvin:


EIPS/eip-ondemand-data.md
```diff
@@ -0,0 +1,152 @@
...
+interface IOracleContract {
+  function fulfillOracleData(bytes oracleQuery, bytes signedOffchainData);
+}
+```
+
+## Rationale
+
+This proposal is essentially an alternative to [EIP-3668](./eip-3668.md) with a few important distinctions:
+
+- EIP-3668 requires URIs to be encoded on-chain. While this can work well for static assets (such as IPFS hashes for assets related to NFTs and merkle trees), it is not ideal for retrieving data that must be fresh like cross-chain data retrieval or price feeds.
```
ali-bahjati, line 50: it also has url maintenance + centralization risks

EIPS/eip-ondemand-data.md
```diff
@@ -0,0 +1,152 @@
...
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-07-26
+---
+
+## Abstract
+
+Contracts may require off-chain data that is not already available on chain during execution. A smart contract function could implement the standard proposed here by reverting with `error OracleDataRequired(bytes32 oracleId, bytes oracleQuery, address oracleContract)`. Clients supporting this standard would recognize this error message during a simulation of the request, query the specified decentralized oracle network for signed data, and instead stage a transaction with a multicall that prepends the verification of the required off-chain data. The data would be written on-chain during verification to a smart contract for the subsequent call to read, avoiding the error.
```
ali-bahjati, line 14: Afaik multicalls are not a standard in Ethereum. So it would be good if you refer the multicall contract and explain it a bit.

EIPS/eip-ondemand-data.md
```diff
@@ -0,0 +1,152 @@
...
+  function fulfillOracleData(bytes oracleQuery, bytes signedOffchainData);
+}
+```
+
+## Rationale
+
+This proposal is essentially an alternative to [EIP-3668](./eip-3668.md) with a few important distinctions:
+
+- EIP-3668 requires URIs to be encoded on-chain. While this can work well for static assets (such as IPFS hashes for assets related to NFTs and merkle trees), it is not ideal for retrieving data that must be fresh like cross-chain data retrieval or price feeds.
+- By relying on a multicall rather than callbacks, it is much simpler to handle situations in which nested calls require different off-chain data. By the standard proposed here, end users (including those using clients that implement account abstraction) always need to simply sign a transaction, regardless of the complexity of the internal structure of the call being executed. The client can automatically prepend any necessary off-chain data to the transaction for the call to succeed.
```
ali-bahjati, line 51: Although multicall is a simpler approach, it is more expensive gas-wise.

EIPS/eip-ondemand-data.md
```diff
@@ -0,0 +1,152 @@
...
+}
+```
+
+Note that the `add` function can simply call the cross-chain function on the Oracle Contract. The developer of the `CrosschainAdder` contract does not need to be concerned with the implementation of this standard.
+
+Cross-chain functions like this could also be leveraged to avoid O(n) (and greater) loops on-chain. For example, `chainIdA` and `chainIdB` could reference the same chain that the `CrosschainAdder` contract is deployed on with `functionA` and `functionB` as view functions with computationally intensive loops.
+
+## Security Considerations
+
+This standard has similar security considerations to [EIP-3668](./eip-3668.md). However, regarding HTTP requests, it is more secure as the client is not expected to query an arbitrary HTTP endpoint specified by the contract. It is instead the responsibility of the client to decide which gateway should be used to connect to the decentralized oracle network specified in the error data.
```
ali-bahjati, line 144: I like to see how this is gonna work out practically. i.e: from where the clients get the urls? how clients make sure there is no conflicing id? Is there a registry of some sort? Do clients get the urls upon instantiation?

EIPS/eip-ondemand-data.md
```diff
@@ -0,0 +1,152 @@
...
+    _verify(signedOffchainData);
+    latestVerifiedData[keccak256(oracleQuery)] = signedOffchainData;
+  }
+
+  function retrieveCrossChainData(uint chainId, address contractAddress, bytes payload) internal returns (bytes) {
+    bytes memory oracleQuery = abi.encode(chainId, contractAddress, payload);
+    (uint timestamp, bytes response) = abi.decode(latestVerifiedData[oracleQuery], (uint, bytes));
+
+    if(timestamp < block.timestamp - STALENESS_TOLERANCE){
+      revert OracleDataRequired(ORACLE_ID, oracleQuery, address(oracleContract));
```
jayantk, line 107: nitpick: missing definition for the constant ORACLE_ID above

EIPS/eip-ondemand-data.md
jayantk: yeah it's probably good to say a word about this. can likely copy the state of affairs for chain ids

jayantk (commented):
This proposal makes a lot of sense to me. Left you some comments inline in the doc.

EIPS/eip-ondemand-data.md
```diff
@@ -0,0 +1,152 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7412
```
Assigning EIP number using PR number.  Please also update the filename

noahlitvin: All set, thanks!

EIPS/eip-ondemand-data.md
```diff
@@ -57,6 +66,8 @@ Another major advantage of this standard is that oracles can charge fees in the
...
 
+It is the responsibility of the client to decide how to construct the multicall. Wallets that support account abstraction (per [ERC-4337](./eip-4337.md)) should already have the ability to generate atomic multi-operations. For EOA support, a multicall contract may be used. This may need to be a protocol-specific multicall proxy contract if the function being called references `msg.sender`. Here, the protocol should honor the reported `msg.sender` value passed from the multicall proxy contract.
```
jayantk: how do you think clients will implement this in practice? it seems like they will now need to know how each protocol supports multicalls, which will require another registry similar to the oracle one. 

if there's a way for the contract to suggest to the client how to do the multicall, that could help make this more generic.

noahlitvin: Yea, we discussed potentially having a multicall address in the revert. Realistically, I think support for this will all be built in at the application layer until wallets support account abstraction, but it's hard to say.

jayantk: yeah... i guess if you think we're going to get account abstraction soon, then it's a moot point.

noahlitvin: Another solution could be valuable, but I think it'd be a different EIP? Like if the `CrosschainAdder` contract had functions with `msg.sender`, it would need to be able to communicate 'use `x` function to multicall me' (and it may want to do this for reasons other than this standard too). The `OracleContract` implementing this standard wouldn't be able to help. (Or maybe there's something clever I'm not considering...)

noahlitvin: Practically, I was imagining there'd be a community-maintained TS library that integrates with viem/ethers.js/etc. There'd be an extendable config for oracle id resolution and an extendable config with a mapping of chainId:contractAddress => multicall func. If you're trying to call a contract that isn't in mapping, you'd need to specify that it should use the generic multicall contract or a custom one.

EIPS/eip-7412.md
```diff
@@ -0,0 +1,167 @@
...
+discussions-to: https://ethereum-magicians.org/t/erc-7412-on-demand-off-chain-data-retrieval/15346
+status: Draft
+type: Standards Track
+category: ERC
+created: 2023-07-26
+---
+
+## Abstract
+
+Contracts may require off-chain data that is not already available on chain during execution. A smart contract function could implement the standard proposed here by reverting with `error OracleDataRequired(address oracleContract, bytes oracleQuery)`. Clients supporting this standard would recognize this error message during a simulation of the request, query the specified decentralized oracle network for signed data, and instead stage a transaction with a multicall that prepends the verification of the required off-chain data. The data would be written on-chain during verification to a smart contract for the subsequent call to read, avoiding the error.
```
SamWilsn, line 15:
```suggestion
Contracts may require off-chain data that is not already available on-chain during execution. A smart contract function could implement the standard proposed here by reverting with `error OracleDataRequired(address oracleContract, bytes oracleQuery)`. Clients supporting this standard would recognize this error message during a simulation of the request, query the specified decentralized oracle network for signed data, and instead stage a transaction with a multicall that prepends the verification of the required off-chain data. The data would be written on-chain during verification to a smart contract for the subsequent call to read, avoiding the error.
```

EIPS/eip-7412.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Motivation
+
+Ethereum's scaling roadmap involves a series of separate execution contexts for smart contract code (including layer two and layer three scaling solutions). This makes the ability to read data across multiple chains crucial to the construction of scalable applications. Also, for decentralized finance protocols that rely on price data, it is not reasonable to expect oracle networks will be able to continuously push fresh data to every layer two and layer three network for an arbitrary number of price feeds.
+
+Cross-chain bridges are being developed, adhering to standards such as [ERC-6170](./eip-6170.md), where smart contract functions can write data to other chains. There is a need for a similar standard that enables reading data from other chains. This standard can be generalized for reading any off-chain data from a decentralized oracle network, including price feeds.
```
SamWilsn, line 21: You are allowed to reference other proposals, **but** doing so means your proposal cannot advance past their statuses. So, for example, if ERC-6170 is a Draft, this proposal wouldn't be able to advance to Review.

If you're fine with that, you can leave this reference in. I just wanted to point it out.

EIPS/eip-7412.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Specification
+
+A contract implementing this standard MUST revert with the following error whenever off-chain data is required:
+
+```solidity
+error OracleDataRequired(address oracleContract, bytes oracleQuery)
+```
+
+`oracleQuery` specifies the off-chain data that is being required. Valid data formats for this parameter are specific to the oracle ID used. This might include chain id, contract address, function signature, payload, and timestamp/"latest" for cross-chain reads. For price feeds, it could include a ticker symbol and timestamp/"latest".
```
SamWilsn, line 35: "oracle ID" hasn't been defined yet, so maybe this is more clear:

```suggestion
`oracleQuery` specifies the off-chain data that is being required. Valid data formats for this parameter are specific to the oracle contract used. This might include chain id, contract address, function signature, payload, and timestamp/"latest" for cross-chain reads. For price feeds, it could include a ticker symbol and timestamp/"latest".
```

I could also be entirely incorrect here :rofl:

SamWilsn: Ah, I was entirely incorrect. I'd _maybe_ suggest moving the definition of oracle ID up a bit so that readers don't make the same mistake I did.

EIPS/eip-7412.md
```diff
@@ -0,0 +1,167 @@
...
+
+`oracleId` is a unique identifier that references the decentralized oracle network that generates the desired signed off-chain data. Oracle IDs would be analogous to Chain IDs in the Ethereum ecosystem. Clients are expected to resolve a gateway that corresponds to an Oracle ID, similar to how clients are expected to resolve an RPC endpoint based on a Chain ID.
+
+The contract implementing the `IOracleContract` interface MUST revert with the following error message if it requires payment to fulfill the oracle data query:
+
+```solidity
+error OracleFeeRequired(uint feeAmount)
+```
+
+`feeAmount` specifies the amount of native gas tokens required to execute the `fulfillOracleData` function, denominated in wei. This error MUST be resolved if the caller provides sufficient gas such that the fee amount can be collected by the oracle contract. The contract MAY NOT return gas tokens if they are provided in excess of the `feeAmount`. In practice, we would expect the fee amount to remain relatively stable, if not constant.
```
SamWilsn, line 54: How the fee is collected seems a bit vague, at least to me, a non-solidity expert. If it's intentionally left out, maybe include a sentence to that effect?

EIPS/eip-7412.md
```diff
@@ -0,0 +1,167 @@
...
+
```
SamWilsn, line 55: The specification section doesn't really explain the "multicall" required to set up the data. For example: does it have to be done in a single transaction, or can the wallet submit the data in one transaction, then make the call in a second one?

EIPS/eip-7412.md
```diff
@@ -0,0 +1,167 @@
...
+To prevent data becoming too stale for a request between the simulation and a call's execution, ideally a contract could also emit the following event: `event OracleDataUsed(address oracleContract, bytes oracleQuery, uint expirationTime)` Here, `expirationTime` is the time after which the `OracleDataRequired` error would be thrown by the contract. (This would typically be a calculation involving a staleness tolerance and `block.timestamp`). Client applications that implement this standard would be able to recognize this event during simulation and estimate if an additional update will still be necessary, taking into account the speed of the chain. For example, the oracle query may request the latest quote available for a particular price feed and the expiration time may signal that the price cannot be older than three seconds prior to the current timestamp recognized by the blockchain. This has been omitted from the standard because there isn't a practical way to retrieve event data during transaction simulations on most JSON-RPC APIs at this time.
```
SamWilsn, line 68: Similarly, this seems like either an implementation suggestion or a security consideration.

Charria12 (approved):

Charria12 (approved):

Charria12 (approved):

PR #7419, opened 2023-07-27
Update EIP-7377: Fix some nits

jtraglia:
While reading this EIP, I noticed a few mistakes I would like to fix:

* Minor typo in sentence.
* Minor table formatting issue.
* Unnecessary comma in sentence.
* `storageTuples` should be `storage` to match table of fields.
* An "it's" v.s. "its" problem.
* A duplicate word in "migration transaction transaction."
* I think "MigrationTransaction" should be "migration transaction."

By the way, the transaction type format in the abstract uses `storage` instead of `storageTuples` too:

https://github.com/ethereum/EIPs/blob/bf74be7bd9e11082e3fb1a73cf31aa19b2287b57/EIPS/eip-7377.md?plain=1#L16


lightclient (approved):
amazing thank you sir

PR #7432, opened 2023-07-31
Add EIP: Non-Fungible Token Roles

karacurt:
This standard introduces role management for NFTs. Each role assignment is associated with a single NFT and expires
automatically at a given timestamp. Inspired by `ERC-5982` roles are defined as `bytes32` and feature a
custom `_data` field of arbitrary size to allow customization.

The NFT Roles interface aims to establish a standard for role management in NFTs. Tracking on-chain roles enables
decentralized applications (dApps) to implement access control for privileged actions, e.g., minting tokens with a role
(airdrop claim rights).

NFT roles can be deeply integrated with dApps to create a utility-sharing mechanism. A good example is in digital real
estate. A user can create a digital property NFT and grant a `keccak256("PROPERTY_MANAGER")` role to another user,
allowing them to delegate specific utility without compromising ownership. The same user could also grant multiple
`keccak256("PROPERTY_TENANT")` roles, allowing the grantees to access and interact with the digital property.

There are also interesting use cases in decentralized finance (DeFi). Insurance policies could be issued as NFTs, and 
the beneficiaries, insured, and insurer could all be on-chain roles tracked using this standard.

### DIFF
--- a/EIPS/eip-7213.md
+++ b/EIPS/eip-7213.md
@@ -1,5 +1,4 @@
 ---
-# EIP-7213: Non-Fungible Token Roles
 eip: 7213
 title: Non-Fungible Token Roles
 description: Role Management for NFTs. Enables accounts to share the utility of NFTs via expirable role assignments.  
@@ -15,7 +14,7 @@ requires: 165
 ## Abstract
 
 This standard introduces role management for NFTs. Each role assignment is associated with a single NFT and expires
-automatically at a given timestamp. Inspired by [EIP-5982](./eip-5982.md), roles are defined as `bytes32` and feature a
+automatically at a given timestamp. Inspired by [ERC-5982](./eip-5982.md), roles are defined as `bytes32` and feature a
 custom `_data` field of arbitrary size to allow customization.
 
 ## Motivation
@@ -37,7 +36,7 @@ the beneficiaries, insured, and insurer could all be on-chain roles tracked usin
 The keywords “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, 
 “NOT RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
-ERC-7213 compliant contracts MUST implement the following interface:
+[ERC-7213](./eip-7213.md) compliant contracts MUST implement the following interface:
 
 ```solidity
 /// @title ERC-7213 Non-Fungible Token Roles
@@ -160,11 +159,11 @@ interface IERC7213 /* is ERC165 */ {
 * The `hasRole` function SHOULD return `false` if `_supportsMultipleAssignments` is `false` and last role assignment is not to
 `_grantee`.
 * The `roleData` function MAY be implemented as `pure` or `view`.
-* Compliant contracts SHOULD implement [EIP-165](./eip-165.md) identifier.
+* Compliant contracts SHOULD implement [ERC-165](./eip-165.md) identifier.
 
 ## Rationale
 
-ERC-7213 IS NOT an extension of [ERC-721](./eip-721.md) or [ERC-1155](./eip-1155.md). The main reason behind this
+[ERC-7213](./eip-7213.md) IS NOT an extension of [ERC-721](./eip-721.md) or [ERC-1155](./eip-1155.md). The main reason behind this
 decision is to keep the standard agnostic of any NFT implementation. This approach also enables the standard to be
 implemented externally or on the same contract as the NFT, and allow dApps to use roles with immutable NFTs.
 
@@ -261,7 +260,7 @@ Updated Metadata Schema:
 }
 ```
 
-The following JSON is an example of ERC-7213 Metadata:
+The following JSON is an example of [ERC-7213](./eip-7213.md) Metadata:
 
 ```js
 {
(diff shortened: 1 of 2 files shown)


EIPS/eip-7213.md
```diff
@@ -0,0 +1,328 @@
+---
+eip: 7213
```
abcoathup, line 2:
```suggestion
eip: 7432
```
Assigning EIP number as the PR number.  #7213 is an unrelated PR.
Please also change the filename

karacurt: Done, Thank you! @abcoathup

ernanirst:
Hi @xinbenlv 

Yes! We did take a very good look at EIP-5982, we even reference it on the abstract. Just like in EIP-5982, roles are represented as `bytes32`, the key difference with this approach is that we need to associate roles with NFTs (`tokenAddress` + `tokenId`).

We also took an off-chain approach towards metadata, which aligns with the existing metadata extensions of EIP-721 and EIP-1155, but differs from EIP-5982.

ernanirst:
@xinbenlv we also reviewed #7303 and it seems we have different goals. EIP-7303 aims to represent roles as NFTs, while this EIP intends to create roles **for NFTs**. Creating roles for NFTs enable "role users" to use the NFT in a specific way without compromising ownership.

ernanirst:
@abcoathup we have created a reference implementation [here](https://github.com/OriumNetwork/ERC7432-reference-implementation). Is it possible to include this link the docs? I know external URLs are not allowed, but not even Github links?

EIPS/eip-7432.md
```diff
@@ -0,0 +1,346 @@
...
+type: Standards Track
+category: ERC
+created: 2023-07-14
+requires: 165
+---
+
+## Abstract
+
+This standard introduces role management for NFTs. Each role assignment is associated with a single NFT and expires
+automatically at a given timestamp. Inspired by [ERC-5982](./eip-5982.md), roles are defined as `bytes32` and feature a
```
SamWilsn, line 17: Linking to another proposal means this proposal (ERC-7432) cannot advance past ERC-5982's status. So if 5982 is Review, 7432 cannot go to Last Call.

You are more than welcome to keep this link, just wanted to give you a heads up.

EIPS/eip-7432.md
```diff
@@ -0,0 +1,346 @@
...
+
+There are also interesting use cases in decentralized finance (DeFi). Insurance policies could be issued as NFTs, and 
+the beneficiaries, insured, and insurer could all be on-chain roles tracked using this standard.
+
+## Specification
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",
+"NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC-2119 and RFC-8174.
+
+[ERC-7432](./eip-7432.md) compliant contracts MUST implement the following interface:
```
SamWilsn, line 39:
```suggestion
Compliant contracts MUST implement the following interface:
```

EIPS/eip-7432.md
```diff
@@ -0,0 +1,346 @@
...
+MUST return:
+
+|     Role Type     | `_supportsMultipleAssignments` |  Is last Role granted?  | Result of `hasRole` |
+|:-----------------:|:------------------------------:|:-----------------------:|:-------------------:|
+|  Non-Unique Role  |            `false`             |       Irrelevant        |       `true`        |
+|    Unique Role    |             `true`             |         `true`          |       `true`        |
+|    Unique Role    |             `true`             |         `false`         |       `false`       |
+
+In conclusion, the `_supportsMultipleAssignments` argument only affects the result when `true`, and if the queried
+assignment is not the last one granted.
```
SamWilsn, line 198: This belongs in the specification section, since it defines behaviour. In rationale, for example, you might want to explain _why_ you support unique and non-unique roles, and why the uniqueness boolean is set when checking the role instead of when assigning it.

EIPS/eip-7432.md
```diff
@@ -0,0 +1,346 @@
...
+```
+
+The `roles` array properties are SUGGESTED, and developers should add any other relevant information as necessary (e.g.,
+an image for the role). However, it's highly RECOMMENDED to include the `supportsMultipleAssignments` property, as this
+field is required in the `hasRole` function (refer back to [Unique and Non-Unique Roles](#unique-and-non-unique-roles)).
+
+It's also important to highlight the importance of the `inputs` property. This field describes the parameters that
+should be encoded and passed to the `grantRole` function. It's RECOMMENDED to use the properties `type` and `components`
+defined on the Solidity ABI Specification, where `type` is the canonical type of the parameter, and `components` is used
+for complex tuple types.
```
SamWilsn, line 315: This is also specification, so it should go into that section.

EIPS/eip-7432.md
```diff
@@ -0,0 +1,346 @@
...
+roles of immutable NFTs. It also helps with backward compatibility as NFT-specific functions such as `ownerOf` and
+`balanceOf` aren't required. Consequently, this design ensures a more straightforward integration with different
+implementations of NFTs.
+
+## Reference Implementation
+
+NFT Roles - A reference implementation created by Orium Network.
+
+* CC0 License.
+* 100% test coverage.
```
SamWilsn, line 330: Please either include a minimal reference implementation in the `assets` directory, or remove the mention of it.

EIPS/eip-7432.md
ernanirst: Removed mentions to ERC-5982

EIPS/eip-7432.md
ernanirst: Fixed

EIPS/eip-7432.md
ernanirst: Moved!

EIPS/eip-7432.md
ernanirst: Included a couple solidity files for reference

EIPS/eip-7432.md
ernanirst: As you suggested over the call, we split the `hasRole` function in two: `hasRole` and `hasUniqueRole`. This also removed the `_supportsMultipleAssignments` which greatly simplified the rationale/specification

SamWilsn (approved):

PR #7437, opened 2023-08-01
Update EIP-6551: Move to Review

jaydenwindle:

 - The build passes.

### DIFF
--- a/EIPS/eip-6551.md
+++ b/EIPS/eip-6551.md
@@ -28,7 +28,7 @@ Unlike other forms of on-chain identity, NFTs cannot act as agents or associate 
 - An automated investment portfolio composed of multiple fungible assets
 - A punch pass membership card granting access to an establishment and recording a history of past interactions
 
-This proposal grants every NFT the full capabilities of an Ethereum account. It does so by deploying unique, deterministically-addressed smart contract accounts for each NFT via a permissionless registry. Importantly, this proposal does not require any smart contract changes to the NFT smart contracts themselves. As a result, this proposal is backwards compatibile with all previously deployed NFTs.
+This proposal grants every NFT the full capabilities of an Ethereum account, allowing them to own assets and take actions. It does so by deploying unique, deterministically-addressed smart contract accounts for each NFT via a registry. Importantly, this proposal does not require any smart contract changes to the NFT smart contracts themselves. As a result, this proposal is backwards compatibile with all previously deployed NFTs.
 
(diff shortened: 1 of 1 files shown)


jaydenwindle:
Here is a summary of the changes included in this PR: https://ethereum-magicians.org/t/erc-6551-non-fungible-token-bound-accounts/13030/185?u=jay

Khademr7 (approved):

EIPS/eip-6551.md
```diff
@@ -81,44 +79,50 @@ Each token bound account proxy SHALL delegate execution to a contract that imple
 
 The registry contract is permissionless, immutable, and has no owner. The complete source code for the registry can be found in the [Registry Implementation](#registry-implementation) section below. The registry SHALL be deployed at address `TBD` using Nick's Factory (`0x4e59b44847b379578588920cA78FbF26c0B4956C`) with salt `0x6551655165516551655165516551655165516551655165516551655165516551`.
```
SamWilsn, line 80: Since this is going from Draft to Review, the EIP should be "complete". Having "TBD" here seems like it isn't complete?

It's a bit weird here since you won't know what the contract address is until the final contract is deployed, so I'm fine leaving this until you move into Last Call, but I wanted to point it out.

SamWilsn (approved):

EIPS/eip-6551.md
jaydenwindle: @SamWilsn thanks for highlighting this, I agree it's a bit of an edge case. We kept this as TBD in order to avoid confusion/migration issues when the address changes. We plan to release a finalized version of the registry and include the address in the proposal prior to moving to Last Call.

sullof: Whatever would be the ideal situation, it is very unlikely that there will be only a single ERC6551 Registry. I suggest you remove that requirement from the specification, which would also solve the issue raised by @SamWilsn

jaydenwindle: @sullof the inclusion of a well-known registry in the proposal was actually @SamWilsn's suggestion.

There can be many registry contracts (hence the "should" in [this section](https://github.com/ethereum/EIPs/blob/830708a049fc982fd595cb0c4dca703aebefd003/EIPS/eip-6551.md?plain=1#L138)), but applications are only required to support the well-known registry. Individual applications may support additional registries alongside the canonical registry if they choose to. The benefits of this approach are outlined in the rationale section and have been discussed at length on ETH Magicians.

So far there have been no use cases mentioned that would require a custom registry contract in order to use the token bound account pattern (with the notable exception of centralized registry control, which is antithetical to the ethos of this proposal). If any use cases are raised, the included registry can be updated to accommodate them.

PR #7441, opened 2023-08-01
Add EIP: Upgrade block proposer election to Whisk

dapplion:

 - The build passes.

### DIFF
--- a/EIPS/eip-7441.md
+++ b/EIPS/eip-7441.md
@@ -12,7 +12,7 @@ created: 2023-09-01
 
 ## Abstract
 
-Upgrades the block proposer election mechanism to Whisk, a single secret leader election (SSLE) protocol. Currently, block proposers are publicly known in advance, sufficiently to allow sequential DoS attacks that could disable Ethereum. This upgrade makes the next block proposer unknown until its block is published.
+Upgrades the block proposer election mechanism to Whisk, a single secret leader election (SSLE) protocol. Currently, block proposers are publicly known in advance, sufficiently to allow sequential DoS attacks that could disable Ethereum. This upgrade allows the next block proposer to remain secret until its block is published.
 
 ## Motivation
 


Pandapip1 (changes requested):
Please fix the issue with the title

YaroslavPysanskyi:
@YaroslavPysanskyi

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
+---
+eip: 7441
+title: Upgrade block proposer election to Whisk SSLE
```
abcoathup, line 3:
```suggestion
title: Upgrade block proposer election to Whisk
```
Suggestion to fit 44 character limit

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-7441-upgrade-block-proposer-election-to-whisk-ssle/15316
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-09-01
+---
+
+## Abstract
+
+Upgrades the block proposer election mechanism to Whisk, a single secret leader election (SSLE) protocol. Currently, block proposers are publicly known in advance, sufficiently to allow sequential DoS attacks that could disable Ethereum. This upgrade makes the next block proposer unknown until its block is published.
```
g11tech, line 15:
```suggestion
Upgrades the block proposer election mechanism to Whisk - a single secret leader election (SSLE) protocol. Currently block proposers are publicly known in advance sufficiently to allow sequential DoS attacks that could halt the Ethereum blockchain. This upgrade makes the next block proposer unknown until the current block is published.
```

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Motivation
+
+The beacon chain currently elects the next 32 block proposers at the beginning of each epoch. The results of this election are public and everyone gets to learn the identity of those future block proposers.
+
+This information leak enables attackers to launch DoS attacks against each proposer sequentially in an attempt to disable Ethereum.
```
g11tech, line 21:
```suggestion
This information leak enables attackers to launch DoS attacks against each proposer sequentially in an attempt to disable or halt Ethereum blockchain. Further it also allows collusion between block proposers and/or users.
```

there was an EIP which described how proposers and/or users can collude to avoid paying higher base fee and pocket more tips leading to irregular block size, will see which was that . this EIP can potentially address that scenario as well

EIPS/eip-7441.md
g11tech: yea it was correct. `compromise Ethereum's liveness` sounds great. updated the recommend. also would it be correct to replace `its block` with `the block` then? `its` sounds a bit confusing as to what it refers to

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Execution layer
+
+This requires no changes to the Execution Layer.
+
+### Consensus layer
+
+The protocol can be summarized in the following concurrent steps:
+
+- Validators register a tracker on their first proposal after the fork
```
g11tech, line 33:
```suggestion
- Validators register a unique tracker on their first proposal after the fork
```

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+## Specification
+
+### Execution layer
+
+This requires no changes to the Execution Layer.
+
+### Consensus layer
+
+The protocol can be summarized in the following concurrent steps:
+
```
g11tech, line 32:
```suggestion
- Proposals are now divided into shuffling phases of `WHISK_EPOCHS_PER_SHUFFLING_PHASE` epochs
```

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+- Validators register a tracker on their first proposal after the fork
+- At the start of a shuffling phase a list of candidate trackers is selected using public randomness from RANDAO
+- During one shuffling phase each proposer shuffles a subset of the candidate trackers using private randomness
```
g11tech, line 35:
```suggestion
- During a shuffling phase each proposer shuffles a subset of the candidate trackers using private randomness
```

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+- After one shuffling phase an ordered list of proposer trackers is selected from the candidate set using RANDAO
```
g11tech, line 36:
```suggestion
- After each shuffling phase an ordered list of proposer trackers is selected from the candidate set using RANDAO
```

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+Upgrades the block proposer election mechanism to Whisk, a single secret leader election (SSLE) protocol. Currently, block proposers are publicly known in advance, sufficiently to allow sequential DoS attacks that could disable Ethereum. This upgrade makes the next block proposer unknown until its block is published.
+
+## Motivation
+
+The beacon chain currently elects the next 32 block proposers at the beginning of each epoch. The results of this election are public and everyone gets to learn the identity of those future block proposers.
+
+This information leak enables attackers to launch DoS attacks against each proposer sequentially in an attempt to disable Ethereum.
+
+## Specification
+
```
g11tech, line 24: a `Parameters` section can be added to refer to constants relevant to describe the EIP for e.g. `WHISK_EPOCHS_PER_SHUFFLING_PHASE`

g11tech: a `<!-- TODO: add parameters. -->` kind of a TODO can be added which the bots can track post drafts so that the draft acceptance can become less burdensome

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+- Add `get_shuffle_indices` to compute pre-shuffle candidate selection
+- Add `process_shuffled_trackers` to submit shuffled candidate trackers.
+- Add `process_whisk` to block processing logic.
+- Modify `apply_deposit` to register an initial unique tracker and commitment without entropy.
+
+## Rationale
+
+### Fields per validator
+
+Whisk requires appending one tracker `(rG,krG)` and one commitment `kG` per validator. Both are updated at max once through a validator's lifetime.
```
g11tech, line 56:
```suggestion
Whisk requires having one unique tracker `(rG,krG)` and one commitment `kG` per validator. Both are updated at max once through a validator's lifetime whenever the validator gets to propose first block after the fork.
```

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+
+Whisk does identity binding by storing a commitment to the tracker's secret `kG` in the validator record. Storing the commitment also ensures the uniqueness of `k`.
+
+Alternatively, identity binding can be achieved by forcing the hash prefix of `hash(kG)` to match its validator index. However, validators would have to brute force `k` making bootstrap of the system harder for participants with fewer computational resources.
+
+Identity binding can also be achieved by setting `k = hash(nonce + pubkey)`. However, proposers will need to reveal `k` and be de-anonymized for repeated proposals on adjacent shuffling phases.
+
+### Alternative: non-single secret election
+
+Secret non-single leader election is based on protocol engineering rather than cryptography, thus much simpler and cheap than Whisk. However, it complicates the fork-choice and opens it up to potential MEV time-buying attacks, making it an unsuitable option at the time of writing.
```
g11tech, line 72:
```suggestion
Secret non-single leader election could be based on protocol engineering rather than cryptography, thus much simpler and cheaper than Whisk. However, it complicates the fork-choice and opens it to potential MEV time-buying attacks, making it an unsuitable option at the time of writing.
```

EIPS/eip-7441.md
dapplion: So the meaning is that a proposer is only revealed when specifically the block it is proposing is revealed, and not any other. The word "its" implied that relationship between what block carries the reveal, just "the block" does not specify which block.

EIPS/eip-7441.md
dapplion: With whisk, each proposer is aware of its future proposer slots much more in advance. So if there's collusion and proposers want to reveal such information to selected parties, Whisk makes the attack you describe easier.

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+
+PBS participants (e.g. builders) will not know the next proposer validator index to use a specific pre-registered fee recipient; unless the proposer chooses to reveal itself ahead of time. Block explorers and tooling will not be able to attribute missing slots to a specific validator index.
+
+## Security Considerations
+
+The shuffling strategy is analyzed in a companion paper and considered sufficiently safe for Whisk's use case. The data and computational complexity of this EIP are significant but constant, thus does not open new DoS vectors. 
+
+### Anonymity set
+
+The anonymity set in Whisk equals `WHISK_CANDIDATE_TRACKERS_COUNT - WHISK_PROPOSER_TRACKERS_COUNT` (8,192) which are the candidates that did not get selected as proposers. That count of validators corresponds to a smaller number of p2p nodes. Assuming a Pareto principle where "20% of the nodes run 80% of the validators" the anonymity corresponds to 2,108 nodes on average. A bigger candidate pool could make the shuffling strategy unsafe while shuffling more trackers per round would increase the cost of the ZK proofs.
```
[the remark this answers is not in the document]
dapplion: I had a version of the EIP defining all constants in this document, but there are so many, see https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/whisk/beacon-chain.md#constants

Should this be expressed in words to decouple it from those specific constants?

EIPS/eip-7441.md
dapplion: I'm not sure if this sentence is necessary. TBH this mini summary of the protocol is not necessary either. It does not provide any specification and is only a general overview for readers. Should it be removed altogether?

EIPS/eip-7441.md
dapplion: If we can avoid having to list of constants that would be better, it would duplicate a lot of content from the specs

EIPS/eip-7441.md
g11tech: aha this makes sense. could we change it to `This upgrade allows the next block proposer to stay secret and hidden until its block is published.` 

else current could also work but will provide less context to someone starting from zero background.

EIPS/eip-7441.md
g11tech: may be then we can add this in security considerations :)

EIPS/eip-7441.md
g11tech: so the flow of someone looking into this feature would be (or atleast how we want it to be) : Read EIP => Read consensus specs => get into client implementations if someone wants to follow through on this. So mini summary helps making that jump

keeping that in mind, what kind of mini summary would you deem good to fit here. else if we have that somewhere in consensus specs (`fork.md`?) then we can refer it here i guess

EIPS/eip-7441.md
```diff
@@ -0,0 +1,101 @@
...
+The protocol can be summarized in the following concurrent steps:
+
+- Validators register a tracker and unique commitment on their first proposal after the fork
+- At the start of a shuffling phase a list of candidate trackers is selected using public randomness from RANDAO
+- During each shuffling phase each proposer shuffles a subset of the candidate trackers using private randomness
+- After each shuffling phase an ordered list of proposer trackers is selected from the candidate set using RANDAO
+
+The full specification of the proposed change can be found in [`/_features/whisk/beacon-chain.md`](https://github.com/ethereum/consensus-specs/blob/a39abe388bc2d1abd5b4fd62fd18aed497956b30/specs/_features/whisk/beacon-chain.md). In summary:
+
+- Update `BeaconState` with fields needed to track validator trackers, commitments, and the two rounds of candidate election.
```
g11tech, line 40: may be we don't need this detail as its defined in the `full specification` already refered and could do away with this

EIPS/eip-7441.md
g11tech: i guess it makes sense , resolving this comment

EIPS/eip-7441.md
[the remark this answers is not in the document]
g11tech: no this is good for a person reading EIP to make easy transition to the specs... resolving the comment

EIPS/eip-7441.md
dapplion: Would be good define some guidelines on what this section is expected to include or not? Other editors can chip-in and @djrtwo

g11tech (dismissed):
lgtm,  discussions for how to best summarize EIP ideas can be done separately i guess.

g11tech:
@Pandapip1 is there anything pending from your side?

EIPS/eip-7441.md
dapplion: I don't think it's a significant change into the status quo security in this regard. With RANDAO based shuffling actors can still collude if they have proposer slots in a row

EIPS/eip-7441.md
dapplion: I don't think such summary would fit anywhere in the specs. Since it's derived information it mostly fits here or in a blog / etheresearch post introduction.

g11tech (approved):

PR #7444, opened 2023-08-01
Add EIP: Time Locks Maturity

thanhtrinh2003:
Add EIP: Timelock Maturity markdown file


bitcoinfinancier (approved):

sambacha:
this timelock should encode chainId information as well as explicitly state what timestamps are acceptable (ie only posix epoch timestamps).

jhweintraub:
> Why are we using bytes32 and not uint256? @jhweintraub would appreciate your input on such

Not every application uses `uint` as the identifier and a uint can be easily cast to a `bytes32` when needed. Given that the identifier is a hash of important information like `msg.sender`, `chainId`, etc. this is much easier to create and store. However I can be convinced the other way.

thanhtrinh2003:
> this timelock should encode chainId information as well as explicitly state what timestamps are acceptable (ie only posix epoch timestamps).

Will add the specific standard for timestamp use! However, could you please explain why it should encode the chainId information ?

xinbenlv (approved):

PR #7445, opened 2023-08-01
Update EIP-4788: Mention genesis block with no existing beacon block root case

holgerd77:
This just came up in a discussion on how to populate the new beacon block root header field for the case where Cancun starts at genesis and there is no beacon block root present for the genesis block.

### DIFF
--- a/EIPS/eip-4788.md
+++ b/EIPS/eip-4788.md
@@ -75,6 +75,8 @@ Validity of the parent beacon block root is guaranteed from the consensus layer,
 
 When verifying a block, execution clients **MUST** ensure the root value in the block header matches the one provided by the consensus client.
 
+For a genesis block with no existing parent beacon block root the 32 zero bytes are used as a root placeholder.
+
 ### EVM changes
 
 #### Block processing


jochem-brouwer (approved):

EIPS/eip-4788.md
```diff
@@ -75,6 +75,8 @@ Validity of the parent beacon block root is guaranteed from the consensus layer,
 
 When verifying a block, execution clients **MUST** ensure the root value in the block header matches the one provided by the consensus client.
 
+For a genesis block with no existing parent beacon block root the 32 zero bytes are used as a root placeholder.
```
ralexstokes, line 78:
```suggestion
For a genesis block with no existing parent beacon block root, the root consisting of 32 zero bytes should be used in the block header.
```

ralexstokes (approved):
lgtm!

EIPS/eip-4788.md
holgerd77: Ah, now the old version still made it into the merged spec, was this intended? 🤔

PR #7451, opened 2023-08-02
Update EIP-6780: Move to Review

gballet:


SamWilsn (commented):
I've opened #7460 to withdraw EIP-6046, but the faster solution is to just remove the reference to it here. Otherwise this looks good.

SamWilsn:
Looks like this PR has no changes after updating the branch, so I'm going to close it. Feel free to re-open if I am mistaken.

PR #7453, opened 2023-08-03
Update EIP-6963: use data URI for icon value in Wallet Provider implementation example

jiexi:
Spec was updated to specify that `icon` must be data URI in [32fbcd8d7422958a9afe51f6ef02e96044af8594](https://github.com/ethereum/EIPs/commit/32fbcd8d7422958a9afe51f6ef02e96044af8594)

This updates the Wallet Provider implementation example to use a data uri for its `icon` value


kdenhartog (approved):
Seems like a good idea to have the non-normative example be an SVG if that's what we want to recommend in the spec.

Issue #167, opened 2016-10-29
Scratch space


vbuterin:
### Specification

If `block.number >= FORK_BLKNUM`, add two precompile contracts at addresses 0x0e and 0x0f. Both contracts shard 1 gas per byte of calldata plus 20 base gas cost. The contracts work as follows:
- `0x0e`: All bytes of the calldata except for the first 32 are saved in a new key/value map using `sha3(sender) xor sha3(target)` as a key where `target` is the address in the first 32 bytes and `sender` is the sender of the call.
- `0x0f`: Immediately return the calldata saved in the key/value map using `sha3(sender) xor sha3(target)` as a key where `target` is the address in the first 32 bytes of calldata as a key and `sender` is the sender of the call.

The key/value map is reset at the end of processing every transaction.
### Rationale

This allows for certain kinds of applications to be built more cheaply, safely and efficiently. For example, consider the use case of sending ERC20 tokens along with a call. The current workflow is:
1. Sender authorizes token to withdraw funds
2. Sender calls contract
3. Contract calls token to withdraw funds, and checks the response to make sure that the withdrawal succeeded. If it did then it continues executing.

However, this is ugly, requiring ERC20 tokens to have a complex withdrawal feature, and is gas-inefficient, requiring four SSTOREs: (i) creating the withdrawal authorization, (ii) decreasing sender's ERC20 token balance, (iii) increasing the contract's ERC20 token balance, (iv) removing the withdrawal authorization. Also, it requires a call to the token contract as the first operation of the contract's code, before other state changes; if the token is untrusted this may lead to re-entrancy vulnerabilities.

Using scratch space, we can change the mechanism to the following:
1. Sender calls token to send funds to the contract. Token saves in the recipient's scratch space that this was the last transaction that took place.
2. Sender calls contract
3. Contract checks scratch space and makes sure that the last transaction of the token sent the correct number of coins, from the sender, to their account. If the check passed, the contract empties the scratch space and continues executing.

This now only requires two SSTOREs, to do the actual balance change, and so has a gas cost that is proportional to the actual cost of what took place. Re-entrancy is no longer an issue.

[...]

PR #7456, opened 2023-08-04
Update EIP-4788: initial stab at v2

lightclient:
Based on the conversation in [ACD #167](https://github.com/ethereum/pm/issues/836), I wanted to take a look at rewriting this EIP to use a regular contract instead of a precompile. The main motivating factor for this is to reduce the surface of consensus failures. Instead of `N` implementations of the precompile, we can rely on the same EVM logic as we do for any other contract execution and execute a standardized bytecode implementation of the beacon roots contract.

The changes are as follows:

* require `timestamp` input to be _exactly_ 32 bytes
* revert if `timestamp` input does not match stored value (instead of returning zeroed word)
* remove precompile concept, use regular smart contract with provided bytecode

The main open question here is where and how to deploy the code for the beacon roots contract. As I see it, there are three options:

1) deploy as a normal contract
2) predeploy the code via enshirement during the hard fork to the same address as the original precompile (`0x0b`)
3) predeploy the code to an arbitrary address in the address space (maybe a create/create2 calc'd address?)

I prefer option 1) as it aligns with the original motivation of the v2 change, which is to minimize the surface of consensus failures and reuse EVM infrastructure where possible. The addresses in the EIP were generated using [this](https://gist.github.com/lightclient/668353c6ba0d85ce901edb7048bd1562) script.

--

I have also taken a stab at writing an assembly version of the beacon roots contract. It can be found here: https://github.com/lightclient/4788asm

I think that is roughly the minimal bytecode representation of the logic we wish to enshrine.


EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
 
 For a genesis block with no existing parent beacon block root the 32 zero bytes are used as a root placeholder.
 
-### EVM changes
+#### Beacon roots contract
 
-#### Block processing
+The beacon roots contract has two operations: `get` and `set`. The input itself is not used to determine which function to execute, for that the result of `caller` is used. If `caller` is equal to `SYSTEM_ADDRESS` then the operation to perform is `set`. Otherwise, `get`.
```
djrtwo, line 84: oh, funny. seems okay though

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
 
-At the start of processing any execution block where `block.timestamp >= FORK_TIMESTAMP` (i.e. before processing any transactions),
-write the parent beacon root provided in the block header into the storage of the contract at `HISTORY_STORAGE_ADDRESS`.
+##### `get`
 
-In order to bound the storage used by this precompile, two ring buffers are used: one to track the latest timestamp at a given index in the ring buffer and another to track
-the latest root at a given index.
+* If `caller` is equal to `SYSTEM_ADDRESS`, the contract must revert.
```
djrtwo, line 88: This seems unnecessary. The outer of `get`/`set` does the routing *based* on the `caller` so by definition the `caller` won't be `SYSTEM_ADDRESS`

(same for parallel in `set`)

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+* Finally, the beacon root associated with the timestamp is accessed at returned to the user. It is stored at `timestamp % HISTORICAL_ROOTS_MODULUS + HISTORICAL_ROOTS_MODULUS`.
 
-To derive the index `timestamp_index` into the timestamp ring buffer, the timestamp (a 64-bit unsigned integer value) is reduced modulo `HISTORICAL_ROOTS_MODULUS`.
-To derive the index `root_index` into the root ring buffer, add `HISTORICAL_ROOTS_MODULUS` to the index into the timestamp ring buffer.
-Both resulting 64-bit unsigned integers should be encoded as 32 bytes in big-endian format when writing to the storage.
+##### `set`
 
-The timestamp from the header, encoded as 32 bytes in big-endian format, is the value to write behind the `timestamp_index`.
-The 32 bytes of the `parent_beacon_block_root` (as provided) are the value to write behind the `root_index`.
+* If `caller` is not equal to `SYSTEM_ADDRESS`, the contract must revert.
```
djrtwo, line 97: ditto

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+* Caller provides the parent beacon block root as calldata to the contract.
+* Set the storage value at `header.timestamp % HISTORICAL_ROOTS_MODULUS` to be `header.timestamp`
```
djrtwo, line 99: I kind of want to make the case for avoiding  a system-TX and just setting these storage values at the top of the block. Then the code is just `get` and we don't have to test/specify how to do system-TXs 

weaker held opinion than the deploy method

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+  "type": "0x2",
+  "chainId": "0x1",
+  "nonce": "0x0",
+  "to": null,
+  "gas": "0xd4f8",
+  "gasPrice": null,
+  "maxPriorityFeePerGas": "0x9c7652400",
+  "maxFeePerGas": "0xe8d4a51000",
+  "value": "0x0",
+  "input": "0x605a8060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604257602036146024575f5ffd5b620180005f350680545f351415603d576201800001545f525b60205ff35b42620180004206555f3562018000420662018000015500",
```
djrtwo, line 220: I might put many of these values behind a VAR -- e.g. input and the stubbed `r`, `s`

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+  "accessList": [],
+  "v": "0x0",
+  "r": "0x539",
```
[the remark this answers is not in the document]
djrtwo: > Although this style of contract creation is not tied to any specific initcode like create2 is, the synthetic address is cryptographically bound to the input data of the transaction (e.g. the initcode).

and then we don't have to think about things like this.

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+add
+sstore
+stop
+```
 
-If the `timestamp` **does** match, the client **MUST** read the root from the contract storage and return those 32 bytes in the caller's return buffer.
+#### Deployment
+
+The beacon roots contract is deployed like any other smart contract. A special synthetic address is generated
+by working backwards from the desired deployment transaction:
```
djrtwo, line 207: This should be specified as "at the first block where  `block.timestamp >= FORK_TIMESTAMP`, this deploy TX is transparently inserted at the start of the block" right?

Right now the spec doesn't say if/when to deploy the bytecode via this method

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
-else:
-    timestamp_extended = timestamp_reduced + HISTORICAL_ROOTS_MODULUS
-    root_index = to_uint256_be(timestamp_extended)
-    root = sload(HISTORY_STORAGE_ADDRESS, root_index)
-    evm.returndata[:32].set(root)
-```
+* the call must execute to completion, therefore the available gas can be considered as infinite
+* the call does not count against the block's gas limit
+* the call does not follow the [EIP-1559](./eip-1559.md) burn semantics - no value should be transferred as part of the call
+* if no code exists at `BEACON_ROOTS_ADDRESS`, the call must fail silently
```
djrtwo, line 238: Assuming the prior section specifies when to deploy the bytecode, this bullet will become unnecessary

djrtwo (commented):
nice!

EIPS/eip-4788.md
```diff
@@ -77,76 +79,165 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+```json
+{
+  "type": "0x2",
+  "chainId": "0x1",
+  "nonce": "0x0",
+  "to": null,
+  "gas": "0xd4f8",
+  "gasPrice": null,
+  "maxPriorityFeePerGas": "0x9c7652400",
+  "maxFeePerGas": "0xe8d4a51000",
```
jochem-brouwer, line 218: The semi-problem with this fake sig tx is (as raised by @holiman in a discussion we had), that if you want to deploy on other chains, all items are fixed, so also these gas fees, which might be problematic to deploy it on another chain.

EIPS/eip-4788.md
jochem-brouwer: This somewhat seems to imply that this contract has an ABI (which it has not, at least not in current assembly). I would really like to do the ABI-like approach as in https://github.com/lightclient/4788asm/pull/5 since this will also be very helpful for solidity users (you can now `BeaconRootContract(address).get(timest)`.

EIPS/eip-4788.md
lightclient: We could do make uint256 for `maxFeePerGas` if we use an EIP-1559 tx. Ultimately it isn't important how other chains deploy the opcode. The EIP specifies the mainnet behavior, but I assume the address will ultimately become a genesis configuration option much like the deposit contract. Chains could easily etch the code directly into the state by an irregular state transition or via genesis enshrinement.

EIPS/eip-4788.md
lightclient: I think it is a bad idea to enshrine solidity ABI behavior. It is ubiquitous today, but in the future I expect other language will have different calling conventions. Solidity should directly support this contract like they do for `ecrecover`.

EIPS/eip-4788.md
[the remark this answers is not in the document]
lightclient: >  rather than trying to use a TX of sorts but without gas or signature verification which requires a lot more exceptional logic

To be clear, this tx does still abide by gas and signature verification. The signature is constructed arbitrarily and therefore the sender is a one-time sender address (pk is not known).

--

In an effort to minimize protocol baggage, I think it is better to deploy the contract using standard methods. Once we are satisfied with the bytecode we'd like to deploy, it could be deployed by anyone. We simply have to agree on the address at which it is deployed. It doesn't have to be a synthetic tx, although I do slightly prefer this method as it is permissionless (anyone can fund the account and submit the tx).

EIPS/eip-4788.md
lightclient: good point - updated

EIPS/eip-4788.md
lightclient: Submission of the deployment tx is permissionless and can occur at anytime before or after fork.

EIPS/eip-4788.md
jochem-brouwer: Ok that actually makes a lot of sense!

EIPS/eip-4788.md
lightclient: As mentioned in other thread, I don't think it is important to specify the deployment. But I suppose the bullet is unnecessary because that is the semantics of an evm call anyways.

EIPS/eip-4788.md
lightclient: I think if we're going to use a standard contract we should embrace it and allow it to have storing functionality.

EIPS/eip-4788.md
jochem-brouwer: (Small side note: we actually cannot do max uint256 since this violates the balance check of 1559. But this does not matter because you are right, other chains can just "expand" this EIP to just deploy it anywhere as a special state transition)

lightclient: good point 🫠

EIPS/eip-4788.md
```diff
@@ -77,76 +79,163 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
-write the parent beacon root provided in the block header into the storage of the contract at `HISTORY_STORAGE_ADDRESS`.
+##### `get`
 
-In order to bound the storage used by this precompile, two ring buffers are used: one to track the latest timestamp at a given index in the ring buffer and another to track
-the latest root at a given index.
+* Callers provide the `timestamp` they are querying encoded as 32 bytes in big-endian format.
+* If the input is not exactly 32 bytes, the contract must revert.
+* Given `timestamp`, the contract computes the storage index in which the timestamp is stored by computing the modulo `timestamp % HISTORICAL_ROOTS_MODULUS` and reads the value.
+* If the `timestamp` does not match, the contract must revert.
+* Finally, the beacon root associated with the timestamp is accessed at returned to the user. It is stored at `timestamp % HISTORICAL_ROOTS_MODULUS + HISTORICAL_ROOTS_MODULUS`.
```
ralexstokes, line 92: ?


```suggestion
* Finally, the beacon root associated with the timestamp is returned to the user. It is stored at `timestamp % HISTORICAL_ROOTS_MODULUS + HISTORICAL_ROOTS_MODULUS`.
```

EIPS/eip-4788.md
```diff
@@ -77,76 +79,163 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+```json
+{
+  "type": "0x2",
+  "chainId": "0x1",
+  "nonce": "0x0",
+  "to": null,
+  "gas": "0xd4f8",
+  "gasPrice": null,
+  "maxPriorityFeePerGas": "0x9c7652400",
+  "maxFeePerGas": "0xe8d4a51000",
```
ralexstokes, line 216: where do these gas values come from?

to echo what others have said, it seems fragile to enshrine a "synthetic transaction"

easier to just drop byte code a la EIP-1011

lightclient: Just arbitrary values that would be included on mainnet. It isn't really enshrining the tx. I would think more of it as two steps: 1) deploy contract 2) set address in clients to call as part of pre block processing. A synthetic tx is just one way of deploying the contract. I think it is best because it is clear, transparent, and permissionless. We could just as easily have someone hand deploy the contract before the fork (now even, if the byte code is ready) and then set the address in client configurations.

EIPS/eip-4788.md
holiman: Wrote this already in discord, posting here too for visibility. 

With the spec being 'system call', it means a reference client can just do a simple call. A client implementor can still choose not to, and instead bypass the whole call and update the values directly. 

If the eip standardizes 'direct update', however, and omits the system-update path from the contract, then we remove all optionality, and force more special-case code into clients

Therefore I think the 'system call' approach is the best way forward.

holiman: I think the EIP should mention that "direct update" is a valid/possible client implementation/optimization detail.

holiman:
I also think that the 'get'-path should have an abi signature: `getBeaconRoot(uint256)`, thus takign `36` bytes input. 
It can either _check_ the abi, or if we for some reason want to, we can omit the signature-check and ignore the first four bytes. But I don't see any reason to be overly "clever". 

The ABI encoding is ubiquitous, not merely solidity, it's the calling-convention used in the ethereum application layer, and I think it makes sense to have the precompile use that callpattern. 

With abi, even with zero compiler support, this is possible: 

```javascript
root = BeaconRoot( address ).getBeaconRoot( tstamp );
``` 
As opposed to something like this: 
```javascript
address.call().input( to_bytes32(tstamp ) ).output(0,32)
```
(I'm happy to update that example, I'm not really sure of the syntax for doing a call that way. )

EIPS/eip-4788.md
g11tech: sounds like a hack imo

EIPS/eip-4788.md
g11tech: having a system-tx would be a good generic functionality to have in long term imo

holiman:
In the original EIP, the contract was a precompile, and this 'subsidized' such that for call-purposes, the address is considered already included in the transaction access list. The difference is, as far as I can tell, `700` instead of `2600` gas. 

I don't really care which way we go, but the "least changes to the original eip" route suggests that the hardfork should add the address to the access list.

EIPS/eip-4788.md
```diff
@@ -77,76 +79,163 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
-    root = sload(HISTORY_STORAGE_ADDRESS, root_index)
-    evm.returndata[:32].set(root)
-```
+* the call must execute to completion, therefore the available gas can be considered as infinite
+* the call does not count against the block's gas limit
+* the call does not follow the [EIP-1559](./eip-1559.md) burn semantics - no value should be transferred as part of the call
+* if no code exists at `BEACON_ROOTS_ADDRESS`, the call must fail silently
 
-The precompile costs `G_beacon_root` gas to reflect the two (2) implicit `SLOAD`s from the precompile's state.
+Client may decide to omit an explicit EVM call and directly set the storage values.
```
LukaszRozmej, line 238: This would make any potential modifications to the contract in the future harder or support some other contract on a different chain.

holiman: Yes. This sentence should have some YMMV warnings: doing this may be a workable implementation of the mainnet contract, but might lead to issues on other networks, if some other contract is used. 

(The case where no contract at all exists should be fine, since setting the values directly doesn't affect the state -- the slots are cleaned up due to the `empty`-rule at the end of block anyway. )

Perhaps something like this? 

```suggestion
Clients may decide to omit an explicit EVM call and directly set the storage values. Note: While this is a valid optimization for Ethereum mainnet, it could be problematic on non-mainnet situations in case a different contract is used.
```

yperbasis:
> In the original EIP, the contract was a precompile, and this 'subsidized' such that for call-purposes, the address is considered already included in the transaction access list. The difference is, as far as I can tell, `700` instead of `2600` gas.
> 
> I don't really care which way we go, but the "least changes to the original eip" route suggests that the hardfork should add the address to the access list.

Yes, adding the address to the access list (from Cancun) makes sense.

EIPS/eip-4788.md
```diff
@@ -77,76 +79,163 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
```
pinges: I think that we have to decide whether we want to do the EVM call or to directly set the storage values. I understand that when we set the value directly and the the account is empty we do remove the account at the end of the block, but there are the edge case where someone puts some eth into the account, which means it is not empty anymore. 
I would prefer to set the values directly, as that is 10x faster and we do not require all clients to implement the system operation.

pinges (changes requested):

EIPS/eip-4788.md
holiman: > I think that we have to decide whether we want to do the EVM call

The consensus-definition is that an EVM call is made. A client implementor may choose to optimize this, and do a direct update instead. This choice may cause incompatibilities in non-mainnet scenarios, e.g. if the code is replaced with different code. 

> I understand that when we set the value directly and the the account is empty we do remove the account at the end of the block, but there are the edge case where someone puts some eth into the account, which means it is not empty anymore.

No, there is now code on the account, so it would not be removed, since it is not `empty`, according to the definition of `empty`. 

> I would prefer to set the values directly, as that is 10x faster and we do not require all clients to implement the system operation.

You are at liberty to do so, since "we do not require all clients to implement the system operation" -- however, the **consensus correct behaviour, in any given scenario, is given by the system-call behaviour**

IMO this is a pretty good middle ground.

EIPS/eip-4788.md
```diff
@@ -77,76 +79,164 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+  "to": null,
+  "gas": "0x27eac",
+  "gasPrice": "0xe8d4a51000",
+  "maxPriorityFeePerGas": null,
+  "maxFeePerGas": null,
+  "value": "0x0",
+  "input": "0x60588060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5ffd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b6201800042064281555f359062018000015500",
+  "v": "0x1b",
+  "r": "0x539",
+  "s": "0x1337",
```
holiman, line 221: Alternative suggestion: 
```suggestion
  "s": "0x133700cb4fbb1368012b31",
```

Then we get a `bEaC0` -prefixed address `0xbEaC0B09a6ed668e6BC5d29e29013aCb9286bF5F` with the transaction
```json
{
  "type": "0x0",
  "nonce": "0x0",
  "to": null,
  "gas": "0x27eac",
  "gasPrice": "0xe8d4a51000",
  "maxPriorityFeePerGas": null,
  "maxFeePerGas": null,
  "value": "0x0",
  "input": "0x60588060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5ffd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b6201800042064281555f359062018000015500",
  "v": "0x1b",
  "r": "0x539",
  "s": "0x133700cb4fbb1368012b31",
  "hash": "0xa71141ed853de6fdd7c9d9355293f255011e9e512cb833a03b3ff20d8045f5b9"
}
```

holiman: Update: `BeAC00` - prefix:

Address: `0xBeAC00541d49391ED88ABF392bfC1F4dEa8c4143`
Tx:
```json
{
  "type": "0x0",
  "nonce": "0x0",
  "to": null,
  "gas": "0x27eac",
  "gasPrice": "0xe8d4a51000",
  "maxPriorityFeePerGas": null,
  "maxFeePerGas": null,
  "value": "0x0",
  "input": "0x60588060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5ffd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b6201800042064281555f359062018000015500",
  "v": "0x1b",
  "r": "0x539",
  "s": "0x1337007db5d301ac899c9e",
  "hash": "0x6923750e3b16ef9bfe1a70820d0d0628d7d4e84906b19ef91840fe913d688759"
}
```

lightclient: amazing!

lightclient: Ah wait did you intend for this to be the sender of the transaction or the address of the deployed contract? Because I am recovering `0xBeAC00541d49391ED88ABF392bfC1F4dEa8c4143` as the EOA address and `0xC221266EE238DbA34c6df82f75516a27087C0Bd9` as the deployed address.

holiman: Ah silly me. Let me get back to you on this

holiman: How about this then? 
_deploy_ address: `0xbEac00dDB15f3B6d645C48263dC93862413A222D`

```json
{
  "type": "0x0",
  "nonce": "0x0",
  "to": null,
  "gas": "0x27eac",
  "gasPrice": "0xe8d4a51000",
  "maxPriorityFeePerGas": null,
  "maxFeePerGas": null,
  "value": "0x0",
  "input": "0x60588060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5ffd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b6201800042064281555f359062018000015500",
  "v": "0x1b",
  "r": "0x539",
  "s": "0x133700f3a77843802897db",
  "hash": "0x14789a20c0508b81ab7a0287a12a3a41ca960aa18244af8e98689e37ed569f07"
}
```

EIPS/eip-4788.md
```diff
@@ -77,76 +79,162 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+    storage.set(timestamp_idx, evm.timestamp)
+    storage.set(root_idx, evm.calldata)
+```
 
-If the `timestamp` **does NOT** match, the client **MUST** return the "zero" word -- the 32-byte value where each byte is `0x00`.
+##### Bytecode
+
+The exact initcode to deploy is shared below.
+
+```asm
```
g11tech, line 135: i think it would be really nice and convenient to have bytecode hex as well for below to leave no room for any mismatch in client

EIPS/eip-4788.md
```diff
@@ -77,76 +79,164 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
+  "input": "0x60588060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5ffd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b6201800042064281555f359062018000015500",
+  "v": "0x1b",
+  "r": "0x539",
+  "s": "0x133700f3a77843802897db",
+  "hash": "0x14789a20c0508b81ab7a0287a12a3a41ca960aa18244af8e98689e37ed569f07"
+}
+```
 
-If the `timestamp` **does** match, the client **MUST** read the root from the contract storage and return those 32 bytes in the caller's return buffer.
+The sender of the transaction can be calculated as `0x3A700F76Ddec97c4640FdAF5a0fF1632E50A10D8`. The address of the first contract deployed from the account is `rlp([sender, 0])` which equals `0xBeAC00541d49391ED88ABF392bfC1F4dEa8c4143`. This is how `BEACON_ROOTS_ADDRESS` is determined. Although this style of contract creation is not tied to any specific initcode like create2 is, the synthetic address is cryptographically bound to the input data of the transaction (e.g. the initcode).
```
holiman, line 226:
```suggestion
The sender of the transaction can be calculated as `0x3e266d3c3a70c238bdddafef1ba06fbd58958d70`. The address of the first contract deployed from the account is `rlp([sender, 0])` which equals `0xbEac00dDB15f3B6d645C48263dC93862413A222D`. This is how `BEACON_ROOTS_ADDRESS` is determined. Although this style of contract creation is not tied to any specific initcode like create2 is, the synthetic address is cryptographically bound to the input data of the transaction (e.g. the initcode).
```

jaydub74:
James Carroll

EIPS/eip-4788.md
```diff
@@ -25,18 +26,19 @@ restaking constructions, smart contract bridges, MEV mitigations and more.
...
-|---                           |---                                           |---
+| constants                    | value                                        |
+|---                           |---                                           |
 | `FORK_TIMESTAMP`             | TBD                                          |
-| `HISTORY_STORAGE_ADDRESS`    | `Bytes20(0xB)`                               |
-| `G_beacon_root`              | 4200                                         | gas
-| `HISTORICAL_ROOTS_MODULUS`   | 98304                                        |
+| `HISTORICAL_ROOTS_MODULUS`   | `98304`                                      |
+| `SYSTEM_ADDRESS`             | `0xfffffffffffffffffffffffffffffffffffffffe` |
+| `BEACON_ROOTS_ADDRESS`       | `0xBeAC00541d49391ED88ABF392bfC1F4dEa8c4143` |
```
holiman, line 34:
```suggestion
| `BEACON_ROOTS_ADDRESS`       | `0xbEac00dDB15f3B6d645C48263dC93862413A222D` |
```

holiman:
> BEACON_ROOTS_ADDRESS should be 0xBeAC00541d49391ED88ABF392bfC1F4dEa8c4143 now AFAIK

@yperbasis nope, should be `0xbEac00dDB15f3B6d645C48263dC93862413A222D`. Sorry for the mixups

holiman (approved):
LGTM

EIPS/eip-4788.md
```diff
@@ -77,76 +79,164 @@ When verifying a block, execution clients **MUST** ensure the root value in the
...
-```python
-timestamp = evm.calldata[:32]
-if len(timestamp) != 32:
-    evm.revert()
-    return
+At the start of processing any execution block where `block.timestamp >= FORK_TIMESTAMP` (i.e. before processing any transactions), call `BEACON_ROOTS_ADDRESS` as `SYSTEM_ADDRESS` with the 32-byte input of `header.parent_beacon_block_root`. This will trigger the `set()` routine of the beacon roots contract. This is a system operation and therefore:
 
-timestamp_reduced = to_uint64_be(timestamp) % HISTORICAL_ROOTS_MODULUS
-timestamp_index = to_uint256_be(timestamp_reduced)
+* the call must execute to completion, therefore the available gas can be considered as infinite
```
holiman, line 232: Why? 
I would rather specify some hard limit, like `30M` gas, as suggested by @yperbasis . I suspect most actual implementation will need to set a limit anyway.

EIPS/eip-4788.md
holiman: This bullet is unnecessary, other than as clarification: any non-value call to account with non-existing code has no discernible effect on state.  
If you prefix the bullet with `Note:` or `Clarification:`, it becomes apparent that this bullet does not intend to add any behavioural changes.

holiman:
Why no merge? I thought this was mostly agreed upon on by ACD?

yperbasis:
Please change the gas limit to 30M. Geth, Erigon, and probably Nethermind have implementations with the gas limit of 30M.

holiman:
I agree with @yperbasis . To clarify what I said earlier: 

>  I suspect most actual implementation will need to set a limit anyway.

Doing metered calls is the natural way that clients implement calls. Doing an "infinite-gas" call, or "unmetered call" is _not_ a common thing, and it wouldn't surprise me is there are clients which do not even support unmetered calls. So _some_ limit needs to be set, in those cases, and in many cases `maxuint64` is the closes to infinite gas there is. So let's just set it to something and avoid the confusion.

ralexstokes:
> Why no merge? I thought this was mostly agreed upon on by ACD?

I see two outstanding changes we want to make:

1) Set gas limit of the "system execution" to 30M
2) change `HISTORICAL_ROOTS_MODULUS` to `HISTORY_BUFFER_LENGTH` (following a review w/ @djrtwo )

ralexstokes:
here's a PR to address the gas limit:

> Set gas limit of the "system execution" to 30M

https://github.com/lightclient/EIPs/pull/4

ralexstokes (approved):
lots of work has gone into this update, thanks everyone for taking a look and contributing!

going to go ahead and merge, if we need further tweaks we can update in future PRs

yperbasis:
@lightclient Section "Gas cost of precompile" is obsolete now and should be removed. Also, "precompile" is still mentioned in a couple of other places in the Rationale.

EIPS/eip-4788.md
ritzdorf: @holiman But shouldn't it then rather say:

* if no code exists at `BEACON_ROOTS_ADDRESS`, the call must succeed as with any other account without code

I find "fail" a bit misleading here.

namiloh: I agree: fail or succeed depends on your point of view. Whole bullet is moot and should not exist

PR #7459, opened 2023-08-08
Add EIP: Gasless Transactions

JKincorperated:


EIPS/eip-TBD.md
```diff
@@ -0,0 +1,50 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: TBD
```
An EIP number won't be issued until there is a specification.

PR #7464, opened 2023-08-08
Update EIP-6963: Add RDNS section for motivation and requirements

everdimension:
This update adds a section to provide requirements and motivation for the `rdns` property.

I was originally planning to add example UI flows where a DApp would use this value, but I'm not sure this belongs in the spec. I'm thinking that saying that the DApps can rely on this property being stable might be enough.

### DIFF
--- a/EIPS/eip-6963.md
+++ b/EIPS/eip-6963.md
@@ -80,7 +80,7 @@ The `icon` string MUST be a data URI as defined in [RFC-2397]. The image SHOULD 
 The **`rdns`** (Reverse-DNS) property serves to provide an identifier which DApps can rely on to be stable between sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions.
 The Reverse-DNS convention implies that the value should start with a reversed DNS domain name controlled by the Provider. The domain name should be followed by a subdomain or a product name. Example: `com.example.MyBrowserWallet`.
 
-- The DNS part of the `rdns` value MUST BE a valid [RFC-1034] Domain Name;
+- The `rdns` value MUST BE a valid [RFC-1034] Domain Name;
 - The DNS part of the `rdns` value SHOULD BE an active domain controlled by the Provider;
 - DApps MAY reject the Providers which do not follow the Reverse-DNS convention correctly;
 - DApps SHOULD NOT use the `rnds` value for feature detection as these are self-attested and prone to impersonation or bad incentives without an additional verification mechanism; feature-discovery and verification are both out of scope of this interface specification.
--- a/EIPS/eip-6963.md
+++ b/EIPS/eip-6963.md
@@ -83,7 +83,7 @@ The Reverse-DNS convention implies that the value should start with a reversed D
 - The DNS part of the `rdns` value MUST BE a valid [RFC-1034] Domain Name;
 - The DNS part of the `rdns` value SHOULD BE an active domain controlled by the Provider;
 - DApps MAY reject the Providers which do not follow the Reverse-DNS convention correctly;
-- DApps MUST NOT use the `rnds` value for feature detection;
+- DApps SHOULD NOT use the `rnds` value for feature detection as these are self-attested and prone to impersonation or bad incentives without an additional verification mechanism; feature-discovery and verification are both out of scope of this interface specification.
 
 ### Provider Detail
 
--- a/EIPS/eip-6963.md
+++ b/EIPS/eip-6963.md
@@ -75,6 +75,16 @@ data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12
 
 The `icon` string MUST be a data URI as defined in [RFC-2397]. The image SHOULD be squared with 96x96px minimum resolution. The image format is RECOMMENDED to be either lossless or vector based such as PNG, WebP or SVG to make the image easy to render on the DApp. Since SVG images can execute Javascript, applications and libraries MUST render SVG images using the `<img>` tag to ensure no untrusted Javascript execution can occur.
 
+#### RDNS
+
+The **`rdns`** (Reverse-DNS) property serves to provide an identifier which DApps can rely on to be stable between sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions.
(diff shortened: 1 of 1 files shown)


EIPS/eip-6963.md
```diff
@@ -75,6 +75,16 @@ data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12
...
 
+#### RDNS
+
+The **`rdns`** (Reverse-DNS) property serves to provide an identifier which DApps can rely on to be stable between sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions.
+The Reverse-DNS convention implies that the value should start with a reversed DNS domain name controlled by the Provider. The domain name should be followed by a subdomain or a product name. Example: `com.example.MyBrowserWallet`.
+
+- The DNS part of the `rdns` value MUST BE a valid [RFC-1034] Domain Name;
+- The DNS part of the `rdns` value SHOULD BE an active domain controlled by the Provider;
+- DApps MAY reject the Providers which do not follow the Reverse-DNS convention correctly;
+- DApps MUST NOT use the `rnds` value for feature detection;
```
everdimension, line 86: Should the spec also prescribe against prioritization?
For example, a DApp with a Connect Screen that chooses to show some wallet options at the top, and chooses to hide other wallets behind the "show more" button.
Such UI patterns _may_ lead to the "browser wars" problem, which this particular proposal is trying to avoid in the first place.

bumblefudge: I believe that falls under "economic behaviors" and definitely hard to banish with a MUST NOT.  I would say it's better to phrase any guidance in technical terms and, at most, remind implementers that such design decisions create incentives for bad security, i.e. impersonation.  For example:

```suggestion
- DApps SHOULD NOT use the `rnds` value for feature detection as these are self-attested and prone to impersonation or bad incentives without an additional verification mechanism; feature-discovery and verification are both out of scope of this interface specification.
```

EIPS/eip-6963.md
```diff
@@ -75,6 +75,16 @@ data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12
 
 The `icon` string MUST be a data URI as defined in [RFC-2397]. The image SHOULD be squared with 96x96px minimum resolution. The image format is RECOMMENDED to be either lossless or vector based such as PNG, WebP or SVG to make the image easy to render on the DApp. Since SVG images can execute Javascript, applications and libraries MUST render SVG images using the `<img>` tag to ensure no untrusted Javascript execution can occur.
 
+#### RDNS
+
+The **`rdns`** (Reverse-DNS) property serves to provide an identifier which DApps can rely on to be stable between sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions.
```
bumblefudge, line 80:
```suggestion
The **`rdns`** (Reverse-DNS) property serves to provide an identifier which DApps can rely on to be stable between sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions and to organize the namespace of these identifiers intuitively for developers.
```

EIPS/eip-6963.md
```diff
@@ -75,6 +75,16 @@ data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12
...
+The Reverse-DNS convention implies that the value should start with a reversed DNS domain name controlled by the Provider. The domain name should be followed by a subdomain or a product name. Example: `com.example.MyBrowserWallet`.
+
+- The DNS part of the `rdns` value MUST BE a valid [RFC-1034] Domain Name;
```
bumblefudge, line 83:
```suggestion
- The `rdns` value MUST BE a valid [RFC-1034] Domain Name;
```
I feel like exempting the "subdomain or product name" segment from RFC-1034 just invites problems, while requiring product names to be encoded as valid subdomains will be a stitch in time if future verification specs define garden-variety `/well-known/` lookups :D

EIPS/eip-6963.md
```diff
@@ -75,6 +75,16 @@ data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12
...
+- The DNS part of the `rdns` value SHOULD BE an active domain controlled by the Provider;
```
bumblefudge, line 84: Likewise
```suggestion
- The root domain of the `rdns` value SHOULD BE an active domain controlled by the Provider;
```
Can't be a must because we don't provide a mechanism for making this testable, but hints that a future extension spec might.

EIPS/eip-6963.md
everdimension: Here I used the wording "DNS part of the `rdns`" only to mean the original non-reversed string.
We can't say that the literal `rnds` value needs to be a valid domain, because "com.example" isn't. It's the reversed-to-the original string that needs to be a valid domain :) 
I'm not sure how to say this without making the sentence unnecessarily long 

I agree that there's no need to exempt the "product name" part from being a valid part of the domain rfc.

At the same time I feel we have to highlight that, given a value of `com.example.wallet-name`, the following is expected:
* `example.com` SHOULD be an _active_ domain
* `wallet-name.example.com` MUST BE a _valid domain name_,
* `wallet-name.example.com` is NOT required to be an _active_ domain.

Perhaps this is better done with an example?

EIPS/eip-6963.md
everdimension: Good point!

EIPS/eip-6963.md
everdimension: I don't think we've defined a "_root domain_" thought 🤔

kdenhartog (approved):

EIPS/eip-6963.md
bumblefudge:
```suggestion
- The DNS part of the `rdns` value SHOULD BE an active apex domain controlled by the Provider;
```
Googling around, I realized "root" isn't really a stable or defined term.  This recent RFC supersedes the older ones and defines `apex` as the "highest delegation point owned by someone", basically: 
https://www.rfc-editor.org/rfc/rfc8499

PR #7469, opened 2023-08-09
Update EIP-1: Rewrite EIP-1 with content from the charter discussions

SamWilsn:


EIPS/eip-1.md
```diff
@@ -11,484 +11,231 @@ created: 2015-10-27
...
+##### Transitive Mutability
 
-**Final** - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.
+Keeping a proposal immutable upon completion is an important property of the EIP process. Mentioning/linking to other proposals or external resources pokes a hole in that immutability.
 
-A PR moving an EIP from Last Call to Final SHOULD contain no changes other than the status update. Any content or editorial proposed change SHOULD be separate from this status-updating PR and committed prior to it.
+In practice, this means:
 
-**Stagnant** - Any EIP in `Draft` or `Review` or `Last Call` if inactive for a period of 6 months or greater is moved to `Stagnant`. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to `Draft` or it's earlier status. If not resurrected, a proposal may stay forever in this status.
+ * Proposals in immutable statuses may only mention/link to other immutable proposals;
```
bumblefudge, line 84:
```suggestion
 - Proposals in immutable statuses may only mention/link to other immutable proposals;
```

Pandapip1 (changes requested):
It was mentionned that there is no written record that only Final EIPs should be used in production. Can we fix this?

SamWilsn:
> It was mentionned that there is no written record that only Final EIPs should be used in production. Can we fix this?

Is this something we want to codify? I usually suggest that EIPs in Last Call should be implemented and in-use before moving to Final.

PR #7480, opened 2023-08-11
Add EIP: EOF - Data section access instructions

gumb0:


Aryan9592 (approved):

EIPS/eip-7480.md
```diff
@@ -0,0 +1,92 @@
...
+2. Pushes the size of the data section of the active container to the stack.
+3. Deducts 2 gas.
+
+### `DATACOPY`
+
+1. Pops three values from the stack: `mem_offset`, `offset`, `size`.
+2. Performs memory expansion to `mem_offset + size` and deducts memory expansion cost.
+3. Deducts `3 * ((size + 31) // 32)` gas for copying.
+3. If `offset + size` is greater than data section size of the active container, execution results in exceptional halt.
+4. Reads `[offset:offset+size]` segment from the data section and writes it to memory starting at offset `mem_offset`.
```
hugo-dc, line 67: Shouldn't this also require validating `offset+size <= data_section_size`?

gumb0: This is in point 3 above?

hugo-dc: You're right

EIPS/eip-7480.md
```diff
@@ -0,0 +1,92 @@
...
+
+Clear separation between code and data is one of the main features of EOF1. Data section may contain anything, e.g. compiler's metadata, but to make it useful for smart contracts, EVM has to have instructions that allow to read from data section. Previously existing instructions for bytecode inspection (`CODECOPY`, `CODESIZE` etc.) are deprecated in EOF1 and cannot be used for this purpose.
+
+`DATALOAD`, `DATASIZE`, `DATACOPY` instructions combination follows the design of existing instructions for reading other kinds of data (return data, call data).
+
+`DATALOADN` is an optimized version of `DATALOAD`, where data offset to read is set at compilation time, and therefore need not be validated at run-time, which makes the instruction cheaper.
+
+## Specification
+
+We introduce four new instructions on the same block number [EIP-3540](./eip-3540.md) is activated on:
```
g11tech, line 28: i think it would be better to specify activation independently (and this EIP already specifies that it depends on 3540). Their activation is dependent on inclusion of the EIP in a particular hardfork which would be an independent decision by the ACD.

```suggestion
We introduce four new instructions:
```

EIPS/eip-7480.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Abstract
+
+Four new instrutions are introduced, that allow to read EOF container's data section: `DATALOAD` loads 32-byte word to stack, `DATALOADN` loads 32-byte word to stack where the word is addressed by a static immediate argument, `DATASIZE` loads data section size and `DATACOPY` copies a segment of data section to memory. 
+
+## Motivation
+
+Clear separation between code and data is one of the main features of EOF1. Data section may contain anything, e.g. compiler's metadata, but to make it useful for smart contracts, EVM has to have instructions that allow to read from data section. Previously existing instructions for bytecode inspection (`CODECOPY`, `CODESIZE` etc.) are deprecated in EOF1 and cannot be used for this purpose.
+
+`DATALOAD`, `DATASIZE`, `DATACOPY` instructions combination follows the design of existing instructions for reading other kinds of data (return data, call data).
```
axic, line 22:
```suggestion
The `DATALOAD`, `DATASIZE`, `DATACOPY` instruction pattern follows the design of existing instructions for reading other kinds of data (i.e. returndata and calldata).
```

EIPS/eip-7480.md
axic: Most of EOF must be activated at the same time, otherwise the functionality is very limited.

lightclient (approved):
LGTM

ProphetZX:
Just a short question: Why do you plan to implement a DATALOAD and DATALOADN method/function? Is it as a backup version, so if either gets rejected in  a specific situation, you‘d want to switch to the other one?

ProphetZX:
Just have some doubts it‘s really necessary. :)

gumb0:
> Just a short question: Why do you plan to implement a DATALOAD and DATALOADN method/function? Is it as a backup version, so if either gets rejected in a specific situation, you‘d want to switch to the other one?

It's in Motivation:

> `DATALOADN` is an optimized version of `DATALOAD`, where data offset to read is set at compilation time, and therefore need not be validated at run-time, which makes the instruction cheaper.

`DATALOADN` can be used when the offset is known at compile-time, `DATALOAD` is needed to read from offset determined at run-time.

PR #7484, opened 2023-08-14
Add EIP: Registries and Adapters for Smart Accounts

kopy-kat:
This proposal standardises the interface and functionality of Module Registries, allowing modular smart contract accounts to verify the security of modules using a Registry Adapter. It also provides a reference implementation of a Singleton Module Registry.


EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+### Interfaces
+
+#### IRegistry.sol
+
+##### `check`
+
+Checks the attestation of a given module address and attestations made by a single attester.
+
+Returns two timestamps. `listedAt` is a `uint48` timestamp of when the attestation was made.
+The second return parameter `revokedAt` is a `uint48` timestamp of when an attestation was revoked.
```
rmeissner, line 61: Is there any expectation how this value behaves? I.e. can revokedAt be > 0 and then go back to 0 later?

EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+- Is RECOMMENDED to query the registry before module installation.
+- MAY query the registry before module execution.
+
+### Interfaces
+
+#### IRegistry.sol
+
+##### `check`
+
+Checks the attestation of a given module address and attestations made by a single attester.
```
rmeissner, line 58: What happens if the attester doesn't exist?

EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+_EXAMPLE_: If threshold of 2 is set and 5 attesters are queried, the function will return `true` if 2 or more of the selected attesters have issued an attestation that has not been revoked.
+
+**NOTE**:
+
+- `verified` of false MUST be treated as a security risk.
+
+```solidity
+function verify(
+        address module,
+        address[] memory attesters,
```
rmeissner, line 93: Are there any gas limits involved when iterating the calls for the attestations?

I.e. if the second `attester` uses up all gas, would this block any further execution? 

Should the caller make any assumptions here (or set any gas limit)?

EIPS/eip-7484.md
kopy-kat: If the attester doesn't exist then listedAt will be 0

EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+
+![Adapter flow](../assets/eip-7484/flow.png)
+
+An Adapter is a piece of functionality that allows smart accounts to query the module registry before using a module. In order to be compatible with this proposal, a smart account MUST implement the adapter either natively in the account or as a default module that is installed during account deployment.
+
+The adapter:
+
+- MUST query the registry at least once before calling a module for the first time.
+- MUST use one of the functions on the IRegistry interface when querying the registry.
+- Is RECOMMENDED to query the registry before module installation.
```
rmeissner, line 49: Since the status of an attestation can change (i.e. the `revokedAt`) should it also be recommended to call the registry before any/every interaction

EIPS/eip-7484.md
kopy-kat: Yes, the expectation (including code in the reference implementation) is that revokedAt can change from 0 to > 0 but not the other way around

rmeissner: is this a MUST or just a SHOULD? this feels like an important detail for such an interface

EIPS/eip-7484.md
rmeissner: should this be documented? As another solution would be that the call reverts, and it would probably make sense to avoid different behaviors. With ERC-20 there was a similar issue where it was not defined how `transfer` should behave on failure and then different implementations evolved.

EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Interfaces
+
+#### IRegistry.sol
+
+##### `check`
+
+Checks the attestation of a given module address and attestations made by a single attester.
+
+Returns two timestamps. `listedAt` is a `uint48` timestamp of when the attestation was made.
```
rmeissner, line 60: While for most it might be clear why `uint48` is used here (and below), I think it would make sense to mention the motivation for this, as the block timestamp is a `uint256`.

EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+- **Smart account** - An ERC-4337 compliant smart contract account that has a modular architecture.
+- **Module** - Self-contained smart account functionality.
+- **Attestation** - Onchain assertions made about the security of a module.
+- **Registry** - An onchain list of attestations about modules.
+- **Adapter** - Smart account functionality that handles the fetching and validation of attestations from the registry.
+- **Attester** - An entity that makes an attestation about a module.
+
+### Overview
+
+![Adapter flow](../assets/eip-7484/flow.png)
```
rmeissner, line 41: Would it make sense to use a sequence diagram here?

```mermaid
sequenceDiagram
    autonumber
    participant M as Module
    participant SA as Smart Account<br>w/ Adapter
    participant MR as Module Registry
    actor AT as Attester
    AT ->> MR: attests to module
    SA ->> MR: check module
    SA -->> M: installs
    opt Execution Check
       SA ->> MR: check module
    end
    SA -->> M: executes
    
```

EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+
+The reference implementation of the registry is heavily inspired by the Ethereum Attestation Service. The specific use-case of this proposal, however, required some custom modifications and additions to EAS, meaning that using the existing EAS contracts as the module registry was sub-optimal.
+
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Reference Implementation
+
+### IAttestation.sol
```
rmeissner, line 143: The data of the `IAttestation.sol` doesn't seem to be used. Would it make sense to remove it to make it easier to understand the reference code?

EIPS/eip-7484.md
kopy-kat: You're right this is quite an important detail for alternate registries, we'll explicitly mention this as a MUST in the next draft

EIPS/eip-7484.md
```diff
@@ -0,0 +1,239 @@
...
+
+**NOTE**:
+
+- `listedAt` of 0 MUST be treated as a security risk.
+- `revokedAt` of non-0 MUST be treated as a security risk.
+
+```solidity
+function check(
+        address module,
+        address attester
```
filmakarov, line 71: Will be there a public list to take attesters to check from?

Also, what if Smart Account does not care about which attester provided attestation for a module but only cares certain cumulative threshold of 'authority points' is reached?
Is it meant to be handled by an Adapter internally? Like maintaining such a list of attesters with their authority level and allowing enabling (executing) modules only after some threshold?

sgilflores3 (approved):

bitcoinfinancier (approved):

EIPS/eip-7484.md
kopy-kat: I'm not entirely sure what you mean. In the reference implementation, neither `check` nor `verify` read the attestation data so I'm unsure how the call could run out of gas (unless the gas limit was not properly set on the frontend). 

If you refer to cases in which data is being read and excessively long attestations could cause the wallet to run out of gas then I'm not sure whether that's a feasible attack vector since creating these attestations would be quite expensive.

EIPS/eip-7484.md
```diff
@@ -0,0 +1,217 @@
...
+
+    function checkModule(address module, address trustedAttester) internal {
+       // check module implementation address on registry
+        (uint48 listedAt, uint48 flaggedAt) = registry.check(module, trustedAttester);
+
+        // revert if module was ever flagged or was never attested to
+        require(listedAt != 0 && flaggedAt == 0, "Module is insecure");
+    }
+
+    function verifyModule(address module, address[] memory attesters, uint256 threshold) internal {
```
ankurdubey521, line 173: Based on the interface defined here, it looks like checks are made based on whether an Attestation exists or not, i.e. an attestation can be considered to be a boolean. 
What are your thoughts on supporting checks based on the data inside an Attestation? Say, for example, an Attestation stores the number of audits done for a particular module and a smart contract wallet wishes to set a security policy where only modules with >= K audits are installed.

EIPS/eip-7484.md
```diff
@@ -0,0 +1,217 @@
...
+requires: 4337
+---
+
+## Abstract
+
+This proposal standardises a Registry Adapter for modular smart contract accounts. This Adapter allows the account to query and verify security attestations about a module through an Attestation Registry. The adapter is responsible for querying the registry and correctly handling the return values.
+
+## Motivation
+
+[ERC-4337](./eip-4337.md) standardises the execution flow of contract accounts and [ERC-6900](./eip-6900.md) aims to standardise the modular implementation of these accounts, allowing any developer to build modules for these modular accounts (hereafter smart accounts). However, adding third-party modules into smart accounts unchecked opens up a wide range of attack vectors on these accounts.
```
SamWilsn, line 20: Just so you're aware, mentioning ERC-6900 will create an implicit link between this EIP and that one. ERC-7484 won't be able to progress past the status of ERC-6900 (currently Draft.) That's totally fine if that's what you want, I just wanted to point it out.

EIPS/eip-7484.md
```diff
@@ -0,0 +1,217 @@
...
+        ...
+    }
+
+    ...
+}
+```
+
+### Registry
+
+See `https://github.com/rhinestonewtf/registry/`
```
SamWilsn, line 209: Please remove any external links.

EIPS/eip-7484.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Security Considerations
+
+Needs discussion.
```
SamWilsn, line 213: If you use an HTML style comment, the linter will make sure you replace it before you go to review:

```suggestion
Needs discussion. <!-- TODO -->
```

EIPS/eip-7484.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Motivation
+
+[ERC-4337](./eip-4337.md) standardises the execution flow of contract accounts and [ERC-6900](./eip-6900.md) aims to standardise the modular implementation of these accounts, allowing any developer to build modules for these modular accounts (hereafter smart accounts). However, adding third-party modules into smart accounts unchecked opens up a wide range of attack vectors on these accounts.
+
+A proposed solution to these security considerations is a permission framework that allows module developers to define the permissions that their modules require. The counterpiece of these permissions are attestations that assert statements about the security of modules and can be queried onchain.
+
+This proposal is independent of the exact implementation of the permissioning system and could even be used across different kinds of permissioning systems. Instead, the goal of this proposal is to outline a standard, but flexible, way to create onchain attestations and standardise how these attestations are queried by the smart account during module installation or execution.
+
```
SamWilsn, line 25: I'd like to see, in your Motivation, some justification for why this needs an EIP. Generally EIPs exist to coordinate multiple vendors that have to integrate, but this looks like it'll be a single registry contract that everyone integrates with.

kopy-kat: Implemented in newest version

EIPS/eip-7484.md
kopy-kat: Done

EIPS/eip-7484.md
```diff
@@ -0,0 +1,272 @@
...
+- Allow attesters to specify an expiry date for their attestations and revert during `check` or `checkN` if an attestation is expired.
+- Implement a view function that allows an adapter or offchain client to read the data for a specific attestation.
+
+#### `check`
+
+Takes two arguments: `module` and `attester`.
+
+- The Registry MUST revert if the `attester` has not made an attestation on the `module`.
+- The Registry MUST revert if the `attester` has revoked their attestation on the `module`.
```
tomarsachin2271, line 76: There should be one more point here IMO

The Registry MUST revert if the `attestation` provided is expired.

Same holds for checkN method.

kopy-kat: Currently, we think it makes more sense to have this be an optional (but encouraged) requirement since some registries might choose to not implement expiration

lightclient (approved):

PR #7487, opened 2023-08-15
Update EIP-6122: don't skip genesis forks

lightclient:
This addresses Felix's feedback on eth magician: https://ethereum-magicians.org/t/eip-6122-forkid-checks-based-on-timestamps/12130/2


lightclient:
I think this PR would be a nice change to the forkid mechanism in general, but it isn't really right to change 6122 and not change 2364. So I will go ahead and close this.

PR #7493, opened 2023-08-17
Update EIP-6909: update eip-6909

banteg:
- fix incorrect types in `Transfer` event
- add a `bool` return value for `transfer`, `transferFrom`, `approve`, `setOperator` because it makes calls from contracts cheaper


jtriley-eth (approved):
lgtm, ty!

jtriley-eth (changes requested):
*correction: can you also add the boolean return to the reference implementation in the eip?

PR #7495, opened 2023-08-17
Add EIP: SSZ PartialContainer

etan-status:


jonesk7734 (approved):
😀

EIPS/eip-7495.md
```diff
@@ -0,0 +1,165 @@
...
+For the purpose of serialization, `PartialContainer[T, N]` is always considered ["variable-size"](https://github.com/ethereum/consensus-specs/blob/67c2f9ee9eb562f7cc02b2ff90d92c56137944e1/ssz/simple-serialize.md#variable-size-and-fixed-size) regardless of `T`.
+
+### Stability guarantees
+
+The serialization and merkleization of a `PartialContainer[T, N]` remains stable as long as:
+
+- The maximum capacity `N` does not change
+- The order of fields within `T` does not change
+- New fields are always added to the end of `T`
+- Required fields of `T` remain required `E`, or become an `Optional[E]`
```
g11tech, line 48: can this happen?: required => optional => required (and so on...) as this and below points imply infinite change

EIPS/eip-7495.md
```diff
@@ -0,0 +1,165 @@
...
+
+Merkleization `hash_tree_root(value)` of an object `value` is then extended with:
+
+- `mix_in_object(merkleize(([hash_tree_root(element) if is_active_field(element) else Bytes32() for element in value.data] + [Bytes32()] * N)[:N]), value.active_fields)` if `value` is a `PartialContainer[T, N]`.
+
+## Rationale
+
+### What are the problems solved by `PartialContainer[T, N]`?
+
+Current SSZ types are only stable within one version of a specification, i.e., one fork of Ethereum. This is alright for messages pertaining to a specific fork, such as attestations or beacon blocks. However, it is a limitation for messages that are expected to remain valid across forks, such as transactions or receipts. In order to support evolving the features of such perpetually valid message types, a new SSZ scheme needs to be defined.
```
g11tech, line 122: sounds amazing property :clap:

g11tech (approved):
lgtm

EIPS/eip-7495.md
etan-status: Yes, a new fork could relax validity constraints, while a subsequent fork could require the field to be present again. Whether that makes sense is up to the application, but the serialization format proposed here will be stable across both directions (from optional to required and vice versa), essentially providing maximum flexibility when it comes to future design space.

In fact, an implementation could simply treat all fields as optional always, if they prefer to enforce constraints in the application layer instead of the serialization library. Another implementation may go the other direction and also push down an invariant callback to ensure that only valid combinations are accepted; e.g., a receipt can contain _either_ an intermediate state root, _or_ a status code, but not both at the same time. Yet another implementation may try to use union types to represent various valid combinations; point being is, none of these implementation details leak into the serialization representation.

etan-status: --> https://ethereum-magicians.org/t/eip-7495-ssz-partialcontainer/15476 is a better location for discussing technical points.

PR #7516, opened 2023-08-23
Update EIP-6900: Spec update 3

jaypaik:
- Updated shared components diagram to more accurately depict modular functions and their relationship to plugins.
- Added "native function" definition under Terms.
- Added the `IPlugin` interface that describes the functions a plugin must implement.
    - Plugins must implement ERC-165 and support this interface.
    - Associated functions in the interface take a `uint8` function identifier to support multiple implementations of the same type of associated function.
- Removed the `FunctionReference` custom type, as function identifiers do not represent function selectors.
- Removed the hook group concept and its associated types.
- Replaced the `IPluginUpdate` interface with the `IPluginManager` interface.
    - The function `updatePlugins` has been replaced by the functions `installPlugin` and `uninstallPlugin`.
- Updated the `IPluginLoupe` interface to reflect the above changes.
- Added the `IPluginExecutor` interface, which introduces new functions `executeFromPlugin` and `executeFromPluginExternal` that allow plugins to interact with execution functions on the modular account as well as external contracts.
- Standard execution functions and `executeFromPluginExternal` should not allow calls to contracts that implement the `IPlugin` interface to maintain the integrity of account and plugin configuration.
- Added a section describing the plugin manifest, and the expected behavior for `installPlugin` and `uninstallPlugin`.
- Added a section describing the expected behavior for calls made from plugins.
- Style updates and typo fixes.


EIPS/eip-6900.md
```diff
@@ -88,190 +79,334 @@ Each step is modular, supporting different implementations for each execution fu
...
+    /// @notice Run the post execution hook specified by the `functionId`.
+    /// @dev To indicate the entire call should revert, the function MUST revert.
+    /// @param functionId An identifier that routes the call to different internal implementations, should there be more than one.
+    /// @param preExecHookData The context returned by its associated pre execution hook.
+    function postExecutionHook(uint8 functionId, bytes calldata preExecHookData) external;
+
+    /// @notice Describe the contents and intended configuration of the plugin.
+    /// @dev This function MUST NOT make any state changes.
+    /// @return A manifest describing the contents and intended configuration of the plugin.
+    function pluginManifest() external pure returns (PluginManifest memory);
```
rmeissner, line 234: `pure` indicates that it does not read any state at all, would that mean that the manifest data is kept in the immutable storage (I think this is not directly possible for structs right now)? Also this does not align with  `This function MUST NOT make any state changes.`, which is the description for a `view` function.

jaypaik: Thanks for the thorough review!

Manifest data should be declared within the `pluginManifest()` function itself. With plugin dependencies provided at installation time, the plugin does not need to read from storage.

Agree that `/// @dev This function MUST NOT make any state changes.` is redundant / not apt here given `pure`.

The manifest hash check during installation is meant to prevent installation of malicious plugins that may swap its manifest to a different one in between user confirmation and actual execution.

rmeissner: Thanks for the explanation, this makes sense :)

Since there is a manifest hash check it is not super critical that this is actually a pure function, right? Since a change to the manifest would be detected.

jaypaik: That's correct. I do like `pure` in the sense that it conveys to the developer that what should be returned should be fixed and not change over time (and possibly read from mutable state). However, I do see the possibility that `pure` might be difficult to work with in certain scenarios.

Do you see practical benefits to this being relaxed to `view`?

rmeissner: > Do you see practical benefits to this being relaxed to view?

Good question: My comment was primarily because I was confused by the slight mismatch of natspec and modifier. 

Personally I would probably relax it as it is not a real security benefit (due to the hash check). This will give the developer the flexibility to store the data however they want. That being said they can already do this by "just not following the interface 100%", as there is no opcode to enforce `pure` (like staticall for `view`). Also I cannot provide a concrete use case were storing it in memory would be required, therefore not strong preference ;)

jaypaik: Got it thanks for your thoughts! Will continue to iterate on this.

PR #7517, opened 2023-08-23
Update EIP-7329: Move to Last Call

shemnon:
Last call for EIP-7329 re-review.


SamWilsn (changes requested):
I can't comment on individual lines. The only change that's necessary is removing "I've written a script...". That's mostly an implementation detail, and not really relevant to future readers

gcolvin (approved):

SamWilsn (approved):

PR #7526, opened 2023-08-28
Revert "Add EIP: SSZ PartialContainer"

ProphetZX:
Reverts ethereum/EIPs#7495

Creates conflicts with implemented state.


etan-status:
What conflict?

Pandapip1:
EIPs should not be deleted, they should be withdrawn. Also, this EIP does not seem to have any conflicts with anything -- if it did, a PR modifying the spec to make it compatble should be made instead.

@ProphetZX, if you're incompetent... well I've told you to read EIP-1 already. Please re-read it. If you're being malicious, then please, for the love of god, stop. Taking advantage of the fact that we will never ban anyone so you can be a troll is shameful. Either way, your behavior is disruptive, and I beg you--do something to make it less disruptive.

PR #7530, opened 2023-08-28
Update EIP-4844: 4844 - Add scenario for blob txs submitted via RPC

acolytec3:
Following discussion on the Eth R&D discord, it feels clear that the current 4844 specs are not 100% clear with regard to handling of blob transactions submitted to a node via `eth_sendRawTransaction` and this proposes an update to the spec that clarifies this (based on the actual implementations in the EthereumJS and Geth clients).


lightclient:
Hi @acolytec3 sorry the EIP isn't the correct place to specify this logic. We should add a note in the method description in the execution-apis. Thanks!

PR #7637, opened 2023-09-04
Moves 5114 to Last Call

MicahZoltu:
Set the last call deadline quite a bit in the future because I'm not in a big hurry and I want to leave lots of time for any last minute feedback, but I can move this forward if desired by the editors.


Allyszz:
> 
> 
> ### File `EIPS/eip-5114.md`
> 
> Requires 1 more reviewers from @axic, @pandapip1, @samwilsn, @xinbenlv
> 

Reply

SamWilsn:
Your `last-call-deadline` was in the past, so I bumped it to two weeks from now.

SamWilsn (approved):

MicahZoltu:
Bleh, meant to be 2024-01-01.  Any problem with me updating it to be later?  I want plenty of time for Last Call unless there is a desire to keep it short.

PR #7665, opened 2023-09-06
Website: Remove Networking and Interface categories

Pandapip1:
I don't know why it's picking this up as a website change vs an EIP-1 change. This is an EIP-1 change.

This PR removes the Networking category and changes all Networking EIPs to Core, and recategorizes all Interface EIPs. It also redefines Core and ERC to encompass those new proposals.

This is meant as a very temporary measure to speed up the split. Once we split, it should be way easier to submit more specific CFIs to the WG that don't require all editors to contribute.


lightclient:
Networking changes don't require hard fork and would therefore not fall under the core category.

Pandapip1:
> Networking changes don't require hard fork and would therefore not fall under the core category.

I am proposing to redefine what Core means.

Pandapip1:
@eth-bot rerun

karalabe:
Jesus

karalabe:
Fucking

karalabe:
Christ

karalabe:
You've

karalabe:
sent

karalabe:
52

karalabe:
notifications

karalabe:
to

karalabe:
every

karalabe:
member

karalabe:
this

karalabe:
repo

karalabe:
Can

karalabe:
you

karalabe:
please

karalabe:
stop

karalabe:
pushing

karalabe:
commits

karalabe:
1

karalabe:
by

karalabe:
1

karalabe:
?

Pandapip1:
Guess what? You've also just sent a bunch of notifications! I also stopped a few minutes before you sent those comments, so your chain wasn't even needed.

pcaversaccio:
sers, I just got 80 emails...

Pandapip1:
> sers, I just got 80 emails...

It should have been less than 60. Weird.

Still, forgot that GitHub sends people emails for notifications and that they don't rate limit them. Sorry!

ethDreamer (approved):

Pandapip1:
> Strongly oppose rolling changes that do not require a hardfork into core. Protocol changes are the most important modifications to Ethereum and it should be easy for the community to follow what is being proposed and accepted. By rolling network changes and other related changes into core, people will need to wade through much more junk to keep their eye on the protocol.

@SamWilsn claimed that networking changes are capable of creating forks. Would @SamWilsn mind explaining? (I think I know, but I also don't want to get this wrong and look like an idiot)

lightclient:
I would rather deprecate them and leave the existing EIPs as they are.

kdenhartog:
> Strongly oppose rolling changes that do not require a hardfork into core.

In general I agree with this and the reasoning behind it.

> I would rather deprecate them and leave the existing EIPs as they are.

+1 with the addition that going forward I think it makes sense to have a core (optional) and core (HF) given that most changes to the protocol today do assume a HF, but it's not a requirement. This would make it that much easier for core devs to identify critical changes while also also creating an avenue in the future for optional extensions.

It seems reasonable to combine these somewhat while still keeping clear signals since the key people who would need to discuss and ultimately implement would still be core devs in a lot of cases.

Pandapip1:
> I would rather deprecate them and leave the existing EIPs as they are.

This is not an option. All EIPs have to have a working group, and you have already stated you don't want the core devs to be in control of these (as a reminder: we're splitting ERCs, so anything that's not an ERC will effectively become the responsibility of the Core WG, which you have already stated you don't want!). Again, would you prefer that we move them to the ERC WG or leave them in the Core WG?

> I think it makes sense to have a core (optional) and core (HF) given that most changes to the protocol today do assume a HF, but it's not a requirement.

+1 to this. This was my motivation for expanding the scope of Core, with the assumption that it would later be recategorized. However, since the track for these sort of changes more closely resembles ERCs, I guess it could make sense to categorize them as such.

Pandapip1:
It's not too heated. That was just me being dumb and sending a bunch of people 47 emails and @karalabe adding 20-ish more. It's all resolved now.

Pandapip1:
> Anything is an option, we literally make the rules.

This is not false. The universe makes the rules. We can't make a rule that requires the halting problem to be solved (obviously this is hyperbole, but you get the point I'm trying to make here).

Those EIPs need to go *somewhere*, and you have expressed that it should not wherever the Core EIPs are going. Therefore, are you okay with me moving the relevant EIPs to ERC to resolve this?

lightclient:
You haven't explained why they can't stay in the current repo for historical reasons beyond "those are the rules". They aren't ERCs and therefore shouldn't be recategorized.

Pandapip1:
> You haven't explained why they can't stay in the current repo for historical reasons beyond "those are the rules".

I mean, it could stay around in this repo, but I thought you explicitly stated you didn't want them being comingled with Core EIPs.

> They aren't ERCs and therefore shouldn't be recategorized.

Ah-I think you misunderstand me. I'm proposing that ERCs be redefined so that they *can* be recategorized.

EIPS/eip-2159.md
```diff
@@ -5,7 +5,7 @@ author: Adrian Sutton (@ajsutton)
 discussions-to: https://ethereum-magicians.org/t/common-chain-metrics/3415
 status: Final
 type: Standards Track
-category: Interface
+category: Cire
```
kdenhartog, line 8: Core
```suggestion
category: Core
```

EIPS/eip-5593.md
```diff
@@ -6,7 +6,7 @@ author: Yan Zhu (@diracdeltas), Brian R. Bondy (@bbondy), Andrea  Brancaleoni (@
 discussions-to: https://ethereum-magicians.org/t/rfc-limiting-provider-object-injection-to-secure-contexts/10670
 status: Draft
 type: Standards Track
-category: Interface
+category: ERC
```
kdenhartog, line 9: ERC doesn't seem right here. The group working on ERCs is distinctly different from the group working on wallets and wallet providers and I think we want to make that distinction clear still.

EIPS/eip-6963.md
```diff
@@ -6,7 +6,7 @@ author: Pedro Gomes (@pedrouid), Kosala Hemachandra (@kvhnuke), Richard Moore (@
 discussions-to: https://ethereum-magicians.org/t/eip-6963-multi-injected-provider-interface-aka-mipi/14076
 status: Review
 type: Standards Track
-category: Interface
+category: ERC
```
kdenhartog, line 9: same here regarding ERC vs Wallets

Pandapip1:
@kdenhartog the point of moving them all to ERC is because the processes for the two are the same, wheras the process for Core EIPs is different. I'm still in favor of then splitting ERCs into different categories, but first they need to get out of what will effectively be limbo.

kdenhartog:
> @kdenhartog the point of moving them all to ERC is because the processes for the two are the same, wheras the process for Core EIPs is different. I'm still in favor of then splitting ERCs into different categories, but first they need to get out of what will effectively be limbo.

Cool, I'm fine with a double move. We are prepping 6963 to move towards final within wallet so we probably will have to address the difference in categories sooner rather than later.

MicahZoltu:
> I would rather update them to say that they are deprecated and no further networking/interface proposals will be accepted.

What is the proposed alternative mechanism for specifying and coordination around networking changes?

lightclient:
> What is the proposed alternative mechanism for specifying and coordination around networking changes?

There is no proposed alternative, we are _actively_ using the ethereum/devp2p repo to specify and coordinate networking change.

fjl:
Maybe a bit more about this. We have so far always created EIPs for networking changes. However, it seems nobody wants to deal with these EIPs because they do not easily fit the workflow, can't be scheduled in forks etc.

For recent p2p protocol changes, we have just included the change to networking as a part of the Core EIP. That works too, in the end we just need the EIP as a reference in the spec to justify the change.

When it comes to more fundamental p2p protocol changes that do not affect the core data structures, there is actually no real need to agree on them using EIPs. It is an implementation concern and the implementers can find ways to agree. The Ethereum blockchain does not require p2p networking to function. It does in practice, but the definition of the chain protocol does not depend on it.

Pandapip1:
> there is actually no real need to agree on them using EIPs. It is an implementation concern and the implementers can find ways to agree

EIPs are all about interop, so I'd say EIPs would still be helpful here.

Pandapip1:
> I think they should stay in this repo and I don't think we should broaden the scope of core category. So, again, I oppose this change to the categories. I would rather update them to say that they are deprecated and no further networking/interface proposals will be accepted

@lightclient, would you be okay with broadening the scope of ERCs?

lightclient:
Not to include networking EIPs

Pandapip1:
> Not to include networking EIPs

Okay, what about Interface EIPs?

lightclient:
I mean I also don't think ERC should include interface EIPs. But slightly less strongly than networking EIPs.

Pandapip1:
This will be done during the split.

PR #7668, opened 2023-09-07
Add EIP: Add Max Epoch Churn Limit

dapplion:

 - The build passes.

### DIFF
--- a/EIPS/eip-7668.md
+++ b/EIPS/eip-7668.md
@@ -1,0 +1,71 @@
+---
+eip: 7668
+title: Add Max Epoch Churn Limit
+description: Modify the churn limit function to upper bound it to a max value
+author: dapplion (@dapplion)
+discussions-to: https://ethereum-magicians.org/t/eip-7668-add-max-epoch-churn-limit/15709
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-09-07
+---
+
+## Abstract
+
+Limit Beacon chain growth from exponential to linear to mitigate the risks of very high levels of total ETH supply staked.
+
+## Motivation
+
+With a full deposit queue, the share of ETH supply staked will reach 50% by May 2024, 75% by September 2024, and 100% by December 2024. While yields decrease as the validator set size increases, the APY at 100% ETH supply staked is 1.6% for consensus rewards only excluding MEV. This yield can be sufficient to attract capital such that the equilibrium point of the validator set size is close to its maximum possible.
+
+Reaching very high levels of total ETH supply staked can have negative externalities. Liquid staking derivatives (LSDs) are likely to absorb most of the staking demand, jeopardizing the economic guarantees of the Beacon chain. If LSD tokens overtake the native token in liquidity, they permeate systemic risks into most Ethereum applications.
+
+It's unclear if the Ethereum network security benefits from having so much stake. After some upper bound, attacks become so economically unfeasible that additional stake does not increase the network's safety. In that case, the network will be overpaying for security. Note that Beacon chain rewards are dictated by an opinionated function, chosen before its launch in 2020. Much has changed since then, including the Beacon chain achieving unprecedented success. Beacon chain rewards should be adjusted to match today's reality, and stop incentiving actors into increasing the risks listed above at no benefit to the network's security.
+
+If one agrees that Beacon chain rewards should be adjusted down, it's important to forecast the social cost of doing so. The more ETH is staked, the more actors may depend on today's reward regime and oppose the change. Also, the network could reach some catastrophic tipping point before it can agree to a new rewards function.
+
+This proposal accepts the complexities of changing the rewards curve and is meant only to slow down growth. It wants to mitigate the negative externalities of very high levels of total ETH supply staked before a proper solution is implemented.
+
+## Specification
+
+### Constants
+
+| Name | Value |
+| ---- | ----- |
+| `MAX_PER_EPOCH_CHURN_LIMIT` | 12 |
+
+### Execution layer
+
+This requires no changes to the Execution Layer.
+
+### Consensus layer
+
+- Modify `get_validator_churn_limit` function adding an upper bound `MAX_PER_EPOCH_CHURN_LIMIT`
+
+## Rationale
+
+### `MAX_PER_EPOCH_CHURN_LIMIT` value
+
+Depending on the specific constant selection the churn can _decrease_ at the activation fork epoch. The Beacon chain spec can handle this without issues. During 2023 Q4 (projected Dencun activation) the churn value will range 14-16. The table below compares the projected validator set assuming a continuous full deposit queue.
+
+| `MAX_PER_EPOCH_CHURN_LIMIT` | max validator set (% of ETH supply) at 2024-07 |
+| --------------------------- | ---------------------------- |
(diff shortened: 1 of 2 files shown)


EIPS/eip-7668.md
```diff
@@ -0,0 +1,71 @@
...
+created: 2023-09-07
+---
+
+## Abstract
+
+Limit Beacon chain growth from exponential to linear to mitigate the risks of very high levels of total ETH supply staked.
+
+## Motivation
+
+With a full deposit queue, the share of ETH supply staked will reach 50% by May 2024, 75% by September 2024, and 100% by December 2024. While yields decrease as the validator set size increases, the APY at 100% ETH supply staked is 1.6% for consensus rewards only excluding MEV. This yield can be sufficient to attract capital such that the equilibrium point of the validator set size is close to its maximum possible.
```
djrtwo: I would use the following language to make it more explicit the assumption:
> In the event that the deposit queue stays 100% full

djrtwo: "1.6% due to issuance, excluding transaction fees/MEV"

djrtwo: > This yield can be sufficient to attract capital such that the equilibrium point of the validator set size is close to its maximum possible.

I don't think this claim is substantiated. 1.6% yield is not necessarily sufficient to attract capital. At just 1.6%, capital can easily find way less risky yields -- e.g. bonds and t-notes.

"This small yield does not necessarily dissuade additional capital staking due to the often much higher and unpredictable yields from MEV"

EIPS/eip-7668.md
```diff
@@ -0,0 +1,71 @@
...
+
+Reaching very high levels of total ETH supply staked can have negative externalities. Liquid staking derivatives (LSDs) are likely to absorb most of the staking demand, jeopardizing the economic guarantees of the Beacon chain. If LSD tokens overtake the native token in liquidity, they permeate systemic risks into most Ethereum applications.
```
djrtwo: In this EIP, I might stay closer to the system impact of the linear load (state size and attestations) and the security and economic impacts more broadly of 50%+ staked. I don't think there are 100% clear answers but there is a lot of worry and debate at these thresholds, so instead of making explicit claims about a particular thing -- e.g., LSDs -- I think it simpler to make claims more broadly about the unintended consequences of very high stake rates.

Also, on this particular item, it's not clear that just reducing total stake rate would reduce the % that goes into LSDs and I don't think you should attempt to make that argument here

```suggestion
Reaching very high levels of total ETH supply staked can have negative externalities. Liquid staking derivatives (LSDs) might continue to grow at outsized pace, jeopardizing the economic guarantees of the Beacon chain. If LSD tokens overtake the native token in liquidity, they permeate systemic risks into most Ethereum applications.
```

EIPS/eip-7668.md
```diff
@@ -0,0 +1,71 @@
...
+
+It's unclear if the Ethereum network security benefits from having so much stake. After some upper bound, attacks become so economically unfeasible that additional stake does not increase the network's safety. In that case, the network will be overpaying for security. Note that Beacon chain rewards are dictated by an opinionated function, chosen before its launch in 2020. Much has changed since then, including the Beacon chain achieving unprecedented success. Beacon chain rewards should be adjusted to match today's reality, and stop incentiving actors into increasing the risks listed above at no benefit to the network's security.
```
djrtwo: You could make a note that PoS research, rewards curves, and target stake rate were performed in a pre-MEV world with an intention of targeting 10 to 30% (MAX! ha) stake rate. The impact of MEV, LSDs, and other things have turned that on it's head.

EIPS/eip-7668.md
```diff
@@ -0,0 +1,71 @@
...
+
+If one agrees that Beacon chain rewards should be adjusted down, it's important to forecast the social cost of doing so. The more ETH is staked, the more actors may depend on today's reward regime and oppose the change. Also, the network could reach some catastrophic tipping point before it can agree to a new rewards function.
```
djrtwo: Again -- there are economic tipping points, but also practical network/engineering tipping points that we want to avoid. I would give at least as much weight to the latter in these discussions

EIPS/eip-7668.md
```diff
@@ -0,0 +1,71 @@
...
+
+This proposal accepts the complexities of changing the rewards curve and is meant only to slow down growth. It wants to mitigate the negative externalities of very high levels of total ETH supply staked before a proper solution is implemented.
```
djrtwo:
```suggestion
This proposal does not change the rewards curve but instead only slows down validator set size growth. This is in an attempt to stall the negative externalities of very high levels of total ETH supply staked while a proper solution is researched and specified.
```

EIPS/eip-7668.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Specification
+
+### Constants
+
+| Name | Value |
+| ---- | ----- |
+| `MAX_PER_EPOCH_CHURN_LIMIT` | 12 |
```
djrtwo, line 35: I know this is the current value, but have you considered reducing it to a more modest number?
Why 12 and not, say, 8 or 6?

djrtwo: Is there consideration in only applying this to activations and not exits?

djrtwo (commented):
nice! thanks for cranking this out for review

EIPS/eip-7668.md
```diff
@@ -0,0 +1,71 @@
+---
+eip: 7668
```
abcoathup, line 2:
```suggestion
eip: 7514
```

EIPS/eip-7668.md
mkalinin: > Is there consideration in only applying this to activations and not exits?

At the first glance such imbalance doesn't affect security properties, but it might cause the need of additional analysis before the proposal can be accepted. Considering short time frames we might not want to impose any additional complexity to make this proposal pass through the governance process. Thus, capping both queues is preferable, IMHO

0xScratch (approved):

EIPS/eip-7668.md
dapplion: Removed LSD claims from EIP text

EIPS/eip-7668.md
dapplion: Updated on current version

EIPS/eip-7668.md
dapplion: Outdated line on current version

EIPS/eip-7668.md
dapplion: Included a note for this

EIPS/eip-7668.md
dapplion: The lower the number the higher delta there will be in queue capacity at the fork epoch. A lower number increases the "unfairness" argument, but it increases the effectiveness of this EIP. I pick 12 without strong reasoning, just because it yields good results without resulting in a dramatic drop in churn.

mkalinin: One of the potential arguments in favour of `12` is that this is likely the limit when the Deneb fork happens on the mainnet assuming the same market conditions, so the decrease of the churn limit is unlikely to happen

g11tech (approved):
lgtm

PR #7670, opened 2023-09-08
Update EIP-7053: Move to Last Call

bafu:
After the conversations with the community and the industrial experts, we propose to change this EIP from Review to Last Call.


eip-automerger:

 - Trying to change EIP 7053 state from Review to Last Call

PR #7672, opened 2023-09-08
Update EIP-4788: post audit tweaks

lightclient:
*repost from https://github.com/lightclient/4788asm/pull/16*

We've had a few really nice audits done on this code and things have generally been positive. There are a few pieces of feedback that I think are worth addressing before we settle on the final code though. Here is a list, ordered by priority:

1. Verify timestamp is non-zero -- currently the implementation allows an input of zero to return a zero root. This is because the current mainnet timestamp is not perfectly divisible by our modulus. Although no proof could be authenticated against the zero root, it's still an ugly correctness edge case that I think should be addressed.
2. ~~Update `buflen` to `93600` -- this value will share 500-1000[^1] more remainders if the slot time were to change versus the current modulus, decreasing the overall storage use of the contract (h/t @adietrichs).~~ updated to `8191` as the number is prime and there is 100% utilization for all potential slot time values.
3. Load calldata once -- minor nit, but in case the cost of retrieving calldata changes in the future, it would make more sense to only access it via `calldatacopy` a single time, instead of paying for that cost at each load. This require a bit of dup'ing and swap'ing, but still extremely manageable. As a part of this I was also able to delete the `get_input` macro (h/t @adietrichs).

[^1]: https://gist.github.com/lightclient/820a0d5c5861ed09c5cae5e384e9779e

### DIFF
--- a/EIPS/eip-4788.md
+++ b/EIPS/eip-4788.md
@@ -29,9 +29,9 @@ restaking constructions, smart contract bridges, MEV mitigations and more.
 | constants                    | value                                        |
 |---                           |---                                           |
 | `FORK_TIMESTAMP`             | TBD                                          |
-| `HISTORY_BUFFER_LENGTH`      | `8191`                                      |
+| `HISTORY_BUFFER_LENGTH`      | `8191`                                       |
 | `SYSTEM_ADDRESS`             | `0xfffffffffffffffffffffffffffffffffffffffe` |
-| `BEACON_ROOTS_ADDRESS`       | `0xbEAC020008aFF7331c0A389CB2AAb67597567d7a` |
+| `BEACON_ROOTS_ADDRESS`       | `0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02` |
 
 ### Background
 
@@ -224,14 +224,14 @@ by working backwards from the desired deployment transaction:
   "input": "0x60618060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604d57602036146024575f5ffd5b5f35801560495762001fff810690815414603c575f5ffd5b62001fff01545f5260205ff35b5f5ffd5b62001fff42064281555f359062001fff015500",
   "v": "0x1b",
   "r": "0x539",
-  "s": "0x1337005f06a8b6a0a0b1f4",
-  "hash": "0x7186d3f6803018a58578464132f120d9420c0a00e37ed42c117fdb3d650e1dac"
+  "s": "0x1b9b6eb1f0",
+  "hash": "0xdf52c2d3bbe38820fff7b5eaab3db1b91f8e1412b56497d88388fb5d4ea1fde0"
 }
 ```
 
 Note, the input in the transaction has a simple constructor prefixing the desired runtime code.
 
-The sender of the transaction can be calculated as `0x4f6Da9ba1Cc8c37d9CE52311d4baFfc43bA42D0e`. The address of the first contract deployed from the account is `rlp([sender, 0])` which equals `0xbEAC020008aFF7331c0A389CB2AAb67597567d7a`. This is how `BEACON_ROOTS_ADDRESS` is determined. Although this style of contract creation is not tied to any specific initcode like create2 is, the synthetic address is cryptographically bound to the input data of the transaction (e.g. the initcode).
+The sender of the transaction can be calculated as `0x0B799C86a49DEeb90402691F1041aa3AF2d3C875`. The address of the first contract deployed from the account is `rlp([sender, 0])` which equals `0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02`. This is how `BEACON_ROOTS_ADDRESS` is determined. Although this style of contract creation is not tied to any specific initcode like create2 is, the synthetic address is cryptographically bound to the input data of the transaction (e.g. the initcode).
 
 ### Block processing
 
--- a/EIPS/eip-4788.md
+++ b/EIPS/eip-4788.md
@@ -31,7 +31,7 @@ restaking constructions, smart contract bridges, MEV mitigations and more.
 | `FORK_TIMESTAMP`             | TBD                                          |
 | `HISTORY_BUFFER_LENGTH`      | `8191`                                      |
 | `SYSTEM_ADDRESS`             | `0xfffffffffffffffffffffffffffffffffffffffe` |
-| `BEACON_ROOTS_ADDRESS`       | `0xbEaC02001Aedb23b4088c853e4C05c57491c8bCA` |
+| `BEACON_ROOTS_ADDRESS`       | `0xbEAC020008aFF7331c0A389CB2AAb67597567d7a` |
 
 ### Background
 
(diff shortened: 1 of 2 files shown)


EIPS/eip-4788.md
```diff
@@ -210,20 +224,20 @@ by working backwards from the desired deployment transaction:
...
   "maxPriorityFeePerGas": null,
   "maxFeePerGas": null,
   "value": "0x0",
-  "input": "0x60588060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5ffd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b6201800042064281555f359062018000015500",
+  "input": "0x60618060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604d57602036146024575f5ffd5b5f35801560495762016da0810690815414603c575f5ffd5b62016da001545f5260205ff35b5f5ffd5b62016da042064281555f359062016da0015500",
   "v": "0x1b",
   "r": "0x539",
-  "s": "0x133700f3a77843802897db",
-  "hash": "0x14789a20c0508b81ab7a0287a12a3a41ca960aa18244af8e98689e37ed569f07"
+  "s": "0x133700018971c643803096",
```
shemnon, line 235: Do the `r` and `s` values have any particular meaning?  I'm more curious as to why they s value had to change.  Was it searched to find the first valid signature of any address from "zero" or is there some other meaning?

lightclient: The `s` value is like the "nonce" for our address generator. We keep trying new `s` values until we get a deploy address with `beac02` prefix.

shemnon: Got it.  

Is there a canonical definition of what a synthetic address is and best practices to grind out and S value?  

In this particular case (r being 1337 in decimal, and s being 0x1337 left shifted 72 bytes, and then search upwards from there) it is obvious it should not have a known public key.  But I think the industry would benefit from a best practice for showing a transaction is a synthetic address transaction with a know r producing a corresponding s.

lightclient: I think this post is the pinnacle of our knowledge on the subject and it does not give any explicit instructions: https://weka.medium.com/how-to-send-ether-to-11-440-people-187e332566b7

SamWilsn: Are you talking about [Nothing-up-my-sleeve number](https://en.wikipedia.org/wiki/Nothing-up-my-sleeve_number)s?

adietrichs:
> this value will share 500-1000 if slot were to change

Not sure I understand this, but I created my own gist to compare: https://gist.github.com/adietrichs/4f41090c37d292a96b08c8cdbd076d90
Active means percent of slots actively used under that slot time, Inactive counts the extra slots that were written to under 12s, that are now no longer used.

For many slot times, `93600` results in a lower number of active (and total, taking into account the inactive ones) slots.

adietrichs:
I talked with @lightclient, and out of that came the idea to maybe go the other way: I updated the gist with a third option of a modulus of `8191`. That way we are always at 100% active slots, but across a much smaller total range (picked so that at 12s slot time, it still covers ~1d). I'll open a PR for that change.

adietrichs:
I have opened the PR: https://github.com/lightclient/EIPs/pull/7.
To keep conversation in one place, here a copy of the description:

`HISTORY_BUFFER_LENGTH` determines the overall size of the history region of the 4788 contract in storage. Both current proposed modulus, as well as the replacement proposed in the branch this PR is against, share prime factors with the current slot time (12s). This means that the usage of the history region will be sparse (only 1/12 of the storage slots in the region will be used). Furthermore, if we change the slot time in the future, this will change in somewhat arbitrary ways (depending on how many factors the new slot time and the modulus share). While this does not affect the functioning of the EIP, it does add unnecessary variance to two aspects, namely the total storage requirements and the total history depth available.

This PR changes the modulus to a prime number, so that there can never be any shared factors with the slot time. This way, 100% of the history region will be utilized, allowing for simple reasoning about total storage requirements and the total history depth, even in the case of future slot time changes. The specific value (`8191 = 2**13 - 1`) was chosen so that in the 12s slot time case it matches the total history depth of the original modulus as closely as possible.

EIPS/eip-4788.md
```diff
@@ -210,20 +224,21 @@ by working backwards from the desired deployment transaction:
...
   "to": null,
-  "gas": "0x27eac",
+  "gas": "0x3d090",
   "gasPrice": "0xe8d4a51000",
   "maxPriorityFeePerGas": null,
   "maxFeePerGas": null,
   "value": "0x0",
-  "input": "0x60588060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604457602036146024575f5ffd5b620180005f350680545f35146037575f5ffd5b6201800001545f5260205ff35b6201800042064281555f359062018000015500",
+  "input": "0x60618060095f395ff33373fffffffffffffffffffffffffffffffffffffffe14604d57602036146024575f5ffd5b5f35801560495762016da0810690815414603c575f5ffd5b62016da001545f5260205ff35b5f5ffd5b62
+016da042064281555f359062016da0015500",
```
ryanschneider, line 233: minor: just noticed the newline mid `input` is this intentional?

EIPS/eip-4788.md
```diff
@@ -158,24 +162,29 @@ push0
...
+push1 0x3c
 jumpi
 
 push0
 push0
 revert
 
 jumpdest
-push3 0x018000
+push3 0x016da0
```
marioevz, line 187: This is the part of the bytecode that was mentioned on ACD, it should be the same number as mentioned in line 32.

lightclient:
New beacon root address found by @spencer-tb! Thanks!

lightclient:
Final beacon root address is `0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02`.

EIPS/eip-4788.md
```diff
@@ -29,9 +29,9 @@ restaking constructions, smart contract bridges, MEV mitigations and more.
 | constants                    | value                                        |
 |---                           |---                                           |
 | `FORK_TIMESTAMP`             | TBD                                          |
-| `HISTORY_BUFFER_LENGTH`      | `98304`                                      |
+| `HISTORY_BUFFER_LENGTH`      | `8191`                                       |
 | `SYSTEM_ADDRESS`             | `0xfffffffffffffffffffffffffffffffffffffffe` |
-| `BEACON_ROOTS_ADDRESS`       | `0xbEac00dDB15f3B6d645C48263dC93862413A222D` |
+| `BEACON_ROOTS_ADDRESS`       | `0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02` |
```
chfast, line 34: Why was the address changed?

EIPS/eip-4788.md
```diff
@@ -158,24 +154,29 @@ push0
...
+push1 0x3c
 jumpi
 
 push0
 push0
 revert
 
 jumpdest
-push3 0x018000
+push3 0x001fff
```
chfast, line 179: This should be push2 now.

holiman: yeah, this was a miss. However, not worth fixing, the runtime execution gascost is identical, the only difference is that the code is one byte larger than it would have needed to be.

chfast: This repeats 4x so 4 bytes wasted. Forever :)

namiloh: Yeah. Now there are 16 bits less left in the world.

🚻 ☮️

PR #7676, opened 2023-09-09
Update EIP 7212: Clarifications and improvements

nalinbhardwaj:
(in order of importance)
- specify "Required Checks in Verification" more precisely
- make choice of verification vs. recovery clearer
- todo add links to implementation
- one ??? that i didn't understand

### DIFF
--- a/EIPS/eip-7212.md
+++ b/EIPS/eip-7212.md
@@ -1,7 +1,7 @@
 ---
 eip: 7212
 title: Precompiled for secp256r1 Curve Support
-description: Proposal to add precompiled contract that allows signature verifications in the “secp256r1” elliptic curve.
+description: Proposal to add precompiled contract that performs signature verifications in the “secp256r1” elliptic curve.
 author: Ulaş Erdoğan (@ulerdogan), Doğan Alpaslan (@doganalpaslan)
 discussions-to: https://ethereum-magicians.org/t/eip-7212-precompiled-for-secp256r1-curve-support/14789
 status: Review
@@ -12,7 +12,7 @@ created: 2023-06-22
 
 ## Abstract
 
-This proposal creates a precompiled contract that allows signature verifications in the “secp256r1” elliptic curve by given parameters of message hash, `r` - `s` components of the signature, and `x` - `y` coordinates of the public key. So that, any EVM chain -principally Ethereum rollups- will be able to integrate this precompiled contract easily.
+This proposal creates a precompiled contract that performs signature verifications in the “secp256r1” elliptic curve by given parameters of message hash, `r` and `s` components of the signature, and `x`, `y` coordinates of the public key. So that, any EVM chain - principally Ethereum rollups - will be able to integrate this precompiled contract easily.
 
 ## Motivation
 
@@ -35,23 +35,23 @@ As of `FORK_TIMESTAMP` in the integrated EVM chain, add precompiled contract `P2
 “secp256r1” is a specific elliptic curve, also known as “P-256” and “prime256v1” curves. The curve is defined with the following equation and domain parameters:
 
 ```
-# curve:
+# curve: short weierstrass form
 y^2 ≡ x^3 + ax + b
 
-# p: specifies reduced elliptic group
+# p: curve prime field modulus
 0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff
 
-# a: elliptic curve coefficient
+# a: elliptic curve short weierstrass first coefficient
 0xffffffff00000001000000000000000000000000fffffffffffffffffffffffc
 
-# b: elliptic curve coefficient
+# b: elliptic curve short weierstrass second coefficient
 0x5ac635d8aa3a93e7b3ebbd55769886bc651d06b0cc53b0f63bce3c3e27d2604b
 
 # G: base point of the subgroup
 (0x6b17d1f2e12c4247f8bce6e563a440f277037d812deb33a0f4a13945d898c296,
  0x4fe342e2fe1a7f9b8ee7eb4a7c0f9e162bce33576b315ececbb6406837bf51f5)
 
-# n: order of the subgroup
+# n: subgroup order (number of points)
 0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551
 
 # h: cofactor of the subgroup
@@ -68,7 +68,7 @@ The signature verifying algorithm takes the signed message hash, the signature c
 # pubKey = (public key of the signer private key)
 
 # Calculate the modular inverse of the signature proof:
-s1 = s^(−1)(modn)
+s1 = s^(−1) (mod n)
 
 # Recover the random point used during the signing:
 R' = (h * s1) * G + (r * s1) * pubKey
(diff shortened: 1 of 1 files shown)


EIPS/eip-7212.md
```diff
@@ -108,17 +108,17 @@ The use of signature verification cost by `P256VERIFY` is `3450` gas. Following
...
+“secp256r1” ECDSA signatures consist of `v`, `r`, and `s` components. While the `v` value makes it possible to recover the public key of the signer, most signers do not generate the `v` component of the signature since `r` and `s` are sufficient for verification. In order to provide an exact and more compatible implementation, verification is preferred over recovery for the precompile.
 
-The signature values in `r` - `s` and the public key coordinates in the `x`- `y` provides direct computations in signing and verification part, so these formats are chose in the input data format which are 32 bytes.
+The signature values in `r` and `s` and the public key `(x, y)` provides direct computations in signing and verification part, so these formats are chosen in the input data format which are 32 bytes. (Nalin: ??? what does this mean ???)
```
nalinbhardwaj, line 113: it's unclear to me what this paragraph means: would be nice to clarify

EIPS/eip-7212.md
```diff
@@ -81,13 +81,13 @@ r' == r
...
 
-Some requirements have to be checked to understand if the given signature components are valid:
+Following requirements **must** be checked by the precompile to verify signature components are valid:
 
-- Verify that both `r` - `s` values are greater than 0 and less than the curve order.
-- Verify that s is equal to or less than half of the order of the subgroup to prevent signature malleability.
-- Verify that the point formed by (`x`, `y`) values is on the curve and both components are in between 0 and the `p` value of the curve.
+- Verify that `r` and `s` values are valid scalar field elements, i.e. they are in `(0, n)` (exclusive) where `n` is the order of the subgroup.
+- Verify that `s <= n // 2`, where `n` is the order of the subgroup to prevent signature malleability.
+- Verify that the point formed by `(x, y)` is on the curve and both `x` and `y` are in `(0, p)` (exclusive) where `p` is the prime field modulus.
```
nalinbhardwaj, line 90: I would actually suggest that these be deleted from this EIP to match compatibility exactly with the NIST specs which does not specify this. In particular, if I understand correctly, at the moment a valid Implementation of P256 signing according to the NIST spec can fail this verifier. This would not be ideal for the purpose of this precompile.

It is anyway relatively straightforward/cheap to add these checks separately in EVM byte code if a contract implementation does want them.

EIPS/eip-7212.md
```diff
@@ -81,13 +81,13 @@ r' == r
...
+### Required Checks in Verification
 
-Some requirements have to be checked to understand if the given signature components are valid:
+Following requirements **must** be checked by the precompile to verify signature components are valid:
 
-- Verify that both `r` - `s` values are greater than 0 and less than the curve order.
-- Verify that s is equal to or less than half of the order of the subgroup to prevent signature malleability.
-- Verify that the point formed by (`x`, `y`) values is on the curve and both components are in between 0 and the `p` value of the curve.
+- Verify that `r` and `s` values are valid scalar field elements, i.e. they are in `(0, n)` (exclusive) where `n` is the order of the subgroup.
+- Verify that `s <= n // 2`, where `n` is the order of the subgroup to prevent signature malleability.
```
nalinbhardwaj, line 89: I would actually suggest that this be deleted from this EIP to match compatibility exactly with the NIST specs which does not specify the second. In particular, if I understand correctly, at the moment a valid Implementation of P256 signing according to the NIST spec can fail this verifier. This would not be ideal for the purpose of this precompile.

It is anyway relatively straightforward/cheap to add these checks separately in EVM byte code if a contract implementation does want them.

EIPS/eip-7212.md
```diff
@@ -169,7 +169,7 @@ geomean                                               33.00        7.000
 
 ## Reference Implementation
 
-Implementation of the `P256VERIFY` precompiled contract is applied to go-ethereum client to create a reference. Also, an “secp256r1” package has already been included in the Besu Native library which is used by Besu client. Other client implementations are in the future roadmap.
+Implementation of the `P256VERIFY` precompiled contract is applied to go-ethereum client to create a [reference](TODOLINK). Also, an “secp256r1” package [has already been included in the Besu Native library](TODOLINK) which is used by Besu client. Other client implementations are in the future roadmap.
```
nalinbhardwaj, line 172: could someone share the source for these?

EIPS/eip-7212.md
dcposch: i think they mean something like

> Existing P256 implementations verify (x, y, r, s) directly. We've chosen to match this style here, encoding each argument for the EVM as a `uint256`.

This is different from `ecrecover`, with the advantage that it 1. follows the NIST spec, 2. matches the rest of the (large) P256 ecosystem, and most importantly 3. allows execution clients to use existing well-vetted verifier implementations.

nalinbhardwaj: Nice explanation. Thanks, added your paragraph as-is to this PR.

EIPS/eip-7212.md
ulerdogan: I respect the idea of having compatibility between the standard and precompiled contract implementation, but I think that if we remove the malleability check, most of the applications using this code will require additional storage and checks. As in `ecrecover`, the malleability issue is not in the standards, but implemented in the precompiled contract. Is there any case which we need malleable signatures?

EIPS/eip-7212.md
ulerdogan: It's the [reference implementation](https://github.com/ethereum/go-ethereum/pull/27540), but external links are not allowed in the EIP itself. We can add as just text, not hyperlink.

EIPS/eip-7212.md
nalinbhardwaj: https://github.com/daimo-eth/daimo as an example application uses this precompile with AA wallets and does not require non-malleability (since nonces/userops are unique).

The case that strongly concerns me is that making this check inside the precompile means we would disagree from the NIST spec, and by extension signatures that by default are marked as valid by the most popular P256 implementations like [subtlecrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) and [noble](https://github.com/paulmillr/noble-curves) and all the test suites like [wycheproof](https://github.com/google/wycheproof) will be marked as invalid. Further there could exist device secure enclave’s that return signatures marked as invalid unless the user first runs a transform on them. This would be really confusing for users and annoying for implementation devs, and seems like unnecessary complexity.

Also, adding this check seperately as-needed in some sort of wrapper contract would not require any extra storage. As far as computation cost goes, users are likely to be calling a hash function before calling this precompile anyway, so the cost of one comparison seems trivial compared to that.

Overall, I think the point of this precompile should be to be able to onboard popular implementations to Ethereum (that all follow the NIST spec), so we should match the spec. If you would like, we can add an example usage or note to the EIP where we emphasise the malleability.

EIPS/eip-7212.md
nalinbhardwaj: gotcha, thanks, removing this then.

ulerdogan (approved):

PR #7716, opened 2023-09-14
Update EIP-7516: removes remaning references to 0x49

CarlBeek:
As @smartprogrammer93 pointed out in #7714, there is an opcode conflict with `BLOBHASH` with `0x49`. #7714 fixed this by changing `BLOBBASEFEE` to `0x4a`, but they missed a few references, this PR fixes them.

### DIFF
--- a/EIPS/eip-7516.md
+++ b/EIPS/eip-7516.md
@@ -13,7 +13,7 @@ requires: 3198, 4844
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


smartprogrammer93:
My bad!

CarlBeek:
> My bad!

All good, I missed in when reviewing your PR too

PR #7725, opened 2023-09-16
Add EIP: Atomic Storage Operations SCREDIT and SDEBIT

shemnon:
Two new opcodes that atomically mutate smart contract storage are proposed:
SCREDIT and SDEBIT. Overflow and underflow errors are enforced, reverting when
an unsigned 256-bit integer would overflow and underflow the limits of the
integer.

### DIFF
--- a/EIPS/eip-tbd.md
+++ b/EIPS/eip-tbd.md
@@ -1,4 +1,5 @@
 ---
+eip: TBD
 title: Atomic Storage Operations - SCREDIT and SDEBIT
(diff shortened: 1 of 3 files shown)


EIPS/eip-tbd.md
```diff
@@ -0,0 +1,199 @@
+---
+eip: TBD
```
abcoathup, line 2:
```suggestion
eip: 7519
```

g11tech (approved):
lgtm

PR #7736, opened 2023-09-19
Update EIP-7409: Move to Last Call

ThunderDeliverer:
Since the EIP-7409 is ready to be finalized, with discussion settling down, we are proposing it be moved to Last Call.


SamWilsn (approved):

eip-automerger:

 - Trying to change EIP 7409 state from Review to Last Call

PR #7743, opened 2023-09-20
Add EIP: OIDC ZK Verifier for AA Account

dongshu2013:

 - The build passes.


EIPS/eip-7743.md
```diff
@@ -0,0 +1,120 @@
+---
+eip: 7743
```
abcoathup, line 2:
```suggestion
eip: 7522
```

Please also update the filename & Eth Magicians post.

EIPS/eip-7743.md
```diff
@@ -0,0 +1,120 @@
...
+created: 2023-09-20
+---
+
+## Abstract
+
+Account Abstraction facilitates new use cases for smart accounts, empowering users with the ability to tailor authentication and recovery mechanisms to their specific needs. To unlock the potential for more convenient verification methods such as social login, we inevitably need to connect smart accounts and OpenID Connect(OIDC), given its status as the most widely accepted authentication protocol. In this EIP, we proposed a new model to use OIDC identity as the recovery mechanism for smart accounts. Users can link their accounts with OIDC identities and authorize an OIDC verifier to mutate their account state, such as resetting operators, by performing ZK verification against the linked OIDC identity. The corresponding interfaces are also defined in the spec.
+
+## Motivation
+
+Connecting OIDC identity and smart accounts has been a very interesting but challenging problem. The most common ways of tackling the issue is by utilizing Multi-Party Computation(MPC). However, the limitation of the MPC solution is obvious:
```
xinbenlv, line 19: Since ZKP and OpenID Connect is the central piece, it would be helpful you establish these two concepts as context properly, so I'd recommend sufficient introduction of them.

EIPS/eip-7743.md
```diff
@@ -0,0 +1,120 @@
...
+}
+
+interface IRecoverableAccount is IOpenIdAccount {
+    /// @notice get the operator of the account
+    /// @return the address of operator
+    function getOperator() external view returns(address);
+ 
+    /// @notice set operator of the account
+    /// @params the new operator to set
+    function setOperator(address operator) external;
```
xinbenlv, line 98: [Security] does it imply giving whoever could `setOperator` a very high privilegde? Should the setter get compromise, regardless of it being `IOpenIdVerifier` or other kind of contract accounts, is it additional risk?

[Technical Feedback] It seems the setOperator could be achieved by https://eips.ethereum.org/EIPS/eip-173 Ownable's `transferOwnership` or https://eips.ethereum.org/EIPS/eip-5982 for grantRole if needs more flexible ACL

EIPS/eip-7743.md
```diff
@@ -0,0 +1,120 @@
...
+    // @notice get verification key of the open id authenticator
+    function getVerificationKey() external view returns(bytes memory);
+ 
+    // @notice get id hash of account
+    function getIdHash(address account) external view returns(bytes32);
+
+    // @notice the function verifies the proof and executes the action if it's valid
+    // @params action: the action to execute
+    //         proof: data to prove the account ownership
+    function verifyAndCall(
```
xinbenlv, line 70: This could be used beyond ZKP use case but more general. Meanwhile, regular permission mechanism could also be used to approve things with ZKP proofs. Help us understand better what ZKP+AA makes the requirement of the ERC interface special?

EIPS/eip-7743.md
dongshu2013: 1. It really depends on how the account contract use it. This EIP is more about the IOpenIdVerifier not the account implementation. One may have very complicated account implementation with better permission control over setOperator function.
2. yeah I think ERC173 is sufficient as the example, will update the interface to inherit from ERC173 directly.

dongshu2013: For the first question, remember that IOpenIdVerifier is permissionless so the only way to compromise it is to hack your OpenId identity. One common enhancement is to add additional verifiers so the account can only be hacked if all linked open ids are hacked, or to add extra locking period to ensure account ownership cannot be transferred for a period of time.

EIPS/eip-7743.md
dongshu2013: The design is intentional. Regular permission mechanism per account is hard to deal with aggregated ZK proof, in which case we will need a third party contract(the verifier in this case) to process multiple accounts' recovery in batch. Another option is to merge it with the ERC4337 Aggregator in which case we will not need verifyAndCall but just verify function and the verifier has to be compatible with the IAggregator interface at ERC4337

EIPS/eip-7743.md
dongshu2013: It's hard to illustrate a specific ZKP proof structure since different ZKP algorithm may generate very different proofs. One of the goal of the design is to be ZK compatible but it can be used to support non-zk proofs as well. I will update the description to focus on the OpenID verification. How we do the OpenID verification is explained in the workflow and Rationale section but I can add some more context.

dongshu2013:
> Left some editorial and non-editorial comments.

Updates:
1. More contexts about ZKP and general OpenID verification as you suggested
2. The interface is updated to emphasize format of ZKP public input and ERC-4337 compatibility
3. Update the workflow to a more detailed version with how it can interact with ERC-4337 entry point and aggregator

xinbenlv (approved):

PR #7751, opened 2023-09-21
Update EIP-7425: Fix typos

omahs:


jimstir (commented):
Thanks for catching this.

PR #7757, opened 2023-09-21
Update EIP-7519: Fix typos

omahs:


shemnon (approved):
Please use "Fix tpyos" as the title next time.  For the lulz.

PR #7766, opened 2023-09-23
Update EIP-1767: Move to Draft

xinbenlv:
Resume the ERC-1767 drafting. Context: https://x.com/nicksdjohnson/status/1705476652124217748?s=20

### DIFF
--- a/EIPS/eip-1767.md
+++ b/EIPS/eip-1767.md
@@ -1,7 +1,7 @@
 ---
 eip: 1767
 title: GraphQL interface to Ethereum node data
-description: a GraphQL schema for accessing data stored on an Ethereum node to complete replacement to the read-only information exposed via the present JSON-RPC interface
+description: a GraphQL schema and interface to expose readonly Ethereum node data
 author: Nick Johnson (@arachnid), Raúl Kripalani (@raulk), Kris Shinn (@kshinn), Zainan Victor Zhou (@xinbenlv)
 discussions-to: https://ethereum-magicians.org/t/graphql-interface-to-ethereum-node-data/2710
 status: Review
--- a/EIPS/eip-1767.md
+++ b/EIPS/eip-1767.md
@@ -409,8 +409,9 @@ This schema implements the bulk of the current read-only functionality provided 
 
 For specific reasoning behind omitted functionality, see the Rationale section.
 
-## Implementation
+## Reference Implementation
 
+### Adoption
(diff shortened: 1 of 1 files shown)


lightclient:
We don't track the API changes by the EIP anymore and so this is now out-of-date. You can see the latest here: https://github.com/ethereum/execution-apis/blob/main/graphql.json

xinbenlv:
Thank! In that case, what's the best way to propose change for API?

lightclient:
I would make a PR to `execution-apis`. Here is a document outlining the process: https://github.com/ethereum/execution-apis/blob/main/docs/making-changes.md

xinbenlv:
Got it. 

- I checked the file for the right way to do it, it seems the last update was https://github.com/ethereum/execution-apis/pull/400/files#diff-a44507d2ad2482bb3ed863208b9e1e0393069edfdbd686593bac598901f36770R2169
which still cites an EIP (EIP-4895)?
- it seems EIP-1767 presents the rule change, and is being cited by https://geth.ethereum.org/docs/interacting-with-geth/rpc/graphql and a few other places which is why I was trying to move this to final to reflect the status, can we allow this EIP which is already adopted?
- At the end of the day, the current EIP-1 has still accept proposal about Interface, if the rule to propose change has recently it's been updated, and we don't want to finalize this EIP, maybe we shall update EIP-1 first?

lightclient:
> which still cites an EIP (EIP-4895)?

it's just referring to the eip where withdrawals were introduced to the protocol, same is done for tx types

> can we allow this EIP which is already adopted?

What exactly do you want to allow? You're free to try to update this, bring it to final, etc. But it won't change the "spec" for graphql. A better use of time is probably removing references to this EIP wherever you see it.

> maybe we shall update EIP-1 first?

What would the update be? I'd rather go ahead and deprecate this category all together and note that we've moved to `execution-apis` for these types of things, as we're discussing in the other PR.

xinbenlv:
Thanks for answering. Just to give a bit context: the reason that I am interested in help championing this EIP is because onto of that I like to further promote JSONRPC / GraphQL interface for native support of querying ERC-20 or ERC-721 transfer graph. In that direction, since it's not necessary that GraphQL interface provider needs to be EL clients, they could be pure indexer as well, I am wondering whether `execution-apis` is the best place for such proposal. 

For that EIP (JSONRPC/GraphQL for ERC-20/721 transfer graph) to move forward, it's easier if we have EIP-1767 finalized first and use it as a baseline and precedents to further propose

PR #7790, opened 2023-09-27
Update EIP-6963: make icon requirement wording consistent

jiexi:
Noticed this inconsistency with `SHOULD` and `MUST` for the icon aspect ratio and resolution. I believe this should be `SHOULD`


kdenhartog:
The way I interpret the MUST is that "it MUST be 96 x 96 or larger" and it SHOULD be a square. AKA

95x95 is not allowed
97x97 is allowed and expected
97x96 is allowed, but not expected

I think what's written is consistent currently

bpierre:
I think it would be preferable to have both requirements as `MUST`, so that UI authors can rely on the icon to be squared.

jiexi:
I don't think we have consensus on this one yet. So will hold off on making any changes

kdenhartog (approved):
Seems like we don't have any strong opposition to this change in the TG channel. This language seems more consistent now as well.

PR #7795, opened 2023-09-29
Update EIP-721: Fix typos

mitjat:


mitjat:
Giving up on this; the existing EIP text fails many different linters. I tried fixing the first wave (= make all mentions of ERC-XX into links) but more's needed, including a reqrite to match the chapter structure prescribed in EIP-1.

PR #7817, opened 2023-10-05
Add EIP: Getting Staked ERC-721 Ownership Recognition

sullof:
The ownership of ERC721 tokens when staked in a pool presents challenges, particularly when it involves older, non-lockable NFTs like, for example, Crypto Punks or Bored Ape Yacht Club (BAYC) tokens. This proposal introduces an interface to address these challenges by allowing staked NFTs to be recognized by their original owners, even after they've been staked.

Recent solutions involve retaining NFT ownership while "locking" the token when staked. However, this requires the NFT contract to implement lockable functionality. Many early vintage NFTs like CryptoPunks or Bored Ape Yacht Club were not originally designed as lockable.

When these non-lockable NFTs are staked, ownership transfers fully to the staking pool contract. This prevents the original owner from accessing valuable privileges and benefits associated with their NFTs.

For example:

A BAYC holder would lose access to the BAYC Yacht Club and member events when staked.
A CryptoPunks holder may miss out on special airdrops or displays only available to verified owners.
Owners of other early NFTs like EtherRocks would lose the social status of provable ownership when staked.
By maintaining a record of the original owner, the proposed interface allows these original perks to remain accessible even when the NFT is staked elsewhere. This compatibility is critical for vintage NFT projects lacking native locking mechanisms.

The interface provides a simple, elegant way to extend staking compatibility to legacy NFTs without affecting their core functionality or benefits of ownership.

### DIFF
--- a/EIPS/eip-7817.md
+++ b/EIPS/eip-7817.md
@@ -1,7 +1,7 @@
 ---
 eip: 7817
-title: Resolving Staked ERC721 Ownership Recognition
-description: An interface for recognizing original NFT ownership when staked to unsupported contracts.
+title: Getting Staked ERC-721 Ownership Recognition
+description: An interface for recognizing NFT ownership when staked to unsupported contracts.
 author: Francesco Sullo (@sullof)
 discussions-to: https://ethereum-magicians.org/t/eip-xxxx-resolving-staked-erc721-ownership-recognition/15967
 status: Draft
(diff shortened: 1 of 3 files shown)


EIPS/eip-7817.md
```diff
@@ -0,0 +1,73 @@
+---
+eip: 7817
```
abcoathup, line 2:
```suggestion
eip: 7531
```

sullof: Sure. Thanks

SamWilsn:

PR #7824, opened 2023-10-07
Update EIP-6963: name change + term consistency + overall copy edits

darrylyeo:
EIP-6963 is all about the **relationship between "Dapps" and "Wallet Providers"** – I want to ensure the copy and code examples reflect this clearly.

Naming things is _hard_. This PR should make the explanations more concise and terminology more consistent.

### New EIP title:

* "Multi Injected Provider Discovery"
  * → "Event-Driven Discovery of Wallet Providers"

(The previous title does not do this EIP justice in my opinion – "provider" is an overloaded term, and  "injected" is ambiguous as it could refer to the EIP-1193 provider object or the content script of a browser extension. Plus, it needs "Wallet" in the name!)

### Notable terminology clarifications:

* "DApp" / "application"
  * → "Dapp" (the "a" in "dapp" is [lowercase on Ethereum.org](https://ethereum.org/en/dapps/))
* "Wallet" / "provider" / "wallet provider"
  * → "Wallet Provider" (explicitly defined as a "browser extension" for the purposes of this spec)
* "library" / "Provider Discovery Library"
  * → "Wallet Provider Discovery Library" (usually interchangeable with "Dapp" for the purposes of this spec)
* "injected Wallet Provider" / "Ethereum provider" / "Ethereum interface" / "EIP-1193 provider" / "Provider object" 😵‍💫
  * → "EIP-1193 provider object"
* "injected script"
  * → "injected [content script](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts)", "content script injected by a browser extension", etc.
* "`window.ethereum`"
  * → "global JavaScript variable `window.ethereum`", "EIP-1193 provider object injected at `window.ethereum`", "legacy `window.ethereum` behavior", etc.
* "web page" / "session" / "lifetime of the page" / "over time"
  * → "page session" or "webpage session"
* "`Eip6963ProviderDetail` object" / "Wallet Provider"
  * → "EIP-6963 compliant Wallet Provider object", "`Eip6963ProviderDetail` object sent by Wallet Provider", etc.
* Events (reference by their JavaScript event name instead of TypeScript interfaces specific to this spec):
  * "`Eip6963AnnounceProviderEvent`" → "`eip6963:announceProvider` event"
  * "`Eip6963RequestProviderEvent`" → "`eip6963:requestProvider` event"
* "this EIP"
  * → "EIP-6963"

### Minor structural changes:
* "Wallet Provider info" section:
  * Link to "RDNS identifiers" sub-section (mirroring link to "Icon images" sub-section)
* "Events" section:
  * Add "Event loop" sub-heading

### <s>Minor spec changes:</s>
* <s>Allow `Eip6963RequestProviderEvent` to be an `Event` OR `CustomEvent`</s>
* <s>Explicitly disallow Dapps from rendering `<svg>` tags as-is</s>
* <s>Allow Dapps to render SVGs as an `` tag OR XSS-resistant "equivalent means" (such as a CSS background)</s>


glitch-txs (commented):
I'm a little bit against making breaking changes to the API unless they are really needed, since there are some wallets already supporting the EIP.
I also think "EIP" is the right way to call it since it's an acronym :)

kdenhartog (changes requested):
While this has some good editorial changes to it, I'm generally against making substantial changes 2 days before finalization which would require at least another 10 day period for wide review. Especially when it comes to modifying normative language which can have subtle implications on implementers.

For this reason, I do not thing this should be merged. In the future, changes like this should come much earlier within the EIP process in my opinion rather than at the very end of last call.

darrylyeo:
The only breaking change here would be disallowing Dapps from rendering `<svg>` tags as-is.

Everything else is a naming/stylistic/organizational choice and shouldn't affect new or existing implementers of the spec.

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
+- **`icon`** - a [URI][RFC-3986] referencing an image to represent the Wallet Provider. The image SHOULD be a square with 96x96px minimum resolution. See [Icon images](#icon-images) below for additional requirements.
+- **`rdns`** - a registered domain name from the Domain Name System associated with the Wallet Provider, in reverse domain name notation (e.g `com.example.subdomain`). It is RECOMMENDED this identifier remain the same throughout the development of the Wallet Provider. See [RDNS identifiers](#rdns-identifiers) below for additional requirements.
 
 ```typescript
 /**
- * Represents the assets needed to display a wallet
+ * The information needed for a Dapp to identify and display a Wallet Provider
  */
-interface EIP6963ProviderInfo {
+interface Eip6963ProviderInfo {
```
glitch-txs, line 58:
```suggestion
interface EIP6963ProviderInfo {
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
+### Wallet Provider announcement detail
 
-The `EIP6963ProviderDetail` is used as a composition interface to announce a Wallet Provider and related metadata about the Wallet Provider. The `EIP6963ProviderDetail` MUST contain an `info` property of type `EIP6963ProviderInfo` and a `provider` property of type `EIP1193Provider` defined by [EIP-1193](./eip-1193.md).
+The `Eip6963ProviderDetail` interface composes a Wallet Provider's display information and [EIP-1193][EIP-1193] provider object into a single object to be announced by the Wallet Provider. The `Eip6963ProviderDetail` MUST contain an `info` property of type `Eip6963ProviderInfo` and a `provider` property of type `Eip1193Provider` defined by [EIP-1193][EIP-1193].
 
 ```typescript
-interface EIP6963ProviderDetail {
-  info: EIP6963ProviderInfo;
-  provider: EIP1193Provider;
+interface Eip6963ProviderDetail {
```
```suggestion
interface EIP6963ProviderDetail {
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
+  info: Eip6963ProviderInfo;
```
glitch-txs, line 100:
```suggestion
  info: EIP6963ProviderInfo;
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
+  provider: Eip1193Provider;
```
glitch-txs, line 101:
```suggestion
  provider: EIP1193Provider;
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
-The `EIP6963AnnounceProviderEvent` interface MUST be a `CustomEvent` object with a `type` property containing a string value of `eip6963:announceProvider` and a `detail` property with an object value of type `EIP6963ProviderDetail`. The `EIP6963ProviderDetail` object SHOULD be frozen by calling `Object.freeze()` on the value of the `detail` property.
+#### Wallet Provider: `eip6963:announceProvider` event
+
+The `Eip6963AnnounceProviderEvent` MUST be a `CustomEvent` object with a `type` property containing a string value of `"eip6963:announceProvider"` and a `detail` property with an object value of type `Eip6963ProviderDetail`. The `Eip6963ProviderDetail` object SHOULD be frozen with `Object.freeze()`.
 
 ```typescript
-// Announce Event dispatched by a Wallet
-interface EIP6963AnnounceProviderEvent extends CustomEvent {
+// "Announce" event dispatched by a Wallet Provider
+interface Eip6963AnnounceProviderEvent extends CustomEvent {
```
glitch-txs, line 121:
```suggestion
interface EIP6963AnnounceProviderEvent extends CustomEvent {
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
   type: "eip6963:announceProvider";
-  detail: EIP6963ProviderDetail;
+  detail: Eip6963ProviderDetail;
```
glitch-txs, line 123:
```suggestion
  detail: EIP6963ProviderDetail;
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
 }
 ```
 
-The `EIP6963RequestProviderEvent` interface MUST be an `Event` object with a `type` property containing a string value of `eip6963:requestProvider`.
+#### Dapp: `eip6963:requestProvider` event
+
+The `Eip6963RequestProviderEvent` MUST be an `Event` (or `CustomEvent`) object with a `type` property containing a string value of `"eip6963:requestProvider"`.
```
glitch-txs, line 129:
```suggestion
The `EIP6963RequestProviderEvent` MUST be an `Event` (or `CustomEvent`) object with a `type` property containing a string value of `"eip6963:requestProvider"`.
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
 
 ```typescript
-// Request Event dispatched by a DApp
-interface EIP6963RequestProviderEvent extends Event {
+// "Request" event dispatched by a Dapp
+interface Eip6963RequestProviderEvent extends Event {
```
glitch-txs, line 133:
```suggestion
interface EIP6963RequestProviderEvent extends Event {
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
+The Wallet Provider MUST announce the `eip6963:announceProvider` event to the Dapp via the `window.dispatchEvent()` method. The Wallet Provider MUST add an event listener to catch an `eip6963:requestProvider` event dispatched from the Dapp. This event listener MUST use a handler that will re-dispatch an `eip6963:announceProvider` event.
+
+This re-announcement by the Wallet Provider is necessary in case the first announcement is delayed or fired before the Dapp has a chance to initialize its event listener.
+
+The Wallet Provider dispatches the `eip6963:announceProvider` event with immutable contents and listens to the `eip6963:requestProvider` event:
 
 ```typescript
-let info: EIP6963ProviderInfo;
-let provider: EIP1193Provider;
+let info: Eip6963ProviderInfo;
```
glitch-txs, line 147:
```suggestion
let info: EIP6963ProviderInfo;
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
+let provider: Eip1193Provider;
```
glitch-txs, line 148:
```suggestion
let provider: EIP1193Provider;
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
 
-const announceEvent: EIP6963AnnounceProviderEvent = new CustomEvent(
+const announceEvent: Eip6963AnnounceProviderEvent = new CustomEvent(
```
glitch-txs, line 150:
```suggestion
const announceEvent: EIP6963AnnounceProviderEvent= new CustomEvent(
```

EIPS/eip-6963.md
```diff
@@ -14,195 +14,211 @@ requires: 1193
...
-The DApp MUST listen for the `EIP6963AnnounceProviderEvent` dispatched by the Wallet via a `window.addEventListener()` method and MUST NOT remove the Event Listener for the lifetime of the page so that the DApp can continue to handle Events beyond the initial page load interaction. The DApp MUST dispatch the `EIP6963RequestProviderEvent` via a `window.dispatchEvent()` function call after the `EIP6963AnnounceProviderEvent` handler has been initialized.
+The Dapp MUST listen for the `eip6963:announceProvider` event dispatched by the Wallet via the `window.addEventListener()` method and MUST NOT remove the event listener for the lifetime of the page session in order to continue handling events beyond the initial page load interaction. The Dapp MUST dispatch the `eip6963:requestProvider` via the `window.dispatchEvent()` method after the `eip6963:announceProvider` handler has been initialized.
 
 ```typescript
-// The DApp listens to announced providers
+// The Dapp listens for announced Wallet Providers
 window.addEventListener(
   "eip6963:announceProvider",
-  (event: EIP6963AnnounceProviderEvent) => {}
+  (event: Eip6963AnnounceProviderEvent) => {}
```
glitch-txs, line 172:
```suggestion
  (event: EIP6963AnnounceProviderEvent) => {}
```

glitch-txs (changes requested):
there are some left, but I consider these as breaking changes.

I also agree with the comment about, seems like there are a lot of changes and we're hitting final in a few days. I think it would be better to make changes only if it's really needed.

kdenhartog:
> The only breaking change here would be disallowing Dapps from rendering `<svg>` tags as-is.
>
> The other spec changes loosen the requirements to use `Event` / `` exclusively.
>
> Everything else is a naming/stylistic/organizational choice and shouldn't affect new or existing implementers of the spec.

Any breaking change without large consensus here is unnecessary and potentially harmful at this point. At this point every change should be questioned substantially whether it warrants necessary change. If we think of this spec as concrete drying we're at the point where the tape is still up but it's basically ready to walk on. Now is not the time to try and spray a bunch of water on it so that we can resurface the concrete. 

Also, we shouldn't assume that just because a change is non normative that it doesn't matter. The implications of subtle interpretation differences can be profound. Especially when making changes to examples which are by definition non-normative but often times the most important parts of the spec.

In my opinion we shouldn't be making changes at this point unless extremely necessary in order to enhance conformance. I do not believe these changes lead to enhanced conformance so therefore am against making these changes.

darrylyeo:
Reverting all proposed normative changes to avoid restarting the Last Call period.

darrylyeo:
> there are some left, but I consider these as breaking changes.

@glitch-txs As I understand it, this spec is JavaScript-based and the TypeScript syntax is used illustratively, so changing the TypeScript interface names used as internal references shouldn't break any JavaScript implementations.

Using TitleCase as in `Eip6963` is my personal style preference. Not the most important part of my proposal though – happy to leave as-is fully capitalized as `EIP6963` if others prefer :)

darrylyeo:
> Also, we shouldn't assume that just because a change is non normative that it doesn't matter. The implications of subtle interpretation differences can be profound. Especially when making changes to examples which are by definition non-normative but often times the most important parts of the spec.

@kdenhartog Agreed with these points – my hope is that by consolidating terminology and applying them more consistently across the document, we can prevent these subtle interpretation differences after the EIP moves to Final.

I took great care to preserve the original intent of the non-normative text as much as possible – happy to address any specific parts you feel I have not!

darrylyeo:
If it helps peopl read the code diff more easily — my revisions stem from using the **four terms under the "Definitions" section** as consistently as possible.

If anyone strongly prefers an alternative wording for any of the four terms, let me know and I can easily find-and-replace.

EIPS/eip-6963.md
```diff
@@ -38,171 +38,189 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
...
+Since SVG images can execute JavaScript, Dapps and Wallet Provider Discovery Libraries MUST render SVG images using the `<img>` tag to ensure no untrusted JavaScript execution can occur.
 
-#### RDNS
+#### RDNS identifiers
 
-The **`rdns`** (Reverse-DNS) property serves to provide an identifier which DApps can rely on to be stable between sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions.
-The Reverse-DNS convention implies that the value should start with a reversed DNS domain name controlled by the Provider. The domain name should be followed by a subdomain or a product name. Example: `com.example.MyBrowserWallet`.
+The `rdns` (reverse-DNS) property serves as an identifier which Dapps can rely on to be stable between page sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions. The domain should be controlled by the Wallet Provider and be followed by a subdomain or a product name. Example: `com.example.MyBrowserWallet`.
```
bumblefudge, line 87:
```suggestion
The `rdns` (reverse-DNS) property serves as an identifier which Dapps can rely on to be stable between page sessions. The Reverse Domain Name Notation is chosen to prevent namespace collisions. The domain SHOULD be controlled by the Wallet Provider and be followed by a subdomain or a product name. Example: `com.example.MyBrowserWallet`.
```

EIPS/eip-6963.md
```diff
@@ -38,171 +38,189 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
...
+
+The orchestration of events described above guarantees that the Dapp is able to discover the Wallet Provider, regardless of whether the Wallet Provider code or the Dapp code executes first. Additionally, multiple Wallet Providers can respond to requests from a single Dapp without polluting the `window.ethereum` namespace, which has historically resulted in non-deterministic behavior where a different Wallet Provider is connected for each page session.
 
-To follow the Javascript event name conventions, the names are written in present tense and are prefixed with the number of this document (`EIP6963`). 
+### Event names
+
+Event names are written in present tense in accordance with JavaScript conventions, and are prefixed with the number of this document (`eip6963:`) to avoid namespace collisions.
```
bumblefudge, line 192: i'm not sure i understand the insertion of the `:` namespacing symbol here?

glitch-txs:
I think we should keep the previous name for types. Imo it is a breaking change since most of us use typescript and have this EIP as reference/source of truth.

About the title I just wanted to point out that Wagmi has a library called `mpid` and some wallets are already using it, sharing it just to be aware of early implementations.

darrylyeo:
@glitch-txs I'd be inclined to agree if the EIP explicitly mentioned using TypeScript as a normative requirement. While most existing/early implementations today are TypeScript-based it may not be that way in the future.

I'm trying to avoid being overly prescriptive, especially for current or future implementers who don't prefer to use TypeScript.

EIPS/eip-6963.md
darrylyeo: This is referring to the JavaScript event names `eip6963:requestProvider` and `eip6963:announceProvider`.

PR #7834, opened 2023-10-11
Update EIP-6956: Fix typos

xiaolou86:


steven2308 (approved):

mpeyfuss (approved):

tbergmueller:
Hi all,
@xiaolou86  many thanks for fixing the typos!

Since if somebody takes the time to fix typos, I assume that also interest in the content may be there or it may be even implemented already.

Saftey advise in case you're implementing: 
We plan to simplify the EIP in the coming weeks/month significantly. In a nutshell
-  the mapping ``anchor <> tokenID` will become optional, since in many use-cases it is not needed and a randomized tokenId (leveraging the full uint256) has the same level of security. 
- This is expected to decrease gas significantly and simplifies interaction with the contract. 
 - this should be possible without breaking changes.
 - (Events may only carry tokenId, some may become obsolete) 

Will also add this to https://ethereum-magicians.org/t/erc-6956-asset-bound-non-fungible-tokens/14056

tbergmueller (approved):

shuoer86:
@tbergmueller thank you for your introduction. I am a newbie and studying about web3.

xiaolou86:
@tbergmueller thank you for introduction. I am beginning to study web3.

PR #7836, opened 2023-10-11
Update EIP-6800: pedersen hash endianness and group to field fixes

jsign:
This PR clarifies some important details:

How 16-bytes chunks used in the trie-key hashing should be interpreted when mapping them to scalar-field elements.
How points are transformed into scalar-field elements. This is in the context of mapping points to vector elements in the tree nodes.

### DIFF
--- a/EIPS/eip-6800.md
+++ b/EIPS/eip-6800.md
@@ -42,7 +42,7 @@ def group_to_field(point: Point) -> int:
     if point == bandersnatch.Z:
         return 0
     else:
-        return point.map_to_scalar_field()
+        return int.from_bytes(point.map_to_base_field().to_bytes(32, 'little'), 'little') % BANDERSNATCH_MODULUS
     
 def compute_commitment_root(children: Sequence[int]) -> Point:
     o = bandersnatch.Z
--- a/EIPS/eip-6800.md
+++ b/EIPS/eip-6800.md
@@ -35,7 +35,7 @@ BANDERSNATCH_MODULUS = \
 PEDERSEN_BASIS = [....]
 VERKLE_NODE_WIDTH = len(PEDERSEN_BASIS)
 
-def group_to_field(point: Point) -> int:
+def group_to_scalar_field(point: Point) -> int:
     # Not collision resistant. Not random oracle. 
     # Binding for Pedersen commitments.
     assert isinstance(point, Point)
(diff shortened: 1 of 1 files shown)


EIPS/eip-6800.md
```diff
@@ -144,7 +144,7 @@ def pedersen_hash(inp: bytes) -> bytes32:
     # Interpret input as list of 128 bit (16 byte) integers
     ext_input = inp + b"\0" * (255 * 16 - len(inp))
     ints = [2 + 256 * len(inp)] + \
-           [int.from_bytes(ext_input[16 * i:16 * (i + 1)]) for i in range(255)]
+           [int.from_bytes(ext_input[16 * i:16 * (i + 1)], 'little') for i in range(255)]
```
jsign, line 147: By default `int.from_bytes` endianness interpretation is big-integer, but currently both the Go and Rust implementation matches in little-endian interpretation. Also, all other `int.from_bytes` calls in this EIP use the little-endian interpretation.

It might be fair to ask if the EIP should change or the implementations should change, but from what I've asked it looks like it might be a mistake in the EIP.

References on how this works today in implementations:
- [Geth](https://github.com/gballet/go-ethereum/blob/kaustinen-with-shapella/trie/utils/verkle.go#L107-L110)
- [Rust entrypoint](https://github.com/crate-crypto/rust-verkle/blob/master/verkle-spec/src/lib.rs#L44), which then maps to exactly "this line" in the EIP [here](https://github.com/crate-crypto/rust-verkle/blob/master/verkle-spec/src/util.rs#L97).

The fact that both (and others like Nethermind) match is pretty obvious since they've already agreed on previous testnets. If that wasn't the case, it would be immediately obvious when clients sync.

Coming back on changing the impl or changing specs, I'll ask for the final call to any EIP authors.

EIPS/eip-6800.md
```diff
@@ -42,7 +42,7 @@ def group_to_field(point: Point) -> int:
     if point == bandersnatch.Z:
         return 0
     else:
-        return int.from_bytes(point.serialize(), 'little') % BANDERSNATCH_MODULUS
+        return int.from_bytes(point.map_to_base_field().to_bytes(32, 'little'), 'little') % BANDERSNATCH_MODULUS
```
jsign, line 45: This was incorrect since we don't serialize the point to map a point to a scalar-field element. (i.e: we do `X/Y`  (affine form)). Despite `serialize()` isn't strictly defined, it clearly (IMHO) doesn't refer to this.

To avoid messing with coordinate systems, I "hand wave" it with a `map_to_base_field()` call, which does what it says: maps the point to a base-field element (i.e: `X/Y`). Then we do the transformation of that to bytes, and re-interpret as a scalar-field.

All the above can be also be summarized in code as shown in [go-ipa here](https://github.com/crate-crypto/go-ipa/blob/master/banderwagon/element.go#L259-L264).

Whenever a cryptography-spec repo exists, we can decide if this is OK. Personally, it'd be better to do a `map_to_scalar_field()` API here; but if I do that _today_ it explains literally "nothing" about what to do. The current proposal is a bit less magical (for now).

EIPS/eip-6800.md
gballet: some historical context here: that was agreed informally during meetings and never specified. Changing the endianness will cause a lot of issues we experienced when we settled down on little endian (at the time, we were using different KZG libraries, each with their own endianness).

PR #7845, opened 2023-10-12
Update EIP-7015: allow arbitratry `TYPEHASH` and correponding fields for `structHash`

oveddan:
This updates EIP-7015 to allow contracts to specify the TYPEHASH and corresponding encoded `structHash` as desired, as long as the `TYPEHASH` follows the EIP-712 standard.

This provides some benefits:
* This allows contracts to specify their own fields for the `TYPEHASH` rather than conforming it to always be `CreatorAttribution(bytes32 structHash)`
* When signing the message in the client, we can now use `signTypedData` and the explicit types to sign can be defined, and the popup that opens in the web3 wallet shows the explicit fields that are signed.

Without this change, they'd just be signing the hashed `structHash` without actually knowing what they're signing.

This update also provides example code for how to sign the message and decode the `CreatorAttribution` event.


EIPS/eip-7015.md
```diff
@@ -29,22 +29,29 @@ This EIP takes advantage of the fact that contract addresses can be precomputed
...
-Creator consent is given by signing an [EIP-712](./eip-712.md) compatible message; all signatures compliant with this EIP MUST include all fields defined. The struct signed is:
+Creator consent is given by signing an [EIP-712](./eip-712.md) compatible message; all signatures compliant with this EIP MUST include all fields defined. The struct signed can be any arbitrary data that defines how to create the token; it must hashed in an EIP-712 compatible format with a proper EIP-712 domain. 
+
+The following shows some examples of structs that could be encoded into `structHash` (defined below):
 
 ```solidity
+// defines that a token can be created with a specified id and metadata uri
 struct TokenCreation {
-	bytes32 structHash;
+  uint256 tokenId;
```
strollinghome, line 39: Can we remove the `tokenId` the standard's attribution applies to all token-ids so this example might be a little misleading/confusing.

oveddan: done, made it more generic

strollinghome:
lgtm!

strollinghome (approved):

eip-automerger:

 - File assets/eip-7015/contracts/DelegatedErc721.sol is not an EIP
 - File assets/eip-7015/contracts/EIP7015.sol is not an EIP
 - File assets/eip-7015/js-test/abi.ts is not an EIP
 - File assets/eip-7015/js-test/delegatedErc721.test.ts is not an EIP
 - File assets/eip-7015/package.json is not an EIP

PR #7850, opened 2023-10-14
Update EIP-6551: Add ERC6551 address

SebastienGllmt:
Many guides and tools already use this hard-coded registry address across many chains (with 2500+ txs on this address on Ethereum mainnet)

Given the registry address is meant to be static and the fact it's actively used, it may be good to add it to the EIP. If a change to the spec causes the address to change in the future, probably this address should be mentioned anyway as a migration path for any tool that used it

### DIFF
--- a/EIPS/eip-6551.md
+++ b/EIPS/eip-6551.md
@@ -77,7 +77,7 @@ For example, the token bound account with implementation address `0xbebebebebebe
 
 Each token bound account proxy SHALL delegate execution to a contract that implements the `IERC6551Account` interface.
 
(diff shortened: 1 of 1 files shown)


jaydenwindle:
Thanks for the PR @SebastienGllmt! An [upcoming update](https://github.com/jaydenwindle/EIPs/blob/a3f7e71090042ff051fb4aadc65030d8de8894e2/EIPS/eip-6551.md?plain=1#L58) to the proposal is going to include the finalized registry address. Previously deployed Registry contracts have always been subject to change (see the warning in the [reference repo](https://github.com/erc6551/reference#erc-6551-reference-implementation)) and will be deprecated once this proposal is finalized.

All previously deployed Registry contract addresses are documented in the [release history](https://github.com/erc6551/reference/releases) of the reference repo. Since they are documented elsewhere, I don't think that older Registry addresses should be included in the proposal document itself.

SamWilsn:

PR #6485, opened 2023-02-09
Add EIP 223: Token standard with event handling implementation

Dexaran:
---
title: Token Standard
description: ERC223 token implementing a communication model for interaction between contracts
author: <Dexaran (@Dexaran) <dexaran@ethereumclassic.org>
discussions-to: https://github.com/ethereum/EIPs/issues/223
status: Review
type: <Standards Track>
category: <ERC>
created: <2017-05-03>
---

PR #7856, opened 2023-10-16
Update EIP-223: Remove `standard` function

Pandapip1:
ERC-223 needs an errata:

1. In https://github.com/ethereum/EIPs/pull/6485#discussion_r1103853086, I noted that the `standard` function was out-of-scope and insisted that it be removed
2. In https://github.com/ethereum/EIPs/pull/6485#discussion_r1124870613, @Dexaran acknowledged that it was indeed out-of-scope and removed it
3. In https://github.com/ethereum/EIPs/pull/7466 / https://github.com/ethereum/EIPs/commit/ff3ec32c03398c523e57df2ababcc2da7f0a0e65, the function was mistakenly re-introduced
4. The optional function remains out-of-scope and inadequately specified.

Removing this function is a backward compatible change, as applications cannot expect it to be present.

### DIFF
--- a/EIPS/eip-223.md
+++ b/EIPS/eip-223.md
@@ -70,16 +70,6 @@ Returns the number of decimals of the token. The functionality of this method is
 
 OPTIONAL - This method can be used to improve usability, but interfaces and other contracts MUST NOT expect these values to be present. 
 
-##### `standard`
-
-```solidity
-function standard() view returns (string memory)  
-```
-
-Returns the identifier of the standard. If the token is [ERC-223](./eip-20.md) then it must return `"223"`.
-
-OPTIONAL - This method can be used to improve usability, but interfaces and other contracts MUST NOT expect these values to be present.
-
 ##### `balanceOf`
 
 ```solidity


Dexaran:
I would approve this changes but the removal of this function now would require the update of the reference implementation code as well since the function is also presented there. The reference implementation was based on the 


I'm also working on a Token Standard Converter as according to [EIP-7417](https://github.com/ethereum/EIPs/pull/7418) (btw that would help if you could merge the EIP-7417 PR, I'm waiting for it for 3 weeks without any response)

The `standard()` function is also part of the token converter now https://github.com/Dexaran/TokenStandardConverter/blob/main/TokenConverter.sol#L135 as it's the best way to identify if a token is supposed to work as a ERC-20 token or as a ERC-223 token. EIP-165 does not provide a deterministic answer since ERC-20 and ERC-223 token interfaces can match even though the logic of the transferring is different.

Even though the "standard recognition" can be reinstated as a separate informational EIP here is the question - is it worth the effort to remove it from a final EIP and reference code just  to make it a separate EIP?

Pandapip1:
> I would support this changes but the removal of this function now would require the update of the reference implementation code as well since the function is also presented there.

Actually, it's perfectly fine for you to keep the standard function there! If you'd like it removed, however, I'd be happy to comply.

> I'm also working on a Token Standard Converter as according to https://github.com/ethereum/EIPs/pull/7418 (btw that would help if you could merge the EIP-7417 PR, I'm waiting for it for 3 weeks without any response)

Sorry about the wait. I'll look into it!

> The standard() function is also part of the token converter (https://dexaran.github.io/token-converter/) now https://github.com/Dexaran/TokenStandardConverter/blob/main/TokenConverter.sol#L135 as it's the best way to identify if a token is supposed to work as a ERC-20 token or as a ERC-223 token. ERC-165 does not provide a deterministic answer since ERC-20 and ERC-223 token interfaces can match even though the logic of the transferring is different.

The interface ID is the XOR of all the function selectors, so they *should* be different. https://eips.ethereum.org/EIPS/eip-5269 and https://eips.ethereum.org/EIPS/eip-1046 might also interest you, although the latter applies less directly to your token converter EIP.

> Even though the "standard recognition" can be reinstated as a separate informational EIP here is the question - is it worth the effort to remove it from a final EIP and reference code just to make it a separate EIP?

Considering most if not all of the work would be done by editors, I'd say yes, this is worth it. Take a look at ERC-165 and/or ERC-5269 and let me know what you think! ERC-5269 is still in Review, so if you have any strong feelings, @xinbenlv will probably be able to collaborate with you to refine them!

SamWilsn:

PR #7867, opened 2023-10-17
Update EIP-6956: fix a couple of typos

shuoer86:


tbergmueller (approved):
Thank you @shuoer86  (once again) for fixing typos! 
Since you 're now the second one to compensate our grammatical shortcomings, we'll certainly take better care in the upcoming changes ;)

PR #7886, opened 2023-10-20
Update EIP-20: Remove external implementation links

Dexaran:
I am proposing to remove external links to the reference implementations of ERC-20 tokens and replace them with a code written in the text of the original proposal. This PR removes the links.

There are the following reasons to do this:

### 1. Deprecation

The links are outdated significantly and the code is written in Solidity 0.4.x versions. It is not advised to use it in 2023 and we are no longer relying on `.add(...)`, `.sub(...)` expressions. Now these should be considered anti-patterns as they reduce the readability of the code.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9b3710465583284b8c4c5d2245749246bb2e0094/contracts/token/ERC20/ERC20.sol#L105

### 2. MIT license

The content in the links is licensed under MIT while the EIP process requires that the content would be licensed under CC0.

https://github.com/Consensys/Tokens/blob/fdf687c69d998266a95f15216b1955a4965a0a6d/LICENSE

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9b3710465583284b8c4c5d2245749246bb2e0094/LICENSE

### 3. Lack of neutrality

The developers of the initial reference implementation benefit from this commercially. Being presented as the original implementer and having a link to your service in the EIP text at the time when all your competitors can't do the same is a significant marketing advantage that Ethereum community gives to the authors of the old code. For example new EIP authors / implementers can't promote themselves using the EIP procedure in this way. Being presented in the text of the original EIP may allow the implementor to gain the traffic for their auditing services (which are not guaranteed to be the highest quality auditing services in the industry however).

The fact that one party implemented the first version of the code does not automatically guarantee that the code they provide today is somehow better than the code their competitors may provide.

I would recommend replacing a link with a more neutral text-written code. If this code would not suffice the needs of EIP reader then he will do the research on their own to find the best available implementation which would allow for healthier competition.


Pandapip1 (approved):
+1. The links provide no value and present certain implementations as being more 'correct.'

abcoathup:
Final ERCs should be final.  

Other ERCs have links/references to specific implementations e.g. https://eips.ethereum.org/EIPS/eip-721#implementations
If the approach is to remove implementations for the reasons given above then this should be done for every ERC which has them.  Alternatively the implementations section could link to a wiki where modern, supported implementations can be listed and maintained.

SamWilsn (changes requested):
Requesting changes so the bot doesn't merge until https://github.com/ethereum-cat-herders/EIPIP/issues/287 is resolved.

Dexaran:
>Final ERCs should be final.

If we want to develop a healthy ecosystem this shouldn't apply to codes.

Hardforks and specifications can be immutable. Codes can not. It is quite a common situation that a flaw or a vulnerability is discovered in a program that was "finalized" years ago. Software evolves over time.

>If the approach is to remove implementations for the reasons given above then this should be done for every ERC which has them.

Yes, I think it's a good idea to replace all links with text-written codes attached directly to the .md file of the EIP.

SamWilsn:

PR #7906, opened 2023-10-22
Update EIP-7542: Update EIP-7542 to reflect its alignment with EIP-4444

smartprogrammer93:
Quoting 4444 from the user experience section:
> In the first phase, clients don’t prune historical data by default. They introduce a command line option similar to `geth’s --txlookuplimit` that users can use if they want to prune historical data.


domothyb:
should an explicit dependency on 4444 be listed on line 11?

smartprogrammer93:
@domothyb I don't believe it is. as this will need to be implemented in first stage before 4444 is fully implemented. Also, the network could benefit from this EIP for the syncing peers regardless if we prune history or not. 
As such i don't see the dependency exist, but I can be wrong.

lightclient:
@smartprogrammer93 fyi there is an issue when you undraft a PR the eth-bot doesn't review it. So sometime you need to open and close it.

smartprogrammer93:
Thank you @lightclient  🙏

PR #7932, opened 2023-10-28
Update EIP-4844: fix typo on blob batch verification function name

ralexstokes:
Use latest function name for batched blob verification


ppopth (approved):
lgtm

ppopth:
This reminds me that the EIP is not often updated

protolambda (approved):
LGTM

PR #7949, opened 2023-11-03
Add EIP: Separated Payer Transaction

adietrichs:
Create an EIP for a new separated payer transaction type.

Todo before merge:

- [ ] write rationale

### DIFF
--- a/EIPS/eip-XXXX.md
+++ b/EIPS/eip-XXXX.md
@@ -1,0 +1,93 @@
+---
+eip: XXXX
+title: Separated Payer Transaction
+author: Tomasz Stanczak (@tkstanczak), Ansgar Dietrichs (@adietrichs)
+discussions-to: tbd
+status: Draft
+type: Standards Track
+category: Core
+created: 2023-11-03
+---
+
+## Simple Summary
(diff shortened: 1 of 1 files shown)


EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,93 @@
+---
+eip: XXXX
```
abcoathup, line 2:
```suggestion
eip: 7553
```

domothyb:
I quite like the idea behind this EIP! I believe it could be enhanced significantly if there was an additional EVM opcode that returns the address of the payer, so the sender's call could involve any arbitrary logic, but most commonly sending any token/stablecoin to `tx.payer`, whoever that happens to be. Or have DEXs integrate `tx.payer` natively so the user doesn't have to even think about gas fees at all. 

EOAs could then broadcast a signed partially filled transaction of this type to the mempool, then it's up to whoever wants to to fill out the rest and sign as the sender and get it included in a block and pay the actual gas fee in ETH

lightclient:
@domothyb can you post this on eth mag please? We try to avoid EIP discussion in the PR.

PR #7984, opened 2023-11-23
Update EIP-5656: Move to Last Call

chfast:


SamWilsn (changes requested):
You'll need to fill in the security considerations section before we can merge. Aside from maybe some gas or buffer re-use problems, I doubt there's anything of substance? You can just put "None" if there aren't any.

g11tech (changes requested):
ahh sorry I see changed requested by @SamWilsn

PR #7995, opened 2023-11-28
Update EIP-7516: Move to Last Call

CarlBeek:


CarlBeek:
Closing PR until 4844 is Final Review and we have a Dencun fork date.

PR #8000, opened 2023-12-01
Update EIP-6953: Move to Last Call

timbeiko:


poojaranjan:
This EIP contains the list of post-merge upgrades (Line 67). 

Do we intend to update this EIP to include Dencun and beyond upgrades? 

If so, should we consider moving it to `Living` instead of `Last Call`?

SamWilsn (approved):
Once this goes to Final, no further edits will be permitted. Just want to make sure that the description will still make sense ("exhaustive")?

PR #8011, opened 2023-12-02
Update EIP-3074: update 3074 MAGIC byte

anna-carroll:
per [this comment](https://ethereum-magicians.org/t/eip-3074-auth-and-authcall-opcodes/4880/133?u=anna-carroll), EIP-4844 blob transactions use `0x03` as the EIP-2718 TransactionType prefix. it was suggested that EIP-3074 `MAGIC` byte should choose a different prefix to avoid any possibility of collision.


lightclient (approved):
Definitely should shift this now given 4844. Thank you!

PR #8058, opened 2023-12-24
Add EIP: Increase default block gas target to 20M

benaadams:
Increase default block gas target by 33.3% from 15M to 20M (and block gaslimit from 30M to 40M)


EIPS/eip-7581.md
```diff
@@ -0,0 +1,68 @@
+---
+eip: 7581
+title: Increase block gas target and gaslimit
```
g11tech, line 3: it would be better so specify final target numbers here so as to be unambiguous w.r.t. future limit raising EIPs

EIPS/eip-7581.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+## Specification
+
+* Increase default block gaslimit from 30M to 45M
+* Target remains half of that increasing from 15M to 22.5M 
+
+## Rationale
+
+While increasing block speed could also be an option it may have issues due to assumptions that have crept in or larger consensus times (which [EIP-7251](./eip-7251.md) seeks to address to some degree)
```
g11tech, line 30: this is not rationale, could you pls give some analysis of why exactly these numbers are being suggested

EIPS/eip-7581.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Security Considerations
+
+[EIP-1153](./eip-1153.md) `TSTORE` presents a new way to allocate memory on a node with linear cost. In other words, each TSTORE allows the developer to store 32 bytes for 100 gas, excluding any other required operations to prepare the stack. Given 45 million gas, the maximum amount of memory that can be allocated using TSTORE is:
```
g11tech, line 38: how is this a security consideration?

lightclient (commented):
Raising the gas limit isn't a consensus change, so I'm not sure if this should be a Core EIP. Really more informational for validators to choose.

EIPS/eip-7581.md
benaadams: Done

benaadams:
> Raising the gas limit isn't a consensus change, so I'm not sure if this should be a Core EIP. Really more informational for validators to choose.

While raising the actual gaslimit is in the hands of the validators; this is a proposal to raise the "default" gaslimit in all the execution clients for mainnet. So the mainnet limit that would be used in a vanilla install without the gaslimit being explicitly specified.

I'm going via EIP process to
1. Achieve consensus with ACD if/that they believe their execution clients can support the increased limit (and come to a more refined value).
2. Start a more focused discussion about the target block gaslimit where all interested parties (core devs, validators, blockbuilders, users, dapp devs, etc) can give feedback rather than shouting into the twitter algorithm.
3. Adjust in a more coordinated fashion with the previous considerations.

While its true 51% of the validators could just raise the limit; this hasn't happened in over two years which suggests the network is likely operating conservatively; i.e. it's working, don't change things in case it doesn't.

This EIP if accepted (depending on feedback) will achieve 4 things:
1. Signal to the validators that the core devs are confident that the software can handle the higher load
2. Imply a vote to increase the limit to any validators that don't explicitly set the gaslimit on following software upgrade
3. Indicate to blockbuilders to start building bigger blocks
4. Be a clear sign to other 3rd parties to update other materials, like docs https://ethereum.org/en/developers/docs/blocks/#block-size where its full specified that 15M/30M is the blocksize
image

---

## Aside on blocksize vs blockspeed

### Blocksize = **Throughput** 

Increasing block gaslimit (gently) shouldn't be too problematics as bandwidth and storage capacities are always increasing at technological rates.

For example looking at the hardware side of things; at that time of EIP-1559 a 2TB SAMSUNG 980 PRO SSD 2TB NVMe Gen 4 cost $429; it now costs $139; and you can now buy a 4TB SAMSUNG 990 PRO SSD for $319; which is double the space for 25% less than the price of a recommended drive at the time of EIP-1559

image


### Blockspeed = **Latency** 

Leave block times the same; speed of light isn't increasing, future protocol changes may require extra time and can't slow down later; and dependencies may have crept into the system. Leave it to L2s to have faster block times.

Also on mainnet block builders use that time to speculatively run hundreds of blocks searching for the best construction https://payload.de/data/
image

EIPS/eip-7581.md
benaadams: Is increasing the amount of memory that can be allocated by a single transaction which could be a DoS vector if unreasonably large or EL clients not prepared for it. Though not sure it would be an issue as isn't that large in the scheme of things

benaadams:
Note: Also think its important to see the effect of 4844 prior to sending this signal

Gemk83 (approved):

PR #8095, opened 2024-01-08
Update EIP-4844: Rename gasPrice to baseFee

MicahZoltu:
Type 2 transactions already use the "base fee" nomenclature.  This change makes 4844 transactions consistent with that naming scheme.

### DIFF
--- a/EIPS/eip-4844.md
+++ b/EIPS/eip-4844.md
@@ -48,8 +48,8 @@ Compared to full data sharding, this EIP has a reduced cap on the number of thes
 | `POINT_EVALUATION_PRECOMPILE_GAS` | `50000` |
 | `MAX_BLOB_GAS_PER_BLOCK` | `786432` |
 | `TARGET_BLOB_GAS_PER_BLOCK` | `393216` |
-| `MIN_BLOB_GASPRICE` | `1` |
-| `BLOB_GASPRICE_UPDATE_FRACTION` | `3338477` |
+| `MIN_BLOB_BASE_FEE` | `1` |
+| `BLOB_BASE_FEE_UPDATE_FRACTION` | `3338477` |
 | `GAS_PER_BLOB` | `2**17` |
 | `HASH_OPCODE_BYTE` | `Bytes1(0x49)` |
 | `HASH_OPCODE_GAS` | `3` |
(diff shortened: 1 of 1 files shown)


lightclient (commented):
I'm not necessarily against, but the data cost isn't exactly a "base" of the fee because it is exactly what the user will pay. There is no priority component of the data gas price. 

Curious if other authors have thoughts on this?

MicahZoltu:
For type 2 transactions, the "base fee" is the fee that is automatically determined and charged by the protocol and burnt.  The same is true for this fee IIUC, which is why it feels like it should share naming convention.  Just as with type 2 transactions, other fees are likely necessary to incentivize inclusion, but that feels out of scope to me for determining the name of this fee.

roberto-bayardo:
FWIW I've been using this terminology already in some of the OP-stack 4844 work as I find it more clear, e.g.: https://github.com/ethereum-optimism/docs/pull/386

lightclient:
We're talking about a header name though (or I guess in this gas, a value computed from a header value), the tx name just follows the value in the header. In 1559, there are multiple fee components, thus "base fee" name. In 4844, only a single fee component, therefore "gasprice".

In both cases, the tx name is "max_xxxx_per_gas".

MicahZoltu:
In hindsight, I think EIP-1559 should have named it `protocolFeePerGas`, which better represents what is being paid I think (the protocol enforced fee for a transaction).  Unfortunately, that ship has sailed, but if we conceptually think of `baseFee` as a synonym for `protocolFee`, then I believe the 4844 fee we are referring to is also a `protocolFee` and thus, also a `baseFee`.

g11tech:
i think constants defintely can be changed becuase they are base fee part of constants anyway
regarding the other changes, i think since the blob priority will not happen as its priority is already rolled into normal priority (and block proposers in the end don't care how tips are divided), i am fine with the name change to `base` or `protocol`

roberto-bayardo:
I guess the opcode is already using the BLOBBASEFEE terminology https://eips.ethereum.org/EIPS/eip-7516

s1na:
I want to add that internally in Geth this is called blob base fee which I agree is the more accurate name. blobGasPrice can be confusing because of the legacy tx's gas price which goes to the miner. Or we call this "burntFee", so the intention becomes clear.

MicahZoltu:
I like `blobBaseFee` to be consistent with type 2 transaction and London block header nomenclature.  In hindsight, I would have preferred `protocolFee` instead of `baseFee`, but it is too late to change type 2 transactions or block header for post-London blocks.

g11tech (approved):
the rename was accepted/not contested in Jan 18 ACD, so lgtm

lightclient (approved):

PR #8101, opened 2024-01-09
Update EIP-7212: Move to Moved

CarlBeek:
As an artifact of the new RIP process, EIP-7212 has been converted from an EIP to an RIP. As RIP-7212 has already been created and there are no EIPs that depend on EIP-7212, it makes sense to just delete here to complete the move. CC @poojaranjan @ulerdogan


ulerdogan (changes requested):
Hello Carl. Thanks for applying decided changes on EIP->RIP process. I recommend also deleting following files which are referenced by the EIP:
- ./assets/eip-7212/p256Verify_benchmark_test
- ./assets/eip-7212/ecrecover_benchmark_test

CarlBeek:
ACK, thanks @ulerdogan, `./assets/eip-7212/` removed too

ulerdogan (approved):
Thanks, I think ready to go!

RIP EIP-7212 😢🪦

poojaranjan:
Perhaps we can "move" and not "remove". 

At the time of EIP/ERC split many ERCs were moved out of EIPs repo but ERC history was preserved. Like [this](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6982.md).

CarlBeek:
Interesting, I like it. I will update to do that instead

EIPS/eip-7212.md
```diff
@@ -1,188 +1,7 @@
...
 eip: 7212
-title: Precompile for secp256r1 Curve Support
-description: Proposal to add precompiled contract that performs signature verifications in the “secp256r1” elliptic curve.
-author: Ulaş Erdoğan (@ulerdogan), Doğan Alpaslan (@doganalpaslan), DC Posch (@dcposch), Nalin Bhardwaj (@nalinbhardwaj)
-discussions-to: https://ethereum-magicians.org/t/eip-7212-precompiled-for-secp256r1-curve-support/14789
-status: Review
-type: Standards Track
-category: Core
-created: 2023-06-22
+category: EIP
```
CarlBeek, line 3: Question: should this be `category: RIP` instead?

```suggestion
category: RIP
```

ulerdogan:
We don't have `RIP` category defined on the EIP-1. Do we need a change there? I think we can keep as it is in the `Core` category as EIP itself was a 'Core EIP' similar to the moved ERCs' categories were still ERC.

poojaranjan:
@ulerdogan 

> We don't have RIP category defined on the EIP-1. Do we need a change there? 

As we progress with a proposal, an author gets more clarity on the Type/Category. Changing either or both for a proposal before it is moved to `Final` is fine

There is a general agreement on adding RIP as a Standard Track Category discussed in [EIPIP Meeting 95](https://www.youtube.com/watch?v=qgcphJxvVDs). 

We were hoping to have the PR to EIP-1 coming in by RIP working group. But, it can be made by anyone.

HoangTranMap:
OK all

Vào 7:20, Th 6, 3 thg 5, 2024 github-actions[bot] ***@***.***>
đã viết:

Scamreno (approved):

Scamreno (approved):
#8101

Scamreno (approved):
#8101 #8650 #8649 #8650 #8546

Scamreno:
#8101

Scamreno (approved):

Scamreno:
Just trying to find a way to get some of my money coming in because I haven't received anything since I started this in 2013

PR #8102, opened 2024-01-10
Add EIP: BLS signed transactions

MariusVanDerWijden:
This EIP proposes adding a new transaction type that is signed with BLS signatures


EIPS/bls-sig.md
```diff
@@ -0,0 +1,114 @@
+---
+eip: TBD
```
abcoathup, line 2:
```suggestion
eip: 7591
```

EIPS/eip-7591.md
```diff
@@ -0,0 +1,114 @@
...
+
+The block building algorithm needs to be changed in order to built the aggregated signature of all BLS signed transactions in the block. 
+All transactions in the block will be added without the signature field set. 
+
+Blocks with transactions containing the `signature` field MUST be rejected.
+
+On block verification, the `verifyAggregate` algorithm is used as follows:
+
+```
+valid = verifyAggregate(sender_1, ... sender_n, tx_hash_1, ... tx_hash_n, aggregated_sig)
```
mratsim, line 81: Note that contrary to "batch_verify" we cannot avoid signature forgeries at the cryptographic level with aggregate_verify:
- https://ethresear.ch/t/fast-verification-of-multiple-bls-signatures/5407
- https://ethresear.ch/t/security-of-bls-batch-verification/10748
- https://github.com/status-im/nim-blscurve/blob/d091a57/tests/t_batch_verifier.nim#L146-L156

So either:
- the protocol ensures that the sender_1 pubkey is correct (I think it is)
- or the verification generates "randomness", similar to fiat-shamir construct, by hashing all transactions for example. And then the pubkey and aggregate sig are adjusted to be multiplied by the 64 most significant bits of repeated hash of that randomness, see: https://github.com/mratsim/constantine/blob/f3e08ed/constantine/signatures/bls_signatures.nim#L421-L430

g11tech (approved):

EIPS/eip-7591.md
Dropkick13: []()

Dropkick13 (commented):
[]()

PR #8103, opened 2024-01-10
Add EIP: Precompile for Falcon Signature Verification

christam96:
Add a precompile for efficient verification of Falcon, a post-quantum digital signature algorithm.


EIPS/eip-8000.md
```diff
@@ -0,0 +1,76 @@
+---
+eip: 8000
```
abcoathup, line 2:
```suggestion
eip: 7592
```

christam96: EIP number and filename has been updated, thank you for the suggestion.

PR #8119, opened 2024-01-17
Update EIP-6963: fix typo

Thabokani:


glitch-txs (approved):

kdenhartog (approved):
Because this is finalized it's unlikely to get merged, but good catch still

PR #8120, opened 2024-01-18
Update EIP-6913: Move to Draft

wjmelements:
This simplification has the same properties as the belongs-to spec but is clearer and combines separate rules into a general rule.
#### Changes
The general rule is now to allow when the currently executing code equals the code of the executing account.
This encapsulates the relevant cases of `DELEGATECALL` and `CREATE`, while keeping the same security properties as the more-complex belongs-to.

### DIFF
--- a/EIPS/eip-6913.md
+++ b/EIPS/eip-6913.md
@@ -36,8 +36,8 @@ Given the upcoming deprecation of `SELFDESTRUCT`, `SETCODE` introduces a better 
 ## Specification
 
 When within a read-only execution scope like the recursive kind created by `STATICCALL`, `SETCODE` causes an exceptional abort.
-When inside of a `CREATE`-like execution scope that returns new code for the executing address (the account returned by `ADDRESS`), `SETCODE` causes an exceptional abort.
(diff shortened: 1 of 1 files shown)


RonnyGiron:
Abstract sum of all accounts, total balance of all USD currency, and crypto
currency+ ***@***.***_ string.
Point of entry>>Ethereum layer 2 (web3)-deploy smart
contracts(deposit)-post.
Owner: Ronny A. Giron_string.

RonnyGiron:
https://photos.app.goo.gl/p1mb4xqKp8iDKKNr5

Laceduppanda (approved):

gaudren (commented):
- When editing a stagnant EIP you need to move it back into its last status. (Draft)
- Your description is a little vague. You could mention that you're proposing an instruction.
- In the Gas subsection: Gselfdestruct and Gcodedeposit should be enclosed in backticks.
- Similarly Gcreate in the Rationale section should be enclosed in backticks.
- A lot of the information here is presented like a bulleted list. Although not against any rule, it makes it more difficult to read.

g11tech (approved):

wjmelements:
Thanks @g11tech

PR #8138, opened 2024-01-29
Update EIP-4762: Move to Draft

gballet:
This is a proposal not to charge the witness costs for some basic operations, that break a lot of tooling.

These witness costs that would not be charged are:
- The tx's `from` and `to` addresses

### DIFF
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -236,11 +236,13 @@ When executing a transaction, maintain four sets:
 
 When an **access** event of `(address, sub_key, leaf_key)` occurs, perform the following checks:
 
+ * If `address` is either the transaction's originator or target, skip the following steps.
  * If `(address, sub_key)` is not in `accessed_subtrees`, charge `WITNESS_BRANCH_COST` gas and add that tuple to `accessed_subtrees`.
  * If `leaf_key` is not `None` and `(address, sub_key, leaf_key)` is not in `accessed_leaves`, charge `WITNESS_CHUNK_COST` gas and add it to `accessed_leaves`
 
 When a **write** event of `(address, sub_key, leaf_key)` occurs, perform the following checks:
 
+ * If `address` is either the transaction's originator or target, skip the following steps.
  * If `(address, sub_key)` is not in `edited_subtrees`, charge `SUBTREE_EDIT_COST` gas and add that tuple to `edited_subtrees`.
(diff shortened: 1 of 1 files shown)


g11tech (dismissed):
lgtm

EIPS/eip-4762.md
```diff
@@ -236,11 +236,13 @@ When executing a transaction, maintain four sets:
 
 When an **access** event of `(address, sub_key, leaf_key)` occurs, perform the following checks:
 
+ * If `address` is either the transaction's originator or target, skip the following steps.
  * If `(address, sub_key)` is not in `accessed_subtrees`, charge `WITNESS_BRANCH_COST` gas and add that tuple to `accessed_subtrees`.
  * If `leaf_key` is not `None` and `(address, sub_key, leaf_key)` is not in `accessed_leaves`, charge `WITNESS_CHUNK_COST` gas and add it to `accessed_leaves`
 
 When a **write** event of `(address, sub_key, leaf_key)` occurs, perform the following checks:
 
+ * If `address` is either the transaction's originator or target, skip the following steps.
```
gfukushima, line 245: The way this edit was made implicates that we're not going to change CHUNK_FILL_COST in case `address` is the tx originator or target . Is that correct?
I think @gballet and @dankrad agreed to keep CHUNK_FILL_COST during the last VIC.

gballet: yeah but not for this use case, adding elements to the witness for accessing the basic fields of `to` and `from` are not going to be charged as it's assumed to be covered by the intrinsic gas. Happy to discuss this again during the next VIC if you think I'm wrong.

EIPS/eip-4762.md
```diff
@@ -236,11 +236,17 @@ When executing a transaction, maintain four sets:
 
 When an **access** event of `(address, sub_key, leaf_key)` occurs, perform the following checks:
 
+ * Perform the following steps unless `address` is either:
+   * the transaction's originator;
+   * the transactions's target;
+   * the address of the originator of a `CALL`, `STATICCALL`, `DELEGATECALL` and `CALLCODE` instruction; or,
+   * the address of the target of a `CALL`, `STATICCALL`, `DELEGATECALL` and `CALLCODE` instruction.
```
gballet, line 243:
```suggestion
```

g11tech: so any access related to originator and target won't show up? how will stateless clients run?

gballet: they are added to the witness, but no access cost is charged for them

g11tech (approved):
lgtm

PR #8154, opened 2024-01-31
Update EIP-7480: match to EOF Megaspec

pdobacz:
Sync with [megaspec](https://github.com/ipsilon/eof/blob/main/spec/eof.md)

1. update opcodes to not conflict with `EXCHANGE`
2. DATALOADN is validated against the "static data" section only
3. include rationale for not having `EXTDATACOPY` (will be linked to from the "Data-only contracts" section in EIP-3540
4. DATACOPY gas rule was missing the static 3 gas

### DIFF
--- a/EIPS/eip-7480.md
+++ b/EIPS/eip-7480.md
@@ -85,7 +85,7 @@ It is benefitial to avoid exceptional failures, because compilers can employ opt
 
 ### Lack of `EXTDATACOPY`
 
-`EXTCODECOPY` instruction is deprecated and rejected in EOF contracts and does not copy anything when being called in legacy with an EOF contract as target. A replacement instruction `EXTDATACOPY` has been considered, but decided against in order to reduce the scope of changes.
+`EXTCODECOPY` instruction is deprecated and rejected in EOF contracts and does not copy contract code when being called in legacy with an EOF contract as target. A replacement instruction `EXTDATACOPY` has been considered, but decided against in order to reduce the scope of changes.
 
 Data-only contracts which previously relied on `EXTCODECOPY` are thereby discouraged, but if there is a strong need, support for them can be easily brought back by introducing `EXTDATACOPY` in a future upgrade.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7480.md
```diff
@@ -83,6 +83,12 @@ Existing instructions for reading other kinds of data implicitly pad with zeroes
 
 It is benefitial to avoid exceptional failures, because compilers can employ optimizations like removing a code that copies data, but never accesses this copy afterwards, but such optimization is possible only if instruction never has other side effects like exceptional abort.
 
+### Lack of `EXTDATACOPY`
+
+`EXTCODECOPY` opcode is deprecated and rejected in EOF contracts and does not copy anything when being called in legacy with an EOF contract as target. An `EXTDATACOPY` replacement counterpart `EXTDATACOPY` has been considered, but decided against in order to reduce the scope of changes.
```
gumb0, line 88:
```suggestion
`EXTCODECOPY` opcode is deprecated and rejected in EOF contracts and does not copy anything when being called in legacy with an EOF contract as target. An `EXTCODECOPY` replacement counterpart `EXTDATACOPY` has been considered, but decided against in order to reduce the scope of changes.
```

pdobacz: Good catch, but I think I garbled this sentence, will rework

EIPS/eip-7480.md
```diff
@@ -83,6 +83,12 @@ Existing instructions for reading other kinds of data implicitly pad with zeroes
 
 It is benefitial to avoid exceptional failures, because compilers can employ optimizations like removing a code that copies data, but never accesses this copy afterwards, but such optimization is possible only if instruction never has other side effects like exceptional abort.
 
+### Lack of `EXTDATACOPY`
+
+`EXTCODECOPY` instruction is deprecated and rejected in EOF contracts and does not copy anything when being called in legacy with an EOF contract as target. A replacement instruction `EXTDATACOPY` has been considered, but decided against in order to reduce the scope of changes.
```
gumb0, line 88: Maybe 
```suggestion
`EXTCODECOPY` instruction is deprecated and rejected in EOF contracts and does not copy contract code when being called in legacy with an EOF contract as target. A replacement instruction `EXTDATACOPY` has been considered, but decided against in order to reduce the scope of changes.
```
because it copies fixed `EF00` bytes

pdobacz: ah, missed it, good catch

gumb0 (approved):

PR #8166, opened 2024-02-05
Update EIP-2935: Move to Draft

gballet:
This heretofore stagnant eip is now necessary for stateless clients to be able to execute the `BLOCKHASH` opcode. This PR:

 * moves the eip from a stagnant status to a draft status
 * change how the PR is activated since forks are now timestamp-based
 * adds additional history persist step on fork activation so that new retrieval logic can be transitioned to from first block itself

### DIFF
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -1,9 +1,9 @@
 ---
 eip: 2935
 title: Save historical block hashes in state
-author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak)
+author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (gballet), Gajinder Singh (@g11tech)
 discussions-to: https://ethereum-magicians.org/t/eip-2935-save-historical-block-hashes-in-state/4565
-status: Stagnant
+status: Draft
 type: Standards Track
 category: Core
 created: 2020-09-03
@@ -26,12 +26,14 @@ Additional secondary motivations include:
 
 | Parameter | Value |
 | - | - |
-| `FORK_BLKNUM` | TBD |
+| `FORK_BLKTIME` | TBD |
 | `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe`|
 
-At the start of processing any block where `block.number > FORK_BLKNUM` (ie. before processing any transactions), run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`.
+At the start of processing any block where `block.timestamp > FORK_BLKTIME` (ie. before processing any transactions), run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`.
 
-When `block.number > FORK_BLKNUM + 256`, change the logic of the `BLOCKHASH` opcode as follows: if `FORK_BLKNUM <= arg < block.number`, return `sload(HISTORY_STORAGE_ADDRESS, arg)`. Otherwise return 0.
+When `sload(HISTORY_STORAGE_ADDRESS, max(block.number - 256, 0)) != 0`, change the logic of the `BLOCKHASH` opcode as follows: return `sload(HISTORY_STORAGE_ADDRESS, arg)`.
+
+Note that this means that slot `max(block.number - 256, 0)` is accessed during block execution.
 
 ## Rationale
 
@@ -41,6 +43,11 @@ Very similar ideas were proposed before in EIP-98 and EIP-210. This EIP is a sim
 2. Writing the EIP in EVM code
 
 The former was intended to save space. Since then, however, storage usage has increased massively, to the point where even eg. 5 million new storage slots are fairly negligible compared to existing usage. The latter was intended as a first step toward "writing the Ethereum protocol in EVM" as much as possible, but this goal has since been de-facto abandoned.
+
+The touching of slot `max(block.number - 256, 0)` during block execution is necessary so that:
+
+ * detecting the activation of a fork now that forks are timestamp-activated; and
+ * stateless clients can determine if the fork has happened, and if the new behavior can be activated.
 
 ## Backwards Compatibility
 
@@ -52,7 +59,8 @@ TBD
 
 ## Implementation
 
-TBD
+ * PR 28878 of go-ethereum
+ * Active on verkle-gen-devet-3 for its verkle implementation
 
 ## Security Considerations
 
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -10,7 +10,7 @@ category: Core
 created: 2020-09-03
 ---
 
-## Simple Summary
+## Abstract
 
 Store historical block hashes in a contract, and modify the `BLOCKHASH (0x40)` opcode to read this contract.
 
@@ -63,7 +63,7 @@ The range of `BLOCKHASH` is increased by this opcode, but behavior within the pr
 
 TBD
 
-## Implementation
+## Reference Implementation
 
  * PR 28878 of go-ethereum
  * Active on verkle-gen-devet-3 for its verkle implementation
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -40,7 +40,7 @@ At the start of processing any block where `block.timestamp > FORK_TIMESTAMP` (i
 
 Edge cases:
 
- * For verkle at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
+ * For the fork to be activated at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
  * for verkle activated at block `1`, only genesis hash will be written at slot `0` as no additional history needs to be persisted.
  * for verkle activated at block `32`, block `31`'s hash will be written to slot `31` and additonal history for `0..30`'s hashes will be persisted, so all in all `0..31`'s hashes.
 
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -1,7 +1,7 @@
 ---
 eip: 2935
 title: Save historical block hashes in state
-author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech)
+author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria)
 description: store previous block hashes as storage slots of a system contract to allow for stateless execution
 discussions-to: https://ethereum-magicians.org/t/eip-2935-save-historical-block-hashes-in-state/4565
 status: Draft
@@ -27,14 +27,19 @@ Additional secondary motivations include:
 
 | Parameter | Value |
 | - | - |
-| `FORK_BLKTIME` | TBD |
+| `FORK_TIMESTAMP` | TBD |
 | `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe`|
+| `HISTORY_SERVE_WINDOW` | `256` |
 
-At the start of processing any block where `block.timestamp > FORK_BLKTIME` (ie. before processing any transactions), run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`.
+At the start of processing any block where `block.timestamp > FORK_TIMESTAMP` (ie. before processing any transactions), 
+* run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`. 
+* **additionally** if this is the first block of the fork, then store last `min(HISTORY_SERVE_WINDOW-1, block.number-1)` **blocks' prevhash** into their corresponding slots i.e persist the entire `HISTORY_SERVE_WINDOW` block hashes (up until the genesis) that can be accessed in the block.
+* change the logic of the `BLOCKHASH` opcode as follows: return `sload(HISTORY_STORAGE_ADDRESS, arg)`.
 
-When `sload(HISTORY_STORAGE_ADDRESS, max(block.number - 256, 0)) != 0`, change the logic of the `BLOCKHASH` opcode as follows: return `sload(HISTORY_STORAGE_ADDRESS, arg)`.
-
-Note that this means that slot `max(block.number - 256, 0)` is accessed during block execution.
+Edge cases:
+* For verkle at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
+* for verkle activated at block `1`, only genesis hash will be written at slot `0` as no additional history needs to be persisted.
+* for verkle activated at block `32`, block `31`'s hash will be written to slot `31` and additonal history for `0..30`'s hashes will be persisted, so all in all `0..31`'s hashes.
 
 ## Rationale
 
@@ -45,10 +50,7 @@ Very similar ideas were proposed before in EIP-98 and EIP-210. This EIP is a sim
 
 The former was intended to save space. Since then, however, storage usage has increased massively, to the point where even eg. 5 million new storage slots are fairly negligible compared to existing usage. The latter was intended as a first step toward "writing the Ethereum protocol in EVM" as much as possible, but this goal has since been de-facto abandoned.
 
-The touching of slot `max(block.number - 256, 0)` during block execution is necessary so that:
-
- * detecting the activation of a fork now that forks are timestamp-activated; and
- * stateless clients can determine if the fork has happened, and if the new behavior can be activated.
+Storing of all last `HISTORY_SERVE_WINDOW` block hashes alleviates the need to detect fork activation height to transition to the new logic as the entire required history will be available from the first block of the fork itself. The cost of doing so is marginal considering the `HISTORY_SERVE_WINDOW` being small.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -1,8 +1,8 @@
 ---
 eip: 2935
-title: Save historical block hashes in state
-description: store previous block hashes as storage slots of a system contract to allow for stateless execution
-author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria)
+title: Serve historical block hashes from state
+description: Store and serve last 8192 block hashes as storage slots of a system contract to allow for stateless execution
+author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria), Ignacio Hagopian (@jsign), Jochem Brouwer (@jochem-brouwer)
 discussions-to: https://ethereum-magicians.org/t/eip-2935-save-historical-block-hashes-in-state/4565
 status: Draft
 type: Standards Track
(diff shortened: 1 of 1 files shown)


EIPS/eip-2935.md
```diff
@@ -1,9 +1,9 @@
 ---
 eip: 2935
 title: Save historical block hashes in state
-author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak)
+author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (gballet), Gajinder Singh (@g11tech)
```
gballet, line 4:
```suggestion
author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech)
```

EIPS/eip-2935.md
```diff
@@ -1,9 +1,10 @@
 ---
 eip: 2935
 title: Save historical block hashes in state
-author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak)
+author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria)
+description: store previous block hashes as storage slots of a system contract to allow for stateless execution
```
gballet, line 5:
```suggestion
description: store previous block hashes as storage slots of a system contract to allow for stateless execution
author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria)
```

EIPS/eip-2935.md
```diff
@@ -26,12 +27,19 @@ Additional secondary motivations include:
...
+* run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`. 
+* **additionally** if this is the first block of the fork, then store last `min(HISTORY_SERVE_WINDOW-1, block.number-1)` **blocks' prevhash** into their corresponding slots i.e persist the entire `HISTORY_SERVE_WINDOW` block hashes (up until the genesis) that can be accessed in the block.
+* change the logic of the `BLOCKHASH` opcode as follows: return `sload(HISTORY_STORAGE_ADDRESS, arg)`.
 
-When `block.number > FORK_BLKNUM + 256`, change the logic of the `BLOCKHASH` opcode as follows: if `FORK_BLKNUM <= arg < block.number`, return `sload(HISTORY_STORAGE_ADDRESS, arg)`. Otherwise return 0.
+Edge cases:
+* For verkle at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
+* for verkle activated at block `1`, only genesis hash will be written at slot `0` as no additional history needs to be persisted.
+* for verkle activated at block `32`, block `31`'s hash will be written to slot `31` and additonal history for `0..30`'s hashes will be persisted, so all in all `0..31`'s hashes.
```
gballet, line 42:
```suggestion
Edge cases:

 * For verkle at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
 * for verkle activated at block `1`, only genesis hash will be written at slot `0` as no additional history needs to be persisted.
 * for verkle activated at block `32`, block `31`'s hash will be written to slot `31` and additonal history for `0..30`'s hashes will be persisted, so all in all `0..31`'s hashes.
```

g11tech (approved):
lgtm pending previous authors' approval

EIPS/eip-2935.md
```diff
@@ -26,22 +27,34 @@ Additional secondary motivations include:
...
+| `FORK_TIMESTAMP` | TBD |
 | `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe`|
+| `HISTORY_SERVE_WINDOW` | `256` |
 
-At the start of processing any block where `block.number > FORK_BLKNUM` (ie. before processing any transactions), run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`.
+At the start of processing any block where `block.timestamp > FORK_TIMESTAMP` (ie. before processing any transactions):
 
-When `block.number > FORK_BLKNUM + 256`, change the logic of the `BLOCKHASH` opcode as follows: if `FORK_BLKNUM <= arg < block.number`, return `sload(HISTORY_STORAGE_ADDRESS, arg)`. Otherwise return 0.
+ * run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`. 
+ * **additionally** if this is the first block of the fork, then store last `min(HISTORY_SERVE_WINDOW-1, block.number-1)` **blocks' prevhash** into their corresponding slots i.e persist the entire `HISTORY_SERVE_WINDOW` block hashes (up until the genesis) that can be accessed in the block.
```
jochem-brouwer, line 37:
```suggestion
 * **additionally** if this is the first block of the fork, then store last `min(HISTORY_SERVE_WINDOW, block.number)` **blocks' prevhash** into their corresponding slots i.e persist the entire `HISTORY_SERVE_WINDOW` block hashes (up until the genesis) that can be accessed in the block.
```

I think there is an off-by-one error here. If this EIP gets activated at block 0, then at block 1 we get `min(255, 0)` (thus 0) and therefore the "last 0 **blocks' prevhash**" get written to the contract. From other changes to the EIP I get the impression that we want the genesis hash to be written as well.

In general I get the idea that the idea of the BLOCKHASH opcode does not change during the 256 block-window of the fork happening, therefore I think we have this off-by-one-error.

In case of the fork happening at block 256: we would still have access to the `BLOCKHASH` of the blocks 0-255 (so 256 blocks in total), so we need to write the latest 256 blocks (so 0-255) to the contract, not the latest 255 blocks (so 1-255) (`min(256-1, 256-1) = min(255,255) = 255)` per original).

EIPS/eip-2935.md
```diff
@@ -26,22 +27,34 @@ Additional secondary motivations include:
...
+ * change the logic of the `BLOCKHASH` opcode as follows: return `sload(HISTORY_STORAGE_ADDRESS, arg)`.
+ 
+
+Edge cases:
+
+ * For verkle at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
```
jochem-brouwer, line 43: Nit: this EIP is not specific to Verkle, so
```suggestion
 * For the fork to be activated at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
```

(also for the other examples)

EIPS/eip-2935.md
```diff
@@ -26,22 +27,34 @@ Additional secondary motivations include:
...
-| `FORK_BLKNUM` | TBD |
+| `FORK_TIMESTAMP` | TBD |
 | `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe`|
+| `HISTORY_SERVE_WINDOW` | `256` |
 
-At the start of processing any block where `block.number > FORK_BLKNUM` (ie. before processing any transactions), run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`.
+At the start of processing any block where `block.timestamp > FORK_TIMESTAMP` (ie. before processing any transactions):
 
-When `block.number > FORK_BLKNUM + 256`, change the logic of the `BLOCKHASH` opcode as follows: if `FORK_BLKNUM <= arg < block.number`, return `sload(HISTORY_STORAGE_ADDRESS, arg)`. Otherwise return 0.
+ * run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`.
```
jochem-brouwer, line 36: Nit: there is currently no note on how to avoid EIP-158 state clearing. If an account has zero balance, zero nonce, no code, but storage, it is still considered dead, so any touch-action on this contract will delete the entire state. (So either pre-fund this account, or bump the nonce to one, this should be safe since no one should be able to deploy a contract here / have the keys of this contract)

jochem-brouwer (commented):
Great updates, I left some comments

EIPS/eip-2935.md
g11tech: this step is previous to additional step, so block 1 writes its prevhash i.e. 0 => genesis hash, and then no additional things get written

EIPS/eip-2935.md
```diff
@@ -26,22 +27,34 @@ Additional secondary motivations include:
...
-When `block.number > FORK_BLKNUM + 256`, change the logic of the `BLOCKHASH` opcode as follows: if `FORK_BLKNUM <= arg < block.number`, return `sload(HISTORY_STORAGE_ADDRESS, arg)`. Otherwise return 0.
+ * run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`. 
+ * **additionally** if this is the first block of the fork, then store last `min(HISTORY_SERVE_WINDOW-1, block.number-1)` **blocks' prevhash** into their corresponding slots i.e persist the entire `HISTORY_SERVE_WINDOW` block hashes (up until the genesis) that can be accessed in the block.
+ * change the logic of the `BLOCKHASH` opcode as follows: return `sload(HISTORY_STORAGE_ADDRESS, arg)`.
+ 
+
+Edge cases:
+
+ * For the fork to be activated at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
+ * for verkle activated at block `1`, only genesis hash will be written at slot `0` as no additional history needs to be persisted.
```
g11tech, line 44:
```suggestion
 * for activation at block `1`, only genesis hash will be written at slot `0` as no additional history needs to be persisted.
```

EIPS/eip-2935.md
```diff
@@ -26,22 +27,34 @@ Additional secondary motivations include:
...
+ * for verkle activated at block `32`, block `31`'s hash will be written to slot `31` and additonal history for `0..30`'s hashes will be persisted, so all in all `0..31`'s hashes.
```
g11tech, line 45:
```suggestion
 * for activation at block `32`, block `31`'s hash will be written to slot `31` and additonal history for `0..30`'s hashes will be persisted, so all in all `0..31`'s hashes.
```

EIPS/eip-2935.md
jochem-brouwer: Yup, I missed the `additional` point

gballet: it's correct, because in the general case, you have to add 255 blocks after adding the one just before (notice the **additionally** in bold).

In the case of genesis, you also get `min(255, 0) == 0` so no extra block get inserted.

I think we should switch to a pseudo-python implementation, because you're the second person to be confused by this (I was the first :D)

gballet: so something like:

```python3
def process_block_hash_history(block :Block, state: State):
    if block.timestamp >= FORK_TIMESTAMP:
        state.insert_slot(HISTORY_STORAGE_ADDRESS, block.number-1, block.parent.hash)

    # If this is the first fork block, add the parent's direct 255 ancestors as well
    if block.parent.timestamp < FORK_TIMESTAMP:
        ancestor = block.parent
        for i in range(255):
            ancestor = ancestor.parent
            state.insert_slot(HISTORY_STORAGE_ADDRESS, ancestor.number, ancestor.hash)
```

EIPS/eip-2935.md
```diff
@@ -26,22 +27,56 @@ Additional secondary motivations include:
 
 | Parameter | Value |
 | - | - |
-| `FORK_BLKNUM` | TBD |
+| `FORK_TIMESTAMP` | TBD |
 | `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe`|
+| `HISTORY_SERVE_WINDOW` | `256` |
 
-At the start of processing any block where `block.number > FORK_BLKNUM` (ie. before processing any transactions), run `sstore(HISTORY_STORAGE_ADDRESS, block.number - 1, block.prevhash)`.
+At the start of processing any block where `block.timestamp > FORK_TIMESTAMP` (ie. before processing any transactions), update the history in the following way:
```
jochem-brouwer, line 34:
```suggestion
At the start of processing any block where `block.timestamp >= FORK_TIMESTAMP` (ie. before processing any transactions), update the history in the following way:
```

Current text does not match pseudocode

EIPS/eip-2935.md
```diff
@@ -26,22 +27,56 @@ Additional secondary motivations include:
...
+        for i in range(HISTORY_SERVE_WINDOW - 1):
+            # stop at genesis block
+            if ancestor.number == 0:
+                break
+
+            ancestor = ancestor.parent
+            state.insert_slot(HISTORY_STORAGE_ADDRESS, ancestor.number, ancestor.hash)
+```
+
+Note that if this is the fork block, then it persists the additional requisite history that could be needed while resolving `BLOCKHASH` opcode for all of the `HISTORY_SERVE_WINDOW` ancestors (up untill genesis).
```
jochem-brouwer, line 53:
```suggestion
Note that if this is the fork block, then it persists the additional requisite history that could be needed while resolving `BLOCKHASH` opcode for all of the `HISTORY_SERVE_WINDOW` ancestors (up until genesis).
```

EIPS/eip-2935.md
```diff
@@ -26,22 +27,56 @@ Additional secondary motivations include:
...
+
+For resolving the `BLOCKHASH` opcode this fork onwards (`block.timestamp > FORK_TIMESTAMP`), switch the logic to:
```
jochem-brouwer, line 55:
```suggestion
For resolving the `BLOCKHASH` opcode this fork onwards (`block.timestamp >= FORK_TIMESTAMP`), switch the logic to:
```

EIPS/eip-2935.md
```diff
@@ -26,22 +27,56 @@ Additional secondary motivations include:
...
+
+```python
+def resolve_blockhash(block: Block, state: State, arg: uint64):
+  assert(arg < block.number)
```
jochem-brouwer, line 59: This does not match current `BLOCKHASH`. If `arg >= block.number` then return 0 (push 0 to stack)

Our code:

```typescript
      const number = runState.stack.pop()

      const diff = runState.interpreter.getBlockNumber() - number
      // block lookups must be within the past 256 blocks
      if (diff > BIGINT_256 || diff <= BIGINT_0) {
        runState.stack.push(BIGINT_0)
        return
      }
```

jochem-brouwer (changes requested):

EIPS/eip-2935.md
gballet: yeah but that's precisely why we want to use the contract: be able to get further history than 256 blocks.

jochem-brouwer: Yes, the point is about when you request the current block number hash, or a block number higher, I would assume that `assert` throws? (And thus push nothing to stack and exit call frame?)

EIPS/eip-2935.md
```diff
@@ -26,22 +27,65 @@ Additional secondary motivations include:
...
+ * for activation at block `32`, block `31`'s hash will be written to slot `31` and additonal history for `0..30`'s hashes will be persisted, so all in all `0..31`'s hashes.
+
+### EIP-158 exception
+
+This address is currently exempt from [EIP-158](./eip-158.md) cleanup in Kaustinen Verkle Testnet but there are two ways this could be addressed before this EIP is adopted by ACD:
+
+* Update the nonce to 1 in the fork block, or
+* Deploy a contract à la [EIP-4788](./eip-4788.md) with `BLOCKHASH` opcode delegating call to this contract with appropriate args.
+
+While the second option looks more elegant, but it has higher complexity as well as gas consumption considerations.
```
jochem-brouwer, line 76:
```suggestion
While the second option looks more elegant, it has a higher complexity as well as gas consumption considerations.
```

EIPS/eip-2935.md
g11tech: applied the feedback

EIPS/eip-2935.md
```diff
@@ -26,37 +27,88 @@ Additional secondary motivations include:
...
+```
+
+Note that if this is the fork block, then it persists the additional requisite history that could be needed while resolving `BLOCKHASH` opcode for all of the `MIN_HISTORY_SERVE_WINDOW` ancestors (up until genesis).
+
+For resolving the `BLOCKHASH` opcode this fork onwards (`block.timestamp >= FORK_TIMESTAMP`), switch the logic to:
+
+```python
+def resolve_blockhash(block: Block, state: State, arg: uint64):
+  # check the wrap around range
+  if arg >= block.number or arg < max(block.number - MAX_HISTORY_SERVE_WINDOW, 0)
```
jochem-brouwer, line 61: In EVM, the requested `block.number` cannot be negative or be higher than 2**256, so just the check `arg >= block.number` (assuming that `block` is the current block we are running on) is sufficient here.

jochem-brouwer: `arg` should thus also be of type `uint256`

g11tech: yes we need to discuss the wrap around thing and arg sizes

EIPS/eip-2935.md
```diff
@@ -1,62 +1,122 @@
...
-* Improving cleanness of the protocol, as the BLOCKHASH opcode would then access state and not history.
+A side benefit of this approach could be that it allows building/validating proofs related to last 256 ancestors directly against the current state.
 
 ## Specification
 
 | Parameter | Value |
 | - | - |
-| `FORK_BLKNUM` | TBD |
+| `FORK_TIMESTAMP` | TBD |
 | `HISTORY_STORAGE_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe`|
```
s1na, line 30: This conflicts with the system address from EIP-4788.

EIPS/eip-2935.md
```diff
@@ -1,62 +1,122 @@
...
+    if block.timestamp >= FORK_TIMESTAMP:
+        state.insert_slot(HISTORY_STORAGE_ADDRESS, (block.number-1) % HISTORY_SERVE_WINDOW , block.parent.hash)
+
+    # If this is the first fork block, add the parent's direct 255 ancestors as well
+    if block.parent.timestamp < FORK_TIMESTAMP:
+        ancestor = block.parent
+        for i in range(HISTORY_SERVE_WINDOW - 1):
+            # stop at genesis block
+            if ancestor.number == 0:
+                break
```
gballet, line 49: @g11tech @jochem-brouwer this seems incorrect, as if the fork happens at block 1, the genesis will not be stored, either by this method, or by the regular block operation. I think this check should be moved AFTER `state.insert_slot`

jochem-brouwer: Why? If fork block is 1, then 

```
    if block.timestamp >= FORK_TIMESTAMP:
        state.insert_slot(HISTORY_STORAGE_ADDRESS, (block.number-1) % HISTORY_SERVE_WINDOW , block.parent.hash)
```

This will be ran (so genesis hash is stored).

EIPS/eip-2935.md
```diff
@@ -1,62 +1,122 @@
...
+```python
+def process_block_hash_history(block: Block, state: State):
+    if block.timestamp >= FORK_TIMESTAMP:
+        state.insert_slot(HISTORY_STORAGE_ADDRESS, (block.number-1) % HISTORY_SERVE_WINDOW , block.parent.hash)
+
+    # If this is the first fork block, add the parent's direct 255 ancestors as well
+    if block.parent.timestamp < FORK_TIMESTAMP:
+        ancestor = block.parent
+        for i in range(HISTORY_SERVE_WINDOW - 1):
+            # stop at genesis block
```
gballet, line 47: The wording here is confusing, I suggest the following:

```suggestion
            #  if ancestor.number == 0, this means we inserted the
            # genesis block hash in the previous iteration. Stop here.
```

EIPS/eip-2935.md
g11tech: i think L51 makes one hop to the parent,so at number=1, L51 will hop to genesis block and save it, may be we can simplify it

EIPS/eip-2935.md
```diff
@@ -1,62 +1,122 @@
...
+            if ancestor.number == 0:
+                break
+
+            ancestor = ancestor.parent
+            state.insert_slot(HISTORY_STORAGE_ADDRESS, ancestor.number % HISTORY_SERVE_WINDOW, ancestor.hash)
```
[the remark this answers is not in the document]
g11tech: this makes it look more analogous to L41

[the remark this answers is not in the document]
gballet: actually, I think we should just rewrite it more clearly:

```python
for i in range(block.number-1, min(block.number-HISTORY_SERVE_WINDOW, 0)):
    state.insert_slot(HISTORY_STORAGE_ADDRESS, i % HISTORY_SERVE_WINDOW, ancestor.parent.hash)
    ancestor = ancestor.parent
```

then, the way it's implemented is left to the implementer.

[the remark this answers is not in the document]
g11tech: i think this is correct  storing at (`i-1`) than `i`, since 'i` is ancestor

or we shouldn't use i, just `ancestor.number -1` since that relationship is more clear (we always store ancestor's parent)

```python
    state.insert_slot(HISTORY_STORAGE_ADDRESS, (ancestor.number -1) % HISTORY_SERVE_WINDOW, ancestor.parent.hash)
```

lightclient:
@g11tech @gballet how are the block hashes written to this contract? The assembly I see has no `sstore`. I think it should follow the 4788 pattern where it is spec'd as the system makes a call as `0xff..ff` and the update logic is gated on that.

tkstanczak (approved):

lightclient (approved):

gballet:
> @g11tech @gballet how are the block hashes written to this contract? The assembly I see has no `sstore`. I think it should follow the 4788 pattern where it is spec'd as the system makes a call as `0xff..ff` and the update logic is gated on that.

We already addressed this question during ACD: if anything, the 4788 system should be changed as it unnecessarily adds code chunks to the witness. We do not want to do that, and so the 2935 contract retains its original mode of operation

PR #8172, opened 2024-02-07
Update EIP-7568: Move to Last Call

timbeiko:


eip-automerger:

 - Trying to change EIP 7568 state from Review to Last Call

PR #8214, opened 2024-02-13
Update EIP-4844: simplify `calc_excess_blob_gas`

bkellerman:
Less code and easier to understand from the view of constraining `excess_blob_gas`. `excess_blob_gas` is a summation of  `parent.blob_gas_used - TARGET_BLOB_GAS_PER_BLOCK` that must always be non-negative.


lightclient:
Hi @bkellerman thanks for this PR. While I do see how this could be considered a bit simpler, I'm not sure it's enough to warrant the change at this point.

PR #8228, opened 2024-02-16
Update EIP-7002: convert to system contract

lightclient:
First stab at converting 7002 to a system contract. A few notable changes:

* Instead of directly reading storage to compute the exits in the block, this has be combined in the system call to the contract. The exits are returned as normal return data. They should be packed in 68 byte chunks.
* The contract now allows for callers to read the current `excess_exits` count by calling the contract with no calldata.
* The contract does not return an over payment of the fee. Instead, callers should use the aforementioned getter to compute the exact fee before calling the contract.

The bytecode is created from the following repo: https://github.com/lightclient/7002asm

### DIFF
--- a/EIPS/eip-7002.md
+++ b/EIPS/eip-7002.md
@@ -38,7 +38,7 @@ Note, 0x00 withdrawal credentials can be changed into 0x01 withdrawal credential
 
 | Name | Value | Comment |
 | - | - | - |
-| `VALIDATOR_EXIT_ADDRESS` | `0x229F8EdAF3c9C852E29034AF9EA3ce16b50AE017` | Where to call and store relevant details about exit mechanism |
+| `VALIDATOR_EXIT_ADDRESS` | `0x0f1ee3e66777F27a7703400644C6fCE41527E017` | Where to call and store relevant details about exit mechanism |
 | `SYSTEM_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe` | Address used to invoke system operation on contract
 | `EXCESS_EXITS_STORAGE_SLOT` | 0 | |
 | `EXIT_COUNT_STORAGE_SLOT` | 1 | |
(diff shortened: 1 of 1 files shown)


EIPS/eip-7002.md
```diff
@@ -147,137 +138,402 @@ def fake_exponential(factor: int, numerator: int, denominator: int) -> int:
...
+dup3
+add
+gt
+push2 0x014
+jumpi
+
+pop
+pop
+push0
+push2 0x014
```
mkalinin, line 460: This offset seems wrong. If we assume the offset in line#454 (`0x0140`) is correct, then this one should be `0x0140 + 0x6`

mkalinin, line 460:
```suggestion
push2 0x0146
```

EIPS/eip-7002.md
```diff
@@ -147,137 +138,402 @@ def fake_exponential(factor: int, numerator: int, denominator: int) -> int:
...
+push1 0xa0 
+shr
+or
+dup3
+push1 0x44
+mul
+push1 0x20
+add
+mstore
+push20 0xff
```
mkalinin, line 404:
```suggestion
push20 0xffffffff00000000000000000000000000000000
```

mkalinin:
`EXCESS_RETURN_GAS_STIPEND` should be removed as well as no longer used

lightclient:
@mkalinin updated based on your review, PTAL.

mkalinin (approved):

mkalinin:
Awesome work!!

loga4:
Based on the recent changes to the EIP-7002, 2 questions here:

1. The ability to return unspent gas to the sender has been removed, but the Utilizing CALL block remains https://eips.ethereum.org/EIPS/eip-7002#utilizing-call-to-return-excess-payment. Is this option supposed to appear in the future or did you forget to remove it from the document?

2. Is it correct - there is no restriction on incoming queue, meaning I can send a exit request of 100/1000 validators for the same price?

(there are no open PRs about 7002, so I tried to find an relevant place to ask these questions)

lightclient:
You can in general direct questions to https://ethereum-magicians.org/t/eip-7002-execution-layer-triggerable-exits/14195

1) This is an omission, thank you for letting us know. It shall be removed.
2) The fee grows exponentially but it updates each block. You can send as many exits as you like for the fee of a certain block. However, the dequeue process only reads a fixed number per block.

PR #8190, opened 2024-02-09
Add EIP: Precompile Falcon512 generic verifier

eum602:
Propose to add a **generic** Postquantum signature verification algorithm based on Falcon-512.

PR #8245, opened 2024-02-19
Update EIP-1: add external links for post-quantum algorithm implementations

eum602:
Hello, this PR proposes to add new PQClean implementations. The intention of this PR is to allow such references in new EIP proposals such as [Precompile Falcon512 generic verifier](https://github.com/ethereum/EIPs/pull/8190)

### DIFF
--- a/EIPS/eip-1.md
+++ b/EIPS/eip-1.md
@@ -356,6 +356,22 @@ Permitted Yellow Paper URLs must anchor to a specific commit, and so must match 
 ^(https://github\.com/ethereum/yellowpaper/blob/[0-9a-f]{40}/paper\.pdf)$
 ```
 
+### PQClean
+
+Links to PQClean (clean implementations of the post-quantum schemes that are in the NIST post-quantum project) may be included using normal markdown syntax, such as:
+
+```markdown
+[PQClean-Falcon-512](https://github.com/PQClean/PQClean/commit/8e220a87308154d48fdfac40abbb191ac7fce06a)
+```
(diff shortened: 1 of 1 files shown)


SamWilsn:
I've created a "Call for Input" to collect input from EIP Editors: https://github.com/ethcatherders/EIPIP/issues/318

SamWilsn:
It might be easier if you can provide a source with a DOI number (eg. [`10.1007/3-540-48071-4_1`](https://doi.org/10.1007/3-540-48071-4_1)). We allow any links that have DOIs without any special exceptions.

eum602:
> It might be easier if you can provide a source with a DOI number (eg. [`10.1007/3-540-48071-4_1`](https://doi.org/10.1007/3-540-48071-4_1)). We allow any links that have DOIs without any special exceptions.

Thanks for pointing it out @SamWilsn .  I couldn't find a DOI for Falcon; since all information was originally published in the author's page  (https://falcon-sign.info/)  which is the main page that implementers refer to. There the key differentiator is the specification version (20211101) . For example that number is used in the [OpenQuantumSafe project](https://openquantumsafe.org/liboqs/algorithms/sig/falcon) to point the specification version they use. Just leaving this information to hopefully find a way to address the observation.

SamWilsn:
Sorry for the runaround on this @eum602; thanks a ton for putting up with us!

You mentioned that you're going to copy the implementation into your assets directory instead, so I'm going to close this PR.

PR #8266, opened 2024-02-28
Add EIP: Cease serving history before PoS

lightclient:
This EIP proposes that execution layer clients stop server historical data from before the merge over p2p.


EIPS/7639.md
```diff
@@ -0,0 +1,81 @@
+---
+eip: 7639
```
abcoathup, line 2: @lightclient need to rename the file to eip-7639.md

EIPS/7639.md
```diff
@@ -0,0 +1,81 @@
...
+As of 2024, historical data in clients has grown to around 500 GB. Nearly 400 GB
+of that is from block data before PoS was activated in the Paris upgrade. Long
+term, Ethereum plans to bound the amount of data nodes must store. This EIP
+proposes the first steps to achieve such goal.
+
+## Specification
+
+Clients must not make or respond to p2p queries about blocks before block 15537393.
+
+### Header Accumulator
```
smartprogrammer93, line 29: This feels out of context. even though i know what this is, it does not seem connected to the rest of the EIP. maybe we need to add some explaination about how the header accumulator will be used. and how users will be able to get this data using era1 one files?

EIPS/7639.md
```diff
@@ -0,0 +1,81 @@
...
+eip: 7639
+title: Cease serving history before PoS
+description: Execution layer clients will no longer serve block data before Paris over p2p.
+author: lightclient (@lightclient)
+discussions-to: https://ethereum-magicians.org/t/cease-serving-history-before-pos/18991
+status: Draft
+type: Standards Track
+category: Networking
+created: 2024-02-13
+---
```
smartprogrammer93, line 11: Maybe add dependency on eip-6110?

lightclient: I don't think this is dependent on 6110? This only removes pre-merge data. 6110 only modifies the behavior of exits after Prague, so clients still need to sort out how to deal with the loss of pre-merge receipts.

EIPS/7639.md
lightclient: fixed, thx

EIPS/7639.md
lightclient: I think you're right. I have moved it to a separate EIP: https://github.com/ethereum/EIPs/pull/8277

lightclient:
mr. eip editor, my eip is ready

erayack:
Do you see this as a step up to EIP-4444, or will it replace it?

lightclient:
Hi @erayack this would be step up to full 4444.

EIPS/eip-7639.md
```diff
@@ -0,0 +1,55 @@
...
+instead focuses on the shape of the data.
+
+## Backwards Compatibility
+
+After this EIP is activated, nodes will no longer be able to full sync from the
+devp2p network. To continue doing so, they must retrieve the data out-of-band.
+
+## Security Considerations
+
+TODO
```
SamWilsn, line 51:
```suggestion
<!-- TODO -->
```

SamWilsn (approved):

PR #8271, opened 2024-02-29
Add EIP: eth/69 - Drop pre-merge fields

MariusVanDerWijden:
Draft of eth70
Todo:
- check if we should do 69 or 70
- Any additional things we can remove?


jochem-brouwer:
Quick reminder that here `eth/68` was updated: https://github.com/ethereum/devp2p/commit/6b259a7003b4bfb18365ba690f4b00ba8a26393b 

On this issue it was discussed:

https://github.com/ethereum/devp2p/issues/245

See: https://github.com/ethereum/devp2p/issues/245#issuecomment-1668118158

So maybe we want to do `eth/69` first and then do `eth/70` here?

MariusVanDerWijden:
I've renamed the proposal to `eth/69` now after discussing with @smartprogrammer93 as this will likely go in before the proposal that drops history. Thanks again!

MariusVanDerWijden:
@jochem-brouwer Do you think we should change the protocol to use the previously proposed mechanism? I would prefer leaving it as is and enshrining the "bug"

EIPS/eip-7642.md
```diff
@@ -0,0 +1,105 @@
...
+description: Drop unnecessary fields after the merge
+author: Marius van der Wijden (@MariusVanDerWijden)
+discussions-to: https://ethereum-magicians.org/t/eth-70-drop-pre-merge-fields-from-eth-protocol/19005
+status: Draft
+type: Standards Track
+category: Networking
+created: 2024-02-29
+requires: 5793
+---
+
```
lightclient, line 13:
```suggestion
```

MariusVanDerWijden: addressed in https://github.com/ethereum/EIPs/pull/8271/commits/e719980ae2d17c3c847de14b542b0ad5e4fe7729 thanks!

smartprogrammer93 (approved):

EIPS/eip-7642.md
lightclient: have a blessed day!

jochem-brouwer:
> @jochem-brouwer Do you think we should change the protocol to use the previously proposed mechanism? I would prefer leaving it as is and enshrining the "bug"

I have no strong opinion here, I think just leaving it in as bug is fine :)

MariusVanDerWijden:
I did some more benchmarks and determined the floor for bandwidth improvements to be at ~95GiB. 
I converted the receipts of a node from non-bloom to with bloom (and compressed them with snappy) and the disk usage for receipts goes up from 179.07GiB to 273.28 GiB, so up by 94.21 GiB. The theoretical increase should be 530GB, so the snappy compression is quite good. We can not expect the same level of compression on the wire though, but this gives us a good floor for this improvement in terms of bandwidth

EIPS/eip-7642.md
```diff
@@ -0,0 +1,104 @@
...
+## Abstract
+
+After the merge a few fields (`td`) and messages (`NewBlockHashes`, `NewBlock`) in the networking protocol became obsolete.
+This EIP modifies the networking messages such that these fields are not sent anymore.
+Additionally we propose to remove the `Bloom` field from the receipts networking messages.
+
+## Motivation
+
+We recently discovered that none of the clients store the `Bloom` field of the receipts as it can be recomputed on demand.
+However the networking spec requires the `Bloom` field to be send over the network.
```
holiman, line 23:
```suggestion
However the networking spec requires the `Bloom` field to be sent over the network.
```

MariusVanDerWijden: @ mods theres someone typo-farming on my PR

lightclient: they have been removed from the face of the earth

lightclient (approved):

PR #8274, opened 2024-02-29
Update EIP-2537: Editorial

khovratovich:
Edited a few sentences that were ambiguous. 

Added the Fp, Fp2, G1, G2 definitions.


EIPS/eip-2537.md
```diff
@@ -67,10 +69,10 @@ Twist type: M
 B coefficient for twist c0 = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004
 B coefficient for twist c1 = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004
 Generators:
-G1:
+H1:
```
asanso, line 72: why H and not G?

khovratovich: Just to resolve ambiguity. Because G1 is used for a group of points.

EIPS/eip-2537.md
```diff
@@ -80,33 +82,43 @@ Pairing parameters:
 x is negative = true
 ```
 
-One should note that the base field modulus is equal to `3 mod 4` that allows an efficient square root extraction, although as described below, the gas cost of decompression is larger than the gas cost of supplying decompressed point data in `calldata`.
+One should note that base field modulus $p$ is equal to `3 mod 4` that allows an efficient square root extraction, although as described below gas cost of decompression is larger than gas cost of supplying decompressed point data in `calldata`.
```
ralexstokes, line 85:
```suggestion
One should note that base field modulus `p` is equal to `3 mod 4` that allows an efficient square root extraction, although as described below gas cost of decompression is larger than gas cost of supplying decompressed point data in `calldata`.
```

ralexstokes (commented):
generally looks good

shamatar:
If clarification are in the process, then we should double check (as I no longer remember if it was well covered in the original documents, it was too long ago):
- subgroup checks for non-pairing operations: I think they should be on the caller's responsibility, so implementation should only do "on curve" check
- number of input-pairs into pairing operation and multiexp operation: empty input must return an error and not zero-point/identity

I will check a the original document, but second confirmation will be great

khovratovich:
> * subgroup checks for non-pairing operations: I think they should be on the caller's responsibility, so implementation should only do "on curve" check
> * number of input-pairs into pairing operation and multiexp operation: empty input must return an error and not zero-point/identity

Both are true in the current document.

EIPS/eip-2537.md
```diff
@@ -320,7 +333,8 @@ There are no backward compatibility questions.
 
 ### Subgroup checks
 
-A subgroup check **is mandatory** during the pairing call. Implementations *should* use fast subgroup checks: at the time of writing, multiplication gas cost is based on the `double-and-add` multiplication method that has a clear "worst case" (all bits are equal to one). For pairing operations, it is expected that implementations use faster subgroup checks, e.g. by using the wNAF multiplication method for elliptic curves that is ~ `40%` cheaper with windows size equal to 4. (Tested empirically. Savings are due to lower hamming weight of the group order and even lower hamming weight for wNAF. Concretely, subgroup check for both G1 and G2 points in a pair are around `35000` combined).
+The subgroup check for point `H` verifies that `H` multiplied by scalar `q` yields the infinity point on the respective curve.
```
asanso, line 336: rather than use this naive expensive subgroup check (in particular for G2) I'd rather suggest to point to https://eprint.iacr.org/2021/1130.pdf

asanso:
> > * subgroup checks for non-pairing operations: I think they should be on the caller's responsibility, so implementation should only do "on curve" check
> > * number of input-pairs into pairing operation and multiexp operation: empty input must return an error and not zero-point/identity
> 
> Both are true in the current document.

I agree with @khovratovich that the current situation is ambiguous. Moreover, it is even more convoluted than that, as leaving the responsibility to callers to check the subgroup for non-pairing operations opens the door to subtle issues. For example, in order to expedite scalar multiplication, implementers might be tempted (within their rights) to apply GLV endomorphism optimization. However, the problem arises because the GLV is defined and valid only within the 'magic subgroup'.

EIPS/eip-2537.md
khovratovich: Can we provide complexity estimates for this method? What is `u`?

asanso: it's all in the paper. 'u' is the seed for generating BLS12-381 and pretty short ( -0xd201000000010000)

khovratovich: Reformulated the text with your link @asanso

shamatar:
If no one knows a good reasonable case for explicit wanting to mix-in not-main-subgroup point for non-pairing cases, I can perform some basic gas measurements in case of fast subgroup check. Hopefully costs should not change too much for multiplication and addition since they are dominated by one final inversion

khovratovich:
The CI validator does not like a link to ePrint. Probably we can ignore this error.

asanso:
@khovratovich  IMHO it would be good to separate the really good editorial changes and typos you corrected from the subgroup check discussion. How about slit this PR in 2 so we could merge the editorial changes/typos first ?

asanso (approved):

PR #8276, opened 2024-03-01
Update to eth/70

smartprogrammer93:
Updating to eth/70 as another eth protocol update will pick 69 as a version


lightclient:
Hi @smartprogrammer93, 7642 is merged now. I'm not sure why this is still failing but do you mind rebasing? I think that will resolve the issue and then this should merge.

PR #8277, opened 2024-03-01
Add EIP: History accumulator for pre-PoS data

lightclient:
This EIP defines a historical hashes accumulator for pre-merge data and commits to the root `0xec8e040fd6c557b41ca8ddd38f7e9d58a9281918dc92bdb72342a38fb085e701`.


EIPS/eip-7643.md
```diff
@@ -0,0 +1,1983 @@
...
+HistoricalHashesAccumulator = Container[
+    historical_epochs: List[bytes32, max_length=MAX_HISTORICAL_EPOCHS],
+    current_epoch: EpochRecord,
+]
+```
+
+### Pre-PoS Root
+
+The hash tree root of `HistoricalHashesAccumulator` for data before block 15537393 is
+`0xec8e040fd6c557b41ca8ddd38f7e9d58a9281918dc92bdb72342a38fb085e701`.
```
axic, line 57: Can you mine a fancy number (by shifting the data structure around)? 😅

lightclient: unfortunately not a lot a of malleability in the SSZ object 😅

EIPS/eip-7643.md
```diff
@@ -0,0 +1,1983 @@
...
+
+The total difficulty allowed so that clients may return the value for specific
+JSON-RPC methods which support it. It is also useful for verifying the TTD of
+the final proof-of-work block.
+
+## Backwards Compatibility
+
+n/a
+
+## Test Cases
```
g11tech, line 71:
```suggestion
## Mainnet PoW accumulator
```
or this is just some test data?

EIPS/eip-7643.md
```diff
@@ -0,0 +1,1983 @@
...
+01887,0xdededef33d83e607868d87e8f058511a152de71d8a04b12499d5f1666b89071d
+01888,0xcdbce5e36cfff749d0ee9ccb6dc344a822ed097e4b8eb812ada19521890d5663
+01889,0x4a09fe43bd97064993f18d214062005a1ffc76d03495b2b7c65c6137d3dd055e
+01890,0xff7a1b1127af6429b799637ff6cffb9169aa5000f01b67a54e2c7f9a155e7027
+01891,0x0f7bffdd015be88c78d30778d5bd66b75e86f9867e221f857352d0cb95c927a1
+01892,0x9671b1edc7dc58206e18de1e4d107bf2e8508483017749146ff5c82c674408e6
+01893,0x1b07973b9e818e0e014e52c587149c9775fd44bd5ccd095dee3221ba7c7a9efe
+01894,0x804008940c025a4e8a00ea42a659b484ba32c14dff133e9d3b7bf3685c1e54de
+01895,0x3f81607c8cb3f0448a11cab8df0e504b605581f4891a9a35bd9c0dd37a71834f
+01896,0xe6ebe562c89bc8ecb94dc9b2889a27a816ec05d3d6bd1625acad72227071e721
```
g11tech, line 1971: epoch 1897 should also be here for completeness?

EIPS/eip-7643.md
```diff
@@ -0,0 +1,1983 @@
...
+
+The historical hashes accumulator commits to the set of pre-merge headers and their
+associated total difficulty. The format for this data is defined as:
+
+```python
+EPOCH_SIZE = 8192 # blocks
+MAX_HISTORICAL_EPOCHS = 131072  # 2**17
+
+# An individual record for a historical header.
+HeaderRecord = Container[block_hash: bytes32, total_difficulty: uint256]
```
g11tech, line 43: just an idea: throw in all other header fields in there (with default values where fields didn't exist before a particular hardfork)?

lightclient: I'm really just creating an EIP around something that is already live on the portal network, so don't exactly have the ability to change the format further.

EIPS/eip-7643.md
acolytec3: Looks like it matches the mainnet accumulator so I'd definitely refer to it as such.

EIPS/eip-7643.md
acolytec3: @g11tech what's your motivation for adding the other header fields to the record?  We just need a unique record for each header and arguably even the blockhash should be sufficient.  There was some discussion in the past on possibly removing the total difficulty since it didn't really add any value for purposes of constructing the overall accumulator.

EIPS/eip-7643.md
acolytec3: I think the indices here just start with 0.  The hash for 01896 here matches the epoch hash root for [1897](https://github.com/ethereumjs/ultralight/blob/52a60520310608ef9de4b555cd9ee55805d72c13/packages/portalnetwork/src/networks/history/data/master.ts#L1898) in Ultralight

EIPS/eip-7643.md
g11tech: well i was wondering about having the utility of TD, but then i thought that if TD is useful may be other fields are useful too.

but TD is an accumulator so i guess I am fine

EIPS/eip-7643.md
g11tech: so what is `0xec8e040fd6c557b41ca8ddd38f7e9d58a9281918dc92bdb72342a38fb085e701` then?

acolytec3: I haven't constructed it locally to confirm but I'm assuming that's the hash tree root of the master accumulator SSZ container.  If it's urgent to verify, I can assemble it locally.  Maybe @lightclient can confirm?

lightclient: I don't think it is urgent, AFAIK it has been verified by 3 different implementations (portal, geth, and nimbus el)

g11tech (approved):
lgtm

PR #8287, opened 2024-03-05
Update EIP-7069: Rename and update semantics

pdobacz:
- The rename as in https://github.com/ipsilon/eof/pull/64
- The execution semantics to be more aligned with original CALL flow and with `evmone` implementation and most recent EOF conversations


EIPS/eip-7069.md
```diff
@@ -50,31 +50,32 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
...
-9. Perform the call with the available gas and configuration.
-10. Push a status code on the stack:
-    11a. `0` if the call was successful.
-    11b. `1` if the call has reverted.
-    11c. `2` if the call has failed.
-11. Gas not used by the callee is returned to the caller.
+7. Calculate the gas available to callee as caller's remaining gas reduced by `max(floor(gas/64), MIN_RETAINED_GAS)`.
+8. Fail with error if the gas available to callee at this point is less than `MIN_CALLEE_GAS`.
+9. If `value` is non-zero, fail lightly if the balance of the current account is less than `value`.
+10. Fail lightly if call stack depth equals `1024`.
```
shemnon, line 72: What does "fail Lightly" mean?  dose "Fail with error" mean revert and take all available gas and "fail lightly" mean fail consuming only already charged gas?

EIPS/eip-7069.md
```diff
@@ -50,31 +50,32 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
...
+11. Perform the call with the available gas and configuration.
+12. Push a status code on the stack:
+    12a. `0` if the call was successful.
+    12b. `1` if the call has reverted.
+    12c. `2` if the call has failed.
```
shemnon, line 77: This doesn't render as indended.

```suggestion
12. Push a status code on the stack:
    0. `0` if the call was successful.
    1. `1` if the call has reverted.
    2. `2` if the call has failed.
```

EIPS/eip-7069.md
pdobacz: yes (except "Fail with error" is not a revert but an exceptional abort, I presume this was the original intention), I'll specify these.

But you've uncovered an interesting issue - what should be pushed to stack in this case? For a light failure in EOFCREATE/TXCREATE we return `0` (with a different meaning), same for CREATE/CREATE2, but for revamped calls that would mean success.

I also see that `evmone` would return `0`, so making a self note to amend this (I'm working on that now in parallel).

pdobacz: Proposing a resolution (`3` on stack for light failure) https://github.com/ethereum/EIPs/pull/8287/commits/54f6dcd92c95a21841e68fd7888dab5e02f6f51a to push the discussion forward. WDYT?

EIPS/eip-7069.md
```diff
@@ -50,31 +50,33 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
...
-9. Perform the call with the available gas and configuration.
-10. Push a status code on the stack:
-    11a. `0` if the call was successful.
-    11b. `1` if the call has reverted.
-    11c. `2` if the call has failed.
-11. Gas not used by the callee is returned to the caller.
+7. Calculate the gas available to callee as caller's remaining gas reduced by `max(floor(gas/64), MIN_RETAINED_GAS)`.
+8. Halt with exceptional failure if the gas available to callee at this point is less than `MIN_CALLEE_GAS`.
+9. If `value` is non-zero, fail lightly if the balance of the current account is less than `value` (push `3` on the stack, only consume gas charged until this point).
+10. Fail lightly if call stack depth equals `1024` (push `3` on the stack, only consume gas charged until this point).
```
gumb0, line 72: Call depth check is actually superseded by 63/64 rule, see Rationale.

pdobacz: I thought you questioned the correctness of that initial removal of call depth check.

Isn't it a compatibility problem, that call-stack-depth-checking calls can alternate with non-checking ones in the call stack?

Why wasn't the call depth check removed already in legacy calls, if superseded by the 63/64 rule?

EIPS/eip-7069.md
pdobacz: https://github.com/ethereum/EIPs/pull/8287/commits/477e269900f21a4cdd6f48dfe08f36a8a750af79 adheres to what we spoke about on the EOF implementers call - `1` as status code for light failure

EIPS/eip-7069.md
pdobacz: for future reference - decided on the EOF Implementers call to keep it
1. least change to legacy
2. being cautious about the air-tightness of the (complex) gas-based 63/64 rule, and providing a simpler hard cap
3. being open to gas limits which are huge for various use cases beyond mainnet

EIPS/eip-7069.md
```diff
@@ -50,31 +50,32 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
 
 We introduce four new instructions:
 
-- `CALL2` (`0xf8`) with arguments `(target_address, input_offset, input_size, value)` 
-- `DELEGATECALL2` (`0xf9`) with arguments `(target_address, input_offset, input_size)`
-- `STATICCALL2` (`0xfb`) with arguments `(target_address, input_offset, input_size)`
+- `EXTCALL` (`0xf8`) with arguments `(target_address, input_offset, input_size, value)` 
+- `EXTDELEGATECALL` (`0xf9`) with arguments `(target_address, input_offset, input_size)` - delegate call
+- `EXTSTATICCALL` (`0xfb`) with arguments `(target_address, input_offset, input_size)` - static call
```
axic, line 55: The `- static call` comment I think is pointless now?

EIPS/eip-7069.md
```diff
@@ -50,31 +50,32 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
...
-7. Calculate the gas available to callee as caller's remaining gas reduced by `max(ceil(gas/64), MIN_RETAINED_GAS)` (`MIN_RETAINED_GAS` is 5000).
-8. Fail with error if the gas available to callee at this point is less than `MIN_CALLEE_GAS` (2300).
-9. Perform the call with the available gas and configuration.
-10. Push a status code on the stack:
-    11a. `0` if the call was successful.
-    11b. `1` if the call has reverted.
-    11c. `2` if the call has failed.
-11. Gas not used by the callee is returned to the caller.
+7. Calculate the gas available to callee as caller's remaining gas reduced by `max(floor(gas/64), MIN_RETAINED_GAS)`.
+8. Halt with exceptional failure if the gas available to callee at this point is less than `MIN_CALLEE_GAS`.
```
gumb0, line 70: Why is this exceptional failure? Seems could be light failure

EIPS/eip-7069.md
```diff
@@ -50,31 +50,32 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
...
+9. If `value` is non-zero, fail lightly if the balance of the current account is less than `value` (push `1` on the stack, only consume gas charged until this point).
```
gumb0, line 71: kinda redundant
```suggestion
9. Fail lightly if the balance of the current account is less than `value` (push `1` on the stack, only consume gas charged until this point).
```

EIPS/eip-7069.md
```diff
@@ -50,31 +50,32 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
...
+10. Fail lightly if call stack depth equals `1024` (push `1` on the stack, only consume gas charged until this point).
```
gumb0, line 72:
```suggestion
10. Fail with `1` error code returned on stack, if current call stack depth equals `1024`. Only gas charged until this point is consumed.
```
(just "fail lightly" is not very commonly known term)

EIPS/eip-7069.md
```diff
@@ -114,10 +115,13 @@ Before the 63/64th rule was introduced, it was required to calculate available g
...
 
-1. Removed the call depth check.
-2. Use the 63/64th rule, but
-    2a. ensure that at least 5000 gas is retained prior to executing the callee,
-    2b. ensure that at least 2300 gas is available to the callee.
+1. Use the 63/64th rule, but
+    - Ensure that at least `MIN_RETAINED_GAS` gas is retained prior to executing the callee,
+    - Ensure that at least `MIN_CALLEE_GAS` gas is available to the callee.
+
+Removing the call stack depth check was initially considered, but this would be incompatible with the original `*CALL` instructions, as well as `CREATE*` instructions, which can be intertwined with the new `EXT*CALL` instructions in the call stack. As such, keepeing the call stack depth involves the least change for legacy code.
```
gumb0, line 122: maybe
```suggestion
Removing the call stack depth check was initially considered, but this would be incompatible with the original `*CALL` instructions, as well as `CREATE*` instructions, which can be intertwined with the new `EXT*CALL` instructions in the call stack. As such, keeping the call stack depth involves no change affecting legacy code.
```

EIPS/eip-7069.md
pdobacz: The PR only changes the language on this one. And I see it as a sensible failure mode - consistent with an OOG. But light failure also makes sense on 2nd thought. @chfast @axic @shemnon any more opinions on this, any objections to changing to light failure?

EIPS/eip-7069.md
pdobacz: wait, is caller nonce ever implemented for calls? Looks like too much copied from the similar CREATE case, I'll leave the nonce mention out.

chfast (approved):

EIPS/eip-7069.md
gumb0: Ah right, nonce is only relevant for creates, sorry for confusion.

EIPS/eip-7069.md
```diff
@@ -50,31 +50,33 @@ Lastly, the introduction of the `RETURNDATA*` instructions ([EIP-211](./eip-211.
...
-Execution semantics of `*CALL2`:
+Execution semantics of `EXT*CALL`:
 
 1. Charge `WARM_STORAGE_READ_COST` (100) gas.
-2. Pop required arguments from stack, fail with error on stack underflow.
-4. If `value` is non-zero:
-    3a. Fail with error if the current frame is in `static-mode`.
-    3b. Fail with error if the balance of the current account is less than `value`.
-    3c. Charge `CALL_VALUE_COST` gas.
+2. Pop required arguments from stack, halt with exceptional failure on stack underflow.
```
gumb0, line 61: It seems at this point EIP tries to maintain compatibility with both legacy and EOF, so maybe
```suggestion
2. Pop required arguments from stack, halt with exceptional failure on stack underflow.
  - *Note: When implemented in EOF, stack underflow check is done during stack validation, and runtime check is omitted.* 
```

EIPS/eip-7069.md
```diff
@@ -114,10 +116,13 @@ Before the 63/64th rule was introduced, it was required to calculate available g
 
 We have changed the ruleset:
 
-1. Removed the call depth check.
-2. Use the 63/64th rule, but
-    2a. ensure that at least 5000 gas is retained prior to executing the callee,
-    2b. ensure that at least 2300 gas is available to the callee.
+1. Use the 63/64th rule, but
+    - Ensure that at least `MIN_RETAINED_GAS` gas is retained prior to executing the callee,
+    - Ensure that at least `MIN_CALLEE_GAS` gas is available to the callee.
```
gumb0, line 121: to remove weird 1-item numbered list
```suggestion
The 63/64th rule is still applied, but
    - At least `MIN_RETAINED_GAS` gas is retained prior to executing the callee,
    - At least `MIN_CALLEE_GAS` gas is available to the callee.
```

EIPS/eip-7069.md
```diff
@@ -114,10 +116,13 @@ Before the 63/64th rule was introduced, it was required to calculate available g
...
+
+Removing the call stack depth check was initially considered, but this would be incompatible with the original `*CALL` instructions, as well as `CREATE*` instructions, which can be intertwined with the new `EXT*CALL` instructions in the call stack. As such, keeping the call stack depth check involves no change affecting legacy code.
+
+Also, we find the simple (as opposed to the complex 63/64th rule) hard cap reassuring, that the call stack depth is limited, in case the gas rules can be bypassed. Lastly, the amount of gas to reach depth of 1024 is huge, but not absurdly huge and we want to avoid constrain ourselves by relying this check on current gas limits.
```
gumb0, line 125:
```suggestion
Also, we find the simple (as opposed to the complex 63/64th rule) hard cap reassuring, that the call stack depth is limited, in case the gas rules can be bypassed. Lastly, the amount of gas to reach depth of 1024 is huge, but not absurdly huge, and we want to avoid constraining ourselves by dependency of this check on current gas limits.
```

EIPS/eip-7069.md
```diff
@@ -129,18 +134,20 @@ Current call instructions return a boolean value to signal success: 0 means fail
 
 We change the value from boolean to a status code, where `0` signals success and thus it will be possible to introduce more non-success codes in the future, if desired.
 
+Status code `1` is used for both reverts coming from the callee frame and light failures encountered in the execution of the instructions. The reason for combining them is keeping the semantics similar to the original CALLs - both scenario preserve unused gas and continue being indistinguishable to the caller.
```
gumb0, line 137:
```suggestion
Status code `1` is used for both reverts coming from the callee frame and light failures encountered in the execution of the instructions. The reason for combining them is keeping the semantics similar to the original CALLs - both scenarios preserve unused gas and continue being indistinguishable to the caller.
```

EIPS/eip-7069.md
pdobacz: The idea is for this EIP to be fully legacy-complete and EOF-independent. EOF (Megaspec) lists the differences required it to be EOF-complete (which is lack of runtime stack underflow check and banning new opcodes in legacy).

Is there any other piece which is EOF-only in this EIP?

(this is just a question to your comment, I have no objections towards adding this note for completeness, and I'll add it)

gumb0 (approved):

PR #8302, opened 2024-03-11
Update EIP-6800: change tree key hashing

jsign:
This PR changes the way we do tree key hashing, switching from base field point serialization to mapping to a scalar field and byte serialization.

### DIFF
--- a/EIPS/eip-6800.md
+++ b/EIPS/eip-6800.md
@@ -42,7 +42,7 @@ def group_to_scalar_field(point: Point) -> int:
     if point == bandersnatch.Z:
         return 0
     else:
-        return int.from_bytes(point.map_to_base_field().to_bytes(32, 'little'), 'little') % BANDERSNATCH_MODULUS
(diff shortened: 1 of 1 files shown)


EIPS/eip-6800.md
```diff
@@ -139,13 +139,17 @@ def old_style_address_to_address32(address: Address) -> Address32:
 These are the positions in the tree at which block header fields of an account are stored.
 
 ```
+def hash_point_to_bytes(point: Point) -> int:
+    return group_to_scalar_field(point).to_bytes(32, 'little')
```
jsign, line 143: For more info: https://github.com/gballet/go-ethereum/pull/401 & https://github.com/ethereum/go-verkle/pull/428

jsign: Note that `group_to_scalar_field` is defined on L38.

kevaundray (approved):
LGTM

PR #8305, opened 2024-03-12
Update EIP-3670: Align with EOF Megaspec and other EIPs

pdobacz:
Updates the language of the EIP-3670, to align with [EOF Megaspec](https://github.com/ipsilon/eof/blob/main/spec/eof.md) and other changes in EOF EIPs. Mainly makes independent from the exact spec of EOF contract creation.

The gist of the EIP should remain the same.

### DIFF
--- a/EIPS/eip-3670.md
+++ b/EIPS/eip-3670.md
@@ -38,12 +38,10 @@ The EOF1 format provides following forward compatibility properties:
 
 ## Specification
 
-*Remark:* We rely on the notation of *initcode*, *code* and *creation* as defined by [EIP-3540](./eip-3540.md).
-
 This feature is introduced on the same block EIP-3540 is enabled, therefore every EOF1-compatible bytecode MUST be validated according to these rules.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-3670.md
```diff
@@ -65,45 +63,48 @@ This change poses no risk to backwards compatibility, as it is introduced at the
...
 ### Contract creation
 
-Each case should be tested for creation transaction, `CREATE` and `CREATE2`.
+Each case should be tested by submitting an EOF container to EOF contract creation (as specced out in a separate EIP).
 
-- Invalid initcode
-- Valid initcode returning invalid code
-- Valid initcode returning valid code
+- EOF container with an invalid code at zeroth code section
+- EOF container with an invalid code at code sections placed in various positions
```
chfast, line 69: I don't understand this.

pdobacz: Invalid code can be found in code section 0 or some other, test cases should cover that. It is phrased clunkily, I'll rework this. How about

```
- EOF container with invalid code in one of its code sections (test various code section indices)
```
?

pdobacz: I'm a terrible wordsmith here, how about fa60e768 ?

chfast (approved):

PR #8152, opened 2024-01-31
Update EIP-3540: Move to Review

pdobacz:
First step to update to be in sync with [Megaspec](https://github.com/ipsilon/eof/blob/main/spec/eof.md).

Opening as draft first to get a round of reviews on these changes, then update at least the EIP-7480 and add the CREATE3/CREATE4 EIP to keep references working. (EDIT: both done, CREATE3/4 EIP is EIP-7620)

Let's focus the first round of reviews on overall structure and scope of this doc, and also any potential issues with the spec itself, which may come to mind when looking through these changes -- if anything looks suspicious here please report.

Note that many of the removed portions concern contract creation which ~will be described from ground up in a new EIP (current specification is inside the [Megaspec](https://github.com/ipsilon/eof/blob/main/spec/eof.md)~ is EIP-7620).

PR #6261, opened 2023-01-05
EIP-4750: Change checks to asserts

chfast:
Change runtime checks to asserts coming from EOF verification.

PR #8312, opened 2024-03-13
Update EIP-4750: align with EOF Megaspec

pdobacz:
1. Adjust to current spec where the return stack no longer has the stack height
2. Update stack requirement logic
3. ~Update gas to 4 for RETF~
4. RETF from top frame

### DIFF
--- a/EIPS/eip-4750.md
+++ b/EIPS/eip-4750.md
@@ -92,7 +92,7 @@ If the code is valid EOF1, the following execution rules apply:
 4. Pops nothing and pushes nothing to operand stack.
 5. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.
 
-*Note:* EOF validation requirement defined below guarantees that return stack cannot be empty before `RETF`.
+*Note:* EOF validation requirement for 0th section to be non-returning (non-returning sections introduced in a separate EIP) guarantees that return stack cannot be empty before `RETF`.
 
 ### Code Validation
 
@@ -103,8 +103,7 @@ In addition to container format validation rules above, we extend code section v
 3. `RJUMP`, `RJUMPI` and `RJUMPV` immediate argument value (jump destination relative offset) validation:
     1. Code section is invalid in case offset points to a position outside of section bounds.
     2. Code section is invalid in case offset points to one of two bytes directly following `CALLF` instruction.
-4. First code section is invalid if it contains a `RETF` instruction.
-5. No unreachable sections are allowed, i.e. every section can be reached from section 0 with a series of `CALLF` instructions (section 0 is always reachable).
+5. No unreachable sections are allowed, i.e. every section can be reached from section 0 with a series of `CALLF` / `JUMPF` (`JUMPF` introduced in a separate EIP) instructions (section 0 is always reachable).
 
 ### Disallowed instructions
 
@@ -132,7 +131,7 @@ Alternative logic for `RETF` in the top frame could be to allow it during code v
 - end execution if the return stack is emptied by the `RETF` or
 - exceptionally halt if the return stack is empty before the `RETF`.
 
-This has been superseded with the validation rule of top frame never containing `RETF`, because such rule is valuable by itself for other reasons. Therefore all considerations of runtime behavior of `RETF` in the top frame were obsoleted.
+This has been superseded with the validation rule of top frame being non-returning (non-returning sections introduced in a separate EIP), because validating non-returning status of functions is valuable by itself for other reasons. Therefore all considerations of runtime behavior of `RETF` in the top frame were obsoleted.
 
 ### Code section limit and instruction size
 
--- a/EIPS/eip-4750.md
+++ b/EIPS/eip-4750.md
@@ -86,7 +86,7 @@ If the code is valid EOF1, the following execution rules apply:
 #### `RETF`
 
 1. Does not have immediate arguments.
-3. Charges 4 gas.
+3. Charges 3 gas.
 4. Pops nothing and pushes nothing to operand stack.
 5. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.
 
--- a/EIPS/eip-4750.md
+++ b/EIPS/eip-4750.md
@@ -67,7 +67,7 @@ If the code is valid EOF1, the following execution rules apply:
 #### `CALLF`
 
 1. Has one immediate argument,`target_section_index`, encoded as a 16-bit unsigned big-endian value.
-2. EOF validation guarantees that operand stack has enough items to use as callee's inputs.
+2. *Note:* EOF validation [EIP-5450](./eip-5450.md) guarantees that operand stack has enough items to use as callee's inputs.
 3. If operand stack size exceeds `1024 - type[target_section_index].max_stack_height + type[target_section_index].inputs` (i.e. if the called function may exceed the global stack height limit), execution results in exceptional halt. This also guarantees that the stack height after the call is within the limits.
 4. If return stack already has `1024` items, execution results in exceptional halt.
 5. Charges 5 gas.
@@ -81,18 +81,16 @@ If the code is valid EOF1, the following execution rules apply:
    
    Under `PC_post_instruction` we mean the PC position after the entire immediate argument of `CALLF`.
    
-   *Note:* Code validation rules of [EIP-5450](./eip-5450.md) guarantee there is always an instruction following `CALLF` (since terminating instruction or unconditional jump is required to be final one in the section), therefore `PC_post_instruction` always points to an instruction inside section bounds.
+   *Note:* EOF validation [EIP-5450](./eip-5450.md) guarantees there is always an instruction following `CALLF` (since terminating instruction or unconditional jump is required to be final one in the section), therefore `PC_post_instruction` always points to an instruction inside section bounds.
 8. Sets `current_section_index` to `target_section_index` and `PC` to `0`, and execution continues in the called section.
 
 #### `RETF`
 
 1. Does not have immediate arguments.
-2. EOF validation guarantees that operand stack has exact number of items to use as outputs.
+2. *Note:* EOF validation [EIP-5450](./eip-5450.md) guarantees that operand stack has exact number of items to use as outputs.
 3. Charges 3 gas.
 4. Pops nothing and pushes nothing to operand stack.
 5. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.
-
-Note: EOF validation requirement for 0th section to be non-returning guarantees that return stack cannot be empty before `RETF`.
 
 ### Code Validation
 
@@ -120,18 +118,19 @@ Dynamic jump instructions `JUMP` (`0x56`) and `JUMPI` (`0x57`) are invalid and t
 
 1. Execution starts at the first byte of the first code section, and PC is set to 0.
 2. Return stack is initialized empty.
-3. Stack underflow check is not performed anymore. EOF stack validation guarantees that it cannot happen at run-time.
+3. Stack underflow check is not performed anymore. *Note:* EOF validation [EIP-5450](./eip-5450.md) guarantees that it cannot happen at run-time.
 3. Stack overflow check is not performed anymore, except during `CALLF` as specified above.
 
 ## Rationale
 
-### `RETF` in the top frame ends execution vs exceptionally halts vs is just not allowed
+### `RETF` in the top frame ends execution vs exceptionally halts vs is not allowed during validation
 
-Alternative logic for `RETF` in the top frame could be to allow it during code validation and make it end execution. This would mean that return stack is initialized with a pair of zeros, and `RETF` ends execution when return stack is empty.
+Alternative logic for `RETF` in the top frame could be to allow it during code validation and make it either:
 
-An argument in favor of such alternative (as opposed to exceptionally halt) was always having at least one item in the return stack, and avoiding a special case for empty stack in the interpreter loop return stack underflow check.
+- end execution if the return stack is emptied by the `RETF` or
+- exceptionally halt if the return stack is empty before the `RETF`.
 
-This has however been superseded with the validation rule of top frame never containing `RETF`, thereby obsoleteing all considerations of runtime behavior of `RETF` in top frame.
+This has been superseded with the validation rule of top frame never containing `RETF`, because such rule is valuable by itself for other reasons. Therefore all considerations of runtime behavior of `RETF` in the top frame were obsoleted.
 
 ### Code section limit and instruction size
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-4750.md
```diff
@@ -103,6 +99,8 @@ In addition to container format validation rules above, we extend code section v
 3. `RJUMP`, `RJUMPI` and `RJUMPV` immediate argument value (jump destination relative offset) validation:
     1. Code section is invalid in case offset points to a position outside of section bounds.
     2. Code section is invalid in case offset points to one of two bytes directly following `CALLF` instruction.
+4. First code section is invalid if it contains a `RETF` instruction.
+5. No unreachable sections are allowed, i.e. every section is referenced by at least one non-recursive `CALLF`, and section 0 is implicitly reachable.
```
pdobacz, line 103: not sure if this belongs here or in 5450

EIPS/eip-4750.md
```diff
@@ -103,6 +99,8 @@ In addition to container format validation rules above, we extend code section v
 3. `RJUMP`, `RJUMPI` and `RJUMPV` immediate argument value (jump destination relative offset) validation:
     1. Code section is invalid in case offset points to a position outside of section bounds.
     2. Code section is invalid in case offset points to one of two bytes directly following `CALLF` instruction.
+4. First code section is invalid if it contains a `RETF` instruction.
```
pdobacz, line 102: This is superseded by non-returning functions in 6206, but those aren't a thing in 4750 yet, so this rule needs to be added for completeness

EIPS/eip-4750.md
gumb0: I would say here, as it's not related to stack validation.

pdobacz: okay, good. I asked b/c I thought I've seen such checks included in the spec/algo for stack validation, maybe  I confused them with unreachable code within sections. Leaving this point in here then.

EIPS/eip-4750.md
```diff
@@ -58,41 +58,37 @@ If the code is legacy bytecode, any of these instructions results in an *excepti
...
+- `type[i].max_stack_height = type_section_contents[i * 4 + 2:i * 4 + 4]` - maximum operand stack height of ith section
 
 If the code is valid EOF1, the following execution rules apply:
 
 #### `CALLF`
 
 1. Has one immediate argument,`target_section_index`, encoded as a 16-bit unsigned big-endian value.
-2. EOF validation guarantees that operand stack has at least `caller_stack_height + type[target_section_index].inputs` items.
-3. If operand stack size exceeds `1024 - type[target_section_index].max_stack_height` (i.e. if the called function may exceed the global stack height limit), execution results in exceptional halt. This also guarantees that the stack height after the call is within the limits.
+3. If operand stack size exceeds `1024 - type[target_section_index].max_stack_height + type[target_section_index].inputs` (i.e. if the called function may exceed the global stack height limit), execution results in exceptional halt. This also guarantees that the stack height after the call is within the limits.
```
gumb0, line 70: I would leave point 2, with some rephrasing, it was added already with stack validation in mind (https://github.com/ethereum/EIPs/pull/6261/files)
```suggestion
2. EOF validation guarantees that operand stack has enough items to use as callee's inputs.
3. If operand stack size exceeds `1024 - type[target_section_index].max_stack_height + type[target_section_index].inputs` (i.e. if the called function may exceed the global stack height limit), execution results in exceptional halt. This also guarantees that the stack height after the call is within the limits.
```

EIPS/eip-4750.md
```diff
@@ -39,9 +39,9 @@ Refer to [EIP-3540](./eip-3540.md) to see the full structure of a well-formed EO
 
 ### New execution state in EVM
 
-A return stack is introduced, separate from the operand stack. It is a stack of items representing execution state to return to after function execution is finished. Each item is comprised of: code section index, offset in the code section (PC value), calling function stack height.
+A return stack is introduced, separate from the operand stack. It is a stack of items representing execution state to return to after function execution is finished. Each item is comprised of: code section index, offset in the code section (PC value).
```
gumb0, line 42:
```suggestion
A return stack is introduced, separate from the operand stack. It is a stack of items representing execution state to return to after function execution is finished. Each item is comprised of code section index and offset in the code section (PC value).
```

EIPS/eip-4750.md
```diff
@@ -58,41 +58,37 @@ If the code is legacy bytecode, any of these instructions results in an *excepti
...
+   Under `PC_post_instruction` we mean the PC position after the entire immediate argument of `CALLF`.
    
    *Note:* Code validation rules of [EIP-5450](./eip-5450.md) guarantee there is always an instruction following `CALLF` (since terminating instruction or unconditional jump is required to be final one in the section), therefore `PC_post_instruction` always points to an instruction inside section bounds.
 8. Sets `current_section_index` to `target_section_index` and `PC` to `0`, and execution continues in the called section.
 
 #### `RETF`
 
 1. Does not have immediate arguments.
-2. EOF validation guarantees that operand stack has exactly `caller_stack_height + type[current_section_index].outputs` items.
 3. Charges 3 gas.
```
gumb0, line 89:
```suggestion
2. EOF validation guarantees that operand stack has exact number of items to use as outputs.
3. Charges 3 gas.
```

EIPS/eip-4750.md
```diff
@@ -58,41 +58,37 @@ If the code is legacy bytecode, any of these instructions results in an *excepti
...
 4. Pops nothing and pushes nothing to operand stack.
 5. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.
```
gumb0, line 91:
```suggestion
5. Pops an item from return stack and sets `current_section_index` and `PC` to values from this item.

Note: EOF validation requirement for 0th section to be non-returning guarantees that return stack cannot be empty before `RETF`.
```

EIPS/eip-4750.md
gumb0: Stack validation also is defined in a later EIP, yet we're speccing this out having it in mind. So I think it's fine to refer to unreachableness, I added a suggestion about to `RETF` section. This item here I would probably remove.

(it's also not complete, code section is also invalid if it contains `JUMPF` to returning)

EIPS/eip-4750.md
gumb0, line 103: To reflect new aspect that was discovered recently
```suggestion
5. No unreachable sections are allowed, i.e. every section can be reached from section 0 with a series of `CALLF`/`JUMPF` instructions.
```

EIPS/eip-4750.md
```diff
@@ -117,17 +115,19 @@ Dynamic jump instructions `JUMP` (`0x56`) and `JUMPI` (`0x57`) are invalid and t
 ### Execution
 
 1. Execution starts at the first byte of the first code section, and PC is set to 0.
-2. Return stack is initialized to contain one item: `(code_section_index = 0, offset = 0, stack_height = 0)`
-3. If any instruction access the operand stack item below `caller_stack_height`, execution results in exceptional halt. This rule replaces the old stack underflow check.
-4. No change in stack overflow check: if any instruction causes the operand stack height to exceed `1024`, execution results in exceptional halt.
+2. Return stack is initialized empty.
+3. Stack underflow check is not performed anymore.
```
gumb0, line 119:
```suggestion
3. Stack underflow check is not performed anymore. EOF stack validation guarantees that it cannot happen at run-time.
```

EIPS/eip-4750.md
```diff
@@ -117,17 +115,19 @@ Dynamic jump instructions `JUMP` (`0x56`) and `JUMPI` (`0x57`) are invalid and t
...
+### `RETF` in the top frame ends execution vs exceptionally halts vs is just not allowed
 
-Alternative logic for executing `RETF` in the top frame could be to exceptionally halt execution, because there is arguably no caller for the starting function. This would mean that return stack is initialized as empty, and `RETF` exceptionally aborts when return stack is empty.
+Alternative logic for `RETF` in the top frame could be to allow it during code validation and make it end execution. This would mean that return stack is initialized with a pair of zeros, and `RETF` ends execution when return stack is empty.
 
-We have decided in favor of always having at least one item in the return stack, because it allows to avoid having a special case for empty stack in the interpreter loop stack underflow check. We keep the stack underflow rule general by having `caller_stack_height = 0` in the top frame.
+An argument in favor of such alternative (as opposed to exceptionally halt) was always having at least one item in the return stack, and avoiding a special case for empty stack in the interpreter loop return stack underflow check.
```
gumb0, line 128: I don't think it makes sense to mention this obsolete motivation, I would shrink this section:

```
Alternative logic for `RETF` in the top frame could be to allow it during code validation and make it either end execution or exceptionally halt. This would mean that return stack is initialized with a pair of zeros, and `RETF` ends execution or aborts when return stack is empty.

This has been superseded with the validation rule of top frame being non-returning, because validating non-returning status of functions is valuable by itself for other reasons. Therefore all considerations of runtime behavior of `RETF` in the top frame were obsoleted.
```

EIPS/eip-4750.md
pdobacz: > (it's also not complete, code section is also invalid if it contains JUMPF to returning)

But JUMPF in particular is not yet known, and is not required for this spec (also including 5450, as linked and mentioned) to be complete.

I think the point (4.) makes the spec self-contained and independent of at least one EIP (6206). One can also argue that it makes 4750 compatible with 6206 - the rule in 6206 is a generalization of this (4.) one, and by that, 6206 doesn't invalidate code which would be otherwise valid without 6206.

I would prefer to strive to decouple the EIPs wherever possible, like here we're making 6206 an optional extension to 4750. 5450 is much harder to leave out from this EIP than 6206

EIPS/eip-4750.md
pdobacz: 1. as above, I would leave JUMPF out. I also don't think it's ok to mention an entire new opcode not introduced here, without linking 6206. And the latter we don't want to do for obvious reasons
2. I think we should leave "section 0 being reachable" in, if we're pedantic.

EIPS/eip-4750.md
pdobacz: sounds good, with the caveat I'd not mention the "top frame being non-returning" term, same reasons as above. "Top frame not containing RETF" should be the equivalent term not leveraging 6206. But agreed with compressing the reasoning. Let's discuss the 6206 thing first, and then I'll fix here accordingly

pdobacz: on second read I noticed that the shrinked version is somewhat imprecise, b/c it conflates two earlier designs:
- empty return stack on `RETF` => abort
- emptied return stack after `RETF` => end execution.

I'll try to rephrase this

EIPS/eip-4750.md
pdobacz: @gumb0 I pushed a version of the document still decoupled from 6206, as I would suggest to proceed, but have a version with 6206 referenced (w/ JUMPF and non-returning fs, more in line with your suggestion) stashed away.

PTAL, I would opt to remain decoupled from 6206 as long as reasonably possible.

gumb0: I think trying to make EIPs independent of each other is a lost cause by now. 

My concern is potential confusion for the people reading two EIPs, seeing these two different versions of the validation rule and needing to go through mental effort to figure out how these versions relate to each other and what supersedes what. It's not trivial to figure out, it's formulated very differently in 6206.

pdobacz: > I think trying to make EIPs independent of each other is a lost cause by now.

For 3540, after feedback received, we strived to remove as many dependencies as possible, see [here](https://github.com/ethereum/EIPs/pull/8152#pullrequestreview-1922905668) and [here](https://github.com/ethereum/EIPs/pull/8152/commits/b45d6076c2e9bd59b211eb9e4d1e0d85eff089ed).

I think we're giving up too early on this 4750->6206 reference, and should also make the effort to decouple. That's also the main reason for the EIPs - to spec out new concepts in isolation, without needing to jump around different docs.

> My concern is potential confusion for the people reading two EIPs, seeing these two different versions of the validation rule and needing to go through mental effort to figure out how these versions relate to each other and what supersedes what. It's not trivial to figure out, it's formulated very differently in 6206.

This is a fair point, there should be no confusion, but I think it is not that bad, as the 4750 rules are generalized (not superseded really, I used a wrong word) and made implicit by 6206 rules. I reread 6206 and it seems fine to me.

I pushed the "6206-ish" version in a commit [a12f35ec](https://github.com/pdobacz/EIPs/commit/a12f35ec0f900c33748f8cbbbad199f7e16a67d5) for comparison. The diff is really small and shows the 6206 rules generalize 4750 rules nicely. And if there still remains risk of confusion, maybe better to clarify in 6206 ("rule x generalizes rule y from 4750" or just generally "Some rules generalize rules introduced in 4750")

gcolvin:
I'll offer a high-level comment here:  the Motivation section could be a lot stronger and more detailed than just "restricts them from making optimisations" and " improve analysis opportunities."  To sell this proposal it must be stronger.

As I've said too often, in the presence of dynamic jumps and the absence of function calls it is currently _impossible_ to traverse the control flow of EVM code in linear time.  Which makes many kinds of deploy-time analyses or rewrites of the code subject to DoS.  That includes the validations proposed here, translations to more efficiently interpreted code, compilation to machine code, and more.  

Dynamic jumps also make some offline static analyses take quadratic time, (and I'm told it makes some take exponential time, and even makes some undecidable.) And it forces some ZK EVMs into parallel execution.  And no doubt there many other problems I don't know about.

This is all unsurprising to me and the people dealing with these problems, but I've found it difficult to explain.

But to really get it you have to know
* what a directed graph is,
* what a cyclic directed graph is,
* how to traverse a cyclic graph acyclically, and
* why dynamic jumps make that traversal non-linear.

And given that,
* how code gets represented as a directed control-flow graph,
* how that representation is used in many ways for many purposes,
* including things we want to do at deploy time, and
* why non-linear traversal time makes these things a DoS surface.

And indeed if you haven't studied this stuff you might not get it.

So I don't know if it's worth a brief "tutorial" here or elsewhere, but I'm sure it would be easier than explaining elliptic curve cryptography to me.

gumb0 (approved):

pdobacz:
> I'll offer a high-level comment here...

@gcolvin Thanks for posting this, and apologies about not responding earlier. All great points but beyond the scope of this PR, let's have a separate PR if we want to strengthen the Motivation section.

gcolvin:
> @gcolvin Thanks for posting this, and apologies about not responding earlier. All great points but beyond the scope of this PR, let's have a separate PR if we want to strengthen the Motivation section.

Understood.  I have the misfortune to have very little time to devote to a topic dear to my heart -- no PR is coming from me.  So I just dropped this off while passing through, as it were.

PR #8317, opened 2024-03-14
Update EIP-7516: fix blobbasefee opcode number in nominal test case

enitrat:
Fixes the number of the BLOBBASEFEE opcode in the nominal test case. 0x4900 -> 0x4a00 for (BLOBBASEFEE, STOP)


enitrat:
:(

PR #8322, opened 2024-03-15
Update EIP-2537: Rephrased subgroup check part

asanso:

--

 - The build passes.


asanso:
@shamatar, @ineffectualproperty, @ralexstokes I rephrased a bit the subgroup check section including the fast subgroup check. 
Once we are on it, I would like to continue the discussion started in https://github.com/ethereum/EIPs/pull/8274#issuecomment-1973386322. 
Namery IMHO we need to decide if to employ subgroup checks only for pairing or for all operations.
Thoughts ?

Queenkaz (approved):

shamatar:
The nature of suggestion is great, but now phrasing mentioning double and add and subgroup check is misleading. If you just say that
- one must check
- one should use fast subgroup checks
- default gas pricing assumes naive double-and-add multiplication

it'll be clear enough

shamatar:
Separately on mandatory checks: if no one found any use case when out of group elements are interesting for multiplication/addition then we can make subgroup checks mandatory iff fast subgroup checks are also mandatory. Of course, gas schedule will change, but most likely change will only be substantial for G1 add

Queenkaz (approved):

shamatar:
> any suggestion on the words to use ? the "double-and-add" part was already present in the original text

@asanso I meant that we can just remove/simplify that paragraph and put compact information
- implementation MUST perform subgroup checks
- default gas schedule uses double-and-add multiplication cost as a cost of subgroup check
- nevertheless, implementers SHOULD (or at least encouraged) to use faster subgroup checks

asanso:
bq. default gas schedule uses double-and-add multiplication cost as a cost of subgroup check

@shamatar  do we really want to go that route? For Fp2  double-and-add multiplication is super expensive and if we add subrgroup checking for any operations well gas schedule becomes prohibitive!

shamatar:
I meant that in the current cost of pairing (and only pairing) there is an included cost of subgroup check assuming double-and-add for the check implementation. I do not refer to a global approach change to make subgroup checks mandatory, but only to rephrasing a current paragraph

asanso:
ok my proposal is a bit wider than that. I meant to change the document mandate subgroup check to all operations as previously discussed and you seemed to agree :)

asanso:
@shamatar I adjusted the text to include ALL the calls and added a new separate document (similar to the existing one for hash to curve) in https://github.com/asanso/EIPs/blob/patch-9/assets/eip-2537/fast_subgroup_checks.md .

Thoughts ? cc @ineffectualproperty @ralexstokes

shamatar:
Actually I found an example where we may need to allow out-of-main-subgroup points https://github.com/ethyla/bls12-381-hash-to-curve/blob/main/src/readable_Hash_to_curve.sol

asanso:
@shamatar what exactly in that file?

shamatar:
It's a full hash-to-curve implementation (starting from byte string and then using field element -> curve point precompile). I'm checking how is map-to-curve specified now. If it clears a cofactor internally then such case would not be applicable, but as far as I remember it's not. 

Full author's comment https://ethereum-magicians.org/t/eip-2537-bls12-precompile-discussion-thread/4187/60

shamatar:
At least my implementation clears a cofactor for fp-to-g1. Now checking the spec

shamatar:
Spec is also defined to clear a cofactor

asanso:
So one reason more to work in the subgroup for all the operations correct ?

shamatar:
Kind of. I think original IETF draft for mapping to curve suggests to get two points, add them and then clear cofactor, that makes sense because clearing cofactor is amortized this way. But if we want to keep a number of exposed precompile function manageable, we can build-in those cofactor cleanups. 

My only concern is at some point we will see a use case where secondary subgroup point is needed. Yes, it's hard to produce it in EVM because precompiles will give only main subgroup, but I do not claim a knowledge of all possible cases

mratsim:
I've also mentioned concerned about subgroup checks here: https://ethereum-magicians.org/t/eip-2537-bls12-precompile-discussion-thread/4187/63

The issue is that without them, libraries cannot use endomorphism acceleration or the result will be wrong. They might also not be able to use projective coordinates with complete formula (from Renes 2015), though Jacobian coordinates would work OK.

As mentioned in https://ethereum-magicians.org/t/eip-2537-bls12-precompile-discussion-thread/4187/66, there are 3 paths ahead:

1. No change approach. no subgroup check, price like double-and-add and Pippenger today and add test vectors. For MSMs, in both Gnark and Constantine, endomorphism acceleration was not a problem or was even slower due to probably decomposition overhead + memory bandwidth / hardware cache limitations (at least on x86, Apple memory likely behaves differently).
2. Misuse resistance approach. See [@asanso](https://ethereum-magicians.org/u/asanso), enforce subgroup check except on addition.
3. Modular approach or performance approach, provide subgroup_check and clear_cofactor as separate primitives. People can cache subgroup_checks and ensure their cost is only paid once. Issue: they need to learn what a subgroup and cofactor is.

Some important point is that for large MSMs, endomorphism acceleration doesn't help, hence subgroup check cost is paid in vain. For single scalar mul, it represents about 1/3 (G1) and 1/5 (G2) more operations. (benchmarks in the link)

I do agree with @asanso that having them on all operations is reasonable due to the risk of some crypto libraries defaulting to endomorphism acceleration and outputting wrong result.

EIPS/eip-2537.md
```diff
@@ -338,13 +338,14 @@ There are no backward compatibility questions.
...
-
+The check **is mandatory** during **ALL** the calls. Implementations *should* use fast subgroup checks. It can be stated that it **must** be performed by fast subgroup checks to achieve a speedup, resulting in a discount over naive implementations based on the `double-and-add` multiplication method, which has a clear "worst-case" scenario where all bits are equal to one.
+Before accepting an input, it is recommended to subject the input to the appropriate endomorphism test as described in https://eprint.iacr.org/2021/1130.pdf.
```
mratsim, line 342: MUST perform a subgroup check, and reject invalid point.

mratsim:
https://github.com/mratsim/constantine/pull/368#issuecomment-2047473465

> For @asanso on gas pricing
> 
> gas costs in ratio of G1 scalarmul image
> 
> original: https://gist.github.com/mratsim/6785a29e72865cfa94e1174fae1e1168 image
> 
> Reproduction
> 
> ```shell
> git clone https://github.com/mratsim/constantine
> cd constantine
> git checkout eip2537
> CC=clang nimble bench_eip2537_subgroup_checks_impact
> ```

EIPS/eip-2537.md
asanso: @mratsim  what do you mean? is there anything wrong with this part of the text ?

mratsim: AFAIK, in RFC **MUST** is the way to specify that something is mandatory, and the spec should explicitly mention that it returns as an error on failure.

asanso: @mratsim  Is the problem the capitalization of 'must' rather than using lowercase? Or am I missing something else? :)

mratsim: To be clear, I have 2 issues:

1. Instead of **is mandatory**: "scalar multiplications, MSMs and pairings MUST perform a subgroup check. Implementations SHOULD use the fast subgroup check method from 2021/1130 detailed in _link to fast subgroupcheck markdown file_
2. "On inputs that fail the subgroup check the precompile MUST return an error."

Rationale:
- MUST is how mandatory steps are usually specified. I added details since EC add and hashing-to-curve do not have subgroup checks. And now we have a dedicated document for fast subgroup-checks.
- Similar to the paragraph on _Field elements encoding_ with "On inputs that can not be a valid encodings of field elements the precompile must return an error.", it's important to spell what happens on failure or people have to look at the implementations to make sure. For example what if for MSM, the offending point and its corresponding coefficient were skipped?

I can make a PR to this PR.

asanso: >I can make a PR to this PR.

@mratsim  this would be great.
One you are on it I everyone agrees you can already to the PR to this PR with this strategy:

2. Misuse resistance approach. See [@asanso](https://ethereum-magicians.org/u/asanso), enforce subgroup check except on addition.

assuming @shamatar @ralexstokes @ineffectualproperty agree

PR #8324, opened 2024-03-15
Update EIP-7547: Change modification to payload attribute SSE event

mcdee:
Change so that the `payload_attributes` SSE event output includes the inclusion list summary.

(I believe that this is what is meant, feel free to close this if not.)

### DIFF
--- a/EIPS/eip-7547.md
+++ b/EIPS/eip-7547.md
@@ -126,7 +126,7 @@ class BeaconBlockBody(Container):
 
 **Builder and API spec:**
 
-- ***Modified** payload attribute SSE endpoint:* Adjust the payload attribute SSE endpoint to include payload summaries.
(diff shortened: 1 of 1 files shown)


EIPS/eip-7547.md
```diff
@@ -126,7 +126,7 @@ class BeaconBlockBody(Container):
 
 **Builder and API spec:**
```
terencechain, line 127: Feels odd for the EIP to reference a builder API, given that it is an element outside of the protocol. As I'm not well-versed in the EIP process, is it common for such external to be mentioned? 

Additionally, regarding the endpoint modification, builder could just monitor the inclusion list through the gossip network so no additional modification of end point is required.

mcdee: Okay; I've updated the PR to remove the offending entry.

terencechain (approved):
lgtm! might want to update the title

PR #8343, opened 2024-03-21
Add EIP: Stepwise Blob Throughput Increase

adietrichs:
todo:

- [ ] add section on impact of update fraction changes on blob basefee
- [ ] potentially specify epoch parameter logic more explicitly (how to derive epoch from header)
- [ ] additional storage requirements for blobs


EIPS/eip-XXXX.md
```diff
@@ -0,0 +1,93 @@
+---
+eip: tbd
```
abcoathup, line 2:
```suggestion
eip: 7659
```
Numbering changed to sequential from 7500.

PR #8354, opened 2024-03-25
Update EIP-6800: remove gas reform section

jsign:
Remove section that is the [same as EIP-4762 "Gas reform" section](https://eips.ethereum.org/EIPS/eip-4762#gas-reform) which explain gas changes for Verkle.


gballet (commented):
I think it's a pretty good paragraph that should stay, with a link to eip 4762 to be added. At the very least, this paragraph should be moved to eip 4762.

jsign:
> Remove section that is the [same as EIP-4762 "Gas reform" section](https://eips.ethereum.org/EIPS/eip-4762#gas-reform) which explain gas changes for Verkle.

Clarifying here, too, since it might be confusing for others. The same paragraph is in EIP-4762. The link in the PR description can confuse you since, considering it's the same paragraph (with an extra parenthesis), you might think it's EIP-6800, but it's EIP-4762; so no information is lost.

PR #8361, opened 2024-03-29
Update EIP-7620: Update spec

lukasrosario:

--

 - The build passes.

### DIFF
--- a/EIPS/eip-5792.md
+++ b/EIPS/eip-5792.md
@@ -2,7 +2,7 @@
 eip: 5792
 title: Wallet Call API
 description: Adds JSON-RPC methods for sending multiple calls from the user's wallet, and checking their status
-author: Moody Salem (@moodysalem)
+author: Moody Salem (@moodysalem), Lukas Rosario (@lukasrosario), Wilson Cusack (@wilsoncusack), Dror Tirosh (@drortirosh)
 discussions-to: https://ethereum-magicians.org/t/eip-5792-wallet-abstract-transaction-send-api/11374
 status: Stagnant
 type: Standards Track
--- a/EIPS/eip-5792.md
+++ b/EIPS/eip-5792.md
@@ -1,7 +1,7 @@
 ---
 eip: 5792
-title: Wallet Function Call API
-description: Adds JSON-RPC methods for sending multiple function calls from the user's wallet, and checking their status
+title: Wallet Call API
+description: Adds JSON-RPC methods for sending multiple calls from the user's wallet, and checking their status
 author: Moody Salem (@moodysalem)
 discussions-to: https://ethereum-magicians.org/t/eip-5792-wallet-abstract-transaction-send-api/11374
 status: Stagnant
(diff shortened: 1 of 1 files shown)


EIPS/eip-5792.md
```diff
@@ -1,390 +1,267 @@
...
-additional effort, as well as provide a better UX for sending bundles of function calls (
-e.g. [EIP-20](./eip-20.md) `#approve` followed by a contract call).
+Defines new JSON-RPC methods for apps to send batches of calls from a user's wallet, as well as check on the status of those calls. The new methods have a simple core while enabling additional transaction features (e.g. sponsored transactions with a paymaster) via the introduction of capabilities. To support this, there is also a new method for querying for the capabilities a given wallet supports. The new methods are more abstract in regard to the underlying transactions than the existing transaction sending APIs to allow for differences between wallet implementations, e.g. smart accounts utilizing [ERC-4337](./eip-4337.md) or EOA wallets that support batched transactions via [EIP-3074](./eip-3074.md). Apps may use this more abstract interface to support different kinds of wallets without additional effort, as well as provide a better UX for sending batches of function calls (e.g. [ERC-20](./eip-20.md) `approve` followed by a contract call).
```
wilsoncusack, line 15: Not perfect, but I think this would improve 

```suggestion
Defines new JSON-RPC methods which enable apps to ask a wallet to process a batch of onchain write calls and to check on the status of those calls. Calls can be specified to make use of various wallet-defined capabilities, and there is also a new RPC to enable apps to discover which capabilities a wallet supports. 
```

EIPS/eip-5792.md
```diff
@@ -1,390 +1,267 @@
...
-existing EOA wallets. EOA wallets may send bundles of function calls as individual transactions. The goal of the new
-interface is to be a more stable and flexible interface that enables a better user experience, while wallet
-implementations evolve over time.
+The current methods used to send transactions from the user wallet and check their status are `eth_sendTransaction` and `eth_getTransactionReceipt`.
+
+The current methods used to send transactions from the user wallet and check their status do not meet modern developer demands and cannot accomodate new transaction formats. Even the name–`eth_sendTransaction`–is an artifact of a time when nodes served as wallets. 
+
+Today, developers want to send batch transactions, which many smart accounts can support atomically. Developers also want to make use of features afforded by new transaction formats, like paymasters in [ERC-4337](./eip-4337.md) transactions. `eth_sendTransaction` offers no way to do these things. 
+
+In updating to a new set of `wallet_` RPCs, our main goals are to enforce a clean separation of wallet and app concerns, and create a clear way for more features to be added over time with minimal coordination.
```
wilsoncusack, line 25:
```suggestion
In updating to a new set of `wallet_` RPCs, our main goals are to enforce a clean separation of wallet and app concerns, enable developers to make use of things like paymasters and batch transactions, and to create a clear way for more features to be added over time with minimal coordination.
```

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+* MAY reject the request if the request chain ID does not match the currently selected chain ID.
+* MAY reject the request if the from address does not match the enabled account.
+* MAY reject the request if one or more calls in the batch is expected to fail, when simulated sequentially
+
+#### `wallet_sendCalls` RPC Specification
+
+```typescript
+type SendCallsParams = {
+  version: string;
+  chainId: `0x{string}`; // Hex chain id
```
nateReiners, line 94: nit: dropped a $ here

vvvv666:
> Updating EIP-5792 according to https://hackmd.io/@lsr/proposal-for-new-wallet-rpcs and resulting discussions.
> 
> Main changes:
> 
> * Name changes
> * Introduction of capabilities
> * Changes to the shape of call sending method
> * Atomic batch capability spec

moodysalem (approved):
haven't fully reviewed it, just a few nits in the first parts, but approving b/c ok with coinbase team taking ownership of the eip from here on

vvvv666:
> Updating EIP-5792 according to https://hackmd.io/@lsr/proposal-for-new-wallet-rpcs and resulting discussions.
> 
> Main changes:
> 
> * Name changes
> * Introduction of capabilities
> * Changes to the shape of call sending method
> * Atomic batch capability spec

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
 
 ## Specification
 
-Three new JSON-RPC methods are added. Dapps may begin using these methods immediately, falling back
-to `eth_sendTransaction` and `eth_getTransactionReceipt` when they are not available.
+Four new JSON-RPC methods are added. Apps may begin using these methods immediately, falling back to `eth_sendTransaction` and `eth_getTransactionReceipt` when they are not available.
+
+We define one capability related to call execution atomicity.
+
+### `wallet_getCapabilities`
```
arein, line 33: I suggest to remove this method and instead to reference CAIPs (which you can now reference from EIPs/ERCs the way I did when I [updated this same EIP a few months back](https://github.com/arein/dapp-smart-account/pull/1/files#diff-2ce6a27d54a5083f40751babba9d77c5529f0515cd5f8dc31a379406d52ff6cdR38).

This solution I argue is better as it doesn't ever require a roundtrip to discover the capabilities.

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+
+```json
+{
+  "0x2105": {
+    "paymasterService": {
+      "supported": true
+    },
+    "sessionKeys": {
+      "supported": true
+    }
```
arein, line 61: what does this mean? Should this not refer to other EIPs? e.g. `eth_sendUserOperation` to me depicts support for Paymaster.

Session keys is the TBD standard we are discussing in Telegram.

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
 
 ## Backwards Compatibility
 
-Wallets that do not support the following methods should return error responses to the new JSON-RPC methods.
-Dapps MAY attempt to send the same bundle of calls via `eth_sendTransaction` when they receive a not implemented
-call, or otherwise indicate to the user that their wallet is not supported.
+Wallets that do not support the following methods should return error responses to the new JSON-RPC methods. Apps MAY attempt to send the same batch of calls via `eth_sendTransaction` when they receive a not implemented call, or otherwise indicate to the user that their wallet is not supported.
```
arein, line 262: In your proposal an app needs to call `wallet_getCapabilities` first. And handling errors if it's not supported. That's an extra roundtrip and odd API (error handling vs. explicit).

Following my suggestion above that would be saved.

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
-wallets without considering the differences between wallet implementations. These methods are backwards compatible with
-existing EOA wallets. EOA wallets may send bundles of function calls as individual transactions. The goal of the new
-interface is to be a more stable and flexible interface that enables a better user experience, while wallet
-implementations evolve over time.
+The current methods used to send transactions from the user wallet and check their status are `eth_sendTransaction` and `eth_getTransactionReceipt`.
+
+The current methods used to send transactions from the user wallet and check their status do not meet modern developer demands and cannot accomodate new transaction formats. Even the name–`eth_sendTransaction`–is an artifact of a time when nodes served as wallets. 
+
+Today, developers want to send batch transactions, which many smart accounts can support atomically. Developers also want to make use of features afforded by new transaction formats, like paymasters in [ERC-4337](./eip-4337.md) transactions. `eth_sendTransaction` offers no way to do these things.
```
arein, line 23: It'd be a good though-experiment if this can also support EIP-3074

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+    },
+    "sessionKeys": {
+      "supported": true
+    }
+  },
+  "0x14A34": {
+    "paymasterService": {
+      "supported": true
+    }
+  }
```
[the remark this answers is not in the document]
lukasrosario: this is something we went back and forth on. i think this is nice because you can save a round trip if your app is multichain. but yeah have gotten this feedback before, maybe we allow for both -- i.e. without a chain id to get all chains, and with a chain id to get capabilities for just the request chain id.

@jxom curious if you have thoughts on hex chain id as keys. viem / wagmi use numbers but i believe this could be problematic because chain id can be uint256? and ts makes a fuss about bigint as keys.

EIPS/eip-5792.md
wilsoncusack: I think this is just an example of how wallet capabilities could be formatted. But maybe should be a different example as we don't specify any more on this, here.

EIPS/eip-5792.md
wilsoncusack: Can you make clear your suggestion: what call will the app make to the wallet in order to discover capabilities?

EIPS/eip-5792.md
wilsoncusack: Should definitely be able to!

EIPS/eip-5792.md
wilsoncusack: FWIW I am open to the idea that `eth_requestAccounts` as the implicit connection method should possibly be replaced by some like `wallet_connect` RPC that shares accounts + capabilities + other info. Don't necessarily want to block this EIP on getting to clarity there, but if we can align on something better quickly we should!

User7223141801:
``

User7223141801:
1. 

> _**

EIPS/eip-5792.md
arein: Yes I saw you commented on https://github.com/arein/dapp-smart-account/pull/1/files#diff-2ce6a27d54a5083f40751babba9d77c5529f0515cd5f8dc31a379406d52ff6cdR38 and I linked the reuest there as you asked.

Just making very clear that this has nothing to do with _WalletConnect_. _WalletConnect_ has many things, one of them is a transport layer for remote wallets.

Our SDKs use the transport layer and support CAIPs first class. But many browser extension wallets use CAIPs without using our transport layer (even though some do). Our SDKs will also support other transport layers.

EIPS/eip-5792.md
jxom: What do you mean by “context switch”? 

`eth_getCapabilities` should mostly be a synchronous call just like `eth_accounts`. CAIP-25 is an orthogonal abstraction to this, that said,  consumers + wallets using CAIP-25 could include `eth_getCapabiltities` in their required/optional methods.

EIPS/eip-5792.md
jxom: What if the wallet doesn’t support `eth_sendUserOperation`, but supports sending a paymaster via `wallet_sendTransaction`.

Detecting capability support via RPC methods seems like a leaky abstraction to me.

EIPS/eip-5792.md
jxom: What do you mean by “context switch”?

`eth_getCapabilities` should mostly be a synchronous call just like `eth_accounts`. CAIP-25 is an orthogonal abstraction to this, that said, consumers + wallets using CAIP-25 could include `eth_getCapabilities` in their required/optional methods.

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+      "supported": true
+    },
+    "sessionKeys": {
+      "supported": true
+    }
+  },
+  "0x14A34": {
+    "paymasterService": {
+      "supported": true
+    }
```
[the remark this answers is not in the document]
wilsoncusack: As in you're concerned about bytes on the wire? I think it's worth it for clarity. Also at one point I think we had a description field here @lukasrosario ?

EIPS/eip-5792.md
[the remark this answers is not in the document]
wilsoncusack: > maybe we allow for both -- i.e. without a chain id to get all chains, and with a chain id to get capabilities for just the request chain id.

I think I would prefer to be more opinionated. @moodysalem is your main concern just how chainID is formatted? Isn't this already fairly standardized for other requests?

EIPS/eip-5792.md
McOso: Anything capabilities or permissions related should not be on this EIP. The Permission Standard is still in discussion and this should wait for that to move forward, since this seems like it would be dependent on the outcome of the Permission Standard. This was originally an EIP for batched function calls, imho it should stay that way, unless you can find a way to remove the capabilities, for the time being. I think it would be best to wait for Permissions Standard to move forward first.

EIPS/eip-5792.md
Jrachman: gm gm sers! Batched `AUTHCALL`s are definitely doable here. Happy to collaborate where I can as a rep for 3074.

EIPS/eip-5792.md
wilsoncusack: @McOso what permissions standard are you referring to?

If the one Derek and others are working on, we became aware of that after we had already published our plans here. I think we have different goals and designs, and I don't think it's fair to ask us to wait on something someone else is working on which is independent of what we're doing :)  

If we can find a common best path that's great. But doesn't seem to be the case so far.

If you are sincerely concerned on the change to this EIP, the alternative path was a new proposal with these same RPCs. We took over this EIP because of the similarity in the RPCs and Moody asked us to. Otherwise this EIP is stagnant and nothing is going to happen with it. 

We should not solve for batch transactions as a one off. We need to fix several things here.

EIPS/eip-5792.md
wilsoncusack: Thanks. So is the rpc provider_authorize?

EIPS/eip-5792.md
jxom: The outcome of any "permission standard" EIP would fit into this type of method, no?

IIRC, the goal of the "permission standard" is to define a standard for a consumer to **asynchronously request permissions**,
Whereas, the goal of this method is to define a standard for a consumer to **synchronously discover capabilities**.

If there is a way to synchronously discover capabilities in the permissions standard discussions, then I am all ears and maybe these efforts should be combined!

arein: This standard needs no mention of the WIP session key/permissions standard @McOso is referring to.

For session keys the dApp will send the UserOps/Transactions. They can still use these new RPC methods to do so.

In fact I see these two things in harmony.

The [UserOpBuilder ERC](https://zerodev.notion.site/ERC_UserOpBuilder-753e73db33bd4c8c8a9fbeeaf160a27c?pvs=4) may deserve a mention here. But since it's just going into draft this week it's probably better not to.

danfinlay:
wallet_getCapabilities is a fingerprinting vector. The modern browser landscape is riddled with identity leaks, and each distinguishing characteristic makes this worse, and developers of browser standards should be vigilant about not adding new fingerprinting surface unless absolutely necessary.

Providing a paymaster wouldn’t require feature detection anyway: It could just be an optional parameter.

In addition to privacy concerns, the proposed wallet_getCapabilities return type does not account for a wallet that has support for multiple account types. For example, since atomic batching capability may vary account by account (as it already can in MetaMask Flask, which supports account plugins), and may not be a global wallet characteristic, any atomic batching detection must be done by account.

I think the design goals for the batching feature can be achieved without this kind of privacy-leaking feature-declaration. An alternate approach could be to add some type definition parameters to wallet_getPermissions, as described in [EIP-2255](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2255.md). That EIP was even written with that type of use in mind. This allows a Ðapp to request required account features at connection time, ensuring that any account the wallet allows the user to connect is one that is compatible with the site.

jxom:
@danfinlay – yeah, I think `eth_getCapabilities` is definitely the most contentious concept of this proposal, and I have been thinking maybe it shouldn't be a concern of this proposal as I believe it can still function without (ie. paymaster as an optional param). Did you have any strong opinions there @lukasrosario + @wilsoncusack?

wilsoncusack:
Hey @danfinlay, thanks for the feedback! 

I see a difference between `capabilities` and `permissions`, and I wonder if you do, too. I personally don't think a user is interested in or has the information required to screen things like `atomicBatch` and `paymasterService`. As a user, I would be annoyed to have to check boxes here. 

Re the fingerprinting concern, I think the info exposed is pretty minimal compared to exposing your account address. I know you also prefer the latter isn't done, but the reality is that's how most apps work today, and I think it's out of scope to litigate that here. Would you feel better if we made it explicit that `wallet_getCapabilities` should not return anything until after some initial connection has been made, e.g. via `eth_requestAccounts`?

As [I said](https://github.com/ethereum/EIPs/pull/8361#discussion_r1545491324), I am open to the possibility that we should instead be proposing some like `wallet_connect` in which capabilities and accounts are shared, but that seems like it could be an EIP unto itself and I think it maybe has more direct conflicts with the permissions approach you mention? 

> since atomic batching capability may vary account by account 

This is a good point and maybe fixing this is the easy way to require the user has already connected: perhaps `getCapabilities` should take an account? Or return on a per account basis, for all connected accounts. 

Re the question of if this proposal could work without `getCapabilities` (@jxom), I think it could but it's a worse API. E.g. an app like Uniswap may have different call flows based on the `atomicBatch` capability (permit + swap vs. approve + swap). I think it's a bummer if `sendCalls` takes capabilities but the app has little to no knowledge as to which are supported. 

Thanks again for taking time to review!

EIPS/eip-5792.md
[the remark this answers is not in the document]
lukasrosario: yeah i think the clarity is worth it here, savings would be pretty minimal. no description field here, dont see the value in that since this is just for wallet <> app communication. maybe misunderstanding though so lmk if you think we do need something else here

danfinlay:
> I see a difference between capabilities and permissions, and I wonder if you do, too.

Without getting into the way that the word "capabilities" is used in different contexts as either "feature detection" or very much [as permission](https://en.wikipedia.org/wiki/Capability-based_security), no I am not suggesting that we treat each feature to detect as a user-facing permission.

When [I said](https://github.com/ethereum/EIPs/pull/8361#issuecomment-2029496535) that [EIP-2255](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2255.md) provided a solution to this problem, I did not mean adding more permissions. I was referring to the part of EIP-2255 that is specifically about parameterized permissions requests. If you refer to [the first code example of the EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2255.md), it actually demonstrates an example of exactly what we're talking about: Requesting an account that supports a specific method:

```
provider.request({
  method: 'requestPermissions',
  params: [
    {
      'eth_accounts': {
        requiredMethods: ['signTypedData_v3']
      }
    }
  ]
});
```

A wallet receiving this request would only allow the user to select a wallet that supported `signTypedData_v3` (something that was desirable at the time, as some DEXes had adopted it earlier than some hardware wallet companies). You could just as easily imagine `wallet_sendBatch`, or _some new parameters_. That options bag on the permission was designed to be extended with _specifiers_ enabling the requester to narrow the valid responses to the request, which is what you want here.

By doing this, you're able to:
- Never bother the user with permissions related to feature detection
- Never leak user data related to what features their wallet supports until the user makes a _related_ permissions request.

> Would you feel better if we made it explicit that wallet_getCapabilities should not return anything until after some initial connection has been made, e.g. via eth_requestAccounts?

Kinda, but I think that would be annoying! Why make users connect an account just to then tell them the account isn't compatible? With the parameterized permissions request approach, we can guarantee that a connected account is a compatible account, all while also not leaking any _other_ information about the connected account!

> Re the fingerprinting concern, I think the info exposed is pretty minimal compared to exposing your account address.

The `wallet_getCapabilities` method is exposed to dapps _before_ a user connects their address, and that's _particularly terrible_. It may be not as bad as an address, but *at least the user chose to disclose their address!* Bumping the capability detection to after the account connection only creates the unpleasant experience described above.

Anyways, I hope that clears things up, it sounds like we had a miscommunication. I think there's a way to basically have your intended ideal UX while minimizing data leakage, which should be a serious priority for any browser standard.

wilsoncusack:
Thanks @danfinlay! I read the proposal and looked at that code sample but tbh didn't quite follow that meaning. This is interesting. We'll discuss and come back!

wilsoncusack:
@danfinlay I'm not totally sure how this would work. You have 
```
provider.request({
  method: 'requestPermissions',
  params: [
    {
      'eth_accounts': {
        requiredMethods: ['signTypedData_v3']
      }
    }
  ]
});
```

But here we are talking capabilities, not methods, so I think I'd want like 
```
provider.request({
  method: 'requestPermissions',
  params: [
    {
      'eth_accounts': {
        requiredCapabilities: ['paymasterService', 'atomicBatch]
      }
    }
  ]
});
```

But then, I also don't want to `require` these, I just want to know if the wallet supports and react?

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
-##### `wallet_getBundleStatus` Example Parameters
+type GetCallsResult = {
+  status: 'PENDING' | 'CONFIRMED'
+  receipts?: {
+    logs: {
+      address: `0x${string}`
+      data: `0x${string}`
+      topics: `0x${string}`[]
+    }[]
+    success: boolean
```
jxom, line 164: Should we avoid flags like this? Should we align with the `status` in the receipt schema in the execution specs? https://github.com/ethereum/execution-apis/blob/main/src/schemas/receipt.yaml#L110-L113

wilsoncusack: Aligning with status spec seems good. @lukasrosario?

lukasrosario: sgtm. will update here

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+The key of each item per chain is the name of a capability and the value can be of any shape. Capabilities are returned per chain because wallets may support different capabilities across chains.
+
+We expect the community to align on the definition of capabilities in separate ERCs.
+
+#### `wallet_getCapabilities` RPC Specification
+
+```typescript
+type GetCapabilitiesParams = [];
+
+type GetCapabilitiesResult = Record<`0x${string}`, <Record<string, any>>; // Hex chain id
```
jxom, line 46: We should be avoiding the `any` type here. Ideally it should be narrowed, and if not, `unknown`.

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+#### `wallet_sendCalls` RPC Specification
+
+```typescript
+type SendCallsParams = {
+  version: string;
+  chainId: `0x${string}`; // Hex chain id
+  from: `0x${string}`;
+  calls: {
+    to: `0x${string}`;
+    data: `0x${string}`;
```
jxom, line 98:
```suggestion
    data?: `0x${string}` | undefined;
```

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+    value: `0x${string}`; // Hex value
+  }[];
+  capabilities: Record<string, any>;
```
jxom, line 101:
```suggestion
  capabilities?: Record<string, any> | undefined;
```

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+
+```typescript
+type SendCallsParams = {
+  version: string;
+  chainId: `0x${string}`; // Hex chain id
+  from: `0x${string}`;
+  calls: {
+    to: `0x${string}`;
+    data: `0x${string}`;
+    value: `0x${string}`; // Hex value
```
jxom, line 99:
```suggestion
    value?: `0x${string}` | undefined; // Hex value
```

EIPS/eip-5792.md
```diff
@@ -1,390 +1,269 @@
...
+
+#### `wallet_sendCalls` RPC Specification
+
+```typescript
+type SendCallsParams = {
+  version: string;
+  chainId: `0x${string}`; // Hex chain id
+  from: `0x${string}`;
+  calls: {
+    to: `0x${string}`;
```
jxom, line 97:
```suggestion
    to?: `0x${string}` | undefined;
```

can deploy contract?

danfinlay:
> I also don't want to require these, I just want to know if the wallet supports and react?

I'm not against feature detection for connected accounts. If that suits your use case, have at it! I could definitely imagine a case where someone might _require_ atomicity, but if that's not worth your standards time, then I wouldn't bother with it.

Why not just use existing methods, like (pseudocode)?:
```
try {
  provider.sendBatch(batch)
} catch (err) {
  if (err.message === METHOD_NOT_FOUND) { 
    batch.forEach(provider.send);
  }
}
```

My top complaint is 100% leaking user-agent features to every site before a single user act of consent. If the feature detection is coupled to _the relevant object of consent_, then I don't really care one way or the other.

Even looser held opinion, not remotely a blocker: I'd probably prefer a method where the dapp asks about methods, and doesn't get a blank descriptor, though.

lightclient (approved):

PR #8345, opened 2024-03-21
Update EIP-7044: Move to Final

timbeiko:
Moves EIP-7514 to Final, links the test cases and adds security considerations.

PR #8362, opened 2024-03-29
Update EIP-7044: Move to Final

lightclient:
Should have moved to final in #8345, but looks like somewhere along the line of rebasing we lost it.


lightclient:
@dapplion already approved #8345, just going to go ahead and merge this one.

PR #8368, opened 2024-03-31
Create eip-7653.md

vbuterin:
Require the bloom filters in an execution block, including at the top level and in the receipt object, to be empty.

### DIFF
--- a/EIPS/eip-7653.md
+++ b/EIPS/eip-7653.md
@@ -1,0 +1,45 @@
+---
+eip: XXXX
+title: Remove bloom filters
+description: Remove bloom filters from the execution block
+author: Vitalik Buterin (@vbuterin)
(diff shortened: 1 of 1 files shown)


EIPS/eip-7653.md
```diff
@@ -0,0 +1,45 @@
+---
+eip: 7653
```
lightclient, line 2:
```suggestion
eip: 7668
```

lightclient (approved):

QBlockQ:
Thanks. Removing bloom filters from execution blocks and receipt objects is a significant change. It seems aimed at optimizing data storage and retrieval efficiency on the Ethereum blockchain, which is crucial for scalability. Could you elaborate on how this might affect the average transaction processing time? Also, how might developers need to adjust their dApps to accommodate this change? Understanding these aspects can help the community better prepare for the transition.

PR #8374, opened 2024-04-01
Add EIP: EXTSELFDESTRUCT Opcode

parithosh:

--

 - The build passes.


EIPS/tbd.md
```diff
@@ -0,0 +1,112 @@
+---
+eip: 7777
```
lightclient, line 2:
```suggestion
eip: 7665
```

emmanuel-awosika:
Opcode sounds nice, but it would even be better to have a system-level contract--so we can emit something like a `fatality()` event each time a contract is cleared from the state trie. Some other questions related to this proposal:

1) Do contract developers just get points and nothing else? Shouldn't we be thinking about financial incentives, like issuing ETH based on the number of contracts cleared? 

2) The conclusion says "Mechanisms to prevent abuse, such as a validation process for EXTSELFDESTRUCT eligibility, are essential." A hack-y solution is to disallow contracts from interacting with the `target` contract and calling `EXTSELFDESTRUCT` on the target contract in the same block. That way, contracts are disincentivized from triggering (spam) operations that create new storage slots in the target contract in order to clear out contracts whose storage is within bounds. This should protect contract developers that are conscientious about optimizing storage from bad actors. 

How will this work? A requirement to call `EXTSELFDESTRUCT` (after interacting with the target contract) in the next block means anyone can clear out the contract once the storage hits the limit. Since we assume that an attacker bloating the `target` contract's storage wants to accumulate points for itself, this reduces incentive to game the system (anyone can call `EXTSELFDESTRUCT` before it does). 

> The mnemonic MURDER was additionally considered as a possibility, however, it was found to not sufficintly align with the Ethereum ethos of positivity and cooperation.

Good vibes only?

SamWilsn:
Closing because I'm _pretty sure_ this is an April fools joke.

PHOENIXENT (commented):
image

PR #8377, opened 2024-04-01
Update EIP-2255: Fix code examples

danfinlay:
Fixed name of method.


SamWilsn (commented):
Should line 137 also be updated?

Scamreno:
Yes I would laugh PR forward my account information

Scamreno (commented):
Just trying to get some of my assets funded to my bank account because I really need it

PR #7996, opened 2023-11-29
Update EIP-5792: Move to Review

arein:
**ATTENTION: ERC-RELATED PULL REQUESTS NOW OCCUR IN [ETHEREUM/ERCS](https://github.com/ethereum/ercs)**

--

When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #8387, opened 2024-04-04
Update EIP-5792: Editorial updates to EIP-5792

bumblefudge:
No changes to any features or behaviors here, aside from side effects and error-handling and partial-information cases.

That said, because there are a lot of interested parties here, I broke the changes into various commits that could be cherry-picked out in case any of them don't get consensus.

Overview, commit-by-commit:

- Some [general cleanup](https://github.com/ethereum/EIPs/commit/3a74178a19125866e60c5d73c561a701a6f5a3bb), like making shoulds and mays into `SHOULD`s and `MAY`s
- A [two-commit](https://github.com/ethereum/EIPs/commit/cbef48f5bdc9236cb8649ba9545d01d9a1a0e61d) [reorg](https://github.com/ethereum/EIPs/commit/abb6dc056bc7a5c5fded0ee09255e922ba31cf2c) to move the feature discovery section AFTER the titular call-batching methods, tweak headers, and make the `atomicBatch` capability a little more explicitly specified, including an anti-fingerprinting behavior (null case versus false case - see final commit)
- A tiny commit to [add a mention of supplemental ways of expressing/querying capabilities](https://github.com/ethereum/EIPs/commit/14893337c06cb0e66874164befe00a5a6cc68230) other than the novel capability RPC (especially useful during the interim until critical mass of wallets start implementing it)
- A commit [adding derek rein as co-author](https://github.com/ethereum/EIPs/commit/a37e0d63b22cfd51c493984f54665d5fb2c98c7c), since the previous commit was first elaborated [in another PR](https://github.com/ethereum/EIPs/pull/7996) and worked out in dialogue with him and his team
- A commit mentioning [another capability discovery variant already present in the wild](https://github.com/ethereum/EIPs/commit/f9d78ef24244313daf8639b2b5ebfed85741c6ae)
- A commit adding a DRAFT [privacy considerations section](https://github.com/ethereum/EIPs/commit/33a808347bc306ff51bc9b6056a55cb7cd37c0dd) crudely summarizing @danfinlay 's comments on #8361 which I hope he will workshop


EIPS/eip-5792.md
```diff
@@ -2,7 +2,7 @@
 eip: 5792
 title: Wallet Call API
 description: Adds JSON-RPC methods for sending multiple calls from the user's wallet, and checking their status
-author: Moody Salem (@moodysalem), Lukas Rosario (@lukasrosario), Wilson Cusack (@wilsoncusack), Dror Tirosh (@drortirosh), Jake Moxey (@jxom)
+author: Moody Salem (@moodysalem), Lukas Rosario (@lukasrosario), Wilson Cusack (@wilsoncusack), Dror Tirosh (@drortirosh), Jake Moxey (@jxom), Derek Rein (@arein)
```
arein, line 5: grazie

arein (approved):

EIPS/eip-5792.md
wilsoncusack: Sorry, I'm confused. Did Derek write this PR or what's the idea here? Love Derek but there are a lot of people who have contributed to discussion at the same level who we haven't added.

EIPS/eip-5792.md
```diff
@@ -12,61 +12,23 @@ created: 2022-10-17
...
 
-The current methods used to send transactions from the user wallet and check their status do not meet modern developer demands and cannot accomodate new transaction formats. Even the name–`eth_sendTransaction`–is an artifact of a time when nodes served as wallets. 
+The current methods used to send transactions from the user wallet and check their status do not meet modern developer demands and cannot accomodate new transaction formats. Even the name–- `eth_sendTransaction`-– is an artifact of a time when nodes served as wallets. 
 
-Today, developers want to send batch transactions, which many smart accounts can support atomically. Developers also want to make use of features afforded by new transaction formats, like paymasters in [ERC-4337](./eip-4337.md) transactions. `eth_sendTransaction` offers no way to do these things. 
+Today, developers want to send batched transactions, which many smart accounts can receive and execute atomically. Developers also want to make use of features afforded by new transaction formats, like paymasters in [ERC-4337](./eip-4337.md) transactions. `eth_sendTransaction` offers no way to do these things.
```
wilsoncusack, line 23: Should be batched calls, actually. Atomic batched transactions are more like what Flashbots support than what smart wallets support.

EIPS/eip-5792.md
```diff
@@ -12,61 +12,23 @@ created: 2022-10-17
 
 ## Abstract
 
-Defines new JSON-RPC methods which enable apps to ask a wallet to process a batch of onchain write calls and to check on the status of those calls. Calls can be specified to make use of various wallet-defined capabilities, and there is also a new RPC to enable apps to discover which capabilities a wallet supports. 
+Defines new JSON-RPC methods which enable apps to ask a wallet to process a batch of onchain write calls and to check on the status of those calls. Dapps can specify that these onchain calls be executed taking advantage of specific capabilities previously expressed by the wallet; an additional, novel wallet RPC is defined to enable apps to query the wallet for those capabilities.
```
wilsoncusack, line 15: Calls are not necessarily made by a "dapp"

EIPS/eip-5792.md
```diff
@@ -207,23 +169,66 @@ As with the return value of `wallet_sendCalls`, the batch identifier may be any
 }
 ```
 
-### `atomicBatch` Capability
+### `wallet_getCapabilities`
+
+An app calling the above-defined methods without prior knowledge that the wallet supports them risks degrading user experience considerably, so an additional RPC for querying the wallet's capabilities is provided.
```
wilsoncusack, line 174: Capabilities is not for discovering which RPCs that wallet supports.

EIPS/eip-5792.md
```diff
@@ -207,23 +169,66 @@ As with the return value of `wallet_sendCalls`, the batch identifier may be any
...
 
-Indicates that a wallet can execute multiple calls atomically as part of a single transaction.
+In addition to, or instead of, querying the wallet directly for capabilities, the same capability objects could be expressed out-of-band (such as in the `sessionProperty.capabilities` object persisted in a CAIP-25-conformant wallet provider interface) or in a well-known location (such as a URL derived from an [EIP-6963](./eip-6963.md) `rdns` identifier). Provider abstractions may also cache capabilities from previous requests or otherwise inject them from out-of-band. If querying multiple ways leads to different results, or to override caches or provider injections, direct RPC communication with the wallet SHOULD always be taken as the canonical and current expression of capabilities.
```
wilsoncusack, line 176: Does CAIP-25 have a field for this? Or expressly allow new sets of fields defined by the wallet?

EIPS/eip-5792.md
```diff
@@ -207,23 +169,66 @@ As with the return value of `wallet_sendCalls`, the batch identifier may be any
...
+##### `atomicBatch` Capability Specification
 
 ```typescript
 type AtomicBatchCapability = {
   supported: true
 }
 ```
 
-##### `wallet_getCapabilities` Example Return Value
+The only valid JSON-RPC values for this capability are `true` or `false`; if a returned value for "supported" value is typed as a string or number, it SHOULD be treated as malformed.
```
wilsoncusack, line 227: Why? What if I want a capability like "offChainFunding" and want to provide a URI where info can be fetched?

Gemrav (approved):

EIPS/eip-5792.md
wilsoncusack: Definitely not opposed to adding more authors for contributions at author level! Just want to have a clear and fair criteria so we don't end up with like 20 authors.

EIPS/eip-5792.md
wilsoncusack: Or maybe you mean, which capabilities the wallet supports?

EIPS/eip-5792.md
wilsoncusack: e.g. would this conform to the spec? 
> Respondent SHOULD ignore and drop from its response any properties not defined in this document or in another CAIP document extending this protocol which the respondent has implemented in its entirety; similarly, the requiredScopes, optionalScopes, and sessionScopes arrays returned by the respondent SHOULD contain only valid [CAIP-217](https://chainagnostic.org/CAIPs/caip-217) objects, and properties not defined in [CAIP-217](https://chainagnostic.org/CAIPs/caip-217) SHOULD also be dropped from each of those objects.

EIPS/eip-5792.md
wilsoncusack: Tbh I think the whole thing here reads less clearly :)

EIPS/eip-5792.md
bumblefudge: See commit breakdown in opening comment-- this was to close #7996 creditfully.  i can cherrypick out that commit if the other authors don't think 8387 (the PR we're in) merits a co-author cred.

EIPS/eip-5792.md
bumblefudge: good catch! incorporated in [this commit](https://github.com/ethereum/EIPs/pull/8387/commits/956c6ded6dbecc83e57209dacfe6b9863b2e9e77)

EIPS/eip-5792.md
bumblefudge: You're right about dapps-- let's fix that whatever replaces this line.

Sorry for the torturous syntax-- it's been my experience, though, that people allocate a number of minutes they think a doc will take to read and start skimming near the end, so without an abstract that warns them there's a whole section on capability expression near the end, they might not allocate the time/mental bandwidth to get that far before thinking they got the gist.  i'm not wed to the specific language, but AM wed to the abstract signaling that whole section.

How's this, at the other end of the verbosity spectrum from the previous proposal (any co-author drops an emoji thumb and second thumb-up can just commit the suggestion?):

```suggestion
Defines a new JSON-RPC method for apps to send batches of onchain write calls together, which returns a pointer to the execution state, as well as a second method to query that state.  A third application-to-wallet method is included for requesting the capability to use the first two.
```

EIPS/eip-5792.md
bumblefudge: very good feedback.  resolving because of [this commit](https://github.com/ethereum/EIPs/pull/8387/commits/4e7cb461c846050681a322a4bb738ed29c969034) but feel free to comment again if the new prose is just as off-base in another direction!

EIPS/eip-5792.md
bumblefudge: Mmm, reading this as a diff may be obscuring the meaning (or it might be badly written) , i was referring specifically to the `atomicBatch` boolean, not to all other possible object capabilities expressable in this RPC-or-its-equivalents.  any help making this text link that otherwise arbitrary constraint to the privacy considerations below would be appreciated, but I was trying not to overload this section with explanation.

bumblefudge: it's also totally fine to `debate if this should be more tightly specced` in the first place, as i well appreciate that i'm kind of the OCap hardliner in this conversation and also not a co-author 😅

EIPS/eip-5792.md
bumblefudge: > Does CAIP-25 have a field for this? Or expressly allow new sets of fields defined by the wallet?

there is, in fact, a property for exactly this:

> A third object is the sessionProperties object, all of whose properties MUST be interpreted as optional, since requesting applications cannot mandate session variables to providers. Because they are optional, providers MAY respond with all of the requested properties, or a subset of the session properties, or no sessionProperties object at all; they MAY even replace the values of the optional session properties with their own values.

The semantics and syntax of *specific* members of the `sessionProperties`, such as `sessionProperties.capabilities`, are specified by future CAIPs, in the same way that the semantics and syntax additional capability object are expected to be specified in future EIPs.  Ideally, 1:1 equivalence and translatability between the two should be preserved, which is why there aren't any CAIPs defining these yet 😉

bumblefudge: (in the meantime, experimental implementations are breaking one SHOULD, but no MUSTs)

wilsoncusack: Ok, but I think this leaves us in an odd state: we are saying 
> such as in the `sessionProperty.capabilities` object persisted in a CAIP-25-conformant wallet
and then also that 
> The semantics and syntax of specific members of the sessionProperties object, such as sessionProperties.capabilities, are expected to be specified by future CAIPs

So isn't it the case that, today, one could not properly follow CAIP-25 and include `sessionProperty.capabilities` ?

EIPS/eip-5792.md
wilsoncusack: Oh ok sorry, yeah got confused by the diff. That is helpful clarity.

wilsoncusack: > i'm kind of the OCap hardliner in this conversation 

Happy to hear the steelman for more tightly specifying :)

EIPS/eip-5792.md
wilsoncusack: Ah I had not seen this PR, thanks! However it seems that, in that context, CAIP-25 is mentioned for discovering these RPCs? 

> ### Advertising support via [CAIP-25]

Whereas the debate here has been whether CAIP-25 should be used to discover capabilities? The strongest version of the argument being the argument that `wallet_getCapabilities` should not be present.

EIPS/eip-5792.md
```diff
@@ -171,9 +171,9 @@ As with the return value of `wallet_sendCalls`, the batch identifier may be any
 
 ### `wallet_getCapabilities`
 
-An app calling the above-defined methods without prior knowledge that the wallet supports them risks degrading user experience considerably, so an additional RPC for querying the wallet's capabilities is provided.
+An app calling the above-defined methods without prior knowledge that the wallet supports them risks degrading user experience considerably, yet overtly or indirectly querying the wallet for interfaces risks eroding the anonymity of the user-agent (see Privacy Considerations below). The semantics of requesting a capability supports security and privacy by treating unsupported features and unpermitted features the same way, reducing fingerprinting surface. An exemplary wallet RPC is defined below for applications to request the capability to call the RPC methods above, assuming the wallet will process them atomically if the wallet grants that capability.
```
wilsoncusack, line 174: > An app calling the above-defined methods without prior knowledge that the wallet supports them risks degrading user experience considerably

Why? The call should fail more or less synchronously and we've defined fallbacks. 

> yet overtly or indirectly querying the wallet for interfaces risks eroding the anonymity of the user-agent (see Privacy Considerations below)

Disagree that this is a substantive concern. 

> An exemplary wallet RPC is defined below for applications to request the capability to call the RPC methods above

Are you referring to wallet_getCapabilities? A wallet could possible implement wallet_sendCalls and not wallet_getCapabilities. It sounds like you're saying `wallet_getCapabilities` is used to detect `wallet_sendCalls` support?

EIPS/eip-5792.md
bumblefudge: well i was waiting for this EIP to be stable to push that CAIP to draft, but you could CONFORMANTLY follow CAIP-25 as either a wallet or a dapp putting anything you want in `sessionProperties`. I wouldn't advice people do it in prod until there's a CAIP they trust their counterparties to also implement, that's kind of the proleptic temporality of EIP adoption that led us to such proactive providers in the first place, isn't it 😏

EIPS/eip-5792.md
bumblefudge: "exemplary" means this RPC or equivalent should trump any less-direct way of getting the capability, including discovering the feature without requesting the cap.  you asked for an OCap steelman, so i have to do my best mark miller impersonation until dan f shows up!

wilsoncusack: Does a failure require context switching? Wouldn't the user just never see the notification on their phone or whatever?

EIPS/eip-5792.md
wilsoncusack: Roger. Would it make sense to get that CAIP-25 update in and then we can include it more explicitly here as an alternative means of discovery?

EIPS/eip-5792.md
jxom: If a wallet doesn't support an RPC method, then it should synchronously throw a 4200 Unsupported Method error.

EIPS/eip-5792.md
jxom, line 15: Can we explicitly call out the method (feel like a reader just wants to get straight to the point instead of reading too much preamble)?

```suggestion
Defines a new set of JSON-RPC methods that enable Wallet Consumers to request to send bundles of onchain write calls: 

- `wallet_sendCalls`: a method to send a bundle of onchain write calls that returns a reference identifier.
- `wallet_getCallsReceipt`: a method to extract receipts from the sent calls.
- `wallet_getCapabilities`: a method to discover capabilities that a wallet supports.
```

bumblefudge:
```suggestion
Defines a new set of JSON-RPC methods that enable Wallet Consumers to request to send bundles of onchain write calls: 
- `wallet_sendCalls`: a method to send a bundle of onchain write calls that returns a reference identifier.
- `wallet_getCallsReceipt`: a method to extract receipts from the sent calls.
- `wallet_getCapabilities`: a method to request capabilities that a wallet supports.
```
I love it, with only one word changed to honor my eccentric steelman proclivities changed, i'd love it even more (see last line)

jxom: Does "request" sound like is asynchronous? As in, you need to switch to wallet context (which you don't)?

EIPS/eip-5792.md
bumblefudge: oh I would not make that argument at all; I thought [i made the exact opposite argument a few lines below](https://github.com/ethereum/EIPs/pull/8387/files#diff-7a494500c37dc4b57613207d62874322198a4ced215f4a0457a0af908bb6364eR176). I think wallet RPCs should trump any three-party trust model and capability models should trump discovery models, but it's hard to thread that needle through today's business incentives and coordination landscape.

EIPS/eip-5792.md
bumblefudge: well, if by "more explicitly" you mean normatively refer, I don't think it's a good idea, since we're explicitly telling people it's supplemental and _not an alternative or a replacement_ to the wallet_RPC, so the reference should only be NON-normative (and thus linking or not linking doesn't matter).  it's already happening in the world, and you have to name to mark it as supplemental (especially if you'd like to deprecate it later).

i also don't love the idea of removing the reference to it (or the reference to provider-caching, more importantly, which I think wallet devs and dapp devs alike should be given a heads-up about, honestly).

bumblefudge: (and technically, it wouldn't be a CAIP-25 update, it was be a novel _extension_ CAIP that defines sessionProperties.capabilities. I suppose I could tho, if this EIP isn't going to specify anything more concrete, syntactically or semantically, than `The key of each item per chain is the name of a capability and the value can be of any shape. Capabilities are returned per chain because wallets may support different capabilities across chains.`.  It'll be the shortest CAIP yet, until future EIPs add more sub-classes!)

EIPS/eip-5792.md
bumblefudge: oh, sorry, I wasn't thinking in terms of runtime sync semantics, I was just thinking about the nitpicky distinction between requesting a capability (which is both feature and permission to use it) and discovering a feature.  the hair-splitting has to do with the semantics of NULL responses (the wallet might SUPPORT the feature but still send back no capability for permission or trust-level reasons, as Dan keeps saying).

EIPS/eip-5792.md
wilsoncusack: Ok. I think I'm too smooth brain to follow :) 

But it seems like what is being discussed here is not to do with that prior PR? Or what's the part that was closely lifted?

arein (approved):

EIPS/eip-5792.md
bumblefudge: > Whereas the debate here has been whether CAIP-25 should be used to discover capabilities? The strongest version of the argument being the argument that wallet_getCapabilities should not be present.

I am NOT making the strongest version of the CAIP-25 maxi argument, even tho I have been working on CAIP-25 primarily for two years.  But I _AM_ making a strong argument for including non-normative mention of provider caching and provider-facilitated opt-in declaration of capabilities. subtle distinction, you're not smoothbrained for getting lost it's probably because I've been focused on this for 2 years that I can't explain it to a normal person.

EIPS/eip-5792.md
jxom: "discover permitted capabilities"?

bumblefudge: "query for permitted capabilities"?

bumblefudge: (if it's this hard to wordsmith maybe it's not a distinction worth making!)

jxom: anything but "request" :)

PR #8388, opened 2024-04-04
Update EIP-5450: Align with EOF megaspec

gumb0:
Stack validation algorithm is updated to be less restrictive.

### DIFF
--- a/EIPS/eip-5450.md
+++ b/EIPS/eip-5450.md
@@ -28,8 +28,6 @@ The operand stack validation provides several benefits:
 
 - removes the run-time stack underflow check for all instructions,
 - removes the run-time stack overflow check for all instructions except `CALLF` and `JUMPF` (`JUMPF` introduced in a separate EIP) ,
-- allows for multiple forward jump instructions executing at different stack heights to target the same instruction; the stack of target basic stack is validated for all possible heights,
-- does not allow backwards jump instructions to target an instruction, that can be executed with different stack height; this prevents deployment of the loops with unbounded stack pushing or popping,
 - ensures that execution terminates with one of the terminating instructions,
 - prevents deployment of code with unreachable instructions, thereby discouraging the use of code sections for data storage.
 
@@ -111,6 +109,18 @@ The computational and space complexity of this pass is *O(len(code))*. Each inst
 Given the deploy-time validation guarantees, an EVM implementation is not required anymore to have run-time stack underflow nor overflow checks for each executed instruction. The exception is the `CALLF` and `JUMPF` performing operand stack overflow check for the entire called function.
 
 ## Rationale
+
+### Properties of validated code
+
+Any code section validated according to operand stack validation has the following properties:
+
+1. There are no unreachable instructions
+2. There are no instructions reachable only via backwards jump
+3. Operand stack underflow cannot happen.
+4. Operand stack overflow can only happen at `CALLF` or `JUMPF` instruction.
+5. Multiple forward jump instructions executing at different stack heights may target the same instruction; the stack of target basic block is validated for all possible heights.
+6. Any backwards jump instruction can only target an instruction that is executed with equal stack height; this prevents deployment of the loops with unbounded stack pushing or popping.
+7. Final instruction in the code section is either terminating instruction or `RJUMP`.
 
 ### Stack overflow check only in CALLF/JUMPF
 
--- a/EIPS/eip-5450.md
+++ b/EIPS/eip-5450.md
@@ -27,7 +27,7 @@ and preventing the execution and deployment of any invalid code.
 The operand stack validation provides several benefits:
 
 - removes the run-time stack underflow check for all instructions,
-- removes the run-time stack overflow check for all instructions except `CALLF` and `JUMPF`,
+- removes the run-time stack overflow check for all instructions except `CALLF` and `JUMPF` (`JUMPF` introduced in a separate EIP) ,
 - allows for multiple forward jump instructions executing at different stack heights to target the same instruction; the stack of target basic stack is validated for all possible heights,
 - does not allow backwards jump instructions to target an instruction, that can be executed with different stack height; this prevents deployment of the loops with unbounded stack pushing or popping,
 - ensures that execution terminates with one of the terminating instructions,
@@ -62,6 +62,8 @@ In the second validation phase control-flow analysis is performed on the code.
 
 - ending function execution: `RETF`, `JUMPF`, or
 - ending whole EVM execution: `STOP`, `RETURN`, `RETURNCONTRACT`, `REVERT`, `INVALID`.
+
+*note: `JUMPF` and `RETURNCONTRACT` are introduced in separate EIPs.*
 
 Instructions in the code are scanned in a single linear pass over the code. For each instruction the operand stack height bounds are recorded as `stack_height_min` and `stack_height_max`.
 
--- a/EIPS/eip-5450.md
+++ b/EIPS/eip-5450.md
@@ -63,6 +63,9 @@ In the second validation phase control-flow analysis is performed on the code.
 
 *note: `JUMPF` and `RETURNCONTRACT` are introduced in separate EIPs.*
 
+*Forward jump* refers to any of `RJUMP`/`RJUMPI`/`RJUMPV` instruction with relative offset greater than or equal to 0. *Backwards jump* refers to any of `RJUMP`/`RJUMPI`/`RJUMPV` instruction with relative offset less than 0, including jumps to the same jump instruction.
+
+
 Instructions in the code are scanned in a single linear pass over the code. For each instruction the operand stack height bounds are recorded as `stack_height_min` and `stack_height_max`.
 
 The first instruction's recorded stack height bounds are initialized to be equal to the number of inputs to the function type matching the code (`stack_height_min = stack_height_max = type[code_section_index].inputs`).
--- a/EIPS/eip-5450.md
+++ b/EIPS/eip-5450.md
@@ -67,7 +67,8 @@ The first instruction's recorded stack height bounds are initialized to be equal
 
 For each instruction:
 
-1. **Check** if this instruction has recorded stack height bounds. If it does not, it means it was neither referenced by previous forward jump, nor is part of sequential instruction flow, and this code fails validation. 
+1. **Check** if this instruction has recorded stack height bounds. If it does not, it means it was neither referenced by previous forward jump, nor is part of sequential instruction flow, and this code fails validation.
+  - It is a prerequisite to validation algorithm, and code generators are required to order code basic blocks in a way that no block is referenced only by backwards jump.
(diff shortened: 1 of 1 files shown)


EIPS/eip-5450.md
```diff
@@ -58,48 +58,50 @@ In the second validation phase control-flow analysis is performed on the code.
...
+      - for `JUMPF` into returning function both the recorded lower and upper bound must equal exactly `type[current_section_index].outputs + type[target_section_index].inputs - type[target_section_index].outputs`.
+      - for `JUMPF` into non-returning function the recorded stack height lower bound must be at least the number of inputs of the target function according to its type defined in the type section.
+      - for any other instruction the recodrded stack height lower bound must be at least the number of inputs required by instruction
+      -  there is no additional check for terminating instructions other than `RETF` and `JUMPF`, this implies that extra items left on stack at instruction ending EVM execution are allowed.
+   2. For `CALLF` and `JUMPF` **check** for possible stack overflow: if recorded stack height upper bound is greater than `1024 - types[target_section_index].max_stack_height + types[target_section_index].inputs`, validation fails.
+   3. Compute new stack height bounds after the instruction execution. Upper and lower bound are updated by the same value.
```
pdobacz, line 80: specify the value of the update, esp. for CALLF

EIPS/eip-5450.md
```diff
@@ -58,48 +58,50 @@ In the second validation phase control-flow analysis is performed on the code.
...
    2. All targets of a conditional or unconditional jump.
-3. For each successor instruction:
+4. For each successor instruction:
    1. **Check** if the instruction is present in the code (i.e. execution must not "fall off" the code).
-   2. If the instruction does not have stack height recorded (visited for the first time):
-      1. Record the instruction stack height as the value computed in 1.2.
-      2. Add the instruction to the *worklist*.
-   3. Otherwise, **check** if the recorded stack height equals the value computed in 1.2.
+   2. If the successor is reached via forwards jump or sequential flow from previous instruction:
+      1. If the instruction does not have stack height bounds recorded (visited for the first time), record the instruction stack height bound as the value computed in 2.2.
```
pdobacz, line 87:
```suggestion
      1. If the instruction does not have stack height bounds recorded (visited for the first time), record the instruction stack height bound as the value computed in 2.3.
```

EIPS/eip-5450.md
```diff
@@ -58,48 +58,50 @@ In the second validation phase control-flow analysis is performed on the code.
...
+      2. Otherwise instruction was already visited (by previously seen forward jump). Update this instruction's recorded stack height bounds so that they contain the bounds computed in 2.2, i.e. `target_stack_min = min(target_stack_min, current_stack_min)` and `target_stack_max = max(target_stack_max, current_stack_min)`, where `(target_stack_min, target_stack_max)` are successor bounds and `(current_stack_min, current_stack_max)` are bounds computed in 2.2.
```
pdobacz, line 88:
```suggestion
      2. Otherwise instruction was already visited (by previously seen forward jump). Update this instruction's recorded stack height bounds so that they contain the bounds computed in 2.3, i.e. `target_stack_min = min(target_stack_min, current_stack_min)` and `target_stack_max = max(target_stack_max, current_stack_min)`, where `(target_stack_min, target_stack_max)` are successor bounds and `(current_stack_min, current_stack_max)` are bounds computed in 2.3.
```

+ some more occurrences of this below

EIPS/eip-5450.md
```diff
@@ -58,48 +58,50 @@ In the second validation phase control-flow analysis is performed on the code.
...
-While *worklist* is not empty, dequeue an instruction and:
+For each instruction:
 
-1. Determine the effect the instruction has on the operand stack:
-   1. **Check** if the recorded stack height satisfies the instruction requirements. Specifically:
-      - for `CALLF` instruction the recorded stack height must be at least the number of inputs of the called function according to its type defined in the type section.
-      - for `RETF` instruction the recorded stack height must be exactly the number of outputs of the function matching the code.
-   2. Compute new stack height after the instruction execution.
-2. Determine the list of successor instructions that can follow the current instructions:
+1. **Check** if this instruction has recorded stack height bounds. If it does not, it means it was neither referenced by previous forward jump, nor is part of sequential instruction flow, and this code fails validation.
```
pdobacz, line 70: We should mention the ordering of blocks as in eof.md, somewhere above: "Code basic blocks must be ordered in a way that every block is reachable either by a forward jump or sequential flow of instructions. In other words, there is no basic block reachable only by a backward jump."

EIPS/eip-5450.md
```diff
@@ -58,48 +58,50 @@ In the second validation phase control-flow analysis is performed on the code.
...
    3. **Check** if the maximum stack height matches the corresponding code section's `max_stack_height` within the type section.
 
 The computational and space complexity of this pass is *O(len(code))*. Each instruction is visited at most once.
 
 ### Execution
 
-Given the deploy-time validation guarantees, an EVM implementation is not required anymore to have run-time stack underflow nor overflow checks for each executed instruction. The exception is the `CALLF` performing operand stack overflow check for the entire called function.
+Given the deploy-time validation guarantees, an EVM implementation is not required anymore to have run-time stack underflow nor overflow checks for each executed instruction. The exception is the `CALLF` and `JUMPF` performing operand stack overflow check for the entire called function.
 
 ## Rationale
```
pdobacz, line 102: Once we're done with the specifications part, needs adding justification of going down the relaxed validation path

pdobacz (commented):
LGTM, I left some comments. However, I would still opt for not mentioning JUMPF. To me it seems it can just be left out and this EIP is well formed. EIP-6206 does a good job in building upon this foundation. To add to this, mentioning JUMPF requires referencing EIP-6206, we should not weasel around it, and then if we do mention, we get more circular deps :(

Gemrav (approved):

EIPS/eip-5450.md
gumb0: Added.

EIPS/eip-5450.md
gumb0: fixed

EIPS/eip-5450.md
gumb0: Added in Rationale.

EIPS/eip-5450.md
gumb0: I mentioned it, but decided to put it into as a sub-point.

It feels not as important for implementers of the algorithm, and mainly important for code generators, so I thought it fits as a note. wdyt?

pdobacz: totally can be a note :+1:

EIPS/eip-5450.md
```diff
@@ -27,14 +27,16 @@ and preventing the execution and deployment of any invalid code.
 The operand stack validation provides several benefits:
 
 - removes the run-time stack underflow check for all instructions,
-- removes the run-time stack overflow check for all instruction except `CALLF`,
+- removes the run-time stack overflow check for all instructions except `CALLF` and `JUMPF`,
```
pdobacz, line 30: reference a "separate EIP" for JUMPF, like in 4750?

EIPS/eip-5450.md
```diff
@@ -27,14 +27,16 @@ and preventing the execution and deployment of any invalid code.
...
+- allows for multiple forward jump instructions executing at different stack heights to target the same instruction; the stack of target basic stack is validated for all possible heights,
+- does not allow backwards jump instructions to target an instruction, that can be executed with different stack height; this prevents deployment of the loops with unbounded stack pushing or popping,
```
pdobacz, line 32: these two bullets belong more in the Rationale section IMO, unless they are already covered there. At most this "prevents deployment of the loops with unbounded stack pushing or popping" sentence is a Motivation of the EIP as a whole.

EIPS/eip-5450.md
```diff
@@ -58,50 +60,59 @@ In the second validation phase control-flow analysis is performed on the code.
 
 *Terminating instructions* refers to the instructions either:
 
-- ending function execution: `RETF`, or
-- ending whole EVM execution: `STOP`, `RETURN`, `REVERT`, `INVALID`.
+- ending function execution: `RETF`, `JUMPF`, or
```
pdobacz, line 63: "separate EIP" again, the following ones are OK to not have it, but at least once for the Spec section is good to have

pdobacz (approved):
Nice, just some nitpicks/suggestions to consider

EIPS/eip-5450.md
gumb0: should `RETURNCONTRACT` mention have the same then?

pdobacz: oh boy, I've missed RETURNCONTRACT popping up here... my top-level opinion doesn't really change (should not mention at all), but if we mention, we should consistent in how we mention :(. We cannot leave the reader puzzled whether RETURNCONTRACT is something in legacy EVM, a typo, copy-pasta or leftovers from prev revision or whatnot.

But once more, I really think these EIPs should be as independent as possible, rather than trying to be the megaspec and cover all aspects in each thoroughly

EIPS/eip-5450.md
gumb0: I just wanted to have a neat list of all code properties achieved by validation. I put it now into Rationale, but this way it duplicates a bit what is said in other parts, ptal

pdobacz: IMO, much better this way, thank you. Allowing for "multiple forwards jump instructions... etc" sounded weird in Motivation, as it is possible with dynamic JUMP. As such, it is motivation for choosing a certain design over alternatives, rather than the entire EIP over _status quo_.

pdobacz (approved):

EIPS/eip-5450.md
```diff
@@ -58,50 +58,76 @@ In the second validation phase control-flow analysis is performed on the code.
...
+      - for `RETF` instruction both the recorded lower and upper bound must be equal and must be exactly the number of outputs of the function matching the code,
+      - for `JUMPF` into returning function both the recorded lower and upper bound must equal exactly `type[current_section_index].outputs + type[target_section_index].inputs - type[target_section_index].outputs`,
+      - for `JUMPF` into non-returning function the recorded stack height lower bound must be at least the number of inputs of the target function according to its type defined in the type section,
+      - for any other instruction the recodrded stack height lower bound must be at least the number of inputs required by instruction,
+      - there is no additional check for terminating instructions other than `RETF` and `JUMPF`, this implies that extra items left on stack at instruction ending EVM execution are allowed.
+   2. For `CALLF` and `JUMPF` **check** for possible stack overflow: if recorded stack height upper bound is greater than `1024 - types[target_section_index].max_stack_height + types[target_section_index].inputs`, validation fails.
```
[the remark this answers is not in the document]
gumb0: 2.1 are the checks that guarantee no stack underflow,
2.2 is the check that prevents (some) stack overflows.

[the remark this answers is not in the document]
rodiazet: OK

PR #8390, opened 2024-04-04
Update EIP-152: Fix reference test

chfast:
The solidity example with the reference test of https://datatracker.ietf.org/doc/html/rfc7693#appendix-A is at least confusing because it declares values of type `bytes8` using a literal of length of 4 bytes.


lightclient (approved):

xinbenlv (approved):

xinbenlv:
For this kind of fix, can we just lower the requirement to just 2 editor-s approval? or even one should be ok

g11tech (approved):

chfast:
Thank you for handling this.

PR #8392, opened 2024-04-05
Update EIP-5792: Add showCallsStatus

lukasrosario:
- add back `wallet_showCallsStatus`. this is a method that requests a wallet shows information about a call bundle. this was part of the original EIP-5792.
- added semicolons throughout ts
- add an address param to `wallet_getCapabilities`. capabilities can differ per address. also noted that apps should have an already-approved connection to this provided address.

### DIFF
--- a/EIPS/eip-5792.md
+++ b/EIPS/eip-5792.md
@@ -161,23 +161,23 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
 #### `wallet_getCallsStatus` RPC Specification
 
 ```typescript
-type GetCallsParams = string
+type GetCallsParams = string;
 
 type GetCallsResult = {
-  status: 'PENDING' | 'CONFIRMED'
+  status: 'PENDING' | 'CONFIRMED';
   receipts?: {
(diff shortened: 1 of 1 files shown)


EIPS/eip-5792.md
```diff
@@ -207,6 +216,26 @@ As with the return value of `wallet_sendCalls`, the batch identifier may be any
 }
 ```
 
+### `wallet_showCallsStatus`
```
arein, line 219: what's the motivation for this one?

lukasrosario: this was in the original 5792. more advanced wallets (eg 4337) will have the most context on tx statuses (such as pending in bundler mempool). could also do things like speeding up / replacing transactions here

wilsoncusack: @arein idea A simple motivating case would be a user wants to speed up or drop a transaction, and the app wants to help kick that off, this would be a screen where a wallet could provide those actions. 

More generally this is a place where a wallet could show a user a lot more details about what happened, if they want to see it.

PR #8406, opened 2024-04-11
Update EIP-7002: add support for partial withdrawals, rename exits

lightclient:
This updates the EIP to support partial withdrawals in accordance with max eb.

### DIFF
--- a/EIPS/eip-7002.md
+++ b/EIPS/eip-7002.md
@@ -1,6 +1,6 @@
 ---
 eip: 7002
-title: Execution layer triggerable exits and partial withdrawals
+title: Execution layer triggerable exits
 description: Allow validators to trigger exits and partial withdrawals via their execution layer (0x01) withdrawal credentials
 author: Danny Ryan (@djrtwo), Mikhail Kalinin (@mkalinin), Ansgar Dietrichs (@adietrichs), Hsiao-Wei Wang (@hwwhww), lightclient (@lightclient)
 discussions-to: https://ethereum-magicians.org/t/eip-7002-execution-layer-triggerable-exits/14195
--- a/EIPS/eip-7002.md
+++ b/EIPS/eip-7002.md
@@ -1,7 +1,7 @@
 ---
 eip: 7002
-title: Execution layer triggerable exits
-description: Allows validators to trigger exits via their execution layer (0x01) withdrawal credentials
+title: Execution layer triggerable exits and partial withdrawals
+description: Allow validators to trigger exits and partial withdrawals via their execution layer (0x01) withdrawal credentials
 author: Danny Ryan (@djrtwo), Mikhail Kalinin (@mkalinin), Ansgar Dietrichs (@adietrichs), Hsiao-Wei Wang (@hwwhww), lightclient (@lightclient)
 discussions-to: https://ethereum-magicians.org/t/eip-7002-execution-layer-triggerable-exits/14195
 status: Draft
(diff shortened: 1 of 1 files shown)


EIPS/eip-7002.md
```diff
@@ -38,92 +38,95 @@ Note, 0x00 withdrawal credentials can be changed into 0x01 withdrawal credential
...
-    queue_storage_slot = EXIT_MESSAGE_QUEUE_STORAGE_OFFSET + queue_tail_index * 3
-    sstore(VALIDATOR_EXIT_ADDRESS, queue_storage_slot, msg.sender)
-    sstore(VALIDATOR_EXIT_ADDRESS, queue_storage_slot + 1, validator_pubkey[0:32])
-    sstore(VALIDATOR_EXIT_ADDRESS, queue_storage_slot + 2, validator_pubkey[32:48])
-    sstore(VALIDATOR_EXIT_ADDRESS, EXIT_MESSAGE_QUEUE_TAIL_STORAGE_SLOT, queue_tail_index + 1)
+    queue_tail_index = sload(WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS, WITHDRAWAL_REQUEST_QUEUE_TAIL_STORAGE_SLOT)
+    queue_storage_slot = WITHDRAWAL_REQUEST_QUEUE_STORAGE_OFFSET + queue_tail_index * 3
+    sstore(WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS, queue_storage_slot, msg.sender)
+    sstore(WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS, queue_storage_slot + 1, validator_pubkey[0:32])
+    sstore(WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS, queue_storage_slot + 2, validator_pubkey[32:48])
```
mkalinin, line 115:
```suggestion
    sstore(WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS, queue_storage_slot + 2, validator_pubkey[32:48] ++ amount)
```

EIPS/eip-7002.md
```diff
@@ -536,10 +527,10 @@ block.header.exits_root = compute_trie_root_from_indexed_data(block.body.exits)
 
 Sketch of spec:
 
-* New operation `ExecutionLayerExit`
-* Will show up in `ExecutionPayload` as an SSZ List bound by length `MAX_EXITS_PER_BLOCK`
-* New function in `process_execution_layer_exit` that has similar functionality to `process_voluntary_exit` but that can fail validations (e.g. validator is already exited) without the block failing (similar to deposit coming from EL)
-* `process_execution_layer_exit` called in `process_operations` for each `ExecutionLayerExit` found in the `ExecutionPayload`
+* New operation `ExecutionLayerWithdrawalRequest`
+* Will show up in `ExecutionPayload` as an SSZ List bound by length `MAX_WITHDRAWAL_REQUESTS_PER_BLOCK`
+* New function in `process_execution_layer_withdrawal_request` that has similar functionality to `process_voluntary_withdrawal_request` but that can fail validations (e.g. validator is already exited) without the block failing (similar to deposit coming from EL)
```
mkalinin, line 532: There is no `process_voluntary_withdrawal_request` and i think that we can say that a new function is similar to `process_voluntary_exits` because the functionality is extended with partial withdrawals. Maybe just say that the new function can fail validations without block failing and that would be it?

mkalinin (approved):
LGTM! I think that we need to plan at least the following updates to the EIP not related to this change:
1. Change consensus spec to Electra as after the spec was moved from a separate feature to Electra and merged with EIP-7251, it has been changed because of partial withdrawals were added to CL.
2. Mention 0x02 creds in the Motivation section as currently it seems like this EIP is useful for 0x01 creds only.

lightclient:
Okay I've done 1) that you mentioned above. Not familiar with 0x02 cred so maybe you take a stab at that?

PR #8407, opened 2024-04-12
Update EIP-3074: Add `AUTHORIZED` opcode

clabby:
## Overview

Introduces a new opcode to EIP-3074, `AUTHORIZED`, that allows for smart contracts to introspect the `authorized` context variable of the current or parent callframe.

### Rationale

1. Other callframe-level context variables are accessible within the executing context via accessor opcodes.
2. With 3074, checks in smart contracts of the style `msg.sender == tx.origin` to guarantee callframe depth = 0 or that `msg.sender` is in fact an EOA are no longer viable due to `authorized` being able to be set to `tx.origin`. While the rationale of this invariant not being affected by callframe depth is true in the context of an "isEOA" check today, that is only in the case if `msg.sender == parent_callframe.authorized == tx.origin` within an `AUTHCALL` context. It is no longer possible to tell if `msg.sender` is an EOA if the check is performed in an `AUTHCALL` context where `tx.origin` is not the parent callframe's `authorized` context variable.
    - Example:
        - Alice signs an `AUTH` message.
        - Bob relays Alice's `AUTH` message with an `AUTHCALL`, which calls `Contract`.
        - `Contract` checks if `tx.origin == msg.sender`, to determine if `msg.sender` is an EOA.
        - `Bob != Alice`, so `Alice` is wrongfully assumed by this contract to be a smart contract rather than an EOA.
    - Example 2:
        - Bob signs an `AUTH` message.
        - Bob relays his own `AUTH` message with an `AUTHCALL`, which calls `Contract`.
        - `Contract` checks if `tx.origin == msg.sender`, to determine if the current callframe depth is `0` (and therefore, the calldata will be available in the block body).
        - `Bob == Bob`, so the contract will wrongfully assume that the context is the originating callframe. The data that the contract believed would be available within the block body is now only available by executing the transaction and inspecting the memory passed to `AUTHCALL`.
    - Real world example: In the OP Stack's `OptimismPortal` contract, [Address Aliasing](https://specs.optimism.io/protocol/deposits.html?highlight=alias#address-aliasing) is applied for smart contracts that call the `depositTransaction` function [here](https://github.com/ethereum-optimism/optimism/blob/94f04ba950488353afa1881073a65baf2f202b44/packages/contracts-bedrock/src/L1/OptimismPortal.sol#L400-L404). Without being able to introspect the parent callframe's `authorized` context variable, it is now impossible to determine whether or not `msg.sender` is an EOA inside of an `AUTHCALL` frame, and therefore whether or not the `OptimismPortal` should alias the address. 
3. While this opcode can be used by existing contracts that rely on determining whether or not `msg.sender` is an EOA, or that the callframe depth is guaranteed to be zero when combined with `msg.sender == tx.origin`, it can also be used to detect when a message is being relayed on behalf of `msg.sender`.

### DIFF
--- a/EIPS/eip-3074.md
+++ b/EIPS/eip-3074.md
@@ -193,6 +193,44 @@ else:
 
 As with `CALL`, the full gas cost is charged immediately, independently of actually executing the call.
 
+### `AUTHORIZED` (`0x4b`)
+
+A new opcode `AUTHORIZED` shall be created at `0x4b`. It shall take one stack element as input, and it shall return one stack element.
+
+#### Input
+
+##### Stack
+
+| Stack      | Value                       |
+| ---------- | --------------------------- |
+| `top - 0`  | `currentOrParentCallframe`  |
+
+##### Memory 
+
+Memory is not modified by this instruction.
+
+#### Output 
+
+##### Stack
+
+| Stack      | Value                                            |
+| ---------- | ------------------------------------------------ |
+| `top - 0`  | `(currentCallframe\|parentCallframe).authorized` |
+
+##### Memory 
+
+Memory is not modified by this instruction.
+
+#### Behavior 
+
+If `currentOrParentCallframe` is 0, `AUTHORIZED` will return the `authorized` context variable of the current callframe. If `currentOrParentCallframe` is any value greater than 0, `AUTHORIZED` will return the `authorized` context variable of the parent callframe, if it exists.
+
+If `currentOrParentCallframe` is greater than zero, and the current callframe is the top-level callframe, `AUTHORIZED` will return 0.
+
+#### Gas Cost
+
+The fee for `AUTHORIZED` is the same as other context variable accessor opcodes, such as `ADDRESS`, `ORIGIN`, `CALLER`, and `CALLVALUE`, at a constant fee of `2` gas.
+
 ## Rationale
 
 ### Signature in Memory
@@ -335,7 +373,8 @@ Allowing `authorized` to equal `tx.origin` has the possibility to:
  - Break atomic sandwich protections which rely on `tx.origin`;
  - Break reentrancy guards of the style `require(tx.origin == msg.sender)`.
 
-The authors of this EIP believe the risks of allowing `authorized` to equal `tx.origin` are acceptable for the reasons outlined in the Rationale section.
+The authors of this EIP believe the risks of allowing `authorized` to equal `tx.origin` are acceptable for the reasons outlined in the Rationale section. In addition, with the `AUTHORIZED` opcode,
+smart contract authors may choose to reject transactions that are being relayed with `AUTHCALL` by checking if the parent context's `authorized` context variable is set to `msg.sender`.
 
 ## Copyright
 


lightclient:
I'm against this addition. We have spent years warning users of `caller == origin` checks, stating that one should not be able to differentiate between accounts.

We should avoid further differentiation of accounts with this op.

SamWilsn:
> It is no longer possible to tell if `msg.sender` is an EOA

If necessary, you can still use `ecrecover` to prove an account is an EOA, or possibly `EXTCODEHASH`/`EXTCODESIZE` if you need less strict guarantees. Not exactly great solutions, but I believe&mdash;like @lightclient&mdash;that we should avoid further differentiation between account types.

> The data that the contract believed would be available within the block body is now [not.]

The workaround is certainly a bit painful. How badly does it affect your usecase? If I understand correctly, extracting the inner calldata would require the block's state while the outer calldata only needs the block itself?

clabby:
> We should avoid further differentiation of accounts with this op.
> & 
> Not exactly great solutions, but I believe—like @lightclient—that we should avoid further differentiation between account types.

I see where you both are coming from and agree. The focus of this opcode was primarily to allow for smart contracts to continue making assertions about being in the originating callframe as well as detect if they are currently executing within an `AUTHCALL` context. It does have the unfortunate byproduct of also allowing "isEOA" checks to persist, so it may not be the right solution.

> The workaround is certainly a bit painful. How badly does it affect your usecase? If I understand correctly, extracting the inner calldata would require the block's state while the outer calldata only needs the block itself?

Yes, if we continue relying on this pattern, we would need to execute the block to gather the input data, rather than being able to statically guarantee that it will be available for an off-chain actor's consumption by viewing the block body. I think we can work around it by using blobs as DA for the application I have in mind. In the OP Stack's fault proof, we must be able to add data to the host in a verifiable way on-chain, which will be requested by key in a `read` syscall instruction on the MIPS/RISC-V VM during single instruction steps within the dispute.

Right now, the way that we do this for large `keccak256` preimages (i.e. receipts with many logs) is through optimistic proposals and a single-step keccak dispute, where the user streams in the full preimage as well as commitments to the internal state of the hash function for every 1088 bits of the preimage processed by the hash function. We optimistically accept the commitments to the hash function's computation as true, and allow for an off-chain agent to show that they are wrong by executing a single absorbtion / permutation step with a Solidity implementation of the SHA3 permutation.

As the proposal is streamed in, the contract iteratively merkleizes leaves composed of each keccak block, the commitment to the internal hash function state after processing each block, and the index of the block during the proposal process. This tree is built incrementally using the same incremental merkle tree algorithm as the beacon deposit contract.

We rely on the call to adding leaves to the tree being in the top-level callframe so that the raw leaf data is available in the block body, which brings up the need of the contract being able to identify whether or not it is in the originating callframe. The off-chain challenge agent currently only needs a pointer to the block(s) where the leaves were added, and can statically retrieve this data to rebuild the merkle tree within the contract off-chain with only the full block. If the input data to the contract is not available in the block body (i.e. leaves were added by a contract or an authorized EOA in a deeper callframe), the off-chain agent must execute the blocks to find the input, in order to reconstruct the merkle tree off-chain, check its validity, and create a challenge. This creates much more computational overhead on the end of the off-chain challenge agent.

I believe that this issue can be solved in our protocol by using blobs rather than calldata to post the large preimage as well as the intermediate state transitions, and tracking the versioned hashes along the way. We should then be able to create openings on these blobs to reconstruct the current leaf structure from field elements to retain a similar design.

Will close this for now and go back to the drawing board - appreciate your thoughts :)

Issue #3010, opened 2020-09-30
EIP-3009: "Transfer With Authorization" - ERC20 meta-transactions


petejkim:
**Pull Request: [EIP-3009: transferWithAuthorization](https://github.com/ethereum/EIPs/pull/3009)**

This is the place to discuss EIP-3009 "Transfer With Authorization", implemented by the USD Coin (USDC) v2 smart contract alongside [EIP-2612: permit](https://eips.ethereum.org/EIPS/eip-2612).

- - -

---
eip: 3009
title: Transfer With Authorization
author: Peter Jihoon Kim (@petejkim), Kevin Britz (@kbrizzle), David Knott (@DavidLKnott)
discussions-to: https://github.com/ethereum/EIPs/issues/3010
status: Draft
type: Standards Track
category: ERC
created: 2020-09-28
requires: 20, 712
---

## Simple Summary

A contract interface that enables transferring of fungible assets via a signed authorization.

## Abstract

A set of functions to enable meta-transactions and atomic interactions with [ERC-20](./eip-20.md) token contracts via signatures conforming to the [EIP-712](./eip-712.md) typed message signing specification.

This enables the user to:

- delegate the gas payment to someone else,
- pay for gas in the token itself rather than in ETH,
- perform one or more token transfers and other operations in a single atomic transaction,
- transfer ERC-20 tokens to another address, and have the recipient submit the transaction,
- batch multiple transactions with minimal overhead, and
- create and perform multiple transactions without having to worry about them failing due to accidental nonce-reuse or improper ordering by the miner.

## Motivation

There is an existing spec, [EIP-2612](./eip-2612), that also allows meta-transactions, and it is encouraged that a contract implements both for maximum compatibility. The two primary differences between this spec and EIP-2612 are that:

- EIP-2612 uses sequential nonces, but this uses random 32-byte nonces, and that
- EIP-2612 relies on the ERC-20 `approve`/`transferFrom` ("ERC-20 allowance") pattern.

The biggest issue with the use of sequential nonces is that it does not allow users to perform more than one transaction at time without risking their transactions failing, because:

- DApps may unintentionally reuse nonces that have not yet been processed in the blockchain.
- Miners may process the transactions in the incorrect order.

This can be especially problematic if the gas prices are very high and transactions often get queued up and remain unconfirmed for a long time. Non-sequential nonces allow users to create as many transactions as they want at the same time.

[...]

PR #8412, opened 2024-04-12
Update EIP-3074: Update eip-3074.md

SamWilsn:


lightclient (approved):

juli:
Thanks for making the PR @SamWilsn.

 I mentioned this risk during the [initial EIP-3010 discussion](https://github.com/ethereum/EIPs/issues/3010#issuecomment-706444635)  and community members with experience designing and implementing transaction relaying systems, such as @yahgwai agreed.

The attack scenario considered involves competing relay services attacking each other to make a profit. They do this by canceling transactions, which makes their competitors waste gas and lose money.

When designing these systems, it's possible to find a conflict between protecting end users and relay services. However, the risk to end users is lower because if one relay service fails to deliver the expected service, users can easily switch to another with little or no harm. On the other hand, allowing attacks that could potentially put relay services out of business represents a more significant design flaw.

PR #8417, opened 2024-04-13
Update EIP-3074: Fix typos and grammar

omahs:
Update EIP-3074: Fix typos and grammar

### DIFF
--- a/EIPS/eip-3074.md
+++ b/EIPS/eip-3074.md
@@ -266,7 +266,7 @@ In previous iterations of this EIP, it was thought that deducting value from an 
 
 However, after further investigation we found that breaking the invariant is safe. This is mostly due to the fact that the worst case is similar in both instances.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-3074.md
```diff
@@ -266,7 +266,7 @@ In previous iterations of this EIP, it was thought that deducting value from an
...
 
-Currently an attacker can queue many transactions in the tx pool, across many accounts, and invalidate them all at once with a block where each of the queued accounts send a tx moving their entire balance. This attack will become easier and cheaper after this EIP, because it will no longer require direct access to the block builder and will not cost a full 21000 gas to originate each tx. However, the attack does not have a substantial impact on the network, so reducing the difficulty and cost are not of concern.
+Currently an attacker can queue many transactions in the tx pool, across many accounts, and invalidate them all at once with a block where each of the queued accounts sends a tx moving their entire balance. This attack will become easier and cheaper after this EIP, because it will no longer require direct access to the block builder and will not cost a full 21000 gas to originate each tx. However, the attack does not have a substantial impact on the network, so reducing the difficulty and cost is not of concern.
```
lightclient, line 269:
```suggestion
Currently an attacker can queue many transactions in the tx pool, across many accounts, and invalidate them all at once with a block where each of the queued accounts send a tx moving their entire balance. This attack will become easier and cheaper after this EIP, because it will no longer require direct access to the block builder and will not cost a full 21000 gas to originate each tx. However, the attack does not have a substantial impact on the network, so reducing the difficulty and cost is not of concern.
```

lightclient (approved):

PR #8445, opened 2024-04-16
Update EIP-3074: Explain `nonce` and `chainId` better

SamWilsn:


sambacha:
shouldnt this be in the security considerations section regarding chainId?

PR #8476, opened 2024-04-21
Update EIP-7002: Fix after renaming fields (exit->count)

MarekM25:
exit-> count fix, I noticed a small bug introduced in this PR: https://github.com/ethereum/EIPs/pull/8406

cc: @lightclient


lightclient (approved):
Nice catch!

PR #8487, opened 2024-04-23
Update EIP-2935: replace MOD with AND

chfast:


jsign (approved):
Thank you!

chfast:
You should also update the bytecode.

g11tech:
> You should also update the bytecode.

the contract address will also get modified as well  cc @jsign

PR #8488, opened 2024-04-23
Update EIP-2935: note possible ADD overflow

chfast:


g11tech:
ahh thank you for reviewing: have addressed it like this:
 - https://github.com/ethereum/EIPs/pull/8514
 
since subtraction will also involve underflow check etc, and input is not going to go beyond 8 bytes which is the type size assumed in various specs including the CL spec types

PR #8514, opened 2024-05-03
Update EIP-2935: restrict contract input to 8 bytes to prevent overflow in checks

g11tech:
addresses the issue raised in:
 - https://github.com/ethereum/EIPs/pull/8488/files


gballet (approved):
LGTM

PR #8515, opened 2024-05-03
Update EIP-2935: fix the input validation as 32 byte calldataload

g11tech:
realized that the contract uses calldataload which pads right with zeros so previous 8 byte check was incorrect (https://github.com/ethereum/EIPs/pull/8514)

now changes to direct compare against the max 8 byte val as the simplest option

### DIFF
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -71,7 +71,9 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
 Exact evm assembly that can be used for the contract to resolve `BLOCKHASH`
 
 ```
-// check if input > 8 byte value revert
+// check if input > 8 byte value and revert if this isn't the case
+// the check is performed by comparing a the biggest 8 byte number with
+// the call data, which is a right-padded 32 byte number.
 push8 0xffffffffffffffff
 push0
 calldataload
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -71,11 +71,12 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
 Exact evm assembly that can be used for the contract to resolve `BLOCKHASH`
 
 ```
-// check if inputsize>8 revert
-push1 0x08
-calldatasize
+// check if input > 8 byte value revert
+push8 0xffffffff
+push0
+calldataload
 gt
-push1 0x31
+push1 0x39
 jumpi
 
 // check if input > blocknumber-1 then return 0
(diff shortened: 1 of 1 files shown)


EIPS/eip-2935.md
```diff
@@ -71,11 +71,12 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
 Exact evm assembly that can be used for the contract to resolve `BLOCKHASH`
 
 ```
-// check if inputsize>8 revert
-push1 0x08
-calldatasize
+// check if input > 8 byte value revert
+push8 0xffffffff
```
gballet, line 75: this is a 32 bit number, not 64 bit

```suggestion
push8 0xffffffffffffffff
```

EIPS/eip-2935.md
```diff
@@ -71,11 +71,12 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
...
+push0
+calldataload
 gt
-push1 0x31
+push1 0x39
```
gballet, line 79: ok, you removed 3 instructions and added 11, that's 8 indeed.

gballet (changes requested):
I think that the pushed data is incorrect, but apart from that LGTM.

EIPS/eip-2935.md
```diff
@@ -71,11 +71,12 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
 Exact evm assembly that can be used for the contract to resolve `BLOCKHASH`
 
 ```
-// check if inputsize>8 revert
-push1 0x08
-calldatasize
+// check if input > 8 byte value revert
```
gballet, line 74:
```suggestion
// check if input > 8 byte value and revert if this isn't the case
// the check is performed by ANDing a 64bit bitmask with
// the call data, which is a right-padded 32 byte number.
```

EIPS/eip-2935.md
g11tech: yea, but i just checked the new jumpdest destinations via evm codes :) but yes math checks out

EIPS/eip-2935.md
g11tech: ah thanks!

EIPS/eip-2935.md
```diff
@@ -71,11 +71,14 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
 Exact evm assembly that can be used for the contract to resolve `BLOCKHASH`
 
 ```
-// check if inputsize>8 revert
-push1 0x08
-calldatasize
+// check if input > 8 byte value and revert if this isn't the case
+// the check is performed by comparing a the biggest 8 byte number with
```
g11tech, line 75:
```suggestion
// the check is performed by comparing the biggest 8 byte number with
```

PR #8524, opened 2024-05-06
Update EIP-7495: fix property accessors for `Variant[S]`

etan-status:


etan-status:
Integrating this into another PR that also affects the .md file so that it auto approves. If one only changes assets, no auto approve is triggered.

PR #8525, opened 2024-05-06
Update EIP-6493: Convert to `StableContainer` for serialization tests

etan-status:


etan-status:
Integrating this into another PR that also changes an `.md` file, so that it automerges.

PR #8493, opened 2024-04-24
Add EIP: AUTHCREATE opcode

txgyy:
[EIP-7697：AUTHCREATE opcode](https://ethereum-magicians.org/t/eip-7697-authcreate-opcode/19780)
## Abstract

The EIP introduces an EVM instruction `AUTHCREATE`. It allows a deterministic addresses to migrate to a smart contract.

Inspiration comes from [EIP-3074](./eip-3074.md) and [EIP-5003](./eip-5003.md):

- ([EIP-3074](./eip-3074.md)) + [EIP-5003](./eip-5003.md) = (`AUTH` + `AUTHCALL`) + `AUTHUSURP`
- [EIP-7697](./eip-7697.md) = `AUTHCREATE` = `AUTH` + `AUTHUSURP`

## Motivation

For a long time, the EVM ecosystem has been plagued by two issues:

1. Difficulty ensuring consistency across addresses on multiple chains, even when using the same bytecode.
2. EOAs lack contract capabilities, preventing the realization of account abstraction.
    1. Authentication - any form of proving one’s identity.
    2. Authorization - any access policy.
    3. Replay protection - transaction ordering decoupled from replay protection.
    4. Gas payment - gas payment decoupled from the account itself.
    5. Execution - any execution logic.

[EIP-7697](./eip-7697.md) will address these issues.

| Feature                                                      | [EIP-7697](./eip-7697.md)  | [EIP-3074](./eip-3074.md) | [EIP-5003](./eip-5003.md) | [EIP-7377](./eip-7377.md) |
|--------------------------------------------------------------|----------------------------|---------------------------|---------------------------|---------------------------|
| Deploy the same address [ERC-2470](./eip-2470.md)            | ✅                          | ❎                         | ✅ need 3 opcodes          | ✅ only EOA                |
| Help EOA to upgrade to CA                                    | ✅                          | ❎                         | ✅ need 3 opcodes          | ✅                         |
| Support `secp256r1` or more                                  | ✅                          | ✅                         | ❎                         | ❎                         |
| Reuse existing wallet infrastructure                         | ✅ adapt the contract       | ❎                         | ✅ adapt the contract      | ❎ adapt the node rpc      |
| Integrate easily with [ERC-4337](./eip-4337.md) and RIP-7560 | ✅ support factory contract | ❎                         | ❎                         | ❎                         |
| Grant temporary CA capabilities to EOA                       | ❎                          | ✅                         | ❎                         | ❎                         |

PR #8527, opened 2024-05-07
Add EIP: Set EOA account code for one transaction

vbuterin:
Add a new transaction type that adds a `contract_code` field and a signature, and converts the signing account (not necessarily the same as the `tx.origin`) into a smart contract wallet for the duration of that transaction. Intended to offer similar functionality to EIP-3074.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -18,9 +18,9 @@ Add a new transaction type that adds a `contract_code` field and a signature, an
 
 There is a lot of interest in adding short-term functionality improvements to EOAs, increasing the usability of applications and in some cases allowing improved security. Three particular applications include:
 
-* **Batching**: allowing multiple operations from the same user in one atomic transaction. One common example is an ERC20 approval followed by spending that approval, a common workflow in DEXes that requires two transactions today. Advanced use cases of batching occasionally involve dependencies: the outhttps://ethereum-magicians.org/t/eip-7701-native-account-abstraction-with-eof/19893put of the first operation is part of the input to the second operation.
-* **Sponsorship**: account X pays for a transaction on behalf of account Y. Account X could be paid in some other ERC20 for this service, or it could be an application operator including the transactions of its users for free.
-* **Privilege de-escalation**: users can sign sub-keys, and give them specific permissions that are much weaker than global access to the account. For example, you could imagine a permission to spend ERC20 tokens but not ETH, or to spend up to 1% of total balance per day, or to interact only with a specific application.
+* **Batching**: allowing multiple operations from the same user in one atomic transaction. One common example is an ERC-20 approval followed by spending that approval, a common workflow in DEXes that requires two transactions today. Advanced use cases of batching occasionally involve dependencies: the output of the first operation is part of the input to the second operation.
+* **Sponsorship**: account X pays for a transaction on behalf of account Y. Account X could be paid in some other ERC-20 for this service, or it could be an application operator including the transactions of its users for free.
+* **Privilege de-escalation**: users can sign sub-keys, and give them specific permissions that are much weaker than global access to the account. For example, you could imagine a permission to spend ERC-20 tokens but not ETH, or to spend up to 1% of total balance per day, or to interact only with a specific application.
 
 [EIP-3074](./eip-3074.md) solves all of these use cases. However, it has forward-compatibility concerns:
 
@@ -35,10 +35,10 @@ The keywords “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL N
 
 ### Parameters
 
-- `FORK_BLKNUM` = `TBD`
-- `TX_TYPE` = `TBD`
-- `MAGIC` = `TBD`
-- `PER_CONTRACT_CODE_BASE_COST` = `5000`
+* `FORK_BLKNUM` = `TBD`
+* `TX_TYPE` = `TBD`
+* `MAGIC` = `TBD`
+* `PER_CONTRACT_CODE_BASE_COST` = `5000`
 
 As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introduced with `TransactionType` = `TX_TYPE(TBD)`.
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -8,6 +8,7 @@ status: Draft
 type: Standards Track
 category: Core
 created: 2024-05-07
+requires: 2718, 2930
 ---
 
 ## Abstract
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -2,7 +2,7 @@
 eip: 7702
 title: Set EOA account code for one transaction
 description: Add a new tx type that sets the code for an EOA during one transaction execution
-author: Vitalik Buterin (@vbuterin), Ansgar Dietrichs
+author: Vitalik Buterin (@vbuterin), Sam Wilson (@SamWilsn), Ansgar Dietrichs (@adietrichs), Matt Garnett (@lightclient)
 discussions-to: https://ethereum-magicians.org/t/eip-set-eoa-account-code-for-one-transaction/19923
 status: Draft
 type: Standards Track
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -2,7 +2,7 @@
 eip: 7702
 title: Set EOA account code for one transaction
 description: Add a new tx type that sets the code for an EOA during one transaction execution
-author: Vitalik Buterin (@vbuterin)
+author: Vitalik Buterin (@vbuterin), Ansgar Dietrichs
 discussions-to: https://ethereum-magicians.org/t/eip-set-eoa-account-code-for-one-transaction/19923
 status: Draft
 type: Standards Track
@@ -12,7 +12,7 @@ created: 2024-05-07
 
 ## Abstract
 
-Add a new transaction type that adds a `contract_code` field and a signature, and converts the signing account (not necessarily the same as the `tx.origin`) into a smart contract wallet for the duration of that transaction. Intended to offer similar functionality to EIP-3074.
+Add a new transaction type that adds a `contract_code` field and a signature, and converts the signing account (not necessarily the same as the `tx.origin`) into a smart contract wallet for the duration of that transaction. Intended to offer similar functionality to [EIP-3074](./eip-3074.md).
 
 ## Motivation
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -1,0 +1,97 @@
+---
+eip: 7702
+title: Set EOA account code for one transaction
+description: Add a new tx type that sets the code for an EOA during one transaction execution
+author: Vitalik Buterin (@vbuterin)
+discussions-to: https://ethereum-magicians.org/t/eip-set-eoa-account-code-for-one-transaction/19923
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-05-07
+---
+
+## Abstract
+
+Add a new transaction type that adds a `contract_code` field and a signature, and converts the signing account (not necessarily the same as the `tx.origin`) into a smart contract wallet for the duration of that transaction. Intended to offer similar functionality to EIP-3074.
+
+## Motivation
+
+There is a lot of interest in adding short-term functionality improvements to EOAs, increasing the usability of applications and in some cases allowing improved security. Three particular applications include:
+
+* **Batching**: allowing multiple operations from the same user in one atomic transaction. One common example is an ERC20 approval followed by spending that approval, a common workflow in DEXes that requires two transactions today. Advanced use cases of batching occasionally involve dependencies: the outhttps://ethereum-magicians.org/t/eip-7701-native-account-abstraction-with-eof/19893put of the first operation is part of the input to the second operation.
+* **Sponsorship**: account X pays for a transaction on behalf of account Y. Account X could be paid in some other ERC20 for this service, or it could be an application operator including the transactions of its users for free.
+* **Privilege de-escalation**: users can sign sub-keys, and give them specific permissions that are much weaker than global access to the account. For example, you could imagine a permission to spend ERC20 tokens but not ETH, or to spend up to 1% of total balance per day, or to interact only with a specific application.
+
+[EIP-3074](./eip-3074.md) solves all of these use cases. However, it has forward-compatibility concerns:
+
+* It introduces two opcodes, `AUTH` and `AUTHCALL`, that would have no use in an "endgame account abstraction" world where eventually all users are using smart contract wallets (which seems like it must happen eventually, at the least because eventually quantum computers will break the ECDSA that EOAs use)
+* It leads to the development of an "invoker contract" ecosystem that would be separate from the "smart contract wallet" ecosystem, leading to possible fragmentation of effort.
+
+The purpose of this EIP is to enable all of the use cases of EIP-3074, without these two weaknesses.
+
+## Specification
+
+The keywords “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+### Parameters
+
+- `FORK_BLKNUM` = `TBD`
+- `TX_TYPE` = `TBD`
+- `MAGIC` = `TBD`
+- `PER_CONTRACT_CODE_BASE_COST` = `5000`
+
+As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introduced with `TransactionType` = `TX_TYPE(TBD)`.
+
+The [EIP-2718](./eip-2718.md) `TransactionPayload` for this transaction is
+
+```
+rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, data, access_list, [[contract_code, y_parity, r, s], ...], signature_y_parity, signature_r, signature_s])
+```
+
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `16 * non-zero calldata bytes + 4 * zero calldata bytes` over each `contract_code`, plus `PER_CONTRACT_CODE_BASE_COST` times the length of the `contract_code` array.
+
+At the start of executing the transaction, for each `[contract_code, y_parity, r, s]` tuple:
+
+1. Let `signer = ecrecover(keccak(MAGIC + contract_code), y_parity, r, s]`.
+2. Verify that the contract code of `signer` is empty.
+3. Set the contract code of `signer` to `contract_code`.
+
+At the end of the transaction, set the `contract_code` of each `signer` back to empty.
+
+Note that the signer of any of the `contract_code` signatures, and the `tx.origin` of the transaction, are allowed to be different.
+
+## Rationale
+
+### Conversion of EIP-3074 use cases
+
+In this design, it requires fairly little work to convert an existing EIP-3074 workflow. Specifically, AUTH and AUTHCALL would get replaced by calls into the EOA. One way to do this is that the `contract_code` would be a user wallet (which could be a `DELEGATECALL` forwarder to save gas), and would expose two functions, `verify` and `execute`.
+
+* AUTH would be replaced by a code to `verify`, which would use TSTORE to locally set `authorized[msg.sender, ...] = True`.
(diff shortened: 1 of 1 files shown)


EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+
+As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introduced with `TransactionType` = `TX_TYPE(TBD)`.
+
+The [EIP-2718](./eip-2718.md) `TransactionPayload` for this transaction is
+
+```
+rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, data, access_list, [[contract_code, y_parity, r, s], ...], signature_y_parity, signature_r, signature_s])
+```
+
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `16 * non-zero calldata bytes + 4 * zero calldata bytes` over each `contract_code`, plus `PER_CONTRACT_CODE_BASE_COST` times the length of the `contract_code` array.
```
PhABC, line 51: Seems expensive, can users not point to an already deployed contract and reduce data size of txn payload?

rmeissner: minimal proxy accounts have very little bytecode. Using [ERC-1167](https://eips.ethereum.org/EIPS/eip-1167) should be sufficient and keep this EIP lean

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+eip: 7702
+title: Set EOA account code for one transaction
+description: Add a new tx type that sets the code for an EOA during one transaction execution
+author: Vitalik Buterin (@vbuterin), Sam Wilson (@SamWilsn), Ansgar Dietrichs (@adietrichs), Matt Garnett (@lightclient)
+discussions-to: https://ethereum-magicians.org/t/eip-set-eoa-account-code-for-one-transaction/19923
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-05-07
+---
```
SamWilsn, line 11:
```suggestion
requires: 2718, 2930
---
```

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+
+As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introduced with `TransactionType` = `TX_TYPE(TBD)`.
+
+The [EIP-2718](./eip-2718.md) `TransactionPayload` for this transaction is
+
+```
+rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, data, access_list, [[contract_code, y_parity, r, s], ...], signature_y_parity, signature_r, signature_s])
+```
+
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `16 * non-zero calldata bytes + 4 * zero calldata bytes` over each `contract_code`, plus `PER_CONTRACT_CODE_BASE_COST` times the length of the `contract_code` array.
```
lightclient, line 51: Why not 1559 tx? :)

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+* Hence, it avoids the problem of "creating two separate code ecosystems", because to a large extent they would be the same ecosystem. There would be some workflows that require kludges under this solution that would be better done in some different "more native" under "endgame AA", but this is relatively a small subset.
+* It does not require adding any opcodes, that would become dangling and useless in a post-EOA world.
+* It allows EOAs to temporarily convert themselves into contracts to be included in ERC-4337 bundles, in a way that's compatible with the existing `EntryPoint`.
+* Once this is implemented, [EIP-5003](./eip-5003.md) is "only one line of code": just add a flag to not set the code back to empty at the end.
+
+## Backwards Compatibility
+
+This EIP breaks the invariant that an account balance can only decrease as a result of transactions originating from that account. This has consequences for mempool design, and for other EIPs such as inclusion lists. However, these issues are common to any proposal that provides similar functionality, including EIP-3074.
+
+## Security Considerations
```
PowerStream3604, line 91: I assume it is, but just to confirm - this EIP doesn't allow storage modification of the EOA right?
e.g., SSTORE, etc

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+
+```
+rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, data, access_list, [[contract_code, y_parity, r, s], ...], signature_y_parity, signature_r, signature_s])
+```
+
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `16 * non-zero calldata bytes + 4 * zero calldata bytes` over each `contract_code`, plus `PER_CONTRACT_CODE_BASE_COST` times the length of the `contract_code` array.
+
+At the start of executing the transaction, for each `[contract_code, y_parity, r, s]` tuple:
+
+1. Let `signer = ecrecover(keccak(MAGIC + contract_code), y_parity, r, s]`.
```
Amxx, line 55: This means that a signature is reusable and non revocable. That can then be reused and encapsulated into other transactions. 
If the code is vulnerable (for any reason) the signer becomes vulnerable indefinitely

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+* `TX_TYPE` = `TBD`
+* `MAGIC` = `TBD`
+* `PER_CONTRACT_CODE_BASE_COST` = `5000`
+
+As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introduced with `TransactionType` = `TX_TYPE(TBD)`.
+
+The [EIP-2718](./eip-2718.md) `TransactionPayload` for this transaction is
+
+```
+rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, data, access_list, [[contract_code, y_parity, r, s], ...], signature_y_parity, signature_r, signature_s])
```
txgyy, line 48: A small suggestion: when `y_parity` is 0 or 1, it's a temporary setting code. When it's 2 or 3, it's `secp256r1` and a permanent setting code. When it's 4 or 5, it's `secp256k1` and a permanent setting code (currently not supported).

Your idea is very similar to mine, but you chose the `tx type`, so there's no need to add an opcode.
https://github.com/ethereum/EIPs/pull/8493/files

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `16 * non-zero calldata bytes + 4 * zero calldata bytes` over each `contract_code`, plus `PER_CONTRACT_CODE_BASE_COST` times the length of the `contract_code` array.
+
+At the start of executing the transaction, for each `[contract_code, y_parity, r, s]` tuple:
+
+1. Let `signer = ecrecover(keccak(MAGIC + contract_code), y_parity, r, s]`.
+2. Verify that the contract code of `signer` is empty.
+3. Set the contract code of `signer` to `contract_code`.
+
+At the end of the transaction, set the `contract_code` of each `signer` back to empty.
+
```
dangerousfood, line 60: So I presume transaction execution the `EXTCODEHASH` will return the `contract_code`, right?

AmadiMichael:
Nice proposals. I've a few questions though
- is `contract_code` a runtime or creation code?
- What happens gas wise if `contract_code` executes selfdestruct?

lightclient:
just want to remind everyone this PR is going to be merged shortly so the discussion should continue on in the discussions-to link https://ethereum-magicians.org/t/eip-set-eoa-account-code-for-one-transaction/19923

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+Specifically:
+
+* The contract code that users would need to sign could literally be existing ERC-4337 wallet code.
+* The "code pathways" that are used are code pathways that would, in many cases (though perhaps not all), continue to "make sense" in a pure-smart-contract-wallet world.
+* Hence, it avoids the problem of "creating two separate code ecosystems", because to a large extent they would be the same ecosystem. There would be some workflows that require kludges under this solution that would be better done in some different "more native" under "endgame AA", but this is relatively a small subset.
+* It does not require adding any opcodes, that would become dangling and useless in a post-EOA world.
+* It allows EOAs to temporarily convert themselves into contracts to be included in ERC-4337 bundles, in a way that's compatible with the existing `EntryPoint`.
+* Once this is implemented, [EIP-5003](./eip-5003.md) is "only one line of code": just add a flag to not set the code back to empty at the end.
+
```
Amxx, line 86:
```suggestion

### Opcode restriction

For security reasons, some opcodes should not be executed in the context of an EOA:

- `SSTORE` (0x55): Setting storage under an EOA breaks many assumptions. In particular storage set through a delegate transaction could cause issues if the accounts later "migrates" using [EIP-7377](./eip-7377.md) or similar. Additionally, storage may be a source of conflicts if a single EOA uses delegate transactions to target codes that interpret the storage layout under this account differently. For all these reasons, EOA should be forbiden from performing `SSTORE` in the context of a delegate transaction. If a delegate transaction performs a CALL, the target of the call is free to manipulate storage normally.

- `CREATE` (0xF0), `CREATE2` (0xF5) and `SELFDESTRUCT` (0xFF): There may be an expectation that transactions from a given sender should have consecutive nonces. This assumption would be broken if an EOA was able to execute one or multiple operations that alter the sender account's nonce. Consequently, EOA performing a delegate transaction should not be able to use the `CREATE`, `CREATE2` or `SELFDESTRUCT` opcodes. If a delegate transaction performs a CALL, the target of the call is free to create contracts normally.

Any attempts to execute one of these restricted operations in the context of one of the signers will instead throw an exception.
```

EIPS/eip-7702.md
```diff
@@ -0,0 +1,97 @@
...
+At the start of executing the transaction, for each `[contract_code, y_parity, r, s]` tuple:
+
+1. Let `signer = ecrecover(keccak(MAGIC + contract_code), y_parity, r, s]`.
+2. Verify that the contract code of `signer` is empty.
+3. Set the contract code of `signer` to `contract_code`.
+
+At the end of the transaction, set the `contract_code` of each `signer` back to empty.
+
+Note that the signer of any of the `contract_code` signatures, and the `tx.origin` of the transaction, are allowed to be different.
+
```
Amxx, line 62:
```suggestion

### Opcode restriction

For security reasons, some opcodes should not be executed in the context of an EOA:

- `SSTORE` (0x55): Setting storage under an EOA breaks many assumptions. In particular storage set through a delegate transaction could cause issues if the accounts later "migrates" using [EIP-7377](./eip-7377.md) or similar. Additionally, storage may be a source of conflicts if a single EOA uses delegate transactions to target codes that interpret the storage layout under this account differently. For all these reasons, EOA should be forbiden from performing `SSTORE` in the context of a delegate transaction. If a delegate transaction performs a CALL, the target of the call is free to manipulate storage normally.

- `CREATE` (0xF0), `CREATE2` (0xF5) and `SELFDESTRUCT` (0xFF): There may be an expectation that transactions from a given sender should have consecutive nonces. This assumption would be broken if code placed at an EOA's address was to execute one or multiple operations that alter the sender account's nonce. Consequently, EOA performing a delegate transaction should not be able to use the `CREATE`, `CREATE2` or `SELFDESTRUCT` opcodes. If a delegate transaction performs a CALL, the target of the call is free to create contracts normally.

Any attempts to execute one of these restricted operations in the context of one of the signers will instead throw an exception.

This restrictions only apply if the code placed at the signer's address is removed at the end of the transaction. If a flag is set to not set the code back to empty at the end of the transactions, then these restrictions do not apply.

```

EIPS/eip-7702.md
Amxx: https://github.com/ethereum/EIPs/pull/8527/files#r1593099283

EIPS/eip-7702.md
shemnon: Add a new field, don't overload yParity.  We don't need more obscure parsing rules.

EIPS/eip-7702.md
ryanschneider: If the code tuple contained the signer's nonce then the same revocation strategy as 3074 could be used.  While that solves revocation, it's unclear to me if the tuples being reusable is a feature or a bug though.

ryanschneider: 3074 includes the `invokerAddress` in the `AUTH` payload, so auth is only reusable by a single invoker, so perhaps the tuple also an address field to compare `tx.origin` against?

EIPS/eip-7702.md
AmadiMichael: I agree with restricting `CREATE`, `CREATE2` and `SELFDESTRUCT` but not sure what harm `SSTORE` poses. While the code stored at the address is deleted, all written storage slots can be cleared alongside it, that way it isn't persisted across different transactions. 

I'd argue that the gas cost of `SSTORE` and `SLOAD` be the same as that of `TLOAD` and `TSTORE` since they'd behave exactly the same way in this scenario.

wjmelements: I want to be able to `SSTORE`.

EIPS/eip-7702.md
txgyy: > Add a new field, don't overload yParity. We don't need more obscure parsing rules.

OK, sir. Do you agree with my idea?

txgyy: 1. If EIP7702 has already been implemented, the cost of adding a new algorithm isn't high.

2. The risk of permanently upgrading an EOA to a CA is significant (as EOAs carry historical baggage), but addresses generated by algorithms like secp256r1 are not EOAs, so permanently setting code carries no additional risk.

3. People need a more stable way to ensure deterministic deployments.

4. Better backward compatibility: When temporary code setting is prohibited and only permanent code setting is allowed, simply reject that type and enable a new type.

5. Prevent user errors: In the past, Ethereum used block heights to determine certain code behaviors, which wasn't user-friendly. We should directly reject user requests via a flag since upgrading an EOA to a CA is risky.

txgyy: > Add a new field, don't overload yParity. We don't need more obscure parsing rules.

Yeah, like this:
0x00: Temporary upgrade And secp256k1  
0x10: Permanent upgrade And secp256k1  
0x01: Temporary upgrade And secp256r1  
0x11: Permanent upgrade And secp256r1

EIPS/eip-7702.md
```diff
@@ -0,0 +1,98 @@
...
+```
+rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, data, access_list, [[contract_code, y_parity, r, s], ...], signature_y_parity, signature_r, signature_s])
+```
+
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `16 * non-zero calldata bytes + 4 * zero calldata bytes` over each `contract_code`, plus `PER_CONTRACT_CODE_BASE_COST` times the length of the `contract_code` array.
+
+At the start of executing the transaction, for each `[contract_code, y_parity, r, s]` tuple:
+
+1. Let `signer = ecrecover(keccak(MAGIC + contract_code), y_parity, r, s]`.
+2. Verify that the contract code of `signer` is empty.
```
drortirosh, line 57: I think it is important to mention that the outcome of this "verification" is only whether the signer's code is set.
That is: if the ecrecover succeeds (doesn't return zero), AND the returned `signer` address doesn't have code, only then its code is set to `contract_code`.
In any other circumstance, it is left unmodified.

g11tech (approved):
lgtm

EIPS/eip-7702.md
Amxx: I'd also want to use storage, but that would be breaking invariants specified in other EIP

There is a restriction that account with no code should have no storage. If someone can find a référénce that would be great.

AmadiMichael: True, but I think if all written slots are cleared at the end of the transaction it doesn't break any invariants.

During the transaction it doesn't break any invariants too since if the code length, code or codehash is queried, it returns a non-default value.

AmadiMichael: > There is a restriction that account with no code should have no storage. If someone can find a référénce that would be great.


Not so sure about this because a contract that sets storage at initialization but also returns 0 bytes at initialization has no bytecode but has > 0 storage slots set.

See here: https://twitter.com/AmadiMichaels/status/1641918500020137984?t=sy3aLWqR4we0tuA_kXV20w&s=19

sriharikapu (commented):
Thats more elaborate

ghost (commented):
I'm ok

ghost (commented):
Ok

MikeD198 (approved):

Muaytie23:
notification@email.meta.com

Muaytie23:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

aea97968cbae18dfc5fdc721e485f264fe1d67207383d826de9e762b307d450c  Portfile-0.13
d9c010c2cb9f2594894e542f59fd58c7c1effca56ec5fe2681e2cabc61cccec1  Portfile-0.14
8a6083b4851bff6dcde8bf246a9175db6747523f2e337817aa1de725f9f89e91  Portfile-0.14.1
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2.0.22 (GNU/Linux)

iQIcBAEBCAAGBQJkSEX2AAoJEGMtOgZYnaaxBbcQAKOk3G9kAEwmhL/b3nF+U4XQ
sUV804N8dV0geBUXRBT1dETT18i8kY1bzVemnD3Kdc6WjWfEIhVDM4crMW18g3p0
ZfGQWvRgO963ydKkmXIM8ADlrAYrSHcEMcjWb6KjiZ0fVctFjzLsXks3IrPH/UMb
JIPAbThWoFu2FGbZ5ezU0hbe2G750MRjoXKHpMskom5zZerqZCjYwjJCI6edLrM0
aEkDLdcX4Ar7WFo3v+l55DPn9xGEBJ8pUVsOUn7baiS1tjywrSvMRRr50WQNU79W
sg8M6GruTXTsZ0fOBv+vFsXxjJ7N5uEc+6UWPmbJC1UKy6IC9U3A8nHFdRDJF+DS
xPbMbcHPoZA3KSLElnm2ZSuimLWaSAWg9HuODDk8I8Q1yivU7gHtCXxoPHM12zXE
VVkOBAnmeovq5D4W+p7ikD0ky0rpfzUQLw5aVK15rOMrgSbM3DKZXP4q7K/v00HR
gHS1yD0qOAzvUrWcCITYLZKRY7kdlJ23bpYGQwYEXRbUbBGClXZle34ZmMImC74C
UudxdrXQKI115MAxAncRNnJwnbIgl+h0pnr3Ij5/aXltAvgdQV900OOF3CpsqD9f
Ny7BexhoJcXsKxOucZB9CHSc0Vq9RO+oOtoFOz8V+mi/n4ClSzzT/Bn1jNe4m5W2
pAW4MZ7mDgTFZjVT6zyK
=jmKG
-----END PGP SIGNATURE-----

EIPS/eip-7702.md
MikeD198: Ok

MikeD198 (approved):

mandervrik (commented):
ETH Rules!

octavio12345300 (commented):
list

draghtnight:
https://github.com/ethereum/EIPs/pull/8527

draghtnight (commented):
README.md

Ekene1998 (approved):
requires: 2718, 2930

Ekene1998 (commented):
EIPS/eip-7702.md

EIPS/eip-7702.md
```diff
@@ -0,0 +1,98 @@
...
+eip: 7702
+title: Set EOA account code for one transaction
+description: Add a new tx type that sets the code for an EOA during one transaction execution
+author: Vitalik Buterin (@vbuterin), Sam Wilson (@SamWilsn), Ansgar Dietrichs (@adietrichs), Matt Garnett (@lightclient)
+discussions-to: https://ethereum-magicians.org/t/eip-set-eoa-account-code-for-one-transaction/19923
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-05-07
+requires: 2718, 2930
```
525411, line 11: - [ ]

EIPS/eip-7702.md
Slink55: 1 more transaction

PR #8530, opened 2024-05-08
Update EIP-7623: Move to Withdrawn

wjmelements:
#### Changes
* Describe Gas Sheltering in Security Considerations section
* Move status to Withdrawn


nerolation:
I disagree. 
You say in the security section that one needs to consider the worst case blocks after users start playing the games you describe:

Have you ever checked what the worst case block size would then be?
If there's a user that needs EVM computation worth 20m gas and another who wants to add 10m gas worth of data to the chain, then, depite being irrelevant if it's one or two users, the max possible blocksize is still significantly reduced.

PR #8539, opened 2024-05-09
Update EIP-7702: Prohibit `SSTORE`

SamWilsn:


Amxx:
Today, you can create a contract, manipulate storage freely, and selfdestruct it before the end of the transaction. In that case, storage is cleared, effectivelly behaving like TSTORE.

Should we replicate that behavior, considering that the reset at the end of an EIP-7702 transaction is simlilar to a selfdestruct?

wjmelements:
SSTORE would be useful for persisting state across transactions. It won't be necessary for most use-cases, but I would prefer if it is possible. Some uses I can imagine:

* Marking nonces as used
* Approval
* Whitelist, Blacklist

LwkkyLDdubz:
> SSTORE would be useful for persisting state across transactions. It won't be necessary for most use-cases, but I would prefer if it is possible. Some uses I can imagine:
> 
> * Marking nonces as used
> * Approval
> * Whitelist, Blacklist

Reviewing cases

PR #8550, opened 2024-05-13
Update EIP-4762: reworked gas schedule from interop

gballet:
For reference: https://notes.ethereum.org/EHXrhQ8-QdeqnqLH3-KBuw?view

### DIFF
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -41,7 +41,7 @@ We define access events as follows. When an access event takes place, the access
 
 When:
 
- 1. a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `STATICALL`, `AUTHCALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, 
+ 1. a non-precompile address is the target of a `*CALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, 
  2. a non-precompile address is the target address of a contract creation whose initcode starts execution,
  3. any address is the target of the `BALANCE` opcode
  4. a _deployed_ contract calls `CODECOPY`
@@ -205,11 +205,11 @@ Reduce gas cost:
 
 |Constant |Value|
 |-|-|
-|`WITNESS_BRANCH_COST`	|1900|
+|`WITNESS_BRANCH_COST`|1900|
 |`WITNESS_CHUNK_COST`	|200|
 |`SUBTREE_EDIT_COST`	|3000|
-|`CHUNK_EDIT_COST`	    |500|
-|`CHUNK_FILL_COST`	    |6200|
+|`CHUNK_EDIT_COST`    |500|
+|`CHUNK_FILL_COST`    |6200|
 
 When executing a transaction, maintain four sets:
 
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -60,6 +60,7 @@ If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), 
 ```
 
 When calling `EXTCODEHASH`, process the access event:
+
 ```
 (address, 0, CODE_HASH_LEAF_KEY)
 ```
@@ -259,7 +260,7 @@ class AccessSubtree(Container):
     elements: BitVector[256]
 ```
 
-## Block-level operations
+### Block-level operations
 
 * Precompile accounts, system contract accounts and slots of a system contract that are accessed during a system call are considered warm.
 * When (and only when) calling a system contract _via a system call_, the code chunks and account should not appear in the witness.
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -235,6 +235,13 @@ When a **write** event of `(address, sub_key, leaf_key)` occurs, perform the fol
 
 Note that tree keys can no longer be emptied: only the values `0...2**256-1` can be written to a tree key, and 0 is distinct from `None`. Once a tree key is changed from `None` to not-`None`, it can never go back to `None`.
 
+Note that values should only be added to the witness if there is sufficient gas to cover their associated event costs.
+
+`CREATE*` and `*CALL` reserve 1/64th of the gas before the nested execution. Note that, following the behavior of access lists, the 1/64th of the gas is subtracted:
+
+ * **BEFORE** charging the witness costs when performing a `CALL`, `CODECALL`, `DELEGATECALL` or`STATICCALL`
+ * **AFTER** charging th witness costs when performing a `CREATE` or `CREATE2`
+
 ### Replacement for access lists
 
 We replace [EIP-2930](./eip-2930.md) access lists with an SSZ structure of the form:
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -39,23 +39,23 @@ We define access events as follows. When an access event takes place, the access
 
 #### Access events for account headers
 
-When a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, or is the target address of a contract creation whose initcode starts execution, process these access events:
+When:
+
+ 1. a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `STATICALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, 
+ 2. a non-precompile address is the target address of a contract creation whose initcode starts execution,
+ 3. any address is the target of the `BALANCE` opcode
+
+process this access events:
 
 ```
-(address, 0, VERSION_LEAF_KEY)
-(address, 0, NONCE_LEAF_KEY)
-```
-
-For `CALL`, `CALLCODE`, `DELEGATECALL`, `EXTCODESIZE`, `EXTCODECOPY`, process the additional event:
-```
-(address, 0, CODE_SIZE_LEAF_KEY)
+(address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 If a call is value-bearing (ie. it transfers nonzero wei), whether or not the callee is a precompile, process these two access events:
 
 ```
-(caller_address, 0, BALANCE_LEAF_KEY)
-(callee_address, 0, BALANCE_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
+(callee_address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 When calling `EXTCODEHASH`, process the access event:
@@ -66,24 +66,15 @@ When calling `EXTCODEHASH`, process the access event:
 When a contract is created, process these access events:
 
 ```
-(contract_address, 0, VERSION_LEAF_KEY)
-(contract_address, 0, NONCE_LEAF_KEY)
-(contract_address, 0, BALANCE_LEAF_KEY)
-(contract_address, 0, CODE_KECCAK_LEAF_KEY)
-(contract_address, 0, CODE_SIZE_LEAF_KEY)
+(contract_address, 0, BASIC_DATA_LEAF_KEY)
+(contract_address, 0, CODEHASH_LEAF_KEY)
 ```
 
-If the `BALANCE` opcode is called targeting some address, process this access event:
+If the `SELFDESTRUCT` opcode is called by some `caller_address` targeting some `target_address` (regardless of whether it’s value-bearing or not), process access events of the form:
 
 ```
-(address, 0, BALANCE_LEAF_KEY)
-```
-
-If the `SELFDESTRUCT` opcode is called by some caller_address targeting some target_address (regardless of whether it’s value-bearing or not), process access events of the form:
-
-```
-(caller_address, 0, BALANCE_LEAF_KEY)
-(target_address, 0, BALANCE_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
+(target_address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 If the `EXTCODEHASH` opcode is called targeting some address, process an access event of the form:
@@ -133,31 +124,21 @@ We define **write events** as follows. Note that when a write takes place, an ac
 When a nonzero-balance-sending `CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
 
 ```
-(sender, 0, BALANCE_LEAF_KEY)
-(recipient, 0, BALANCE_LEAF_KEY)
+(sender, 0, BASIC_DATA_LEAF_KEY)
+(recipient, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 When a contract creation is initialized, process these write events:
 
 ```
-(contract_address, 0, VERSION_LEAF_KEY)
-(contract_address, 0, NONCE_LEAF_KEY)
-```
-
-Only if the value sent with the creation is nonzero, also process:
-
-```
-(contract_address, 0, BALANCE_LEAF_KEY)
+(contract_address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 When a contract is created, process these write events:
 
 ```
-(contract_address, 0, VERSION_LEAF_KEY)
-(contract_address, 0, NONCE_LEAF_KEY)
-(contract_address, 0, BALANCE_LEAF_KEY)
-(contract_address, 0, CODE_KECCAK_LEAF_KEY)
-(contract_address, 0, CODE_SIZE_LEAF_KEY)
+(contract_address, 0, BASIC_DATA_LEAF_KEY)
+(contract_address, 0, CODEHASH_LEAF_KEY)
 ```
 
 #### Write events for storage
@@ -191,29 +172,22 @@ For `i` in `0 ... (len(code)+30)//31`.
 For a transaction, make these access events:
 
 ```
-(tx.origin, 0, VERSION_LEAF_KEY)
-(tx.origin, 0, BALANCE_LEAF_KEY)
-(tx.origin, 0, NONCE_LEAF_KEY)
-(tx.origin, 0, CODE_SIZE_LEAF_KEY)
-(tx.origin, 0, CODE_KECCAK_LEAF_KEY)
-(tx.target, 0, VERSION_LEAF_KEY)
-(tx.target, 0, BALANCE_LEAF_KEY)
-(tx.target, 0, NONCE_LEAF_KEY)
-(tx.target, 0, CODE_SIZE_LEAF_KEY)
-(tx.target, 0, CODE_KECCAK_LEAF_KEY)
+(tx.origin, 0, BASIC_DATA_LEAF_KEY)
+(tx.origin, 0, CODEHASH_LEAF_KEY)
+(tx.target, 0, BASIC_DATA_LEAF_KEY)
+(tx.target, 0, CODEHASH_LEAF_KEY)
 ```
 
 #### Write events
 
 ```
-(tx.origin, 0, NONCE_LEAF_KEY)
+(tx.origin, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 if `value` is non-zero:
 
 ```
-(tx.origin, 0, BALANCE_LEAF_KEY)
-(tx.target, 0, BALANCE_LEAF_KEY)
+(tx.target, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 ### Witness gas costs
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -185,7 +185,7 @@ if a transaction is a successful contract creation, make this extra access event
 (tx.origin, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-if `value` is non-zero:
+If `value` is non-zero:
 
 ```
 (tx.target, 0, BASIC_DATA_LEAF_KEY)
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -262,10 +262,15 @@ class AccessSubtree(Container):
 
 ### Block-level operations
 
-* Precompile accounts, system contract accounts and slots of a system contract that are accessed during a system call are considered warm.
-* When (and only when) calling a system contract _via a system call_, the code chunks and account should not appear in the witness.
-* The coinbase account is warmed before the block, and the whole account is accessed
-* Withdrawals are cold, but are added to the witness at no cost if they are the target of a transaction.
+None of:
+
+* Precompile accounts, system contract accounts and slots of a system contract that are accessed during a system call,
+* The coinbase account
+* Withdrawal accounts
+
+are warm at the start of a transaction.
+
+Note: When (and only when) calling a system contract _via a system call_, the code chunks and account should not appear in the witness.
 
 ### Account abstraction
 
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -52,31 +52,29 @@ process this access events:
 (address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), whether or not the callee is a precompile, process these additional two access events:
+Note: a non-value-bearing `SELFDESTRUCT` or `*CALL`, targetting a precompile, will not cause the `BASIC_DATA_LEAF_KEY` to be added to the witness.
+
+If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), whether or not the callee is a precompile, process this additional access event:
 
 ```
-(caller_address, 0, BASIC_DATA_LEAF_KEY)
-(callee_address, 0, BASIC_DATA_LEAF_KEY)
+(origin, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-When calling `EXTCODEHASH`, process the access event:
+Note: when checking for the existence of the callee, the existence check is done by validating that there is an extension-and-suffix tree at the corresponding stem, and does not rely on `CODEHASH_LEAF_KEY`.
+
+When calling `EXTCODEHASH` on a non-precompile target, process the access event:
 
 ```
-(address, 0, CODE_HASH_LEAF_KEY)
+(address, 0, CODEHASH_LEAF_KEY)
 ```
+
+Note that precompiles are excluded, as their hashes are known to the client.
 
 When a contract is created, process these access events:
 
 ```
 (contract_address, 0, BASIC_DATA_LEAF_KEY)
 (contract_address, 0, CODEHASH_LEAF_KEY)
-```
-
-If the `SELFDESTRUCT` opcode is called by some `caller_address` targeting some `target_address` (regardless of whether it’s value-bearing or not), process access events of the form:
-
-```
-(caller_address, 0, BASIC_DATA_LEAF_KEY)
-(target_address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 #### Access events for storage
@@ -117,11 +115,17 @@ We define **write events** as follows. Note that when a write takes place, an ac
 
 #### Write events for account headers
 
-When a nonzero-balance-sending `CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
+When a nonzero-balance-sending `*CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
 
 ```
-(sender, 0, BASIC_DATA_LEAF_KEY)
-(recipient, 0, BASIC_DATA_LEAF_KEY)
+(origin, 0, BASIC_DATA_LEAF_KEY)
+(target, 0, BASIC_DATA_LEAF_KEY)
+```
+
+if no account exists at `callee_address`, also process:
+
+```
+(target, 0, CODEHASH_LEAF_KEY)
 ```
 
 When a contract creation is initialized, process these write events:
@@ -133,7 +137,6 @@ When a contract creation is initialized, process these write events:
 When a contract is created, process these write events:
 
 ```
-(contract_address, 0, BASIC_DATA_LEAF_KEY)
 (contract_address, 0, CODEHASH_LEAF_KEY)
 ```
 
@@ -161,6 +164,8 @@ When a contract is created, process the write events:
 
 For `i` in `0 ... (len(code)+30)//31`.
 
+Note: since no access list existed for code up until this EIP, note that no warm costs are charged for code accesses.
+
 ### Transactions
 
 #### Access events
@@ -171,11 +176,6 @@ For a transaction, make these access events:
 (tx.origin, 0, BASIC_DATA_LEAF_KEY)
 (tx.origin, 0, CODEHASH_LEAF_KEY)
 (tx.target, 0, BASIC_DATA_LEAF_KEY)
-```
-
-if a transaction is a successful contract creation, make this extra access event before charging the costs for the bytecode storage:
-
-```
 (tx.target, 0, CODEHASH_LEAF_KEY)
 ```
 
@@ -266,6 +266,10 @@ class AccessSubtree(Container):
 * When (and only when) calling a system contract _via a system call_, the code chunks and account should not appear in the witness.
 * The coinbase account is warmed before the block, and the whole account is accessed
 * Withdrawals are cold, but are added to the witness at no cost if they are the target of a transaction.
+
+### Account abstraction
+
+TODO : still waiting on a final decision between 7702 and 3074
 
 ## Rationale
 
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -43,6 +43,11 @@ When a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECA
 
 ```
 (address, 0, VERSION_LEAF_KEY)
+(address, 0, NONCE_LEAF_KEY)
+```
+
+For `CALL`, `CALLCODE`, `DELEGATECALL`, `EXTCODESIZE`, `EXTCODECOPY`, process the additional event:
+```
 (address, 0, CODE_SIZE_LEAF_KEY)
 ```
 
@@ -51,6 +56,11 @@ If a call is value-bearing (ie. it transfers nonzero wei), whether or not the ca
 ```
 (caller_address, 0, BALANCE_LEAF_KEY)
 (callee_address, 0, BALANCE_LEAF_KEY)
+```
+
+When calling `EXTCODEHASH`, process the access event:
+```
+(address, 0, CODE_HASH_LEAF_KEY)
 ```
 
 When a contract is created, process these access events:
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -2,7 +2,7 @@
 eip: 4762
 title: Statelessness gas cost changes
 description: Changes the gas schedule to reflect the costs of creating a witness by requiring clients update their database layout to match.
-author: Guillaume Ballet (@gballet), Vitalik Buterin (@vbuterin), Dankrad Feist (@dankrad)
+author: Guillaume Ballet (@gballet), Vitalik Buterin (@vbuterin), Dankrad Feist (@dankrad), Ignacio Hagopian (@jsign), Tanishq Jasoria (@tanishqjasoria), Gajinder Singh (@g11tech)
 discussions-to: https://ethereum-magicians.org/t/eip-4762-statelessness-gas-cost-changes/8714
 status: Draft
 type: Standards Track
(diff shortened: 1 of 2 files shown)


EIPS/eip-4762.md
```diff
@@ -39,41 +39,42 @@ We define access events as follows. When an access event takes place, the access
...
 #### Access events for account headers
 
-When a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, or is the target address of a contract creation whose initcode starts execution, process these access events:
+When:
+
+ 1. a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `STATICALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, 
+ 2. a non-precompile address is the target address of a contract creation whose initcode starts execution,
+ 3. any address is the target of the `BALANCE` opcode
+
+process this access events:
```
gballet, line 48: clarify if `CODESiZE` needs to charge the warm costs

EIPS/eip-4762.md
```diff
@@ -39,47 +39,43 @@ We define access events as follows. When an access event takes place, the access
...
+(address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-When a contract is created, process these access events:
+If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), whether or not the callee is a precompile, process these additional two access events:
 
 ```
-(contract_address, 0, VERSION_LEAF_KEY)
-(contract_address, 0, NONCE_LEAF_KEY)
-(contract_address, 0, BALANCE_LEAF_KEY)
```
gabrocheleau, line 61: I'm wondering if converting balance from a `uint256` to a `uint128` might cause breaking changes in some contexts. To be backwards compatible, it also puts languages like Solidity in a position where they will have to treat balances as `uint256` (for e.g. the very commonly used `address.balance` returns a `uint256`.) even though the underlying data structure only contains a uint128. Ethereum's ecosystem also includes a variety of layer-2 solutions, sidechains, and off-chain protocols that may rely on the current uint256 account structure. Changing to uint128 could introduce incompatibilities with these systems.

Not saying we shouldn't do it, and I'm a fan of the simplification & more efficient packed data structure, but just pointing out these potential risks.

EIPS/eip-6800.md
```diff
@@ -115,11 +115,8 @@ Instead of a two-layer structure as in the Patricia tree, in the Verkle tree we
 
 | Parameter             | Value   |
 | --------------------- | ------- |
-| VERSION_LEAF_KEY      | 0       |
-| BALANCE_LEAF_KEY      | 1       |
-| NONCE_LEAF_KEY        | 2       |
+| BASIC_DATA_LEAF_KEY   | 0       |
 | CODE_KECCAK_LEAF_KEY  | 3       |
```
gballet, line 119:
```suggestion
| CODE_KECCAK_LEAF_KEY  | 1       |
```

EIPS/eip-6800.md
```diff
@@ -157,25 +154,26 @@ def get_tree_key(address: Address32, tree_index: int, sub_index: int):
...
 def get_tree_key_for_code_keccak(address: Address32):
     return get_tree_key(address, 0, CODE_KECCAK_LEAF_KEY)
-    
-# Backwards compatibility for EXTCODESIZE
-def get_tree_key_for_code_size(address: Address32):
-    return get_tree_key(address, 0, CODE_SIZE_LEAF_KEY)
 ```
 
-When any account header field is set, the `version` is also set to zero. The `code_keccak` and `code_size` fields are set upon contract creation.
+An account's `version`, `balance`, `nonce` and `code size` fields are packed in the value found at `BASIC_DATA_LEAF_KEY`:
```
jsign, line 165:
```suggestion
An account's `version`, `balance`, `nonce` and `code_size` fields are packed in the value found at `BASIC_DATA_LEAF_KEY`:
```

EIPS/eip-4762.md
```diff
@@ -39,47 +39,43 @@ We define access events as follows. When an access event takes place, the access
...
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
+(callee_address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-If the `BALANCE` opcode is called targeting some address, process this access event:
-
+When calling `EXTCODEHASH`, process the access event:
 ```
-(address, 0, BALANCE_LEAF_KEY)
+(address, 0, CODE_HASH_LEAF_KEY)
```
g11tech, line 64: so we need to remove isEmpty check? 

or we assume that if codehash witness is non null then it isn't empty? and for empty accounts it will come null

EIPS/eip-4762.md
```diff
@@ -251,6 +230,13 @@ When a **write** event of `(address, sub_key, leaf_key)` occurs, perform the fol
 
 Note that tree keys can no longer be emptied: only the values `0...2**256-1` can be written to a tree key, and 0 is distinct from `None`. Once a tree key is changed from `None` to not-`None`, it can never go back to `None`.
 
+Note that values should only be added to the witness if there is sufficient gas to cover their associated event costs.
```
g11tech, line 233: ahh nice

EIPS/eip-4762.md
```diff
@@ -268,6 +254,13 @@ class AccessSubtree(Container):
     elements: BitVector[256]
 ```
 
+## Block-level operations
+
+* Precompile accounts, system contract accounts and slots of a system contract that are accessed during a system call are considered warm.
```
g11tech, line 259: just to reconfirm this is only for the system call?

also system call is:
 i) system updates not related to a tx
 ii) a precompile executing system call code  (blockhash)

EIPS/eip-4762.md
```diff
@@ -268,6 +254,13 @@ class AccessSubtree(Container):
...
+* When (and only when) calling a system contract _via a system call_, the code chunks and account should not appear in the witness.
```
g11tech, line 260: this implies that stateless clients need to maintain system contracts . Also there is an unresolved issue of stateless clients not knowing if the contract has been actually deployed or not.

we should simply declare that stateless clients should assume activated system contracts deployed.

EIPS/eip-4762.md
```diff
@@ -268,6 +254,13 @@ class AccessSubtree(Container):
...
+* The coinbase account is warmed before the block, and the whole account is accessed
```
g11tech, line 261: whole account = basic data + code hash?

EIPS/eip-4762.md
jsign, line 64: In 4762 we have mixed namings with `CODE_HASH_LEAF_KEY` and `CODEHASH_LEAF_KEY`.
And in 6800 we use `CODE_KECCAK_LEAF_KEY`.

I'd say to change everything to `CODE_HASH_LEAF_KEY` or `CODEHASH_LEAF_KEY`.

EIPS/eip-4762.md
tanishqjasoria, line 259: should we add that the code chunks during system calls are not to be warmed? or it might cause additional confusion?

EIPS/eip-4762.md
```diff
@@ -268,6 +254,13 @@ class AccessSubtree(Container):
...
+* Withdrawals are cold, but are added to the witness at no cost if they are the target of a transaction.
```
tanishqjasoria, line 262: we should add the clarification about storage cells here.
If in a trasnaction, storage value is set and then reset to zero, we should insert zero to the tree and not leave it empty because the user paid the gas cost.
Basically if we charge CHUNK_FILL_COST, fill the chunk value.

tanishqjasoria (changes requested):

EIPS/eip-4762.md
g11tech: well if balance can never exceed 16 bytes it doesn't matter since clients stack can still treat it as 32 bytes and only writing to and fro the state is 16 bytes

EIPS/eip-4762.md
vbuterin: I think the big deciding factor is: when ethereum completes its shift to sha256 as its primary hash function (or even moves to something more stark-friendly), would this still be the keccak, or would it change (even if for new contracts or EOF only)? If still keccak, then call it CODE_KECCAK, if not, call it CODE_HASH.

EIPS/eip-4762.md
```diff
@@ -39,47 +39,43 @@ We define access events as follows. When an access event takes place, the access
...
+When:
 
-```
-(address, 0, VERSION_LEAF_KEY)
-(address, 0, CODE_SIZE_LEAF_KEY)
-```
+ 1. a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `STATICALL`, `AUTHCALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, 
+ 2. a non-precompile address is the target address of a contract creation whose initcode starts execution,
+ 3. any address is the target of the `BALANCE` opcode
+ 4. a _deployed_ contract calls `CODECOPY`
```
jsign, line 47: If you are in a deployed contract, isn't read access effect for the target already done by the respective `*CALL` making this clarification redundant?

EIPS/eip-4762.md
```diff
@@ -39,47 +39,43 @@ We define access events as follows. When an access event takes place, the access
...
 #### Access events for account headers
 
-When a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, or is the target address of a contract creation whose initcode starts execution, process these access events:
+When:
 
-```
-(address, 0, VERSION_LEAF_KEY)
-(address, 0, CODE_SIZE_LEAF_KEY)
-```
+ 1. a non-precompile address is the target of a `CALL`, `CALLCODE`, `DELEGATECALL`, `STATICALL`, `AUTHCALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode,
```
jsign, line 44: Do we want to compress `CALL`, `CALLCODE`, `DELEGATECALL`, `STATICALL`, `AUTHCALL` with a `*CALL` as done in other parts of this EIP?

EIPS/eip-4762.md
```diff
@@ -39,47 +39,43 @@ We define access events as follows. When an access event takes place, the access
...
-When a contract is created, process these access events:
+If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), whether or not the callee is a precompile, process these additional two access events:
 
 ```
-(contract_address, 0, VERSION_LEAF_KEY)
-(contract_address, 0, NONCE_LEAF_KEY)
-(contract_address, 0, BALANCE_LEAF_KEY)
-(contract_address, 0, CODE_KECCAK_LEAF_KEY)
-(contract_address, 0, CODE_SIZE_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
```
jsign, line 58: Considering now that `BASIC_DATA_LEAF_KEY` was read-accessed for `SELFDESTRUCT` by [this rule](https://github.com/ethereum/EIPs/pull/8550/files#diff-cd2aaa222faeb6081a41605386a509b3c4839285cd15f1b0ebe009318707914eR44), maybe we can remove this line?

EIPS/eip-4762.md
jsign: Lateral question: the consensus wasn't that we still would have to touch "VERSION" to be forward-compatible? If that's the case, we also need to add here `BASIC_DATA_LEAF_KEY` which can simplify all the questions above (since it comes with all the rest)?

EIPS/eip-4762.md
```diff
@@ -39,47 +39,43 @@ We define access events as follows. When an access event takes place, the access
...
+(contract_address, 0, CODEHASH_LEAF_KEY)
 ```
 
-If the `EXTCODEHASH` opcode is called targeting some address, process an access event of the form:
+If the `SELFDESTRUCT` opcode is called by some `caller_address` targeting some `target_address` (regardless of whether it’s value-bearing or not), process access events of the form:
 
 ```
-(address, 0, CODEHASH_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
+(target_address, 0, BASIC_DATA_LEAF_KEY)
```
jsign, line 78: If this rule applies if it's value-bearing or not, and for all cases (i.e: including precompiles), I think we might be able to remove `SELFDESTRUCT` from [this rule](https://github.com/ethereum/EIPs/pull/8550/files#diff-cd2aaa222faeb6081a41605386a509b3c4839285cd15f1b0ebe009318707914eR44)?

And also being mentioned from [this other rule](https://github.com/ethereum/EIPs/pull/8550/files#diff-cd2aaa222faeb6081a41605386a509b3c4839285cd15f1b0ebe009318707914eR55)?

EIPS/eip-4762.md
```diff
@@ -39,47 +39,43 @@ We define access events as follows. When an access event takes place, the access
...
+(contract_address, 0, BASIC_DATA_LEAF_KEY)
+(contract_address, 0, CODEHASH_LEAF_KEY)
 ```
 
-If the `EXTCODEHASH` opcode is called targeting some address, process an access event of the form:
+If the `SELFDESTRUCT` opcode is called by some `caller_address` targeting some `target_address` (regardless of whether it’s value-bearing or not), process access events of the form:
 
 ```
-(address, 0, CODEHASH_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
```
jsign, line 77: If this rule applies if it's value-bearing or not, and for all cases (i.e: including precompiles), I think we might be able to remove `SELFDESTRUCT` from [this rule](https://github.com/ethereum/EIPs/pull/8550/files#diff-cd2aaa222faeb6081a41605386a509b3c4839285cd15f1b0ebe009318707914eR44)?

EIPS/eip-4762.md
jsign, line 78: So, isn't this L74-L78 rule more general than [this one](https://github.com/ethereum/EIPs/pull/8550/files#diff-cd2aaa222faeb6081a41605386a509b3c4839285cd15f1b0ebe009318707914eR55) for `SELFDESTRUCT`?

If that's the case, we could remove `SELFDESTRUCT` mentioned in that linked rule, and only leave this L74-L78?

jsign: I'm starting to feel if wouldn't be easier for everyone to simply list rules for each independent instruction. Like `*CALL`, `SELFDESTRUCT`, etc, instead of trying to coalesce overlaps and re-clarify later with further rules some accessed extensions.

Yes, separating rules by instructions might have some redundancy but for implementors and testing it will be entirely clear what to do for each instruction. Now you have to read the whole document to see if in an extra scenario the same instruction you're interested in has another rule.

EIPS/eip-4762.md
```diff
@@ -123,31 +119,21 @@ We define **write events** as follows. Note that when a write takes place, an ac
 When a nonzero-balance-sending `CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
```
jsign, line 119: Should say `*CALL`?

EIPS/eip-4762.md
```diff
@@ -123,31 +119,21 @@ We define **write events** as follows. Note that when a write takes place, an ac
...
 
 When a contract is created, process these write events:
 
 ```
-(contract_address, 0, VERSION_LEAF_KEY)
-(contract_address, 0, NONCE_LEAF_KEY)
-(contract_address, 0, BALANCE_LEAF_KEY)
-(contract_address, 0, CODE_KECCAK_LEAF_KEY)
-(contract_address, 0, CODE_SIZE_LEAF_KEY)
+(contract_address, 0, BASIC_DATA_LEAF_KEY)
```
jsign, line 135: Is this line needed given that L126 does the same write-access and `initialized < created`?

EIPS/eip-4762.md
```diff
@@ -181,29 +167,22 @@ For `i` in `0 ... (len(code)+30)//31`.
...
 ```
 
 #### Write events
 
 ```
-(tx.origin, 0, NONCE_LEAF_KEY)
+(tx.origin, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 if `value` is non-zero:
```
jsign, line 182:
```suggestion
If `value` is non-zero:
```

EIPS/eip-4762.md
```diff
@@ -223,8 +202,8 @@ Reduce gas cost:
 |`WITNESS_BRANCH_COST`	|1900|
 |`WITNESS_CHUNK_COST`	|200|
 |`SUBTREE_EDIT_COST`	|3000|
-|`CHUNK_EDIT_COST`	|500|
-|`CHUNK_FILL_COST`	|6200|
+|`CHUNK_EDIT_COST`	    |500|
+|`CHUNK_FILL_COST`	    |6200|
```
jsign, line 206: Intended?

EIPS/eip-4762.md
```diff
@@ -251,6 +230,13 @@ When a **write** event of `(address, sub_key, leaf_key)` occurs, perform the fol
 
 Note that tree keys can no longer be emptied: only the values `0...2**256-1` can be written to a tree key, and 0 is distinct from `None`. Once a tree key is changed from `None` to not-`None`, it can never go back to `None`.
 
+Note that values should only be added to the witness if there is sufficient gas to cover their associated event costs.
+
+`CREATE*` and `*CALL` reserve 1/64th of the gas before the nested execution. Note that, following the behavior of access lists, the 1/64th of the gas is subtracted:
+
+ * **BEFORE** charging the witness costs when performing a `CALL`, `CODECALL`, `DELEGATECALL` or`STATICCALL`
```
jsign, line 237: Want to use `*CALL` to also contemplate `AUTHCALL`? (Also note a missing space after `or`)

EIPS/eip-4762.md
```diff
@@ -251,6 +230,13 @@ When a **write** event of `(address, sub_key, leaf_key)` occurs, perform the fol
...
+ * **AFTER** charging th witness costs when performing a `CREATE` or `CREATE2`
```
jsign, line 238:
```suggestion
 * **AFTER** charging the witness costs when performing a `CREATE` or `CREATE2`
```

EIPS/eip-4762.md
jsign, line 206: I can't add this comment in the intended line because there wasn't a change there.
I'm referring to lines L218-L220 and L226.

The intention in both places to say "skip it" is for gas cost charging. But the thing being skipped is also adding it to the underlying access list that are used for the witness creation. If those things also skip adding them to the list, then I think the witness wouldn't be correct?

EIPS/eip-4762.md
tanishqjasoria: yes, i kind of agree with this. Listing every opCode seperately will be a better idea. 
Right now its much easier to miss something.

tanishqjasoria:
i think we should add a section explaing that the account and storage slots that are warmed at the block level should be treated warm for all the transactions. For eg, coinbase and blockhash contract

EIPS/eip-4762.md
gballet: according to etherscan/coinmarketcap, the total ether supply peaked above 120M, which is $120 \times 10^{18 + 6}$ wei, which corresponds to about 87 bits if my math is correct. To reach a situation where this is a problem, we need to multiply the issuance by $2^{42}$ and this would have to land at the same address.

Mortiemi: > I'm wondering if converting balance from a `uint256` to a `uint128` might cause breaking changes in some contexts. To be backwards compatible, it also puts languages like Solidity in a position where they will have to treat balances as `uint256` (for e.g. the very commonly used `address.balance` returns a `uint256`.) even though the underlying data structure only contains a uint128. Ethereum's ecosystem also includes a variety of layer-2 solutions, sidechains, and off-chain protocols that may rely on the current uint256 account structure. Changing to uint128 could introduce incompatibilities with these systems.
> 
> 
> 
> Not saying we shouldn't do it, and I'm a fan of the simplification & more efficient packed data structure, but just pointing out these potential risks.

EIPS/eip-4762.md
gballet: but the adding of stuff in the access list is done in previous sections, whereas this is specific to the charging of costs.

EIPS/eip-4762.md
gballet: this is meant so that the warm costs are charged

EIPS/eip-4762.md
gballet: system calls, right now, are 2935 (post osaka), 4788 and 7002. whatever stuff they use that get accessed, should be considered warm.

I am actually close to reconsidering that, it's a special case.

EIPS/eip-4762.md
gballet: > this implies that stateless clients need to maintain system contracts 

it does not: because there is no obligation to call the contract's EVM, you are free to write the storage slots directly without having a local copy of the code.

>  Also there is an unresolved issue of stateless clients not knowing if the contract has been actually deployed or not.

I also don't see why this would be true. System contract activation is part of the pectra fork, so it should assume that.

> we should simply declare that stateless clients should assume activated system contracts deployed.

indeed. Or do you mean that it should be in the spec itself?

EIPS/eip-4762.md
gballet: yes

gballet: this is because the coinbase might not exist, so to avoid a lot of issues, we decide to touch the whole account so that it's created if it doesn't exist already.

EIPS/eip-4762.md
g11tech: > because there is no obligation to call the contract's EVM, you are free to write the storage slots directly without having a local copy of the code

for system contracts like 7002, even stateless focused clients like ethereumjs just does the EVM call to get the triggered exits  contract to get the exits. But this isn't a deal breaker just something clients need to bundle in verkle.

> System contract activation is part of the pectra fork

deployment is a manual exercise and not really part of client hardfork (that clients deploy this contract on forkblock which imo should have been a better mechanism to deploy the tx if not already deployed). I mean I already had done the mistake of not adding a contract in a local testnet and then debugging why things weren't working. So I would like things foolproof if possible. The problem could become acute as the number of such contracts grow.

> indeed. Or do you mean that it should be in the spec itself?

yes a simple mention of this would clearup the "ruleset" for the stateless client.

EIPS/eip-4762.md
g11tech: system calls totally ignore code chunks imo both for accesses and gas compute and we can clearly mention it to remove confusion

EIPS/eip-4762.md
g11tech: what if there is no tx that puts any value to it, and lets say there is no tips that get added to a non existent coinbase, then it should be 158 cleaned up in the end of block?

EIPS/eip-4762.md
g11tech: lateral question/observation to the above:

this is just because withdrawals are processed at the end of the block, and there is no actual cost charged to the coinbase for processing them?

EIPS/eip-6800.md
```diff
@@ -157,25 +154,26 @@ def get_tree_key(address: Address32, tree_index: int, sub_index: int):
...
+| Name        | Offset | Size |
+| ----------- | ------ | ---- |
+| `version`   | 0      | 1    |
+| `nonce`     | 4      | 8    |
+| `code_size` | 12     | 4    |
+| `balance`   | 16     | 16   |
+
+Bytes `1..3` are reserved for future use.
+
+When any account header field is set, the `version` field is also set to zero. The `code_keccak` and `code_size` fields are set upon contract or EoA creation.
```
g11tech, line 176: should we also add a section about 7702 (eoa temporary contract code) accesses?

EIPS/eip-4762.md
jsign: Post-Osaka-2935 has a long history. I'm not sure accesses to that history via the system contract should be considered warm?

jsign: From what I understand the intention is here, "system call" refers to any instruction (e.g: BLOCKHASH) or system contract execution called during a tx.

So "system call" has to contemplate both instructions and contracts, thus this is the reason why this concept emerges? (just trying to double-check).

EIPS/eip-6800.md
jsign: In this PR already added rules for `AUTHCALL`. So mentioning 7702 would be confusing?

Maybe it's better to not yet assume 3074 nor 7702 in any part of the PR (i.e: remove the existing `AUTHCALL` mentioning) and add a TODO saying `Add 7702/3074 to the spec`.

EIPS/eip-4762.md
g11tech: > reg your last paragraph, system updates from my perspective doesn't make sense to say are cold or warm, since we can't charge anyone for that?

yes I mean whateever system updates access warms for the txs. With this logic only the last entry in the ring buffer would be warmed and others cold so not the rest 8191 hashes

EIPS/eip-4762.md
```diff
@@ -39,47 +39,44 @@ We define access events as follows. When an access event takes place, the access
...
-(contract_address, 0, NONCE_LEAF_KEY)
-(contract_address, 0, BALANCE_LEAF_KEY)
-(contract_address, 0, CODE_KECCAK_LEAF_KEY)
-(contract_address, 0, CODE_SIZE_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
+(callee_address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-If the `BALANCE` opcode is called targeting some address, process this access event:
+When calling `EXTCODEHASH`, process the access event:
```
jsign, line 62: @gballet, I think we should add here "if the target isn't a precompile".

EIPS/eip-4762.md
```diff
@@ -251,6 +236,13 @@ When a **write** event of `(address, sub_key, leaf_key)` occurs, perform the fol
 
 Note that tree keys can no longer be emptied: only the values `0...2**256-1` can be written to a tree key, and 0 is distinct from `None`. Once a tree key is changed from `None` to not-`None`, it can never go back to `None`.
 
+Note that values should only be added to the witness if there is sufficient gas to cover their associated event costs.
+
+`CREATE*` and `*CALL` reserve 1/64th of the gas before the nested execution. In order to match the behavior of this charge with the pre-fork behavior of access lists: 
+
+ * this 1/64th of the gas is subtracted **AFTER** charging the witness costs when performing a `CALL`, `CODECALL`, `DELEGATECALL` or`STATICCALL`
```
jsign, line 243: Nit: in the case of `*CALL` (compared to `CREATE*`) it isn't a strict reservation, but a "cap check" of the max amount of gas you can use. 

If the `*CALL` parameter provided a `gas` parameter that is less 63/64, then _that_ `gas` is reserved.

```suggestion
 * this minimum 1/64th gas reservation is checked **AFTER** charging the witness costs when performing a `CALL`, `CODECALL`, `DELEGATECALL` or`STATICCALL`
```

Anyone reading the spec as it's today will understand the important clarification being made here. So I'm fine if we dismiss this nit.

jsign, line 243: @gballet, I wonder if we're missing an extra clarification here.

The `*CALL` variants that can be value-bearing, should also respect this rule?

Today in geth, we have this order (e.g: `makeCallVariantGasEIP4762(gasCall)`):
- Charge general `*CALL` costs. (`makeCallVariantGasEIP4762(...)`)
- Do the 1/64 stuff ((`gasCall(...)`)
- Charge BALANCE if `transfersValue` (`gasCall(...)`)

So I'm wondering if this requires a clarification, or was intentional. If was intentional it means that today in kaustinenv6-geth we need a change.

matkt:
do we have after the gas rework some logic to charge warm storage when stateless gas cost is zero ? for example https://github.com/ethereum/go-ethereum/compare/kaustinen-with-shapella...gballet:go-ethereum:kaustinen-with-shapella#diff-a6893c63a589eae0dd0ec9d5282aa504592c3024189cba62504b75c43a019658R57
if yes I think it will be nice to have that in the doc

EIPS/eip-4762.md
gballet: I don't think so: looking at 2929, there is no check whether the target is a precompile or not.

EIPS/eip-4762.md
```diff
@@ -181,29 +168,27 @@ For `i` in `0 ... (len(code)+30)//31`.
...
-(tx.target, 0, BALANCE_LEAF_KEY)
-(tx.target, 0, NONCE_LEAF_KEY)
-(tx.target, 0, CODE_SIZE_LEAF_KEY)
-(tx.target, 0, CODE_KECCAK_LEAF_KEY)
+(tx.origin, 0, BASIC_DATA_LEAF_KEY)
+(tx.origin, 0, CODEHASH_LEAF_KEY)
+(tx.target, 0, BASIC_DATA_LEAF_KEY)
+```
+
+if a transaction is a successful contract creation, make this extra access event before charging the costs for the bytecode storage:
```
matkt, line 176: could we clarify if it is before this one https://github.com/ethereum/EIPs/blob/29da7466816d023031fe0171e5adc66204fc6d87/EIPS/eip-4762.md?plain=1#L152 ?

matkt: do we have to access it again even if we did https://github.com/ethereum/EIPs/blob/29da7466816d023031fe0171e5adc66204fc6d87/EIPS/eip-4762.md?plain=1#L72 ?

EIPS/eip-4762.md
jsign: I believe EIP-2929 rules are because you're supposed to load the whole account in one go.

What this rule is saying in this EIP is that adding a precompile codehash to the witness makes sense, which I think it doesn't? In other parts of this EIP we have rules that exclude precompiles.

EIPS/eip-4762.md
matkt: I agree, it's hard to follow specs like that, I also understand the idea of ​​having a concise EIP

EIPS/eip-4762.md
```diff
@@ -39,47 +39,44 @@ We define access events as follows. When an access event takes place, the access
...
-(contract_address, 0, CODE_SIZE_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
+(callee_address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-If the `BALANCE` opcode is called targeting some address, process this access event:
+When calling `EXTCODEHASH`, process the access event:
 
 ```
-(address, 0, BALANCE_LEAF_KEY)
```
jsign, line 69: Shouldn't we also touch the `VERSION` thus `BASIC_DATA_LEAF_KEY`?

EIPS/eip-4762.md
gballet:
```suggestion
When calling `EXTCODEHASH` on a contract that is not a precompile, process the access event:
```

EIPS/eip-4762.md
gballet: TODO: add version

EIPS/eip-4762.md
gballet: TODO: add rationale for this

EIPS/eip-4762.md
gballet: we decided to touch `BASIC_DATA_LEAF_KEY`

gballet: using `CODEHASH` for future-proofiness

EIPS/eip-4762.md
gballet: if value-bearing + precompile : touch balance, if precompile + not value bearing, don't touch it (just like a call)

EIPS/eip-4762.md
gballet:
```suggestion
```

EIPS/eip-4762.md
```diff
@@ -39,47 +39,44 @@ We define access events as follows. When an access event takes place, the access
...
+(contract_address, 0, CODEHASH_LEAF_KEY)
 ```
 
-If the `EXTCODEHASH` opcode is called targeting some address, process an access event of the form:
+If the `SELFDESTRUCT` opcode is called by some `caller_address` targeting some `target_address` (regardless of whether it’s value-bearing or not), process access events of the form:
 
 ```
-(address, 0, CODEHASH_LEAF_KEY)
+(caller_address, 0, BASIC_DATA_LEAF_KEY)
+(target_address, 0, BASIC_DATA_LEAF_KEY)
```
gballet, line 79: Touching the code hash as well in case the target account is created.

```suggestion
(target_address, 0, BASIC_DATA_LEAF_KEY)
(target_address, 0, CODEHASH_LEAF_KEY)
```

EIPS/eip-4762.md
gballet: we decided to not warm anything.

EIPS/eip-4762.md
```diff
@@ -181,29 +168,27 @@ For `i` in `0 ... (len(code)+30)//31`.
...
-(tx.origin, 0, NONCE_LEAF_KEY)
-(tx.origin, 0, CODE_SIZE_LEAF_KEY)
-(tx.origin, 0, CODE_KECCAK_LEAF_KEY)
-(tx.target, 0, VERSION_LEAF_KEY)
-(tx.target, 0, BALANCE_LEAF_KEY)
-(tx.target, 0, NONCE_LEAF_KEY)
-(tx.target, 0, CODE_SIZE_LEAF_KEY)
-(tx.target, 0, CODE_KECCAK_LEAF_KEY)
+(tx.origin, 0, BASIC_DATA_LEAF_KEY)
+(tx.origin, 0, CODEHASH_LEAF_KEY)
```
jsign, line 172: If we have a tx (non-value-bearing) targeting a precompile or system contract, we don't need to touch BASIC_DATA. (i.e: stateless client knows the code-size/hash)

We can still argue that this rule should be avoided to avoid a border case. But considering we didn't do any similar simplification for precompiles/system contracts in other places in the EIP (i.e. we avoid adding in the witness for them as much as possible), then maybe not doing it here is actually _adding_ a border case (we have to remember this is an exception), not removing it.

No strong opinions, just surfacing this.
cc @gballet @g11tech @gabrocheleau @tanishqjasoria

EIPS/eip-4762.md
```diff
@@ -181,29 +168,27 @@ For `i` in `0 ... (len(code)+30)//31`.
...
+(tx.target, 0, BASIC_DATA_LEAF_KEY)
```
gballet, line 173: talk with @jsign: also add the code hash so that it's warmed

gballet: note that this collapses the next paragraph

morph-dev:
Since we are cleaning up the spec, I believe we can simplify `pedersen_hash` function as well.

> ```py
> def pedersen_hash(inp: bytes) -> bytes32:
>     assert len(inp) <= 255 * 16
>     # Interpret input as list of 128 bit (16 byte) integers
>     ext_input = inp + b"\0" * (255 * 16 - len(inp))
>     ints = [2 + 256 * len(inp)] + \
>            [int.from_bytes(ext_input[16 * i:16 * (i + 1)], 'little') for i in range(255)]
>     return compute_commitment_root(ints).hash_point_to_bytes()
> ```

To my understanding, it's only used with input that is exactly 64 bytes long: 32 bytes for address + 32 bytes for tree_index. I understand the idea to be more generic, but does it have to be?

I propose to simplify it to something like this (if we want to keep identical functionality):

```py
def pedersen_hash(address: Address32, tree_index: int) -> bytes32:
    marker = 16386 # 2 + 256 * 64
    ints = [marker,
            int.from_bytes(address[:16], 'little'),
            int.from_bytes(address[16:], 'little'),
            tree_index % (2 ** 128),
            tree_index / (2 ** 128)] + \
           [0] * 251
    return compute_commitment_root(ints).hash_point_to_bytes()
```

I believe that we can simplify it further and just use `2` as the marker.

gballet:
> To my understanding, it's only used with input that is exactly 64 bytes long: 32 bytes for address + 32 bytes for tree_index. I understand the idea to be more generic, but does it have to be?

What you describe is already implemented in e.g. geth but is an optimization. Simplifying it would make it more specific, when this describes a hash function that could be used for more things (and it _was_ in previous versions of this spec, so presumably it can be used even more in the future).

> 
> I propose to simplify it to something like this (if we want to keep identical functionality):
> 
> ```python
> def pedersen_hash(address: Address32, tree_index: int) -> bytes32:
>     marker = 16386 # 2 + 256 * 64
>     ints = [marker,
>             int.from_bytes(address[:16], 'little'),
>             int.from_bytes(address[16:], 'little'),
>             tree_index % (2 ** 128),
>             tree_index / (2 ** 128)] + \
>            [0] * 251
>     return compute_commitment_root(ints).hash_point_to_bytes()
> ```

this is assuming `tree_index` is little-endian

> 
> I believe that we can simplify it further and just use `2` as the marker.

if you assume a fixed input, sure. But if not, [a b c d 0 0 0 0 ... ] and [a b c d] would have the same hash.

Gabo139 (approved):

EIPS/eip-4762.md
fjl: I think we should include a mechanism to handle balances larger than 128bit, even if it is not used. The overflow balance could be stored in an 'additional balance leaf' which will be accessed if the 128bit one is maxed out. This makes the gas cost for reading balance dynamic though, which could be another source of contract bugs, though we wouldn't be hitting this in practice on eth mainnet.

PR #8561, opened 2024-05-15
Update EIP-7702: refine based on discussions

lightclient:
Based on some discussions I've made the following updates:

* deploy code via template (copy from existing address) instead of specifying code directly
* reduced the per auth base cost to `2500`, mostly to account for reading the template account
* sign over the template address instead of code

Additionally, I have added a lot of rationale for why decisions were made.


EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
+| -------------------- | ------ |
+| `SET_CODE_TX_TYPE`   | `0x04` |
+| `MAGIC`              | `0x05` |
+| `PER_AUTH_BASE_COST` | `2500` |
 
-As of `FORK_BLOCK_NUMBER`, a new [EIP-2718](./eip-2718.md) transaction is introduced with `TransactionType` = `TX_TYPE(TBD)`.
+### Set Code Transaction
 
-The [EIP-2718](./eip-2718.md) `TransactionPayload` for this transaction is
+We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction", where the `TransactioType` is `SET_CODE_TX_TYPE` and the `TransactionPayload` is the RLP serialization of the following:
```
vbuterin, line 38:
```suggestion
We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction", where the `TransactionType` is `SET_CODE_TX_TYPE` and the `TransactionPayload` is the RLP serialization of the following:
```

EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
+
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `PER_CONTRACT_CODE_BASE_COST * authorization list length`.
 
 ## Rationale
 
-### Conversion of EIP-3074 use cases
+### No initcode
+
+Running initcode is not desirable for many reasons. The chief concern is it's unnatural. Initcode is intended to initialize and deploy a contracts. With this EIP, it will take on a new role of determine whether it is appropriate to deploy code to the EOA. Suppose a user only wants code deployed to their account if they also have an operation bundled with the general transaction calldata. This gives EOA a unique power to control when and what code executes in their account. Although EIP-7702 as written still allows this to a degree, the lack of programmability in the decision will force wallets to not sign many authorization tuples and instead focus signing only a tuple pointing to a configurable proxy. This affords EOAs a similar experience to smart contract wallets
```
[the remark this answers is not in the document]
lightclient: After the number is assigned it's possible to self ref in the EIP

EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
 
 ## Rationale
 
-### Conversion of EIP-3074 use cases
+### No initcode
+
+Running initcode is not desirable for many reasons. The chief concern is it's unnatural. Initcode is intended to initialize and deploy a contracts. With this EIP, it will take on a new role of determine whether it is appropriate to deploy code to the EOA. Suppose a user only wants code deployed to their account if they also have an operation bundled with the general transaction calldata. This gives EOA a unique power to control when and what code executes in their account. Although [EIP-7702](./eip-7702.md) as written still allows this to a degree, the lack of programmability in the decision will force wallets to not sign many authorization tuples and instead focus signing only a tuple pointing to a configurable proxy. This affords EOAs a similar experience to smart contract wallets
```
smartprogrammer93, line 73:
```suggestion
Running initcode is not desirable for many reasons. The chief concern is it's unnatural. Initcode is intended to initialize and deploy contracts. With this EIP, it will take on a new role of determine whether it is appropriate to deploy code to the EOA. Suppose a user only wants code deployed to their account if they also have an operation bundled with the general transaction calldata. This gives EOA a unique power to control when and what code executes in their account. Although [EIP-7702](./eip-7702.md) as written still allows this to a degree, the lack of programmability in the decision will force wallets to not sign many authorization tuples and instead focus signing only a tuple pointing to a configurable proxy. This affords EOAs a similar experience to smart contract wallets
```

EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
+
+This is not to say that revocation isn't critical. It is simply that smart contracts and EOA wallets are capable of providing the same functionality. Even if the there is no cost to the user for in-protocol revocation, it is undue complexity in the core protocol. We must resist all except which is absolutely necessary to encode in the core protocol.
+
+### Setting code as `tx.origin`
+
+Allowing `tx.origin` to set code enables simple transaction batching, where the sender of the outer transaction would be the signing account. The ERC-20 approve-then-transfer pattern, which currently requires two separate transactions, could be completed in a single transaction with this proposal.
+
+`AUTH` allows for signatures to be signed by `tx.origin`. For any such signatures, subsequent `AUTHCALL`s have `msg.sender == tx.origin` in their first layer of execution. Without EIP-7702, this situation can only ever arise in the topmost execution layer of a transaction. This EIP breaks that invariant and so affects smart contracts containing `require(msg.sender == tx.origin)` checks. This check can be used for at least three purposes:
```
Eikix, line 119: Why are we mentioning AUTH and AUTHCALL here?

EIPS/eip-7702.md
```diff
@@ -88,11 +151,33 @@ Specifically:
...
+The following is a non-exhaustive list of checks/pitfalls/conditions that delegate contracts *should* be wary of and require a signature over from the account's authority:
+
+ * Replay protection -- (ex. a nonce) should be implemented by the delegate and signed over. Without it, a malicious actor can reuse a signature, repeating its effects.
+ * `value` -- without it, a malicious sponsor could cause unexpected effects in the callee.
+ * `gas` -- without it, a malicious sponsor could cause the callee to run out of gas and fail, griefing the sponsee.
+ * `target` / `calldata` -- without them, a malicious actor may call arbitrary functions in arbitrary contracts.
+
+A poorly implemented delegate can *allow a malicious actor to take near complete control over a signer's EOA*.
+
+### Setting code as `tx.origin`
```
Eikix, line 169: What does that mean? Setting the code pointer address as tx.origin?

EIPS/eip-7702.md
lightclient: i lifted some of this rationale from 3074, i fixed this.

EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
-2. Verify that the contract code of `signer` is empty.
-3. Set the contract code of `signer` to `contract_code`.
-4. Add the `signer` account to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
+#### Behavior
 
-At the end of the transaction, set the `contract_code` of each `signer` back to empty.
+At the start of executing the transaction, for each `[address, y_parity, r, s]` tuple:
 
-Note that the signer of any of the `contract_code` signatures, and the `tx.origin` of the transaction, are allowed to be different.
+1. Let `authority = ecrecover(keccak(MAGIC || address), y_parity, r, s]`.
```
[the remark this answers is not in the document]
lightclient: IMO it is better to do ECDSA on keccak everywhere and transition everything at once than it is to have two different formats to support in parallel.

EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
+
+Writing safe and bug-free smart contracts is difficult. However, not all code paths are create equal. There are certain areas of code that are both simple and cannot fail. Revocation logic is one. Authentication logic is another. These are the core components of a wallet. One can imagine extending a wallet further with plugins in way were failure is more tolerable. Suppose a user adds a handful of plugins which bit rot over time. Revoking that plugin should be simple and possible via the core wallet implementation.
+
+It's important to note that revocation is only useful as a mitigation against bugs that have not yet been found. As soon as a vulnerability exists in a wallet, the funds in the account must be considered as MEV and value of the account will be auctioned off.
+
+This is not to say that revocation isn't critical. It is simply that smart contracts and EOA wallets are capable of providing the same functionality. Even if the there is no cost to the user for in-protocol revocation, it is undue complexity in the core protocol. We must resist all except which is absolutely necessary to encode in the core protocol.
```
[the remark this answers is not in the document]
lightclient: looks like a typo there, but in general, even if there is no UX hit for in-protocol revocation (max nonce format), the complexity in the protocol isn't worth it

EIPS/eip-7702.md
lightclient: Yes. I adjusted the title to be a bit clearer.

EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
+
+Additionally, initcode in transaction tends to be propagated inside the transaction. That means it would need to be included in the authorization tuple and signed over. The minimum initcode would be around 15 bytes and that would simply copy the contract code from an external address. The total cost would be `16 * 15 = 240` calldata cost, plus the [EIP-3860](./eip-3860.md) cost of `2 * 15 = 30`, plus the runtime costs of around `150`. So nearly `500` additional gas would be spent simply preparing the account; and even more likely, 1200+ gas if not copying from an external account.
+
+### Creation by template
+
+Initcode or not, there is a question of how users should specify the code they intend to run in their account. The two main options are to specify the bytecode directly in the transaction or to specify a pointer to the code. The simplest pointer would just the address of some code deployed on-chain.
+
+The cost analysis makes the answer clear. The smallest proxy would be around 50 bytes and an address is 20 bytes. The 30 byte difference provides no useful additional functionality and will be inefficiently replicated billions of time on the chain.
```
SamWilsn, line 81:
```suggestion
The cost analysis makes the answer clear. The smallest proxy would be around 50 bytes and an address is 20 bytes. The 30 byte difference provides no useful additional functionality and will be inefficiently replicated billions of times on the chain.
```

derekchiang:
I just want to clarify that, given [the discussions on eth magicians](https://ethereum-magicians.org/t/eip-7702-set-eoa-account-code-for-one-transaction/19923), it's clear that many (most) of the updates in this PR are still in contention (even among the EIP authors), so I personally don't think that this is a good time to merge this PR, considering that we will have another ACDE meeting coming up to discuss 7702 and hopefully align on some of these potential updates.

Merging this PR now might mislead the core devs and other community members into believing that decisions have already been made one way or the other, when in reality we are all still deliberating.

EIPS/eip-7702.md
```diff
@@ -88,11 +155,33 @@ Specifically:
 
 ## Backwards Compatibility
```
drortirosh, line 156: Missing from the list of broken assumptions is signature malleability: 
Specifically, a user might sign an unverified experimental 7702 contract on a testnet under the assumption (which was correct until now) that it is a sandbox, and nothing affects other network.
This becomes disastrous, as this contract can be used anywhere - and currently is un-revocable.
The user never opt-in to this new feature: he merely continued his work with an existing (though updated) wallet.

smartprogrammer93: @drortirosh chainId variable will be added to the signature. Also, this type of signed data has a specific magic. Wallets are expected to warn the user heavily about what they are about to sign.

lightclient: Signing an unverified / experimental 7702 contract would never happen with a wallet that correctly implements 7702.

drortirosh: I don't have any problem with "wallet that correctly implements the standard".

Leaving the chainid out of signature EIP breaks the invariant that "testnet are sandbox and can't affect any other network".
Please add that text into your security consideration, if you think that's acceptable.

drortirosh: > Signing an unverified / experimental 7702 contract would never happen with a wallet that correctly implements 7702.

The only way to implement this is by completely blocking (by the wallet) from ever signing anything un-approved by the wallet, not even with a warning message.
That's the only way you can block this concern.

Even with a warning message such as:

> Beware of this signature: it opens your account to malicious operations not only on this network, but on any other network, and this operation is un-revocable. are you sure you want to continue

And I'm afraid that on testnets, some people will STILL sign this message, as the current narrative is "testnets are sandbox"

lightclient: It would be a nice invariant to have, but it's not our current reality. The safety we have relies on apps and wallets to implement things in a safe manner.

At the protocol level, chain ID is already optional. If the wallets correctly implement chain ID by encoding it in the signature, it can safe. But there is nothing forcing them to

For applications and smart contract wallets, anything goes. There are recommendations, such as 
in permit. It is *recommended* implementors include chain id in the domain separator for 712. If tokens implement this correctly, users are okay.

Smart contract wallets are similar. 4337 recommends wallets sign over the chain id, yet it isn't enforced.

712 is a good example by itself. It is used by many on-chain applications for authorization. Chain id is an optional field enforced only at the application level. Yet most users of 712 do sign over the chain id

We can't from the protocol up ensure safety for users. The best we can do is build a robust platform for developers to build safe tools. If the layers above do not uphold their responsibility then the number of ways users will be taken advantage of will be innumerable.

The point is that signing over the chain id is important, but enforcing it at the protocol level is not. Applications and wallet devs have clearly shown they are capable of enforcing it in the signatures they process.

drortirosh: > At the protocol level, chain ID is already optional. If the wallets correctly implement chain ID by encoding it in the signature, it can safe. But there is nothing forcing them to

After EIP-155, chainid is **mandatory** - that is, it is optional to sign, but the chain itself (most if not all evm-based chains today) enforce it at the protocol level.

> Smart contract wallets are similar. 4337 recommends wallets sign over the chain id, yet it isn't enforced.

ERC-4337 passes **userOpHash** to the account to sign, which already embed the **chainid** inside (wallet implementation MAY ignore it, and sign whatever it likes (or even not sign at all) - but it is important that no **off-chain** entity can decide, only the smart contract itself)

> We can't from the protocol up ensure safety for users. 

We can't **ensure** - but we should **reduce** the risk as much as possible.

lightclient: EIP-155 is not mandatory.

> The currently existing signature scheme using v = 27 and v = 28 remains valid and continues to operate under the same rules as it did previously.

If your wallet doesn't support it correctly, you could unknowingly sign away your assets on another chain.

drortirosh: > EIP-155 chain id is not mandatory.
Is it? Can you submit a non-EIP-155 transaction on mainnet? (or any other chain, for that matter?)
Can you deploy deterministically [Nick's determinisitic deployer](https://github.com/Arachnid/deterministic-deployment-proxy) on new networks? 
All new networks block such "malleable transactions" - and rightly so.
And in order to have this deployer on the same address across networks, they have to submit a "nonstandard state change" (e.g. a fork) to allow this specific non-EIP-155 transaction get through.

This is because malleable transactions are dangerous, and we no longer trust applications or wallets to take this decision.

lightclient: Which networks are banning chain id 0?

drortirosh: > Which networks are banning chain id 0?

The question is, on which network you can submit a non-EIP-155 transaction? All block builders on all networks I know of block them.
That's a de-facto standard (that all TXs are EIP-155).

drortirosh: I'm saying that there is an implicit assumptions today by users, that this EIP breaks:
1. That signing data on testnet stays in testnet and doesn't affect other networks.
2. That signing anything has a one-shot effect, not long-lasting (or permanent)

And I think that breaking these assumptions AT LEAST deserves being mentioned in the "security consideration" (or backward compatibility) section.
After specifying them, you can add your argument why you think it is OK to break them - but at least acknowledge that we're introducing a network change that alters existing users assumptions.
This is not a change that affect new users that use those new features, or new wallets: it affect all existing users that happen to upgrade their existing wallet, and are unaware of such changes.

lightclient: > That signing data on testnet stays in testnet and doesn't affect other networks.

Again, this is not correct. The protocol allows for non EIP-155 txs to be replayed on any chain. If users think they're safe because the wallet says they're on a testnet today, it's because apps, wallets, and RPC providers have done well to create this feeling of security. They could have just as easily continued signing and allowing txs w/o chain id baked in. I have no reason to believe post-7702 things will be any different.

> That signing anything has a one-shot effect, not long-lasting (or permanent)

This would be true for 7702 auths. Happy to put reword the security considerations if it is not clear there.

Jrachman:
Some thoughts I'd like to bring up as I review this revision:

> In the case of multiple tuples for the same authority, set the code specified by the address in the first occurrence. ([ref](https://github.com/lightclient/EIPs/blob/155baa3ff0b5853a46ea92ba769ff0846e9a7593/EIPS/eip-7702.md?plain=1#L63))

So this means that if we have `tupleA` and `tupleB` where `tupleA` comes first in the list and both `ecrecover` calls come out to be `authority`, then the code of `authority` would be set to `tupleA` and `tupleB` would be ignored?

> Furthermore, specifying code directly would again make it possible for EOAs to have a new, unique ability to execute arbitrary code specified in the transaction calldata. ([ref](https://github.com/lightclient/EIPs/blob/155baa3ff0b5853a46ea92ba769ff0846e9a7593/EIPS/eip-7702.md?plain=1#L87))

Isn't this a good thing? I believe the context this is in makes it sound like a bad thing. I bias towards maximal flexibility. This "creation by template" doesn't necessarily mean that what the EOA can sign/execute is constrained right? It is just choosing the way that code should be signed over?

> The main families of instructions where a ban was considered were storage related and contract creation related. ([ref](https://github.com/lightclient/EIPs/blob/155baa3ff0b5853a46ea92ba769ff0846e9a7593/EIPS/eip-7702.md?plain=1#L93))

Potentially dumb question but what happens if an SSTORE opcode is called within a 7702 transaction? Would it just not do anything or act kind of like memory storage because EOA's don't have built-in storage? Also, what are the unnecessary inefficiencies with an external storage contract? - would love to identify them to keep note of them.

> A hotly debated element of this EIP is the need for in-protocol revocation. ([ref](https://github.com/lightclient/EIPs/blob/155baa3ff0b5853a46ea92ba769ff0846e9a7593/EIPS/eip-7702.md?plain=1#L109))

I wrote some thoughts [here](https://ethereum-magicians.org/t/eip-7702-set-eoa-account-code-for-one-transaction/19923/154?u=julrach) around time-based revocation. Not sure if it is at all useful here but thought I'd throw it into the ring.

> The following is a non-exhaustive list of checks/pitfalls/conditions that delegate contracts should be wary of and require a signature over from the account's authority: ([ref](https://github.com/lightclient/EIPs/blob/155baa3ff0b5853a46ea92ba769ff0846e9a7593/EIPS/eip-7702.md?plain=1#L158))

Would this list of params be signed over in the `data` field of the 7702 transaction? Thinking about this like it's what you sign over in `execData` for 3074...

Mortiemi (approved):

frangio:
If I produce and broadcast one 7702 signature without nonce, I can never again use 7702 without the risk of being frontrun with my previous signature to prevent the new code from being deployed on my account. This does seem like something the protocol should prevent.

smartprogrammer93:
> If I produce and broadcast one 7702 signature without nonce, I can never again use 7702 without the risk of being frontrun with my previous signature to prevent the new code from being deployed on my account. This does seem like something the protocol should give tools to prevent.

I am not sure how would be front run. If you send a 7702 tx that you control what authorizations go inside it. Even if there was a 7702 tx before, it wont affect how your tx is executed.

frangio:
Right, I was thinking of a bundle submitted by a sponsor, and "frontrunning" in the sense of including an authorization before another in a 7702 tx authorization list.

I guess that clarifies why it shouldn't be considered a problem? A sponsor can always just choose not to include your signed authorization, so including two of your authoriations as a way to censor one of them is not really a new capability.

smartprogrammer93:
@lightclient can we specify exactly the behavior in case the validations of an authorization fails? my assumption is that it gets ignored but the tx goes through regardless. if that is not the case and tx should fail, do we spend all gas? seems excessive.

yperbasis:
Let's clarify how null nonce (as opposed to zero) is encoded in `authorization_list`. I suggest to encode it as an empty RLP list to differentiate it from a zero nonce, which is encoded as an empty RLP string per the spec.

DONALDSULLENDER (approved):

EIPS/eip-7702.md
```diff
@@ -23,63 +23,126 @@ There is a lot of interest in adding short-term functionality improvements to EO
...
+Note that the signer of an authorization tuple may be different than `tx.origin` of the transaction.
 
-1. Let `signer = ecrecover(keccak(MAGIC + contract_code), y_parity, r, s]`.
-2. Verify that the contract code of `signer` is empty.
-3. Set the contract code of `signer` to `contract_code`.
-4. Add the `signer` account to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
+#### Gas Costs
 
-At the end of the transaction, set the `contract_code` of each `signer` back to empty.
+The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `PER_CONTRACT_CODE_BASE_COST * authorization list length`.
```
Jrachman, line 71: suggesting per-contract code item calldata and the wrapper tx calldata is expected? where is the former supposed to be included and what is the point of the latter if we have the former?

lightclient:
@yperbasis you're correct, as specified it was not possible to represent null nonce value. I've updated it to wrap the nonce in a list object.

PR #8564, opened 2024-05-16
Update EIP-2935: Adjust gas-costs for 2935

MarekM25:
Adjusted gas costs for Blockhash opcode


yperbasis (changes requested):
Should update the Backwards Compatibility section

MarekM25:
@yperbasis thx, true, updated. But I think we should close this PR since we have pectra-devnet-1 version of this EIP

PR #8569, opened 2024-05-17
Website: add status anchors

hanniabu:
This adds status anchor links to the top of the page as can be seen here: https://i.imgur.com/vi2rL3I.png

Unlike in the image, this PR will only add anchor link if there are EIPs with that status (the image shows "Living" when it shouldn't)

### DIFF
--- a/_includes/eiptable.html
+++ b/_includes/eiptable.html
@@ -7,6 +7,16 @@
     width: 33%;
   }
 </style>
+
+{% assign statuses = include.eips | map: "status" %}
+{% for status in statuses %}
+  {% if forloop.last %}
+    <a href="#{{status|slugify}}">{{status}}</a><br><br>
+  {% else %}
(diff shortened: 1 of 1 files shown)


hanniabu:
bump

hanniabu:
This PR is ready for review

g11tech (approved):
makes sense to me

lightclient:
I think the site needs refresh in general. I would rather not update things piecemeal. This snippet also doesn't show the similar looking navbar for categories. Overall feels like it clutters the interface more than it adds value.

hanniabu:
Petition to reopen this since it's been ~2 years and the website still hasn't been refreshed

This snippet doesn't follow the same navbar b/c it's not a general category page, it's a "jump to section" link. That would be like saying you don't want to add a table of contents to an article b/c it doesn't match a website's menu.

PR #8573, opened 2024-05-17
Update EIP-2935: update gas cost & implementation, and specify verkle_fork requirements

gabrocheleau:
Adjusts the following points from the EIP:

- Splits the EIP into two forks happening at `FORK_TIMESTAMP` and `VERKLE_FORK_TIMESTAMP`
- From a user point-of-view, there is no change to  `BLOCKHASH` happening at `FORK_TIMESTAMP` (no change in behavior, no change in gas cost)
- Simplifies requirements at `FORK_TIMESTAMP`. The only expectation is that clients store the blockhashes to the contract storage from `FORK_TIMESTAMP`. There is no backfilling.
- At `FORK_TIMESTAMP`, only the blockhashes of the `BLOCKHASH_OLD_WINDOW` last blocks are served through the `BLOCKHASH` opcode. Blockhashes from the earlier blocks (up to `HISTORY_SERVE_WINDOW`) can be retrieve from the contract storage directly. Blockhashes that fall before the `BLOCKHASH_OLD_WINDOW` will however return `0` if obtained through the `BLOCKHASH` opcode.
- At `VERKLE_FORK_TIMESTAMP`, clients must serve the blockhash history from the contract storage at the updated gas cost. They must also add the storage accesses to the verkle execution witness. 
- At `VERKLE_FORK_TIMESTAMP`, the `BLOCKHASH` instruction will return valid blockhashes for `HISTORY_SERVE_WINDOW`.


EIPS/eip-2935.md
```diff
@@ -24,38 +24,31 @@ A side benefit of this approach could be that it allows building/validating proo
...
-        for i in range(HISTORY_SERVE_WINDOW - 1):
-            # stop at genesis block
-            if ancestor.number == 0:
-                break
+Note that from the `fork_block` (first block where `block.timestamp >= FORK_TIMESTAMP`), the implementation of `BLOCKHASH` does not change. The blockhashes of the last `BLOCKHASH_OLD_WINDOW` blocks can be retrieved as they previously were. `BLOCKHASH` will return `0` for blocks preceding the `BLOCKHASH_OLD_WINDOW`.
 
-            ancestor = ancestor.parent
-            state.insert_slot(HISTORY_STORAGE_ADDRESS, ancestor.number % HISTORY_SERVE_WINDOW, ancestor.hash)
-```
+Blockhashes from the `HISTORY_SERVE_WINDOW` can be retrieved directly from storage for blocks whose `block.number + 1 >= fork_block`.
```
gballet, line 47: at fork block, the parent is inserted, so it's actually `block.number >= fork_block -1`

gabrocheleau: `block.number + 1 >= fork_block` is equivalent to `block.number >= fork_block - 1`, no?

EIPS/eip-2935.md
```diff
@@ -66,9 +59,11 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
   return state.load_slot(HISTORY_STORAGE_ADDRESS, arg % HISTORY_SERVE_WINDOW)
 ```
 
+<!-- TODO: Do we need to update the implementation because there is not history backfilling? -->
```
gballet, line 62: no, because this is only really a problem one verkle activates (presumably people will keep using `BLOCKHASH` until the buffer is filled enough that it becomes interesting to use the contract. I would, however, add a note to explain just what i said.

EIPS/eip-2935.md
```diff
@@ -66,9 +59,11 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
...
+
 ### Contract Implementation
 
-Exact evm assembly that can be used for the contract to resolve `BLOCKHASH`
+Exact evm assembly that can be used for the contract to resolve `BLOCKHASH` from verkle_block_fork
```
gballet, line 66: that's not correct: the contract is available before the prague fork, whether it resolves blockhash or not.

gballet: so I would just remove the reference to `BLOCKHASH`

EIPS/eip-2935.md
```diff
@@ -66,9 +59,11 @@ def resolve_blockhash(block: Block, state: State, arg: uint64):
...
 
 ```
 // check if input > 8 byte value and revert if this isn't the case
```
gballet, line 69: ideally, the contract would be updated to add a `set` function

EIPS/eip-2935.md
```diff
@@ -24,38 +24,31 @@ A side benefit of this approach could be that it allows building/validating proo
...
-| `HISTORY_STORAGE_ADDRESS` | `0x25a219378dad9b3503c8268c9ca836a52427a4fb`|
-| `HISTORY_SERVE_WINDOW` | `8192` |
-| `BLOCKHASH_OLD_WINDOW` | `256` |
+| Parameter                 | Value                                        |
+| ------------------------- | -------------------------------------------- |
+| `FORK_TIMESTAMP`          | TBD                                          |
+| `VERKLE_FORK_TIMESTAMP`   | TBD                                          |
+| `HISTORY_STORAGE_ADDRESS` | `0x25a219378dad9b3503c8268c9ca836a52427a4fb` |
+| `HISTORY_SERVE_WINDOW`    | `8192`                                       |
+| `BLOCKHASH_OLD_WINDOW`    | `256`                                        |
```
g11tech, line 33:
```suggestion
| `BLOCKHASH_SERVE_WINDOW`    | `256`                                        |
```
I think this name would be better since we wouldn't modify blockhash window even when we serve it from state

EIPS/eip-2935.md
```diff
@@ -24,38 +24,31 @@ A side benefit of this approach could be that it allows building/validating proo
...
+```
 
-    # If this is the fork block, add the parent's direct `HISTORY_SERVE_WINDOW - 1` ancestors as well
-    if block.parent.timestamp < FORK_TIMESTAMP:
-        ancestor = block.parent
-        for i in range(HISTORY_SERVE_WINDOW - 1):
-            # stop at genesis block
-            if ancestor.number == 0:
-                break
+Note that from the `fork_block` (first block where `block.timestamp >= FORK_TIMESTAMP`), the implementation of `BLOCKHASH` does not change. The blockhashes of the last `BLOCKHASH_OLD_WINDOW` blocks can be retrieved as they previously were. `BLOCKHASH` will return `0` for blocks preceding the `BLOCKHASH_OLD_WINDOW`.
```
g11tech, line 45:
```suggestion
Note that from the `fork_block` (first block where `block.timestamp >= FORK_TIMESTAMP`), the implementation of `BLOCKHASH` does not change. The blockhashes of the last `BLOCKHASH_SERVE_WINDOW` blocks can be retrieved as they previously were. `BLOCKHASH` will continue to return `0` for blocks preceding the `BLOCKHASH_SERVE_WINDOW`.
```

EIPS/eip-2935.md
g11tech: also blockhash only serves its original 256 window right?

EIPS/eip-2935.md
```diff
@@ -158,37 +153,43 @@ A special synthetic address is generated by working backwards from the desired d
...
 
 Note, the input in the transaction has a simple constructor prefixing the desired runtime code.
 
 The sender of the transaction can be calculated as `0xa4690f0ed0d089faa1e0ad94c8f1b4a2fd4b0734`. The address of the first contract deployed from the account is `rlp([sender, 0])` which equals `0x25a219378dad9b3503c8268c9ca836a52427a4fb`. This is how `HISTORY_STORAGE_ADDRESS` is determined. Although this style of contract creation is not tied to any specific initcode like create2 is, the synthetic address is cryptographically bound to the input data of the transaction (e.g. the initcode).
 
+Some fork (blocks where `block.timestamp >= FORK_TIMESTAMP`) activation scenarios:
 
-Some activation scenarios:
+- For the fork to be activated at genesis, no history is written to the genesis state, and at the start of block `1`, genesis hash will be written as a normal operation to slot `0`.
+- for fork activation at block `1`, only genesis hash will be written at slot `0`.
```
jsign, line 167: Nit: maybe a good moment to uppercase this and the bullets below?

EIPS/eip-2935.md
```diff
@@ -158,37 +153,43 @@ A special synthetic address is generated by working backwards from the desired d
...
+- for fork activation at block `32`, block `31`'s hash will be written to slot `31`. Additional history for block `0..30` hashes will continue to be served, but will not be backfilled into storage.
```
jsign, line 168:
```suggestion
- for fork activation at block `32`, block `31`'s hash will be written to slot `31`. Additional history for block `0..30` hashes will continue to be served by `BLOCKHASH`, but will not be backfilled into the contract storage.
```

EIPS/eip-2935.md
```diff
@@ -158,37 +153,43 @@ A special synthetic address is generated by working backwards from the desired d
...
+- for fork activation at block `10000`, block `9999`'s hash will be written to slot `9999`. Additional history for the the `BLOCKHASH_OLD_WINDOW` (`9744..9998`) will continue to be served, but will not be backfilled into storage. History for the blocks preceding the `HISTORY_SERVE_WINDOW` (`1808..9998`) will not be served, and will not be backfilled into storage. At block `20000`, the BLOCKHASH opcode would continue serving history for the `BLOCKHASH_OLD_WINDOW` (`19744..19999`) and would return 0 for blocks before the `BLOCKHASH_OLD_WINDOW`. The blocks from `HISTORY_SERVE_WINDOW` would be served from state for blocks `11808..19999`. Blocks prior to the `HISTORY_SERVE_WINDOW` would return 0.
```
gballet, line 169: this should be broken into 2 examples and "will not be served" should be replaced by "will return 0" for clarity

EIPS/eip-2935.md
gballet: 🤦 yes

EIPS/eip-2935.md
```diff
@@ -158,37 +153,43 @@ A special synthetic address is generated by working backwards from the desired d
...
+- The storage access is provided in the verkle execution witness.
 
 ### [EIP-158](./eip-158.md) handling
 
 This address is currently exempt from [EIP-158](./eip-158.md) cleanup in Kaustinen Verkle Testnet but we plan to address this in the following way:
 
-* Deploy a contract à la [EIP-4788](./eip-4788.md) which just supports `get` method to resolve the BLOCKHASH as per the logic defined in `resolve_blockhash` (and use the generated address as the BLOCKHASH contract address).
-* While the clients are expected to directly read from state (or maintain and serve from memory) to resolve BLOCKHASH opcode, this contract's `get` could be invoked by transaction (via another contract or directly) leading to a normal contract execution (and gas consumption) as per the semantics of the contract call.
-
+- Deploy a contract à la [EIP-4788](./eip-4788.md) which just supports `get` method to resolve the BLOCKHASH as per the logic defined in `resolve_blockhash` (and use the generated address as the BLOCKHASH contract address).
```
gballet, line 181: should have a `set` as well

EIPS/eip-2935.md
jochem-brouwer, line 47: Can this be clarified? I think the idea is that if you want a block within `HISTORY_SERVE_WINDOW` but larger than `BLOCKHASH_OLD_WINDOW` you now call into the contract - right? Can this sentence then be updated such that it is clear that the contract should now be called and thus that `BLOCKHASH` will still put 0 on stack (even though the data is in the contract storage)

EIPS/eip-2935.md
jochem-brouwer, line 169: The slot is incorrect, it should be `9999 % HISTORY_SERVE_WINDOW`.

jochem-brouwer: Can it be made explicit also that the "served from state" is thus served from state via the contract call and thus NOT via `BLOCKHASH` opcode?

jochem-brouwer (commented):
Some comments

gabrocheleau:
Thanks for all the review comments. I will close this and extract the Verkle stuff into a separate EIP. 

This https://github.com/ethereum/EIPs/pull/8577 can be kept for the Pectra facet of the EIP .

jochem-brouwer:
Linking here for clarity, the "make the EIP less complex" (so basically the changes here but without the specific Verkle changes) are merged in this PR https://github.com/ethereum/EIPs/pull/8577

EIPS/eip-2935.md
```diff
@@ -198,28 +199,27 @@ However after weighing pros and cons, we decided to go with just a limited ring
...
-The cost of doing so is marginal considering the `HISTORY_SERVE_WINDOW` being relatively limited. Most clients write this into their flat db/memory caches and just requires reading last `HISTORY_SERVE_WINDOW` from the chain history.
+<!-- TODO: Explain why we are choosing the former -->
 
 ## Backwards Compatibility
 
-The behavior of `BLOCKHASH` opcode gets extended in backward compatible manner as the history it can serve will get extended upto `HISTORY_SERVE_WINDOW` on the fork block. However the gas charges will also get bumped as per the additional `SLOAD` costs.
+The behavior of `BLOCKHASH` opcode gets extended in a backward compatible manner as its functionality, and its gas cost, remains the same from `FORK_TIMESTAMP`. The blockhashes from `BLOCKHASH_OLD_WINDOW` can still be retrieve with `BLOCKHASH`. The additional history can be directly retrieve from the contract storage. From `VERKLE_FORK_TIMESTAMP`, `BLOCKHASH` will serve history for all blocks from `HISTORY_SERVE_WINDOW` from the contract storage address at the updated gas cost.
```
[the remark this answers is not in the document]
gabrocheleau: Ok, thanks for pointing that out, I had misunderstood and thought that the BLOCKHASH opcode was eventually updated to serve HISTORY_SERVE_WINDOW. Upcoming EIP will be adjusted.

EIPS/eip-2935.md
gabrocheleau: Is this for 2935, or for the verkle one?

PR #8577, opened 2024-05-17
Update EIP-2935: update for Prague devnet1

s1na:
This PR makes the following changes to the EIP:

- Reverts changes to the BLOCKHASH opcode so it retains its prior behavior. I.e. gas cost of 20 stays and it will not read from the contract state anymore. The idea is we should change the implementation of BLOCKHASH at the transition to Verkle tries.
- Backfilling logic is removed. It will take roughly 1 day for the contract storage to fill up with 8192 block hashes.
- EIP-2929 rules clarified w.r.t history contract

### DIFF
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -12,13 +12,13 @@ created: 2020-09-03
 
 ## Abstract
 
-Store last `HISTORY_SERVE_WINDOW` historical block hashes in a contract, and modify the `BLOCKHASH (0x40)` opcode to read and serve from this contract storage.
+Store last `HISTORY_SERVE_WINDOW` historical block hashes in the storage of a system contract as part of the block processing logic.
 
 ## Motivation
 
-Currently `BLOCKHASH` opcode accesses history to resolve hash of the block number in EVM. However a more stateless client friendly way is to maintain and serve these hashes from state.
+EVM implicitly assumes the client has the recent block (hashes) at hand. This assumption is not future-proof given the prospect of stateless clients. Including the block hashes in the state will allow bundling these hashes in the witness provided to a stateless client. This is already possible in the MPT and will become more efficient post-Verkle.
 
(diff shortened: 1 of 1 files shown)


tanishqjasoria (approved):

EIPS/eip-2935.md
```diff
@@ -170,21 +149,16 @@ The sender of the transaction can be calculated as `0xa4690f0ed0d089faa1e0ad94c8
...
-
+The bytecode above will be deployed à la [EIP-4788](./eip-4788.md). It just supports a `get` method to resolve block hashes. As such the account at `HISTORY_STORAGE_ADDRESS` will have code and a nonce of 1, and will be exempt from EIP-158 cleanup.
 
-### Gas costs and witnesses
+### Gas costs
 
-Since now `BLOCKHASH` is served from state, the clients now **additionally** charge the corresponding warm or cold `SLOAD` costs. For verkle based networks this would imply doing and bundling corresponding accesses (and gas charges) of `SLOAD`.
+The gas cost of the `BLOCKHASH` opcode is unchanged. Importantly the processing at the beginning of the block, i.e. `process_block_hash_history`, will not warm the `HISTORY_STORAGE_ADDRESS` account or its storage slots as per [EIP-2929](./eip-2929.md) rules. As such the first call to the contract will pay for warming up the account and storage slots it accesses.
```
jochem-brouwer, line 161: Nit: the first call to the contract is in tx context, not block context.

jochem-brouwer (commented):
One post-merge nit, LGTM otherwise

EIPS/eip-2935.md
s1na: Is your concern talking about the first call in this section at all? Or rather, how would you clarify this?

jochem-brouwer: "As such the first call to the contract will pay for warming up the account and storage slots it accesses."

It is not super clear that we are talking about the "first call in a tx to the contract" and not "first call in a block to the contract" (it could be misinterpreted). However, anyone with knowledge of EIP2929 will know that this is inside the tx context and not the block context.

So as nit, I would clarify it with "As such the first call within a transaction to the contract"...

PR #8578, opened 2024-05-18
Add EIP: Read BLOCKHASH from storage and update cost

gabrocheleau:
This EIP extracts the following verkle-specific behavior into a separate EIP to be included in a subsequent hard-fork (the verkle hard fork).

### Changes

- Updates the behavior of the BLOCKHASH opcode, which should now be served from the system contract storage
- Updates the gas cost of the BLOCKHASH opcode to reflect the cost of an SLOAD operation
- Specifies that BLOCKHASH storage accesses should be contained in the verkle execution witness 

### Links
- [Ethereum Magicians thread](https://ethereum-magicians.org/t/eip-7709-read-blockhash-opcode-from-storage-and-adjust-gas-cost/20052)

### DIFF
--- a/EIPS/eip-7704.md
+++ b/EIPS/eip-7704.md
@@ -1,5 +1,5 @@
 ---
-eip: 7703
+eip: 7704
 title: Adjust the BLOCKHASH opcode to read from storage and update its gas cost
 description: Serve the `BLOCKHASH (0x40)` opcode from the system contract storage from EIP-2935 to allow for stateless execution. Also adjust its gas cost to reflect storage access.
 author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria), Ignacio Hagopian (@jsign), Jochem Brouwer (@jochem-brouwer), Gabriel Rocheleau (@gabrocheleau)
--- a/EIPS/eip-7709.md
+++ b/EIPS/eip-7709.md
@@ -80,9 +80,9 @@ This EIP introduces a significant increase in the cost of `BLOCKHASH`, which cou
 
 ## Test Cases
 
-- If BLOCKHASH is called, there MUST be a storage access gas charge ( and witness if Verkle activated) for the corresponding storage slot ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
-- The gas cost for each `BLOCKHASH` operation should still be charged, in addition to the `SLOAD` cost of each lookup
-- If the storage contract is accessed directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent as per normal contract execution by the EVM as per the current fork/activated EIPs.
+* If BLOCKHASH is called, there MUST be a storage access gas charge ( and witness if Verkle activated) for the corresponding storage slot ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
+* The gas cost for each `BLOCKHASH` operation should still be charged, in addition to the `SLOAD` cost of each lookup
+* If the storage contract is accessed directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent as per normal contract execution by the EVM as per the current fork/activated EIPs.
 
 ## Security Considerations
 
--- a/EIPS/eip-7709.md
+++ b/EIPS/eip-7709.md
@@ -8,18 +8,16 @@ status: Draft
 type: Standards Track
 category: Core
 created: 2024-05-18
-requires: 2935, 4762, 6800
+requires: 2935
 ---
 
 ## Abstract
 
-Update the `BLOCKHASH (0x40)` opcode to read and serve from the system contract storage to allow for stateless execution. Update the `BLOCKHASH (0x40)` opcode gas cost to reflect storage access cost, and include storage accesses in the Verkle block witness.
+Update the `BLOCKHASH (0x40)` opcode to read and serve from the system contract storage to allow for stateless execution and charge the **additional** (cold or warm) gas on the lines of `SLOAD` for the corresponding slot.
 
 ## Motivation
 
 The `BLOCKHASH (0x40)` opcode currently assumes that the client has knowledge of the previous blocks, which in Verkle [EIP-6800](./eip-6800.md) would prevent stateless execution. However with [EIP-2935](./eip-2935.md) blockhashes can be retrieved and served from its system contract storage which allows Verkle blocks to include a storage access witness for stateless execution.
-
-This EIP specifies the transition to the new logic on the Verkle hardfork assuming that [EIP-2935](./eip-2935.md) has been activated sufficiently ahead of Verkle(>= `BLOCKHASH_SERVE_WINDOW` or at genesis for testnets/devnets)
 
 ## Specification
 
@@ -28,40 +26,67 @@ This EIP specifies the transition to the new logic on the Verkle hardfork assumi
 | `FORK_TIMESTAMP`          | TBD    |
 | `HISTORY_STORAGE_ADDRESS` | TBD    |
 | `BLOCKHASH_SERVE_WINDOW`  | `256`  |
-| `HISTORY_SERVE_WINDOW`    | `8192` |
 
-The `BLOCKHASH` opcode semantics remains the same as before. From the `fork_block` (first block where `block.timestamp >= FORK_TIMESTAMP`), the `BLOCKHASH` instruction should be updated to retrieve the requested blockhash from the contract storage.
+The `BLOCKHASH` opcode semantics remains the same as before. From the `fork_block` (first block where `block.timestamp >= FORK_TIMESTAMP`), the `BLOCKHASH` instruction should be updated to resolve block hash in the following manner:
 
-Additional gas costs need to be applied for the corresponding SLOAD operation and the corresponding access witness added to the execution witnesses
+```python
+def resolve_blockhash(block: Block, state: State, arg: uint64):
+  # note that outside the BLOCKHASH_SERVE_WINDOW we continue to return 0
+  # despite the 2935 history contract being able to serve more hashes
+  if arg >= block.number or (arg + BLOCKHASH_SERVE_WINDOW) < block.number
+    return 0
 
-### Deployment
+  # performs an sload on arg % HISTORY_SERVE_WINDOW including gas charges,
+  # warming effects as well as execution accesses
+  #
+  # note that the `BLOCKHASH_SERVE_WINDOW` and the 2935 ring buffer window
+  # `HISTORY_SERVE_WINDOW` for slot calculation are different
+  return state.load_slot(HISTORY_STORAGE_ADDRESS, arg % HISTORY_SERVE_WINDOW)
+```
 
-For the fork_block (blocks where `block.timestamp >= FORK_TIMESTAMP`), we consider that the fork (`FORK_TIMESTAMP`) is already in effect for at least `HISTORY_SERVE_WINDOW` blocks (or from genesis if `fork_block < HISTORY_SERVE_WINDOW`). The following changes take effect:
+ONLY if the `arg` is within the correct `BLOCKHASH` window, clients can choose to either
 
-- `BLOCKHASH` SHOULD retrieve blockhashes from the system contract storage and MUST provide the storage access in the Verkle execution witness
-- The gas cost of the `BLOCKHASH` opcode is updated as per specification
+* do a direct `SLOAD` from state, or
+* do a system call to [EIP-2935](./eip-2935.md) contract `get` or
+* serve from memory or as per current designs if maintaining requisite history (full clients for e.g.)
+
+However the entire semantics and after effects of the `SLOAD` operation needs to be applied as per the current fork if the `arg` is within the correct `BLOCKHASH` window:
+
+* `SLOAD` gas costs (cold or warm) for the `arg % HISTORY_SERVE_WINDOW` slot.
+* `SLOAD` after effects on the slot (warming the slot)
+* `SLOAD` accesses added to execution witnesses if Verkle ([EIP-6800](./eip-6800.md) and [EIP-4762](./eip-4762.md)) is activated
+
+### Activation
+
+This EIP specifies the transition to the new logic on the Verkle hardfork assuming that [EIP-2935](./eip-2935.md) has been activated sufficiently ahead of Verkle(>= `BLOCKHASH_SERVE_WINDOW` or at genesis for testnets/devnets).
 
 ### Gas costs
 
-At FORK_TIMESTAMP, the corresponding storage access witness for the blockhash `SLOAD` lookups must also included in the execution witness, along with their access gas charges. The blockhash that has been inserted at the beginning of the block processing (the parent block hash) is considered warm.
+As described above, if the `arg` to be resolved is within the correct window, the corresponding `SLOAD` charges and accesses are to be applied for the slot `arg % HISTORY_SERVE_WINDOW`. Note that the `HISTORY_SERVE_WINDOW` and `BLOCKHASH_SERVE_WINDOW` are different.
+
+### System contract
+
+Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-2935](./eip-2935.md) contract, the gas cost for the system code execution (and also the code witnesses if Verkle activated) is not applied. Only the effect of `SLOAD` is applied as described above.
 
 ## Rationale
 
-The reason behind the updated gas cost is to match the real cost of the operation, which is equivalent to an SLOAD.
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
+
+* The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which don't choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)
 
 ## Backwards Compatibility
 
-This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork. Otherwise, the blockhashes for the block preceding the `EIP-2935` fork block's parent would return `0`, as they will not have been saved to the system contract storage.
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) would not have saved the requisite history.
 
 ## Test Cases
 
-- If BLOCKHASH is not called within the block, there MUST be no witness related to the storage contract, except the storage slot witness for the parent blockhash just added to the storage contract.
-- If BLOCKHASH is called, there MUST be a storage access witness for each corresponding storage slot. There corresponding account access witnesses MUST NOT be included unless the contract has been accessed explicitly (i.e. not through `BLOCKHASH`). The gas cost for each BLOCKHASH operation should still be charged, in addition to the `SLOAD` cost of each lookup (initially cold, unless it is the parent blockhash, in which case it is initially warm).
-- If the storage contract is accessed directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent with other contract storage accesses, as per [EIP-4762](./eip-4762.md).
+- If BLOCKHASH is called, there MUST be a storage access gas charge ( and witness if Verkle activated) for the corresponding storage slot ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
+- The gas cost for each `BLOCKHASH` operation should still be charged, in addition to the `SLOAD` cost of each lookup
+- If the storage contract is accessed directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent as per normal contract execution by the EVM as per the current fork/activated EIPs.
 
 ## Security Considerations
 
-Having contracts (system or otherwise) with hot update paths (branches) poses a risk of "branch" poisoning attacks where attacker could sprinkle trivial amounts of eth around these hot paths (branches). But it has been deemed that cost of attack would escalate significantly to cause any meaningful slow down of state root updates.
+No Security considerations other than the [EIP-2935](./eip-2935.md) are determined as of now.
 
 ## Copyright
 
--- a/EIPS/eip-7709.md
+++ b/EIPS/eip-7709.md
@@ -83,7 +83,7 @@ This EIP introduces a significant increase in the cost of `BLOCKHASH`, which cou
 * If `BLOCKHASH` is called, there MUST be a storage access gas charge (and corresponding access witness if Verkle is activated) for the storage slot but ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors. This is irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
 * The gas cost for each `BLOCKHASH` operation should still be charged, in addition to the `SLOAD` cost of each lookup (if performed)
 * If the [EIP-2935](./eip-2935.md) contract is called directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent as per normal contract execution by the EVM as per the current fork/activated EIPs.
-* `BLOCKHASH` should be consistently resolved if this eip is activated correctly `>= BLOCKHASH_SERVE_WINDOW` after [EIP-2935](./eip-2935.md)
+* `BLOCKHASH` should be consistently resolved if this EIP is activated correctly `>= BLOCKHASH_SERVE_WINDOW` after [EIP-2935](./eip-2935.md)
 
 ## Security Considerations
 
--- a/EIPS/eip-7709.md
+++ b/EIPS/eip-7709.md
@@ -1,5 +1,5 @@
 ---
-eip: 7704
+eip: 7709
 title: Adjust the BLOCKHASH opcode to read from storage and update its gas cost
 description: Serve the `BLOCKHASH (0x40)` opcode from the system contract storage from EIP-2935 to allow for stateless execution. Also adjust its gas cost to reflect storage access.
 author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria), Ignacio Hagopian (@jsign), Jochem Brouwer (@jochem-brouwer), Gabriel Rocheleau (@gabrocheleau)
--- a/EIPS/eip-7709.md
+++ b/EIPS/eip-7709.md
@@ -1,14 +1,14 @@
 ---
 eip: 7709
-title: Read BLOCKHASH from storage and update gas cost
+title: Read BLOCKHASH from storage and update cost
 description: Read the `BLOCKHASH (0x40)` opcode from the EIP-2935 system contract storage and adjust its gas cost to reflect storage access.
 author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria), Ignacio Hagopian (@jsign), Jochem Brouwer (@jochem-brouwer), Gabriel Rocheleau (@gabrocheleau)
 discussions-to: https://ethereum-magicians.org/t/eip-7709-read-blockhash-opcode-from-storage-and-adjust-gas-cost/20052
 status: Draft
 type: Standards Track
-requires: 2935, 6800
 category: Core
 created: 2024-05-18
+requires: 2935, 6800
 ---
 
 ## Abstract
@@ -100,7 +100,7 @@ stop
 
 Note that the input contract read `32` bytes input as `calldataload`. Users and clients doing EVM call to this contract should left pad the `arg` correctly.
 
-<!-- TODO: bytecode is based off on first version and will be updated once assembley is locked down as it changes contract sender and address -->
+<!-- TODO: bytecode is based off on first version and will be updated once assembly is locked down as it changes contract sender and address -->
 
 Corresponding bytecode:
 `60203611603157600143035f35116029575f356120000143116029576120005f3506545f5260205ff35b5f5f5260205ff35b5f5ffd00`
--- a/EIPS/eip-7709.md
+++ b/EIPS/eip-7709.md
@@ -13,7 +13,7 @@ requires: 2935
 
 ## Abstract
 
-Update the `BLOCKHASH (0x40)` opcode to read and serve from the system contract storage and charge the **additional** (cold or warm) gas on the lines of `SLOAD` for the corresponding slot.
+Update the `BLOCKHASH (0x40)` opcode to read and serve from the system contract storage and charge the **additional** (cold or warm) storage costs.
 
 ## Motivation
 
@@ -69,7 +69,7 @@ The current proposal is to activate this EIP with Verkle to allow for stateless 
 
 As described above, if the `arg` to be resolved is within the correct window, the corresponding `SLOAD` charges and accesses are to be applied for the slot `arg % HISTORY_SERVE_WINDOW`. Note that the `HISTORY_SERVE_WINDOW` and `BLOCKHASH_SERVE_WINDOW` are different.
 
-### System contract
+### Reading from the System contract
 
 Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-2935](./eip-2935.md) contract, the gas cost for the system code execution (and also the code witnesses if Verkle activated) is not applied. Only the effect of `SLOAD` is applied as described above.
 
@@ -77,6 +77,8 @@ Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-29
 
 * The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
 * The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)
+
+Note that `BLOCKHASH` opcode only serves a limited `BLOCKHASH_SERVE_WINDOW` to be backward compatible (and to not extend the above exemptions). For deeper accesses one will need to directly call [EIP-2935](./eip-2935.md) system contract which will lead to a normal contract execution (as well as charges and accesses)
 
 ## Backwards Compatibility
 
@@ -87,7 +89,7 @@ This EIP introduces a significant increase in the cost of `BLOCKHASH`, which cou
 * If `BLOCKHASH` is not called or there is no [EIP-2935](./eip-2935.md) contract call by any transaction, only the [EIP-2935](./eip-2935.md) system update of the parent hash shows up in witnesses if Verkle is activated.
 * If `BLOCKHASH` is called, there MUST be a storage access gas charge (and corresponding access witness if Verkle is activated) for the storage slot but ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors. This is irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
 * The gas cost for each `BLOCKHASH` operation should still be charged, in addition to the `SLOAD` cost of each lookup (if performed)
-* If the [EIP-2935](./eip-2935.md) contract is called directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent as per normal contract execution by the EVM as per the current fork/activated EIPs.
+* If the [EIP-2935](./eip-2935.md) contract is called directly (i.e. not through `BLOCKHASH`), then the witness and gas costs (including those related to contract code) are applied as per normal contract execution of the current fork.
 * `BLOCKHASH` should be consistently resolved if this EIP is activated correctly `>= BLOCKHASH_SERVE_WINDOW` after [EIP-2935](./eip-2935.md)
 
 ## Security Considerations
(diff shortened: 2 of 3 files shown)


EIPS/eip-7704.md
```diff
@@ -0,0 +1,137 @@
...
+<!-- TODO: bytecode is based off on first version and will be updated once assembley is locked down as it changes contract sender and address -->
+
+Corresponding bytecode:
+`60203611603157600143035f35116029575f356120000143116029576120005f3506545f5260205ff35b5f5f5260205ff35b5f5ffd00`
+
+#### Deployment
+
+For the verkle_block (blocks where `block.timestamp >= FORK_TIMESTAMP`), we consider that the fork (`FORK_TIMESTAMP`) is already in effect for at least `HISTORY_SERVE_WINDOW` blocks (or from genesis if `fork_block < HISTORY_SERVE_WINDOW`). The following changes take effect:
+
+- `BLOCKHASH` MUST retrieve blockhashes from the system contract storage
```
shemnon, line 111: Is this really a MUST?  Where you get the data is irrelevant as long as the side effects are the same.  Clients may have a more efficient storage of recent block hashes that don't involve diving into storage, and then batch the trie paths into an accumulator to mark as read later.  

I think the spec would be improved by dropping this point, and moving an expanded point #3 to the fist point.

My main issue is with the word "retrieve" which I read as side-effect free where in this case the side effect of recording a read from storage is essential.  Perhaps making it a SHOULD and include language that if it is retrieved by other means that the side effects of a storage read should still be generated, but recording side effects seems to be the essence of the third point.

JakubPawlukiewicz (approved):

EIPS/eip-7704.md
```diff
@@ -0,0 +1,137 @@
+---
+eip: 7704
```
abcoathup, line 2:
```suggestion
eip: 7709
```

EIPS/eip-7704.md
```diff
@@ -0,0 +1,137 @@
...
+title: Adjust the BLOCKHASH opcode to read from storage and update its gas cost
+description: Serve the `BLOCKHASH (0x40)` opcode from the system contract storage from EIP-2935 to allow for stateless execution. Also adjust its gas cost to reflect storage access.
+author: Vitalik Buterin (@vbuterin), Tomasz Stanczak (@tkstanczak), Guillaume Ballet (@gballet), Gajinder Singh (@g11tech), Tanishq Jasoria (@tanishqjasoria), Ignacio Hagopian (@jsign), Jochem Brouwer (@jochem-brouwer), Gabriel Rocheleau (@gabrocheleau)
+discussions-to: https://ethereum-magicians.org/t/eip-2935-save-historical-block-hashes-in-state/4565
```
abcoathup, line 6: Please create a discussions topic on Eth Magicians for this EIP

EIPS/eip-7704.md
gabrocheleau: Thank you, updated.

EIPS/eip-7704.md
gabrocheleau: Created the discussion here: https://ethereum-magicians.org/t/eip-7709-read-blockhash-opcode-from-storage-and-adjust-gas-cost/20052

Have also updated the link in the EIP.

EIPS/eip-7704.md
gabrocheleau: Good Q. I agree that it does not make sense to require actual reads, we just need to data to be correct + the side effects (gas cost + witness) to be consistent. Will update

gabrocheleau: adjusted

tanishqjasoria (changes requested):

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+
+The motivation behind the updated gas cost is to match the real cost of the operation, which is equivalent to an `SLOAD`.
+
+## Specification
+
+| Parameter                 | Value                                        |
+| ------------------------- | -------------------------------------------- |
+| `FORK_TIMESTAMP`          | TBD                                          |
+| `HISTORY_STORAGE_ADDRESS` | `0x25a219378dad9b3503c8268c9ca836a52427a4fb` |
+| `BLOCKHASH_SERVE_WINDOW`  | `256`                                        |
```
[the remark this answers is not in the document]
shemnon: At nyota it was discussed that it should always have the old range for maximum compataibility.  Contracts that want an extended range cal call the smart contract directly, like they do for beacon root hash.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+#### Deployment
+
+For the fork_block (blocks where `block.timestamp >= FORK_TIMESTAMP`), we consider that the fork (`FORK_TIMESTAMP`) is already in effect for at least `HISTORY_SERVE_WINDOW` blocks (or from genesis if `fork_block < HISTORY_SERVE_WINDOW`). The following changes take effect:
+
+- `BLOCKHASH` SHOULD retrieve blockhashes from the system contract storage and MUST provide the storage access in the Verkle execution witness
+- The gas cost of the `BLOCKHASH` opcode is updated as per specification
+
+### [EIP-158](./eip-158.md) handling
+
+### Gas costs
```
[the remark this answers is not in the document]
shemnon: For prague we are sticking with 20 gas, so we don't have a one-fork orphan schedule.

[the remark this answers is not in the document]
shemnon: For prague we are sticking with 20 gas, so we don't have a one-fork orphan schedule.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+| `BLOCKHASH_SERVE_WINDOW`  | `256`                                        |
+| `HISTORY_SERVE_WINDOW`    | `8192`                                       |
+
+The `BLOCKHASH` opcode semantics remains the same as before. From the `fork_block` (first block where `block.timestamp >= FORK_TIMESTAMP`), the `BLOCKHASH` instruction should be updated to retrieve the requested blockhash from the contract storage.
+
+### Contract Implementation
+
+Exact evm assembly that can be used for the contract to resolve `BLOCKHASH`
+
+```
```
gballet, line 39: this is still missing the `set` function (and the associated method dispatch table), which needs to:

1. check the sender is the system contract, revert otherwise
2. geth the block number + block hash (as a parameter)
3. store it into the right slot

Maybe @g11tech wants to take another shot at it? I can also do it otherwise.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+`60203611603157600143035f35116029575f356120000143116029576120005f3506545f5260205ff35b5f5f5260205ff35b5f5ffd00`
+
+#### Deployment
+
+For the fork_block (blocks where `block.timestamp >= FORK_TIMESTAMP`), we consider that the fork (`FORK_TIMESTAMP`) is already in effect for at least `HISTORY_SERVE_WINDOW` blocks (or from genesis if `fork_block < HISTORY_SERVE_WINDOW`). The following changes take effect:
+
+- `BLOCKHASH` SHOULD retrieve blockhashes from the system contract storage and MUST provide the storage access in the Verkle execution witness
+- The gas cost of the `BLOCKHASH` opcode is updated as per specification
+
+### [EIP-158](./eip-158.md) handling
```
gballet, line 116: is this section needed? we are relseasing a regular contract, so eip158 should not be a problem.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+The `BLOCKHASH (0x40)` opcode currently assumes that the client has knowledge of the previous blocks, which in Verkle, would prevent stateless execution. This EIP assumes that [EIP-2935](./eip-2935.md) has been implemented, and that blockhashes can be retrieved from the systems contract storage. By updating the behavior of `BLOCKHASH (0x40)` to directly read and serve from state through the system contract storage, we allow Verkle blocks to include a storage access witness, which will allow stateless execution.
+
+The motivation behind the updated gas cost is to match the real cost of the operation, which is equivalent to an `SLOAD`.
+
+## Specification
+
+| Parameter                 | Value                                        |
+| ------------------------- | -------------------------------------------- |
+| `FORK_TIMESTAMP`          | TBD                                          |
+| `HISTORY_STORAGE_ADDRESS` | `0x25a219378dad9b3503c8268c9ca836a52427a4fb` |
```
gballet, line 29: I would set this to `TBD` for now, as the code for the contract needs to be finalized and deployed with eip 2935. This part will not change with 7709, as we need to use the same contract for continuity.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Rationale
+
+## Backwards Compatibility
+
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost.
+
+## Test Cases
+
+TBD
```
gballet, line 130: There should be at least the following tests cases:

* if `BLOCKHASH` is not called in the block, and neither is the contract what does the witness look like?
* if `BLOCKHASH` is called but not the contract, what does the witness look like?
* if the contract is called but not `BLOCKHASH`, what does the witness look like?

witness and gas costs, that is.

gballet: also, if for whatever reason a testnet has prague and osaka within 8192 blocks, what are the returned values.

EIPS/eip-7709.md
gabrocheleau: removed

EIPS/eip-7709.md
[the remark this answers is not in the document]
gabrocheleau: So to be clear:
- The storage slot that was insert at the beginning of the block processing (i.e. the parent block hash) is warm. I'll adjust that.
- My understanding was that there was no "account access" associated with these operations (only storage accesses), I therefore would only tend to consider the storage slot "warm" and not the account. Thoughts?

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+For the fork_block (blocks where `block.timestamp >= FORK_TIMESTAMP`), we consider that the fork (`FORK_TIMESTAMP`) is already in effect for at least `HISTORY_SERVE_WINDOW` blocks (or from genesis if `fork_block < HISTORY_SERVE_WINDOW`). The following changes take effect:
+
+- `BLOCKHASH` SHOULD retrieve blockhashes from the system contract storage and MUST provide the storage access in the Verkle execution witness
+- The gas cost of the `BLOCKHASH` opcode is updated as per specification
+
+### [EIP-158](./eip-158.md) handling
+
+### Gas costs
+
+At FORK_TIMESTAMP, the corresponding storage access witness for the blockhash `SLOAD` lookups must also included in the execution witness, along with their access gas charges.
```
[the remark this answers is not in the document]
gabrocheleau: I have considered that we do not add the account witness, but only the storage witness so far. I have heard disagreements about this. Let's perhaps move over this conversation over to ethereum magicians? I've added these open questions here: https://ethereum-magicians.org/t/eip-7709-read-blockhash-opcode-from-storage-and-adjust-gas-cost/20052

EIPS/eip-7709.md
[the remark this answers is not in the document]
shemnon: Is this for when verkle was implemented? If so make that clear.

IIR the discussion on the lawn had all slots and accounts cold after the block hash processing while we still have the MPT.  Warming specific fields globally is a fairly invasive change and shouldn't be done for just.a fork or two.

EIPS/eip-7709.md
shemnon: The old spec was 8192, with backfill, until we changed it after seeing how invasive the changes would have been.

EIPS/eip-7709.md
gabrocheleau: updated

EIPS/eip-7709.md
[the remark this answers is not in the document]
gabrocheleau: I have added a mention that this EIP is for the same hard fork when verkle will be implemented in the Motivation section. 

There does not seem to be consensus as to what (if any) accounts/storage slots should be considered warm. I've adjust the open questions of the eth-magicians thread to reflect this: https://ethereum-magicians.org/t/eip-7709-read-blockhash-opcode-from-storage-and-adjust-gas-cost/20052

EIPS/eip-7709.md
gabrocheleau: Shouldn't that  `set` be part of EIP-2935? This EIP only updates the BLOCKHASH implementation so that it is resolved through the contract storage (so `get`), right?

EIPS/eip-7709.md
gabrocheleau: I don't think the case where prague and osaka happen within 8192 blocks is as concerning as the case where prague and osaka happen within 256 blocks. With the former, the storage access just returns 0, with a proof of absence. This is not breaking because this is consistent with what happens at prague time (only the block hashes from `prague_fork_block - 1` are stored, all previous ones would return 0). However, if the osaka happens within 256 blocks, then we have break in functionality because:
- With EIP-2935 (at prague), we still serve the last 256 blocks from client state as we did in the past (not breaking)
- With EIP-7709 (at osaka), we now expect the block hashes to be retrieved through storage, meaning that blocks before `prague_fork_block - 1` would return `0`. This is breaking unless the Osaka fork happens at least 256 blocks before. Obviously won't be the case on Mainnet, but could certainly cause issues for testnets.

gabrocheleau: Update with the suggested examples from @gballet , and added a mention of the breaking change if there is not enough blocks between prague and osaka.

EIPS/eip-7709.md
gballet: well the bytecode won't be changed, so the bytecode needs to be updated in 2935 and need not be in this eip then.

EIPS/eip-7709.md
shemnon: I only expect this in the case where it is prague at genesis.  We will not be testing transitions from devnets starting at Cancun.  In the case of prague at genesis all available slots for BLOCKHASH will be loaded from genesis.,even if it's less than 256.  But it will be all blocks on the chain.

Mortiemi: > There should be at least the following tests cases:
> 
> 
> 
> * if `BLOCKHASH` is not called in the block, and neither is the contract what does the witness look like?
> 
> * if `BLOCKHASH` is called but not the contract, what does the witness look like?
> 
> * if the contract is called but not `BLOCKHASH`, what does the witness look like?
> 
> 
> 
> witness and gas costs, that is.

EIPS/eip-7709.md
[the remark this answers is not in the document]
tanishqjasoria: okay, we should mention this in the eip, from eip its not clear right now.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+requires: 2935, 4762, 6800
+---
+
+## Abstract
+
+Update the `BLOCKHASH (0x40)` opcode to read and serve from the system contract storage to allow for stateless execution. Update the `BLOCKHASH (0x40)` opcode gas cost to reflect storage access cost, and include storage accesses in the Verkle block witness.
+
+## Motivation
+
+The `BLOCKHASH (0x40)` opcode currently assumes that the client has knowledge of the previous blocks, which in Verkle, would prevent stateless execution. This EIP assumes that [EIP-2935](./eip-2935.md) has been implemented, and that blockhashes can be retrieved from the systems contract storage. This EIP also assumes that the Verkle [EIP-6800](./eip-6800.md) is being implemented in the same hard fork. By updating the behavior of `BLOCKHASH (0x40)` to directly read and serve from state through the system contract storage, we allow Verkle blocks to include a storage access witness, which will allow stateless execution.
```
g11tech, line 20:
```suggestion
The `BLOCKHASH (0x40)` opcode currently assumes that the client has knowledge of the previous blocks, which in Verkle, would prevent stateless execution. This EIP assumes that [EIP-2935](./eip-2935.md) has been activated, and that blockhashes can be retrieved from the systems contract storage. This EIP also assumes that the Verkle [EIP-6800](./eip-6800.md) is being implemented in the same hard fork. By updating the behavior of `BLOCKHASH (0x40)` to directly read and serve from state through the system contract storage, we allow Verkle blocks to include a storage access witness, which will allow stateless execution.
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+
+The motivation behind the updated gas cost is to match the real cost of the operation, which is equivalent to an `SLOAD`.
```
g11tech, line 22: you should remove this from abstract as abstract just specifies that is the EIP idea

this should go in specification:

`Additional gas costs need to be applied for the corresponding SLOAD operation and the corresponding access witness added to the execution witnesses`

In Rational section you should add:

 `The reason behind the updated gas cost is to match the real cost of the operation, which is equivalent to an `SLOAD`.`

EIPS/eip-7709.md
g11tech: this section doesn't belong in this EIP and can be removed, I will update the blockhash bytecode in the blockhash eip

EIPS/eip-7709.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+Note that the input contract read `32` bytes input as `calldataload`. Users and clients doing EVM call to this contract should left pad the `arg` correctly.
+
+<!-- TODO: bytecode is based off on first version and will be updated once assembly is locked down as it changes contract sender and address -->
+
+Corresponding bytecode:
+`60203611603157600143035f35116029575f356120000143116029576120005f3506545f5260205ff35b5f5f5260205ff35b5f5ffd00`
+
+#### Deployment
```
g11tech, line 109: deployment also needs to go

EIPS/eip-7709.md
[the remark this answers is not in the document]
gabrocheleau: My understanding is that BLOCKHASH will only serve `BLOCKHASH_SERVE_WINDOW`. If users want to access the history, they should retrieve it from storage directly.

EIPS/eip-7709.md
gabrocheleau: removed

EIPS/eip-7709.md
gabrocheleau: Are you sure? It seems like it contains relevant information for implementation?

EIPS/eip-7709.md
gabrocheleau: Done

EIPS/eip-7709.md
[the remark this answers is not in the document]
PeakyBlinders52: Doesn't work if their laptop is completely broken and they have no money for a new one

EIPS/eip-7704.md
g11tech: have made things more clear now

EIPS/eip-7709.md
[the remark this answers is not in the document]
g11tech: have clarified this more crisply @tanishqjasoria

EIPS/eip-7709.md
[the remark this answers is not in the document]
g11tech: I have removed this from here as this should be part of 4762 EIP (and has also been proposed there)

whatever slot is warmed or not as per the effect of other EIPs /accesses the charges will follow

EIPS/eip-7709.md
g11tech: > for whatever reason a testnet has prague and osaka within 8192 blocks,


this EIP assumes that 2935 has been sufficiently activated ahead > 256 (so not even 8192), so adding a test case if that happens

EIPS/eip-7709.md
g11tech: yes its part of 2935

EIPS/eip-7709.md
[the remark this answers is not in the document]
g11tech: added the clarification in the main EIP

EIPS/eip-7709.md
```diff
@@ -0,0 +1,95 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-05-18
+requires: 2935
+---
+
+## Abstract
+
+Update the `BLOCKHASH (0x40)` opcode to read and serve from the system contract storage to allow for stateless execution and charge the **additional** (cold or warm) gas on the lines of `SLOAD` for the corresponding slot.
```
gabrocheleau, line 16: Concerning the `to allow for stateless execution` part, while this is a clear benefit, I'm not sure this deserves to be in the abstract. 
```suggestion
Update the `BLOCKHASH (0x40)` opcode to read and serve from the system contract storage and charge the **additional** (cold or warm) gas on the lines of `SLOAD` for the corresponding slot.
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,95 @@
...
+
+However the entire semantics and after effects of the `SLOAD` operation needs to be applied as per the current fork if the `arg` is within the correct `BLOCKHASH` window:
+
+* `SLOAD` gas costs (cold or warm) for the `arg % HISTORY_SERVE_WINDOW` slot.
+* `SLOAD` after effects on the slot (warming the slot)
+* `SLOAD` accesses added to execution witnesses if Verkle ([EIP-6800](./eip-6800.md) and [EIP-4762](./eip-4762.md)) is activated
+
+### Activation
+
+This EIP specifies the transition to the new logic on the Verkle hardfork assuming that [EIP-2935](./eip-2935.md) has been activated sufficiently ahead of Verkle(>= `BLOCKHASH_SERVE_WINDOW` or at genesis for testnets/devnets).
```
gabrocheleau, line 61:
```suggestion
This EIP specifies the transition to the new logic on the Verkle hardfork assuming that [EIP-2935](./eip-2935.md) has been activated sufficiently ahead of Verkle (>= `BLOCKHASH_SERVE_WINDOW` or at genesis for testnets/devnets).
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,95 @@
...
+As described above, if the `arg` to be resolved is within the correct window, the corresponding `SLOAD` charges and accesses are to be applied for the slot `arg % HISTORY_SERVE_WINDOW`. Note that the `HISTORY_SERVE_WINDOW` and `BLOCKHASH_SERVE_WINDOW` are different.
+
+### System contract
+
+Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-2935](./eip-2935.md) contract, the gas cost for the system code execution (and also the code witnesses if Verkle activated) is not applied. Only the effect of `SLOAD` is applied as described above.
+
+## Rationale
+
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
+
```
gabrocheleau, line 74:
```suggestion
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,94 @@
...
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
+* The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)
+
+## Backwards Compatibility
+
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) would not have saved the requisite history.
+
+## Test Cases
+
+* If `BLOCKHASH` is not called or there is no [EIP-2935](./eip-2935.md) contract call by an transaction, only the [EIP-2935](./eip-2935.md) system update of the parent hash show up in witnesses if Verkle is activated.
```
gabrocheleau, line 82:
```suggestion
* If `BLOCKHASH` is not called or there is no [EIP-2935](./eip-2935.md) contract call by any transaction, only the [EIP-2935](./eip-2935.md) system update of the parent hash show up in witnesses if Verkle is activated.
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,94 @@
...
+Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-2935](./eip-2935.md) contract, the gas cost for the system code execution (and also the code witnesses if Verkle activated) is not applied. Only the effect of `SLOAD` is applied as described above.
+
+## Rationale
+
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
+* The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)
+
+## Backwards Compatibility
+
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) would not have saved the requisite history.
```
gabrocheleau, line 78:
```suggestion
This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) system contract would not have saved the requisite history.
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,94 @@
...
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
+* The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)
+
+## Backwards Compatibility
+
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) system contract would not have saved the required history.
+
+## Test Cases
+
+* If `BLOCKHASH` is not called or there is no [EIP-2935](./eip-2935.md) contract call by any transaction, only the [EIP-2935](./eip-2935.md) system update of the parent hash show up in witnesses if Verkle is activated.
```
gabrocheleau, line 82:
```suggestion
* If `BLOCKHASH` is not called or there is no [EIP-2935](./eip-2935.md) contract call by any transaction, only the [EIP-2935](./eip-2935.md) system update of the parent hash shows up in witnesses if Verkle is activated.
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Backwards Compatibility
+
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) system contract would not have saved the required history.
+
+## Test Cases
+
+* If `BLOCKHASH` is not called or there is no [EIP-2935](./eip-2935.md) contract call by any transaction, only the [EIP-2935](./eip-2935.md) system update of the parent hash shows up in witnesses if Verkle is activated.
+* If `BLOCKHASH` is called, there MUST be a storage access gas charge ( and correspending access witness if Verkle activated) for the storage slot but ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors. This is irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
```
gabrocheleau, line 83:
```suggestion
* If `BLOCKHASH` is called, there MUST be a storage access gas charge (and corresponding access witness if Verkle is activated) for the storage slot but ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors. This is irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Test Cases
+
+* If `BLOCKHASH` is not called or there is no [EIP-2935](./eip-2935.md) contract call by any transaction, only the [EIP-2935](./eip-2935.md) system update of the parent hash shows up in witnesses if Verkle is activated.
+* If `BLOCKHASH` is called, there MUST be a storage access gas charge (and corresponding access witness if Verkle is activated) for the storage slot but ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors. This is irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
+* The gas cost for each `BLOCKHASH` operation should still be charged, in addition to the `SLOAD` cost of each lookup (if performed)
+* If the [EIP-2935](./eip-2935.md) contract is called directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent as per normal contract execution by the EVM as per the current fork/activated EIPs.
+* `BLOCKHASH` should be consistently resolved if this eip is activated correctly `>= BLOCKHASH_SERVE_WINDOW` after [EIP-2935](./eip-2935.md)
```
gabrocheleau, line 86:
```suggestion
* `BLOCKHASH` should be consistently resolved if this EIP is activated correctly `>= BLOCKHASH_SERVE_WINDOW` after [EIP-2935](./eip-2935.md)
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,94 @@
...
+* If `BLOCKHASH` is called, there MUST be a storage access gas charge (and corresponding access witness if Verkle is activated) for the storage slot but ONLY if the `BLOCKHASH` query is for the last `BLOCKHASH_SERVE_WINDOW` ancestors. This is irrespective of how the client chooses to resolve the `BLOCKHASH` (directly, via system contract or via memory)
+* The gas cost for each `BLOCKHASH` operation should still be charged, in addition to the `SLOAD` cost of each lookup (if performed)
+* If the [EIP-2935](./eip-2935.md) contract is called directly (i.e. not through `BLOCKHASH`), then the witness and gas costs remain consistent as per normal contract execution by the EVM as per the current fork/activated EIPs.
+* `BLOCKHASH` should be consistently resolved if this EIP is activated correctly `>= BLOCKHASH_SERVE_WINDOW` after [EIP-2935](./eip-2935.md)
+
+## Security Considerations
+
+No Security considerations other than the [EIP-2935](./eip-2935.md) are determined as of now.
```
gabrocheleau, line 90:
```suggestion
No security considerations other than the ones contained in [EIP-2935](./eip-2935.md) are determined as of now.
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
+
+* `SLOAD` gas costs (cold or warm) for the `arg % HISTORY_SERVE_WINDOW` slot.
+* `SLOAD` after effects on the slot (warming the slot)
+* `SLOAD` accesses added to execution witnesses if Verkle ([EIP-6800](./eip-6800.md) and [EIP-4762](./eip-4762.md)) is activated
+
+### Activation
+
+This EIP specifies the transition to the new logic assuming that [EIP-2935](./eip-2935.md) has been activated:
+
+* sufficiently ahead of thie EIP's activation (>= `BLOCKHASH_SERVE_WINDOW`) or
```
gabrocheleau, line 63:
```suggestion
* sufficiently ahead of this EIP's activation (>= `BLOCKHASH_SERVE_WINDOW`) or
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
+
+The current proposal is to activate this EIP with Verkle to allow for stateless execution of the block.
+
+### Gas costs
+
+As described above, if the `arg` to be resolved is within the correct window, the corresponding `SLOAD` charges and accesses are to be applied for the slot `arg % HISTORY_SERVE_WINDOW`. Note that the `HISTORY_SERVE_WINDOW` and `BLOCKHASH_SERVE_WINDOW` are different.
+
+### System contract
+
+Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-2935](./eip-2935.md) contract, the gas cost for the system code execution (and also the code witnesses if Verkle activated) is not applied. Only the effect of `SLOAD` is applied as described above.
```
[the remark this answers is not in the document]
gballet: the title is confusing: it says that this is the system contract, but that only concerns calling the system contract via `BLOCKHASH`. But there is another case for which the system contract code (if used) doesn't end up in the witness: calling the system contract from the system address (formerly known as "`set`"). This should also be mentioned, as well as the idea that if you call the contract from another contract, the code _will_ end up in the witness.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Rationale
+
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
```
gballet, line 78: that's true of calling `BLOCKHASH` but if you want to get more than 256 values, you also pay for a `CALL`.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
+* The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)
```
gballet, line 79: that's also when doing `BLOCKHASH`, not when calling the contract. This should be clarified.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Backwards Compatibility
+
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) system contract would not have saved the required history.
```
gballet, line 83: but the latter sentence is only relevant in the case that 2935 isn't activated at genesis.

You should also specify how it breaks, then: calling the contract would return 0, for any block prior to the 2935 block.

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
+  #
+  # note that the `BLOCKHASH_SERVE_WINDOW` and the 2935 ring buffer window
+  # `HISTORY_SERVE_WINDOW` for slot calculation are different
+  return state.load_slot(HISTORY_STORAGE_ADDRESS, arg % HISTORY_SERVE_WINDOW)
+```
+
+ONLY if the `arg` is within the correct `BLOCKHASH` window, clients can choose to either
+
+* do a direct `SLOAD` from state, or
+* do a system call to [EIP-2935](./eip-2935.md) contract `get` or
```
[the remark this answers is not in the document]
g11tech: we should say `2935 get mechanism` no need to explain mechanism here as the reader will read eip 2935

EIPS/eip-7709.md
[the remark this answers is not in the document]
g11tech: `set` mechanism is not the concern here, but yea we can rename the section to : `reading from system contract`

EIPS/eip-7709.md
g11tech: calling evm contract is not a concern here for > 256, for that user should go to eip 2935, may be we can mention something on the lines of this is Rationle

EIPS/eip-7709.md
g11tech: i think this addresses it:
`..re not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)`

EIPS/eip-7709.md
g11tech: we have already mentioned the activation dependency of 2935

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
```
g11tech, line 50:
```suggestion
* do a system call to [EIP-2935](./eip-2935.md) contract via its `get` mechanism (caller other than `SYSTEM_ADDRESS`) or
```

EIPS/eip-7709.md
```diff
@@ -0,0 +1,99 @@
...
+Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-2935](./eip-2935.md) contract, the gas cost for the system code execution (and also the code witnesses if Verkle activated) is not applied. Only the effect of `SLOAD` is applied as described above.
+
+## Rationale
+
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
+* The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)
+
+## Backwards Compatibility
+
+This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork as the [EIP-2935](./eip-2935.md) system contract would not have saved the required history.
```
g11tech, line 83:
```suggestion
This EIP introduces a significant increase in the cost of `BLOCKHASH`, which could break use-cases that rely on the previous gas cost. Also, this EIP introduces a breaking change in the case where less than `BLOCKHASH_SERVE_WINDOW` elapse between the [EIP-2935](./eip-2935.md) fork and this EIP's fork (unless [EIP-2935](./eip-2935.md)  is activated in genesis for e.g. in testnets/devnets) as the [EIP-2935](./eip-2935.md) system contract would not have saved the required history.
```

HoangTranMap:
hi okayĐược gửi từ iPhone của tôiVào ngày 29 thg 5, 2024, lúc 03:20, g11tech ***@***.***> đã viết:﻿
@g11tech commented on this pull request.



In EIPS/eip-7709.md:
> +* at genesis for testnets/devnets where this EIP could also be activated at genesis
+
+The current proposal is to activate this EIP with Verkle to allow for stateless execution of the block.
+
+### Gas costs
+
+As described above, if the `arg` to be resolved is within the correct window, the corresponding `SLOAD` charges and accesses are to be applied for the slot `arg % HISTORY_SERVE_WINDOW`. Note that the `HISTORY_SERVE_WINDOW` and `BLOCKHASH_SERVE_WINDOW` are different.
+
+### System contract
+
+Even if the clients choose to resolve `BLOCKHASH` through system call to [EIP-2935](./eip-2935.md) contract, the gas cost for the system code execution (and also the code witnesses if Verkle activated) is not applied. Only the effect of `SLOAD` is applied as described above.
+
+## Rationale
+
+* The reason behind the updated gas cost is to match the real operation, which is equivalent to an `SLOAD`.
+* The [EIP-2935](./eip-2935.md) system contract execution charges (and accesses) are not applied to keep the gas low and to keep things simple for clients which choose to resolve `BLOCKHASH` in other ways (directly or though memory/maintained history)

i think this addresses it:
..re not applied to keep the gas low and to keep things simple for clients which choose to resolve BLOCKHASH in other ways (directly or though memory/maintained history)

EIPS/eip-7709.md
g11tech: added the clarity on this

g11tech (approved):

EIPS/eip-7709.md
Orhanoguz3567: 0e03197e2f797a9fa54eb59f9f2796a577fdaa34

PR #8590, opened 2024-05-23
Update EIP-4788: specify the timestamp for 4788

smartprogrammer93:


lightclient:
Unfortunately this EIP is final, so I'm not sure if we can update this.

g11tech:
> Unfortunately this EIP is final, so I'm not sure if we can update this.

we should as this is clearly an editing oversight, on the side note we need to check for TBDs while moving to review

SamWilsn (approved):

g11tech (approved):

lightclient (approved):

PR #8596, opened 2024-05-25
Update EIP-2935: Update EVM bytecode to add the system set update op

g11tech:
TODO
 - [x] update deployment tx
 - [x] mine and update new address
   - deployment details are as calculated and tested in ethereumjs test code
     - https://github.com/ethereumjs/ethereumjs-monorepo/pull/3447


gballet (changes requested):
there is no function dispatcher, which I understand is a solidity-level concept, but that means that a code written in solidity will either have to a) use a low-level call function written in yul, or b) this contract will have to ignore the first 4 bytes of the calldata, just in case they are added by a solidity call.

Also, this effectively says that unless you're the system contract, you're going a `get`. I would rather do it the right way, but I would check with matt how the 4788 contract does it.

g11tech:
> Also, this effectively says that unless you're the system contract, you're going a `get`. I would rather do it the right way, but I would check with matt how the 4788 contract does it.

this is exactly how 4788 does it
image

EIPS/eip-2935.md
```diff
@@ -103,21 +111,32 @@ push1 0x20
...
 push0
 revert
 
+// 0x57: set op - sstore the input to number-1
+jumpdest
+push0
+calldataload
+push1 0x1
+number
+sub
```
gballet, line 126: you need to mod it by 8192 here, in order to sanitize the input. This has a major drawback though: it means that we can not change the value of the buffer at a later time.

gballet: I guess it's better to leave it as is then, but it would be a good idea to add a note about it in the next section.

EIPS/eip-2935.md
```diff
@@ -103,21 +111,32 @@ push1 0x20
...
-Note that the input contract read `32` bytes input as `calldataload`. Users and clients doing EVM call to this contract should left pad the `arg` correctly.
+Corresponding bytecode:
+`3373fffffffffffffffffffffffffffffffffffffffe1460575767ffffffffffffffff5f3511605357600143035f3511604b575f35612000014311604b57611fff5f3516545f5260205ff35b5f5f5260205ff35b5f5ffd5b5f35600143035500`
 
-<!-- TODO: bytecode is based off on first version and will be updated once assembly is locked down as it changes contract sender and address -->
+#### Contract `get` and `set` mechanism
 
-Corresponding bytecode:
-`60203611603157600143035f35116029575f356120000143116029576120005f3506545f5260205ff35b5f5f5260205ff35b5f5ffd00`
+Similar to [EIP-4788](./eip-4788.md) above contract provides a `set` mechanism which clients can choose to update the block's parent hash in the ring buffer instead of direct state update. For this clients should call this contract with `SYSTEM_ADDRESS` as caller and provide `32` bytes parent block hash as input.
```
gballet, line 137:
```suggestion
Similar to [EIP-4788](./eip-4788.md) above contract provides a `set` mechanism which clients can choose to update the block's parent hash in the ring buffer instead of direct state update. For this clients should call this contract with `SYSTEM_ADDRESS` as caller and provide `32` bytes parent block hash as input.

Note that the inputs to the `set` method are not sanitized and it is the caller's responsibility to ensure that the ring buffer isn't overrun. This choice was made to support a change in the ring buffer size without having to redeploy a new contract.
```

gballet (approved):
i added a comment about not sanitizing the input of `set` but apart from that, I think this is good to merge.

EIPS/eip-2935.md
g11tech: fixed/updated

EIPS/eip-2935.md
g11tech: we will also need to change the `get` method as well. but there is another consideration: 

the contracts that would directly call the contract would break if we modify ring buffer length and expcted them to mod the input in a new way

PR #8616, opened 2024-06-03
Update EIP-7639: Fix Paris/TheMerge 1th block number

kdeme:
Paris/TheMerge block number is one off.


g11tech (approved):
makes sense terminal block is: https://etherscan.io/block/15537393

PR #8626, opened 2024-06-06
Update EIP-5792: make atomicBatch chain-specific per-batch, not per-call (both in protocol messages and in normative prose)

bumblefudge:
As per a conversation with @pedrouid and @jxom

### DIFF
--- a/EIPS/eip-5792.md
+++ b/EIPS/eip-5792.md
@@ -246,6 +246,7 @@ The capabilities below are for illustrative purposes.
 ### `atomicBatch` Capability
 
 Like the illustrative examples given above and other capabilities to be defined in future EIPs, the capability to execute calls delivered via the above-defined methods in a single transaction can be attested by the wallet in boolean form.
+This capability is expressed separately on each network and should be interpreted as a guarantee only for batches of transactions on that network; batches including calls to multiple networks is out of scope of this capability and this specification.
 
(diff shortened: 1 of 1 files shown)


lukasrosario (changes requested):
requesting that we make the expected behavior for calls across chains clearer. specifically, do calls on one chain need to land before executing calls on the next chain? or can they be made closer to in parallel?

bumblefudge:
i would _assume_ parallel but what do i know about defi. my intuition is that either way, it should be a SHOULD for now and future EIPs would extend the capability definition (or propose an alternative) to be more explicit or opinionated unless people working on such use cases or mechanisms want to speak up and argue one way or the other?

pedrouid:
Hey @lukasrosario that's a great question but I think it applies for both single-chain and multi-chain Atomic Batching

What is the intended behavior for ordering even in a single-chain scenario? Should the Wallet respect the order of the items in the array? Ordering can be lost depending on the serialization and deserialization process involved

I think the Atomic Batching needs to have clearer expectations even without the changes included in this PR

drortirosh:
i'm not sure how can you provide such behavior: there is no way to guarantee cross-chain atomic batch, so it is "Best effort" at best.

pedrouid:
This can be closed in favor of the already merged #8771

PR #8647, opened 2024-06-11
Update EIP-7251: Move to Review

dapplion:


g11tech:
need required EIPs to also move forward

PR #8649, opened 2024-06-11
Update EIP-7600: Add PeerDAS

timbeiko:
Adds PeerDAS to the list of Pectra EIPs in the Meta EIP. I want to discuss this on https://github.com/ethereum/pm/issues/1069 before merging, as this EIP is a bit unique for two reasons. First, it's not a Core EIP, but "only" a networking change. Second, it's not if the special activation pattern for PeerDAS should be mentioned in this EIP. 

Nevertheless, given the overall consensus is "PeerDAS is in Pectra", the Meta EIP should reflect it.


Dropkick13 (commented):
[]()

Dropkick13 (commented):
[]()

PR #8677, opened 2024-06-19
Update EIP-7702: add delegation designation

lightclient:
_This is an idea originally proposed by @vbuterin._

One aspect of 7702 that hasn't sat right with many involved is the fact that some types of 7702 authorizations do not expire. For delegations that are ephemeral, it's important to have a type of authorization that is replayable (which often means irrevocable), because it allows users to distance their activity from the master key of the account, e.g. delegate control to a smart wallet which can have subkeys with different privileges.

Because previous attempts at account delegation have relied on just-in-time provision of authorizations, the only way to make the authorizations replayable has been to make them irrevocable (or revocation impractical, see `maxNonce` mechanism). If we took a step back and thought about what we *really* want to have, regardless of the technical implications, the mechanism we are trying to create should allow an EOA to list on-chain that it has delegated control of it's account to a smart contract.

We've shied away from making changes to the account structure in the trie many times, thus the design of many of these "make EOAs better" proposals. An alternative route is to use an already existing member of the account structure to designate the code the EOA wants to delegate to. With EIP-3541, we now have a rich sequence of code prefixes that cannot be deployed on-chain and can be used to denote other meanings.

This is the core of the proposal: keep the 7702 auth message similar to how it is (although require nonce in sig) and *actually set* a designator in the EOA's account. The delegation designation would be of the form `0xef01 || address` where `address` is the account whose code should be used in place of the designation.

This mechanism should simplify how we reason about authorizations and revocations. Other than it being more complicated than many of the current proposals, I don't see many downsides to this approach.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -61,7 +61,7 @@ At the start of executing the transaction, for each `[chain_id, address, nonce, 
 6. Increase the nonce of `authority` by one.
 7. Add the `authority` account to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
 
-If any of the above steps fail, immediately stop processing that tuple and continue to the next tuple in the list. It will In the case of multiple tuples for the same authority, set the code specified by the address in the first occurrence.
+If any of the above steps fail, immediately stop processing that tuple and continue to the next tuple in the list. It will In the case of multiple tuples for the same authority, set the code specified by the address in the last occurrence.
 
 Note that the signer of an authorization tuple may be different than `tx.origin` of the transaction.
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -13,7 +13,7 @@ requires: 2718, 2929, 2930, 3541, 3607
 
 ## Abstract
 
-Add a new transaction type that adds a list of `[chain_id, address, nonce, y_parity, r, s]` authorization tuples. For each tuple, write a delegation designator `(0xef0000 ++ address)` to the signing account's code. All code reading operations must load the pointed to by the designator.
+Add a new transaction type that adds a list of `[chain_id, address, nonce, y_parity, r, s]` authorization tuples. For each tuple, write a delegation designator `(0xef0000 ++ address)` to the signing account's code. All code reading operations must load the code pointed to by the designator.
 
 ## Motivation
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -1,7 +1,7 @@
 ---
 eip: 7702
-title: Set EOA account code for one transaction
-description: Add a new tx type that sets the code for an EOA during one transaction execution
+title: Set EOA account code
+description: Add a new tx type that sets the code for an EOA during execution
 author: Vitalik Buterin (@vbuterin), Sam Wilson (@SamWilsn), Ansgar Dietrichs (@adietrichs), Matt Garnett (@lightclient)
 discussions-to: https://ethereum-magicians.org/t/eip-set-eoa-account-code-for-one-transaction/19923
 status: Review
@@ -13,7 +13,7 @@ requires: 2718, 2929, 2930, 3541, 3607
 
 ## Abstract
 
-Add a new transaction type that adds a list of `[address, nonce, y_parity, r, s]` authorization tuples, and deploy a delegation designator to the signing accounts so calls into the accounts execute the code at associated `address`.
+Add a new transaction type that adds a list of `[chain_id, address, nonce, y_parity, r, s]` authorization tuples. For each tuple, write a delegation designator `(0xef0000 ++ address)` to the signing account's code. All code reading operations must load the pointed to by the designator.
 
 ## Motivation
 
@@ -45,7 +45,7 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 The fields `chain_id`, `nonce`, `max_priority_fee_per_gas`, `max_fee_per_gas`, `gas_limit`, `destination`, `value`, `data`, and `access_list` of the outer transaction follow the same semantics as [EIP-1559](./eip-1559.md).
 
-The `authorization_list` is a list of tuples that store the address to code which the signer desires to point to from their EOA.
+The `authorization_list` is a list of tuples that store the address to code which the signer desires to execute in the context of their EOA.
 
 The [EIP-2718](./eip-2718.md) `ReceiptPayload` for this transaction is `rlp([status, cumulative_transaction_gas_used, logs_bloom, logs])`.
 
@@ -155,7 +155,7 @@ Specifically:
 * The "code pathways" that are used are code pathways that would, in many cases (though perhaps not all), continue to "make sense" in a pure-smart-contract-wallet world.
 * Hence, it avoids the problem of "creating two separate code ecosystems", because to a large extent they would be the same ecosystem. There would be some workflows that require kludges under this solution that would be better done in some different "more native" under "endgame AA", but this is relatively a small subset.
 * It does not require adding any opcodes, that would become dangling and useless in a post-EOA world.
-* It allows EOAs to temporarily convert themselves into contracts to be included in ERC-4337 bundles, in a way that's compatible with the existing `EntryPoint`.
+* It allows EOAs to masquerade as contracts to be included in ERC-4337 bundles, in a way that's compatible with the existing `EntryPoint`.
 * Once this is implemented, allowing EOAs to migrate permanently is "only one line of code": just add a flag to not set the code back to empty at the end.
 
 ## Backwards Compatibility
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -71,6 +71,8 @@ The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3
 
 For example, `EXTCODESIZE` would return the size of the code pointed to by `address` instead of `24` which would represent the delegation designation. `CALL` would similarly load the code from `address` and execute it in the context of `authority`.
 
+In case a delegation designator points to another designator, creating a potential chain or loop of designators, clients must retrieve only the first code and then stop following the designator chain.
+
 #### Gas Costs
 
 The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `PER_CONTRACT_CODE_BASE_COST * authorization list length`.
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -8,12 +8,12 @@ status: Review
 type: Standards Track
 category: Core
 created: 2024-05-07
-requires: 2718, 2929, 2930
+requires: 2718, 2929, 2930, 3541, 3607
 ---
 
 ## Abstract
 
-Add a new transaction type that adds a list of `[address, y_parity, r, s]` authorization tuples, and converts the signing accounts (not necessarily the same as the `tx.origin`) into smart contract wallets for the duration of that transaction.
+Add a new transaction type that adds a list of `[address, nonce, y_parity, r, s]` authorization tuples, and converts the signing accounts (not necessarily the same as the `tx.origin`) into smart contract wallets for the duration of that transaction.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


Jrachman:
If an EOA were to have an active authorization with this approach and then the EOA sends a regular, non-7702 transaction after, will that invalidate the active authorization when the EOA then wants to execute a 7702 transaction?

EIPS/eip-7702.md
```diff
@@ -40,38 +40,47 @@ We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction"
...
+
 #### Gas Costs
 
 The intrinsic cost of the new transaction is inherited from [EIP-2930](./eip-2930.md), specifically `21000 + 16 * non-zero calldata bytes + 4 * zero calldata bytes + 1900 * access list storage key count + 2400 * access list address count`. Additionally, we add a cost of `PER_CONTRACT_CODE_BASE_COST * authorization list length`.
 
 The transaction sender will pay for all authorization tuples, regardless of validity or duplication.
 
+#### Transaction Origination
+
+Modify the restriction put in place by [EIP-3607](./eip-3607.md) to allow EOAs whose code is a valid delegation designation, i.e. `0xef01 || address`, to continue to originate transactions. Accounts with any other code values may not originate transactions.
```
shemnon, line 82: To preserve design space I would recommend at least one version byte outside of the EF type byte.

```
<prefix> || <version>  || <address>
<prefix> := '0xef01'
<version> := '0x00'
<address> := the address
```

EIPS/eip-7702.md
```diff
@@ -8,12 +8,12 @@ status: Review
...
 category: Core
 created: 2024-05-07
-requires: 2718, 2929, 2930
+requires: 2718, 2929, 2930, 3541, 3607
 ---
 
 ## Abstract
 
-Add a new transaction type that adds a list of `[address, y_parity, r, s]` authorization tuples, and converts the signing accounts (not necessarily the same as the `tx.origin`) into smart contract wallets for the duration of that transaction.
+Add a new transaction type that adds a list of `[address, nonce, y_parity, r, s]` authorization tuples, and converts the signing accounts (not necessarily the same as the `tx.origin`) into smart contract wallets for the duration of that transaction.
```
rmeissner, line 16: It is slightly confusing that it says "for the duration of that transaction", but as far as I understand this new version the delegation is persisted.

EIPS/eip-7702.md
```diff
@@ -108,9 +117,7 @@ An alternative to adding chain ID could be to sign over the code the address poi
...
 
-A hotly debated element of this EIP is the need for in-protocol revocation. Although it is possible to implement revocation logic within delegated code, for some this isn't sufficient and it is the duty of the protocol to provide a revocation option of last resort.
-
-For this reason, the proposal provides two separate schemes. The first is an eternal delegation to a smart contract. At the protocol level, it is not possible to revoke. However, the contract you delegate to is one which can expect to use in your account for perpetuity; similar to how smart contract wallet users deploy proxy contracts to their account and point the target at a wallet implementation. The second is a scoped delegation with a revocation mechanism based on the EOA's nonce. This is a safer to begin using smart contract code in the context of your EOA without potentially committing to a specific piece of code forever.
+Unlike previous versions of this EIP and EIPs similar, the delegation designation can be revoked at anytime signing and sending a EIP-7702 authorization to a new target with the account's current nonce. Without such action, a delegation will remain valid in perpetuity.
```
rmeissner, line 120: Should there be a way to reset the delegation compeltely?

EIPS/eip-7702.md
```diff
@@ -40,38 +40,47 @@ We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction"
...
+At the start of executing the transaction, for each `[chain_id, address, nonce, y_parity, r, s]` tuple:
 
-1. `authority = ecrecover(keccak(MAGIC || rlp([chain_id, [nonce], address])), y_parity, r, s]`
+1. `authority = ecrecover(keccak(MAGIC || rlp([chain_id, address, nonce])), y_parity, r, s]`
 2. Verify the chain id is either 0 or the chain's current ID.
-3. Verify that the code of `authority` is empty.
-4. If nonce list item is length one, verify the nonce of `authority` is equal to `nonce`.
-5. Set the code of `authority` to code associated with `address`.
-6. Add the `authority` account to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
+3. Verify that the code of `authority` is either empty or already delegated.
```
rmeissner, line 58: `or already delegated` this is basically a check that the code of the `authority` is in the format of `0xef01 || address`, right? 

I assume this check is to prevent that this flow in case that any other code besides the delegation is set, right?

drortirosh: very confusing. 
AFAIK, the whole purpose was to define temporary delegations.
If it is "already delegated" it means:
1. delegation are persisted - which is not defined by this doc (if/when we can change the delegation?)
2. there is no need for an `authorizer` entry, since it is already on-chain
3. if the repeated authorizer signature is the mechanism to **change** an on-chain delegation, then please specify it explicitly. (and it has a whole bunch of security implications)

EIPS/eip-7702.md
lightclient: good point

EIPS/eip-7702.md
lightclient: Yes this is intended to maintain the 3607 requirement that EOAs may not have code. We made a special allowance for delegation designations.

lightclient:
> If an EOA were to have an active authorization with this approach and then the EOA sends a regular, non-7702 transaction after, will that invalidate the active authorization when the EOA then wants to execute a 7702 transaction?

@Jrachman no, the authorization would persist in the account. As spec'd it is currently impossible to "invalidate" a delegation, it is only possible to delegate to something else.

shemnon:
@lightclient then there should be an "undelegate" option.  Such as delegating to yourself or keeping the address portion of the message empty and just signing two bytes.

lightclient:
@shemnon what purpose would an undelegation mechanism serve vs. simply delegating to zero address like I mentioned?

EIPS/eip-7702.md
```diff
@@ -40,38 +40,47 @@ We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction"
...
+4. Verify the nonce of `authority` is equal to `nonce`.
+5. Set the code of `authority` to be `0xef0100 || address`. This is a delegation designation.
+6. Increase the nonce of `authority` by one.
+7. Add the `authority` account to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
 
 If any of the above steps fail, immediately stop processing that tuple and continue to the next tuple in the list. It will In the case of multiple tuples for the same authority, set the code specified by the address in the first occurrence.
```
frangio, line 64: I don't think it's necessary to specify this anymore because the nonce check will automatically fail:

```suggestion
If any of the above steps fail, immediately stop processing that tuple and continue to the next tuple in the list.
```

smartprogrammer93:
@lightclient, if you delegate to zero address, wouldn't any call to your account go to zero address, including ones with value?

Delegating to self also would be problematic as first opcode (ef) would result in an abort. Maybe we should mention how we should deal with this? Maybe disallow it?

shemnon:
It would be trapped out to _delete_ the ef code.  In sone L2s there may be code at zeros, which is why I didn't recommend that.  Removal would be to restore "pure EOA" behavior, where calling the EOA only deposits value.

EIPS/eip-7702.md
```diff
@@ -40,38 +40,47 @@ We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction"
...
 
-At the end of the transaction, set the code of each `authority` back to empty.
-
 Note that the signer of an authorization tuple may be different than `tx.origin` of the transaction.
 
+##### Delegation Designation
+
+The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541) to designate the code has a special purpose. This designator requires all code retrieving operations follow the address pointer to fill the accounts observable code. The following instructions are impacted: `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`.
+
+For example, `EXTCODESIZE` would return the size of the code pointed to by `address` instead of `24` which would represent the delegation designation. `CALL` would similarly load the code from `address` and execute it in the context of `authority`.
```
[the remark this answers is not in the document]
smartprogrammer93: I thought about that and i believe it makes no sense to support chainning. Imo, if EOA1 points to EOA2 where EOA2 has a designated address, it just 3xecut3d `EF` which is abort in evm and that is it. No need to handle this as a special case.

smartprogrammer93:
@shemnon then we can trap out delegating to self as removing and delete the code. But this needs to be specified in the spec then

EIPS/eip-7702.md
drortirosh: ok. so remove the "temporary" in text: it is no longer temporary, but permanent assignment of a "proxy" address.

EIPS/eip-7702.md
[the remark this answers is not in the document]
drortirosh: I think multiple tuples of the same address should be forbidden: the signer has all the information to detect and remove duplicate entries.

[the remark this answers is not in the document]
Scamreno: Ok

Diamondb32 (approved):

EIPS/eip-7702.md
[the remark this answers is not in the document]
frangio: > the signer has all the information to detect and remove duplicate entries.

I don't understand this. The signatures may already have been generated and the signer may not be available to create a new signature that combines them into one.

I see no reason to reject duplicates if each occurrence is paid for.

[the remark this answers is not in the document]
yoavw: Why would the sender of the 7702 transaction include duplicates and pay for them, even if allowed?  It won't make both implementations available, only one of them.  Any further tuples for the same account seem like a waste of gas.

EIPS/eip-7702.md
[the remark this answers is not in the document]
drortirosh: I think the best way is to define that delegation is not recursive: if you delegate to an already-delegated entity, the "magic" doesn't happen twice: you end up pointing to a bad code (starting with `0xef`)

EIPS/eip-7702.md
drortirosh: then I suggest specifying it explicitly.

EIPS/eip-7702.md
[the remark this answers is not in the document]
smartprogrammer93: @drortirosh it is already done here: https://github.com/ethereum/EIPs/pull/8677/commits/358fc70c92e39b536e4b2482893c286d5eee6985

EIPS/eip-7702.md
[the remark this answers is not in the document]
frangio: If I have two signed delegations from the same account with nonces `n` and `n + 1`, and I need to submit an op from this account that requires the latest delegation, I need to submit both... The first one is needed to increment the nonce so that the second one becomes valid.

EIPS/eip-7702.md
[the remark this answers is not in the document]
shemnon: What about delegating to code that is empty? Same as any other empty code account?

That happens in a few  cases (a) an delegated EOA that has delegated and (b) an empty account, (c) a balance-only account (nonce == 0).  This should be called out in the testing section:

Tests that do all combinations of the following:
* Delegate to (an empty account  | a balance only account | an undelegated EOA), which later (gains balance | becomes an EOA | becomes a contract | becomes a delegated EOA)
* call the EOA delegating (in the same TX | in a latter TX after the permutation happens).

EIPS/eip-7702.md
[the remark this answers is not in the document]
frangio: I thought the concerns were about being able to bump nonces by an arbitrary amount. In this case it's always incrementing by 1 (per signature) so I don't imagine there would be an issue with that.

EIPS/eip-7702.md
```diff
@@ -8,12 +8,12 @@ status: Review
...
 category: Core
 created: 2024-05-07
-requires: 2718, 2929, 2930
+requires: 2718, 2929, 2930, 3541, 3607
 ---
 
 ## Abstract
 
-Add a new transaction type that adds a list of `[address, y_parity, r, s]` authorization tuples, and converts the signing accounts (not necessarily the same as the `tx.origin`) into smart contract wallets for the duration of that transaction.
+Add a new transaction type that adds a list of `[address, nonce, y_parity, r, s]` authorization tuples, and deploy a delegation designator to the signing accounts so calls into the accounts execute the code at associated `address`.
```
rojotek, line 16: It would probably make sense to update the title and description of the EIP to match this as well.

EIPS/eip-7702.md
```diff
@@ -40,37 +40,48 @@ We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction"
...
 rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, value, data, access_list, authorization_list, signature_y_parity, signature_r, signature_s])
 
-authorization_list = [[chain_id, address, [nonce], y_parity, r, s], ...]
+authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 ```
 
 The fields `chain_id`, `nonce`, `max_priority_fee_per_gas`, `max_fee_per_gas`, `gas_limit`, `destination`, `value`, `data`, and `access_list` of the outer transaction follow the same semantics as [EIP-1559](./eip-1559.md).
 
-The `authorization_list` is a list of tuples that store the address to code which the signer desires to set in their EOA temporarily. The third element is a list item mimicking an optional value. When the list length is zero, consider the authorization nonce to be null. When the list length is one, consider the single integer value to be the provided nonce authorization. Other lengths and value types in this optional are invalid.
+The `authorization_list` is a list of tuples that store the address to code which the signer desires to set in their EOA temporarily.
```
drortirosh, line 48: According to the text below, the delegation is not "temporary" (for the current transaction) but fixed (changes the account trie): a delegated address is set that can be used in future transactions.
So please remove all "temporary" mentions, and add reasonable cost (first 7702 tx creates an account in the storage trie, and future txs may modify a storage)

EIPS/eip-7702.md
```diff
@@ -1,19 +1,19 @@
...
 category: Core
 created: 2024-05-07
-requires: 2718, 2929, 2930
+requires: 2718, 2929, 2930, 3541, 3607
 ---
 
 ## Abstract
 
-Add a new transaction type that adds a list of `[address, y_parity, r, s]` authorization tuples, and converts the signing accounts (not necessarily the same as the `tx.origin`) into smart contract wallets for the duration of that transaction.
+Add a new transaction type that adds a list of `[chain_id, address, nonce, y_parity, r, s]` authorization tuples. For each tuple, write a delegation designator `(0xef0000 ++ address)` to the signing account's code. All code reading operations must load the pointed to by the designator.
```
ryanschneider, line 16: Minor: last sentence is missing a noun (must load the ??? pointed to)

Parasos (commented):
EIPS/eip-7702.md

EIPS/eip-7702.md
lightclient: fixed thx

lightclient:
Last call on this, will merge later today unless there is more feedback.

frangio:
See https://github.com/ethereum/EIPs/pull/8677#discussion_r1647794701. IMO this part makes no sense in the new design and should be removed:

> In the case of multiple tuples for the same authority, set the code specified by the address in the first occurrence.

lightclient:
@frangio this seems reasonable, I will update it.

drortirosh:
The spec still doesn't specify the cost of the operation.
it references "PER_CONTRACT_CODE_BASE_COST", which is not defined.
"PER_AUTH_BASE_COST" is defined but if that constant should be used, it is too cheap: being only 2500 gas, it is half of the SSTORE "modify" cost and 1/8th of the cost of SSTORE to empty cell

lightclient:
@drortirosh sorry I though I updated to remove the ref to `PER_CONTRACT_CODE_BASE_COST`. It's fixed now.

The cost is not finalized and should definitely be updated now that code is deployed to the account. In particular, we need to update it to deal with creating new accounts (usually 25k gas). I'll leave that for another PR. If you have a suggestion for gas schedule, please post it on FEM.

Issue #8546, opened 2024-05-10
Distinguish between `h2` and `h3` more obviously


SamWilsn:
In the rendered output, the `h2` and `h3` styles are too similar.

![IMG_20240510_085800_315](https://github.com/ethereum/EIPs/assets/57262657/533affe1-a6a4-4c55-94bc-75e4f060703b)

MartiPresa:
Hi, is this issue available?

Dropkick13:
[]()

Scamreno:
#7991

Ali1994Baker:
Hallo

Ali1994Baker:
Hallo lk willen mij zalf mij code van 2FA

Op zo 21 jul 2024 om 18:15 schreef Bossman74222 ***@***.***>

Ali1994Baker:
> Hallo

Ali1994Baker:
Hallo Yes, I am working on it.

Op ma 5 aug 2024 om 02:21 schreef github-actions[bot] <
***@***.***>

SamWilsn:
Still an issue.

poojaranjan:
@MartiPresa 
Please reach us at [EthCatHerders-EIP Editors Discord](https://dsc.gg/ech ) if you have any questions about this Issue.

MukulKolpe:
Hi @poojaranjan @SamWilsn, can I work on this issue if it’s still open?

anushkasomani:
hi, @SamWilsn is this issue still open

martinvibes:
hey @SamWilsn is this issue still open

Shaurya2k06:
Regarding Issue: 
What's being asked here? Need clarification regarding whether its a styling issue or a rendering issue.

```
h2 {
  font-size: 1.8rem;
  font-weight: 600;
}

h3 {
  font-size: 1.3rem;
  font-weight: 500;
}
```

Won't something like this do the trick in styles.scss?

PR #8679, opened 2024-06-20
Update EIP-7702: Reorder authorization fields for authority recovery

onbjerg:
Currently items in the authorization list are ordered:

- Chain ID
- Address
- (Optional) nonce

but in order to recover the authority, we need to reorder the fields to

- Chain ID
- (Optional) Nonce
- Address

which in some cases necessitates a secondary type to encode the authorization for recovery. Using the same order for the fields simplifies implementation slightly.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -53,7 +53,7 @@ The [EIP-2718](./eip-2718.md) `ReceiptPayload` for this transaction is `rlp([sta
 
 At the start of executing the transaction, for each `[chain_id, address, [nonce], y_parity, r, s]` tuple:
 
-1. `authority = ecrecover(keccak(MAGIC || rlp([chain_id, [nonce], address])), y_parity, r, s]`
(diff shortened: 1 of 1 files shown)


lightclient:
This ordering was my mistake, they should be ordered the same throughout. I would like to know from other teams if it's okay to update this for devnet-1? Or will we just update this for future devnets?

onbjerg:
Speaking for reth we're ok with changing it for devnet-1:)

Scamreno:
#8682 #6976 #8546

Dropkick13 (commented):
[]()

sudeepdino008:
> This ordering was my mistake, they should be ordered the same throughout. I would like to know from other teams if it's okay to update this for devnet-1? Or will we just update this for future devnets?

Erigon is okay too.

Dropkick13:
[]()

ak88:
> This ordering was my mistake, they should be ordered the same throughout. I would like to know from other teams if it's okay to update this for devnet-1? Or will we just update this for future devnets?

Nethermind is okay with this too

lightclient (approved):

PR #8693, opened 2024-06-26
Update EIP-7620: Remove initcode/runtime container definitions

pdobacz:
Align with https://github.com/ipsilon/eof/pull/136. Justification of the removal of definitions (from the Megaspec PR):

>  proposed removing the "initcode container" and "runtime container" definitions, as I think they are unwieldy and suggest a particular implementation (the so-called inside-out determination of container kind). Also, these definitions cross paths with those used in the context of initcontainer_index and deploy_container_index of opcodes

The Megaspec PR isn't yet merged, so if anyone feels strongly about keeping the definitions in, please let know


Parasos:
allot information needed Reload

Orhanoguz3567:
B

Oğuz oto elektrik Yönetim


26 Haz 2024 Çar, saat 21:05 tarihinde Para ***@***.***> şunu
yazdı:

PR #8707, opened 2024-06-30
Update EIP-4762: Add clarifications for the system contracts

g11tech:
system contract updates

### DIFF
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -41,8 +41,8 @@ Whenever the state is read, one or more of the access events of the form`(addres
 
 When:
 
- 1. a non-precompile or a system contract is the target of a `*CALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode,
- 2. a non-precompile or a system contract  is the target address of a contract creation whose initcode starts execution,
+ 1. a non-precompile which is also not a system contract is the target of a `*CALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode,
+ 2. a non-precompile which is also not a system contract is the target address of a contract creation whose initcode starts execution,
  3. any address is the target of the `BALANCE` opcode
  4. a _deployed_ contract calls `CODECOPY`
 
@@ -62,7 +62,7 @@ If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), 
 
 Note: when checking for the existence of the `callee`, the existence check is done by validating that there is an extension-and-suffix tree at the corresponding stem, and does not rely on `CODEHASH_LEAF_KEY`.
 
-When calling `EXTCODEHASH` on a non-precompile and neither system contract target, process the access event:
+When calling `EXTCODEHASH`, process the access event:
 
 ```
 (address, 0, CODEHASH_LEAF_KEY)
@@ -255,7 +255,7 @@ When (and only when) calling a system contract either
 
  the system contract code chunks and account accesses should not appear in the witness as these should be known/cached in the clients. However any other accesses and all writes should appear in the witness.
 
-Also Corresponding witness costs need to be charged for _precompile/opcode resolution_ but are not charged in the _system call_.
+Also corresponding witness costs need to be charged for _precompile/opcode resolution_ but are not charged in the _system call_.
 
 ### Account abstraction
 
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -41,8 +41,8 @@ Whenever the state is read, one or more of the access events of the form`(addres
 
 When:
 
- 1. a non-precompile address is the target of a `*CALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode, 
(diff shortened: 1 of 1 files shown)


Gabo139 (approved):

EIPS/eip-4762.md
```diff
@@ -251,7 +251,16 @@ None of:
...
+### System contracts
+
+When (and only when) calling a system contract either
+
+ * _via a system call_ or
+ * _to resolve a precompile/opcode_,
+
+ the system contract code chunks and account accesses should not appear in the witness as these should be known/cached in the clients. However any other accesses and all writes should appear in the witness.
+
+Also Corresponding witness costs need to be charged for _precompile/opcode resolution_ but are not charged in the _system call_.
```
jsign, line 263: Is this clarification needed? As in, who would could charge for system call calls?

EIPS/eip-4762.md
```diff
@@ -52,23 +52,23 @@ process this access event:
 (address, 0, BASIC_DATA_LEAF_KEY)
 ```
 
-Note: a non-value-bearing `SELFDESTRUCT` or `*CALL`, targetting a precompile, will not cause the `BASIC_DATA_LEAF_KEY` to be added to the witness.
+Note: a non-value-bearing `SELFDESTRUCT` or `*CALL`, targetting a precompile or a system contract, will not cause the `BASIC_DATA_LEAF_KEY` to be added to the witness.
```
gballet, line 55: but the gas costs will be charged because of https://github.com/ethereum/EIPs/pull/8707/files#diff-cd2aaa222faeb6081a41605386a509b3c4839285cd15f1b0ebe009318707914eR44 right?

gballet (approved):
LGTM. Left a comment, requesting a clarification in the text.

EIPS/eip-4762.md
gballet: if so, I would add a sentence to confirm it, so that people are not confused.

EIPS/eip-4762.md
```diff
@@ -251,7 +251,16 @@ None of:
...
 
-Note: When (and only when) calling a system contract _via a system call_, the code chunks and account should not appear in the witness.
+### System contracts
+
+When (and only when) calling a system contract either
+
+ * _via a system call_ or
+ * _to resolve a precompile/opcode_,
+
+ the system contract code chunks and account accesses should not appear in the witness as these should be known/cached in the clients. However any other accesses and all writes should appear in the witness.
```
jsign, line 261: > the system contract code chunks and account accesses should not appear in the witness as these should be known/cached in the clients

What does "account accesses" mean? If a system call or resolving a precompile/opcode reads some state, that should be in the witness right?
The only thing that is known/cached is the system contract code.

EIPS/eip-4762.md
g11tech: I think it should be 2929 cold or hot account access cost (and not cold 6800 witness cost) for system contracts

and for precompiles it would be warm access 2929 access cost

EIPS/eip-4762.md
g11tech: account accesses: account header accesses

and yes only of system contract's header access is not to be added. if the contract make any accesses while doing normal evm code of some other address it would be included.

may be i should reword it?

EIPS/eip-4762.md
g11tech: i slightly favor having this for completeness and context sake.

EIPS/eip-4762.md
jsign: So this part is then kind of repeating this said earlier?:
>  1. a non-precompile or a system contract is the target of a `*CALL`, `SELFDESTRUCT`, `EXTCODESIZE`, or `EXTCODECOPY` opcode,

If that's the case, I'd remove it just to remove potential confusion if this is referring to something else. But if you think it helps to reinforce, I'm not strongly against it but I'd prefer to put "account **header** accesses" yes.

EIPS/eip-4762.md
jsign: Fine to me. Just one extra nit: `s/Corresponding/corresponding`

EIPS/eip-4762.md
```diff
@@ -52,23 +52,23 @@ process this access event:
...
+If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), whether or not the `callee` is a precompile or a system contract, process this additional access event:
 
 ```
 (caller, 0, BASIC_DATA_LEAF_KEY)
 ```
 
 Note: when checking for the existence of the `callee`, the existence check is done by validating that there is an extension-and-suffix tree at the corresponding stem, and does not rely on `CODEHASH_LEAF_KEY`.
 
-When calling `EXTCODEHASH` on a non-precompile target, process the access event:
+When calling `EXTCODEHASH` on a non-precompile and neither system contract target, process the access event:
```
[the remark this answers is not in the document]
lu-pinto: or you don't even need this change here, since you clarify on line 71 that this does not apply to system contracts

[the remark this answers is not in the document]
g11tech: good idea, will update, thanks

EIPS/eip-4762.md
g11tech: clarified that leaf won't be accessed for system contract as well

PR #8712, opened 2024-07-02
eip-158 deactivation eip

gballet:


EIPS/eip-template.md
```diff
@@ -0,0 +1,59 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7733
```

EIPS/eip-template.md
```diff
@@ -0,0 +1,59 @@
...
+title: EIP-158 deactivation
+description: Deactivates EIP-158 in order to avoid conflicts in a context in which statelessness and EIP-7702 are active
+author: Guillaume Ballet (@gballet)
+discussions-to: https://ethereum-magicians.org/t/eip-158-deactivation/20445
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-7733-eip-158-deactivation/20445
```

EIPS/eip-template.md
```diff
@@ -0,0 +1,59 @@
+---
+title: EIP-158 deactivation
```
abcoathup, line 2:
```suggestion
title: Deactivate EIP-158
```
Personal preference on order.  (I am not an EIP editor)

Qadeem411 (approved):

Qadeem411 (approved):

Qadeem411 (approved):

EIPS/eip-7733.md
```diff
@@ -0,0 +1,60 @@
...
+requires: 158, 7702
+---
+
+## Abstract
+
+Deactivate [EIP-158](./eip-158.md).
+
+## Motivation
+
+After [EIP-6780](./eip-6780.md), the `SELFDESTRUCT` instruction was neutered, so that contracts could no longer be deleted. Contract deletions can still occur, however, via EIP-158. While [EIP-7702](./eip-7702.md) this can not happen, this latter EIP introduces the possibility for an empty EoA to have state. This is causing some issues when interfacing with verkle, for which state deletion is not possible.
```
jochem-brouwer, line 20: The 7702 sentence does not read very well, I understand what you mean though.

EIPS/eip-7733.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+|Name|Value|
+|----|-----|
+|`FORK_TIME`|TBD|
+
+At `block.timestamp > FORK_TIME`, EIP-158 rules no longer apply.
```
jochem-brouwer, line 32:
```suggestion
At `block.timestamp >= FORK_TIME`, EIP-158 rules no longer apply.
```

jochem-brouwer (commented):
Two points. 

Big support in general, I had some concerns about 158 with 7702, especially with the EOA account being empty but having state (and thus by 158 getting deleted)

jochem-brouwer:
BTW, I think you mean EIP-161 

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-607.md
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md

EIPS/eip-7733.md
```diff
@@ -0,0 +1,60 @@
+---
+eip: 7733
+title: Deactivate EIP-158
+description: Deactivates EIP-158 in order to avoid conflicts in a context in which statelessness and EIP-7702 are active
```
SamWilsn, line 4: I'd recommend removing EIP-7702 from the description. Reading 7702 is not a requirement to understand this proposal; it's more a motivating use case.

EIPS/eip-7733.md
```diff
@@ -0,0 +1,60 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-07-02
+requires: 158, 7702
+---
+
+## Abstract
+
+Deactivate [EIP-158](./eip-158.md).
```
SamWilsn, line 16: You should expand a little bit on your abstract. This certainly satisfies the terse requirement, but doesn't give the reader any information about the content of this document. Would you mind adding a sentence explaining, at a high level, what deactivating EIP-158 entails?

EIPS/eip-7733.md
```diff
@@ -0,0 +1,60 @@
...
+
+EIP-158 was meant as a temporary measure to combat the "Shanghai attacks". Now that this is attack has been mitigated by other means, it can be deactivated.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+|Name|Value|
+|----|-----|
+|`FORK_TIME`|TBD|
```
SamWilsn, line 30:
```suggestion
|`FORK_TIME`|TBD <!-- TODO --> |
```

If you use an HTML-style comment, the linter will make sure you replace it before moving to the Review status.

EIPS/eip-7733.md
SamWilsn, line 32: The wording here is a bit ambiguous. Are you trying to undo this rule as well:

> Zero-value calls and zero-value suicides no longer consume the 25000 account creation gas cost in any circumstance

EIPS/eip-7733.md
```diff
@@ -0,0 +1,60 @@
...
+
+TODO
+
+## Reference Implementation
+
+TODO
+
+## Security Considerations
+
+Needs discussion.
```
SamWilsn, line 56:
```suggestion
## Test Cases

TODO <!-- TODO -->

## Reference Implementation

TODO <!-- TODO -->

## Security Considerations

Needs discussion. <!-- TODO -->
```

EIPS/eip-7733.md
petertdavies, line 32: > At `block.timestamp > FORK_TIME`, EIP-158 rules no longer apply.

EIP-161 is 8 years old. Its behaviour has been assumed by EIP authors and implementers. You really ought to explicitly spell out what you actually want to do, rather than make some vague comments about deactivating it. I'm also fairly confident that you meant something different than what you wrote implies.

EIP-161 (see previous comment) has 4 effects:

a. When an account is `CREATE`d the nonce is incremented.
b. The new account gas cost is not charged if the account wouldn't be created because of rule c.
c. If an account would come into existence in an _empty_ state, it remains non-existent.
d. Under certain triggering conditions, empty accounts that are marked for deletion and deleted at the the end of the transaction.

As it stands, rules a, b and c are uncontroversial. Rule d is now a nullity because all empty accounts have been deleted and it is impossible to create new ones.

Your motivation for writing this EIP seems related to the edgecase that occurs if EIP-7702 code setting is used with an EOA that doesn't exist. EIP-7702 does not explain what happens in this case, which is a significant deficiency. If this edgecase results in the creation of empty accounts, it is a serious DOS vulnerability because an account can be created for only 2500 gas (other methods require at least 21000).

Personally I am opposed to a version of EIP-7702 that creates empty accounts. If it is decided that EIP-7702 should create empty accounts, then we can then discuss changing the rules about them. Until then this EIP doesn't change the spec in a meaningful way.

petertdavies (changes requested):
This EIP is so vague that it is not entirely clear to me what you propose to change.

It seems to have been written to attempt to deal with an unfortunate edgecase that arises from one reading of EIP-7702. You should probably fixing EIP-7702 to handle this edgecase better instead.

EIPS/eip-7733.md
```diff
@@ -0,0 +1,60 @@
+---
+eip: 7733
+title: Deactivate EIP-158
+description: Deactivates EIP-158 in order to avoid conflicts in a context in which statelessness and EIP-7702 are active
+author: Guillaume Ballet (@gballet)
+discussions-to: https://ethereum-magicians.org/t/eip-7733-eip-158-deactivation/20445
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-7733-deactivate-eip-158/20445
```
Updated title

gballet:
> This EIP is so vague that it is not entirely clear to me what you propose to change.

the change is the deactivation of eip 158/161... How could I possibly make it any clearer in your view?
 
> It seems to have been written to attempt to deal with an unfortunate edgecase that arises from one reading of EIP-7702. You should probably fixing EIP-7702 to handle this edgecase better instead.

That is a much more complex fix than simply deactivating eip 158, and it would effectively be the same thing.

EIPS/eip-7733.md
gballet: > As it stands, rules a, b and c are uncontroversial. Rule d is now a nullity because all empty accounts have been deleted and it is impossible to create new ones.

eip 7702 changes that, precisely, as clearly stated in the text of the document and your own comment.

> , it is a serious DOS vulnerability because an account can be created for only 2500 gas (other methods require at least 21000).

that isn't the case, you could only call into an EoA with a 0 nonce with a 7702 tx, and send its balance to a different address. In this case, it would be deleted according to your 'd' rule. You would have to pay the 21000 gas, so no DOS vector.

> Personally I am opposed to a version of EIP-7702 that creates empty accounts. 
the only thing that would prevent that would be to increase the nonce of the target account, which has received a lot of negative feedback since it apparently makes it much more difficult for AA to work properly.

yes, that is what I just heard in ACDE, in which case this EIP is moot. I'm going to move it to withdrawn.

> 
> As it stands, rules a, b and c are uncontroversial.

> Rule d is now a nullity because all empty accounts have been deleted and it is impossible to create new ones.

that is no longer the case once 7702 is rolled out. this is the one that I want to deprecate, in fact.

> 
> (...) it is a serious DOS vulnerability because an account can be created for only 2500 gas (other methods require at least 21000).

this is only possible to trigger if you send a 7702 tx to an empty nonce, and drain its funds. You would have to pay the tx costs and there is no DOS vector.
 
> 
> Personally I am opposed to a version of EIP-7702 that creates empty accounts. If it is decided that EIP-7702 should create empty accounts, then we can then discuss changing the rules about them. Until then this EIP doesn't change the spec in a meaningful way.

The problem is that the only way to fix this would be to increase the nonce of the called contract, which I'm told is wreaking havoc with AA. It's also a problem if an alternate method for the verkle transition is chosen.

PR #8719, opened 2024-07-03
Update EIP-3540: Ensure num_container_sections is correct

jochem-brouwer:
The text itself has this text: 

> the number of container sections must not exceed `256`. The number of container sections may not be `0`, if declared in the header

To encode 255 container sections, `0x0100` should thus be used, since `0` is disallowed. See also https://github.com/ipsilon/eof/pull/141


pdobacz (approved):
LGTM, but need author's approval to merge

chfast (approved):

Magken91:
Thank you

PR #8727, opened 2024-07-07
Add .editorconfig

mazesec:
This pull request adds a .editorconfig file to ensure consistent coding styles for Python, HTML, and Solidity... files.


SamWilsn:
I'm not sure this will have much of an effect unless we also add some kind of formatter, which I'm not opposed to.

PR #8735, opened 2024-07-12
Add EIP: Uncouple blob count between CL and EL

ralexstokes:

--

 - The build passes.

### DIFF
--- a/EIPS/eip-7742.md
+++ b/EIPS/eip-7742.md
@@ -1,0 +1,114 @@
+---
+eip: 7742
+title: Uncouple blob count between CL and EL
+description: Have CL verify blob maximum and have EL get target value from CL
+author: Alex Stokes (@ralexstokes)
+discussions-to: https://ethereum-magicians.org/t/eip-7742-uncouple-blob-count-between-cl-and-el/20550
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-07-12
+requires: 4844
+---
+
+## Abstract
+
+Update blob maximum and target verification from [EIP-4844](./eip-4844.md).
+
+The execution layer no longer verifies the blob maximum and receives the target dynamically from the consensus layer.
+
+## Motivation
+
+Following EIP-4844, the execution layer (EL) maintains a hard-coded blob target value and blob maximum value. Given the relationship
+of the EL and the consensus layer (CL) node software, the verification of the blob maximum is redundant so it can be removed
+entirely without any change in security. This EIP also changes how the EL sources the current blob target value for two reasons:
+
+1) Gain more flexibility over the value, rather than the static `TARGET == MAX // 2` relation in EIP-4844.
+2) Uncouple development and deployment of the CL and EL layers in the event it is desirable to change the blob target value.
+
+## Specification
+
+| constants                    | value                                        |
+|---                           |---                                           |
+| `FORK_TIMESTAMP`             | TBD                                          |
+
+### Background
+
+The data facility introduced via EIP-4844 adds blobs to Ethereum blocks, which are simply fixed sets of data that can be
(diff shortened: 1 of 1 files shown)


EIPS/eip-7733.md
```diff
@@ -0,0 +1,114 @@
+---
+eip: TBD
```
abcoathup, line 2:
```suggestion
eip: 7742
```

EIPS/eip-7733.md
```diff
@@ -0,0 +1,114 @@
...
+1) Gain more flexibility over the value, rather than the static `TARGET == MAX // 2` relation in EIP-4844.
+2) Uncouple development and deployment of the CL and EL layers in the event it is desirable to change the blob target value.
+
+## Specification
+
+| constants                    | value                                        |
+|---                           |---                                           |
+| `FORK_TIMESTAMP`             | TBD                                          |
+
+### Background
```
mkalinin, line 35: IMHO, this should be moved to Motivation or partially to Rationale

EIPS/eip-7733.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Target amount of blobs per block
+
+The target is currently specified as a fixed value in relation to the blob count. The Ethereum community intends to increase
+the blob parameters as part of its scaling strategy and the ability to have a more flexible target value in relation to
+the blob max is desirable to reduce rigidity in this protocol parameter.
+
+Even if the EL keeps a fixed target value based on the max, removing the max implies the EL would not know what the target
+value should be. To address this lack of information, this EIP proposes the CL sends the current target value to the EL
+with each provided payload over the Engine API. The EL block header will also need to be extended with this target value
```
mkalinin, line 63: AFAICS, `target_blob_gas_per_block` is used to compute `excess_blob_gas`. Would it be possible to send the target as a part of `PayloadAttributes` so the EL can compute the excess and make CL validate that the excess is computed correctly given the target and parent’s excess?

ralexstokes: the CL part of this change is here: https://github.com/ethereum/consensus-specs/pull/3800

it hasn't been updated yet but yes, the CL will send the target value in the `PayloadAttributes` so the EL can build a valid block

mkalinin: Sorry, i wasn’t clear enough on the CL validation part, it seems to be possible to me to not include `target_blob_gas_per_block` into EL block by doing the following:
1) Send this parameter to payload build process so EL can correctly compute `excess_blob_gas`
2) Make CL validate that the computation was done correctly by utilising a simple math: `assert payload.excess_blob_gas == min(0, state.latest_execution_payload_header.excess_blob_gas + target_blob_gas_per_block - blob_gas_used)`

ralexstokes: would you still commit the `target_blob_gas` in the EL header?

it seems to break the ability for the EL to independently verify the `excess_blob_gas` (and by extension the blob base fee) if it doesn't have access to the target value

if we keep the header extension then your change is just moving this check from the EL to the CL, which is fine in principle although I could argue violates the separation of concerns of "blob (max) limit check on the CL for DoS reasons" and "blob fee accounting on the EL"

mkalinin: > it seems to break the ability for the EL to independently verify the `excess_blob_gas` (and by extension the blob base fee) if it doesn't have access to the target value

I’d argue that EL doesn’t need to have this ability as long as CL is responsible for verifying the `excess_blob_gas` computation. If we put it this way then we have no change to the EL block structure with a bit of cross-layer logic duplication: EL and CL will have to compute `excess_blob_gas`. But I think this duplication is fine as the `excess_blob_gas` formula is unlikely to change.

PR #8737, opened 2024-07-15
Update EIP-2935: Point to correct EIP-161

jochem-brouwer:
The EIP currently points to EIP-158, which was never deployed, and should point to EIP-161.


Parasos:
Praca

PR #8746, opened 2024-07-22
Update EIP-7702: Add decoding limits for AuthList items

rakita:
Fields inside the Authorization List are accessed (validated) differently than items inside the Transaction. A very big nonce will invalidate the transaction while if the nonce is big in AuthorizationList (By current spec) tx will still be valid. This makes decoding a special case.

Introducing expected limits on items invalidates this edge case.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -45,11 +45,11 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 Transaction is considered invalid if authorization list items can't be decoded as:
 
-* `chain_id` and `nonce`: u64 number.
+* `chain_id` and `nonce`: unsigned 64-bit integer.
 * `address`: 20 bytes array.
-* `y_parity`: single byte with values 0 or 1.
-* `r` and `s`: 32 bytes array.
-* and `s`-value is greater than `secp256k1n/2`, specified in [EIP-2](./eip-2.md).
+* `y_parity`: Value 0 or 1.
+* `r`: 32 bytes array.
+* `s`: 32 bytes array and value less or equal than `secp256k1n/2`, specified in [EIP-2](./eip-2.md).
 
 The fields `chain_id`, `nonce`, `max_priority_fee_per_gas`, `max_fee_per_gas`, `gas_limit`, `destination`, `value`, `data`, and `access_list` of the outer transaction follow the same semantics as [EIP-1559](./eip-1559.md).
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -45,7 +45,8 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 Transaction is considered invalid if authorization list items can't be decoded as:
 
-* `chain_id` and `nonce`: unsigned 64-bit integer.
+* `chain_id`: unsigned 256-bit integer.
+* `nonce`: unsigned 64-bit integer.
 * `address`: 20 bytes array.
 * `y_parity`: Value 0 or 1.
 * `r`: 32 bytes array.
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -44,6 +44,7 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 ```
 
 Transaction is considered invalid if authorization list items can't be decoded as:
+
 * `chain_id` and `nonce`: u64 number.
 * `address`: 20 bytes array.
 * `y_parity`: single byte.
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -8,7 +8,7 @@ status: Review
 type: Standards Track
 category: Core
 created: 2024-05-07
-requires: 2718, 2929, 2930, 3541, 3607
+requires: 2718, 2929, 2930, 3541, 3607, 2
 ---
 
 ## Abstract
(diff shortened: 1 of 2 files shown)


EIPS/eip-7702.md
```diff
@@ -43,6 +43,14 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
 authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 ```
 
+Transaction is considered invalid if authorization list items can't be decoded as:
+
+* `chain_id` and `nonce`: u64 number.
+* `address`: 20 bytes array.
+* `y_parity`: single byte with values 0 or 1.
+* `r` and `s`: 32 bytes array.
+* and `s`-value is greater than `secp256k1n/2`, specified in [EIP-2](./eip-2.md).
```
rakita, line 52:
```suggestion
* and `s`-value is greater than `secp256k1n/2`, as specified in [EIP-2](./eip-2.md).
```

EIPS/eip-7702.md
```diff
@@ -43,6 +43,14 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
 authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 ```
 
+Transaction is considered invalid if authorization list items can't be decoded as:
```
jochem-brouwer, line 46: Invalid, as in should be rejected by txpool? (If yes I think this should be clarified - and I also think this should be the case :smile: )

rakita: It's invalid as it can't be added to the block. And if invalid tx is found in the block, the block is considered invalid.

Valid/Invalid should be the correct term used to spec out this.

It is assumed that txpool rejects invalid transactions.

jochem-brouwer: Alright, yes sorry about this, I sometimes get confused about what terminology maps to what, since "invalid" could also be interpreted as the tx itself is valid but will be reverted immediately. (It would be great at some point to have a glossary EIP of the terminology we use to avoid confusion :) )

EIPS/eip-7702.md
```diff
@@ -43,6 +43,14 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
...
+
+* `chain_id` and `nonce`: unsigned 64-bit integer.
```
lightclient, line 48: `chain_id` should be a uint256

rakita: Why should it be uint256?

there are no chain IDs that are bigger than u64: https://chainid.network/

rakita: Both reth and alloy uses u64 for ChainId

EIPS/eip-7702.md
rakita: > (It would be great at some point to have a glossary EIP of the terminology we use to avoid confusion :) )

true, this would be helpful

EIPS/eip-7702.md
lightclient: It hasn't been specified as a uint64 anywhere. To be fair, an explicit bound hasn't been given afaict. But in these cases we basically always assume it is bound by uint256. Reth /alloy can definitely impl as uint64 if you prefer but I wouldn't put that in the spec.

lightclient: I don't mind limiting the chain id, but I would want that to be a separate proposal that limits the chain id across all instances in the protocol at one time.

rakita, line 48:
```suggestion
* `chain_id`: unsigned 256-bit integer.
* `nonce`: unsigned 64-bit integer.
```

rakita: It seems people responded that they want chain_id U256, so will make a change for that.

EIPS/eip-7702.md
```diff
@@ -44,6 +44,15 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
 authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 ```
 
+Transaction is considered invalid if authorization list items can't be decoded as:
+
+* `chain_id`: unsigned 256-bit integer.
+* `nonce`: unsigned 64-bit integer.
+* `address`: 20 bytes array.
+* `y_parity`: Value 0 or 1.
+* `r`: 32 bytes array.
```
lightclient, line 53: I believe this should also be an unsigned 256-bit integer for RLP purposes?

rakita: Makes sense

EIPS/eip-7702.md
```diff
@@ -44,6 +44,15 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
...
+* `s`: 32 bytes array and value less or equal than `secp256k1n/2`, specified in [EIP-2](./eip-2.md).
```
rakita, line 54:
```suggestion
* `s`: unsigned 256-bit integer and value less or equal than `secp256k1n/2`, specified in [EIP-2](./eip-2.md).
```

EIPS/eip-7702.md
rakita, line 53:
```suggestion
* `r`: unsigned 256-bit integer.
```

rakita: @lightclient it is updated

lightclient (approved):

EIPS/eip-7702.md
```diff
@@ -8,7 +8,7 @@ status: Review
 type: Standards Track
 category: Core
 created: 2024-05-07
-requires: 2718, 2929, 2930, 3541, 3607
```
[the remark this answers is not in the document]
12Thee: Trying to access on my funds

PR #8752, opened 2024-07-23
Add EIP: State conversion to Verkle Tree

jsign:
This PR creates a new EIP for the state conversion to Verkle Tree.

### DIFF
--- a/EIPS/eip-7748.md
+++ b/EIPS/eip-7748.md
@@ -1,9 +1,9 @@
 ---
-eip: PENDING
+eip: 7748
 title: State conversion to Verkle Tree
 description: Describes a state conversion procedure to migrate key-values from the Merkle Patricia Tree to the Verkle Tree.
 author: Guillaume Ballet (@gballet), Ignacio Hagopian (@jsign), Gajinder Singh (@g11tech), Ansgar Dietrichs (@adietrichs), Gottfried Herold (@GottfriedHerold), Jamie Lokier (@jlokier), Tanishq Jasoria (@tanishqjasoria), Parithosh Jayanthi (@parithosh), Gabriel Rocheleau (@gabrocheleau), Karim Taam (@matkt)
-discussions-to: https://ethereum-magicians.org/t/eip-state-conversion-to-verkle-tree/20625
+discussions-to: https://ethereum-magicians.org/t/eip-7748-state-conversion-to-verkle-tree/20625
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-7748.md
+++ b/EIPS/eip-7748.md
@@ -17,7 +17,7 @@ This EIP proposes a procedure to convert, on each block, a fixed number of key-v
 
 ## Motivation
 
-The accounts state is big enough to make inviable waiting for transactions to organically move all of them to the VKT thought the Overlay Tree. Thus, we need a strategy to convert all the state within a reasonable time. The state conversion completion allows removing the Overlay Tree abstraction introduced in [EIP-7612](./eip-7612.md) and use directly the VKT for all state access.
+The accounts state is too large to wait for transactions to organically move all of them to the VKT through the Overlay Tree. Thus, we need a strategy to convert all the state within a reasonable time. The state conversion completion allows removing the Overlay Tree abstraction introduced in [EIP-7612](./eip-7612.md) and use directly the VKT for all state access.
 
 ## Specification
 
--- a/EIPS/eip-x.md
+++ b/EIPS/eip-x.md
@@ -1,0 +1,292 @@
+---
+eip: PENDING
+title: State conversion to Verkle Tree
+description: Describes a state conversion procedure to migrate key-values from the Merkle Patricia Tree to the Verkle Tree.
+author: Guillaume Ballet (@gballet), Ignacio Hagopian (@jsign), Gajinder Singh (@g11tech), Ansgar Dietrichs (@adietrichs), Gottfried Herold (@GottfriedHerold), Jamie Lokier (@jlokier), Tanishq Jasoria (@tanishqjasoria), Parithosh Jayanthi (@parithosh), Gabriel Rocheleau (@gabrocheleau), Karim Taam (@matkt)
+discussions-to: TODO when having EIP number
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-07-23
+requires: 7612
+---
+
+## Abstract
+
+This EIP proposes a procedure to convert, on each block, a fixed number of key-values from the existing  Merkle Patricia Tree (MPT) to the Verkle Tree (VKT).
+
+## Motivation
+
+The accounts state is big enough to make inviable waiting for transactions to organically move all of them to the VKT thought the Overlay Tree. Thus, we need a strategy to convert all the state within a reasonable time. The state conversion completion allows removing the Overlay Tree abstraction introduced in [EIP-7612](./eip-7612.md) and use directly the VKT for all state access.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Constants
+
+| Parameter                    | value | Description                                                    |
+| ---------------------------- | ----- | -------------------------------------------------------------- |
+| `CONVERSION_START_TIMESTAMP` | `TBD` | Timestamp at which the conversion starts.                      |
+| `CONVERSION_STRIDE`          | `TBD` | Maximum number of _conversion units_ to be converted per block |
+
+A _conversion unit_ is:
+
+- A contract storage slot.
+- A contract code. (i.e. all the code is a single _conversion unit_)
+- An account data. (e.g. balance, nonce, code-hash)
+
+### Changes to the execution spec
+
+Include the following code in the existing `apply_body(...)` function:
+
+```python
+def apply_body(state: State, ...) -> Tuple[Uint, Root, Root, Bloom, State, Root]:
+    ...
+    # <new_code>
+    if block_time >= CONVERSION_START_TIMESTAMP and not state._conversion_finished:
+        block_state_conversion(state, CONVERSION_STRIDE)
+    # </new_code>
+    
+    for i, tx in enumerate(map(decode_transaction, transactions)):
+        ...
+    ...
+```
+
+Before executing txs, it calls `block_state_conversion(...)` (described below) which performs a state conversion step for this block.
+
+In `state.py`, add the following code:
+
+```python
+@dataclass
+class StoragePhase:
+    """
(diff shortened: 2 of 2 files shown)


EIPS/eip-x.md
```diff
@@ -0,0 +1,292 @@
+---
+eip: PENDING
```
abcoathup, line 2:
```suggestion
eip: 7748
```

EIPS/eip-x.md
```diff
@@ -0,0 +1,292 @@
...
+title: State conversion to Verkle Tree
+description: Describes a state conversion procedure to migrate key-values from the Merkle Patricia Tree to the Verkle Tree.
+author: Guillaume Ballet (@gballet), Ignacio Hagopian (@jsign), Gajinder Singh (@g11tech), Ansgar Dietrichs (@adietrichs), Gottfried Herold (@GottfriedHerold), Jamie Lokier (@jlokier), Tanishq Jasoria (@tanishqjasoria), Parithosh Jayanthi (@parithosh), Gabriel Rocheleau (@gabrocheleau), Karim Taam (@matkt)
+discussions-to: https://ethereum-magicians.org/t/eip-state-conversion-to-verkle-tree/20625
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-7748-state-conversion-to-verkle-tree/20625
```
Added EIP number to title

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+        # Account code and basic data.
+        else:
+            # Getting the code from the Overlay Tree is fine since promises returning
+            # the Account full code which would come from the MPT or a separate code database.
+            account = get_account(state, curr_account.address)
+            chunked_code = chunkify_code(account.code)
+            
+            for chunk_num in range(len(chunked_code)):
+                state_set_codechunk(state, address, chunk_num, chunked_code[chunk_num])
+                n += 1
```
g11tech, line 210: ok so its possible for n > stride  depending on code

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+                    state.conversion_curr_account.phase = CodePhase(0)
+            else:
+                # There's no storage trie for the account, move directly to
+                # migrating code (if any).
+                state.conversion_curr_account.phase = CodePhase(0)
+        # Account code and basic data.
+        else:
+            # Getting the code from the Overlay Tree is fine since promises returning
+            # the Account full code which would come from the MPT or a separate code database.
+            account = get_account(state, curr_account.address)
```
g11tech, line 205: so we always convert account and its code in one go (makes sense)

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+```
+
+## Rationale
+
+### State conversion step position in block execution
+
+Performing the conversion step before the block txs execution has some benefits:
+
+- If the state conversion step is done after txs execution, there's a possibility that txs execution writes overlap with converted key-values, having to care about them becoming stale in the same block. With the proposed ordering, they can only become stale by writes of previous blocks.
+- It can reduce the complexity of optimizations, such as frontrunning the state conversion for the next block before it arrives.
```
g11tech, line 232: very neat point

g11tech (commented):
looks good, there is a small nit which once addressed we can merge the PR

EIPS/eip-7748.md
jsign: Correct, this clarified in the `Rationale/Account code chunking done in a single step` subsection since might not be obvious why was decided this way.

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+requires: 7612
+---
+
+## Abstract
+
+This EIP proposes a procedure to convert, on each block, a fixed number of key-values from the existing  Merkle Patricia Tree (MPT) to the Verkle Tree (VKT).
+
+## Motivation
+
+The accounts state is big enough to make inviable waiting for transactions to organically move all of them to the VKT thought the Overlay Tree. Thus, we need a strategy to convert all the state within a reasonable time. The state conversion completion allows removing the Overlay Tree abstraction introduced in [EIP-7612](./eip-7612.md) and use directly the VKT for all state access.
```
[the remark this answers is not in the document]
jsign: Thanks, I've incorporated the suggestions.

> It seems to me a second motivation would be, the fact that some zombie parts of state will cause us to have to have the two tree structure forever.

Indeed, that's what I intended to highlight by saying that waiting for "organic" activity would be problematic.

g11tech (approved):
lgtm

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+| Parameter                    | value | Description                                                    |
+| ---------------------------- | ----- | -------------------------------------------------------------- |
+| `CONVERSION_START_TIMESTAMP` | `TBD` | Timestamp at which the conversion starts.                      |
+| `CONVERSION_STRIDE`          | `TBD` | Maximum number of _conversion units_ to be converted per block |
+
+A _conversion unit_ is:
+
+- A contract storage slot.
+- A contract code. (i.e. all the code is a single _conversion unit_)
+- An account data. (e.g. balance, nonce, code-hash)
```
gabrocheleau, line 37: So, would "balance" be a conversion unit, or would "balance+nonce+codehash" be a conversion unit? And how would that play out with the updated verkle leaf design where all version, balance, nonce and codesize are all package within a single slot (BASIC_DATA)? It seems to me like it would make sense to consider a "unit" whatever fits in a single "verkle slot" in this case.

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+    """
+    The account next conversion step continues converting the 
+    storage-slot with key greater or equal next_key. 
+    If there isn't such storage-slot, the account must move to
+    BasicDataPhase.
+    """
+    next_key: Bytes   
+   
+@dataclass
+class BasicDataPhase:
```
gabrocheleau, line 72: The BasicData naming is a bit confusing to me because there is also a field named BasicData in the verkle leaf. However here, if I'm following, BasicData includes not only BasicData from the verkle leaf (version,balance,nonce,codesize) but also codehash and contract code.

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+    """
+    pass
+
+@dataclass
+class CurrentConvertingAccount:
+    """
+    Contains the state conversion next step.
+    """
+    address: Address
+    phase : StoragePhase | BasicDataPhase
```
gabrocheleau, line 87: Later in the `state_convert` method there is a reference to a `CodePhase`. Should this also be added?

EIPS/eip-7748.md
```diff
@@ -0,0 +1,292 @@
...
+        if curr_account.phase is StoragePhase:
+            # Get the storage-slot from _storage_tries which is MPT data.
+            trie = state._storage_tries.get(curr_account.address)
+            
+            if trie is not None:
+                slot_num, slot_value, next_key = trie_get_next_at_key(trie, curr_account.phase.next_key)
+                # The Overlay Tree will write in the VKT. We use the new
+                # only_if_empty parameter to avoid writing stale values.
+                set_storage(state, curr_account.address, slot_num, slot_value, only_if_empty=True)
+                n += 1
```
gabrocheleau, line 188: Are clients able to disambiguate between a storage slot that has been (purposefully) set back to `0`, and a storage slot that has never been touched? Trying to picture an edge case where confusion would arise with regards to stale values, let's say: 
- Conversion starts at time `t`
- Slot `s` previously was assigned the value `31` in the MPT.
- Sometime after the conversion has begun, slot `s` is reset to `0`. The `0` value is then written to the verkle trie.
- At a later block, conversion reaches storage slot `s`. How can a client detect that the `0` value is not stale (and therefore should not be replaced by the `31` value, which is outdated)?

EIPS/eip-7748.md
jsign: A conversion unit is "An account data" which includes all the fields (balance, nonce, code-hash). The rationale for this is that on the MPT side all these values are packed in a single tree key.

The complete space of options also have other variants. As you mention can be thought as "what is a unit" on the VKT side. Or maybe think of each field as a unit. And maybe others.

The option of "what is a unit on the MPT side" is the one chosen here. I see those other options as making the logic more complex, since now it means there're more places where you can partially migrate an account. And I don't see the benefit.

EIPS/eip-7748.md
jsign: Yes, you're right. Probably `BasicData` is falling short if we try to keep the semantics clean and only have a single meaning (i.e: the VKT one).

Probably it will be better to choose another name. Open to suggestions!

EIPS/eip-7748.md
jsign: Oh sorry, this is a mistake. On a previous version, I was moving code-chunks as conversion units. But later we simplified to move all the code in one go.

I'll open a PR to fix this soon (and we can decide on the renaming above in the PR too)

Thanks!

EIPS/eip-7748.md
jsign: From the EIP perspective, the `only_if_empty` is the way to signal that the write should only happen if it wasn't written by some tx. 

From a real impl perspective, it up to the client to figure this out. One option is to do a `Get` in the Verkle Tree and see if is empty or has a value. Since VKT can distinguish between empty and zero, then that's an option.  Accessing the tree can be slow so other faster options can be considered, such as some tricks using flat-dbs to write there if something was written after the conversion started, so you can assume exists in VKT and the MPT value is stale.

EIPS/eip-7748.md
jsign: Let's continue in https://github.com/ethereum/EIPs/pull/8780!

EIPS/eip-7748.md
gabrocheleau: Yeah that makes sense. Definitely, I think we should go for whatever has the smallest complexity burden, and partial account migration is a concern that I had in mind. If we imagine an edge case where an account gets partially migrated at block `n`:
- balance and nonce get migrated, and then we stop the account conversion short of converting `code_size`, having reached "n = CONVERSION_TRIDE` 

Then we have a situation where the "BASIC_DATA" of the verkle tree has been partially updated. We therefore have to keep track of the fact that some BASIC_DATA fields have been converted (e.g. version, balance, nonce), while code_size has not. Clients will therefore have to make that discernment during the transition, and won't be able to rely on e.g. the simple fact that `BASIC_DATA.version !== undefined`. What do you think?

jsign: That's exactly the reason why I prefered the current `AccountDataPhase`. All the fields are moved in one go and there's no partial migration for all those fields.

Currently, the `***Phase` describe all the situations where you can have a partially migrated account. Today we only have `StoragePhase` and `AccountDataPhase` so to push as much as possible to reduce these "partially migrated" situations.

So I believe we're both saying the same? I didn't quite get if your last question implies you're proposing something different than today.

gabrocheleau: Ah it sounds like we agree then! Good

PR #8753, opened 2024-07-23
Update EIP-6800: Clarify BASIC_DATA layout and update naming

jsign:
This PR proposes:
- Updates the name of the code hash field to be aligned with the latest EIP-4762.
- Clarify that packing in `BASIC_DATA` is done using big-endian. This is required for one of the reserved bytes to extend `code_size` to 4 bytes in the future.

PR #8754, opened 2024-07-23
EIP-7612: Update func implementations and add further clarifications

jsign:
This PR proposes:
- Updating the spec to align with the latest EIP-6800 changes regarding multiple account fields being packed in `BASIC_DATA`.
- Add a missing new function to save code-chunks in the state. This is required not only for contract creations, but also for [the EIP](https://github.com/ethereum/EIPs/pull/8752) that explains how to do the MPT->VKT conversion.
- Explain block header changes.

### DIFF
--- a/EIPS/eip-7612.md
+++ b/EIPS/eip-7612.md
@@ -134,6 +134,10 @@ def set_storage(
     state._overlay_tree.set(get_tree_key_for_storage_slot(addr, slot), value)
 ```
 
+### Changes to the block header
+
+At `FORK_TIME` the block header state root is changed from the MPT root to the VKT root.
+
 ## Rationale
 
 This approach doesn't convert the state, which is left to a subsequent EIP. This is meant as a stopgap in case we decide to push the conversion itself to a later time. It has the advantage of simplicity, which means that the Verge fork could happen at the same time as other, simpler EIPs. It also requires no change at the consensus layer.
--- a/EIPS/eip-7612.md
+++ b/EIPS/eip-7612.md
@@ -13,15 +13,15 @@ requires: 4762, 6800, 7545
 
 ## Abstract
 
-This EIP proposes a method to switch the state tree tree format from hexary Merkle Patricia Tree (MPT) to a verkle tree: the MPT tree is frozen, and new writes to the state are stored in a verkle tree “laid over” the hexary MPT. The historical MPT state is left untouched and its eventual migration is handled at a later time.
+This EIP proposes a method to switch the state tree tree format from hexary Merkle Patricia Tree (MPT) to a Verkle Tree (VKT): the MPT tree is frozen, and new writes to the state are stored in a VKT “laid over” the hexary MPT. The historical MPT state is left untouched and its eventual migration is handled at a later time.
 
 ## Motivation
 
-The Ethereum state is growing, and verkle trees offer a good mitigation strategy to stem this growth and enable weak statelessness. Owing to the difficulty of translating contracts with large storage while they are being accessed, proposals for migrating the current MPT state are complex and will require client teams to undergo a long process of refactoring their code to handle this conversion.
+The Ethereum state is growing, and VKTs offer a good mitigation strategy to stem this growth and enable weak statelessness. Owing to the difficulty of translating contracts with large storage while they are being accessed, proposals for migrating the current MPT state are complex and will require client teams to undergo a long process of refactoring their code to handle this conversion.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7612.md
```diff
@@ -13,15 +13,15 @@ requires: 4762, 6800, 7545
 
 ## Abstract
 
-This EIP proposes a method to switch the state tree tree format from hexary Merkle Patricia Tree (MPT) to a verkle tree: the MPT tree is frozen, and new writes to the state are stored in a verkle tree “laid over” the hexary MPT. The historical MPT state is left untouched and its eventual migration is handled at a later time.
+This EIP proposes a method to switch the state tree tree format from hexary Merkle Patricia Tree (MPT) to a Verkle Tree (VKT): the MPT tree is frozen, and new writes to the state are stored in a VKT “laid over” the hexary MPT. The historical MPT state is left untouched and its eventual migration is handled at a later time.
```
jsign, line 16: For "Verkle Tree" do the same as we do for "Merkle Patricia Tree".

EIPS/eip-7612.md
```diff
@@ -31,9 +31,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### Constants
 
-|Parameter|value|Description|
-|-|-|-|
-|`FORK_TIME`|`TDB`|Time at which the verkle, overlay tree is activated.|
```
jsign, line 36: I think TDB was a typo.

EIPS/eip-7612.md
```diff
@@ -45,43 +45,35 @@ def is_fork_block(block):
...
-        tree.set(noncekey, account.nonce)
+        basicdata = bytes(0) # Version
+        basicdata += bytes(4) # Reserved
+        basicdata += len(account.code).to_bytes(3, 'big')
+        basicdata += account.nonce.to_bytes(8, 'big')
+        basicdata += account.balance.to_bytes(16, 'big')
+        tree.set(key, basicdata)
         ckkey = key
-        ckkey[31] = CODE_KECCAK_LEAF_KEY
+        ckkey[31] = CODEHASH_LEAF_KEY
```
jsign, line 55: Align naming to EIP-4762/EIP-6800

EIPS/eip-7612.md
```diff
@@ -45,43 +45,35 @@ def is_fork_block(block):
...
-        balancekey[31] = BALANCE_LEAF_KEY
-        tree.set(balancekey, account.balance)
-        noncekey = key
-        noncekey[31] = NONCE_LEAF_KEY
-        tree.set(noncekey, account.nonce)
+        basicdata = bytes(0) # Version
+        basicdata += bytes(4) # Reserved
+        basicdata += len(account.code).to_bytes(3, 'big')
+        basicdata += account.nonce.to_bytes(8, 'big')
+        basicdata += account.balance.to_bytes(16, 'big')
```
jsign, line 52: Update to do `BASIC_DATA` packing. This code assumes [this PR](https://github.com/ethereum/EIPs/pull/8753) will be accepted where we switch to big-endian encoding.

EIPS/eip-7612.md
```diff
@@ -142,6 +134,17 @@ def set_storage(
     state._overlay_tree.set(get_tree_key_for_storage_slot(addr, slot), value)
 ```
 
+Add the following function which is used when storing a contract in the tree:
+
+```python3
+def state_set_codechunk(state: State, addr: Address, chunk_num: int, chunk: Bytes):
+    state._overlay_tree.set(get_tree_key_for_code_chunk(addr, chunk_num), chunk)
+```
```
gballet, line 142: this isn't used anywhere, so imo we should remove it unless it is used

jsign: It's not referenced in this EIP, but it's referenced in another one which has this one as a dependency, [see here](https://github.com/ethereum/EIPs/pull/8752/files#diff-0476f5dcc1f4d253ea73e86a58163a7e194a1bfdb5464fe76e2f1cf97a3afc75R209).

gballet: I would then put it the referring eip, but I will let the allmighty editors decide on this one. @g11tech ?

jsign: Actually this code should be part of 6800, but that EIP doesn't have this "Python-style". But yep, let's see what @g11tech says and I can remove/move/add as needed.

PR #8755, opened 2024-07-23
Update EIP-7702: fix delegation designator prefix in Abstract

gumb0:


EIPS/eip-7702.md
```diff
@@ -13,7 +13,7 @@ requires: 2718, 2929, 2930, 3541, 3607
 
 ## Abstract
 
-Add a new transaction type that adds a list of `[chain_id, address, nonce, y_parity, r, s]` authorization tuples. For each tuple, write a delegation designator `(0xef0000 ++ address)` to the signing account's code. All code reading operations must load the code pointed to by the designator.
```
gumb0, line 16: I believe it is a typo, as Specification has `0xef0100`

jochem-brouwer: Yes, it should also not be `0xef0000`, since by EOF terminology this means "EOF v0", which is described in the EOF EIPs as "legacy contracts" (the EOF v0), so it should definitely be something else (`0xef0100 ++ address`) would be fine, can also ditch the `00` I think?

gumb0: trailing `00` is for versioning I believe

jochem-brouwer: Oh right, that's fine and actually a great idea.

lightclient (approved):

PR #8762, opened 2024-07-26
Update EIP-7702: Proxy the storage of a delegation to its unique deleterminstic keys

g11tech:
Solve the problem of storage conflicting across delegations

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -121,7 +121,7 @@ An alternative to adding chain ID could be to sign over the code the address poi
 
 Considering that the delegation from one contract to another contract may cause conflicts and corrupt the storage with a broken contract at the least and a compromised contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
 
-To this effect, the delegated EOA's storage should be proixed and accessed/written at `keccak256(authority||address)[0:20]`. This way, if the contract is delegated back to a previous `address`, EVM will be able to _re-attach_ its old storage.
+To this effect, the delegated EOA's storage should be proxied and accessed/written at `keccak256(authority||address)[0:20]`. This way, if the contract is delegated back to a previous `address`, EVM will be able to _re-attach_ its old storage.
 
 #### In-protocol revocation
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -127,7 +127,7 @@ With this method:
 
 * if EOA is delegated to a new address, it (essentially) starts with clean storage
 * if the contract is delegated back to a previous `address`, the EVM will be able to _re-attach_ its old storage.
-* chunks of `256` consecutive keys are mapped again to consecutive keys which is nice for debuggings and more importantly state trie optimizations (a la verkle)
+* chunks of `256` consecutive keys are mapped again to consecutive keys which is nice for debuggings and more importantly state trie optimizations (à la verkle)
 
 #### In-protocol revocation
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -121,7 +121,7 @@ An alternative to adding chain ID could be to sign over the code the address poi
 
 Considering that the delegation from one contract to another contract may cause conflicts and corrupt the storage with a broken contract at the least and a compromised contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
 
-To this effect, the delegated EOA's storage should be proxied and accessed/written at `keccak256(authority||address)[0:20]`. This way, if the contract is delegated back to a previous `address`, EVM will be able to _re-attach_ its old storage.
+To this effect, the delegated EOA's storage should be proxied and accessed or written at `keccak256(authority||address)[0:20]`. With this method, if the contract is delegated back to a previous `address`, the EVM will be able to _re-attach_ its old storage.
 
 #### In-protocol revocation
 
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -121,12 +121,13 @@ An alternative to adding chain ID could be to sign over the code the address poi
 
 Considering that the delegation from one contract to another contract may cause conflicts and corrupt the storage with a broken contract at the least and a compromised contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
 
-To this effect, the delegated EOA's storage keys should be proxied (accessed or written) at `keccak256(key||address)`. So essentially we use this key modification for `SLOAD` and `SSTORE` operations for the delegated EOA's orignal key but still writing into EOA storage.
+To this effect, the delegated EOA's storage keys should be proxied (accessed or written) at `keccak256(key||address)[0:30] || key[30:32]`. So essentially we use this key modification for `SLOAD` and `SSTORE` operations for the delegated EOA's orignal key but still writing into EOA storage.
 
 With this method:
 
 * if EOA is delegated to a new address, it (essentially) starts with clean storage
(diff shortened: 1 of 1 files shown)


EIPS/eip-7702.md
```diff
@@ -117,6 +117,12 @@ One consideration when signing a code pointer is what code might that address po
 
 An alternative to adding chain ID could be to sign over the code the address points to. This seems to have the benefit of both minimizing the on-chain size of auth tuples while retaining specificity of the actual code running in the account. One unfortunate issue of this format though is that it imposes a database lookup to determine the signer of each auth tuple. This imposition itself seems to create enough complexity in transaction propagation that it is decided to avoid and simply sign over address directly.
 
+#### Delegation Storage
+
+Considering that the delegation from one contract to another contract can cause conflict and corruption of the storage with broken contract in the least and a compromized contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
```
philknows, line 122:
```suggestion
Considering that the delegation from one contract to another contract may cause conflicts and corrupt the storage with a broken contract at the least and a compromised contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
```

EIPS/eip-7702.md
```diff
@@ -117,6 +117,12 @@ One consideration when signing a code pointer is what code might that address po
...
+
+To this effect, the delegated EOA's storage should be proixed and accessed/written at `keccak256(authority||address)[0:20]`. This way, if the contract is delegated back to a previous `address`, EVM will be able to _re-attach_ its old storage.
```
philknows, line 124:
```suggestion
To this effect, the delegated EOA's storage should be proxied and accessed or written at `keccak256(authority||address)[0:20]`. With this method, if the contract is delegated back to a previous `address`, the EVM will be able to _re-attach_ its old storage.
```

EIPS/eip-7702.md
```diff
@@ -117,6 +117,12 @@ One consideration when signing a code pointer is what code might that address po
...
+
 #### In-protocol revocation
 
 Unlike previous versions of this EIP and EIPs similar, the delegation designation can be revoked at anytime signing and sending a EIP-7702 authorization to a new target with the account's current nonce. Without such action, a delegation will remain valid in perpetuity.
```
philknows, line 128:
```suggestion
Unlike previous versions of this EIP, the delegation designation can be revoked at anytime by signing and sending an EIP-7702 authorization to a new target with the account's current nonce. Without such action, a delegation will remain valid in perpetuity.
```

EIPS/eip-7702.md
```diff
@@ -117,9 +117,22 @@ One consideration when signing a code pointer is what code might that address po
...
+Considering that the delegation from one contract to another contract may cause conflicts and corrupt the storage with a broken contract at the least and a compromised contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
+
+To this effect, the delegated EOA's storage keys should be proxied (accessed or written) at `keccak256(key||address)`. So essentially we use this key modification for `SLOAD` and `SSTORE` operations for the delegated EOA's orignal key but still writing into EOA storage.
+
+With this method:
+
+* if EOA is delegated to a new address, it (essentially) starts with clean storage
+* if the contract is delegated back to a previous `address`, the EVM will be able to _re-attach_ its old storage.
+
+Again for the purpose of [EIP-158](../eip-158.md) cleanup, this storage should be considered attached to the EOA address.
```
jochem-brouwer, line 131: This line can be removed

jochem-brouwer (approved):
LGTM!

EIPS/eip-7702.md
```diff
@@ -121,12 +121,13 @@ An alternative to adding chain ID could be to sign over the code the address poi
 
 Considering that the delegation from one contract to another contract may cause conflicts and corrupt the storage with a broken contract at the least and a compromised contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
 
-To this effect, the delegated EOA's storage keys should be proxied (accessed or written) at `keccak256(key||address)`. So essentially we use this key modification for `SLOAD` and `SSTORE` operations for the delegated EOA's orignal key but still writing into EOA storage.
+To this effect, the delegated EOA's storage keys should be proxied (accessed or written) at `keccak256(key||address)[0:30] || key[30:32]`. So essentially we use this key modification for `SLOAD` and `SSTORE` operations for the delegated EOA's orignal key but still writing into EOA storage.
```
jochem-brouwer, line 124: Off by one error? 
```suggestion
To this effect, the delegated EOA's storage keys should be proxied (accessed or written) at `keccak256(key||address)[0:30] || key[31:31]`. So essentially we use this key modification for `SLOAD` and `SSTORE` operations for the delegated EOA's orignal key but still writing into EOA storage.
```

Assuming that `[x:y]` notation is inclusive of `x` and `y`

g11tech: ahh thank you, fixed

jochem-brouwer (approved):
LGTM!

MustardseedX (commented):
is there no way to fix the html and dns ?  everything else seems to be good im just curious if it will cause depression in other branches that may be bigger than what is specified its simple a concern i thought should be addressed or voiced  that being said i currently am unfortunately without a workaround or fix at this moment ...

X

lightclient:
I don't support this change. It creates division between 7702 accounts and full smart contract wallets. I contend that this problem should be solved at the app layer. Even if people disagree with that, a protocol solution should absolutely solve the problem for both full smart contract wallets and 7702 accounts.

EIPS/eip-7702.md
```diff
@@ -117,9 +117,21 @@ One consideration when signing a code pointer is what code might that address po
...
+
+Considering that the delegation from one contract to another contract may cause conflicts and corrupt the storage with a broken contract at the least and a compromised contract at the worst, it is prudent to separate the storage of each delegation in deterministic way.
+
+To this effect, the delegated EOA's storage keys should be proxied (accessed or written) at `keccak256(key[0:30]||address)[0:30] || key[31:31]`. So essentially we use this key modification for `SLOAD` and `SSTORE` operations for the delegated EOA's orignal key but still writing into EOA storage.
+
+With this method:
+
+* if EOA is delegated to a new address, it (essentially) starts with clean storage
+* if the contract is delegated back to a previous `address`, the EVM will be able to _re-attach_ its old storage.
+* chunks of `256` consecutive keys are mapped again to consecutive keys which is nice for debuggings and more importantly state trie optimizations (a la verkle)
```
gballet, line 130:
```suggestion
* chunks of `256` consecutive keys are mapped again to consecutive keys which is nice for debuggings and more importantly state trie optimizations (à la verkle)
```

g11tech:
> I don't support this change. It creates division between 7702 accounts and full smart contract wallets. I contend that this problem should be solved at the app layer. Even if people disagree with that, a protocol solution should absolutely solve the problem for both full smart contract wallets and 7702 accounts.

imo we shouldn't deliver a bad UX when it is simple enough to solve the problem. compromising one featureset for parity with other sounds like a bad idea

either this or the storage of 7702 should be ephemeral and not banning create so that contracts can create their own storage islands to store their data

thogard785:
When the ability to upgrade an EOA to a real SCW finally does happen, it’s not feasible to assume that every user and / or every wallet will go through an EOA’s 7702 history and modify the SCW’s code to avoid storage conflicts. 

One bandaid now is preferable to twenty bandaids later, especially since we don’t (and can’t) know the full extent of how 7702 will be used.

g11tech:
closing PR as suggestion was not accepted

PR #8763, opened 2024-07-26
Update EIP-7702: Add `tx_origin` field to authority

jochem-brouwer:


EIPS/eip-7702.md
```diff
@@ -121,6 +122,10 @@ An alternative to adding chain ID could be to sign over the code the address poi
 
 Unlike previous versions of this EIP and EIPs similar, the delegation designation can be revoked at anytime signing and sending a EIP-7702 authorization to a new target with the account's current nonce. Without such action, a delegation will remain valid in perpetuity.
 
+### Explicit `tx.origin` signing
+
+A signer might desire that only a specific EOA account sends the EIP-7702 transaction. If this is the case, the signer should set the `tx_origin` field of the authority list to that address. If any EOA account is allowed to send the authority transaction, the `tx_origin` field should be set to `null` (empty bytes). Note that without `tx.origin`, once a EIP-7702 transaction enters the transaction pool, this delegation authority is now immediately available to all accounts, and could thus lead to undesired or unwanted results such as being frontran by MEV.
```
g11tech, line 127:
```suggestion
A signer might desire that only a specific EOA account sends the EIP-7702 transaction. If this is the case, the signer should set the `tx_origin` field of the authority list to that address. If any EOA account is allowed to send the authority transaction, the `tx_origin` field should be set to `null` (empty bytes). Note that without `tx.origin`, once a EIP-7702 transaction enters the transaction pool, this delegation authority is now immediately available to all accounts, and could thus lead to undesired or unwanted results such as being frontrun by MEV.
```

jochem-brouwer: Thanks!

g11tech (approved):
+1 for this small security feature

EIPS/eip-7702.md
```diff
@@ -121,6 +122,10 @@ An alternative to adding chain ID could be to sign over the code the address poi
 
 Unlike previous versions of this EIP and EIPs similar, the delegation designation can be revoked at anytime signing and sending a EIP-7702 authorization to a new target with the account's current nonce. Without such action, a delegation will remain valid in perpetuity.
 
+### Explicit `tx.origin` signing
+
+A signer might desire that only a specific EOA account sends the EIP-7702 transaction. If this is the case, the signer should set the `tx_origin` field of the authority list to that address. If any EOA account is allowed to send the authority transaction, the `tx_origin` field should be set to `null` (empty bytes). Note that without `tx.origin`, once a EIP-7702 transaction enters the transaction pool, this delegation authority is now immediately available to all accounts, and could thus lead to undesired or unwanted results such as being frontrun by MEV.
```
lightclient, line 127: This either requires a trusted party to relay or self-relay. I don't support protocol features that are only useful if you trust a third party. If you need to self-relay, just gate initialization of your contract with `ORIGIN`.

g11tech: if i understand your comment correctly, this is not a security feature for the wallet provider (their gated contract), but for the user doing the delegation.

The basic security principal is keep the attack surface minimal, and this is what this does. it only allows trusted wallet provider/party to use the authorization to prevent timing/reuse attacks (of a tx that couldn't make to block but is in mempool or failed)

This anyway doesn't hinder someone to sign a public authorization by allowing origin to be null.

PR #8769, opened 2024-07-29
Update EIP-6789: Move to Review

pcaversaccio:
`mana` is now an official alias for `gas` in Vyper: https://github.com/vyperlang/vyper/pull/3713. Thus, this EIP is obviously not anymore stagnant but in the "Review" phase.


SamWilsn (approved):
Need to move back into Draft before it can go into Review or it messes up the statistics.

PR #8771, opened 2024-07-29
Update EIP-5792: multi-chain wallet_sendCalls

pedrouid:
Changes proposed:
- move `chainId` from top-level to call-level
- add comment on `atomicBatch` regarding multi-chain

### DIFF
--- a/EIPS/eip-5792.md
+++ b/EIPS/eip-5792.md
@@ -62,7 +62,6 @@ type SendCallsParams = {
     value?: `0x${string}` | undefined; // Hex value
     chainId?: `0x${string}` | undefined; // Hex chain id
   }[];
-  parallel?: boolean;
   capabilities?: Record<string, any> | undefined;
 };
 
--- a/EIPS/eip-5792.md
+++ b/EIPS/eip-5792.md
@@ -55,12 +55,13 @@ The wallet:
 ```typescript
 type SendCallsParams = {
   version: string;
-  chainId: `0x${string}`; // Hex chain id
+  
   from: `0x${string}`;
   calls: {
     to?: `0x${string}` | undefined;
     data?: `0x${string}` | undefined;
     value?: `0x${string}` | undefined; // Hex value
+    chainId?: `0x${string}` | undefined; // Hex chain id
   }[];
   capabilities?: Record<string, any> | undefined;
 };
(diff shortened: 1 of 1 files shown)


EIPS/eip-5792.md
```diff
@@ -40,13 +40,13 @@ The capabilities field is how an app can communicate with a wallet about capabil
 
 The wallet:
 
-* MUST send these calls in the order specified in the request
-* MUST send the calls on the same chain identified by the request's `chainId` property
+* MUST send the calls in the order specified in the request
+* MUST send the calls on the same chain identified by the call's `chainId` property
+* MUST NOT await for any calls to be finalized to complete the batch
```
lukasrosario, line 45: can you explain why you think we should enforce this? I'd expect this to be up to the wallet, until we have the other capability we talked about which would enforce waiting for calls to be mined across chains.

EIPS/eip-5792.md
```diff
@@ -55,13 +55,14 @@ The wallet:
...
   version: string;
-  chainId: `0x${string}`; // Hex chain id
   from: `0x${string}`;
   calls: {
     to?: `0x${string}` | undefined;
     data?: `0x${string}` | undefined;
     value?: `0x${string}` | undefined; // Hex value
+    chainId?: `0x${string}` | undefined; // Hex chain id
   }[];
+  parallel?: boolean;
```
lukasrosario, line 65: seems like leftover? i think we should remove this. lmk if we're not aligned here though

EIPS/eip-5792.md
```diff
@@ -246,6 +250,8 @@ The capabilities below are for illustrative purposes.
 
 Like the illustrative examples given above and other capabilities to be defined in future EIPs, the capability to execute calls delivered via the above-defined methods in a single transaction can be attested by the wallet in boolean form.
 
+This capability is expressed separately on each chain and should be interpreted as a guarantee only for batches of transactions on that chain; batches including calls to multiple chains is out of scope of this capability and this specification.
```
lukasrosario, line 253: 👍

EIPS/eip-5792.md
pedrouid: My understanding is that we would enforce the behavior to always be "parallel" by default 

So the new capability would be "sequential" if available

But my take is that either way it should be deterministic by the spec rather than wallet-dependent

lukasrosario (approved):

EIPS/eip-5792.md
wilsoncusack: Hmm just seeing this. I do feel like generally order matters to devs, but trying to sequence multi chain calls is probably too difficult. Too hard to know when the output of one will land on the other chain, if I am, say, funding my account on chain A from chain B

PR #8780, opened 2024-08-02
Update EIP-7748: fix namings

jsign:
This PR fixes:
- A phase naming which was referring to a phase that doesn't exist (in the final proposed version of the EIP).
- A rename of the `BasicDataPhase`, since as mentioned by @gabrocheleau it might create some confusion since the semantics of `BasicData` are different in VKT.

### DIFF
--- a/EIPS/eip-7748.md
+++ b/EIPS/eip-7748.md
@@ -64,12 +64,12 @@ class StoragePhase:
     The account next conversion step continues converting the 
     storage-slot with key greater or equal next_key. 
     If there isn't such storage-slot, the account must move to
-    BasicDataPhase.
+    AccountDataPhase.
     """
     next_key: Bytes   
    
 @dataclass
-class BasicDataPhase:
+class AccountDataPhase:
     """
     The account next conversion step continues migrating the account 
     code (if any) and basic data. After processing, the account must 
@@ -84,7 +84,7 @@ class CurrentConvertingAccount:
     Contains the state conversion next step.
     """
     address: Address
-    phase : StoragePhase | BasicDataPhase
+    phase : StoragePhase | AccountDataPhase
 ```
 
 These new structures allows `State` to track where we're in the conversion process.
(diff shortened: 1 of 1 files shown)


EIPS/eip-7748.md
```diff
@@ -193,11 +193,11 @@ def state_convert(state: State, stride: int):
                 else:
                     # No more storage-slots. Move to the code migration starting
                     # at chunk-number zero.
-                    state.conversion_curr_account.phase = CodePhase(0)
+                    state.conversion_curr_account.phase = BasicDataPhase()
```
jsign, line 196: We've pending in this PR to decide a better name for `BasicDataPhase`.

gabrocheleau: The best name I can think of is `AccountDataPhase`. It is not perfect as an account can be interpreted as including storage, but in verkle they are conceptually more separate than in the MPT since storage isn't nested, and is (mostly) distributed throughout the tree.

EIPS/eip-7748.md
```diff
@@ -64,12 +64,12 @@ class StoragePhase:
...
     storage-slot with key greater or equal next_key. 
     If there isn't such storage-slot, the account must move to
-    BasicDataPhase.
+    AccountDataPhase.
     """
     next_key: Bytes   
    
 @dataclass
-class BasicDataPhase:
+class AccountDataPhase:
```
jsign, line 72: With @gabrocheleau, we decided to rename it `AccountDataPhase` since `BasicDataPhase` is a bit off, considering that in VKT, it usually leaves out the code-hash (which, in case of the migration, is also moved).

We'd like at least one extra approval for this renaming (or new suggestion!) from another author.

cc @gballet @g11tech @tanishqjasoria

(Note: avoiding switching from Draft to avoid automerge)

EIPS/eip-7748.md
```diff
@@ -193,11 +193,11 @@ def state_convert(state: State, stride: int):
                 else:
                     # No more storage-slots. Move to the code migration starting
                     # at chunk-number zero.
```
gabrocheleau, line 195: Given the CodePhase -> AccountDataPhase update, is this comment still accurate? Do we still start converting accounts at code chunk number 0

jsign: Nope, was a left over. We move to "Account data" migration (i.e: all code, and account fields).

gabrocheleau (approved):

gballet (approved):
LGTM

jsign:
> LGTM. I actually wonder if `AccountPhase` wouldn't be better, since "account data" could be interpreted as "account storage".

Thing is that `AccountPhase` mentions `Account`. In case of smart-contract accounts, `Account` means "everything".

tanishqjasoria (approved):

PR #8790, opened 2024-08-07
Update EIP-7702: Switch order of authority warm access

rakita:
Small change, it inserts step 8. `Add authority to accessed_addresses (as defined in [EIP-2929](./eip-2929.md).)` . to become step 3. (Before nonce and bytecode checks)

This is a better order as we load the account first and add authority to the access list before checking if the nonce is equal or if acc is empty/delegated.


lightclient (approved):
LGTM

PR #8791, opened 2024-08-07
Update EIP-5792: multi-chain wallet_sendCalls"

Mortiemi:
Reverts ethereum/EIPs#8771


lukasrosario:
can you share more on why you want to revert this change?

PR #8816, opened 2024-08-20
Update EIP-2935: update mechanism via system call

s1na:
Based on the discussion here: https://github.com/ethereum/go-ethereum/pull/29465#issuecomment-2296819860. I took the liberty to expand on block processing and the set/get functions.

### DIFF
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -238,7 +238,7 @@ The first option is recommended until the Verkle fork, to stay consistent with [
 
 This EIP introduces backwards incompatible changes to the block validation rule set. But neither of these changes break anything related to current user activity and experience.
 
-## Test Case
+## Test Cases
 
 TBD
 
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -34,15 +34,9 @@ A side benefit of this approach could be that it allows building/validating proo
 
 This EIP specifies for storing last `HISTORY_SERVE_WINDOW` block hashes in a ring buffer storage of `HISTORY_SERVE_WINDOW` length. Note that `HISTORY_SERVE_WINDOW` > `BLOCKHASH_SERVE_WINDOW` (which remains unchanged).
 
-At the start of processing any block where `block.timestamp >= FORK_TIMESTAMP` (ie. before processing any transactions), update the state directly in the following way:
+At the start of processing any block where `block.timestamp >= FORK_TIMESTAMP` (ie. before processing any transactions), perform a system call to the contract `set` mechanism defined in the following section. This system call will follow the same convention as in [EIP-4788](./eip-4788.md). The calldata will equal the 32-byte input of `block.parent.hash`.
 
-```python
-def process_block_hash_history(block: Block, state: State):
-    if block.timestamp >= FORK_TIMESTAMP:
-        state.insert_slot(HISTORY_STORAGE_ADDRESS, (block.number-1) % HISTORY_SERVE_WINDOW , block.parent.hash)
-```
-
-Alternatively clients can also choose to do a system update via a system call to the contract `set` mechanism defined in the following sections.
+Alternatively clients can choose to do directly write to the storage of the contract. Note: this could prove problematic on non-mainnet networks e.g. if the contract is not deployed at activation of this EIP.
 
 Note that, it will take `HISTORY_SERVE_WINDOW` blocks after `FORK_TIMESTAMP` to completely fill up the ring buffer. The contract will only contain the parent hash of the fork block and no hashes prior to that.
 
--- a/EIPS/eip-2935.md
+++ b/EIPS/eip-2935.md
@@ -43,14 +43,14 @@ At the start of processing any block where `block.timestamp >= FORK_TIMESTAMP` (
 * the call does not follow the [EIP-1559](./eip-1559.md) burn semantics - no value should be transferred as part of the call
 * if no code exists at `HISTORY_STORAGE_ADDRESS`, the call must fail silently
 
-Note: Alternatively clients can choose to do directly write to the storage of the contract. However this could prove problematic on non-mainnet networks e.g. if the contract is not deployed at activation of this EIP.
+Note: Alternatively clients can choose to directly write to the storage of the contract but EVM calling the contract remains preferred. Refer to the rationale for more info.
 
 Note that, it will take `HISTORY_SERVE_WINDOW` blocks after `FORK_TIMESTAMP` to completely fill up the ring buffer. The contract will only contain the parent hash of the fork block and no hashes prior to that.
 
 
 ### EVM Changes
 
-The `BLOCKHASH` opcode semantics remains the same as before. However the clients which want to leverage the history from state may do so making sure the query is within `BLOCKHASH_SERVE_WINDOW` and is available in the contract.
+The `BLOCKHASH` opcode semantics remains the same as before.
 
 ### Block hash history contract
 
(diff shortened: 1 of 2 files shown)


lightclient (dismissed):

lightclient (dismissed):

EIPS/eip-2935.md
```diff
@@ -200,6 +217,23 @@ Second concern was how to best transition the BLOCKHASH resolution logic post fo
 
 We choose to go with the former. It simplifies the logic greatly. It will take roughly a day to bootstrap the contract. Given that this is a new way of accessing history and no contract depends on it, it is deemed a favorable tradeoff.
 
+### Inserting the parent block hash
+
+Clients have generally two options for inserting the parent block hash into state:
+
+1. Performing a sytstem call to `HISTORY_STORAGE_ADDRESS` and letting that handle the storing in state.
```
gabrocheleau, line 224:
```suggestion
1. Performing a system call to `HISTORY_STORAGE_ADDRESS` and letting that handle the storing in state.
```

EIPS/eip-2935.md
```diff
@@ -200,6 +217,23 @@ Second concern was how to best transition the BLOCKHASH resolution logic post fo
...
+
+The latter option is as follows:
+
+```python
+def process_block_hash_history(block: Block, state: State):
+    if block.timestamp >= FORK_TIMESTAMP:
+        state.insert_slot(HISTORY_STORAGE_ADDRESS, (block.number-1) % HISTORY_SERVE_WINDOW , block.parent.hash)
+```
+
+The first option is recommended until the Verkle fork, to stay consistent with [EIP-4788](./eip-4788.md) and to issues for misconfigured networks where this EIP is activated but history contract hasn't been deployed. The recommendation may be reconsidered at the Verkle fork to avoid working around attaching the history contract to every block's Verkle witness.
```
gabrocheleau, line 235:
```suggestion
The first option is recommended until the Verkle fork, to stay consistent with [EIP-4788](./eip-4788.md) and to mitigate issues for misconfigured networks where this EIP is activated but history contract hasn't been deployed. The recommendation may be reconsidered at the Verkle fork to avoid working around attaching the history contract to every block's Verkle witness.
```
Not sure if this is what is meant, seems like there's a word missing

gabrocheleau, line 235: I'm not sure I fully get this. 
"issues for misconfigured networks where this EIP is activated but history contract hasn't been deployed."
Why do we care about supporting misconfigured networks? 

"The recommendation may be reconsidered at the Verkle fork to avoid working around attaching the history contract to every block's Verkle witness."

Do you mean that at verkle, we may want to directly update the state rather than access the system contract in order to avoid having to include the system contract to every block's verkle witness?

SamWilsn:
This just needs an author approval to go through.

EIPS/eip-2935.md
s1na: > Why do we care about supporting misconfigured networks?

It's not about supporting misconfigured networks rather reducing damage in case contract is not deployed timely on a network. Calling the contract vs setting the store will behave differently if the contract is not deployed causing a network split. And this and 4788 are the only EIPs that in addition to activating the EIP someone needs to do an extra action of sending the transaction. Easy to miss or get wrong. Not using this as an argument but as an anecdote: a bug in geth's dev mode for 4788 until [recently](https://github.com/ethereum/go-ethereum/issues/29539).

> Do you mean that at verkle, we may want to directly update the state rather than access the system contract in order to avoid having to include the system contract to every block's verkle witness

This suggestion existed in the EIP (I believe put there by @gballet), I moved it to the rationale section as I believe it doesn't exist in the specification.

gabrocheleau: Makes sense concerning network split, thanks for explaining.

Concerning the last sentence, can we clarify the phrasing? Maybe it's just me, but there's a double negative in there that confused me, what does  "avoid working around attaching" mean?

s1na: I think now it is simplified.

EIPS/eip-2935.md
```diff
@@ -200,6 +217,23 @@ Second concern was how to best transition the BLOCKHASH resolution logic post fo
...
+
+The latter option is as follows:
+
+```python
+def process_block_hash_history(block: Block, state: State):
+    if block.timestamp >= FORK_TIMESTAMP:
+        state.insert_slot(HISTORY_STORAGE_ADDRESS, (block.number-1) % HISTORY_SERVE_WINDOW , block.parent.hash)
+```
+
+The first option is recommended until the Verkle fork, to stay consistent with [EIP-4788](./eip-4788.md) and to issues for misconfigured networks where this EIP is activated but history contract hasn't been deployed. The recommendation may be reconsidered at the Verkle fork if the cost of attaching the history contract in each block's Verkle witness is deemed too high.
```
gballet, line 235:
```suggestion
The first option is recommended until the Verkle fork, to stay consistent with [EIP-4788](./eip-4788.md) and to issues for misconfigured networks where this EIP is activated but history contract hasn't been deployed. The recommendation may be reconsidered at the Verkle fork if filtering the system contract code chunks is deemed too complex.
```

gballet (approved):

Alkhtri77:
How much does

EIPS/eip-2935.md
```diff
@@ -34,21 +34,42 @@ A side benefit of this approach could be that it allows building/validating proo
...
+### Block processing
 
-```python
-def process_block_hash_history(block: Block, state: State):
-    if block.timestamp >= FORK_TIMESTAMP:
-        state.insert_slot(HISTORY_STORAGE_ADDRESS, (block.number-1) % HISTORY_SERVE_WINDOW , block.parent.hash)
-```
+At the start of processing any block where `block.timestamp >= FORK_TIMESTAMP` (ie. before processing any transactions), call to `HISTORY_STORAGE_ADDRESS` as `SYSTEM_ADDRESS` with the 32-byte input of `block.parent.hash`, a gas limit of `30_000_000`, and `0` value. This will trigger the `set()` routine of the history contract. This is a system operation following the same convention as [EIP-4788](./eip-4788.md) and therefore:
+
+* the call must execute to completion
```
holiman, line 41: What does this mean? Why even specify it? Should evm-implementors use `30MGas` or some infinite supply? IMO this sentence only brings confusion

EIPS/eip-2935.md
```diff
@@ -34,21 +34,42 @@ A side benefit of this approach could be that it allows building/validating proo
...
+* the call does not count against the block's gas limit
+* the call does not follow the [EIP-1559](./eip-1559.md) burn semantics - no value should be transferred as part of the call
+* if no code exists at `HISTORY_STORAGE_ADDRESS`, the call must fail silently
```
holiman, line 44: I'd have preferred something which shows that this is not a "new thing", rather just a clarification of existing semantics. Something like

> if no code exists at `HISTORY_STORAGE_ADDRESS`, the call is a no-op, like any non-value-transferring call.

I also think it is incorrect to say that the call must **fail**. It's a no-op, it doesn't put a failure-symbol on any stack anywhere.

EIPS/eip-2935.md
s1na: The system call semantics are directly copied from EIP-4788 for consistency since they should behave in the same manner. It applies to your next comment as well. I personally think it doesn't harm as long as the limit is high enough. Also makes me feel better when there is a limit even tho the contract is very simple and has been fully vetted.

holiman: Don't get me wrong: I think the limit is _good_. That's why I think the sentence "must execute to completion" is bad, because it goes _against_ the limit. Because if it must execute to completion, that implies infinite gas. If someone places an infinite loop there, then, without this sentence here, all clients would agree to burn 30M gas and done. With the sentence "must execute to completion", it can arguably be correct to loop forever.

Jcpayvoice23 (commented):
My secretly apologies go out to everyone who has been there

PR #8826, opened 2024-08-25
Update EIP-5792: Propose changes to EIP-5792

forshtat:
List of changes proposed in this PR and their motivations:

1. Rephrase "The items in the `calls` field are only those that are shared by all transaction types" and "method only returns a subset of the fields that `eth_getTransactionReceipt` returns, excluding any fields that may differ across wallet implementations".

There is a rapidly growing list of transaction types, and the definition of the "shared by all" can become confusing.
Specifically, most transaction fields are removed in RIP-7560 transaction type which is part of the Account Abstraction roadmap.

2. Add optional 'capabilitiesData' field to individual 'calls' entities.

It seems likely that more advanced capabilities will need to provide some data on a per-call basis. One example could be an "optional flag" capability that would allow a Smart Contract Wallet to proceed with the batch in case of a revert, which would otherwise default to requiring success from all calls. Whether this particular capability is needed or not, it serves as an example of why such an option might be needed in the future.

3. Specify that an unsupported 'capability' cannot be ignored and the operation must be rejected.

I did not see it in the text anywhere, and ignoring data provided in the capabilities field is obviously extremely dangerous.

4. Add 'chainId' back to the 'SendCallsParams'.

For now, it is defined as optional, but probably shouldn't be. This has been debated lately and I don't know what the conclusion is.

5. Specify that the 'identifier' must be a 32 byte hex.

One reason is that "any string" seems to be too vague of a definition for an ERC. What about "", "ለከበሩጓደኛመርህመንገድይሆናል", or "😎🤏🕶😳", are these strings okay?
Also, in other programming languages it could be more convenient to use the same identifier type as for transactions, blocks and all other things in Ethereum. If there is a motivation for leaving it as "any string", please at least specify it in the ERC.

6. Added 'capabilitiesData' object to the 'wallet_getCallsStatus' response.

It seems very plausible that 'capabilities' will need to provide some additional information to the response. This was previously impossible, which could lead to wallets creating non-standard ways of querying that information.

7. Made the 'receipts' field requirements stricter.

These rules were probably implied anyway, but now they are explicit.

8. Rename 'status' to 'batchStatus'

There is a 'status' field in each 'receipt', which is a hex string. Having another 'status' field, that is not of the same type, appears to be confusing.

9. Specified 5 separate 'batchStatus' values

It seems like the previous two values did not cover the entire set of possible 'sendCalls' outcomes. The new status values are: PENDING, SUCCESS, PARTIAL, FAILURE and DISCARDED.

10. Added 'chainId' array to the 'wallet_getCapabilities' request

It seems like a wallet that supports a very long list of chains would be forced to return an unusually large JSON object as a response. However, as most dapps know which chains they may run on, it makes sense for them to limit the scope of the request.

11. Added SemVer 'version' parameter to capabilities.

It is not likely that all 'capabilities' will reach their final version from day one.
As there was no built-in mechanism for versioning, and having no centralized "capabilities authority", it would be pretty likely to see capabilities like 'paymasterServiceV2' and 'atomicBatch_3' proliferate, creating a lot of confusion.


EIPS/eip-5792.md
```diff
@@ -122,26 +136,41 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
+##### Status Codes for `batchStatus` field
+
+The purpose of the `batchStatus` field is to provide a short summary of the current status of the batch.
+It provides some off-chain context to the array of inner transaction `receipts`.
+
+| Name      | Description                                                                                                          |
+|-----------|----------------------------------------------------------------------------------------------------------------------|
+| PENDING   | Batch has been received by the wallet but has not completed execution on-chain                                       |
+| SUCCESS   | Batch has been included on-chain without reverts, receipts array contains info of all calls                          |
+| PARTIAL   | Batch has been included on-chain only partially, the rest is either reverted or invalid                              |
```
shahafn, line 167:
```suggestion
| PARTIAL   | Batch has been included on-chain only partially, the rest is either reverted or invalid. Receipts array contains info of all successful calls             |
```

EIPS/eip-5792.md
```diff
@@ -122,26 +136,41 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
     blockHash: `0x${string}`;
     blockNumber: `0x${string}`;
     gasUsed: `0x${string}`;
     transactionHash: `0x${string}`;
   }[];
+  capabilitiesData?: Record<string, any> | undefined;
 };
 ```
 
+##### Status Codes for `batchStatus` field
```
shahafn, line 158: Maybe we should define what a batch is first? it's the first time we use it

EIPS/eip-5792.md
```diff
@@ -35,33 +35,41 @@ We also define one capability expression for use with the fourth method, which f
...
 type SendCallsParams = {
   version: string;
   from: `0x${string}`;
+  chainId: `0x${string}` | undefined; // Hex chain id
   calls: {
     to?: `0x${string}` | undefined;
     data?: `0x${string}` | undefined;
     value?: `0x${string}` | undefined; // Hex value
-    chainId?: `0x${string}` | undefined; // Hex chain id
+    capabilitiesData?: Record<string, any> | undefined;
```
wilsoncusack, line 72: hmm, I don't love how this kind of muddles the simplicity of the calls struct. Is the complaint that capabilities makes it hard to specify info on a per call basis. Could we get a better example of when this would be needed 🤔 

> One example could be an "optional flag" capability that would allow a Smart Contract Wallet to proceed with the batch in case of a revert

This should be possible with a general capability? Or you're saying one call in particular can revert?

EIPS/eip-5792.md
```diff
@@ -122,26 +136,41 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
+  capabilitiesData?: Record<string, any> | undefined;
 };
 ```
 
+##### Status Codes for `batchStatus` field
+
+The purpose of the `batchStatus` field is to provide a short summary of the current status of the batch.
+It provides some off-chain context to the array of inner transaction `receipts`.
+
+| Name      | Description                                                                                                          |
```
wilsoncusack, line 163: I like this idea, thoughts @jxom @lukasrosario ?

EIPS/eip-5792.md
```diff
@@ -35,33 +35,41 @@ We also define one capability expression for use with the fourth method, which f
...
-* MUST send the calls on the same chain identified by the call's `chainId` property
+* MUST send the calls on the same chain identified by the request's `chainId` property
 * MUST NOT await for any calls to be finalized to complete the batch
 * MUST NOT send any calls from the request if the user rejects the request
 * MAY revert all calls if any call fails
 * MAY send all calls as part of one or more transactions, depending on wallet capability
 * SHOULD stop executing the calls if any call fails
 * MAY reject the request if the from address does not match the enabled account
 * MAY reject the request if one or more calls in the batch is expected to fail, when simulated sequentially
+* MUST reject the request if it contains a `capability` that is not supported by the wallet
```
wilsoncusack, line 59: Hmm I wonder if this is best 🤔 . E.g. why not allow specifying a paymaster which the app would like to be used but is not a necessity.

EIPS/eip-5792.md
```diff
@@ -122,26 +136,41 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
       topics: `0x${string}`[];
     }[];
     status: `0x${string}`; // Hex 1 or 0 for success or failure, respectively
-    chainId: `0x${string}`;
     blockHash: `0x${string}`;
     blockNumber: `0x${string}`;
     gasUsed: `0x${string}`;
     transactionHash: `0x${string}`;
   }[];
+  capabilitiesData?: Record<string, any> | undefined;
```
wilsoncusack, line 154: I find it kinda confusing that this is the same phrase as on the request, but would likely be different data?

EIPS/eip-5792.md
drortirosh: You don't have to implement it, if our wallet doesn't have any per-call capability. 
Also, an app doesn't have to know about it, unless it needs that capability.
The idea that if someone ever want to add per-call capability, it doesn't mess directly with the "call" structure, or with such params of another capability.

EIPS/eip-5792.md
drortirosh: How a wallet can tell if the cap (e.g. paymaster) is required by this app or optional?
The app can always query the wallet caps first, to decide how to continue.

lukasrosario: capability specs should specify whether they're required or not / how optional ones are handled. paymaster, for example, notes that the app must not assume the provided paymaster service is the one that ends up being used.

so if an app wants the capability to be required, the capability spec should include a `required` field, and that is how a wallet can tell if the capability is required by the app. i dont think we should blanket all capabilities to be required by default.

EIPS/eip-5792.md
lukasrosario: yeah i think this makes sense if we're talking about single chain. would be tough if multichain.

also not clear to me where `DISCARDED` would come up. @forshtat can you share more on how you were thinking this might come up?

EIPS/eip-5792.md
forshtat: > Is the complaint that capabilities makes it hard to specify info on a per call basis?

Yes, that pretty much sums it up. Of course, the `capability` could provide an array of details, which would have to match the `calls` array's length exactly and contain `null` values for calls that do not require extra parameters, which is functionally equivalent but seems like a much more quirky API to me.

> Or you're saying one call in particular can revert?

In my example, yes, I was talking about an "atomic" batch with a single "optional" call that should not revert the batch;
It is hard to reason about potential future additional examples, so these are a little hypothetical, but in my opinion, worth considering:
If a "per-call gas limit capability" is defined, it can be used to provide a per-call gas limit;
In case a "multi-chain capability" is defined, the 'capabilitiesData' field can be used to specify a per-call 'chainId' and L2-specific parameters, something like `gas_per_pubdata_limit`;
In case a "conditional call capability" is defined, it can provide some description of a condition to include or exclude a call from the batch;
etc.

Does this make sense? Or should dapps never need to go to the trouble of defining per-call parameters for a `sendCalls` request?

EIPS/eip-5792.md
forshtat: There are two issues raised here:

1. App providing a "capability" that is *not supported* by the wallet:

There is a `wallet_getCapabilities` method that provides the application with a list of *supported* "capabilities".
Shouldn't the app that ignores that list and submits a `wallet_sendCalls` with a "capability" that is not supported be considered to behave incorrectly? I think it should reject the call, which would prevent dapps from yoloing their `sendCalls` requests.

2. Wallet discretion to use or ignore a *supported* "capability" requested by the dapp:

As the API defined here deals with real value and assets, making capabilities `required` by default makes a lot of sense to me. If using a paymaster for a transaction is indeed optional, it is trivial to specify `optional: true`, but forgetting to set `required: true` for a security-oriented capability may be catastrophic. 

What do you think?

EIPS/eip-5792.md
forshtat: I fully support renaming this parameter. `capabilitiesResultsData` is a bit long, how about just `capabilitiesResults` or `capabilitiesStatus`?

forshtat:
> > Specify that the 'identifier' must be a 32 byte hex.
> 
> @forshtat I'm ok with constraining it to hex, but 32 bytes makes things difficult. For example, in our case the calls id packs the chain id and user op hash together, which we wouldn't be able to do with 32 bytes. Especially with conversations around allowing calls across multiple chains, it would be inevitable that wallets would need to have some more complex state to store + report statuses, which they would not be too happy about.
> 
> how would you feel about hex of any length?

Hmm. I assumed the `identifier` was just some kind of a hash of the request, so there is never a need to parse the identifier to learn any information, but to simply look it up in memory.
I guess the `identifier` could be used to contain some data, but intuitively it feels like a bit of an anti-pattern.
In your case the identity of a batch is `{chainId, userOpHash}`, right? Does it create a lot of overhead if you had to "remember" the `chainId`-of-the-`userOpHash` mapping for the period that the `wallet_getCallsStatus` is expected to remember the `identifier`?
If I recall correctly, the `userOpHash` includes a `chainId` anyways so there is basically no chance of hash collision between two User Operations on different chains.

Anyways, maybe instead of "any length" we should just pick a length that makes most sense and just left-pad it with zeroes?

EIPS/eip-5792.md
forshtat: Yes, it sounds similar to "FAILURE" but the difference I had in mind was that a "FAILURE" means the action failed due to on-chain reverts, and "DISCARDED" means the wallet gave up on trying to send the batch to the network.
Maybe the names do not make the distinction clear enough, or do you think the distinction is not worth the extra status code?

One example could be a batch that specifies a `paymasterService` capability, and given the wallet need to go through these steps before sending anything on-chain:
Screenshot 2024-08-28 at 18 16 29

The wallet cannot block the `wallet_sendCalls()` for all this time, so returns some `identifier` and proceeds with preparing the UserOperation. 
During the preparation phase the `wallet_getCallsStatus` returns PENDING as nothing really happened.
However, if the "Paymaster Service" fails with an error at some stage, wallet could keep retrying to reach it for some time, and the transaction has not technically "failed", but after 2 minutes of trying to reach the "Paymaster Service" it gives up.


Another example is a UserOperation that has some condition in its validation function. Canonical ERC-4337 limits this use-case a lot, but does not completely prevent it. A wallet may want to accept a batch that is not yet valid, check if it has become valid during the next 2 hours, and mark the entire operation as "DISCARDED" after that.

Does this makes sense?

lukasrosario:
> Does it create a lot of overhead if you had to "remember" the chainId-of-the-userOpHash mapping for the period that the wallet_getCallsStatus is expected to remember the identifier?

It does. Because in our case, and in the case of some other wallet teams we've been chatting with, getCallsStatus will route to a server so an app can get a status without opening an extension / popup / app. So either:
1. The relevant information needs to be packed into the identifier or 
2. This server needs to store a mapping of an arbitrary identifier to the relevant ops.

Maybe it's inevitable, but we (and presumably other wallets) would like to avoid (2) if possible, as it would require introducing some storage into these servers.

64 bytes would be sufficient if:
1. We keep calls on a single chain and
2. We are talking about 4337 wallets.

While user op hashes hash over the chain id, we would still need to pack the chain id and user op hash together in the identifier so we know which bundler to call.

This falls about if we start to talk about making the RPC multichain though, as we could never guarantee all the info we need would fit into whatever max length we specify. It also does not work well for EOAs that would need to send multiple transactions for multiple calls and keep track of multiple transaction hashes.

forshtat:
> RE: 64 bytes would be sufficient if... This fails about if we start to talk about making the RPC multichain though, as we could never guarantee all the info we need would fit into whatever max length we specify.

Doesn't it raise a concern that the `identifier` size may end up growing out of control as long as extra information has to be injected into it? That basically the ability to use unrestrictedly large byte arrays as identifiers would be seriously abused?
Like, technically, `Buffer.from(JSON.stringify(userOperationsArray)).toString('hex')` would also be a valid "any length hex identifier", wouldn't it?

EIPS/eip-5792.md
wilsoncusack: So @forshtat is the conclusion here that each capability *could* have an `optional: ` flag? But passing a capability the wallet doesn't have would be hard failure?

EIPS/eip-5792.md
wilsoncusack: I prefer clarity even if verbose so I think I like `capabilitiesResultsData`

lukasrosario:
> Doesn't it raise a concern that the identifier size may end up growing out of control as long as extra information has to be injected into it?

Yeah, that's why I'm starting to think it's inevitable that wallets will need to have some type of storage in their backends. Could alternatively tag on extra calldata with relevant info (unfortunately at the cost of the user).

For now though assuming we stay with one chain as you have here can we say 64 bytes? This would be sufficient for wallets that can do atomic batching.

EIPS/eip-5792.md
forshtat: > So @forshtat is the conclusion here that each capability _could_ have an `optional: ` flag? But passing a capability the wallet doesn't have would be hard failure?

In my personal opinion, yes, this seems pretty reasonable.

EIPS/eip-5792.md
forshtat: > I prefer clarity even if verbose so I think I like `capabilitiesResultsData`

Done.

forshtat:
> For now though assuming we stay with one chain as you have here can we say 64 bytes?

Done.

EIPS/eip-5792.md
```diff
@@ -122,38 +136,55 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
 type GetCallsParams = string;
 
 type GetCallsResult = {
-  status: 'PENDING' | 'CONFIRMED';
+  batchId: `0x${string}`;
```
lukasrosario, line 139: callsId

lukasrosario: should keep the name the same

EIPS/eip-5792.md
```diff
@@ -122,38 +136,55 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
 ```
 
+##### Status Codes for `batchStatus` field
+
+The purpose of the `batchStatus` field is to provide a short summary of the current status of the batch.
+It provides some off-chain context to the array of inner transaction `receipts`.
+
+| Name      | Description                                                                                                          |
+|-----------|----------------------------------------------------------------------------------------------------------------------|
+| PENDING   | Batch has been received by the wallet but has not completed execution on-chain                                       |
```
ajhodges, line 165: How do you feel about possibly expanding `PENDING` to capture more of the offchain context around the pending state?

For example

- `PENDING`: Batch has been received by the wallet but has not completed execution on-chain 
- `UNDERPRICED`: Batch has been received by the wallet but the gas prices are too low for the current chain conditions
- `STALLED`: Batch has been received by the wallet but is stuck for non-pricing related reasons; i.e. ERC-4337 bundler is down. The definition of 'stuck' is left up to the wallet, i.e. batch is priced appropriately but has not been included in >60s

lukasrosario: @forshtat would be good to get your thoughts here. imo `UNDERPRICED` specifically seems like a reasonable addition, as it could be retried.

@ajhodges i'm not a huge fan of the ambiguity for `STALLED` though. to me this is effectively still `PENDING`, and each wallet should have their own interpretations of `PENDING` taking "too long". if we and others feel strongly about it being its own status, i would prefer to be more opinionated in the spec and define when a `PENDING` batch becomes `STALLED`, instead of leaving it open-ended.

jxom: What if we just left it as `PENDING` and `SUCCESS`, and had a sibling `statusCode` or `detail` property that could be extensible enough to accommodate for what ever addition? 
This would make it a lot easier to coerce on frontend apps.

forshtat: > `UNDERPRICED` specifically seems like a reasonable addition, as it could be retried.

I don't have any strong feelings against `UNDERPRICED` or any other status code we can clearly define.

About `UNDERPRICED` specifically, I guess its difference with the regular `PENDING` state is that the user needs to confirm a modification to the original request? And if the wallet could just increase the price internally and retry, it would not be different from `PENDING`, would it?

If so, maybe a better broader category would be `ACTION_REQUIRED` or something like that, what do you think?

EIPS/eip-5792.md
forshtat: Done.

EIPS/eip-5792.md
jxom: Yeah, I guess `statusCode` might be more realistic than `detail` as an application can choose which message they want to display to an end-user (if anything). 

Perhaps this could be a similar pattern to HTTP status codes where `2xx` ≈ success, `4xx` ≈ client error, `5xx` ≈ server error, etc. For this case, maybe `1xx` represents `PENDING` sub-statuses, `2xx` represents `CONFIRMED` sub-statuses, and so on. For example, a `status` of `PENDING` and a `statusCode` of `101` could mean “stalled”, and an application could either choose to show a generic loading indicator, or a loading indicator with “this tx is taking a while, pls wait”. 

I guess the motivation of a sub-status field like `statusCode` is that it would make the branching logic within an application so much easier since 5792 is an ERC for applications.

ajhodges: >I guess the motivation of a sub-status field like statusCode is that it would make the branching logic within an application so much easier — they can rely on the simplicity of status and choose to show extra information by utilizing statusCode.

Yep, exactly. As an app, I want to know if things are healthy. Is my transaction stuck? Is there anything I can do about it if it's stuck? (i.e. underpriced vs stalled). I think the display logic/messaging to the user should be left up to the app, but the wallet can oftentimes provide additional context about what happened to make the UX a little better.

I agree `UNDERPRICED` is maybe a bit vague. But the wallet should be the expert on the current chain conditions and the bundle pricing so it's best suited to detect if the calls should be retried (and we shouldn't prescribe any conditions in this EIP). So maybe `RETRYABLE` is a better sub-status than `UNDERPRICED`, signifying that the calls might go through quicker if you retry. I think I like this better than `ACTION_REQUIRED` because technically an underpriced tx can still go through if chain conditions change

EIPS/eip-5792.md
```diff
@@ -35,33 +35,41 @@ We also define one capability expression for use with the fourth method, which f
...
+Requests that a wallet submits a batch of calls. `from` and `chainId` are top-level properties rather than per-call properties because all calls should be sent from the same sender and on the same chain, identified by [EIP-155](./eip-155.md) integers expressed in hexadecimal notation.
+The items in the `calls` field are simple `{to, data, value}` tuples.
 
 The capabilities field is how an app can communicate with a wallet about capabilities a wallet supports. For example, this is where an app can specify a paymaster service URL from which an [ERC-4337](./eip-4337.md) wallet can request a `paymasterAndData` input for a user operation.
 
+Each capability defined in the "capabilities" member can define global or call specific fields.
+These fields are set inside this capability's entry in the `capabilitiesData` object.
```
jxom, line 44: I’m curious why is this termed as `capabilitiesData` instead of just `capabilities` for consistency? Isn’t this the same thing as global capabilities but on a per-call level?

EIPS/eip-5792.md
```diff
@@ -122,38 +136,55 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
       topics: `0x${string}`[];
     }[];
     status: `0x${string}`; // Hex 1 or 0 for success or failure, respectively
-    chainId: `0x${string}`;
     blockHash: `0x${string}`;
     blockNumber: `0x${string}`;
     gasUsed: `0x${string}`;
     transactionHash: `0x${string}`;
   }[];
+  capabilitiesResultsData?: Record<string, any> | undefined;
```
jxom, line 154: Should this just be called `capabilities` for consistency?

EIPS/eip-5792.md
jxom: Super nit: kinda think `id` is even sufficient enough and maybe more consistent (ie. we don’t call it `transactionHash` on the transaction schema)

EIPS/eip-5792.md
```diff
@@ -212,17 +242,18 @@ If any of these supplemental expressions of capabilities are contradicted by cap
 
 Capabilities are returned in key/value pairs, with the key naming a capability and a value conforming to a shape defined for that name, in an object keyed to the relevant [EIP-155](./eip-155.md) `chainId` expressed in hexadecimal notation.
 Capabilities are nested in per-chain objects because wallets may support different capabilities across multiple chains authorized in a given session.
+All values for the "capability" fields MUST contain Semantic Version value `version`.
```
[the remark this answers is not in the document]
jxom: Yeah, would make it simpler.

EIPS/eip-5792.md
```diff
@@ -122,38 +147,63 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
 type GetCallsParams = string;
 
 type GetCallsResult = {
-  status: 'PENDING' | 'CONFIRMED';
+  id: `0x${string}`;
+  chainId?: `0x${string}`;
+  status: string; // See "Status Codes"
```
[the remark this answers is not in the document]
lukasrosario: ah true, just missed this when making the status change, will update.

EIPS/eip-5792.md
```diff
@@ -191,14 +240,12 @@ type ShowCallsParams = string; // Call bundle identifier returned by wallet_send
...
-]
+["0xe670ec64341771606e55d6b4ca35a1a6b75ee3d5145a99d05921026d1527331"]
 ```
 
 ### `wallet_getCapabilities`
 
-This RPC allows an application to request capabilities from a wallet (e.g. batch transactions, paymaster communication), without distinct discovery and permission requests. For more on the difference between requesting capabilities and discoverying features, see the ["Privacy Considerations" section](#privacy-considerations).
+This RPC allows an application to request capabilities from a wallet (e.g. batch transactions, paymaster communication), without distinct discovery and permission requests. For more on the difference between requesting capabilities and discovering features, see the ["Privacy Considerations" section](#privacy-considerations).
 
 This method SHOULD return an error if the user has not already authorized a connection between the application and the requested address.
```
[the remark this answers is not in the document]
forshtat: There is no definition of specific errors or error codes for any of the APIs, we could either define all of them explicitly or leave it to the wallets. Should we?

EIPS/eip-5792.md
```diff
@@ -251,9 +298,7 @@ The capabilities below are for illustrative purposes.
 
 Like the illustrative examples given above and other capabilities to be defined in future EIPs, the capability to execute calls delivered via the above-defined methods in a single transaction can be attested by the wallet in boolean form.
 
-This capability is expressed separately on each chain and should be interpreted as a guarantee only for batches of transactions on that chain; batches including calls to multiple chains is out of scope of this capability and this specification.
-
-If a wallet has affirmatively expressed this `atomicBatch` capability to a calling application, it MUST submit calls submitted with `wallet_sendCalls` as part of a single transaction.
+Note that this capability is purely declarative. I.e. this capability is not meant to be paired with a `wallet_sendCalls` request (at either top-level nor call-level), as this EIP specifies that multiple calls MUST, by default, be executed as part of a single transaction.
```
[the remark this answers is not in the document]
forshtat: This section has been removed now.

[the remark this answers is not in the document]
forshtat: This section has been removed now.

EIPS/eip-5792.md
```diff
@@ -251,9 +298,7 @@ The capabilities below are for illustrative purposes.
 
 Like the illustrative examples given above and other capabilities to be defined in future EIPs, the capability to execute calls delivered via the above-defined methods in a single transaction can be attested by the wallet in boolean form.
```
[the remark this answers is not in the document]
forshtat: This section has been removed now.

EIPS/eip-5792.md
```diff
@@ -35,38 +35,52 @@ We also define one capability expression for use with the fourth method, which f
...
+Each entity in the `calls` field may contain an optional `capabilities` object.
+This object allows the applications to attach a capability-specific metadata to individual calls.
+
 The wallet:
 
 * MUST send the calls in the order specified in the request
-* MUST send the calls on the same chain identified by the call's `chainId` property
+* MUST send the calls on the same chain identified by the request's `chainId` property
 * MUST NOT await for any calls to be finalized to complete the batch
+* MUST submit multiple calls as part of a single transaction, unless instructions for sequential execution are provided via a `capability`.
```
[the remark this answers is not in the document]
forshtat: Please see:
https://github.com/ethereum/EIPs/pull/8826/commits/0de3f96d90cf95929bb85b193eb62bb7f0bff33a

EIPS/eip-5792.md
```diff
@@ -76,43 +90,54 @@ type SendCallsResult = string;
...
   }
 ]
 ```
 
+Note that since the `paymasterService` `capability` is marked as optional, wallets that do not support it can still process and handle the request to the best of their ability. If this `optional` field were set to `false` or absent from the request, wallets that do not support the `capability` MUST reject the request.
+
 ##### `wallet_sendCalls` Example Return Value
 
-The identifier can be any string. The only requirement is that for a given session, users should be able to call `wallet_getCallsStatus` with this value and expect a call-batch status to be returned.
+The identifier MUST be unique 64 bytes represented as a hex encoded string.
```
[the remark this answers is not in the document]
forshtat: I think a 64 bit identifier can be assumed globally unique for the purpose of this EIP, but we can add "for the current session".
But, as you've mentioned, we did not yet define what a "session" actually is for the `wallet_sendCalls`.

EIPS/eip-5792.md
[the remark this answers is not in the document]
lukasrosario: for this one specifically i would suggest 4100 unauthorized. i dont think any others need to be defined ? as i believe they are covered by the other 1193 errors + getCallsStatus errors

EIPS/eip-5792.md
```diff
@@ -76,43 +90,54 @@ type SendCallsResult = string;
...
+Returns the status of a call batch that was sent via `wallet_sendCalls`.
+The identifier of the batch is the value returned from the `wallet_sendCalls` RPC.
+Note that the `receipts` objects of this method's response is a strict subset of the object returned by `eth_getTransactionReceipt`.
+
+The `capabilities` object allows the wallets to attach a capability-specific metadata to the response.
 
-* If a wallet does not execute multiple calls atomically (i.e. in multiple transactions), the receipts in the `receipts` field MUST be in order of the calls sent.
+* The receipts in the `receipts` field MUST be in order they are included onchain.
+* The `receipts` field MUST contain receipts for all calls in a batch that were included onchain, including reverted ones.
 * If a wallet executes multiple calls atomically (i.e. in a single transaction), `wallet_getCallsStatus` MUST return a single receipt, corresponding to the transaction in which the calls were included.
```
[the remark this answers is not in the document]
drortirosh: IIUC:
- if the sendCalls performs an atomic transaction (as is the default for this ERC), then there is a single receipt for the transaction itself - not for each inner call.
- if there are separate transactions for the calls (which is not fully specified in this ERC) - then there is a receipt for each call.
> multisig wallet, where each signature is recorded in a transaction, then applied atomically once the quorum is reached.
If each multisig owner separately submits a transaction with its own signature, then each of them is a separate transaction, originated by separate wallets. The last owner signature is batched transaction with the multisig execute that uses all these partial signatures.

EIPS/eip-5792.md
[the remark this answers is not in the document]
forshtat: https://github.com/ethereum/EIPs/pull/8826/commits/1be7d7d6b8b1e028ed64b27f27c3e0a48f208f04

EIPS/eip-5792.md
```diff
@@ -122,38 +147,63 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
+
+The purpose of the `status` field is to provide a short summary of the current status of the batch.
+It provides some off-chain context to the array of inner transaction `receipts`.
+
+| Code | Description |
+|------|-------------|
+| 100  | Batch has been received by the wallet but has not completed execution onchain (pending) |
+| 200  | Batch has been included onchain without reverts, receipts array contains info of all calls (confirmed) |
+| 400  | Batch has not been included onchain and wallet will not retry (offchain failure) |
+| 500  | Batch reverted completely and only changes related to gas charge may have been included onchain (chain rules failure) |
```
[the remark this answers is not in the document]
forshtat: Added a 600 status code, what do you think? We can also make it, say, `501`.
https://github.com/ethereum/EIPs/pull/8826/commits/7701a6445691009d5ffd16b54b87de6796802683

EIPS/eip-5792.md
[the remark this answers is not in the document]
forshtat: Rephrased a little:
https://github.com/ethereum/EIPs/pull/8826/commits/a3d61d369dce4e9fe1fa3847942f23a6e9fd2135

EIPS/eip-5792.md
```diff
@@ -76,43 +90,54 @@ type SendCallsResult = string;
...
 ]
 ```
 
+Note that since the `paymasterService` `capability` is marked as optional, wallets that do not support it can still process and handle the request to the best of their ability. If this `optional` field were set to `false` or absent from the request, wallets that do not support the `capability` MUST reject the request.
+
 ##### `wallet_sendCalls` Example Return Value
 
-The identifier can be any string. The only requirement is that for a given session, users should be able to call `wallet_getCallsStatus` with this value and expect a call-batch status to be returned.
+The identifier MUST be unique 64 bytes represented as a hex encoded string.
+For a given session, users should be able to call `wallet_getCallsStatus` with this value and expect a call-batch status to be returned.
```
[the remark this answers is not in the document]
forshtat: Well, there is no strong definition for a "session" in this case. Is a 24-hour window a better suggestion? The rule is just a "SHOULD" so it is not binding anyways.

https://github.com/ethereum/EIPs/pull/8826/commits/77f32cc65c2bf6c2827d25ac19bef07ed046ec52

EIPS/eip-5792.md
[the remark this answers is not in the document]
forshtat: We could also make it explicitly more limited, something like "The identifier MUST be unique for a given RPC provider within a 24 hour window. It is 64 bytes long and is represented as a hex encoded string."

What do you think? The 24 hour window comes from this commit:

https://github.com/ethereum/EIPs/commit/77f32cc65c2bf6c2827d25ac19bef07ed046ec52

EIPS/eip-5792.md
```diff
@@ -35,38 +35,52 @@ We also define one capability expression for use with the fourth method, which f
...
   from: `0x${string}`;
+  chainId: `0x${string}` | undefined; // Hex chain id
   calls: {
     to?: `0x${string}` | undefined;
     data?: `0x${string}` | undefined;
     value?: `0x${string}` | undefined; // Hex value
-    chainId?: `0x${string}` | undefined; // Hex chain id
+    capabilities?: Record<string, any> | undefined;
   }[];
   capabilities?: Record<string, any> | undefined;
```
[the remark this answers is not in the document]
forshtat: https://github.com/ethereum/EIPs/pull/8826/commits/534c81b454d1906edce159ee99b4aed4d0f463f6

EIPS/eip-5792.md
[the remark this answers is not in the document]
jxom: What is the rationale behind a strict 64-byte identifier? 🤔

EIPS/eip-5792.md
```diff
@@ -308,7 +324,7 @@ Wallets that do not support the methods defined here SHOULD return error respons
...
+App developers MUST treat each call in a batch as if the call were an independent transaction. In other words, there may be additional untrusted transactions between any of the calls in a batch. The calls in the batch MAY also be included in separate, non-contiguous blocks. There is no constraint over how long it will take all the calls in a batch to be included. Apps MUST encode deadlines and timeout behaviors in the smart contract calls, just as they do today for transactions, including ones otherwise bundled. App developers MUST NOT assume that all calls will be sent in a single transaction.
```
SamWilsn, line 327: > MUST submit multiple calls as an atomic unit in a single transaction

This can be removed, but I'll steal it for flow-control

EIPS/eip-5792.md
```diff
@@ -76,44 +97,55 @@ type SendCallsResult = string;
...
 ]
 ```
 
+Note that since the `paymasterService` `capability` is marked as optional, wallets that do not support it will still process and handle the request as if the capability had not been present. If this `optional` field were set to `false` or absent from the request, wallets that do not support the `capability` MUST reject the request.
+
 ##### `wallet_sendCalls` Example Return Value
 
-The identifier can be any string. The only requirement is that for a given session, users should be able to call `wallet_getCallsStatus` with this value and expect a call-batch status to be returned.
+The identifier MUST be unique 64 bytes represented as a hex encoded string.
+Within 24 hours from the corresponding `wallet_sendCalls`, wallets SHOULD return a call-batch status when `wallet_getCallsStatus` is called with this value.
```
frangio, line 128: This line should be moved from the example subsection to the spec part.

borislav-itskov (approved):

EIPS/eip-5792.md
```diff
@@ -31,42 +31,77 @@ In updating to a new set of `wallet_` RPCs, our main goals are to enforce a clea
...
+
+If provided, the wallet MUST respect the `id` field and return it in the response.
+
+Identifiers, whether provided by the app or generated by the wallet, MUST be a unique string up to 4096 bytes (8194 characters including leading `0x`).
+
+App-provided `id`s MUST be unique per sender per app, where each "app" SHOULD be identified by their domain.
+
+Wallets MUST reject requests with duplicate `id`s.
+
+Within 24 hours from the corresponding `wallet_sendCalls`, wallets SHOULD return a call-batch status when `wallet_getCallsStatus` is called with the same `id`.
```
jakubuid, line 74: Does it mean that the wallet should store the `wallet_sendCalls` result for 24hours? What's the reasoning for that?

forshtat: But, if the wallets do not store and return the results of `wallet_sendCalls`, how would an app ever get them?

jakubuid: My question is about the reasoning for 24 hours. Wondering if it's just an arbitrary number or sth else.
I'd even consider changing SHOULD to MUST due to having `wallet_getCallsStatus` calls guaranteed

forshtat: Ok, got it. Yes, "24 hours" is an arbitrary number, and that is why I wrote SHOULD - if the value is not available after, say, 19 hours it is probably not such a big deal. I guess in the vast majority of cases the `wallet_getCallsStatus` will be only relevant for a couple of minutes. 
We can change it to MUST if we want to be strict about it, and we can increase the time window if there is a potential use-case for it.

Issalk (approved):

bumblefudge (approved):

EIPS/eip-5792.md
```diff
@@ -31,42 +31,77 @@ In updating to a new set of `wallet_` RPCs, our main goals are to enforce a clea
...
+Unless these requirements are explicitly overriden by a certain `capability`, the wallet must adhere to the following rules.
+Note that such a `capability` is not in the scope of this EIP and should be defined in its own ERC.
+
 The wallet:
 
 * MUST send the calls in the order specified in the request
-* MUST send the calls on the same chain identified by the call's `chainId` property
+* MUST send the calls on the same chain identified by the request's `chainId` property
+* MUST send the calls from the address specified in the request's `from` property, if provided
+  * If `from` is not provided the wallet SHOULD present the user with an opportunity to view and change the `from` address during confirmation
```
bumblefudge, line 55: This is a little hard to parse-- if it's not provided there's nothing to change? do you mean:
```suggestion
  * If `from` is not provided the wallet SHOULD present the user with an opportunity to select a `from` address during confirmation
```

EIPS/eip-5792.md
bumblefudge: it's odd grammatically, because it could be read as "wallets are expected to return a call status at some point in the next 24 hours" or it cold be read as "apps can assume they're never getting a call-batch if 24 hours pass". maybe worth breaking wallet behavior and app behavior into 2 distinct sentences?

EIPS/eip-5792.md
```diff
@@ -122,38 +163,66 @@ Returns the status of a call batch that was sent via `wallet_sendCalls`. The ide
...
       topics: `0x${string}`[];
     }[];
     status: `0x${string}`; // Hex 1 or 0 for success or failure, respectively
-    chainId: `0x${string}`;
     blockHash: `0x${string}`;
     blockNumber: `0x${string}`;
     gasUsed: `0x${string}`;
     transactionHash: `0x${string}`;
   }[];
+  capabilities?: Record<string, any>;
```
bumblefudge, line 182: what happens if these caps changed since last interaction? i'm curious what these are doing here.  just for matching?

EIPS/eip-5792.md
forshtat: > "wallets are expected to return a call status at some point in the next 24 hours" or it cold be read as "apps can assume they're never getting a call-batch if 24 hours pass".

Frankly, I feel fine leaving any behaviour after the 24-hour window udefined, as this is an extremely long time for a transaction.
As for the meaning within the 24 window, clearly the wallet will want to return a status as soon as possible, and the dapps will not be able to work with a wallet that is not able to do so.
If you can suggest a more clear definition of wallet/app expectations, I would appreciate if you could make a PR.

EIPS/eip-5792.md
forshtat: Replaced 'change' with 'select' to indicate this field is not user-editable

EIPS/eip-5792.md
forshtat: > L153: The `capabilities` object allows the wallets to attach a capability-specific metadata to the response.

Meaning, this field contains any data specific to a requested "capability", not the list of supported capabilites as in `wallet_getCapabilities`.

lukasrosario (approved):

eip-automerger:

 - Trying to change EIP 5792 state from Review to Last Call

PR #8829, opened 2024-08-27
Update EIP-7702: remove outdated note on forward compatibility

k1rill-fedoseev:
I was reviewing the updated semantics of EIP-7702 after #8677, which now sets the account code permanently in the trie. One of the forward compatibility notes refers to the earlier version of this EIP and seems confusing / no longer relevant.


lightclient (approved):
Thank you

PR #8832, opened 2024-08-28
Update EIP-7702: selfdestruct opcode behaviour change in eip-7702

sudeepdino008:
corresponds to this [discord comment](https://discord.com/channels/595666850260713488/718596092828057631/1275484533965459456)

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -63,8 +63,8 @@ The [EIP-2718](./eip-2718.md) `ReceiptPayload` for this transaction is `rlp([sta
 
 At the start of executing the transaction, after incrementing the sender's nonce, for each `[chain_id, address, nonce, y_parity, r, s]` tuple do the following:
 
-1. Verify the chain id is either 0 or the chain's current ID.
-2. `authority = ecrecover(keccak(MAGIC || rlp([chain_id, address, nonce])), y_parity, r, s]`
+1. `authority = ecrecover(keccak(MAGIC || rlp([chain_id, address, nonce])), y_parity, r, s]`
+2. Verify the chain id is either 0 or the chain's current ID.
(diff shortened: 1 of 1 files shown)


lightclient:
This clarification makes sense and is in general how I anticipated the EIP to be implemented. Currently the EIP doesn't specify that the auth list accounts should be considered "created in this tx" as defined by EIP-6780:

> A contract is considered created at the beginning of a create transaction or when a CREATE series operation begins execution (CREATE, CREATE2, and other operations that deploy contracts in the future). If a balance exists at the contract’s new address it is still considered to be a contract creation.

jochem-brouwer (approved):

drortirosh:
1. SELFDESTRUCT indeed need to be specified explicitly for 7702 - namely, remove the delegate at the end of the transaction, so that the EOA is left as EOA for future transactions.
2. to prevent replay, it should NOT revert the nonce-bump

gumb0:
Suggestion for two cases that the tests should cover:

1. Set code transaction creates delegation (originally to empty account), same transaction creates contract at delegate address, then calls the delegated account, the code executes SELFDESTRUCT.
2. Delegation exists in prestate, regular transaction creates contract at delegate address, then calls the delegated account, the code executes SELFDESTRUCT.

RyanDinh8:
> corresponds to this [discord comment](https://discord.com/channels/595666850260713488/718596092828057631/1275484533965459456)

Hi @sudeepdino008 
This is Ryan from SubWallet - a multichain wallet supporting EVM. We're building a POC for EIP-7702. It seems I'm not able to open the Discord link. Could you please share the link again?

sudeepdino008:
I've been out of loop on this for a while. I suppose we should merge this PR? There's been no change on this behavior.

PR #8833, opened 2024-08-28
Update EIP-7702: Restrict chain_id to chain's id or zero

rakita:
As we already checking if `y_parity` and `s` values additionally adding a check for correct `chain_id` seems logical.

This would make the EIP-7702 transaction invalid if it is not zero (chain agnostic) or has chain_id different from the Ethereum-s chain_id.

Would propose that this change is added after `devnet-3` to not disrupt testing/developing efforts.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -46,7 +46,7 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 Transaction is considered invalid if authorization list items can't be decoded as:
 
-* `chain_id`: Value 1.
+* `chain_id`: Value 0 or 1.
 * `nonce`: unsigned 64-bit integer.
 * `address`: 20 bytes array.
 * `y_parity`: Value 0 or 1.
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -46,7 +46,7 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 Transaction is considered invalid if authorization list items can't be decoded as:
 
-* `chain_id`: unsigned 256-bit integer.
+* `chain_id`: Value 1.
 * `nonce`: unsigned 64-bit integer.
 * `address`: 20 bytes array.
 * `y_parity`: Value 0 or 1.
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -46,7 +46,7 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 Transaction is considered invalid if authorization list items can't be decoded as:
 
-* `chain_id`: Value 0 or 1.
+* `chain_id`: chain's current ID or zero
 * `nonce`: unsigned 64-bit integer.
 * `address`: 20 bytes array.
 * `y_parity`: Value 0 or 1.


EIPS/eip-7702.md
```diff
@@ -46,7 +46,7 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 Transaction is considered invalid if authorization list items can't be decoded as:
 
-* `chain_id`: unsigned 256-bit integer.
+* `chain_id`: Value 1.
```
rakita, line 49:
```suggestion
* `chain_id`: Value 0 or 1.
```

lightclient:
We want the chain id to equal the chain's current ID or zero, no just zero and one.

EIPS/eip-7702.md
```diff
@@ -46,7 +46,7 @@ authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 
 Transaction is considered invalid if authorization list items can't be decoded as:
 
-* `chain_id`: unsigned 256-bit integer.
+* `chain_id`: Value 0 or 1.
```
rakita, line 49:
```suggestion
* `chain_id`: chain's current ID or zero
```

rakita: cc @lightclient ?

lightclient: since the chain's ID can really be an uint256, I don't see the need to add this

jochem-brouwer: We should check for chain_id or zero here, not for 0 or 1 (because this would invalidate all txs on chain ids which are not 1, and which txs are specifically signed for that chain) - this would invalidate all specific devnet-txs signed with nonzero chain id.

etan-status:
https://ethereum-magicians.org/t/eip-2294-explicit-bound-to-chain-id/11090/16?u=etan-status

I have asked about whether zero is being used sometime early 2023 here. Seems to be actually unused, but it would have to be formally reserved to indicate `chain-agnostic` (it doesn't have that special meaning right now)

gumb0:
Should the check 1. in `Behaviour` be removed then?

lightclient:
I think the way these checks are being added is a bit wrong. We want to first define the serialization of auth list, giving reasonable bounds, and then define the validity of the transaction given certain auth tuples.

For that reason, I think chain id should to be considered valid as long as it is less than `2**256 - 1`. I don't think we should make the entire tx invalid just because an inner element has an invalid chain id -- therefore I am not in support of this PR.

bumblefudge:
I haven't been following closely enough to agree or disagree with lightclient's point above, but I _will_ just point out that the use of chainId 0 for offchain/EOA use-cases is documented [here](https://namespaces.chainagnostic.org/eip155/caip10#special-case-of-eoa), if anyone wants to normref it (note that the EIP link policy would require linking to the [commit-specific markdown on github](https://github.com/ChainAgnostic/namespaces/blob/01e33efff282bad38ff04c77ee3ecdaf060a31d0/eip155/caip10.md#special-case-of-eoa), not the rendered jekyll output).  it sounds like everyone has bigger fish to fry, tho, so tag me later if it comes back up in clean-up after the major construction is done!

lightclient:
I think we've agreed to keep the validity check minimal and be based on the integer width.

PR #8840, opened 2024-08-30
Update EIP-7702: adjust tx validity

lightclient:
The past few weeks we've been discussing validity conditions for 7702 transactions. Originally @rakita added the decoding limits for auth list items in #8746. Now we're looking at further constraining validity in #8833 to disallow chain ids other than zero or the current chain's id.

My philosophy here is 1) tx validity should be as simple as possible to ascertain and 2) we should retain as much similarity across tx types as possible.

The motivation for 1) is that tx originators are incentivized to not send junk authorizations with their txs, therefore they won't. What's the point of adding more constraints into the protocol when they aren't necessary?

For 2), I believe it's extremely important to follow existing conventions when possible. If every single tx type is designed as an ornate protocol fixture by core devs of that era, we will have a mess of formats that are extremely difficult to reason about as a whole or onboard new developers to understanding. Of course this doesn't mean we shouldn't improve the status quo when possible (see: separating chain id back out from `v` value), but unless the benefit is significant, let's not deviate from other tx formats.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -44,18 +44,18 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
 authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 ```
 
-Transaction is considered invalid if authorization list items can't be decoded as:
-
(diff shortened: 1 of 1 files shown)


lightclient:
Please see #8845 -- I didn't mean for this PR to automatically merged and it has been reverted.

PR #8841, opened 2024-08-30
Update EIP-6800: remove text around SELFDESTRUCT deprecation

fjl:
Since EIP-6780 was included in the Cancun fork, and that fork is deployed on mainnet, removal of SELFDESTRUCT is no longer a backwards-compatibility breaking change.


gballet (approved):
LGTM

PR #8845, opened 2024-08-30
Update EIP-7702: 7702 validity

lightclient:
Repost from #8840 as it should not have automatically merged.

> The past few weeks we've been discussing validity conditions for 7702 transactions. Originally @rakita added the decoding limits for auth list items in #8746. Now we're looking at further constraining validity in #8833 to disallow chain ids other than zero or the current chain's id.

> My philosophy here is 1) tx validity should be as simple as possible to ascertain and 2) we should retain as much similarity across tx types as possible.

> The motivation for 1) is that tx originators are incentivized to not send junk authorizations with their txs, therefore they won't. What's the point of adding more constraints into the protocol when they aren't necessary?

> For 2), I believe it's extremely important to follow existing conventions when possible. If every single tx type is designed as an ornate protocol fixture by core devs of that era, we will have a mess of formats that are extremely difficult to reason about as a whole or onboard new developers to understanding. Of course this doesn't mean we shouldn't improve the status quo when possible (see: separating chain id back out from `v` value), but unless the benefit is significant, let's not deviate from other tx formats.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -44,18 +44,18 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
 authorization_list = [[chain_id, address, nonce, y_parity, r, s], ...]
 ```
 
-Transaction is considered invalid if authorization list items can't be decoded as:
-
-* `chain_id`: unsigned 256-bit integer.
-* `nonce`: unsigned 64-bit integer.
-* `address`: 20 bytes array.
-* `y_parity`: Value 0 or 1.
-* `r`: unsigned 256-bit integer.
-* `s`: unsigned 256-bit integer and value less or equal than `secp256k1n/2`, specified in [EIP-2](./eip-2.md).
-
 The fields `chain_id`, `nonce`, `max_priority_fee_per_gas`, `max_fee_per_gas`, `gas_limit`, `destination`, `value`, `data`, and `access_list` of the outer transaction follow the same semantics as [EIP-4844](./eip-4844.md). *Note, this means a null destination is not valid.*
 
 The `authorization_list` is a list of tuples that store the address to code which the signer desires to execute in the context of their EOA. The transaction is considered invalid if the length of `authorization_list` is zero.
+
+The transaction is also considered invalid when any integer field in an
+authorization list cannot fit within the following bounds:
+
+* `chain_id`: `2**256-1`
+* `nonce`:`2**64-1`.
+* `y_parity`: `2**256-1`.
+* `r`: `2**256-1`.
+* `s`: `2**256-1`.
 
 The [EIP-2718](./eip-2718.md) `ReceiptPayload` for this transaction is `rlp([status, cumulative_transaction_gas_used, logs_bloom, logs])`.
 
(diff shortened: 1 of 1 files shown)


rakita:
My stance on this is:
* Sanity limits need to be made. We i think all agree on this.
* @jochem-brouwer asked to add checks on `s-value` and `v`. I don't have a stance on this.
   * But **If** `s-value` and `v` are checked, I would request that we add the check for `chain_Id`
 
 
Side question for this PR, why not use u8 for `v`?

lightclient:
Agree sanity limits need to exist, I just think we should keep them minimal and inline with existing types.

> But If s-value and v are checked, I would request that we add the check for chain_Id

All of these things should be checked, they just shouldn't affect the transactions validity.

> why not use u8 for v?

We bound `v < 2**256-1` everywhere else, it seems weird to do something different here.

EIPS/eip-7702.md
```diff
@@ -44,19 +44,22 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
...
 
 The `authorization_list` is a list of tuples that store the address to code which the signer desires to execute in the context of their EOA. The transaction is considered invalid if the length of `authorization_list` is zero.
 
+The transaction is also considered invalid when any field in an authorization
+tuple cannot fit within the following bounds:
+
+```python
+assert auth.chain_id < 2**256
+assert auth.nonce < 2**64
+assert len(auth.address) == 20
```
drortirosh, line 57: shouldn't it be `len(auth.address)<=20` ?

lightclient: no, address is a fixed length item.

EIPS/eip-7702.md
```diff
@@ -44,19 +44,22 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
...
+assert auth.y_parity < 2**256
+assert auth.r < 2**256
+assert auth.z < 2**256
```
marioevz, line 60:
```suggestion
assert auth.s < 2**256
```

lightclient:
Seems like there was good support for this in ACDE 196. Still need to look more into constraining the `v` value, but this PR can be merged in the meantime.

EIPS/eip-7702.md
```diff
@@ -44,19 +44,22 @@ rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, dest
...
 The fields `chain_id`, `nonce`, `max_priority_fee_per_gas`, `max_fee_per_gas`, `gas_limit`, `destination`, `value`, `data`, and `access_list` of the outer transaction follow the same semantics as [EIP-4844](./eip-4844.md). *Note, this means a null destination is not valid.*
 
 The `authorization_list` is a list of tuples that store the address to code which the signer desires to execute in the context of their EOA. The transaction is considered invalid if the length of `authorization_list` is zero.
 
+The transaction is also considered invalid when any field in an authorization
+tuple cannot fit within the following bounds:
+
+```python
+assert auth.chain_id < 2**256
+assert auth.nonce < 2**64
```
jochem-brouwer, line 56: Have a slight nit about this. (Sorry, just noticed after-merge). By https://eips.ethereum.org/EIPS/eip-2681 should this not be `2**64 - 1`? Or is this checkout outside tx validity, but instead inside EVM?

chfast: It is, but specific to the tx sender. The idea here is to only give numeric boundaries here and leave interpretation to other subsystems.

gumb0: > It is, but specific to the tx sender.

The nonce of authority is also incremented during delegation setting, so it is important to check `nonce < 2**64 - 1` here.

I've made this change in https://github.com/ethereum/EIPs/pull/8905

PR #8854, opened 2024-09-04
Update EIP-7685: change requests hash to flat hash

fjl:
There is no need to create a tree hash over the requests, since proofs related to requests will be made over the CL block. The EL block body stores unvalidated requests and will never be used for proofs. Removing the tree hash simplifies things.

### DIFF
--- a/EIPS/eip-7685.md
+++ b/EIPS/eip-7685.md
@@ -64,18 +64,11 @@ block_body_rlp = rlp([
 
 #### Block Header
 
-Extend the header with a new 32 byte value `requests_root`.
-
-Let `requests_root` be the root of a Merkle-Patricia trie keyed by the index in
-the list of `requests`. This is equivalent to how the transaction trie root is
-computed.
-
- The `requests_root` field value is therefore computed as follows:
+Extend the header with a new 32 byte value `requests_hash`:
 
 ```python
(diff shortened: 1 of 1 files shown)


lightclient:
This makes sense to me in general. We won't be seeing light client proofs against this data as it will be included natively in the CL.

mkalinin:
Moreover, we have requests data in logs https://github.com/lightclient/sys-asm/pull/17

EIPS/eip-7685.md
```diff
@@ -64,18 +64,11 @@ block_body_rlp = rlp([
...
-
- The `requests_root` field value is therefore computed as follows:
+Extend the header with a new 32 byte value `requests_hash`:
 
 ```python
-def compute_trie_root_from_indexed_data(data):
-    trie = Trie.from([(i, obj) for i, obj in enumerate(data)])
-    return trie.root
+def compute_requests_hash(list):
+    return keccak256(rlp.encode([ rlp.encode(req) for req in list]))
```
lightclient, line 71:
```suggestion
    return keccak256(rlp.encode([rlp.encode(req) for req in list]))
```

lightclient (approved):

Scamreno (approved):

PR #8856, opened 2024-09-04
Update EIP-6110: change request to flat encoding

fjl:
I am proposing to change the encoding of deposit requests to be a flat concatenation of the fields returned by the contract. The extra layer of RLP encoding is not necessary, and by removing it, we can avoid defining the structure of these requests in the execution layer client implementation.

### DIFF
--- a/EIPS/eip-6110.md
+++ b/EIPS/eip-6110.md
@@ -94,9 +94,9 @@ def event_data_to_deposit_request(deposit_event_data) -> bytes:
     deposit_data = parse_deposit_data(deposit_event_data)
     pubkey = Bytes48(deposit_data[0])
     withdrawal_credentials = Bytes32(deposit_data[1])
-    amount = deposit_data[2]   # 4 bytes uint64 LE
+    amount = deposit_data[2]   # 8 bytes uint64 LE
     signature = Bytes96(deposit_data[3])
-    index = deposit_data[4]    # 4 bytes uint64 LE
+    index = deposit_data[4]    # 8 bytes uint64 LE
 
     return DEPOSIT_REQUEST_TYPE + pubkey + withdrawal_credentials + amount + signature + index
 
--- a/EIPS/eip-6110.md
+++ b/EIPS/eip-6110.md
@@ -62,19 +62,15 @@ The structure denoting the new deposit operation consists of the following field
 4. `signature: Bytes96`
 5. `index: uint64`
 
-Deposits are a type of [EIP-7685](./eip-7685.md) request, therefore the encoding of the structure must be computed using the `DEPOSIT_REQUEST_TYPE` byte:
+Deposits are a type of [EIP-7685](./eip-7685.md) request, with the following encoding:
 
 ```python
-deposit_request_rlp = DEPOSIT_REQUEST_TYPE + rlp([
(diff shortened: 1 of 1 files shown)


lightclient:
Relies on https://github.com/ethereum/EIPs/pull/8854

EIPS/eip-6110.md
```diff
@@ -94,17 +90,15 @@ def parse_deposit_data(deposit_event_data) -> bytes[]:
 def little_endian_to_uint64(data: bytes) -> uint64:
     return uint64(int.from_bytes(data, 'little'))
 
-def event_data_to_deposit_request_rlp(deposit_event_data) -> bytes:
+def event_data_to_deposit_request(deposit_event_data) -> bytes:
     deposit_data = parse_deposit_data(deposit_event_data)
     pubkey = Bytes48(deposit_data[0])
     withdrawal_credentials = Bytes32(deposit_data[1])
-    amount = little_endian_to_uint64(deposit_data[2])
+    amount = deposit_data[2]   # 4 bytes uint64 LE
```
mkalinin, line 97: Not making the endianness conversion here defers this responsibility to the Engine API handlers

mkalinin:
> we can avoid defining the structure of these requests in the execution layer client implementation.

I believe the structure would still need to be required to serve Engine API requests, wouldn’t it?

EIPS/eip-6110.md
lightclient: The idea is that this is already valid SSZ.

fjl:
The corresponding engine API update is here: https://github.com/ethereum/execution-apis/pull/577

EIPS/eip-6110.md
fjl: Yeah, AFAIK the number is already converted to LE by the contract.

Magken91:
CNN someone tell how do this so I can use funds to pay things

lightclient (approved):

lightclient:
> I believe the structure would still need to be required to serve Engine API requests, wouldn’t it?

I think the idea from Felix's engine api PR is that the data will be "opaque" as far as the EL is concerned.

EIPS/eip-6110.md
```diff
@@ -94,17 +90,15 @@ def parse_deposit_data(deposit_event_data) -> bytes[]:
...
-    amount = little_endian_to_uint64(deposit_data[2])
+    amount = deposit_data[2]   # 8 bytes uint64 LE
     signature = Bytes96(deposit_data[3])
-    index = little_endian_to_uint64(deposit_data[4])
+    index = deposit_data[4]    # 8 bytes uint64 LE
 
-    return DEPOSIT_REQUEST_TYPE + rlp([
-        pubkey, withdrawal_credentials, amount, signature, index
-    ])
+    return DEPOSIT_REQUEST_TYPE + pubkey + withdrawal_credentials + amount + signature + index
```
lucassaldanha, line 101: Do we need to prepend the type? My understanding is that each element does not have the type, and the type is only used when calculating the commitment hash.

lucassaldanha: Ah just realised this has changed in other PRs :)

PR #8867, opened 2024-09-06
Update EIP-4762: unconditionnal account write during contract init

gballet:
In the current state of the spec, there are a few instances where the basic leaf can be created but not the code hash leaf.

This unconditionally write-touch all leaves upon contract initialization, as this will result in an account no matter what happens. The only difference is that the value of the code hash will either be the hash of the contract code, or the empty hash, but that does not impact the gas costs.

### DIFF
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -132,11 +132,6 @@ When a contract creation is initialized, process these write events:
 
 ```
 (contract_address, 0, BASIC_DATA_LEAF_KEY)
-```
-
-When a contract is created, process these write events:
-
-```
 (contract_address, 0, CODEHASH_LEAF_KEY)
 ```
 


EIPS/eip-4762.md
```diff
@@ -132,11 +132,6 @@ When a contract creation is initialized, process these write events:
 
 ```
 (contract_address, 0, BASIC_DATA_LEAF_KEY)
-```
```
jsign, line 135: I wonder if "creation is initialized" is well defined. As in: after collision check.

g11tech (commented):
lgtm

tanishqjasoria:
lgtm

lu-pinto:
lgtm too - we were already doing this in a way

EIPS/eip-4762.md
lu-pinto: I was understanding it as in when the code is set against the contract address after being initialized. That's when hash is computed at least for us

PR #8869, opened 2024-09-06
Update EIP-7702: add instruction to remove code at end of tx

NiloCK:
Original version of the EIP had:

> At the end of the transaction, set the `contract_code` of each `signer` back to empty.

But currently no similar instruction is present. This PR replaces it.


lightclient:
This is specifically no longer a goal of 7702

NiloCK:
> This is specifically no longer a goal of 7702

Interesting. In that case, should the point

> Once this is implemented, allowing EOAs to migrate permanently is "only one line of code": just add a flag to not set the code back to empty at the end.

be removed from the [Forward-compatibility](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md#forward-compatibility-with-future-account-abstraction) section?

NiloCK:
Closing - opened #8871.

PR #8871, opened 2024-09-06
Update EIP-7702: Remove reference to post-tx code deletion

NiloCK:
Prior versions of the EIP called for deleting the set code at the end of the transaction. This PR removes some dangling references to that deletion. 

@lightclient - following up from your comment on #8869.


lightclient (approved):
Thank you

PR #8873, opened 2024-09-08
Update EIP-7623: homogenize calculation examples

MariusVanDerWijden:
This orders the clauses in the calculation s.th. it matches the other calculation better


nerolation (approved):
yeah, better.

PR #8872, opened 2024-09-06
Update EIP-6404: Move to Draft

etan-status:
EIP-6493 mixes the introduction of SSZ transactions converted from RLP with native SSZ transactions, implying that the corresponding Receipt definition and engine API changes also must be part of EIP-6493.

Splitting the RLP converted transactions into the other SSZ transaction EIP-6404 achieves a clearer balance, and is also closer to practical implementation order where an EL initially may choose to support only RLP converted transactions before adding the native SSZ profiles.

PR #8877, opened 2024-09-12
Update EIP-6465: Move to Draft

etan-status:
To unstuck #8872, have to move this EIP to draft so that EIP-6493 which this refers to can be edited.


etan-status:
Done as part of #8872.

Mikhai56 (approved):

PR #8884, opened 2024-09-16
Add EIP: SSZ Transaction / Receipt proofs

etan-status:


Mikhai56:
error[preamble-req]: preamble is missing header(s): `eip`

abcoathup:
@etan-status my understanding is that API standards should be directed to https://github.com/ethereum/execution-apis rather than create an EIP/ERC

etan-status:
OK, can move it there. What is the EIP Interface section used for?

abcoathup:
> What is the EIP Interface section used for?

I don't think it is used for much any more, since ERCs were split off and APIs were directed to execution-apis.  (Note: I am not an EIP editor).

g11tech:
> > What is the EIP Interface section used for?
> 
> I don't think it is used for much any more, since ERCs were split off and APIs were directed to execution-apis. (Note: I am not an EIP editor).

makes sense, execution apis is more reasonable place :+1:

PR #8886, opened 2024-09-16
Update EIP-6404: Rename `from` signature fields

etan-status:
Avoid `signature.ecdsa_signature` in usage code (`signature` twice) and allow switching to different ECDSA curves by renaming:

- `tx.signature.from_` --> `tx.from_.address`
- `tx.signature.ecdsa_signature` --> `tx.from_.secp256k1_signature`


etan-status:
@eth-bot rerun

PR #8891, opened 2024-09-18
Add EIP: Standard Test Addresses for Development Frameworks

CedarMist:
This proposed Informational EIP standardizes the well known test addresses used by many frameworks and developers.

Hopefully it should spur discussion around clearly demarcating these addresses where UI/UX confusion is possible.


abcoathup:
@CedarMist this should be an ERC rather than an EIP.  Suggest opening a PR in https://github.com/ethereum/ERCs

CedarMist:
Thanks, will move to ERCs

PR #8895, opened 2024-09-20
Add EIP: Partial Gas Sponsorship Interface

limyeechern:


abcoathup:
@limyeechern this should be an ERC rather than an EIP. Suggest opening a PR in https://github.com/ethereum/ERCs

PR #8896, opened 2024-09-20
EIP-4762: tweaks to support 7702

gballet:
This is a proposed tweak to the stateless gas costs EIP in order to take into account the extra cases introduced in EIP 4762.

### DIFF
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -54,6 +54,14 @@ process this access event:
 
 Note: a non-value-bearing `SELFDESTRUCT` or `*CALL`, targetting a precompile or a system contract, will not cause the `BASIC_DATA_LEAF_KEY` to be added to the witness.
 
+If a `*CALL` or `SELFDESTRUCT`, `EXTCODESIZE`, `EXTCODEHASH` or `EXTCODECOPY` targetting a non-precompile, non-system contract, also process the access event:
+
+```
+(address, 0, CODEHASH_LEAF_KEY)
+```
+
+Note: this latter access is mean to check whether the contract is a [EIP-7702](./eip-7702.md) contract, or an EOF contract.
+
 If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), whether or not the `callee` is a precompile or a system contract, process this additional access event:
 
 ```
@@ -62,10 +70,10 @@ If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei), 
 
 Note: when checking for the existence of the `callee`, the existence check is done by validating that there is an extension-and-suffix tree at the corresponding stem, and does not rely on `CODEHASH_LEAF_KEY`.
 
-When calling `EXTCODEHASH`, process the access event:
+When calling `*CALL`, `SELFDESTRUCT`, `EXTCODESIZE`, `EXTCODECOPY` or `EXTCODEHASH` on an EIP-7702 delegating contract, process the additional access event:
 
 ```
-(address, 0, CODEHASH_LEAF_KEY)
+(delegated_address, 0, CODEHASH_LEAF_KEY)
 ```
 
 Note that precompiles and system contracts are excluded, as their hashes are known to the client.
@@ -108,6 +116,8 @@ In the conditions below, “chunk chunk_id is accessed” is understood to mean 
      * Example 3: for a `CODECOPY` with start position 1500, read size 2000, `code_size = 3100`, `x = 1500` and `y = 3099`
  * `CODESIZE`, `EXTCODESIZE` and `EXTCODEHASH` do NOT access any chunks.
     When a contract is created, access chunks `0 ... (len(code)+30)//31`
+
+In the case of an EIP-7702 delegating contract, replace `address` with `delegated_address`.
 
 ### Write Events
 
@@ -186,6 +196,29 @@ If `value` is non-zero:
 (tx.target, 0, BASIC_DATA_LEAF_KEY)
 ```
 
+#### Setcode transactions
+
+For EIP-7702 transactions, for each `[[chain_id, address, nonce, y_parity, r, s], ...]` entry in the authorization list, process the additional access events:
+
+```
+(address, 0, BASIC_DATA_LEAF_KEY)
+(address, 0, CODEHASH_LEAF_KEY)
+(address, 0, CODE_OFFSET)
+```
+
+If the authorization is valid, process the following write events:
+
+```
+(address, 0, BASIC_DATA_LEAF_KEY)
+(address, 0, CODE_OFFSET)
+```
+
+If `address` is not present in the state, also process the write event:
+
+```
+(address, 0, CODEHASH_LEAF_KEY)
+```
+
 ### Witness gas costs
 
 Remove the following gas costs:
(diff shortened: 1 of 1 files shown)


EIPS/eip-4762.md
```diff
@@ -186,13 +196,37 @@ If `value` is non-zero:
...
+```
+
 ### Witness gas costs
 
 Remove the following gas costs:
 
  * Increased gas cost of `CALL` if it is nonzero-value-sending
  * [EIP-2200](./eip-2200.md) `SSTORE` gas costs except for the `SLOAD_GAS`
  * 200 per byte contract code cost
+ * `PER_EMPTY_ACCOUNT_COST` and `PER_AUTH_BASE_COST`
```
gballet, line 229: This is potentially a controversial move, but since we have the stateless costs to cover for the writes to the state, that should match. Currently, they are covered by the `21000` intrinsic gas. Are these extra costs related to the transaction itself, or the state being written?

 - If it's the state being written, then I think we should drop them in verkle
 - if it's to disincentivize growing the transaction payload, then this line needs to be removed.

EIPS/eip-4762.md
```diff
@@ -54,6 +54,14 @@ process this access event:
 
 Note: a non-value-bearing `SELFDESTRUCT` or `*CALL`, targetting a precompile or a system contract, will not cause the `BASIC_DATA_LEAF_KEY` to be added to the witness.
 
+If a `*CALL` or `SELFDESTRUCT`, `EXTCODESIZE`, `EXTCODEHASH` or `EXTCODECOPY` targetting a non-precompile, non-system contract, also process the access event:
+
+```
+(address, 0, CODEHASH_LEAF_KEY)
+```
+
+Note: this latter access is mean to check whether the contract is a [EIP-7702](./eip-7702.md) contract, or an EOF contract.
```
gballet, line 63: So this is the weird bit, I am using the code hash to determine if it's a 7702 contract. Come to think of it, it should be reading the first code chunk directly because then you will also know what is the delegation address.

So yeah, the question is: apart from `EXTCODEHASH`, is there any reason to read the code hash for all these extra instructions?

jsign: Yes, I believe we should ask to read the first code-chunk to check if there's a delegation. Also, it might worth clarifying (as 7702 does) that we shouldn't follow a chain of delegations.

> So yeah, the question is: apart from EXTCODEHASH, is there any reason to read the code hash for all these extra instructions?

I don't think so.

EIPS/eip-4762.md
```diff
@@ -62,10 +70,10 @@ If a `*CALL` or `SELFDESTRUCT` is value-bearing (ie. it transfers nonzero wei),
 
 Note: when checking for the existence of the `callee`, the existence check is done by validating that there is an extension-and-suffix tree at the corresponding stem, and does not rely on `CODEHASH_LEAF_KEY`.
 
-When calling `EXTCODEHASH`, process the access event:
+When calling `*CALL`, `SELFDESTRUCT`, `EXTCODESIZE`, `EXTCODECOPY` or `EXTCODEHASH` on an EIP-7702 delegating contract, process the additional access event:
 
 ```
-(address, 0, CODEHASH_LEAF_KEY)
+(delegated_address, 0, CODEHASH_LEAF_KEY)
```
jsign, line 76: I'm a bit confused by this change.
For example, if you're executing an `EXTCODESIZE` as mentioned above:
- We should provide the first-code chunk to allow a stateless client to detect if there's a delegation and setup the right context of the target.
- If there is a delegation, we'll include the `BASIC_DATA` of the delegated address which contains the code size.

Why there is a need for this `CODEHASH_LEAF_KEY` of the delegated address?

jsign: More generally, I think we should explain impact on the witness for the delegation resolving step. Then that might change the context of which address is used for the EXTCODEHASH/EXTCODESIZE/etc and simply apply already existing rules to include the witness for that address (which now can be `address` or `delegated_address`)

EIPS/eip-4762.md
```diff
@@ -186,13 +196,37 @@ If `value` is non-zero:
...
 ```
 
+#### Setcode transactions
+
+For EIP-7702 transactions, for each `[[chain_id, address, nonce, y_parity, r, s], ...]` entry in the authorization list, process the additional access events:
+
+```
+(address, 0, BASIC_DATA_LEAF_KEY)
+(address, 0, CODEHASH_LEAF_KEY)
+(address, 0, CODE_OFFSET)
```
jsign, line 206: What is `CODE_OFFSET`?

gballet: `CODE_OFFSET` is defined in eip-6800 at `128`: it's the start of the code.

gballet: i.e.  the first code chunk is being read

beage666:
What dose this mean

Aces42020 (approved):

Aces42020 (approved):

EIPS/eip-4762.md
beage666: Ok

beage666 (approved):

EIPS/eip-4762.md
beage666: bc1qdfs6vkttw7fj7t4m73humszxmgeq78a8we4ncf that's my address wallet

Scamreno:
Okay 👍

EIPS/eip-4762.md
Scamreno: Thanks

tanishqjasoria, line 206: this does not look right to me, it should be `authority`, insted of `address`

EIPS/eip-4762.md
```diff
@@ -186,13 +196,37 @@ If `value` is non-zero:
...
+```
+
+If the authorization is valid, process the following write events:
+
+```
+(address, 0, BASIC_DATA_LEAF_KEY)
+(address, 0, CODE_OFFSET)
+```
+
+If `address` is not present in the state, also process the write event:
```
tanishqjasoria, line 216: we should always process the write event 
`(authority, 0, CODEHASH_LEAF_KEY)`

EIPS/eip-4762.md
tanishqjasoria: also, its done is a different way and i dont understand the reason behind that
we charge the `PER_EMPTY_ACCOUNT_COST` always and then refund `PER_AUTH_BASE_COST` if the account exists
looking at how the gas costs are implemented in this EIP. I think we should keep the gas costs as it is and deal with them in the way we deal with intrisic gas costs for transactions right now

EIPS/eip-4762.md
```diff
@@ -186,13 +196,37 @@ If `value` is non-zero:
...
+(address, 0, BASIC_DATA_LEAF_KEY)
+(address, 0, CODEHASH_LEAF_KEY)
+(address, 0, CODE_OFFSET)
+```
+
+If the authorization is valid, process the following write events:
+
+```
+(address, 0, BASIC_DATA_LEAF_KEY)
+(address, 0, CODE_OFFSET)
```
tanishqjasoria, line 213:
```suggestion
(authority, 0, BASIC_DATA_LEAF_KEY)
(authority, 0, CODE_OFFSET)
(authority, 0, CODEHASH_LEAF_KEY)
```

EIPS/eip-4762.md
tanishqjasoria, line 206:
```suggestion
(authority, 0, BASIC_DATA_LEAF_KEY)
(authority, 0, CODEHASH_LEAF_KEY)
(authority, 0, CODE_OFFSET)
```

beage666: 😃

Skippyy9 (approved):

daniela2000111 (approved):

PR #8897, opened 2024-09-23
Update EIP-6404: Add support for SetCode transactions

etan-status:
Update EIP-6404: Add support for SetCode transactions

EIP-7702 adds SetCode transactions, adding another RLP format that now
needs conversion into SSZ.

`ChainId` 0 is reserved for authorizations to indicate chain-agnostic
authorizations. To reflect that special meaning, the SSZ type uses
`None` in authorizations to match the encoding of replayable legacy tx.

Further, `SSZ_TX_TYPE` is changed to `0x1f` to avoid clashing with the
new envelope type `0x04` that is used for SetCode transactions.

- `0x05` is blocked because it's used as the signature domain for
  EIP-7702 authorizations
- `0x19` is blocked because it's used as the signature domain for
  general non-transaction messages
- Using a value above `0x19` avoids clashing with any RLP tx types.
- Using a value below `0x20` avoids giving bits `0x60` special meaning.

A commitment to `magic` is stored in the SSZ authorization object to
remember whether an authorization was originally made in RLP or a future
scheme.


g11tech (dismissed):
lgtm

Aces42020 (approved):

etan-status:
@eth-bot rerun

lucentlabz (commented):
👍

PR #8905, opened 2024-09-25
Update EIP-7702: Do not allow authorization nonce equal to 2**64 - 1

gumb0:
It is important to not allow `authorization.nonce == 2**64 - 1`, because in this case incrementing authority nonce will result in nonce overflowing 64 bit, which is not allowed since [EIP-2681](https://eips.ethereum.org/EIPS/eip-2681)

~~Alternative to adding this to the transaction validity rules could be additional check during delegation setting, but it is simpler this way (fewer checks) and more aligned with existing transaction types validity.~~ It is checked at authorization list processing now.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -67,11 +67,11 @@ The [EIP-2718](./eip-2718.md) `ReceiptPayload` for this transaction is `rlp([sta
 At the start of executing the transaction, after incrementing the sender's nonce, for each `[chain_id, address, nonce, y_parity, r, s]` tuple do the following:
 
 1. Verify the chain id is either 0 or the chain's current ID.
-2. `authority = ecrecover(keccak(MAGIC || rlp([chain_id, address, nonce])), y_parity, r, s]`
-3. Add `authority` to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
-4. Verify the code of `authority` is either empty or already delegated.
-5. Verify the nonce of `authority` is equal to `nonce`.
-6. Verify the nonce of `authority` is less than `2**64 - 1`.
+2. Verify the `nonce` is less than `2**64 - 1`.
+3. `authority = ecrecover(keccak(MAGIC || rlp([chain_id, address, nonce])), y_parity, r, s]`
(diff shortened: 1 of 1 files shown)


lightclient:
> but it is simpler this way (fewer checks) and more aligned with existing transaction types validity.

Is this true? I feel like add bounds outside of basic type widths might add more complexity since it isn't natively part of decoding (e.g. value too big for type!).

gumb0:
> > but it is simpler this way (fewer checks) and more aligned with existing transaction types validity.
> 
> Is this true? I feel like add bounds outside of basic type widths might add more complexity since it isn't natively part of decoding (e.g. value too big for type!).

I guess you're right, it's probably the same number of checks, so either extra check at transaction validity, or one at processing authorizations. I don't mind it either way.

gumb0:
I changed it now to be checked at authorization list processing.

EIPS/eip-7702.md
```diff
@@ -71,9 +71,10 @@ At the start of executing the transaction, after incrementing the sender's nonce
 3. Add `authority` to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
 4. Verify the code of `authority` is either empty or already delegated.
 5. Verify the nonce of `authority` is equal to `nonce`.
-6. Add `PER_EMPTY_ACCOUNT_COST - PER_AUTH_BASE_COST` gas to the global refund counter if `authority` exists in the trie.
-7. Set the code of `authority` to be `0xef0100 || address`. This is a delegation designation.
-8. Increase the nonce of `authority` by one.
+6. Verify the nonce of `authority` is less than `2**64 - 1`.
```
jochem-brouwer, line 74: I think we should do this check earlier for optimization. We can check that `nonce < 2**64-1` immediately. If this is the nonce reported in the tuple field then we can immediately skip this authorization. Note that if the nonce of the `authority` is `2**64-1` and the nonce field itself is not, then rule 5 will fail. I suggest we put this after rule 1.

This also optimizes a lookup in the state database, since we now don't have to pull `authority` out of the state.

(Also, thanks for opening this PR :smile: )

gumb0: Sounds good, moved it to step 2.

lightclient:
LGTM. Needs a rebase - I just did in in #8938. Closing

PR #8906, opened 2024-09-25
Update EIP-7702: add several clarifications to align spec with tests

gumb0:
cc @jochem-brouwer

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -89,6 +89,8 @@ For example, `EXTCODESIZE` would return the size of the code pointed to by `addr
 
 In case a delegation designator points to a precompile address, retrieved code is considered empty and `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL` instructions targeting this account will execute empty code, i.e. succeed with no execution given enough gas.
 
+`EXTCODEHASH` instruction in line with [EIP-1052](./eip-1052.md) puts `0` on stack if delegation designation target does not exist in the trie, or is empty as defined by [EIP-161](./eip-161.md). Emptiness is determined using nonce, balance and code of delegation designation target account. In case target is not empty, but has empty code, keccak256 hash of empty data is put on stack.
+
 In case a delegation designator points to another designator, creating a potential chain or loop of designators, clients must retrieve only the first code and then stop following the designator chain.
 
 #### Gas Costs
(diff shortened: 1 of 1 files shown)


EIPS/eip-7702.md
```diff
@@ -79,12 +79,18 @@ If any of the above steps fail, immediately stop processing that tuple and conti
...
+The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541.md) to designate the code has a special purpose. This designator requires all code retrieving operations follow the address pointer to fill the accounts observable code. The following instructions are impacted: `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, as well as transactions with `destination` targeting the code with delegation designation. 
 
 For example, `EXTCODESIZE` would return the size of the code pointed to by `address` instead of `23` which would represent the delegation designation. `CALL` would similarly load the code from `address` and execute it in the context of `authority`.
 
+In case a delegation designator points to a precompile address, retrieved code is considered empty and `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL` instructions targeting this account will execute empty code, i.e. succeed with no execution given enough gas.
```
gumb0, line 90: Personally would like to change this rule to the opposite, i.e. executing precompiles, but this is what tests currently expect.

lightclient: What would be the argument for doing this? IMO that conflates the trie structure and it's values with the convenience feature of the EVM of having "precompiles" at specific addresses.

gumb0: Argument would be being consistent in how if execution is targeting certain address, it behaves the same regardless if it was 7702-delegated.

I.e. currently there is a special case "If execution targets a precompile, don't execute it if target was delegated to".

lightclient: Hmm I just don't see 7702 operating at that level. To me it's more about reading the trie and setting certain values in the trie. It's then, separately, the job of the EVM to execute the code it's provided.

gumb0: I guess it depends on where the boundary between EVM and the rest of the client lies in the particular implementation. If EVM is in charge of resolving precompile addresses, it's natural to do the same resolution for 7702 case. If precompile resolution is outside of EVM and EVM gets only code, then it could be more natural to give it empty code in this case.

Allzz77123 (approved):

lightclient (approved):
thank you

lightclient:
shoot now this has merge conflicts

lightclient:
replaced with #8940 so i could resolve the merge conflicts. thanks for this

PR #8908, opened 2024-09-26
Update EIP-7685: remove requests from block body

lightclient:
This PR removes the requests from the block body. This change is motivated by a simplification to the engine api: https://github.com/ethereum/execution-apis/pull/591

>  Make EL validate requestsHash as a part of blockHash validation, note there is no need to keep requests_hash in the beacon block. CLs that run blockHash validation on their own will still be able to do that by implementing the algo EL uses to compute requestsHash
Make EL validate requestsHash in the EL block header against requests obtained from transaction execution

### DIFF
--- a/EIPS/eip-7685.md
+++ b/EIPS/eip-7685.md
@@ -55,7 +55,7 @@ Extend the header with a new 32 byte value `requests_hash`:
 def compute_requests_hash(list):
     return keccak256(rlp.encode([rlp.encode(req) for req in list]))
 
-block.header.requests_root = compute_requests_hash(requests)
+block.header.requests_hash = compute_requests_hash(requests)
 ```
 
 ### Consensus Layer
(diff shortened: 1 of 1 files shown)


EIPS/eip-7685.md
```diff
@@ -70,7 +55,7 @@ Extend the header with a new 32 byte value `requests_hash`:
 def compute_requests_hash(list):
     return keccak256(rlp.encode([rlp.encode(req) for req in list]))
 
-block.header.requests_root = compute_requests_hash(block.body.requests)
+block.header.requests_root = compute_requests_hash(requests)
```
marioevz, line 58:
```suggestion
block.header.requests_hash = compute_requests_hash(requests)
```
There's another use of `"requests_root"` in the "Ordering" section below, I think we should take the opportunity to update that too.

marioevz (commented):
One small comment.

EIPS/eip-7685.md
lightclient: fixed. tysm

lightclient:
superseded by https://github.com/ethereum/EIPs/pull/8924

PR #8910, opened 2024-09-26
Update EIP-2537: double each discount table value in the MSM gas cost model

jwasinger:
Based on discussion during ACDE, we are repricing the bls MSM precompiles to be twice as expensive.


jwasinger:
> Maybe you did discusse it but it didn't make any sense now: up to 7 inputs the cost is higher than computing the multiplication one by one.

The rationale from my side is: doubling the entries in the discount table is the simplest repricing we can do such that the pricing of g1 msm is at least that of the ecrecover precompile in the worst case (on Geth, concurrency disabled for MSM, benchmarked across several machines - x86/arm).

I need to do some further research to come up with a justification for the reason that MSM is higher cost than naive method (muls) up to a threshold.  From what I know right now, there is some setup overhead with MSM that makes it perform worse than muls for small input sizes.

For example,  here is the ratio of the runtime for the naive method vs the runtime of gnark's g1 msm on my M2 Macbook Pro:

```
1 pairs - 2.451
2 pairs - 1.535
3 pairs - 1.239
4 pairs - 1.07
5 pairs - 0.996
6 pairs - 0.942
7 pairs - 0.878
```

Mikhai56 (approved):

PR #8914, opened 2024-09-30
Add EIP: BURN opcode

itsdevbear:

--

 - The build passes.

### DIFF
--- a/EIPS/eip-7775.md
+++ b/EIPS/eip-7775.md
@@ -102,11 +102,11 @@ This EIP introduces a new opcode and thus must be activated via a scheduled hard
 
 ## Test Cases
 
-- Coming soon
+<!-- TODO -->
 
 ## Reference Implementation
 
-- Coming Soon
+<!-- TODO -->
 
 ## Security Considerations
 
--- a/EIPS/eip-7775.md
+++ b/EIPS/eip-7775.md
@@ -48,15 +48,15 @@ def op_burn(pc, interpreter, scope):
     
     # If the value to be burned is greater than the balance, revert
     if value_to_burn > balance:
-        return None, "ErrInsufficientBalance"
+        return "ErrInsufficientBalance"
     
     # If the value to be burned is 0, do not revert
     if value_to_burn == 0:
-        return None, None
+        return None
     
     # Subtract the value from the current address's balance
     interpreter.evm.state_db.sub_balance(current_address, value_to_burn)
-    return None, None
+    return None
 ```
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
--- a/EIPS/eip-7775.md
+++ b/EIPS/eip-7775.md
@@ -22,6 +22,8 @@ The motivation for this proposal is to provide a standardized and efficient way 
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
+### Behaviour
+
 The `BURN` opcode `(0xFC)` is introduced with the following behavior:
 
 1. Pops one 32-byte word from the top of the stack, treating these bytes as the `uint256` amount of native ether to be burned.
@@ -35,12 +37,24 @@ When `BURN` is used in the context of a DELEGATECALL or CALLCODE, the contract w
 
 When `BURN` is used in the context of a STATICCALL the call `MUST` revert.
 
-When the `BURN` opcode is processed, 0 gas MUST be consumed.
+### Gas Cost
+
+The base gas cost for the `BURN` opcode is 100 gas. The dynamic gas cost is determined as follows:
+
+1. If the value to be burned is 0, the dynamic gas cost is 0.
+2. Otherwise, the dynamic gas cost is 100.
+
+The total gas cost for the `BURN` opcode is the sum of the base gas cost and the dynamic gas cost.
+
+### Pseudocode
 
 Example pseudocode for the `BURN` opcode:
 
 ```python
 def op_burn(pc, interpreter, scope):
+    # Consume the base gas cost
+    interpreter.consume_gas(100)
+    
     # Pop the value to be burned from the stack
     value_to_burn = scope.stack.pop()
 
@@ -60,6 +74,9 @@ def op_burn(pc, interpreter, scope):
 
     # Subtract the value from the current address's balance
     interpreter.evm.state_db.sub_balance(current_address, value_to_burn)
+   
+    # Consume the dynamic gas cost
+    interpreter.consume_gas(100)
     return None
 ```
 
--- a/EIPS/eip-7775.md
+++ b/EIPS/eip-7775.md
@@ -42,7 +42,8 @@ When `BURN` is used in the context of a STATICCALL the call `MUST` revert.
 The base gas cost for the `BURN` opcode is 100 gas. The dynamic gas cost is determined as follows:
 
 1. If the value to be burned is 0, the dynamic gas cost is 0.
-2. Otherwise, the dynamic gas cost is 100.
+2. If the account doesn't exist, or the balance of the account is 0, the dynamic gas cost is 0.
+3. Otherwise, the dynamic gas cost is 2800.
 
 The total gas cost for the `BURN` opcode is the sum of the base gas cost and the dynamic gas cost.
 
@@ -71,12 +72,16 @@ def op_burn(pc, interpreter, scope):
     # If the value to be burned is greater than the balance, revert
     if value_to_burn > balance:
         return "ErrInsufficientBalance"
+    
+    # If the account balance is 0, return.
+    if balance == 0:
+        return None
 
     # Subtract the value from the current address's balance
     interpreter.evm.state_db.sub_balance(current_address, value_to_burn)
-   
-    # Consume the dynamic gas cost
-    interpreter.consume_gas(100)
+    
+    # The account is known to exist at this point, thus we consume "warm" gas.
+    interpreter.consume_gas(2800)
     return None
 ```
 
--- a/EIPS/eip-7775.md
+++ b/EIPS/eip-7775.md
@@ -16,7 +16,7 @@ This proposal introduces a new `BURN` opcode, designed to burn native tokens at 
 
 ## Motivation
 
-The motivation for this proposal is to provide a standardized and efficient way to burn native tokens directly within the EVM. Historically, contracts such as the BeaconDepositContract have "burned" ether by making it irrecoverable from the given address. This approach can lead to confusion and potential misuse. By introducing a dedicated `BURN` opcode, we can ensure a clear and consistent method for burning native tokens. This could become useful for Ethereum L2s when transferring ether back to the L1, as well as other EVM L1 chains that could leverage this for their cryptoeconomics. Additionally, it opens up an interesting design space for EIP-7685 use cases, providing a more robust and standardized mechanism for token management across different layers of the Ethereum ecosystem.
+The motivation for this proposal is to provide a standardized and efficient way to burn native tokens directly within the EVM. Historically, contracts such as the BeaconDepositContract have "burned" ether by making it irrecoverable from the given address. This approach can lead to confusion and potential misuse. By introducing a dedicated `BURN` opcode, we can ensure a clear and consistent method for burning native tokens. This could become useful for Ethereum L2s when transferring ether back to the L1, as well as other EVM L1 chains that could leverage this for their cryptoeconomics. Additionally, it opens up an interesting design space for [EIP-7685](./eip-7685.md) use cases, providing a more robust and standardized mechanism for token management.
 
 ## Specification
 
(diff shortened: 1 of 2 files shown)


EIPS/eip-7xxx.md
```diff
@@ -0,0 +1,101 @@
+---
+eip: 7xxx
+title: New opcode BURN
+description: An opcode to burn native token at the given address
+author: Dev Bear (@itsdevbear)
+discussions-to: <URL>
```
abcoathup, line 6: Please create a discussions topic on Eth Magicians using the template: https://ethereum-magicians.org/c/eips/5

itsdevbear:
@abcoathup anyone I can contact regarding this error, I assume its because my account is relatively new.
image

EIPS/eip-7775.md
```diff
@@ -0,0 +1,100 @@
+---
+eip: 7775
+title: New opcode BURN
+description: An opcode to burn native token at the given address
+author: Dev Bear (@itsdevbear)
+discussions-to: <URL>
```
abcoathup, line 6: Please create a topic on Eth Magicians using the template: https://ethereum-magicians.org/c/eips/5

EIPS/eip-7775.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Specification
+
+The `BURN` opcode `(0xF8)` is introduced with the following behavior:
+
+1. Pops one 32-byte word from the top of the stack, treating this value as the `uint256` value of native token to be burned.
+2. Retrieves the current address from the EVM execution context.
+3. Checks the balance of the current address.
+4. If the value to be burned is greater than the balance of the current address, the opcode `MUST` revert.
+5. If the value to be burned is 0, the execution `MUST NOT` revert.
```
jochem-brouwer, line 29: `revert` when used in EVM, does not cost any gas (except gas for memory expansion). However, INVALID consumes all gas in the current context. What should it be here? It is clear that the call frame must be aborted, but not clear how much gas should be consumed. In case of a CALL opcode where the `value` is higher than the current balance, the call frame is also aborted and all gas is consumed.

EIPS/eip-7775.md
```diff
@@ -0,0 +1,100 @@
...
+
+This proposal introduces a new `BURN` opcode, designed to burn native tokens at the given address.
+
+## Motivation
+
+The motivation for this proposal is to provide a standardized and efficient way to burn native tokens directly within the EVM. Historically, contracts such as the BeaconDepositContract have "burned" ether by making it irrecoverable from the given address. This approach can lead to confusion and potential misuse. By introducing a dedicated `BURN` opcode, we can ensure a clear and consistent method for burning native tokens. This could become useful for Ethereum L2s when transferring ether back to the L1, as well as other EVM L1 chains that could leverage this for their cryptoeconomics. Additionally, it opens up an interesting design space for [EIP-7685](./eip-7685.md) use cases, providing a more robust and standardized mechanism for token management.
+
+## Specification
+
+The `BURN` opcode `(0xF8)` is introduced with the following behavior:
```
jochem-brouwer, line 23: Note: `0xF8` is used in EOF contracts for `EXTCALL`. It is likely this opcode should also be in EOF contracts? If that is the case, then I think it is better to use an opcode which is not used by EOF.

EIPS/eip-7775.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. Pops one 32-byte word from the top of the stack, treating this value as the `uint256` value of native token to be burned.
+2. Retrieves the current address from the EVM execution context.
+3. Checks the balance of the current address.
+4. If the value to be burned is greater than the balance of the current address, the opcode `MUST` revert.
+5. If the value to be burned is 0, the execution `MUST NOT` revert.
+6. Subtracts the value from the current address's native token balance.
+
+When `BURN` is used in the context of a DELEGATECALL or CALLCODE, the contract whose balance is to be manipulated is the contract that issued the DELEGATECALL or CALLCODE instruction.
```
jochem-brouwer, line 32: I think a note here that it is not possible to use `BURN` in a context where `STATICCALL` is used should be added.

EIPS/eip-7775.md
itsdevbear: @abcoathup when I try to post:

image


any suggestions?

abcoathup: Not sure why you can't access Eth Magicians: https://ethereum-magicians.org

Can you try again. If you haven't already created an account you will need to read for a little while before you can create a new topic.

EIPS/eip-7775.md
itsdevbear: changed terminology to `amount`

EIPS/eip-7775.md
```diff
@@ -0,0 +1,102 @@
...
+## Specification
+
+The `BURN` opcode `(0xFC)` is introduced with the following behavior:
+
+1. Pops one 32-byte word from the top of the stack, treating these bytes as the `uint256` amount of native token to be burned.
+2. Retrieves the current address from the EVM execution context.
+3. Checks the balance of the current address.
+4. If the amount to be burned is greater than the balance of the current address, the opcode `MUST` revert.
+5. If the amount to be burned is 0, the execution `MUST NOT` revert.
+6. Subtracts the amount from the current address's native token balance.
```
jochem-brouwer, line 30: The gas cost of this opcode is not specified, and this should be specified.

EIPS/eip-7775.md
```diff
@@ -0,0 +1,102 @@
+---
+eip: 7775
+title: BURN opcode
+description: An opcode to burn native token at the given address
```
jochem-brouwer, line 4: I find the word "token" here confusing. This can easily be confused with an ERC20 token. As I understand it, the "native cryptocurrency" which the evm uses (so for Ethereum: ETH) is used as "native token" here.

EIPS/eip-7775.md
```diff
@@ -0,0 +1,102 @@
...
+created: 2024-09-30
+---
+
+## Abstract
+
+This proposal introduces a new `BURN` opcode, designed to burn native tokens at the given address.
+
+## Motivation
+
+The motivation for this proposal is to provide a standardized and efficient way to burn native tokens directly within the EVM. Historically, contracts such as the BeaconDepositContract have "burned" ether by making it irrecoverable from the given address. This approach can lead to confusion and potential misuse. By introducing a dedicated `BURN` opcode, we can ensure a clear and consistent method for burning native tokens. This could become useful for Ethereum L2s when transferring ether back to the L1, as well as other EVM L1 chains that could leverage this for their cryptoeconomics. Additionally, it opens up an interesting design space for [EIP-7685](./eip-7685.md) use cases, providing a more robust and standardized mechanism for token management.
```
jochem-brouwer, line 19: I do not understand what EIP 7685 here has to do with it. What request is triggered by the EL here which should be forwarded to the CL?

EIPS/eip-7775.md
```diff
@@ -0,0 +1,102 @@
...
+    # If the value to be burned is 0, do not revert
+    if value_to_burn == 0:
+        return None
+    
+    # Subtract the value from the current address's balance
+    interpreter.evm.state_db.sub_balance(current_address, value_to_burn)
+    return None
+```
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```
jochem-brouwer, line 62: For clarity, this sentence should be moved above the first uses of any of these keywords.

EIPS/eip-7775.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Rationale
+
+The introduction of the `BURN` opcode helps clean up a piece of weird semantics in the Ethereum. Historically, burning native tokens involved sending them to an address from which they could not be recovered, such as the zero address or a contract with no withdrawal functionality. This method is not only inefficient but also confusing for indexers and other tools that track token movements. By providing a dedicated `BURN` opcode, we eliminate this ambiguity and ensure that the act of burning tokens is explicit and standardized.
```
jochem-brouwer, line 66: I don't understand this sentence, it talks about native tokens. I interpret the native token to be Ether in case of the Ethereum chain. I am not aware of an Ethereum-like chain which has two or more tokens.

Note that it is not possible to create an opcode which burns ERC20 tokens.

EIPS/eip-7775.md
```diff
@@ -0,0 +1,102 @@
...
+
+- Does not help remove unrecoverable ether sitting in existing contracts.
+- New code in the clients
+- New concept needed to be added to the yellow paper.
+
+## Backwards Compatibility
+
+No backwards compatibility issues.
+
+No backward compatibility issues found.
```
jochem-brouwer, line 84: This is not the case, this EIP introduces a new opcode and thus changes the consensus, and therefore needs a hardfork of the chain.

jochem-brouwer (changes requested):

EIPS/eip-7775.md
itsdevbear: yup you got it

EIPS/eip-7775.md
itsdevbear: 👍

EIPS/eip-7775.md
itsdevbear: Removed, I was more giving an example of what could be an interesting usecase. The `BURN` opcode being utilized within a SystemCall, which could utilize 7685 to relay info to the CL.

EIPS/eip-7775.md
```diff
@@ -12,62 +12,38 @@ created: 2024-09-30
...
-    if value_to_burn == 0:
-        return None
-    
-    # Subtract the value from the current address's balance
-    interpreter.evm.state_db.sub_balance(current_address, value_to_burn)
-    return None
-```
-
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+When the `BURN` opcode is processed, 0 gas MUST be consumed.
```
jochem-brouwer, line 38: This is definitely too low. The reason is that this operation does a state change: it will change the balance of this contract. This is an expensive operation as it will hit the disk. In case of "just sending ether" (CALL opcode) at least 9100 gas has to be paid. The BURN operation here is simpler, because it will not try to invoke another contract, it will only change the balance of the current contract.

EIPS/eip-7775.md
```diff
@@ -12,62 +12,38 @@ created: 2024-09-30
...
 
 ## Rationale
 
-The introduction of the `BURN` opcode helps clean up a piece of weird semantics in the Ethereum. Historically, burning native tokens involved sending them to an address from which they could not be recovered, such as the zero address or a contract with no withdrawal functionality. This method is not only inefficient but also confusing for indexers and other tools that track token movements. By providing a dedicated `BURN` opcode, we eliminate this ambiguity and ensure that the act of burning tokens is explicit and standardized.
+The introduction of the `BURN` opcode helps clean up a piece of weird semantics in the Ethereum. Historically, burning native ether involved sending them to an address from which they could not be recovered, such as the zero address or a contract with no withdrawal functionality. This method is not only inefficient but also confusing for indexers and other tools that track token movements. By providing a dedicated `BURN` opcode, we eliminate this ambiguity and ensure that the act of burning tokens is explicit and standardized.
```
jochem-brouwer, line 68: (Informational): This is not entirely correct, in the past it was actually possible to burn ether. In order to do so, you had to SELFDESTRUCT to the current address. In https://eips.ethereum.org/EIPS/eip-6780 this was removed, see:

> Note that if the target is the same as the contract calling SELFDESTRUCT there is no net change in balances. Unlike the prior specification, Ether will not be burnt in this case.

(This info should not be added, but just wanted to note :) )

jochem-brouwer (changes requested):

EIPS/eip-7775.md
jochem-brouwer: Also note that simple operations like additions cost 3 gas, and this is an operation with much more complexity

itsdevbear: @jochem-brouwer what do you suggest, I was putting 0 as a placeholder for now? Is there a general philsophy for how gas costs are determined for new opcodes?

jochem-brouwer: I am not entirely aware how gas costs are decided on in the end, I think EIPs initially propose costs which are likely too high. These could then be adjusted downward later.

For instance see the gas spec of 1153 motivation:

> Gas cost for TSTORE is the same as a warm SSTORE of a dirty slot (i.e. original value is not new value and is not current value, currently 100 gas), and gas cost of TLOAD is the same as a hot SLOAD (value has been read before, currently 100 gas). Gas cost cannot be on par with memory access due to transient storage’s interactions with reverts.

( https://eips.ethereum.org/EIPS/eip-1153, specification section )

I would set 9100 gas here with the comment that this is the same as the (minimal) gas used for a CALL operation with value. Benchmarks/experiments could show that these costs are too high and that it is safe to set this to a lower number.

itsdevbear: @jochem-brouwer I added a simple algorithm, let me know what you think!

EIPS/eip-7775.md
itsdevbear: Very interesting!

EIPS/eip-7775.md
itsdevbear: kept it reasonable, and used SLOAD, SSTORE as a reference, since `BURN` would be effectively reading and writing 32 bytes.

EIPS/eip-7775.md
```diff
@@ -0,0 +1,119 @@
...
+- New code in the clients
+- New concept needed to be added to the yellow paper.
+
+## Backwards Compatibility
+
+This EIP introduces a new opcode and thus must be activated via a scheduled hardfork.
+
+## Test Cases
+
+- Coming soon
```
g11tech, line 105: put a html tag `<!-- TODO -->` which is tracked by the bots to make sure its addressed before going to review etc

itsdevbear: 👍

EIPS/eip-7775.md
jochem-brouwer: 100 is too low. Please check a gas schedule for state-changing operations, such as `CALL` or `SSTORE`, a good gas overview can be found on https://evm.codes

image

See for instance SSTORE changing 10000 -> 500. This costs 2900 gas. One could argue this would be equal to a contract having 10000 balance and now burning 9500. (So 2900 is reasonable here).

EIPS/eip-7775.md
```diff
@@ -0,0 +1,119 @@
...
+
+When `BURN` is used in the context of a DELEGATECALL or CALLCODE, the contract whose balance is to be manipulated is the contract that issued the DELEGATECALL or CALLCODE instruction.
+
+When `BURN` is used in the context of a STATICCALL the call `MUST` revert.
+
+### Gas Cost
+
+The base gas cost for the `BURN` opcode is 100 gas. The dynamic gas cost is determined as follows:
+
+1. If the value to be burned is 0, the dynamic gas cost is 0.
```
jochem-brouwer, line 44: 0 is also too low, also for a no-op. A no-op `JUMPDEST` costs 1 gas. Also note we have to pop one item from stack here. `3` would be reasonable here (same as the `EQ` operator, which pops two items and compares those)

itsdevbear: This is just the dynamic cost. The min cost of the code is always 100

itsdevbear: @jochem-brouwer thoughts?

jochem-brouwer: The base cost is too low, see: https://github.com/ethereum/EIPs/pull/8914#discussion_r1798660732

jochem-brouwer: Or wait, maybe dynamic should be jumped if the target is cold. But the cost is too low for cold accounts.

EIPS/eip-7775.md
itsdevbear: :+1:

EIPS/eip-7775.md
itsdevbear: @jochem-brouwer I think the dyn can be 0 if account is cold, since we can short circuit on balance. Updated to reflect. LMK what u think.

lightclient (approved):

PR #8916, opened 2024-09-30
Update EIP-7685: Update based on latest discussions

lucassaldanha:
Taking a stab at updating this EIP to capture the design decisions we made on https://github.com/ethereum/execution-apis/pull/591.

### DIFF
--- a/EIPS/eip-7685.md
+++ b/EIPS/eip-7685.md
@@ -14,8 +14,7 @@ created: 2024-04-14
 
 This proposal defines a general purpose framework for storing contract-triggered
 requests. It extends the execution header with a single field as a commitment to
-the request. Requests are later on exposed to the consensus layer via the
-Engine API.
+the request. Requests are later on exposed to the consensus layer.
 
 Requests are not stored on the execution block, only a commitment to the requests
 is stored in the execution block.
@@ -77,8 +76,8 @@ As part of the EL validation for a new block, the EL MUST check that the paramet
 
 Each proposal may choose how to extend the beacon chain types to include new EL request types.
 
-As part of beacon block validation, the CL MUST to calculate the hash of the requests
-in the beacon block body, and pass the calculated hash as the `executionRequestsHash`
+As part of beacon block validation, the CL MUST calculate the hash of the requests
+in the beacon block body, and pass the calculated hash to the EL to be incorporated into the `block_hash` validation process.
 parameter when calling `engine_newPayloadV4`.
 
 To calculate the hash of the requests, the CL appends the a byte representing the request
@@ -108,6 +107,7 @@ prepended by their type. They MUST be ordered by their first byte (the `request_
 order used in the EL.
 
 The execution_hash can be calculated as:
+
 ```
 requests_hash = sha256(requests)
 ```
--- a/EIPS/eip-7685.md
+++ b/EIPS/eip-7685.md
@@ -53,7 +53,7 @@ requests = request_data_0 ++ request_data_1 ++ ... ++ request_data_n
 ```
 
 Note: each `request_data` represents a different list of requests from a particular
-smart contract, prepended by their type. They must be ordered by their first byte
+smart contract, prepended by their type. They MUST be ordered by their first byte
 (the `request_type`).
 
 The ordering of requests within a type is to be defined by each request type and it
@@ -70,10 +70,47 @@ def compute_requests_hash(requests):
 block.header.requests_hash = compute_requests_hash(requests)
 ```
 
+As part of the EL validation for a new block, the EL MUST check that the parameter
+`executionRequestsHash` matches their commitment for that block (`block.header.requests_hash`).
+
 ### Consensus Layer
 
-Each proposal may choose how to extend the beacon chain types to include the new
-EL request.
+Each proposal may choose how to extend the beacon chain types to include new EL request types.
+
+As part of beacon block validation, the CL MUST to calculate the hash of the requests
+in the beacon block body, and pass the calculated hash as the `executionRequestsHash`
+parameter when calling `engine_newPayloadV4`.
+
+To calculate the hash of the requests, the CL appends the a byte representing the request
+type, followed by the SSZ encoded list of requests of that type, repeating this process for
+each one of the different request types the CL has on the `execution_requests` field of the
+beacon block body and following the same ordering rules used on the EL (ordered by `request_type`).
+
+Let `requests_list` be the list of requests of a type (one of many types from `beacon_block.body.execution_requests`).
+
+`request_data` consists of two parts: `request_type` (single byte) and
+`request_list_ssz` (ssz list encoding of `requests_list`) as following:
+
+```
+request_list_ssz = ssz(requests_list)
+request_data = request_type ++ request_list_ssz
+```
+
+Let `requests` be the concatenation of each `request_data` for each request type in
+of the requests on `beacon_block.body.execution_requests`:
+
+```
+requests = request_data_0 ++ request_data_1 ++ ... ++ request_data_n
+```
+
+Note: each `request_data` represents a different list of requests from `beacon_block.body.execution_requests`,
+prepended by their type. They MUST be ordered by their first byte (the `request_type`), matching the
+order used in the EL.
+
+The execution_hash can be calculated as:
+```
+requests_hash = sha256(requests)
+```
 
 ## Rationale
 
@@ -86,7 +123,7 @@ to use SSZ as requests are consumed by the CL.
 ### Request source and validity
 
 This EIP makes no strict requirement where a request may come from nor when/how
-a request must be validated. This is to provide future protocol designers
+a request MUST be validated. This is to provide future protocol designers
 maximum flexibility.
 
 The authors' recommendations on source and validity of requests are:
--- a/EIPS/eip-7685.md
+++ b/EIPS/eip-7685.md
@@ -13,9 +13,12 @@ created: 2024-04-14
 ## Abstract
 
 This proposal defines a general purpose framework for storing contract-triggered
-requests. It extends the execution header and body with a single field each to
-store the request information. This inherently exposes the requests to the
-consensus layer, which can then process each one.
+requests. It extends the execution header with a single field as a commitment to
+the request. Requests are later on exposed to the consensus layer via the
+Engine API.
+
+Requests are not stored on the execution block, only a commitment to the requests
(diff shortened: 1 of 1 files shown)


lucassaldanha:
cc @lightclient

EIPS/eip-7685.md
```diff
@@ -13,64 +13,61 @@ created: 2024-04-14
...
 ```
 
-Let `requests` be the list of all `request` objects in the block in ascending
-order by type. For example:
+Let `requests` be the concatenation of each `request_data` returned from each request
+contract:
 
 ```
-[0x00_request_0, 0x01_request_0, 0x01_request_1, 0x02_request_0, ...]
+requests = request_data ++ request_data ++ ... ++ request_data
```
mkalinin, line 52:
```suggestion
requests = request_data_0 ++ request_data_1 ++ ... ++ request_data_n
```

EIPS/eip-7685.md
```diff
@@ -13,64 +13,61 @@ created: 2024-04-14
...
 
 #### Block Header
 
 Extend the header with a new 32 byte value `requests_hash`:
 
 ```python
-def compute_requests_hash(list):
-    return keccak256(rlp.encode([rlp.encode(req) for req in list]))
+def compute_requests_hash(requests):
+    return sha256(requests_data)
```
mkalinin, line 68:
```suggestion
    return sha256(requests)
```

EIPS/eip-7685.md
```diff
@@ -13,64 +13,61 @@ created: 2024-04-14
...
-    field_n,
-    [request_0, ..., request_k],
-])
-```
+The ordering of requests within a type is to be defined by each request type and it
+is the smart contract responsibility to choose.
 
 #### Block Header
 
 Extend the header with a new 32 byte value `requests_hash`:
```
mkalinin, line 64: IMO, it is important to add that matching the requests obtained from the block execution to the `requests_hash` from the block header is a new block validity condition

EIPS/eip-7685.md
```diff
@@ -13,83 +13,117 @@ created: 2024-04-14
...
+As part of the EL validation for a new block, the EL MUST check that the parameter
+`executionRequestsHash` matches their commitment for that block (`block.header.requests_hash`).
+
 ### Consensus Layer
 
-Each proposal may choose how to extend the beacon chain types to include the new
-EL request.
+Each proposal may choose how to extend the beacon chain types to include new EL request types.
+
+As part of beacon block validation, the CL MUST to calculate the hash of the requests
```
mkalinin, line 80:
```suggestion
As part of beacon block validation, the CL MUST calculate the hash of the requests
```

EIPS/eip-7685.md
```diff
@@ -13,83 +13,117 @@ created: 2024-04-14
...
+in the beacon block body, and pass the calculated hash as the `executionRequestsHash`
```
mkalinin, line 81:
```suggestion
in the beacon block body, and pass the calculated hash to the EL to be incorporated into the `block_hash` validation process.
```
Prefer to abstract the spec from the Engine API specifics

EIPS/eip-7685.md
```diff
@@ -13,83 +13,117 @@ created: 2024-04-14
 ## Abstract
 
 This proposal defines a general purpose framework for storing contract-triggered
-requests. It extends the execution header and body with a single field each to
-store the request information. This inherently exposes the requests to the
-consensus layer, which can then process each one.
+requests. It extends the execution header with a single field as a commitment to
+the request. Requests are later on exposed to the consensus layer via the
+Engine API.
```
mkalinin, line 18:
```suggestion
the request. Requests are later on exposed to the consensus layer.
```

EIPS/eip-7685.md
```diff
@@ -13,83 +13,117 @@ created: 2024-04-14
...
+
+```
+requests = request_data_0 ++ request_data_1 ++ ... ++ request_data_n
+```
+
+Note: each `request_data` represents a different list of requests from `beacon_block.body.execution_requests`,
+prepended by their type. They MUST be ordered by their first byte (the `request_type`), matching the
+order used in the EL.
+
+The execution_hash can be calculated as:
```
lucassaldanha, line 110:
```suggestion
The execution_hash can be calculated as:

```

EIPS/eip-7685.md
```diff
@@ -13,83 +13,117 @@ created: 2024-04-14
...
+Let `request_data` be the output of a system call to a Request Smart Contract.
 
-A `request` consists of a `request_type` prepended to an opaque byte array
-`request_data`:
+`request_data` consists of two parts: `request_type` (single byte) and
+`request_list_ssz` (ssz list encoding of many requests) as following:
 
 ```
-request = request_type ++ request_data
+request_data = request_type ++ request_list_ssz
```
mkalinin, line 44: It might worth mentioning a special case for an empty list, where `request_data = requests_type`, to make the spec more explicit

EIPS/eip-7685.md
```diff
@@ -13,83 +13,117 @@ created: 2024-04-14
...
 
-### Opaque byte array rather than an RLP array
+### Ssz byte array rather than an RLP array
 
-By having the bytes of `request_data` array from second byte on be opaque bytes, rather
-than an RLP (or other encoding) list, we can support different encoding formats for the
-request payload in the future such as SSZ, LEB128, or a fixed width format.
+Because the EL does not need to handle `request_data` or decode it. We can make let
+the smart contract decide what encoding to use. For existing requests they are going
+to use SSZ as requests are consumed by the CL.
```
etan-status, line 121: I don't see any benefit for using SSZ here. The CL consumes this data via JSON (engine API), and uses a different data structure internally (`ExecutionRequests`) with strong typing.

What I find useful is the grouping by type, as in, having separate lists per request type. However, they can be a regular RLP array.

```py
0x00_requests = 0x00_request_0 + 0x00_request_1 + 0x00_request_2   # without request type prefixes
0x01_requests = 0x01_request_0 + 0x01_request_1 + 0x01_request_2   # without request type prefixes
0x02_requests = 0x02_request_0 + 0x02_request_1 + 0x02_request_2   # without request type prefixes
requests_hash = keccak256(rlp([0x00_requests, 0x01_requests, 0x02_requests])) 
```

Engine would have a list of the concatenated requests:

```js
{
    requests: [
        "0x...... 0x00_requests",
        "0x...... 0x01_requests",
        "0x...... 0x02_requests"
    ]
}
```

Each of the individual requests items coincides with the encoding for `List[XyzRequest, AnyLimit]` in SSZ, as long as the request data per item is constant sized. This is regardless of what the list limit actually is (SSZ doesn't encode the list length but computes it from the data length). Therefore, the CL can efficiently parse each of the request lists, and enables the EL to treat the data as opaque, while avoiding the introduction of random SSZ `List` types that are EL-only.

_If_ the goal is a `requests_root` SSZ commitment in the EL, the root should match the `hash_tree_root(beacon_block.body.execution_requests)` to have any advantage. The advantage would be that in `newPayload`, instead of sending the full `requests`, the CL could just send the `requests_root` (self-computed), reducing the amount of JSON encoding that's necessary for block validation, and reducing the number of hashes that the EL has to do to validate `newPayload`.

mkalinin: Encoding of a new request list is supposed to be implemented as a part of the corresponding smart contract, like it is done today already for withdrawals and consolidations. So, SSZ is the encoding that we assume can be implemented in a smart contract even if a request data will be of a dynamic size in the future. Given that request encoding becomes opaque for EL clients which is nice as any new request doesn’t require any work on EL side besides adding a new requests smart contract.

The benefit of `requestsHash = sha256(0x00_requests || 0x01_requests || 0x02_requests)` is that CL can already send `requestsHash` to EL in the Engine API instead of sending request data, and this approach is easy to implement without RLP.

etan-status: `sha256(0x00_requests || 0x01_requests || 0x02_requests)` is not the same as `hash_tree_root(beacon_block.body.execution_requests)`.

lucassaldanha: > I don't see any benefit for using SSZ here. The CL consumes this data via JSON (engine API), and uses a different data structure internally (ExecutionRequests) with strong typing.

The benefit is that each request list can be decoded to a list of corresponding request types. Each one of the lists of requests in`ExecutionRequest` has its container type defined, so we can directly decode this list into a ssz list with container elements of the corresponding request type (`DepositRequest`, `ConsolidationRequest` and `WithdrawalRequest`).

etan-status: it may be trivial for the CL given that CL already uses SHA256 for SSZ hashes, but it is a second hash over the request data, so if the request data is large it can be expensive; but if the data is small, the optimization may be premature? Realistically, the extra requests data tops out around 4k when JSON encoded?

on ETH R&D Discord we have also identified a hash collision problem. imagine that all request types consist of just a single `uint8` for these examples:

example 1:
type 0 -> 0x01, 0x02, 0x02, 0x02
type 1 -> none
type 2 -> none

example 2
type 0 -> none
type 1 -> 0x02
type 2 -> 0x02, 0x01, 0x02

both hash as `sha256(0x00, 0x01, 0x02, 0x02, 0x02, 0x01, 0x02)`. So, if one were to swap out example 1 and example 2 in a beacon block, the EL would deem both as VALID.

I'm not sure how easy it is to construct a collision with the initial 3 types and their limits, but as I understand the idea of this request bus is to be generic and extensible, and at the very least the security section of the EIP should address the hashing method, and should prove that collisions are not possible at least for the proposed request types.

in my opinion, the bandwith savings on newPayload are not large enough to warrant introducing an experimental hashing method. it feels off to me that we are fine with exchanging the full request data via libp2p among beacon nodes, which is an inefficient link where latency matters, but then focus on optimizing a local connection where it is deemd fine to waste 50% of data on encoding binary data as hex strings. I understand that it _is_ ugly to send unnecessary data to the EL, but we can also address the optimization in a future fork with alternatives such as SSZ block headers.

_if_ we want to go with a special purpose hash just for requests, from a CL perspective I think it should be SHA256 based, and exclude MPT, RLP, and keccak (so the current proposal would satisfy that).

lucassaldanha: PSA: more on this convo on https://discord.com/channels/595666850260713488/1288408367068610585

lucassaldanha:
Closing in favour of https://github.com/ethereum/EIPs/pull/8924

PR #8923, opened 2024-10-02
Update EIP-658: fix grammar

RenanSouza2:
In this PR one grammar mistake is fixed

### DIFF
--- a/EIPS/eip-658.md
+++ b/EIPS/eip-658.md
@@ -13,7 +13,7 @@ requires: 140
 This EIP replaces the intermediate state root field of the receipt with a status code indicating if the top-level call succeeded or failed.
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


SamWilsn (changes requested):
Are we sure this isn't ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if)?

g11tech (approved):
might require force merge @SamWilsn

RenanSouza2:
> Are we sure this isn't ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if)?

I agree with you that "if and only if" is better but it's not really my place to approve the changes

SamWilsn:
> I agree with you that "if and only if" is better but it's not really my place to approve the changes

I mean more that "iff" is a commonly accepted shorthand for "if and only if", so getting rid of the "f" would change the meaning.

lightclient:
Agree with Sam. Will err on side of caution and simply close this PR.

PR #8924, opened 2024-10-02
Update EIP-7685: group requests into request-data

lightclient:
Hopefully this encapsulates the discussion on discord and in #8916.

Essentially, we now refer to `request_data` as the opaque data output by the system contract instead of a request's individual fields and we modify the `request_hash` calculation to be the following:

```
sha256(sha256(0x00 || request_data_00) || sha256(0x01 || request_data_01) || ...)
```

### DIFF
--- a/EIPS/eip-7685.md
+++ b/EIPS/eip-7685.md
@@ -13,9 +13,9 @@ created: 2024-04-14
 ## Abstract
 
 This proposal defines a general purpose framework for storing contract-triggered
-requests. It extends the execution header and body with a single field each to
-store the request information. This inherently exposes the requests to the
-consensus layer, which can then process each one.
+requests. It extends the execution header with a single field to store the
+request information. Requests are later on exposed to the consensus layer, which
+then processes each one.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


18211308 (commented):
example: "/ip4/7.7.7.7/udp/30303/p2p/QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N"

g11tech (approved):
lgtm

EIPS/eip-7685.md
```diff
@@ -13,70 +13,62 @@ created: 2024-04-14
...
-Extend the header with a new 32 byte value `requests_hash`:
+Or in pseudocode:
 
 ```python
-def compute_requests_hash(list):
-    return keccak256(rlp.encode([rlp.encode(req) for req in list]))
+def compute_requests_hash(requests):
+    m = sha256()
+    for r in requests:
+        m.update(sha256(r))
```
mkalinin, line 62: According to the PR’s description, there is a type byte hashed in front of every list. In this particular case not having a “domain” byte is fine because the order already does this implicitly. Just want to confirm this algo in the spec is the one that it was intended to be

lightclient: maybe not the clearest, but each `r` in `requests` is the prefixed data already

PR #8925, opened 2024-10-03
detalisation of what pairing check actually does

NikZak:
I had hard time to figure what the pairing check actually checks because it is not written in the PR. Beside there was a typo - like if pairing check always outputs 1. So I had to go and [check in the underlying library](https://github.com/Consensys/gnark-crypto/blob/11d2ed41a3d3470f9cf21a8e1cd4ca5d60a3983f/ecc/bls12-381/pairing.go#L45) what is being checked and changed the description here

### DIFF
--- a/EIPS/eip-2537.md
+++ b/EIPS/eip-2537.md
@@ -25,8 +25,8 @@ The motivation of this precompile is to add a cryptographic primitive that allow
 ### Constants
 
 | Name                | Value | Comment            |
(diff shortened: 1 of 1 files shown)


asanso:
@NikZak, thanks a lot for the PR. Could you please remove the other edits (e.g., in the table) and focus only on the affected sentence?

NikZak:
done

asanso (approved):

PR #8933, opened 2024-10-06
Add EIP: Add Controlled Gas Limit Increase Strategy

Giulio2002:
Adds the possibility to have a gradual increase in gas target over time:
1. This EIP does not require an hard fork so in case of a DDOS vector, a fixed gas target can still be set
2. It gives time to analyze network behaviour and bandwidth usage as the gas target increases, instead of setting the increasing the gas target in a cliff-like manner.
3. It is less prone to cause issue when raising the gas target due to it being a gradual change

### DIFF
--- a/EIPS/eip-7779.md
+++ b/EIPS/eip-7779.md
@@ -77,6 +77,11 @@ Add a new "Gas Limit" selection strategy that takes in Block Number `N` and spit
 
 **No Hard Fork Required**
 
+## **Security Considerations**
+
+
+- The controlled gas limit increase strategy is designed to prevent sudden changes that could lead to network instability or security vulnerabilities.
+- The fact that validators can re-adjust the gas limit in case of a DDOS attack or other issues, makes the network more secure than to just increase the gas limit manually in a cliff-like manner.
 
 ## **Copyright**
 
--- a/EIPS/eip-7779.md
+++ b/EIPS/eip-7779.md
@@ -1,12 +1,11 @@
 ---
 eip: 7779
 title: Add Controlled Gas Target Increase Strategy
-description: Add a controlled gas target increase strategy to determine the gas limit of a specific block for a gradual increase in block size, instead than having cliff-like changes.
+description: Adds a controlled gas target increase strategy.
 author: Giulio Rebuffo (@Giulio2002)
 discussions-to: https://ethresear.ch/t/eip-7779-discussion-thread/20574
 status: Draft
-type: Standards Track
-category: Informational
+type: Informational
 created: 2024-10-06
 ---
 
@@ -43,12 +42,12 @@ Add a new "Gas Target" selection strategy that takes in Block Number `N` and spi
  - The gas target `GL_t` at block `t` is calculated as:
 ```GT_t = GT_0 + (t - t_0)*r```
    Where:
-   - `GT_0` is the initial gas target (say 15 million gas).
-   - `r` is gas increase per block (for example 2.89).
-   - `t` is the number of blocks since the implementation of the EIP.
+    - `GT_0` is the initial gas target (say 15 million gas).
+    - `r` is gas increase per block (for example 2.89).
+    - `t` is the number of blocks since the implementation of the EIP.
     - `t_0` is a start block number for the gradual increase.
     - `t_last? is the block number at which the gas target increase stops.
-   - The increase halts once `t` reaches the block `t_last`. 
+    - The increase halts once `t` reaches the block `t_last`. 
 
 If `GT_0` is set to 15 million gas, `r` to 2.89, `t_0` to 0, and `t_last` to  5184000 (two years), the gas target will increase by 2.89 gas per block for two years, reaching 30 million gas at the end of a 2 years period.
 
--- a/EIPS/eip-7779.md
+++ b/EIPS/eip-7779.md
@@ -1,6 +1,6 @@
 ---
 eip: 7779
-title: Add A Controlled Gas Target Increase Strategy
+title: Add Controlled Gas Target Increase Strategy
 description: Gradual Gas Limit Increase strategy To Determine Gas Limit over a two-year period.
 author: Giulio Rebuffo (@Giulio2002)
 discussions-to: https://ethresear.ch/t/eip-7779-discussion-thread/20574
--- a/EIPS/eip-7779.md
+++ b/EIPS/eip-7779.md
@@ -3,7 +3,7 @@ eip: 7779
 title: Add Controlled Gas Target Increase Strategy
 description: Adds a controlled gas target increase strategy.
 author: Giulio Rebuffo (@Giulio2002)
-discussions-to: https://ethresear.ch/t/eip-7779-discussion-thread/20574
+discussions-to: https://ethereum-magicians.org/t/eip-7779-add-controlled-gas-target-increase-strategy/21282
 status: Draft
 type: Informational
 created: 2024-10-06
--- a/EIPS/eip-7779.md
+++ b/EIPS/eip-7779.md
@@ -41,11 +41,12 @@ The EIP proposes the introduction of a new gas limit management mechanism that a
 Add a new "Gas Limit" selection strategy that takes in Block Number `N` and spits out the Gas Limit `GL` for that block. The strategy is as follows:
 
  - The gas limit `GL_t` at block `t` is calculated as:
-```GL_t = GL_0 * (1 + r)^t```
+```GL_t = GL_0 + (t - t_0)*t```
    Where:
    - `GL_0` is the initial gas limit (say 30 million gas).
    - `r` is the per-block increase rate (0.00475% or 0.0000475).
    - `t` is the number of blocks since the implementation of the EIP.
+    - `t_0` is start block number.
    - The increase halts once `t` reaches the block corresponding to two years (approximately 5,184,000 blocks, assuming 12-second block times). this parameters are tuned so that there is a doubling of the gas limit in two years.
 
 ### **Implementation**
--- a/EIPS/eip-7779.md
+++ b/EIPS/eip-7779.md
@@ -1,0 +1,106 @@
+---
+eip: 7779
+title: Controlled Gas Limit Increase
+description: Gradual Gas Limit Increase with Scheduled Deactivation Block
+author: Giulio Rebuffo (@Giulio2002)
+discussions-to: [Forum URL for Discussion]
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-10-06
+---
+
+This proposal introduces a mechanism to gradually increase the block gas limit over two years, doubling from 30 million to 60 million gas, with an automatic deactivation at a specified block height. The purpose is to provide a controlled and predictable growth of block capacity to meet increasing network demand while ensuring network stability.
+
+## **Abstract**
+
+The EIP proposes a system to increase the block gas limit by a fixed amount at regular intervals over a two-year period. The block gas limit will increase incrementally from 30 million gas to 60 million gas, with a safeguard "deactivation block" that halts the increase after two years. This system is designed to accommodate network growth without imposing abrupt changes, giving developers, node operators, and users time to adapt. The incremental increase is fully automated, ensuring transparent and predictable block size growth.
+
+## **Motivation**
+
+### **Predictable Gas Limit Growth**
+
+- **Current Issue:**
+  - The Ethereum network faces increasing demand, but changes to the gas limit are often manually adjusted by miners or validators based on their preferences, which may cause unpredictable block sizes and network performance issues.
+
+- **Need for Change:**
+  - A systematic and predictable increase of the gas limit will help scale the network while giving the ecosystem time to adjust to larger block sizes, without needing to rely on ad hoc decisions by network participants.
+
+### **Gradual Increase with Deactivation Safeguard**
+
+- **Controlled Growth:**
+  - Instead of sudden or unpredictable changes, this EIP proposes incremental gas limit increases over two years, which will gradually double the block capacity.
+  
+- **Automatic Deactivation:**
+  - A safeguard deactivation block will halt the increase after two years, ensuring that the gas limit does not grow indefinitely without further consensus from the community.
+
+## **Specification**
+
+### **Incremental Gas Limit Increase**
+
+1. **Initial Block Gas Limit:**
+   - At the start of the proposal, the block gas limit is set to 30 million gas.
+
+2. **Increase Rate:**
+   - The block gas limit will increase by 0.00475% per block (approximately 1.5% per day), such that after approximately 730 days (two years), the gas limit reaches 60 million.
+
+3. **Deactivation Block:**
+   - The gas limit increase will automatically halt at a specific block height, calculated as the block occurring roughly two years after the proposal's implementation. This "deactivation block" ensures no further increases occur after this period.
+
+4. **Formula for Gas Limit Increase:**
+   - The gas limit `GL_t` at block `t` is calculated as:
+   ```python
+   GL_t = GL_0 * (1 + r)^t
+   ```
+   Where:
+   - `GL_0` is the initial gas limit (30 million gas).
+   - `r` is the per-block increase rate (0.00475% or 0.0000475).
+   - `t` is the number of blocks since the implementation of the EIP.
+   - The increase halts once `t` reaches the block corresponding to two years (approximately 5,184,000 blocks, assuming 12-second block times).
+
+5. **Post-Deactivation Block Limit:**
+   - After the deactivation block, the gas limit remains fixed at 60 million gas unless further action is taken by the community.
+
+### **Implementation**
+
+- **Client Updates:**
+  - Ethereum clients must implement the automatic gas limit increase according to the specified rate and formula, with enforcement of the deactivation block.
+  
+- **Gas Limit Management:**
+  - Miners and validators are not required to adjust the gas limit manually during this period, as the increase will be automated at the protocol level.
+
+## **Rationale**
+
+### **Predictable Growth**
+
+- **Systematic Adjustment:**
+  - The gradual increase avoids sudden surges in gas limit that could destabilize the network. Instead, it provides a smooth transition, giving the ecosystem time to adapt to larger block sizes.
+
+### **Controlled Limit with Deactivation Block**
+
+- **Automatic Safeguard:**
+  - The inclusion of a deactivation block ensures that the gas limit does not increase indefinitely, preventing potential negative impacts on network performance beyond the planned growth.
+
+- **Post-Deactivation Flexibility:**
+  - After the two-year period, the community can reassess the network's needs and decide on further gas limit adjustments if required.
+
+- **Community Consensus:**
+  - The deactivation block serves only as a checkpoint for the community to evaluate the impact of the gas limit increase, however, in the two-year period, the community can decide to halt the increase at any time within any hard fork.
+
+## **Backwards Compatibility**
+
+- **Hard Fork Required:**
+  - The changes to the gas limit management will require a hard fork. Node operators must update their clients to support the new automated gas limit increase mechanism.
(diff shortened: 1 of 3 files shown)


EIPS/eip-7779.md
```diff
@@ -0,0 +1,106 @@
...
+4. **Formula for Gas Limit Increase:**
+   - The gas limit `GL_t` at block `t` is calculated as:
+   ```python
+   GL_t = GL_0 * (1 + r)^t
+   ```
+   Where:
+   - `GL_0` is the initial gas limit (30 million gas).
+   - `r` is the per-block increase rate (0.00475% or 0.0000475).
+   - `t` is the number of blocks since the implementation of the EIP.
+   - The increase halts once `t` reaches the block corresponding to two years (approximately 5,184,000 blocks, assuming 12-second block times).
```
jochem-brouwer, line 59: Plugging in the numbers:

`30_000_000 * (1 + 0.0000475) ^ (5_184_000) = 2.6016850295000575e+114`

I don't think this is correct :sweat_smile:

EIPS/eip-7779.md
```diff
@@ -0,0 +1,106 @@
...
+
+## **Specification**
+
+### **Incremental Gas Limit Increase**
+
+1. **Initial Block Gas Limit:**
+   - At the start of the proposal, the block gas limit is set to 30 million gas.
+
+2. **Increase Rate:**
+   - The block gas limit will increase by 0.00475% per block (approximately 1.5% per day), such that after approximately 730 days (two years), the gas limit reaches 60 million.
```
jochem-brouwer, line 45: For non-integer outputs of the exponential formula, we should round the result down to the nearest integer - right?

EIPS/eip-7779.md
jochem-brouwer: If you want to double the block limit after 5184000 blocks, you need `2 ^ 1/5184000` as the base (so the `1 + r` part). This would result in the formula; `GL_0 * (1 + 000000133708956) ^ t`

EIPS/eip-7779.md
```diff
@@ -0,0 +1,106 @@
...
+5. **Post-Deactivation Block Limit:**
+   - After the deactivation block, the gas limit remains fixed at 60 million gas unless further action is taken by the community.
+
+### **Implementation**
+
+- **Client Updates:**
+  - Ethereum clients must implement the automatic gas limit increase according to the specified rate and formula, with enforcement of the deactivation block.
+  
+- **Gas Limit Management:**
+  - Miners and validators are not required to adjust the gas limit manually during this period, as the increase will be automated at the protocol level.
```
jochem-brouwer, line 70: If miners decide to adjust the gas limit, this would yield an invalid block according to this EIP, so miners are unable to adjust the gas limit.

EIPS/eip-7779.md
```diff
@@ -0,0 +1,106 @@
...
+### **Controlled Limit with Deactivation Block**
+
+- **Automatic Safeguard:**
+  - The inclusion of a deactivation block ensures that the gas limit does not increase indefinitely, preventing potential negative impacts on network performance beyond the planned growth.
+
+- **Post-Deactivation Flexibility:**
+  - After the two-year period, the community can reassess the network's needs and decide on further gas limit adjustments if required.
+
+- **Community Consensus:**
+  - The deactivation block serves only as a checkpoint for the community to evaluate the impact of the gas limit increase, however, in the two-year period, the community can decide to halt the increase at any time as part of any hard fork during the two-year period.
```
jochem-brouwer, line 88: This halting process will take a very long time. If it is decided that this EIP should be paused, in order to get all clients on board for the fork and the testing, even if the fork is only this EIP it will likely take weeks. (Which is problematic if we absolutely need to pause this mechanism)

jochem-brouwer (commented):
Some comments.

In general I am not in favor of this approach. Because we currently have the freedom to let miners change the gas limit, this means that without a hardfork we can decide as community to scale up the gas limit. In case that a problem is found (for instance, a new DoS vector on blocks, like the Shanghai attacks https://ethos.dev/shanghai-attacks ) we can signal to miners that the gas limit should be scaled down (which can be done without a fork).

However, this EIP will remove this flexibility. If a problem is found and we need to decrease the gas limit as fast as possible, this means we have to fork, which will take a substantial amount of time.

As a general suggestion for this EIP: I would suggest linear increase of the gas limit, instead of an exponential increase. The exponential variant could lead to some rounding errors when rounding down to the nearest integer, which could lead to consensus bugs. The linear increase (constant gas limit bump per block) is also slightly easier to code.

paulmillr:
validators are already voicing their concern with increased bandwidth requirements. Increasing gas to 60M will make the situation worse.

Also, why would this be necessary with blob-based roadmap? Gas prices have been low since the last hardfork.

#8931 seems like a better solution.

Giulio2002:
> validators are already voicing their concern with increased bandwidth requirements. Increasing gas to 60M will make the situation worse.
> 
> Also, why would this be necessary with blob-based roadmap? Gas prices have been low since the last hardfork.
> 
> #8931 seems like a better solution.

https://github.com/ethereum/EIPs/pull/8931 does not anything for bandwidth.

EIPS/eip-7779.md
Giulio2002: Fixed your concerns in the updated version of the EIP

EIPS/eip-7779.md
Giulio2002: Fixed your concerns in the updated version of the EIP

EIPS/eip-7779.md
Giulio2002: Yes

EIPS/eip-7779.md
Giulio2002: ooops

Giulio2002: fixed

Giulio2002:
> Some comments.
> 
> In general I am not in favor of this approach. Because we currently have the freedom to let miners change the gas limit, this means that without a hardfork we can decide as community to scale up the gas limit. In case that a problem is found (for instance, a new DoS vector on blocks, like the Shanghai attacks https://ethos.dev/shanghai-attacks ) we can signal to miners that the gas limit should be scaled down (which can be done without a fork).
> 
> However, this EIP will remove this flexibility. If a problem is found and we need to decrease the gas limit as fast as possible, this means we have to fork, which will take a substantial amount of time.
> 
> As a general suggestion for this EIP: I would suggest linear increase of the gas limit, instead of an exponential increase. The exponential variant could lead to some rounding errors when rounding down to the nearest integer, which could lead to consensus bugs. The linear increase (constant gas limit bump per block) is also slightly easier to code.

Can you re-comment on the updated version?

paulmillr:
@Giulio2002 in the #8931 first comment it clearly says "decreases peak bandwidth".

Giulio2002:
> @Giulio2002 in the #8931 first comment it clearly says "decreases peak bandwidth".

It actually does not - also this EIP introduces a new way to raise gas limit that has less DDOS vectors

EIPS/eip-7779.md
```diff
@@ -0,0 +1,91 @@
+---
+eip: 7779
```
abcoathup, line 2:
```suggestion
eip: 7783
```

EIPS/eip-7779.md
```diff
@@ -0,0 +1,91 @@
...
+title: Add Controlled Gas Target Increase Strategy
+description: Adds a controlled gas target increase strategy.
+author: Giulio Rebuffo (@Giulio2002)
+discussions-to: https://ethresear.ch/t/eip-7779-discussion-thread/20574
```
abcoathup, line 6: Please create a discussions topic in Eth Magicians.  https://ethereum-magicians.org/c/eips/5

EIPS/eip-7779.md
```diff
@@ -0,0 +1,91 @@
...
+    - `t_0` is a start block number for the gradual increase.
+    - `t_last? is the block number at which the gas target increase stops.
+    - The increase halts once `t` reaches the block `t_last`. 
+
+If `GT_0` is set to 15 million gas, `r` to 2.89, `t_0` to 0, and `t_last` to  5184000 (two years), the gas target will increase by 2.89 gas per block for two years, reaching 30 million gas at the end of a 2 years period.
+
+### **Implementation**
+
+- **Client Updates:**
+  - Ethereum clients must implement the automatic gas target increase according to the specified rate and formula, with enforcement of the deactivation block.
```
benaadams, line 57:
```suggestion
  - Ethereum clients must implement the automatic gas target increase according to the specified rate and formula, with enforcement of the deactivation block.
  
  If the gasTarget has been locally set by the validator that will override the formula and take precedence.
```

EIPS/eip-7779.md
Giulio2002: yes - i finally managed to

Thompson1985 (approved):

Thompson1985 (approved):

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+
+## **Backwards Compatibility**
+
+**No Hard Fork Required**
+
+## **Security Considerations**
+
+
+- The controlled gas target increase strategy is designed to prevent sudden changes that could lead to network instability or security vulnerabilities.
+- The fact that validators can re-adjust the gas target in case of a DDOS attack or other issues, makes the network more secure than to just increase the gas target manually in a cliff-like manner.
```
benaadams, line 94:
```suggestion
- The fact that validators can re-adjust the gas target in case of a DOS attack or other issues, makes the network more secure than to just increasing the gas target manually in a cliff-like manner.
```

benaadams (approved):

EIPS/eip-7779.md
```diff
@@ -0,0 +1,89 @@
+---
+eip: 7779
+title: Add Controlled Gas Target Increase Strategy
+description: Gradual Gas Limit Increase strategy To Determine Gas Limit over a two-year period.
+author: Giulio Rebuffo (@Giulio2002)
+discussions-to: https://ethresear.ch/t/eip-7779-discussion-thread/20574
+status: Draft
+type: Standards Track
+category: Core
```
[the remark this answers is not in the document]
ayanwaar: hello its me

potuz (commented):
I don't understand the impact of this EIP. It is changing the gas target without changing the gas limit per block, and without changing therefore the pricing mechanism of EIP-1559 that will base the gas price as a function of the parent gas limit
```
		parent_gas_target = self.parent(block).gas_limit // ELASTICITY_MULTIPLIER
```
If the intention on the other hand is to change the gas limit instead (to effectively  reduce the price per tx at the same gas usage) then that could lead to invalid blocks on honest validators that propose with this feature if a majority of operators set the gas limit manually. In this case a clause should be added to cap the maximum change with respect to the parent gas limit as in EIP-1559

Giulio2002:
@potuz I think you have a misunderstanding. Gas Target is a parameter chosen by the validator - it is not protocol enforced and this EIP does not enforce it (as a matter of fact, there is no hard fork needed). Additionally, if most validators have an high gas target - gas limit is also affected as those validators will propose blocks with higher gas limit, thus driving it up.

Let me re-state the EIP motivation. Right now, validators can only choose a fixed gas target. this makes the increase the gas limit to have a cliff-like shape with no protection against DOS. this EIP proposes to use another mechanism for validator to manage their gas target (and thus the gas limit of the blocks they propose), which is instead a gradual increase.

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+## **Specification**
+
+### **Incremental Gas Target Increase Strategy**
+
+Add a new "Gas Target" selection strategy that takes in Block Number `N` and spits out the Gas Target `GT` for that block. The strategy is as follows:
+
+ - The gas target `GT_t` at block `t` is calculated as:
+
+```python
+def compute_gas_target(blockNum: int, blockNumStart: int, initialGasTarget: int, r: int, gasTargetCap: int) -> int:
```
g11tech, line 45: and gasLimit stays same or limit also grows with the 2X relation?

g11tech: if gasLimit is also changing in same proportion, then `2r` has to be less < gasLimit/1024 which is easily statisfied i guess, but still needs a mention for context of `r`

Giulio2002: yes - if the gas target is increased in a large amount sample of validators, yes

g11tech:
> Gas Target is a parameter chosen by the validator - it is not protocol enforced and this EIP does not enforce it

i think potuz is right, gasLimit is chosen and provided by validator and it has be +- parents limit / gasLimitBoundDivisor

gasTarget directly affects consensus as its calculated from the limit and impacts fees and hence consensus

EIPS/eip-7783.md
Giulio2002: I mentioned that gas limit still grows 2x in proportion to the gas target. tell me if it is satisfiable

Giulio2002:
@g11tech @potuz sorry guys - I was confused. what I was talking about is the gas limit choosen by the miner

potuz:
> @g11tech @potuz sorry guys - I was confused. what I was talking about is the gas limit choosen by the miner. I did this mistake because internally we call it `gas target limit`

Then I think you need to take into account what I mention about the cap and possible invalid blocks from honest nodes.

Giulio2002:
> 
> Then I think you need to take into account what I mention about the cap and possible invalid blocks from honest nodes.

So actually - the CL only sends it to the builder API and not to the engine api so this does not affect the EIP

g11tech (dismissed):

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+
+Where:
+
+- `blockNum` is the block number for which the gas limit is being calculated.
+- `blockNumStart` is the block number at which the gas limit increase starts.
+- `initialGasLimit` is the initial gas limit at block `blockNumStart`.
+- `r` is the rate at which the gas limit increases per block.
+- `gasLimitCap` is the maximum gas limit that can be reached.
+
+If we set `blockNumStart` to the current block number, `initialGasLimit` to the current gas limit (`30_000_000`), `r` to 6, and `gasLimitCap` to `60_000_000`, the gas limit will increase by 3 gas per block for two years, reaching 30 million gas at the end of a 2 years period.
```
g11tech, line 60:
```suggestion
If we set `blockNumStart` to the current block number, `initialGasLimit` to the current gas limit (`30_000_000`), `r` to 6, and `gasLimitCap` to `60_000_000`, the gas limit will increase by 6 gas per block for two years, reaching 30 million gas at the end of a 2 years period. 
```
although approved, can you correct this, just noticed

g11tech (approved):

g11tech:
i think it requires an approval from @SamWilsn or @lightclient to merge

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+- **Current Issue:**
+  - The Ethereum network faces increasing demand, but changes to the gas limit are often manually adjusted by miners or validators based on their preferences, which may cause unpredictable block sizes and network performance issues.
+
+- **Need for Change:**
+  - A systematic and predictable increase of the gas limit will help scale the network while giving the ecosystem time to adjust to larger block sizes, without needing to rely on ad hoc decisions by network participants.
+
+### **Gradual Increase with Deactivation Safeguard**
+
+- **Controlled Growth:**
+  - Instead of sudden or unpredictable changes, this EIP proposes incremental gas Limit increases over a specified amount of time, ensuring a smooth transition to higher transaction throughput, while still keeping the governance of the gas Limit in the hand of the community.
```
jochem-brouwer, line 31:
```suggestion
  - Instead of sudden or unpredictable changes, this EIP proposes incremental gas limit increases over a specified amount of time, ensuring a smooth transition to higher transaction throughput, while still keeping the governance of the gas limit in the hand of the community.
```

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+
+Where:
+
+- `blockNum` is the block number for which the gas limit is being calculated.
+- `blockNumStart` is the block number at which the gas limit increase starts.
+- `initialGasLimit` is the initial gas limit at block `blockNumStart`.
+- `r` is the rate at which the gas limit increases per block.
+- `gasLimitCap` is the maximum gas limit that can be reached.
+
+If we set `blockNumStart` to the current block number, `initialGasLimit` to the current gas limit (`30_000_000`), `r` to 6, and `gasLimitCap` to `60_000_000`, the gas limit will increase by 6 gas per block for two years, reaching 30 million gas at the end of a 2 years period.
```
jochem-brouwer, line 60: Nit: note that this is ~694 days (~1.9 years). Caluculation: 

```
blocksNecessary = (30_000_000 / 6)
timeNeeded = blocksNecessary * 12
days = timeNeeded / (24 * 60  * 60)
```

This assumes 12s block time.

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+- `initialGasLimit` is the initial gas limit at block `blockNumStart`.
+- `r` is the rate at which the gas limit increases per block.
+- `gasLimitCap` is the maximum gas limit that can be reached.
+
+If we set `blockNumStart` to the current block number, `initialGasLimit` to the current gas limit (`30_000_000`), `r` to 6, and `gasLimitCap` to `60_000_000`, the gas limit will increase by 6 gas per block for two years, reaching 30 million gas at the end of a 2 years period. 
+
+### **Implementation**
+
+- **Client Updates:**
+  - Ethereum clients must implement the automatic gas limit increase according to the specified rate and formula, with enforcement of the deactivation block. However, If the GasLimit has been locally set by the validator that will override the formula and take precedence.
```
jochem-brouwer, line 65:
```suggestion
  - Ethereum clients must implement the automatic gas limit increase according to the specified rate and formula, with enforcement of the deactivation block. However, if the gas limit has been locally set by the validator that will override the formula and take precedence.
```

I also find the second sentence "However..." formatted a bit weird :sweat_smile:

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+  
+- **Gas Limit Management:**
+  - Miners and validators are not required to adjust the gas limit manually during this period, as the increase will be automated at the protocol level.
```
jochem-brouwer, line 68: I do not understand this, the gas limit formula is optional, so nothing is enforced by the protocol. What is meant by this sentence?

EIPS/eip-7783.md
```diff
@@ -0,0 +1,99 @@
...
+Where:
+
+- `blockNum` is the block number for which the gas limit is being calculated.
+- `blockNumStart` is the block number at which the gas limit increase starts.
+- `initialGasLimit` is the initial gas limit at block `blockNumStart`.
+- `r` is the rate at which the gas limit increases per block.
+- `gasLimitCap` is the maximum gas limit that can be reached.
+
+If we set `blockNumStart` to the current block number, `initialGasLimit` to the current gas limit (`30_000_000`), `r` to 6, and `gasLimitCap` to `60_000_000`, the gas limit will increase by 6 gas per block for two years, reaching 30 million gas at the end of a 2 years period. 
+
```
jochem-brouwer, line 61: I would note somewhere that this formula is the current gas limit which we "want". The gas limit is still subject to certain checks, namely: the proposed gas limit should stay within bounds of the consensus rules. I think, but have to check, that currently the gas limit can only decrease / increase by 1/1024 of the parent blocks' gas limit.

jochem-brouwer (changes requested):
Some comments and suggestions

EIPS/eip-7783.md
Giulio2002: ooops - sorry it is a left over

jochem-brouwer (approved):
LGTM!

g11tech (approved):

g11tech (approved):

PR #8939, opened 2024-10-08
Update EIP-6404: Add `from`, `authorities`, and drop `logs_bloom`

etan-status:
The `from` address was removed from EIP-6404 transactions, computing it is now considered to be part of transaction processing and assumed to be an outcome of its execution. As transactions are stored by both CL & EL, this reduces storage cost, and simplifies transaction validation.

To keep providing a commitment to the `from` adress on-chain, it is now added to the `Receipt`. EIP-7702 `authorities` are also added to sync with the SetCode transaction type.

`Receipt` profiles are removed, there is only a single `Receipt` profile at any given block height. Forward compatibility is provided by EIP-7495 `StableContainer`, similarly as to how EIP-7688 defines it for CL data.

The `logs_bloom` is removed from receipts, it is practically unused in its current form (EIP-7668). It makes sense to free up the data and repurpose it with logs for plain ETH transfers (EIP-7708), both taking about the same space (logs bloom is about twice the size of a log).


Mikhai56 (approved):

etan-status:
@eth-bot rerun

PR #8942, opened 2024-10-09
Update EIP-4762: value-bearing clarification

jsign:
This PR downgrades a generic reference to `*CALL` instructions to only `CALL`.

### DIFF
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -115,7 +115,7 @@ We define **write events** as follows. Note that when a write takes place, an ac
 
 #### Write events for account headers
 
-When a nonzero-balance-sending `*CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
+When a nonzero-balance-sending `CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
 
 ```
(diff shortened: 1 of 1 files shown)


EIPS/eip-4762.md
```diff
@@ -115,7 +115,7 @@ We define **write events** as follows. Note that when a write takes place, an ac
 
 #### Write events for account headers
 
-When a nonzero-balance-sending `*CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
+When a nonzero-balance-sending `CALL` or `SELFDESTRUCT` with a given sender and recipient takes place, process these write events:
```
jsign, line 118: In this EIP `*CALL` refers to `CALL`, `CALLCODE`, `DELEGATECALL` and `STATICCALL`, but:
- Value-bearing `CALLCODE` doesn't produce any effective transfer since the context contract is still the same one that executes the instruction.
- `DELEGATECALL` and `STATICCALL` can't be value bearing.

This means that this rule only makes sense for `CALL`.

gballet (approved):
LGTM

tanishqjasoria (approved):

g11tech (approved):

PR #8929, opened 2024-10-03
Update EIP-7702: restrict field sizes

lightclient:
placeholder for further devnet changes.

outstanding clarifications to make (from my notes):

* authorization should be invalid is nonce == 2^64 - 1 
* This case (add tx-level delegated account to warm)
* Precompiles will not be run once delegated to (treated as empty accounts)
* EXTCODEHASH puts 0 on stack if the 7702-delegated account it is delegates to does not exist in trie (as opposed to empty hash keccak256("") of accounts which exists in trie, but have no code, i.e. EOAs (non-delegated to))
* delegations are not reverted in case transaction fails
* if delegation target account does not exist, nonce is checked to be 0
* is the price correct at 2500? is there incentive to use as alternative contract deployment?
* limit chain id to uint64 and y_parity to uint8

PR #8948, opened 2024-10-09
Update EIP-7702: fixup gas costs and empty code hash

lightclient:
h/t @jochem-brouwer for suggestions in #8929


jochem-brouwer (commented):
Cannot hit "approve" since it is already merged, but yes this looks good to me :smile: :+1:

PR #8949, opened 2024-10-09
Update EIP-6110: harmonize request eips with 7685 changes

lightclient:
This cleans up 6110, 7002, and 7251 with respect to the modifications to 7685. Specifically, `request_data` now refers to the output of the system call to the contracts, not an individual request object's data.

### DIFF
--- a/EIPS/eip-7002.md
+++ b/EIPS/eip-7002.md
@@ -73,10 +73,8 @@ Note that `amount` is returned by the contract little-endian, and must be encode
 
 ```python
 request_type = WITHDRAWAL_REQUEST_TYPE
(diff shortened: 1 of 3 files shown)


EIPS/eip-7002.md
```diff
@@ -73,11 +73,9 @@ Note that `amount` is returned by the contract little-endian, and must be encode
```
ensi321, line 73: > The [EIP-7685](./eip-7685) encoding of a withdrawal request is computed as follows.

Can you also fix the above link to `[EIP-7685](./eip-7685.md)`? It is missing the md extension

lightclient: fixed. thx

PR #8950, opened 2024-10-10
Update EIP-7702: add s value check

lightclient:
adding back since it was removed by mistake.


jochem-brouwer (approved):
LGTM, have also checked if it was `less than` or `less than or equal to` than `secp256k1/2`. It is the latter. So it is correct :)

PR #8951, opened 2024-10-10
Update EIP-4762: add clarification regarding OOG behavior

jsign:
This PR clarifies in the spec how implementations should treat OOG situations on write/read events.

We found a disagreement on this aspect in `execution-spec-tests`, so it's worth clarifying.

cc @gballet @tanishqjasoria @g11tech


g11tech (approved):
lgtm though was already intutive of OOG behaviors impl in EL

PR #8964, opened 2024-10-14
Update EIP-7002: move ssz serialization into system call

lightclient:
This PR should make it clearer that it is the system contract's responsibility to do the ssz serialization and not the caller (i.e. the execution layer client).


jochem-brouwer (approved):
Great clarification, this will avoid some confusion :+1: :smile:

onbjerg (approved):
this is clearer to me :saluting_face:

PR #8969, opened 2024-10-16
Update EIP-7702: Remove delegation behavior of EXTCODE*

frangio:
Proposal to address latest point brought up in EthMagicians.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -86,13 +86,11 @@ If transaction execution results in failure (any exceptional condition or code r
 
 ##### Delegation Designation
 
-The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541.md) to designate the code has a special purpose. This designator requires all code retrieving operations to follow the address pointer to fill the account's observable code. The following instructions are impacted: `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, as well as transactions with `destination` targeting the code with delegation designation..
+The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541.md) to designate the code has a special purpose. This designator requires all code executing operations to follow the address pointer to get the account's executable code. The following instructions are impacted: `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, as well as transactions with `destination` targeting the code with delegation designation.
 
-For example, `EXTCODESIZE` would return the size of the code pointed to by `address` instead of `23` which would represent the delegation designation. `CALL` would similarly load the code from `address` and execute it in the context of `authority`.
+For example, a `CALL` to an `authority` with a delegation designation to `address` would load the code from `address` and execute it in the context of `authority`.
 
 In case a delegation designator points to a precompile address, retrieved code is considered empty and `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL` instructions targeting this account will execute empty code, i.e. succeed with no execution given enough gas.
-
-`EXTCODEHASH` instruction in line with [EIP-1052](./eip-1052.md) puts `0` on stack if delegation designation target does not exist in the trie, or is empty as defined by [EIP-161](./eip-161.md). Emptiness is determined using nonce, balance and code of delegation designation target account. In case target is not empty, but has empty code, keccak256 hash of empty data is put on stack.
 
 In case a delegation designator points to another designator, creating a potential chain or loop of designators, clients must retrieve only the first code and then stop following the designator chain.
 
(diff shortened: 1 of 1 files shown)


smartprogrammer93 (approved):
i am for this change

lightclient (approved):

gumb0:
> A change of behavior in a contract is currently only possible if its code explicitly allows it (in particular via `DELEGATECALL`), and a change of codehash is only possible in the presence of `SELFDESTRUCT` (and as of Cancun only in the same transaction as contract creation),

Change of behavior without DELEGATECALL is the new reality that EIP-7702 brings regardless of this PR change.

Change of codehash is still possible after this PR, when changing EOA => delegated account.

EIPS/eip-7702.md
```diff
@@ -86,14 +86,12 @@ If transaction execution results in failure (any exceptional condition or code r
 
 ##### Delegation Designation
 
-The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541.md) to designate the code has a special purpose. This designator requires all code retrieving operations to follow the address pointer to fill the account's observable code. The following instructions are impacted: `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, as well as transactions with `destination` targeting the code with delegation designation..
+The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541.md) to designate the code has a special purpose. This designator requires all code executing operations to follow the address pointer to get the account's executable code, and requires all other code reading operations to act only on the first 2 bytes of the designator (`0xef01`). The following instructions are impacted: `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, as well as transactions with `destination` targeting the code with delegation designation.
```
jochem-brouwer, line 89: Post-merge nit: it would be slightly more helpful to group the instructions now in the `executing` and `reading` operations, like: 

```suggestion
The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541.md) to designate the code has a special purpose. This designator requires all code executing operations to follow the address pointer to get the account's executable code, and requires all other code reading operations to act only on the first 2 bytes of the designator (`0xef01`). The following reading instructions are impacted: `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, and the following executing instructions are impacted: `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, as well as transactions with `destination` targeting the code with delegation designation.
```

EIPS/eip-7702.md
```diff
@@ -204,6 +202,10 @@ For most intents and purposes, an account delegated to `0x0` is indistinguishabl
 
 This is not ideal and may be a significant enough concern to impact the overall adoption of the EIP. For these reasons, we have opted to include a mechanism which allow users to restore their EOA to its original pureness.
 
+### Delegation of code execution only
+
+Other code retrieving operations like `EXTCODEHASH` do not automatically follow delegations, they operate on the delegation designator itself. If instead delegations were followed, an account would be able to temporarily masquerade as having a particular codehash, which would break contracts that rely on codehashes as an indicator of possible account behavior. A change of behavior in a contract is currently only possible if its code explicitly allows it (in particular via `DELEGATECALL`), and a change of codehash is only possible in the presence of `SELFDESTRUCT` (and as of Cancun only in the same transaction as contract creation), so choosing to follow delegations in `EXTCODE*` opcodes would have created a new type of account that broke prior assumptions.
```
jochem-brouwer, line 207: Retrieving/reading seems to be mixed in this PR.

Also, note that it is possible to **reset the account delegation** by authorizing the zero address. It is thus possible to change the code hash, from the empty hash, to the `0xef01` hash, and then back to the empty hash. This is possible also after the first authorization (so in two txs). So this introduces a new way to "change" the code hash (from non-empty to empty, which does not have to be in the same tx).

EIPS/eip-7702.md
```diff
@@ -86,14 +86,12 @@ If transaction execution results in failure (any exceptional condition or code r
...
+The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3541.md) to designate the code has a special purpose. This designator requires all code executing operations to follow the address pointer to get the account's executable code, and requires all other code reading operations to act only on the first 2 bytes of the designator (`0xef01`). The following instructions are impacted: `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, as well as transactions with `destination` targeting the code with delegation designation.
 
-For example, `EXTCODESIZE` would return the size of the code pointed to by `address` instead of `23` which would represent the delegation designation. `CALL` would similarly load the code from `address` and execute it in the context of `authority`.
+For example, `EXTCODESIZE` would return `2` (the size of `0xef01`) instead of `23` which would represent the delegation designation, `EXTCODEHASH` would return `0xeadcdba66a79ab5dce91622d1d75c8cff5cff0b96944c3bf1072cd08ce018329` (`keccak256(0xef01)`), and `CALL` would load the code from `address` and execute it in the context of `authority`.
```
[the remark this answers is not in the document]
jochem-brouwer: EOF uses the code `0xef00` when code is being read from EOF contracts (from legacy contracts), so it would be similar to use `0xef01` here. If we would use 3 bytes then in the future the 3rd zero byte in `0xef0100` could change, for instance to note versioning. It is then possible to change the output of EXTCODECOPY and EXTCODEHASH, which is what we want to mitigate here. So, using `0xef01` as code is fine.

PR #8970, opened 2024-10-16
Update EIP-6913: Move to Withdrawn

wjmelements:
I am withdrawing this EIP in favor of EIP-7702 which accomplishes the same purpose and has more support.
#### Changes
Change EIP-6913 Status to Withdrawn


wjmelements:
links in the header don't seem to work

PR #8984, opened 2024-10-21
Add EIP: Verifiable logs

etan-status:


EIPS/eip-####.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Verifiable logs
```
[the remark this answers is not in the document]
g11tech: and corresponding filename chane

g11tech (approved):
lgtm

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Configuration
+
+| Name | Value |
+| - | - |
+| `LOG_CONTRACT_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe` |
```
jochem-brouwer, line 30: I would strongly advise against this, since this is currently the system address used by EIPs 7002 and 7251 (scheduled for Prague).

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Log accumulation
+
+After executing all transactions of a block, commitments of all emitted logs are accumulated into the storage of `LOG_CONTRACT_ADDRESS`. The contract has no code, and its storage layout consists of three slots of type `mapping`. However to prevent [EIP-158](./eip-158.md) cleanup, the contract's nonce is set to `1` at the first write.
```
jochem-brouwer, line 34: This should be EIP-161 (included in SpuriousDragon fork), not EIP-158 (never included)

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+
+| Name | Value | Type |
+| `LOG_ADDRESS_STORAGE_SLOT` | `0` | `mapping(address => bytes32)` |
+| `LOG_TOPICS_STORAGE_SLOT` | `1` | `mapping(bytes32 => bytes32)` |
+| `LOG_ADDRESS_TOPICS_STORAGE_SLOT` | `2` | `mapping(bytes32 => bytes32)` |
```
jochem-brouwer, line 39: This looks like a solidity-like type and should include how to decode the mapping to get the actual storage key needed here.

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+    root = hashlib.sha256()
+    root.update(entry_root)
+    root.update(sload(evm.env.state, LOG_CONTRACT_ADDRESS, key))
+    sstore(evm.env.state, LOG_CONTRACT_ADDRESS, key, root.digest())
+
+def track_log(evm: Evm, entry: LogEntry) -> None:
+    entry_root = entry.hash_tree_root()
+
+    # Allow verification via `address` filter
+    key = keccak256(abi.encode(entry.log.address, LOG_ADDRESS_STORAGE_SLOT))
```
jochem-brouwer, line 75: Ah ok the key encoding is listed here. For the JSON response I would definitely include pseudo-code on how to verify these requests.

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+    log: Log
+```
+
+The `hash_tree_root(LogEntry)` commitments are subsequently tracked as part of `LOG_CONTRACT_ADDRESS`.
+
+```python
+def accumulate_log(evm: Evm, entry_root: Bytes32, key: Bytes32):
+    root = hashlib.sha256()
+    root.update(entry_root)
+    root.update(sload(evm.env.state, LOG_CONTRACT_ADDRESS, key))
```
jochem-brouwer, line 68: For completeness: is the `sload` result always left-padded to 32 bytes? (Especially important for the first case when the slot is empty: are these 0 bytes or 32 bytes with value 0?)

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+class LogEntry(Container):
+    meta: LogMeta
+    log: Log
+```
+
+The `hash_tree_root(LogEntry)` commitments are subsequently tracked as part of `LOG_CONTRACT_ADDRESS`.
+
+```python
+def accumulate_log(evm: Evm, entry_root: Bytes32, key: Bytes32):
+    root = hashlib.sha256()
```
jochem-brouwer, line 66: Sha256/keccak256 is used both in this pseudocode, is this intended?

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+
+Starting from the historical log accumulator from (3), each response entry is applied to it in a way compatible with `accumulate_log` above. If the log accumulator ends up matching the value from (4), the response data is correct and `LogEntry` derived from it can be trusted.
+
+## Rationale
+
+Making the `eth_getLogs` response verifiable adds the necessary security attributes to enable wallets to transition away from relying on trusted data providers, ultimately improving the wallet's privacy guarantees as it is no longer subject to the privacy policy of any given provider.
+
+### Gas cost
+
+The gas cost produced by this scheme is significantly higher than what `LOG#` opcodes produce as of Prague, primarily due to the additional `SLOAD` / `SSTORE` requirement and the double cost of `SHA256` opcodes compared to `KECCAK256` opcodes. The gas cost increases outweigh the savings from dropping logs blooms.
```
jochem-brouwer, line 112: I don't see where any gas cost is incurred. The LOG opcode currently does not get a gas bump, the storage itself seems to be done after txs (thus a "system contract" "pays" for it), and the verification is done over RPC off-chain.

EIPS/eip-7792.md
```diff
@@ -0,0 +1,126 @@
...
+
+If the mechanism turns out to be prohibitively expensive even when optimized, it may be necessary to move the log accumulators to a separate optimized data structure (not in `state_root`), or to an out-of-protocol zk system. Even then, the gas cost for logs should still reflect the actual overall cost to update a typical out-of-protocol accumulator to deter against log spamming.
+
+## Backwards Compatibility
+
+It is still possible to process `eth_getLogs` responses from trusted servers as is, without verifying them. Client applications with strict response validation may need to be updated to allow the additional `blockTimestamp` field.
```
jochem-brouwer, line 118: This change needs a fork since this writes as a system-call like to a system contract and is thus not backwards-compatible.

jochem-brouwer (commented):
Sadly this PR got merged, but here is some feedback which I think should be addressed :smile: :+1:

EIPS/eip-7792.md
etan-status: Intended. Keccak256 for compatibility with how event signatures work in Solidity. SHA256 for the SSZ hash chain.

EIPS/eip-7792.md
etan-status: Well, even if the system contract "pays" for it, that's extra steps that need to be done. Today's LOG opcode   similarly charges the cost for Bloom filter updates despite it being done as part of receipt / block header creation.

EIPS/eip-7792.md
etan-status: Shouldn't interfere I think, that address is used as the sender to pull from the withdrawal and consolidation request queues but does not have storage on its own. However, open to change. Added to outstanding issues list on ethereum-magicians.

EIPS/eip-7792.md
etan-status: Thanks! Good catch.

EIPS/eip-7792.md
etan-status: Yes, a fork is needed to introduce such a scheme if done inside a contract. Backwards compatibility addresses existing application logic, which continues to work as is because the extra writes to the `LOG_CONTRACT_ADDRESS` storage slots do not interfere with existing code. As in, a soft fork is sufficient.

If a zk system is used to store the hash chain state, it may not even need a fork.

Note this is in draft stage at this time, not scheduled for inclusion.

etan-status:
Thanks for the extensive review. Have copied the open points to Ethereum-Magicians for visibility.

PR #8985, opened 2024-10-21
Add EIP: EIP-712 Extensions for Account Abstraction

frangio:


EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,147 @@
...
+
+Each member of the array is an object with the following keys:
+
+- `types`: An object with the same format as EIP-712 `types` with at least an `EIP712Domain` key.
+- `domain`: An object that matches the type `EIP712Domain` described in `types`.
+
+For example: 
+
+1. A dapp requests an EIP-712 signature to a connected account via JSON-RPC. `signingDomains` is empty or undefined.
+2. The connected account is a multisig, so it requests EIP-712 signatures from its signers, appending the domain of the multisig to `signingDomains`, which is now an array of length 1.
```
howydev, line 60: small nit: `append` caused some confusion for me, suggest using `prepend` or `insert at front` instead

frangio: Thanks for the heads up! This is actually left over from an earlier version of the spec.

abcoathup:
@frangio this looks like it should be an ERC so suggest creating the PR in the ERC repo.

frangio:
@abcoathup I believe it needs to be an EIP because it concerns the `eth_signTypedData` JSON-RPC method. Can you confirm?

abcoathup:
@frangio 

> Can you confirm?
Checking with EIP-editors (I am not an editor): https://discord.com/channels/916516304972824576/916713912970412103/1298443139060600833

Feel free to chime in.

frangio:
Moving to https://github.com/ethereum/ERCs/pull/693.

PR #8988, opened 2024-10-22
Add EIP: dApp Security Policy Standard

bernard-wagner:
Very early draft for the related ENSIP:

# ENSIP-XX: Security Policy Records

| **Author** | Bernard Wagner(@bernard-wagner) |
| --- | --- |
| **Status** | Draft |
| **Submitted** | TBC |

### Abstract

This ENSIP extends [ENSIP-5: Text Records](https://docs.ens.domains/ensip/5) and defines a record, `dappsec`, that specifies the location of a security policy as described in EIP-TBC. The text record is intended to be queried by wallets for domains imported using [ENSIP-6: DNS-in-ENS](https://docs.ens.domains/ensip/6), thereby notarising the security policy for a particular domain or subdomain. Wallets can then use the security policy to validate the properties of transaction signing requests based on their HTTP origin. The objective is to mitigate the impact of front-end hacks, such as DNS takeovers or supply-chain compromises of front-end components.

### Motivation

Hackers often target dApp front-ends to coerce users into signing transactions that allow the hacker to transfer victims' funds. By introducing a security policy standard, wallet providers can implement safeguards as described in EIP-XXX to protect users against such attacks.

### Specification

Introduce a well-known global text record that allows wallet providers to discover the security policy for a hosted dApp using on-chain data.

**Example Text Record:**

```diff
dappsec: uri=https://mywebapp.xyz/.well-known/ensip-xx.json hash=0x-hex-string
```

- URI: MUST specify the `https` or `ipfs` schemas.
- Hash: When the location is specified using the `https` schema, it MUST contain the policy document's `keccak256` hash as a 0x-prefixed lowercase hex string. The `hash` property is OPTIONAL for the `ipfs` schema, as the policy's integrity is implicit.

### DIFF
--- a/EIPS/eip-dapp-security.md
+++ b/EIPS/eip-dapp-security.md
@@ -1,0 +1,110 @@
+---
+title: dApp Security Policy Standard
+description: Defines a document standard for defining the expected on-chain behavior of a dApp frontend.
+author: Bernard Wagner (bernard-wagner)
+status: Draft
+type: Meta
+created: 2024-10-22
+---
+
+## Abstract
+
+Introduce a standardized JSON document structure that wallets can interpret to determine whether transaction signing requests should be permitted or rejected. Policy documents should be publicly discoverable using a notary service so that a policy's authenticity can be assumed even when the origin of the transaction signing request may be compromised.
+
+As an initial implementation, DNS-in-ENS can be used as the discovery mechanism as described in ENSIP-TBC.
+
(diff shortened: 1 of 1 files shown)


abcoathup:
@bernard-wagner this looks like it should be an ERC.  Suggest creating the PR in the ERCs repo. https://github.com/ethereum/ERCs

PR #8989, opened 2024-10-22
Update EIP-7685: exclude empty requests data in commitment

fjl:
Here I am proposing to change the `requests_hash` commitment so that empty items are excluded. The point of this is to ensure a stable empty requests hash which is independent of fork. Having such a hash makes testing a lot easier, and it mirrors how other execution-layer commitments behave.

There is also a corresponding engine API change: https://github.com/ethereum/execution-apis/pull/599

### DIFF
--- a/EIPS/eip-7685.md
+++ b/EIPS/eip-7685.md
@@ -45,21 +45,19 @@ Each request type will defines its own `requests` object using with its own
 
 #### Block Header
 
-Extend the header with a new 32 byte value `requests_hash`.
+Extend the header with a new 32 byte commitment value `requests_hash`.
 
-The construction looks like:
+A requests hash list is computed by hashing the elements of the block requests list, but
+only ones where `request_data` is non-empty. The commitment is then computed as the sha256
+hash of the list of requests element hashes.
 
-```
-sha256(sha256(requests_0) ++ sha256(requests_1) ++ ...)
-```
-
-Or in pseudocode:
 
 ```python
-def compute_requests_hash(requests):
+def compute_requests_hash(block_requests):
     m = sha256()
-    for r in requests:
-        m.update(sha256(r))
+    for r in block_requests:
+        if len(r) > 1:
+            m.update(sha256(r))
     return m.digest()
 
 block.header.requests_hash = compute_requests_hash(requests)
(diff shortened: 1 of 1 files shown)


EIPS/eip-7685.md
```diff
@@ -45,30 +45,28 @@ Each request type will defines its own `requests` object using with its own
...
-Or in pseudocode:
 
 ```python
-def compute_requests_hash(requests):
+def compute_requests_hash(block_requests):
     m = sha256()
-    for r in requests:
-        m.update(sha256(r))
+    for r in block_requests:
+        if len(r) > 1:
```
jochem-brouwer, line 59: Ok, this works, but this assumes that `request_type` is one byte, which is not specified.

I would also somewhere note that this thus checks that there is any `request_data` (without `request_data` then `r` len is indeed 1 (assuming that it is a `Bytes1` type)). I initially thought this was off and should be `len(r) >= 1`.

fjl: The type will always be one byte. We should probably add that to the EIP.

lucassaldanha:
> Having such a hash makes testing a lot easier

To me this is already a good reason for the change! :)

EIPS/eip-7685.md
```diff
@@ -45,30 +45,28 @@ Each request type will defines its own `requests` object using with its own
...
+            m.update(sha256(r))
```
etan-status, line 60:
```suggestion
            m.update(sha256(r).digest())
```

MarekM25 (approved):

g11tech (dismissed):

mkalinin (approved):

mkalinin:
With this change the order of requests by type is only mentioned in the Rationale which seems to be wrong. The ordering should be either explicitly specified by saying that `block_requests` passed onto the `compute_requests_hash` call must be ordered by the type byte or implicitly done as a part of `compute_requests_hash`. I prefer the latter because written code is more likely to make it into implementation

EIPS/eip-7685.md
```diff
@@ -45,30 +45,30 @@ Each request type will defines its own `requests` object using with its own
...
-```
-sha256(sha256(requests_0) ++ sha256(requests_1) ++ ...)
-```
+The commitment is then computed as the sha256 hash of the list of element hashes.
 
-Or in pseudocode:
 
 ```python
-def compute_requests_hash(requests):
+def compute_requests_hash(block_requests):
```
mkalinin, line 58:
```suggestion
def compute_requests_hash(block_requests: Sequence[bytes]):
```
This is to highlight that `block_requests = [requests_00, requests_01, …]` and not to be messed up with `requests = request_type ++ request_data`

fjl:
I have pushed a more thorough rewrite of the description of the commitment.

lightclient (approved):
LGTM

PR #8993, opened 2024-10-23
A proposal to reward rollup operators for submitting BLOB data to L1

khajievN:

--

 - The build passes.

### DIFF
--- a/EIPS/BLOB Data and Rollup Operator Rewards.md
+++ b/EIPS/BLOB Data and Rollup Operator Rewards.md
@@ -1,0 +1,48 @@
+---
+title: <BLOB Data and Rollup Operator Rewards>
+description: <A proposal to reward rollup operators for submitting BLOB data to L1>
+author: <Ryo Ha(@sopia19910), Khajiev Nizomjon(@khajievN)>
+discussions-to: <URL>
+status: Draft
+type: <Standards Track>
+category: <Core> # Only required for Standards Track. Otherwise, remove this field.
+created: <2024-10-23>
+requires: <EIP number(s)> # Only required when you reference an EIP in the `Specification` section. Otherwise, remove this field.
+---
+
+## Abstract
+
+This proposal designs a system that rewards rollup operators for using BLOB data and DATA TX. The goal is to allow rollup operators to submit data processed on Layer 2 (L2) to Layer 1 (L1) and receive economic incentives, thereby improving the network's scalability and efficiency.
+
+## Motivation
+
+Rollups play an important role in improving Ethereum's scalability, but there is currently no clear reward mechanism for rollup operators. This proposal suggests using BLOB data and DATA TX to create a reward system for rollup operators, aiming to promote network activation and mitigate the centralization issue in Ethereum staking. By providing rewards based on L2 consensus, this system can encourage more participants and strengthen decentralization.
+
+## Specification
+
+1. BLOB Data Usage: Rollup operators store large-scale data in BLOB space and submit it to L1 for verification.
+2. Reward Mechanism: Once the data is successfully submitted and verified, rollup operators are rewarded with L1 coins (ETH).
+3. DATA TX: The transaction that transfers the data is processed by the rollup protocol, and the result is reflected in L1, where the final reward is granted.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+## Rationale
+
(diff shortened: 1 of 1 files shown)


nerolation:
This is spam.

khajievN:
> This is spam.


Hi, I would like to know, why you are saying "Spam" ?

EIPS/eip-7570.md
```diff
@@ -0,0 +1,48 @@
+---
+eip: 7570
```
abcoathup, line 2:
```suggestion
eip: to be assigned
```
Please note EIP/RIP/ERC numbers are manually assigned by editors & associates.  7570 has already been assigned.

abcoathup:
This doesn't appear to be an EIP (in its current form)

Rollups are already using blobs, with the incentive that it is generally cheaper than calldata: https://www.growthepie.xyz/economics

EIPS/eip-7570.md
sopia19910: Hello,
Thank you for letting me know that it's my first time proposing an EIP.
I will make the necessary corrections.

sopia19910:
Hello,

Ethereum has adopted layer technology, where L2 is connected to the Ethereum blockchain with finality at lower fees. This means that L2 blocks are also recognized as part of the Ethereum blockchain.

In other words, Layer 2 is part of the Ethereum blockchain, and I believe that this scalability should extend not only to blocks but also to validators and the reward system. However, the current focus is solely on block expansion, which has led to the perception that Ethereum is becoming centralized.

Therefore, I believe that if the validator and reward systems are also expanded, Ethereum would not be considered centralized.

To address this issue, our team has made a proposal, and we plan to continue suggesting specific code modifications to support this.

To clarify what we are proposing: ultimately, we are suggesting not just providing rewards to operators who perform rollups, but also creating a system where the consensus algorithm of L2 chains interacts with L1 and receives rewards based on that interaction. We will further elaborate on this in more detail.

Allzz77123 (approved):

Allzz77123 (approved):

ppopth:
This doesn't make sense at all. As @abcoathup said, rollups are already incentivized. There is no need for this EIP. I think this PR can be safely closed.

SamWilsn (changes requested):
This proposal is lacking sufficient detail to be assigned an EIP number.

sopia19910:
Based on the current Layer 2 rollup policies, our proposal may not align fully with traditional approaches. However, considering the future scalability of Layer 3, our proposal is worthy of close examination.

In Layer 3, it is highly likely that **specialized data providers** will emerge. In this scenario, we believe that the Ethereum chain will face two strategic options:

1. **Receive extensive data through layers,** enabling client transactions on the Ethereum chain based on aggregated and statistical data, or
2. **Center aggregation solely on rollup blocks at the topmost layer** for a streamlined approach.
We believe that layer scalability should not be limited to simple block finality but should also include extensible reward policies that interact with the consensus algorithms of lower layers.

This proposal envisions a **content provider layer** supplying data to the Ethereum chain. We plan to further refine and update this proposal with more detailed considerations in the future.

g11tech:
> This proposal is lacking sufficient detail to be assigned an EIP number.

@khajievN could you add more details to this EIP proposal for us to consider it for draft?

sopia19910:
@g11tech As requested, we've added detailed specifications to the EIP proposal regarding a reward mechanism designed to reduce cost burdens and enhance sustainability for Rollup operators uploading L2 block data to L1, and we plan to continue our research and improvements on this topic.

EIPS/eip-00000.md
```diff
@@ -0,0 +1,152 @@
...
+    - It can track multiple addresses tied to the same entity to prevent Sybil attacks.
+
+## Backwards Compatibility
+
+- This proposal targets EIP-4844 (Proto-Danksharding) for blob transactions.
+- It can also work with pre-EIP-4844 calldata, though data storage costs are higher there.
+- A smart contract--based reward system does not conflict with existing Ethereum protocol rules.
+- [EIP-1559](./eip-1559.md), or other improvements do not interfere with this scheme.
+
+## Security Considerations
```
SamWilsn, line 144: You have duplicate sections.

SamWilsn:
This proposal is still kind of a mess (no offence intended). You should cleanly separate any application-level content from the changes that require a hardfork. The application-layer content should be an ERC.

This pull request still needs more detail before we can give it a number; it reads more like a plan to come up with a proposal than a proposal itself. You should specify things like the arguments the opcode pops from the stack, what it pushes, the state changes it makes, and the gas cost for executing it. The motivation should focus on why we should put this opcode into the protocol, and the rationale should justify it being an opcode over (for example) a precompile.

Getting technical agreement from the Editors isn't required. Even if we think it's a terrible idea (and I'm not saying this one is), we'll still assign a number as long as the proposal is ~feasible and sketched out technically.

PR #8994, opened 2024-10-23
Update EIP-7742: update the required EL headers and the gas fee mechanism

g11tech:
an alternate and more simplified proposal to https://github.com/ethereum/EIPs/pull/8979

based on the discussion with @bkellerman here:
 - https://github.com/ethereum/EIPs/pull/8979#issuecomment-2430279573

### DIFF
--- a/EIPS/eip-7742.md
+++ b/EIPS/eip-7742.md
@@ -1,8 +1,8 @@
 ---
 eip: 7742
-title: Uncouple blob count between CL and EL
+title: Uncouple blobs per block between CL and EL
 description: Have CL verify blob maximum and have EL get target value from CL
-author: Alex Stokes (@ralexstokes)
+author: Alex Stokes (@ralexstokes), Gajinder Singh (@g11tech), Bert (@bkellerman)
 discussions-to: https://ethereum-magicians.org/t/eip-7742-uncouple-blob-count-between-cl-and-el/20550
 status: Review
 type: Standards Track
@@ -13,7 +13,7 @@ requires: 4844
 
 ## Abstract
 
-Update blob maximum and target verification from [EIP-4844](./eip-4844.md).
+Update blob maximum, target, and blob gas fee computation from [EIP-4844](./eip-4844.md).
 
 The execution layer no longer verifies the blob maximum and receives the target dynamically from the consensus layer.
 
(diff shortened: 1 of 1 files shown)


bkellerman:
LGTM!

To verify correctness and precision over 1 million full blobspaces, then 1 million empty blobspaces, we can see 

image

EIPS/eip-7742.md
```diff
@@ -44,37 +44,68 @@ and deployment of changes to the blob count easier.
...
+def calc_excess_blob_gas(block: Header) -> int:
+  if(block.parent.timestamp < FORK_TIMESTAMP)
+    normalizedParentExcessBlobGas = block.parent.excess_blob_gas * EXCESS_BLOB_GAS_NORMALIZATION_FACTOR // OLD_TARGET_BLOB_GAS_PER_BLOCK
+    blockTargetBlobGas = TARGET_BLOB_GAS_PER_BLOCK
+  else
+    normalizedParentExcessBlobGas = block.parent.excess_blob_gas
+    blockTargetBlobGas = block.target_blob_count * GAS_PER_BLOB
 
-Validity of the `target_blob_count` is guaranteed from the consensus layer, much like how withdrawals are handled.
+  return (normalizedParentExcessBlobGas + (block.blob_gas_used - blockTargetBlobGas) * EXCESS_BLOB_GAS_NORMALIZATION_FACTOR // blockTargetBlobGas)
```
siladu, line 95: In 4844, this function was `def calc_excess_blob_gas(parent: Header)` - this change now mixes data from the current block with the parent. Wondering what the rationale is for using `block.blob_gas_used` and `block.target_blob_count` is instead of the `parent` values?

siladu: If it's possible use parent for all three values, that would simplify the changes, in Besu at least.

This might also impact eth_feeHistory since we'd need data external to the block in order to calculate the next fee

EIPS/eip-7742.md
```diff
@@ -44,37 +44,68 @@ and deployment of changes to the blob count easier.
...
-with each provided payload over the Engine API. The EL block header will also need to be extended with this target value
-to preserve the security of optimistic sync.
+Even if the EL keeps a fixed target value based on the max, removing the max implies the EL would not know what the target value should be. To address this lack of information, this EIP proposes the CL sends the current target value to the EL with each provided payload over the Engine API. The EL block header will also need to be extended with this target value to preserve the security of optimistic sync.
+
+#### Normalized excess blob gas
+
+At any block we don't want the gas update factor (whether pushing gas up or down) to exceed ~1.125. This is essentially achieved in [EIP-4844](./eip-4844.md) by having an appropriate `BLOB_BASE_FEE_UPDATE_FRACTION` to divide `excess_blob_gas` accumulator while exponetiating it
+
+Since the target from CL could theoratically change block to block, we need to normalize excess gas accumulation because the update fraction is actually directly based on the target (max value `excess_blob_gas` can jump by in a block).
```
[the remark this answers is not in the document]
g11tech: ahh yes can shift to that arrangement

EIPS/eip-7742.md
g11tech: ahh yes can shift to that arrangement

g11tech: updated @siladu  and thanks! 
can you check it it all looks good now?

EIPS/eip-7742.md
```diff
@@ -44,37 +44,68 @@ and deployment of changes to the blob count easier.
...
+| `EXCESS_BLOB_GAS_NORMALIZATION_FACTOR` | `100E6` |
+| `NORMALIZED_BLOB_BASE_FEE_UPDATE_FRACTION` | `849018605` |
+| `OLD_TARGET_BLOB_GAS_PER_BLOCK` | `393216` |
+
+At the fork block (where parent.timestamp < FORK_TIMESTAMP), we convert parent's accumulated `excess_blob_gas` to the normalized `excess_blob_gas` to be used in this block's `excess_blob_gas` accumulator.
+
+```python
+def calc_excess_blob_gas(parent: Header) -> int:
+  if(parent.timestamp < FORK_TIMESTAMP)
+    normalizedParentExcessBlobGas = parent.excess_blob_gas * EXCESS_BLOB_GAS_NORMALIZATION_FACTOR // OLD_TARGET_BLOB_GAS_PER_BLOCK
```
ralexstokes, line 87: would be nice if all of these `camelCase` symbols were `snake_case`, just to follow python conventions

EIPS/eip-7742.md
```diff
@@ -13,7 +13,7 @@ requires: 4844
 
 ## Abstract
 
-Update blob maximum and target verification from [EIP-4844](./eip-4844.md).
+Update blob maximum, target and blob gas fee computes from [EIP-4844](./eip-4844.md).
```
ralexstokes, line 16:
```suggestion
Update blob maximum, target and blob gas fee computation from [EIP-4844](./eip-4844.md).
```

EIPS/eip-7742.md
```diff
@@ -44,37 +44,68 @@ and deployment of changes to the blob count easier.
...
+```python
+def calc_excess_blob_gas(parent: Header) -> int:
+  if(parent.timestamp < FORK_TIMESTAMP)
+    normalizedParentExcessBlobGas = parent.excess_blob_gas * EXCESS_BLOB_GAS_NORMALIZATION_FACTOR // OLD_TARGET_BLOB_GAS_PER_BLOCK
+    targetBlobGas = OLD_TARGET_BLOB_GAS_PER_BLOCK
+    maxExcessGasDiffPossible = OLD_TARGET_BLOB_GAS_PER_BLOCK
+  else
+    normalizedParentExcessBlobGas = parent.excess_blob_gas
+    targetBlobGas = parent.target_blob_count * GAS_PER_BLOB
+    maxExcessGasDiffPossible = max(parent.max_blob_count - parent.target_blob_count, parent.target_blob_count) * GAS_PER_BLOB
```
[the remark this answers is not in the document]
g11tech: we need max for max versioned hash validations and for excess blob gas computation validation in optimistic scenarios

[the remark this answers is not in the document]
mkalinin: > we need max for max versioned hash validations

This validation is run by CL, EL already checks that versioned hashes given by CL match those that are obtained from blob transactions contained by the EL block. But if the max is also involved in the excess gas computation that is done by EL then it should be incorporated into a block header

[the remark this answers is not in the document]
g11tech: but for optimistic block its still needed (and yes for bounding fee jumps)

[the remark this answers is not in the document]
mkalinin: Assuming each payload’s block_hash and versioned_hashes are validated during the optimistic sync (and this is true as either CL sends a payload to EL or runs those checks itself), the following steps conclude the validation of the max value during the optimistic sync phase:
* `assert len(body.blob_kzg_commitments) <= config.MAX_BLOBS_PER_BLOCK` in CL
* `expected_versioned_hashes` matches the actual list obtained from the payload’s transactions
* EL block content including transactions matches the `payload.block_hash`

nerolation:
If I understand it correctly, the proposed mechanism doesn't ensure that the max positive and negative deviation from the target lead to the (more or less) same percentage increase in base fee. Instead, the mechanism scales faster down than up.
What we could to is introduce a scaling factor $s$ ($s = T/(M-T)$) and scale the excess gas with it.

Check this for details:
https://hackmd.io/@Nerolation/HJSGsd1Wkl

EIPS/eip-7742.md
```diff
@@ -44,37 +44,57 @@ and deployment of changes to the blob count easier.
 
 #### Maximum blobs per block
 
-The blob maximum is verified in the CL node and the EL inherits this verification during the consistency check of the
-versioned hashes corresponding to each blob as specified by the Engine API. Because of this, the strict check specified
-by EIP-4844 is unnecessary.
+The max blob count is now provided and controlled by the CL during block production which the ELs must use to bound the maximum blobs it can pack in a block.
```
ralexstokes, line 47: im fine to change the text but Im not sure why the original was edited?

EIPS/eip-7742.md
```diff
@@ -44,37 +44,57 @@ and deployment of changes to the blob count easier.
...
 ## Specification
 
-### Block structure and validity
+### Execution Block structure and validity
 
 Upon activation of this EIP, execution clients **MUST** extend the header schema with an
-additional 64-bit field: the `target_blob_count`. This value is set to the current target blob count. The Engine API
-is modified along with this EIP to provide the `target_blob_count` with each payload and implementations can use this
-value to correctly set the block header field.
+additional 64-bit field: the `target_blob_count`. This values is set to the one provided by the CL with the modified engine api fcU's payload attributes.
```
ralexstokes, line 70: would prefer to keep reference to the Engine API more abstract, e.g. in the event we removed the FCU pattern

could even argue to remove all references to the Engine API (as it is all together an implementation detail)

EIPS/eip-7742.md
```diff
@@ -44,37 +44,57 @@ and deployment of changes to the blob count easier.
...
+
+Validity of these values is guaranteed from the consensus layer, much like how withdrawals are handled. Hence when verifying a block, execution clients **MUST** ensure the target blob count in the block header matches the one provided by the consensus client.
+
+For a genesis block with no existing parent, the value should be set according to the agreed specification for the target blob count given by that genesis block's protocol rule set.
+
+Furthermore `get_base_fee_per_blob_gas` as specified by [EIP-4844](./eip-4844.md) is as modified below:
+
+| Constant | Value |
+| - | - |
+| `BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB` | `834619` |
```
ralexstokes, line 82: where did this number come from?

EIPS/eip-7742.md
```diff
@@ -44,37 +44,57 @@ and deployment of changes to the blob count easier.
...
-Validity of the `target_blob_count` is guaranteed from the consensus layer, much like how withdrawals are handled.
 
-When verifying a block, execution clients **MUST** ensure the target blob count in the block header matches the one
-provided by the consensus client.
+```python
+def get_base_fee_per_blob_gas(header: Header) -> int:
+  if header.timestamp < FORK_TIMESTAMP
+    update_fraction = BLOB_BASE_FEE_UPDATE_FRACTION
+  else
+    update_fraction = BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB * header.target_blob_count
```
ralexstokes, line 90: the way EIPs usually read is that they become live at the hard fork; so we don't need to worry about protocol execution before then
```suggestion
    update_fraction = BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB * header.target_blob_count
```

EIPS/eip-7742.md
ralexstokes: I think for the way you want to use this value (just multiplying some number by the `target_count`) we want a value of `1_112_825`

g11tech: ahh yes, i divied up the original mistakenly by 4

EIPS/eip-7742.md
g11tech: will restore, went through several edits in previous versions

EIPS/eip-7742.md
g11tech: updated

EIPS/eip-7742.md
```diff
@@ -44,52 +44,60 @@ and deployment of changes to the blob count easier.
...
+Validity of these values is guaranteed from the consensus layer, much like how withdrawals are handled. Hence when verifying a block, execution clients **MUST** ensure the target blob count in the block header matches the one provided by the consensus client.
 
-For a genesis block with no existing parent, the value should be set according to the agreed specification for the
-target blob count given by that genesis block's protocol rule set.
+Furthermore `get_base_fee_per_blob_gas` as specified by [EIP-4844](./eip-4844.md) is as modified below:
 
-### Block processing
+| Constant | Value |
+| - | - |
+| `BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB` | `1_112_825` |
```
ralexstokes, line 80:
```suggestion
| `BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB` | `1112825` |
```

I think the underscore notation may confused readers

EIPS/eip-7742.md
```diff
@@ -44,52 +44,60 @@ and deployment of changes to the blob count easier.
...
+
+
+```python
```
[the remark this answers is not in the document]
g11tech: yes this is what it would translate to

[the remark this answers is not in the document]
siladu: Thanks. I think this would be useful to include in this EIP's code section too for clarity/completeness.

EIPS/eip-7742.md
```diff
@@ -44,52 +44,60 @@ and deployment of changes to the blob count easier.
...
-Additionally, any reference to `TARGET_BLOB_GAS_PER_BLOCK` from EIP-4844 can be derived by taking the `target_blob_count` from the CL and multiplying by `GAS_PER_BLOB` as given in EIP-4844.
+  return fake_exponential(
+    MIN_BASE_FEE_PER_BLOB_GAS,
+    header.excess_blob_gas,
+    update_fraction
+    )
+```
 
-Otherwise, the specification of EIP-4844 is not changed. For example, blob base fee accounting and excess blob gas tracking occur in the exact same way.
+Rest the specification of EIP-4844 is not changed
```
abcoathup, line 94:
```suggestion
Rest of the specification of EIP-4844 is not changed
```

g11tech:
fyi: `target_blob_count` has been renamed to `target_blobs_per_block` to resolve conflict with the master EIP change (https://github.com/ethereum/EIPs/commit/a9c9d36a018ad40d1234099150287c90e8c13742)

EIPS/eip-7742.md
```diff
@@ -33,67 +33,81 @@ This EIP also changes how the EL sources the current blob target value for two r
...
-### Block processing
+| Constant | Value |
+| - | - |
+| `BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB` | `1112825` |
+
+
+```python
+def calc_excess_blob_gas(parent: Header) -> int:
+  parent_target_blob_gas = parent.target_blobs_per_block * GAS_PER_BLOB
+  if parent.excess_blob_gas + parent.blob_gas_used < parent_target_blob_gas:
```
bkellerman, line 86: These 4 lines can be replaced with

`return max(parent.excess_blob_gas + parent.blob_gas_used  - parent_target_blob_gas, 0)`

It also more intuitively captures the excess as a non-zero accumulation of `error=parent.blob_gas_used  - parent_target_blob_gas`

g11tech: i guess we can leave it as it is inline with 4844 style

EIPS/eip-7742.md
```diff
@@ -33,67 +33,81 @@ This EIP also changes how the EL sources the current blob target value for two r
...
+
+Even if the EL keeps a fixed target value based on the max, removing the max implies the EL would not know what the target value should be. To address this lack of information, this EIP proposes the CL sends the current target value to the EL with each provided payload over the Engine API. The EL block header will also need to be extended with this `target_blobs_per_block` value to preserve the security of optimistic sync.
+
+#### Updated Fee calculations
+
+With the changing target, we also need to scale the `BLOB_BASE_FEE_UPDATE_FRACTION` from [EIP-4844](./eip-4844.md) accordingly to bound the price jumps by +-12.5%. But this introduces some irregularity on the fork block because excess blob gas is effectively scaled down by this changed factor in the fee calculations.
+
+However we consider the target change events to be few and far between and hence not a cause of concern. The likely effect of this factor update will be to bring down the base fee on the fork block and hence would not negatively impact the transactions in txpool.
```
nerolation, line 63: We shouldn't touch the base fee update fraction in 7742 as the EIP is about something very different (even though it can be combined). Instead, there should be a separate change (going into the eip that changes the target) that we ship on as soon as we agree on changing the blob target. This way it's much cleaner.

g11tech:
closing in favor of https://github.com/ethereum/EIPs/pull/9047

PR #8995, opened 2024-10-23
Add EIP: Double speed for hash_tree_root

etan-status:


EIPS/eip-####.md
```diff
@@ -0,0 +1,97 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7797
```

EIPS/eip-####.md
```diff
@@ -0,0 +1,97 @@
...
+title: Double speed for hash_tree_root
+description: Double the performance of hash_tree_root by customizing SHA-256
+author: Etan Kissling (@etan-status)
+discussions-to: https://ethereum-magicians.org/t/eip-double-speed-for-hash-tree-root/21447
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-7797-double-speed-for-hash-tree-root/21447
```
I added the EIP number to the Eth Magicians title

g11tech:
@twoeths could you help review this new EIP PR having work considerably on ssz sha256 internals

twoeths:
the change makes a lot of sense to me. For lodestar on holesky, it takes ~530ms for `hash_tree_root` on epoch transition with wasm sha256 and going to be 130ms - 150ms with  hashtree so this helps a lot. I'm curious to know the `hash_tree_root` time at epoch transition for other clients, do you have this statistic for Nimbus @etan-status ?

etan-status:
Per block, the main effort is for the epoch participation flags. On epoch transition, the balance changes increase the load. Both of these scale with number of validators. I don't have exact timings for Nimbus, but if necessary we should create a reference state with 2 epochs worth of blocks for proper measurement. I have added this as an open item to the Ethereum Magicians thread.

Giulio2002:
I think we need to evaluate this in a post-MaxEB world. I suspect, it is going to make things better when less hashes change in the merkle tree

g11tech (approved):
lgtm for draft

Deluaney1 (commented):
Deposit

princeOmri84 (commented):
I think this is mine
Omri Hutcherson

PR #9002, opened 2024-10-29
Add EIP: System logs

etan-status:

PR #9003, opened 2024-10-29
Update EIP-7708: Provide more details and add fee payments

etan-status:
Add details about log `address` and suggest concrete `topics[0]` values. Further propose a fee payment log for completeness.

Withdrawals and priority fees could be covered with block level system logs, as proposed here:

- https://github.com/ethereum/EIPs/pull/9002/files

Early prototype is getting implemented at https://fusaka-light.box

### DIFF
--- a/EIPS/eip-7708.md
+++ b/EIPS/eip-7708.md
@@ -20,15 +20,32 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
 
 ## Specification
 
-### Parameters
+### ETH transfer logs
 
-* `MAGIC`: `TBD`
+A log, identical to a LOG3, is issued for:
 
-### Functionality
+- Any transaction (including zero-value-transferring), before any other logs created by EVM execution
+- Any nonzero-value `CALL`, at the time that the value transfer executes
+- Any nonzero-value-transferring `SELFDESTRUCT`, at the time that the value transfer executes
 
-Whenever (i) a nonzero-value `CALL`, (ii) a nonzero-value-transferring `SELFDESTRUCT`, or (iii) a nonzero-value-transferring transaction takes place, issue a log, identical to a LOG3, with three topics: (i) `MAGIC`, (ii) the sender address, (iii) the recipient address. The log data is a big-endian 32-byte encoding of the transfer value.
+| Field | Value |
+| - | - |
+| `address` | `0xfffffffffffffffffffffffffffffffffffffffe` ([`SYSTEM_ADDRESS`](./eip-4788.md)) |
+| `topics[0]` | `0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef`<br />(`keccak256('Transfer(address,address,uint256)')`) |
+| `topics[1]` | `from` address (zero prefixed to fill uint256) |
+| `topics[2]` | `to` address (zero prefixed to fill uint256) |
+| `data` | `amount` in Wei (big endian uint256) |
 
-The `LOG` of a value-transferring transaction should be placed before any logs created by EVM execution. The other two `LOG`s are placed at the time that the value transfer executes.
+### Fee payment logs
+
+A log, identical to a LOG2, is issued for every transaction and appended after all other logs created by EVM execution. The fee amount incorporates the total fee charged during transaction execution, including the base fee, the priority fee, and potential blob fees.
+
+| Field | Value |
+| - | - |
+| `address` | `0xfffffffffffffffffffffffffffffffffffffffe` ([`SYSTEM_ADDRESS`](./eip-4788.md)) |
+| `topics[0]` | `0x7bd3aa7d673767f759ebf216e7f6c12844986c661ae6e0f1d988cf7eb7394d1d`<br />(`keccak256('Fee(address,uint256)')`) |
+| `topics[1]` | `from` address (zero prefixed to fill uint256) |
+| `data` | `amount` in Wei (big endian uint256) |
 
 ## Rationale
 
@@ -36,8 +53,9 @@ This is the simplest possible implementation that ensures that all ETH transfers
 
 ### Open questions
 
-1. Should withdrawals also trigger a log? If so, what should the sender address be specified as?
-2. Should fee payments trigger a log? It would ensure "completeness", in the sense that you can compute the exact current balance table by watching logs, but it would greatly increase the number of logs, perhaps to an unacceptably high amount.
+1. Magic value used for `address` and `topics[0]`?
+2. Should fee payments trigger a log? It would ensure "completeness", in the sense that you can compute the exact current balance table by watching logs, but it would greatly increase the number of logs, perhaps to an unacceptably high amount. The extra disk space required for two extra logs per transaction could be covered by removing logs bloom filters with [EIP-7668](./eip-7668.md).
+3. Should withdrawals also trigger a log? They are not associated with transactions.
 
 ## Backwards Compatibility
 


EIPS/eip-7708.md
```diff
@@ -20,24 +20,42 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
...
 ## Specification
 
-### Parameters
+### ETH transfer logs
 
-* `MAGIC`: `TBD`
+A log, identical to a LOG3, is issued for:
 
-### Functionality
+- Any transaction (including zero-value-transferring), before any other logs created by EVM execution
```
etan-status, line 27: Rationale being that something still happened on the wallet that may be of interest to the user. This is only on the outermost call.

70nyIT:
As proposed [here](https://ethereum-magicians.org/t/eip-7708-eth-transfers-emit-a-log/20034/46?u=metony), I think it's better to use the 0xeee...eee address rather than [EIP-4778](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4788.md)

etan-status:
@razorclient currently looking into this as part of EPF

Carsons-Eels:
EIP-7708 currently has no active champion. I made a separate [PR](https://github.com/ethereum/EIPs/pull/11084) to add myself as an author/champion so that I can help accelerate this process since 7708 was CFId for Amsterdam and I am implementing the spec version. 
Once I am added I will review this ASAP. Thanks for summarizing the EthMagician comments into a PR!

etan-status:
Thanks! Yeah, it has been tricky to get any reaction here (I opened this back in 2024).

Regarding the missing logs, I also made EIP-7799 for prio fees + withdrawal logs, and then also this one to make it only emit 1 log per block for the prio fee credit: https://github.com/ethereum/EIPs/pull/10997

Note that 7708 standalone still improves situations, as it removes the need to trace every single transaction. One still has to process all block headers (for fees + withdrawal logs), but that's a much smaller task. Still, can argue that one should just bump it by a fork (H* e.g.) and do the complete scope there, rather than incrementally.

EIPS/eip-7708.md
```diff
@@ -20,24 +20,42 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
...
 ## Rationale
 
 This is the simplest possible implementation that ensures that all ETH transfers are implemented in some kind of record that can be easily accessed through making RPC calls into a node, or through asking for a Merkle branch that is hashed into the block root. The log type is compatible with the ERC-20 token standard, but does not introduce any overly-specific ERC-20 features (eg. ABI encodings) into the specification.
 
 ### Open questions
 
-1. Should withdrawals also trigger a log? If so, what should the sender address be specified as?
-2. Should fee payments trigger a log? It would ensure "completeness", in the sense that you can compute the exact current balance table by watching logs, but it would greatly increase the number of logs, perhaps to an unacceptably high amount.
+1. Magic value used for `address` and `topics[0]`?
+2. Should fee payments and priority fee credits trigger a log? It would ensure "completeness", in the sense that you can compute the exact current balance table by watching logs, but it would greatly increase the number of logs, perhaps to an unacceptably high amount.
```
etan-status: The fee credit could be made manageable with https://eips.ethereum.org/EIPS/eip-8115

etan-status: And the fee deduction could be made manageable with https://eips.ethereum.org/EIPS/eip-7668

EIPS/eip-7708.md
```diff
@@ -20,24 +20,42 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
...
 
-### Functionality
+- Any transaction (including zero-value-transferring), before any other logs created by EVM execution
+- Any nonzero-value `CALL`, at the time that the value transfer executes
+- Any nonzero-value-transferring `SELFDESTRUCT`, at the time that the value transfer executes
 
-Whenever (i) a nonzero-value `CALL`, (ii) a nonzero-value-transferring `SELFDESTRUCT`, or (iii) a nonzero-value-transferring transaction takes place, issue a log, identical to a LOG3, with three topics: (i) `MAGIC`, (ii) the sender address, (iii) the recipient address. The log data is a big-endian 32-byte encoding of the transfer value.
+| Field | Value |
+| - | - |
+| `address` | `0xfffffffffffffffffffffffffffffffffffffffe` ([`SYSTEM_ADDRESS`](./eip-4788.md)) |
```
benaadams: Is 0x000..000 better for Rlp encoding?

EIPS/eip-7708.md
benaadams: Zero transferring is just noise?

benaadams: Would mean every tx has a ETH transfer log; even if just calling smart contract does it add value over the wallet having a tx? (So is implied)

etan-status: That's the intention, yes. That it is possible to enumerate all interactions that a wallet had with Ethereum.

etan-status: Note just the outermost layer, not inner tx

EIPS/eip-7708.md
etan-status: It's a byte array, not an int, so does not use varlen encoding.

RLP encoding is `0x94 + 20byte address` regardless of value.

There is also an idea of `0xeee...eee`: https://ethereum-magicians.org/t/eip-7708-eth-transfers-emit-a-log/20034/46?u=metony

Right now it just uses the same system address from EIP-4788

EIPS/eip-7708.md
benaadams: I mean might as well be the gas that the wallet spent then rather than 0

e.g. Gas -> 0x0

Otherwise adds nothing over there both being a tx and tx receipt for that wallet which also tells you that

benaadams: i.e. is overloading ETH transfers for a different purpose that is already covered by existing block and receipt data

etan-status: OK, have updated PR

etan-status: So, it should be possible to reconstruct balance by processing every single transaction/receipt/blockheader, but no longer need to run trace on every execution anymore.

benaadams: Still says (including zero transferring)? 🤔

EIPS/eip-7708.md
benaadams: I prefer 0x0 as is ETH the asset emitting the log; rather than the system

But not strongly held

EIPS/eip-7708.md
etan-status: Oops, forgot to stage before commit haha

EIPS/eip-7708.md
etan-status: Sure, worth a broader discussion. Keep in mind can also change the value lateron if there are multiple opinions about it. For now, would just settle with any one so that implementations can continue.

petertdavies:
I'm not sure I support the special fee transfer log proposal. In the ERC20 world, any balance change is reflected by a transfer log (mints transfer from `0x0`, burns to `0x0`). Ideally we would like to be as similar to ERC20 machinery as possible. If we wanted to faithfully match the semantics of fees each transaction would need to emit:

1. At the start of the transaction, `sender_address -> 0x0`, paying the theoretical max fee.
2. At the end of the transaction `0x0 -> sender_address`, refund unused gas.
3. At end of the transaction `0x0 -> coinbase`, paying the gas fee.

The problem is that this is 3 logs per transaction. In the EVM gas schedule, Transfer logs cost 1788 gas each (haven't checked thoroughly), so adding fee logs would cost 5364 gas per transaction. If the extra logs emitted are acceptable that would be the best approach.

It would be nice to know how many logs will be emitted by this EIP. It could easily be more than all other sources of logs.

etan-status:
As is, this EIP contains the minimum so that indexers (exchanges, infra providers) no longer have to run debug trace command on every single transaction to discover hidden internal transfers.

They still need to process _all_ block headers, transactions, and receipts, to get a complete picture, i.e., fee payments, block producer rewards, and CL withdrawals, and testnet genesis balances.

EIP-7799 contains the missing aspecst to make it "complete", so that a complete picture can be gotten with solely eth_getLogs API (with a trustless extension). However, 7708 alone would already be very useful.

benaadams:
> Ideally we would like to be as similar to ERC20 machinery as possible. 

Would also cause significant receipt bloat; as would mean transferring an ERC20 token (with no explicit ETH transfer) would also emit 3 ETH transfer logs

EIPS/eip-7708.md
Carsons-Eels: So, if I have this correctly, these are the rough meanings and benefits of each:

- `0x000...000`: *The null address is often used to mean "not a contract" and might take less storage on disk.*
  - "ETH is the asset emitting the log; rather than the system"
  - Highly compressible (but does not matter for RLP encoding)
- `0xeee...eee`: *Some contracts use this to represent native ether, when an address is required.*
  - Many Eth apps already use it
  - eth_simulateV1 (in multiple clients) logs ETH transfers with it
  - Backwards Compatibility, since it would be the same as ERC-20 transfer logs
  - De facto standard when contracts support both tokens & ETH
  - Follows [ERC-7528](https://eips.ethereum.org/EIPS/eip-7528): ETH (Native Asset) Address Convention
- `0xfff...ffe`: *Since moving ether is done by the "system", the system contract should emit it.*
  - Already in use as a system address for [EIP-4788](https://eips.ethereum.org/EIPS/eip-4788): Beacon block root in the 

For now I'm going to implement `0xfff...ffe` in the spec branch, since that's what this reads. I'll ask for responses from client teams on which they prefer. I can update the spec branch later, we can keep this as is for now or you can update it to match @etan-status your choice. Either way, you're right, needs discussion.

@benaadams can I consider your opinion representative of Nethermind as a whole?

EIPS/eip-7708.md
```diff
@@ -4,7 +4,7 @@ title: ETH transfers emit a log
 description: All ETH transfers emit a log
 author: Vitalik Buterin (@vbuterin), Peter Davies (@petertdavies)
```
Carsons-Eels: Also, IMO it would be appropriate to add yourself and @g11tech here since you're directly contributing to this EIPs definition, and the justification from [g11tech's comment](https://github.com/ethereum/EIPs/pull/11084#issuecomment-3753682781)

EIPS/eip-7708.md
etan-status: Not "my choice", 0xfff..e was the only one depicting a concept of an address for "system originated" in the EIPs.

If the tests use 0xeee..e, can also start with that one. Is confusing if EIP doesn't match the tests

EIPS/eip-7708.md
etan-status: Done

petertdavies (approved):
This all looks good. Agree that restricting this EIP to just `CALL` family and `SELFDESTRUCT` transfers is sensible. Questions about fees, withdrawals etc can be pushed to a future EIP.

Happy to see active authors taking this forward!

EIPS/eip-7708.md
fselmo: > If the tests use 0xeee..e, can also start with that one. Is confusing if EIP doesn't match the tests

It was mentioned by @Carsons-Eels two comments above this that the specs and tests will use `0xfff...e` and is currently in PR ([here](https://github.com/ethereum/execution-specs/pull/2023/changes#diff-26f6bc48cc9c6fabb253819bfdaf2c76812f7422389fa1c0b2092e29b76d40e0R34)). The value in this PR is currently `0xeee...e` ([here](https://github.com/ethereum/EIPs/pull/9003/changes#diff-859c8a79b837627a8e9666253cc808fbe4f1a241fb7362360fd63425d18b91acR33)).

Can we agree before merging? From this thread, the agreement looks to be `0xfff...e`? Let's go with this for now?

etan-status: image

His comment was edited; it used to say 0xeee..e

Will do 0xfff...e then to match his latest comment. Thanks for catching it.

fselmo: > His comment was edited; it used to say 0xeee..e

Ah! That makes sense then :) Thanks!

EIPS/eip-7708.md
```diff
@@ -20,24 +20,31 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
...
-### Parameters
+### ETH transfer logs
 
-* `MAGIC`: `TBD`
+A log, identical to a LOG3, is issued for:
 
-### Functionality
+- Any nonzero-value-transferring transaction, before any other logs created by EVM execution
+- Any nonzero-value `CALL`, at the time that the value transfer executes
+- Any nonzero-value-transferring `SELFDESTRUCT`, at the time that the value transfer executes
```
jwasinger: Some edge cases to think about here:
1. a selfdestruct-to-self for a contract created in the same transaction will immediately burn that balance.
2. a selfdestruct-to-self for a previously-existing contract will result in no net balance difference even if it is "nonzero-value-transferring"

For the sake of simplicity, I suggest we treat both cases as a transfer of value to self.  There is not really a proper way to indicate that the balance is being burned in the case of 1.

etan-status: Is it possible from solely the information in the logs (whatever is proposed to be emitted here), the transactions (amount, max prio fee), the receipts (gas used), the block header (base block fee, withdrawals), and the genesis balance (or, whatever the balance was at the time of the fork when this activates) to reconstruct the account balance accurately, in all cases?

Or do we need a second log for burning ETH in these situations?

jwasinger: I think we would need a special log to indicate burn in the case of 1.

jwasinger: for the case of 2.  we can just mark this as a transfer of zero value to self.  It's the only case where a value transfer log of this type would be allowed.

etan-status: The special log could either be a LOG2 `Burn(from, amount)`, or reuse the Transfer log (with a destination of system address). The `Burn` log would probably be cleaner, as it doesn't actually increase the balance of the system address. The `Transfer` log would be a bit more consistent with ERC20 (no need to combine the result from multiple log topics); but for ERC20 it actually gets credited to zero/burnaddress so it's not the same semantics.

etan-status: Or, `Selfdestruct(from, amount)`? To indicate the special operation? Then, (1) and (2) would be the same?

jwasinger: Yeah, I like that 

> Or, `Selfdestruct(from, amount)`? To indicate the special operation? Then, (1) and (2) would be the same?

This could work.

EIPS/eip-7708.md
```diff
@@ -20,24 +20,31 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
...
+| `data` | `amount` in Wei (big endian uint256) |
 
 ## Rationale
 
 This is the simplest possible implementation that ensures that all ETH transfers are implemented in some kind of record that can be easily accessed through making RPC calls into a node, or through asking for a Merkle branch that is hashed into the block root. The log type is compatible with the ERC-20 token standard, but does not introduce any overly-specific ERC-20 features (eg. ABI encodings) into the specification.
 
 ### Open questions
 
-1. Should withdrawals also trigger a log? If so, what should the sender address be specified as?
+1. Magic value used for `address` and `topics[0]`? For address: (a) `0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee` (same as `eth_simulateV1`), (b) `0xfffffffffffffffffffffffffffffffffffffffe` ([`SYSTEM_ADDRESS`](./eip-4788.md)), (c) zero address
```
jochem-brouwer: What are the alternatives for `topics[0]`? It has no examples here.

EIPS/eip-7708.md
```diff
@@ -20,24 +20,31 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
...
+- Any nonzero-value-transferring transaction, before any other logs created by EVM execution
+- Any nonzero-value `CALL`, at the time that the value transfer executes
+- Any nonzero-value-transferring `SELFDESTRUCT`, at the time that the value transfer executes
 
-Whenever (i) a nonzero-value `CALL`, (ii) a nonzero-value-transferring `SELFDESTRUCT`, or (iii) a nonzero-value-transferring transaction takes place, issue a log, identical to a LOG3, with three topics: (i) `MAGIC`, (ii) the sender address, (iii) the recipient address. The log data is a big-endian 32-byte encoding of the transfer value.
-
-The `LOG` of a value-transferring transaction should be placed before any logs created by EVM execution. The other two `LOG`s are placed at the time that the value transfer executes.
+| Field | Value |
+| - | - |
+| `address` | `0xfffffffffffffffffffffffffffffffffffffffe` ([`SYSTEM_ADDRESS`](./eip-4788.md)) |
```
jochem-brouwer: This row `address` is somewhat confusing as `address` also turns up in `topics[0]` and also in the description of `topics[1]` and `topics[2]`. 

Maybe "address emitting log" or something like that?

EIPS/eip-7708.md
```diff
@@ -20,24 +20,31 @@ Logs are often used to track when balance changes of assets on Ethereum. Logs wo
...
+| `topics[0]` | `0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef` (`keccak256('Transfer(address,address,uint256)')`) |
+| `topics[1]` | `from` address (zero prefixed to fill uint256) |
+| `topics[2]` | `to` address (zero prefixed to fill uint256) |
+| `data` | `amount` in Wei (big endian uint256) |
```
jochem-brouwer, line 37: I think for clarity it should be noted that this layout matches the ERC20 Transfer "event"

jochem-brouwer (commented):
Some minor comments.

Description of this PR says: "Further propose a fee payment log for completeness." I do not see this payment log in the diff though.

EIPS/eip-7708.md
etan-status: Indeed. This was a leftover from when no concrete proposal existed yet.

EIPS/eip-7708.md
etan-status: It's just how the field is named in JSON-RPC / how it's accessed via eth_getLogs. Added remark for clarity.

EIPS/eip-7708.md
etan-status: Good idea, added a remark below the table.

EIPS/eip-7708.md
etan-status: Added `Selfdestruct(from, amount)`.

benaadams: > There is not really a proper way to indicate that the balance is being burned in the case of 1.

Isn't burn for ETH a transfer to 0x0? (to null) like emission is a transfer from 0x0

Symbolic burns transfer to `0xdEaD` but that's not what we are talking about here

etan-status: In a `Transfer`, the value is actually credited to the destination. 
This Selfdestruct doesn't credit anything. One couldn't reconstruct the _actual_ value of the zero / 0xdead address if the special case isn't denoted as such.

PR #9004, opened 2024-10-30
Add EIP: eth/70 - Sharded Blocks Protocol

Giulio2002:
This EIP introduces a method enabling an Ethereum node to communicate its available block spans via a bitlist, where each bit represents a 1-million-block span. Nodes use this bitlist to signal which spans of historical data they store, enabling peers to make informed decisions about data availability when requesting blocks. This aims to improve network efficiency by providing a probabilistic snapshot of data locality across nodes.

The proposal extends the Ethereum wire protocol (eth) with version eth/70, introducing a blockBitlist field in the handshake. Nodes probabilistically retain certain block spans to support data locality across the network.


EIPS/eip-7800.md
```diff
@@ -0,0 +1,71 @@
+---
+eip: 7800
```
abcoathup, line 2:
```suggestion
eip: 7801
```
Self assigned numbers are manually bumped by at least one to avoid perception of special treatment or number sniping.  Numbers are assigned by editors & associates.

EIPS/eip-7800.md
```diff
@@ -0,0 +1,71 @@
...
+title: eth/70 - Sharded Blocks Protocol
+description: Replaces block range with a bitlist representing 1-million-block spans in the handshake, with probabilistic shard retention
+author: Ahmad Bitar (@smartprogrammer93) <smartprogrammer@windowslive.com>, Giulio Rebuffo (@Giulio2002)
+discussions-to: https://ethereum-magicians.org/t/eip-7800-eth-70-sharded-blocks-protocol/21507
```
abcoathup, line 6:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-7801-eth-70-sharded-blocks-protocol/21507
```
Added assigned EIP number and changed category in Eth Magicians.  Please use the discussions template: https://ethereum-magicians.org/t/template-for-discussion-to-threads/20347

Giulio2002: mh - is the discussion template only a suggestion? I saw many discussion not doing it. anyway it seems to me a bit overkill to do it

EIPS/eip-7800.md
Giulio2002: LOL

EIPS/eip-7801.md
```diff
@@ -0,0 +1,71 @@
...
+- Advertise a new `eth` protocol capability (version) at `eth/70`.
+  - The existing `eth/69` protocol will continue alongside `eth/70` until sufficient adoption.
+- Modify the `Status (0x00)` message for `eth/70` to add a `blockBitlist` field after the `forkid`:
+  - Current packet for `eth/69`: `[version: P, networkid: P, blockhash: B_32, genesis: B_32, forkid]`
+  - New packet for `eth/70`: `[version: P, networkid: P, blockhash: B_32, genesis: B_32, forkid, blockBitlist]`,
+  where `blockBitlist` is a bitlist, with each bit representing a 1-million-block span.
+
+- Define node behavior based on the bitlist as follows:
+  - **Bitlist Initialization**: Upon startup, nodes **MUST** set at least one bit in the `blockBitlist` to `on` and backfill that 1-million-block span to ensure data availability for that shard.
+  - **Shard Retention Probability**: Nodes **SHOULD** probabilistically retain new block spans, with an `M%` probability (e.g., 5–10%) to maintain network data locality and avoid frequent pruning of new ranges.
```
ak88, line 37: Would be good with some background for the probability calculation.

Could a simpler alternative to this be, "Nodes must retain the current active shard plus one/two shard(s) immediately preceding the active one". 
The idea being that every shard being retained, gives a set amount of time for fresh syncing nodes to pick that range in "Bitlist Initialization".  The question would then be how much time/shards would be sufficient.

EIPS/eip-7801.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Abstract
+
+This EIP introduces a method enabling an Ethereum node to communicate its available block spans via a bitlist, where each bit represents a 1-million-block span. Nodes use this bitlist to signal which spans of historical data they store, enabling peers to make informed decisions about data availability when requesting blocks. This aims to improve network efficiency by providing a probabilistic snapshot of data locality across nodes.
+
+The proposal extends the Ethereum wire protocol (`eth`) with version `eth/70`, introducing a `blockBitlist` field in the handshake. Nodes probabilistically retain certain block spans to support data locality across the network.
+
+## Motivation
+
+With [EIP-4444](./eip-4444.md), nodes may begin pruning historical data while others continue to serve it. The current approach of connecting and requesting specific blocks to determine data availability is inefficient, consuming unnecessary bandwidth. This EIP addresses this by requiring nodes to retain at least one 1-million-block shard and retain additional shards probabilistically, enabling better-informed and targeted sync processes.
```
ak88, line 22: How does a fresh syncing node verify the blocks it downloads in a given shard? 

I would assume nodes would have to keep all  headers for this purpose.

ak88 (commented):
Overall I like this simple approach. 

One downside to this approach i see, is that it will be harder to setup full sync nodes in remote geolocations, since a minimum number of peers is required, which might be difficult if ping is high.

EIPS/eip-7801.md
Giulio2002: Yes - you still need to keep all headers (500 MB/year of growth)

EIPS/eip-7801.md
Giulio2002: I will re-update shortly, ran some numbers and distribution is good enough

PR #9024, opened 2024-11-07
Update EIP-7002: return fee from getter and add usage example

fjl:
The get operation is meant to be used by contracts to compute the exact amount of ether required to add a withdrawal request. It does not return the fee directly though, but it returns the count of 'excess requests' instead. The caller has to compute the fee themselves by applying the fee formula.

I think this is not great. The fee logic, while reasonably straightforward, is an implementation detail of the contract. Duplicating it into caller contracts could lead to a mismatch in the computed values, and it's not necessary. I propose we change the system contract to return the fee directly. This contract change has also been submitted in this PR: https://github.com/lightclient/sys-asm/pull/33

The Rationale section of the EIP also had some outdated text about returning fee overage to the caller. The contract does not return overage, so I am removing that section here, and adding recommendations & example code for calling the contract.

Related change to EIP-7251: #9025

### DIFF
--- a/EIPS/eip-7002.md
+++ b/EIPS/eip-7002.md
@@ -82,7 +82,7 @@ The contract has three different code paths, which can be summarized at a high l
 
 1. Add withdrawal request - requires a `56` byte input, the validator's public
    key concatenated with a big-endian `uint64` amount value.
-2. Excess withdrawal requests getter - if the input length is zero, return the current excess withdrawal requests count.
+2. Fee getter - if the input length is zero, return the current fee required to add a withdrawal request.
 3. System process - if called by system address, pop off the withdrawal requests for the current block from the queue.
 
 ##### Add Withdrawal Request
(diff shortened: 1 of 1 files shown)


pk910:
The number of requests in the queue is very helpful to calculate when a request gets dequeued and processed by the consensus layer..
I get the Idea of not duplicating the fee logic, and appreciate that.  But the number of entries should still be available somehow, as going back from the fee to that number is impossible.

fjl:
Thanks for bringing this up. My question is, do you really need to access the number of excess entries from within a contract?

fjl:
@pk910 I have discussed your concern with @lightclient. While I understand the change removes the ability for contracts to reason about potential inclusion time of the requests in the beacon chain, it's important to understand that such an estimate  can never be reliable. The beacon chain processes the requests asynchonously, i.e. requests output by the system contract will be added into another queue in the beacon client, which it will then apply into future beacon blocks. See [the beacon spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-process_withdrawal_request) for the code. Since it is an unreliable estimate at best, I feel it's OK to lose this functionality, and proceed with this PR.

mkalinin:
@fjl can we return both values (excess and fee) from the call to the system contract?

mkalinin:
> requests output by the system contract will be added into another queue in the beacon client,

This is true for partial withdrawals and consolidations, full withdrawals are instantly processed. I agree that it is very unlikely for _smart contracts_ to need an access to the excess. Also if an app (including block explorers) wants to get that data it can use `eth_getStorageAt` RPC call.

mkalinin:
shouldn’t this PR also update the smart contract's code and its deployment info?

fjl:
@lightclient already merge my change to the code and updated it in the EIP in https://github.com/ethereum/EIPs/pull/9040.

mkalinin (approved):

PR #9025, opened 2024-11-07
EIP-7251: return fee from getter and add usage example

fjl:
The get operation is meant to be used by contracts to compute the exact amount of ether required to add a request. It does not return the fee directly though, but it returns the count of 'excess requests' instead. The caller has to compute the fee themselves by applying the fee formula.

I think this is not great. The fee logic, while reasonably straightforward, is an implementation detail of the contract. Duplicating it into caller contracts could lead to a mismatch in the computed values, and it's not necessary. I propose we change the system contract to return the fee directly. This contract change has also been submitted in this PR: https://github.com/lightclient/sys-asm/pull/33

The Rationale section of the EIP also had some outdated text about returning fee overage to the caller. The contract does not return overage, so I am removing that section here, and adding recommendations & example code for calling the contract.


pk910:
same comment as here: https://github.com/ethereum/EIPs/pull/9024#issuecomment-2465958735
but let's discuss in the other issue :)

mkalinin (approved):

PR #9026, opened 2024-11-07
Multiple Native Tokens

PaulRBerg:
This EIP introduces Multiple Native Tokens as a backward-compatible extension of the EVM, enabling
fungible tokens to function with native-like properties.


PaulRBerg:
Could we please get a number assigned to this EIP?

We have written an accompanying thread for posting on X/ Farcaster, which would attract more reviewers.

cc @abcoathup

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,352 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7809
```

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,352 @@
...
+title: Multiple Native Tokens
+description: Fungible tokens with native-like properties in the EVM
+author: Paul Razvan Berg (@PaulRBerg), Iaroslav Mazur (@IaroslavMazur)
+discussions-to: https://ethereum-magicians.org/t/eip-xxxx-multiple-native-tokens/21615
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-7809-multiple-native-tokens/21615
```
Updated Eth Magicians title with assigned number.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,352 @@
+---
+title: Multiple Native Tokens
```
abcoathup, line 2:
```suggestion
title: Native Tokens
```
Is multiple superfluous as tokens is plural?

u59149403:
All these functionality already exists in https://eips.ethereum.org/EIPS/eip-223 . Read it carefully, including rationale. In particular pay attention to this phrase: "This standard allows payloads to be attached to transactions using the bytes calldata _data parameter, which can encode a second function call in the destination address, similar to how msg.data does in an ether transaction". In other words, 3rd parameter of `transfer` method may contain another call. I. e. you can perform "call with value", i. e. this is what your `NTCALL` does.

See also: https://eips.ethereum.org/EIPS/eip-1363

And finally: https://eips.ethereum.org/EIPS/eip-1155 . eip-1155 allows to transfer multiple tokens in single call. And send arbitrary data, which (similarly to eip 223) may contain another call.

So, I think existing EIPs already subsume your proposal. (Note: I'm not Ethereum developer)

EIPS/eip-xxxx.md
PaulRBerg: Thank you so much @abcoathup!

EIPS/eip-xxxx.md
PaulRBerg: Good point!

ntop001:
This EIP looks good, but it may lead to the weakening of ETH's dominance as a GAS currency, which will be detrimental to the development of the Ethereum ecosystem in the long run. Especially for the L2 network, this EIP will be beneficial to the L2 network but not to Ethereum.

Dexaran:
@PaulRBerg what this opcode should do?

`NTCALL - 0xb4` `NTCALLCODE - 0xb5`

(I suppose they act similar to call and callcode but the behaviour/logic of the opcodes is not written in the proposal, just the args)

Also, why is the transferring method so overcomplicated? 

- transferred_tokens_length: the number of transferred tokens
- The list of transferred_tokens_length (token_id, token_amount) pairs
- argsOffset: byte offset in the memory in bytes, the calldata of the sub context
- argsSize: byte size to copy (size of the calldata)
- retOffset: byte offset in the memory in bytes, where to store the return data of the sub context
- retSize: byte size to copy (size of the return data)

I guess the idea was to let tokens be transferred with just one transaction but this overcomplicates the logic of transferring significantly and I can envision a huge number of mistakes the implementers will be making with that.

I would be in favour of keeping the logic as clean and simple as possible. Only allow 1 token to be transferred at a time, so that you would only need `gas`, `address`, `token_id`, `quantity` and `data`. No offsets, no lists in args.

wjmelements:
@Dexaran the preferred forum for discussion is specified by `discussions-to`

EIPS/eip-7809.md
```diff
@@ -0,0 +1,355 @@
...
+
+`ETH` becomes the 'Base Token', with its ID and supply initialized to zero. `ETH` is the only NT whose supply is not
+tracked explicitly, i.e., its supply is determined just like it currently is.
+
+For increased security and consistency, the token contracts representing the NTs SHOULD NOT use an upgradeability
+pattern.
+
+### Stack
+
+Since the EVM stack can support only up to 1024 elements, there is a natural limit to the number of tokens that can be
```
g11tech, line 76: need a review over here to substantiate the claim

PR #8774, opened 2024-07-30
Add EIP: Add wallet_signIntendedValidatorData method

YamenMerhi:
- Add wallet_signIntendedValidatorData JSON-RPC method

to discuss and to know more with non standardized sections here: [PatronumLabs/wallet_signIntendedValidatorData](https://github.com/Patronum-Labs/wallet_signIntendedValidatorData)

PR #9027, opened 2024-11-09
Update EIP-7749: Add Test Cases section

YamenMerhi:
As per @SamWilsn suggestions in https://github.com/ethereum/EIPs/pull/8774#discussion_r1723466592.

> An example request/response probably belongs in the Test Cases section. Ideally, we'd like to see a [JSON Schema](https://json-schema.org/) in the specification section describing the endpoint.

            

- Adding a json schema describing the endpoint
- Move example to test cases section


Scamreno:
I approve this

PR #9028, opened 2024-11-09
Add EIP: Introspection opcodes

Brechtpd:
This EIP proposes calculating the transactions trie root while the block is being built, and adding an opcode that exposes the current transactions trie root to the EVM through an opcode.

Execution spec implementation: https://github.com/ethereum/execution-specs/pull/1345

### DIFF
--- a/EIPS/eip-7814.md
+++ b/EIPS/eip-7814.md
@@ -3,7 +3,7 @@ eip: 7814
 title: Introspection precompiles
 description: Introspection precompiles that expose the current block context to the EVM
 author: Brecht Devos (@Brechtpd)
-discussions-to: TBD
+discussions-to: https://ethereum-magicians.org/t/eip-7814-introspection-precompiles/21872
 status: Draft
 type: Standards Track
 category: Core
--- a/EIPS/eip-7814.md
+++ b/EIPS/eip-7814.md
@@ -52,7 +52,9 @@ N/A
 
 ## Security Considerations
 
-<TODO>
+Needs discussion. <!-- TODO -->
+
+This EIP might impact MEV.
 
 ## Copyright
 
--- a/EIPS/eip-7814.md
+++ b/EIPS/eip-7814.md
@@ -1,7 +1,7 @@
 ---
 eip: 7814
-title: Introspection precompiles
-description: Introspection precompiles that expose the current block context to the EVM
+title: Introspection opcodes
+description: Introspection opcodes that expose the current block context to the EVM
 author: Brecht Devos (@Brechtpd)
 discussions-to: https://ethereum-magicians.org/t/eip-7814-introspection-precompiles/21872
 status: Draft
@@ -12,35 +12,35 @@ created: 2024-11-09
 
 ## Abstract
 
-This EIP proposes to add two precompiles that enable introspection of the chain state at arbitrary points within a block in the EVM. Currently, the EVM only has access to the state of previous blocks. No block data is currently exposed to the EVM for the block it's executing in.
+This EIP proposes to add an opcode that enables introspection of the chain state at arbitrary points within a block in the EVM. Currently, the EVM only has access to the state of previous blocks. No block data is currently exposed to the EVM for the block it's executing in.
 
 ## Motivation
 
-The new precompiles aim to enhance introspection capabilities within the EVM, enabling the calculation of the latest chain state offchain at any point in an Ethereum block. This is important to allow general and efficient synchronous composability with L1. Otherwise, to ensure having the latest L1 state, the state would have to be read on L1 and passed in as a separate input. This is expensive and there may be limitations on who can read the state without something like [EIP-2330](https://eips.ethereum.org/EIPS/eip-2330).
+The new opcode aims to enhance introspection capabilities within the EVM, enabling the calculation of the latest chain state offchain at any point in an Ethereum block. This is important to allow general and efficient synchronous composability with L1. Otherwise, to ensure having the latest L1 state, the state would have to be read on L1 and passed in as a separate input. This is expensive and there may be limitations on who can read the state without something like [EIP-2330](./eip-2330).
 
-This proposal allows computing the latest state from the state root of the previous block and the transactions that are in the current block. This data can then be passed into any system requiring the latest chain state where the partial block can be re-executed to compute the latest state in a provable way.
+This proposal allows computing the latest state from the state root of the previous block and the transactions that are in the current block. This data can then be passed into any system requiring the latest chain state where the partial block can be re-executed to compute the latest state before the start of the current transaction in a provable way. The state can be computed at arbitrary points inside the transaction as well by enforcing things inside the current transaction, this requires no additional EVM functionality.
 
 ## Specification
 
-If `block.number >= TBD` two new precompiled contracts shall be created:
-- `TXTRIEROOT` at address `TBD`: This precompile returns the transaction trie root of all transactions in the current block, up to and including the transaction that is currently executing. The tx trie is already calculated for the block header. This EIP just enforces that the trie is constructed incrementally per transaction and exposes the root to the EVM.
-- `OPCODECOUNTER` at address `TBD`: This precompile returns a 4 byte uint in big endian encoding representing the total number of opcodes that have been executed for the current transaction up till (and excluding) the call to this precompile.
+- `TXROOT`: This opcode returns the transaction trie root of all transactions in the current block, excluding the transaction that is currently executing (which could introduce a circular dependency between the merkle tree proof and the transactions hash and thus the merkle trie root). The tx trie is already calculated for the block header. This EIP just enforces that the trie is constructed incrementally per transaction and exposes the root to the EVM.
 
 ### Gas Cost
 
-The gas cost for `TXTRIEROOT` and `OPCODECOUNTER` is a fixed fee of `2`.
+The gas cost for `TXROOT` is a fixed fee of `2`.
 
 ## Rationale
 
 Simple and efficient access to the latest state of a chain is critical for composability. For synchronous composability, we need to be able to immediately prove all offchain work inside the same block. This makes it impossible to delay the proving to a later block where more information about the current block is available.
 
+Onchain functionality to verify arbitrary state like [EIP-2330](./eip-2330) is great, but still expensive because it still requires onchain logic. This EIP makes it possible to move all verfication logic offchain without any limitations or edge cases (we need to be able to make sure ALL state is as expected, which is easy if you can simply re-execute).
+
+It also supersedes [EIP-7793](./eip-7793.md) which only exposes the current transaction index for use in the context of encrypted mempools.
+
+After [EIP-7793](./eip-7825.md) the transaction size will be limited. This EIP makes it possible to chain transactions that are block position dependent (i.e. multiple transactions that need to be top of block after each other).
+
 ### Gas Price
 
 The precompiles are priced to match similar opcodes in the `W_base` set.
-
-### Precompile
-
-Implementing this feature via precompiles instead of opcodes gives L2s flexibility to decide whether to implement it.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-7814.md
+++ b/EIPS/eip-7814.md
@@ -52,7 +52,7 @@ N/A
 
 ## Security Considerations
 
-Needs discussion.
+<TODO>
 
 ## Copyright
 
--- a/EIPS/eip-introspection.md
+++ b/EIPS/eip-introspection.md
@@ -1,0 +1,59 @@
+---
+eip: TBD
+title: Introspection precompiles
+description: Introspection precompiles that expose the current block context to the EVM
+author: Brecht Devos (@Brechtpd)
+discussions-to: TBD
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-11-09
+---
+
+## Abstract
+
+This EIP proposes to add two precompiles that enable introspection of the chain state at arbitrary points within a block in the EVM. Currently, the EVM only has access to the state of previous blocks. No block data is currently exposed to the EVM for the block it's executing in.
+
+## Motivation
+
+The new precompiles aim to enhance introspection capabilities within the EVM, enabling the calculation of the latest chain state offchain at any point in an Ethereum block. This is important to allow general and efficient synchronous composability with L1. Otherwise, to ensure having the latest L1 state, the state would have to be read on L1 and passed in as a separate input. This is expensive and there may be limitations on who can read the state without something like [EIP-2330](https://eips.ethereum.org/EIPS/eip-2330).
+
+This proposal allows computing the latest state from the state root of the previous block and the transactions that are in the current block. This data can then be passed into any system requiring the latest chain state where the partial block can be re-executed to compute the latest state in a provable way.
+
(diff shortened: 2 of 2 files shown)


EIPS/eip-introspection.md
```diff
@@ -0,0 +1,59 @@
+---
+eip: TBD
```
abcoathup, line 2:
```suggestion
eip: 7814
```

EIPS/eip-introspection.md
```diff
@@ -0,0 +1,59 @@
...
+title: Introspection precompiles
+description: Introspection precompiles that expose the current block context to the EVM
+author: Brecht Devos (@Brechtpd)
+discussions-to: TBD
```
abcoathup, line 6: Please create a discussions topic in Eth Magicians using the template: https://ethereum-magicians.org/c/eips/5

EIPS/eip-7814.md
```diff
@@ -0,0 +1,59 @@
...
+created: 2024-11-09
+---
+
+## Abstract
+
+This EIP proposes to add two precompiles that enable introspection of the chain state at arbitrary points within a block in the EVM. Currently, the EVM only has access to the state of previous blocks. No block data is currently exposed to the EVM for the block it's executing in.
+
+## Motivation
+
+The new precompiles aim to enhance introspection capabilities within the EVM, enabling the calculation of the latest chain state offchain at any point in an Ethereum block. This is important to allow general and efficient synchronous composability with L1. Otherwise, to ensure having the latest L1 state, the state would have to be read on L1 and passed in as a separate input. This is expensive and there may be limitations on who can read the state without something like [EIP-2330](https://eips.ethereum.org/EIPS/eip-2330).
```
g11tech, line 19: kindly address the bot failures

EIPS/eip-7814.md
```diff
@@ -0,0 +1,59 @@
...
+
+This proposal allows computing the latest state from the state root of the previous block and the transactions that are in the current block. This data can then be passed into any system requiring the latest chain state where the partial block can be re-executed to compute the latest state in a provable way.
+
+## Specification
+
+If `block.number >= TBD` two new precompiled contracts shall be created:
```
g11tech, line 25: we now don't add hardfork number/timestamp information in the EIPs

EIPS/eip-7814.md
```diff
@@ -0,0 +1,59 @@
...
+- `TXTRIEROOT` at address `TBD`: This precompile returns the transaction trie root of all transactions in the current block, up to and including the transaction that is currently executing. The tx trie is already calculated for the block header. This EIP just enforces that the trie is constructed incrementally per transaction and exposes the root to the EVM.
```
g11tech, line 26: add a html <-- TODO --> tag to be autoflagged at the review PR stage

g11tech, line 26: seems that it would be technically feasible to only compute stateroot till previous tx. kinky confirm that what you wrote is what you intend to propose

EIPS/eip-7814.md
```diff
@@ -0,0 +1,59 @@
...
+- `OPCODECOUNTER` at address `TBD`: This precompile returns a 4 byte uint in big endian encoding representing the total number of opcodes that have been executed for the current transaction up till (and excluding) the call to this precompile.
+
+### Gas Cost
+
+The gas cost for `TXTRIEROOT` and `OPCODECOUNTER` is a fixed fee of `2`.
+
+## Rationale
+
+Simple and efficient access to the latest state of a chain is critical for composability. For synchronous composability, we need to be able to immediately prove all offchain work inside the same block. This makes it impossible to delay the proving to a later block where more information about the current block is available.
```
g11tech, line 35: needs some examples where this proposal would be beneficial. rationale is a bit light imo and needs more support and reasoning.

EIPS/eip-7814.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Gas Price
+
+The precompiles are priced to match similar opcodes in the `W_base` set.
+
+### Precompile
+
+Implementing this feature via precompiles instead of opcodes gives L2s flexibility to decide whether to implement it.
```
g11tech, line 43: do you mean precompiles vs system contracts? precomplies are sort of an opcode implementation

EIPS/eip-7814.md
```diff
@@ -0,0 +1,59 @@
...
+
+Further discussion required.
+
+## Test Cases
+
+N/A
+
+## Security Considerations
+
+Needs discussion.
```
g11tech, line 55: add html <-- TODO --> tag here and else where (discussion pending) to be autoflagged at review status update stage

EIPS/eip-7814.md
Brechtpd: Changed behavior and clarified (hopefully)!

EIPS/eip-7814.md
Brechtpd: Changed to opcode instead now

EIPS/eip-7814.md
```diff
@@ -0,0 +1,61 @@
...
+
+This proposal allows computing the latest state from the state root of the previous block and the transactions that are in the current block. This data can then be passed into any system requiring the latest chain state where the partial block can be re-executed to compute the latest state before the start of the current transaction in a provable way. The state can be computed at arbitrary points inside the transaction as well by enforcing things inside the current transaction, this requires no additional EVM functionality.
+
+## Specification
+
+- `TXROOT`: This opcode returns the transaction trie root of all transactions in the current block, excluding the transaction that is currently executing (which could introduce a circular dependency between the merkle tree proof and the transactions hash and thus the merkle trie root). The tx trie is already calculated for the block header. This EIP just enforces that the trie is constructed incrementally per transaction and exposes the root to the EVM.
```
jochem-brouwer, line 25: I think I am missing something here. What does `TXROOT` calculate? It is clearly not the `block.transactionsRoot`, at least that is what I understand from this motivation.

If we have a block with txs A,B,C, then if I am executing B then `TXROOT` should return the root without B (?) in it. How is this calculated? Should I put A at key 0, C at key 2? Or should I only include A? Since I have to recalculate the root, the `2` gas price also is too low (this would likely be ok if it is indeed the transactions root as calculated for the header). In the current spec it seems that for each transaction, the `TXROOT` will return a different root. Is this correct? How should this be calculated (at which keys should the txs be placed and which txs should be included)? I am also currently not understanding why we would not want the current tx in this root. I do not see how this can create a circular dependency.

EIPS/eip-7814.md
```diff
@@ -0,0 +1,61 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-7814-introspection-precompiles/21872
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-11-09
+---
+
+## Abstract
+
+This EIP proposes to add an opcode that enables introspection of the chain state at arbitrary points within a block in the EVM. Currently, the EVM only has access to the state of previous blocks. No block data is currently exposed to the EVM for the block it's executing in.
```
jochem-brouwer, line 15: I get the idea but this statement is not correct, for instance `TIMESTAMP` and `NUMBER` are accessible.

EIPS/eip-7814.md
```diff
@@ -0,0 +1,61 @@
...
+
+Onchain functionality to verify arbitrary state like [EIP-2330](./eip-2330) is great, but still expensive because it still requires onchain logic. This EIP makes it possible to move all verfication logic offchain without any limitations or edge cases (we need to be able to make sure ALL state is as expected, which is easy if you can simply re-execute).
+
+It also supersedes [EIP-7793](./eip-7793.md) which only exposes the current transaction index for use in the context of encrypted mempools.
+
+After [EIP-7793](./eip-7825.md) the transaction size will be limited. This EIP makes it possible to chain transactions that are block position dependent (i.e. multiple transactions that need to be top of block after each other).
+
+### Gas Price
+
+The precompiles are priced to match similar opcodes in the `W_base` set.
```
jochem-brouwer, line 43: Is this intended to be a precompile or opcode?

EIPS/eip-7814.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Motivation
+
+The new opcode aims to enhance introspection capabilities within the EVM, enabling the calculation of the latest chain state offchain at any point in an Ethereum block. This is important to allow general and efficient synchronous composability with L1. Otherwise, to ensure having the latest L1 state, the state would have to be read on L1 and passed in as a separate input. This is expensive and there may be limitations on who can read the state without something like [EIP-2330](./eip-2330).
+
+This proposal allows computing the latest state from the state root of the previous block and the transactions that are in the current block. This data can then be passed into any system requiring the latest chain state where the partial block can be re-executed to compute the latest state before the start of the current transaction in a provable way. The state can be computed at arbitrary points inside the transaction as well by enforcing things inside the current transaction, this requires no additional EVM functionality.
```
jochem-brouwer, line 21: A very long time ago, the transaction receipts included an intermediate state root. This has changed since Byzantium to the status (0 or 1) of the transaction: https://eips.ethereum.org/EIPS/eip-658. Although the intermediate state root was not accessible in the EVM, would such thing help here if we would reintroduce it?

jochem-brouwer (commented):
Some questions :smile: :+1:

EIPS/eip-7814.md
Brechtpd: Sorry it seems that I should have explained better! I already implemented it in the execution spec here which may help understand: https://github.com/ethereum/execution-specs/pull/1345

So if a block contains 3 transactions A, B, and C, the transactions trie root would be as followed in each transaction:
- A: empty
- B: root([A])
- C: root([A, B])

With this we can find out the following for example:
- We can calculate the L1 state before each transaction, e.g. in transaction B we can calculate the L1 state using prev_block.state_root and re-executing A
- We know in each transaction what its position in the block is. In C we know there are 2 previous transactions so we are at the 3rd tx in the block.

I believe the gas cost of the opcode can be so low because the full transaction trie root needs to be calculated for the block header anyway, and so this EIP only changes so that it now incrementally builds this merkle tree. This is more expensive, but I would say still very small when the number of elements in the trie is small (which is the case for the transactions trie). The opcode would not trigger the calculation of the root (which would always be done), it simply exposes it. It is a valid point though that the opcode could trigger the calculation. Not sure if this is better or not.

The circular dependency would that the tx hash changes based on the tx input. If that input is a merkle proof showing that some tx is in the root, then putting this data in the tx changes the tx hash, which changes the tx trie root, which then would require a different merkle proof to check for inclusion!

EIPS/eip-7814.md
Brechtpd: Sorry first made it a precompile but then though hmmm this does seem better as just an opcode because it's so simple and general. But open for any feedback on this!

EIPS/eip-7814.md
Brechtpd: Oh right yes, I guess I mean any data about a block that is not already known at the very start of the block.

EIPS/eip-7814.md
Brechtpd: Oh interesting, I didn't know that!

Unfortunately this would not help for the cases I am trying to solve. Which is exposing this kind of data to the EVM so that it can be used in proofs that can be verified onchain.

EIPS/eip-7814.md
jochem-brouwer: I think there is some confusion here for the terms used and I am currently not sure if we are talking about the transactions root or the state root. Just to verify here: it seems you are talking about the state root and not the tx root? The transactions trie root is the root of the MPT where you put all the serialized transactions in (in the corresponding order). It seems that here, if I call TXROOT, I get the state root after executing the previous txs (?). If this is not the case and it is actually the root of the previous transactions (so if we are in tx C then we thus apply the serialized A to key 0 and B to key 1) then we cannot know directly that C is the third (key 2) transaction. Because in order to prove this on-chain, we would need a proof that it is indeed at key 2, which means we a MPT proof which would thus contain both the serialized txs (so we can reconstruct the MPT in EVM and then see we are indeed at key 2) or in some cases hashes of the relevant MPT nodes.

Do not interpret this as hostile - I am mainly figuring out the motivation and end goal of this EIP. It seems that you want to supply the state root as it was right before executing the current transaction. Is this correct? If this is the case, how would one use this in practice? If you want to prove things (either that a storage slot changed or a value) you need to upload the MPT proof to the tx (likely as calldata) to verify it in the EVM. This would thus alter the tx (because you alter calldata) and you would thus have to re-sign the tx. This is thus only practical for the builder to do. Is this correct or am I missing something?

Brechtpd: Oh good questions! Appreciate the feedback and points for clarifications, so don't worry. :)

So the root exposed is the transactions root, but yes I'm more interested in the actual state root so I see the confusion!

Ideally the latest state root at any point during EVM execution would be available. However, the state root calculation is very expensive and is already a bottleneck calculating it once per block (which is where the delayed state root calculation EIP comes in to help mitigate this). So this EIP obviously does not want to make things even worse by having the state root calculating multiple times during EVM execution (and then of course making it impossible to delay the state root calculation).

Instead this EIP only exposes the previous transactions in the current block to the EVM. This doesn't require expensive expensive state root calculations, and this can still be used to get the latest state by going from the pre block state and executing the previous transactions on top of that state to get to the state before the current transaction started executing.

> If this is not the case and it is actually the root of the previous transactions (so if we are in tx C then we thus apply the serialized A to key 0 and B to key 1) then we cannot know directly that C is the third (key 2) transaction. Because in order to prove this on-chain, we would need a proof that it is indeed at key 2, which means we a MPT proof which would thus contain both the serialized txs (so we can reconstruct the MPT in EVM and then see we are indeed at key 2) or in some cases hashes of the relevant MPT nodes.

Yes that is one way to do it. Another way would be to use non-inclusion proof to minimize the data needed. If the check is done through a proving system offchain these things don't matter that much.

> It seems that you want to supply the state root as it was right before executing the current transaction. Is this correct?

I want to make it possible to know the latest state in any way that is efficient.

> If this is the case, how would one use this in practice? 

In practice I would use it for [ULTRA TX](https://ethresear.ch/t/ultra-tx-programmable-blocks-one-transaction-is-all-you-need-for-a-unified-and-extendable-ethereum/21673) for synchronous composability with the L1 where I need to know the latest L1 state to be able to generate a proof that everything has executed correctly on top of the latest state.

> If you want to prove things (either that a storage slot changed or a value) you need to upload the MPT proof to the tx (likely as calldata) to verify it in the EVM. This would thus alter the tx (because you alter calldata) and you would thus have to re-sign the tx. This is thus only practical for the builder to do. Is this correct or am I missing something?

For my purposes no MPT data will be onchain, only a proof of correct execution and with the TXROOT as input to the proof. But yes, it would mostly be the builder (or sophisticated searchers) making use of this (which is already an assumption for my purposes).

--

Maybe one way to make this EIP more flexible is to step away from the merkle tree format so that the transaction data can be queried without merkle proofs and so there are fewer dependencies. For example, one could also expose the array of previous transactions directly through something like TX(index) that return the tx hash at the specified index. Maybe could be useful for other use cases though not really necessary for my own.

g11tech (changes requested):
from comments and reading of EIP it seems that it is still in process of being spec-ed in correct and useful manner, so will hold off review of the PR till that completes

PR #9032, opened 2024-11-15
Update EIP-7702: add "hash" to tuple, to allow single-signature for first eip-7702 tx initialization.

drortirosh:
Add an optional "hash" field to the auth tuple, to support single signature when submitted through a wrapper protocol (such as ERC-4337's UserOperation)

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -199,8 +199,8 @@ Specifically:
 ### A signed "hash" value:
 
 The auth tuple contains a `hash` value, which is unused by this EIP, but can be used to authenticate the wrapping transaction.
-This way, a single signature by the user can be used for both eip-7702 delegate and the outer UserOperation.
-It can also be used by a single signature to initialize the eip-7702 delegate and authenticate the "initcode" call
+This way, a single signature by the user can be used for both EIP-7702 delegate and the outer UserOperation.
+It can also be used by a single signature to initialize the EIP-7702 delegate and authenticate the "initcode" call
 
 ### Clearing Delegation Designations
 


lightclient (commented):
I don't really understand his change. How do you observe the hash from within the execution of the transaction?

drortirosh:
> I don't really understand this change. How do you observe the hash from within the execution of the transaction?

During ERC-4337 validation, the account's method is called with the signature 
`validateUserOp(UserOperation userop, bytes32 userOpHash, ...)`
The hash is a keccak over the userop packed structure (excluding the signature field)
The `userop.signature` field can be used to pass authentication info in addition to the actual signature.
For EIP-7702 signature, it should include the nonce and chainId.
Thus the validation function looks something like:
```solidity
function validateUserOp(UserOperation userop, bytes32 userOpHash) { 
   (bytes32 nonce, bytes32 chainid, bytes32 r, bytes32 s, uint8 v) = abi.decode(userop.signature, (bytes32,bytes32,bytes32,bytes32,uint8));
   bytes32 hash = keccak256(eip7702_rlp_encode(nonce, address(this), chainid, userOpHash));
   if ( address(this) == ecrecover(hash, v,r,s) ) return 0;
   return SIG_FAILED
}
```

This way we achieve the UX goal: a single signature by the end user for this transaction is used to validate both the eip-7702 tuple, and also outer transaction (erc-4337) for gas sponsorhip.

drortirosh:
Another use-case for hash: applying initcode to initialize a proxy:
We want to deploy a delegate, and then run some "initcode" to initialize it, without letting a frontrunner separate the two, and perfrom rogue initialization.
The normal flow for that would be a 2nd signature just for this initcode flow.
But using the hash in tuple, we can re-use the same eip-7702 signature to apply the code - so a user signs only once.
Now even if a front-runner ran the eip-7702 tuple in a separate transaction, it can't initialize it for itself, and our transaction will succeed to initialize our delegate (and pay slightly less, since the delegate was pre-installed by the front-runner..)

```solidity
_fallback() {
  if (_implementation() != address(0) {
    _delegate(_implementation());
    return;
  }
  //this is the first call, so calldata contains implementation, target calldata and everything to reuse eip7702 signature)
  (address impl, bytes memory data, bytes32 nonce, bytes32 chainid, bytes32 r, bytes32 s, uint8 v) = abi.decode(msg.data, 
  (address, bytes, bytes32,bytes32,bytes32,bytes32,uint8));
  bytes32 datahash = keccak256(abi.encodePacked(impl,data);
   bytes32 hash = keccak256(eip7702_rlp_encode(nonce, address(this), chainid, datahash));
   require(address(this) == ecrecover(hash, v,r,s) );
   _setImplementation(impl);
   _deletate(data);
}
```

Issue #146, opened 2016-09-02
EVM opcode: CALLGAS


axic:
This is split out of #28.
### Motivation

To allow contracts calculate the gas used effectively.
### Specification

A new opcode, ~~`TXGAS`~~ `CALLGAS` is introduced at `0x3d`. It returns the initial gas allocation for the call. It can be used together with `GAS` (`0x5a`) to calculate the gas used.

Note: this is different to #28 where the opcode `0x5c` was suggested.

chfast:
If this is about current call, I'd name it `MSGGAS`.

How is it better than storing the gas to the memory at the very beginning?

axic:
True! I think `CALLGAS` might be better fitting as we have `CALLVALUE` as well.

lastperson:
This will only solve part of the issue. We also need CALLREFUND counter that will tell how much gas is gathered for refund (storage cleaning) at this point of time.

PR #9047, opened 2024-11-22
Update EIP-7742: update the required EL headers and the gas fee mechanism

g11tech:
as discussed in the devcon meeting of core devs, it was discussed to attempt a pass a simple effective solution to handle the fee mechanism as we scale target blob counts : 

This PR will make sure that when we update the target, previously accumulated excess gas is already correctly scaled to a normalized target so the fees don't drop out when we scale the excess gas update fraction to keep up with the increasing target.

### DIFF
--- a/EIPS/eip-7742.md
+++ b/EIPS/eip-7742.md
@@ -2,7 +2,7 @@
 eip: 7742
 title: Uncouple blobs per block between CL and EL
 description: Have CL verify blob maximum and have EL get target value from CL
-author: Alex Stokes (@ralexstokes), Gajinder Singh (@g11tech), Bert (@bkellerman), Toni Wahrstätter (@nerolation), Ansgar Dietrichs (@adietrichs), Barnabé Monnot (@barnabemonnot)
+author: Alex Stokes (@ralexstokes), Gajinder Singh (@g11tech), Bert (@bkellerman), Toni Wahrstätter (@nerolation), Ansgar Dietrichs (@adietrichs), Barnabé Monnot (@barnabemonnot), Max Resnick (@MaxResnick)
 discussions-to: https://ethereum-magicians.org/t/eip-7742-uncouple-blob-count-between-cl-and-el/20550
 status: Review
 type: Standards Track
--- a/EIPS/eip-7742.md
+++ b/EIPS/eip-7742.md
@@ -112,7 +112,7 @@ def calc_excess_blob_gas(parent: Header) -> int:
 
   # note: multiplication of diff with EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS is before interger 
   # division by target_blobs_per_block to preserve sensitivity
-  return (normalized_parent_excess_blob_gas + ((parent.blob_gas_used - target_blob_gas) * EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS) // target_blobs_per_block)
+  return normalized_parent_excess_blob_gas + (parent.blob_gas_used - target_blob_gas) * EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS // target_blobs_per_block
 
 def get_base_fee_per_blob_gas(header: Header) -> int:
   return fake_exponential(
--- a/EIPS/eip-7742.md
+++ b/EIPS/eip-7742.md
@@ -116,10 +116,10 @@ def calc_excess_blob_gas(parent: Header) -> int:
 
 def get_base_fee_per_blob_gas(header: Header) -> int:
   return fake_exponential(
-    MIN_BASE_FEE_PER_BLOB_GAS,
-    header.excess_blob_gas,
-    BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED
-    )
+      MIN_BASE_FEE_PER_BLOB_GAS,
+      header.excess_blob_gas,
+      BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED
+  )
 ```
 
 Rest of the [EIP-4844](./eip-4844.md) specification is not changed
--- a/EIPS/eip-7742.md
+++ b/EIPS/eip-7742.md
@@ -82,7 +82,7 @@ The community feels that there is a need to increase the min blob base fee for a
 ### Block processing
 
 Upon activation of this EIP, execution clients **MUST** extend the header schema with an
-additional 64-bit field: the `target_blobs_per_block`. This values is set to the one provided by the CL during block production.
+additional 64-bit field: the `target_blobs_per_block`. This value is set to the one provided by the CL during block production.
 
 Any reference to `TARGET_BLOB_GAS_PER_BLOCK` from [EIP-4844](./eip-4844.md) (for e.g. for `excess_blob_gas` calculations) can be derived by taking the `target_blobs_per_block` from the EL block header and multiplying by `GAS_PER_BLOB`.
 
--- a/EIPS/eip-7742.md
+++ b/EIPS/eip-7742.md
@@ -1,8 +1,8 @@
 ---
 eip: 7742
-title: Uncouple blob count between CL and EL
+title: Uncouple blobs per block between CL and EL
 description: Have CL verify blob maximum and have EL get target value from CL
-author: Alex Stokes (@ralexstokes)
+author: Alex Stokes (@ralexstokes), Gajinder Singh (@g11tech), Bert (@bkellerman)
 discussions-to: https://ethereum-magicians.org/t/eip-7742-uncouple-blob-count-between-cl-and-el/20550
 status: Review
 type: Standards Track
@@ -13,7 +13,7 @@ requires: 4844
 
 ## Abstract
 
-Update blob maximum and target verification from [EIP-4844](./eip-4844.md).
+Update blob maximum, target, and blob gas fee computation from [EIP-4844](./eip-4844.md).
 
 The execution layer no longer verifies the blob maximum and receives the target dynamically from the consensus layer.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7742.md
```diff
@@ -33,67 +33,95 @@ This EIP also changes how the EL sources the current blob target value for two r
...
+def calc_excess_blob_gas(parent: Header) -> int:
+  # normalize parent's excess blob gas if this block was fork block
+  if(parent.timestamp < FORK_TIMESTAMP)
+    normalized_parent_excess_blob_gas = parent.excess_blob_gas * EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS // OLD_TARGET_BLOBS_PER_BLOCK
+    target_blobs_per_block = OLD_TARGET_BLOBS_PER_BLOCK
+  else
+    normalized_parent_excess_blob_gas = parent.excess_blob_gas
+    target_blobs_per_block = parent.target_blobs_per_block
+
+  return (normalized_parent_excess_blob_gas + (parent.blob_gas_used - target_blob_gas) * EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS // target_blobs_per_block)
```
bkellerman, line 104: I believe this can be simplified to 
`return (normalized_parent_excess_blob_gas + (parent.blob_gas_used - target_blob_gas) // target_blob_gas)`

With:
`NEW_BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED = BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED /128/GAS_PER_BLOB`

Then there is no need for `EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS ` since it's `=1` and also all the terms are in the same unit, blob gas.

@g11tech wdyt?

g11tech: because (parent.blob_gas_used - target_blob_gas) // target_blob_gas) will be most likely evaluate to 0 int and we would want error accumulation to be some sensitive range which is what the above equation does

EIPS/eip-7742.md
```diff
@@ -33,67 +33,99 @@ This EIP also changes how the EL sources the current blob target value for two r
...
-by EIP-4844 is unnecessary.
+The blob maximum is verified in the CL node and the EL inherits this verification during the consistency check of the versioned hashes corresponding to each blob as specified by the Engine API. Because of this, the strict check specified by EIP-4844 is unnecessary.
 
-#### Target amount of blobs per block
+For optimistically synced blocks from peer nodes, EL may further assume that the entire ancestor EL peer synced chain blocks correctly obeys this condition by the canonicality and correctness of the block they backfilling from as instructed by the CL.
 
-The target is currently specified as a fixed value in relation to the blob count. The Ethereum community intends to increase
-the blob parameters as part of its scaling strategy and the ability to have a more flexible target value in relation to
-the blob max is desirable to reduce rigidity in this protocol parameter.
+Hence we entirely deprecated the `MAX_BLOB_GAS_PER_BLOCK` checks from [EIP-4844](./eip-4844.md) in the EL while letting the maximum blobs per block constraints in the CL stay at `6` equivalent to `MAX_BLOB_GAS_PER_BLOCK`.
```
ralexstokes, line 51: I would say something like this as it is confusing when we later suggest raising the max below

```suggestion
Hence we entirely deprecated the `MAX_BLOB_GAS_PER_BLOCK` checks from [EIP-4844](./eip-4844.md) in the EL while letting the maximum blobs per block constraint to be checked by the CL.
```

EIPS/eip-7742.md
```diff
@@ -33,67 +33,99 @@ This EIP also changes how the EL sources the current blob target value for two r
...
+| `BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB` | `1112825` | used to calculate normalized update fraction for excess gas |
+| `BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED` | `142441600`| excess gas update fraction for the normalized target of `128` blobs |
+
+### Block processing
 
 Upon activation of this EIP, execution clients **MUST** extend the header schema with an
-additional 64-bit field: the `target_blobs_per_block`. This value is set to the current target blob count. The Engine API
-is modified along with this EIP to provide the `target_blobs_per_block` with each payload and implementations can use this
-value to correctly set the block header field.
+additional 64-bit field: the `target_blobs_per_block`. This values is set to the one provided by the CL during block production.
```
ralexstokes, line 85:
```suggestion
additional 64-bit field: the `target_blobs_per_block`. This value is set to the one provided by the CL during block production.
```

EIPS/eip-7742.md
```diff
@@ -33,67 +33,99 @@ This EIP also changes how the EL sources the current blob target value for two r
...
-### Block processing
 
-Upon activating this EIP (i.e. before processing any transactions),
-the verification of the blob maximum as given in EIP-4844 can be skipped. Concretely, this means any logic relating
-to `MAX_BLOB_GAS_PER_BLOCK` as given in EIP-4844 can be deprecated.
-Additionally, any reference to `TARGET_BLOB_GAS_PER_BLOCK` from EIP-4844 can be derived by taking the `target_blobs_per_block` from the CL and multiplying by `GAS_PER_BLOB` as given in EIP-4844.
+```python
+def calc_excess_blob_gas(parent: Header) -> int:
+  # normalize parent's excess blob gas if this block was fork block
+  if(parent.timestamp < FORK_TIMESTAMP)
```
ralexstokes, line 97: there is no way to deploy this rule before the fork and so this clause for `parent.timestamp < FORK_TIMESTAMP` would never be reachable, it should be removed

ralexstokes: if you want to do the conversion logic at the fork boundary, I think this should instead be `if parent.timestamp == FORK_TIMESTAMP`

siladu: At the fork boundary, I think it's possible for `parent.timestamp < FORK_TIMESTAMP` and `header.timestamp >= FORK_TIMESTAMP`.
This new calculation is activated at the fork but still operating on the parent block's data (which doesn't include target_blobs_per_block) so I think it makes sense to use the old hardcoded target here.

g11tech: ahh yes as @siladu pointed out, parent is pre 7742 and this logic runs for calculation of the excess gas for the current fork block

EIPS/eip-7742.md
```diff
@@ -33,67 +33,95 @@ This EIP also changes how the EL sources the current blob target value for two r
...
+
+#### Normalized `excess_gas`
+
+With the changing target, we also need to scale the `BLOB_BASE_FEE_UPDATE_FRACTION` from [EIP-4844](./eip-4844.md) accordingly to bound the price jumps by +-12.5%. But this introduces some irregularity whenever target gets updated because excess blob gas (accumulated via old target) is now effectively scaled down by this changed fraction in the fee calculations.
+
+To mitigate this irregularity, we now save a _normalized_ excess gas with respect to a fixed `EXCESS_BLOB_GAS_NORMALIZATION_TARGET` and use a corresponding `BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED` in the base fee calculations.
+
+#### Increased min blob base fee
+
+The community feels that there is a need to increase the min blob base fee for a fair blob market. To this end the base fee per blob gas will be increased to `MIN_BASE_FEE_PER_BLOB_GAS_EIP_7742`.
```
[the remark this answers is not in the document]
g11tech: yes, but may be i can just add @MaxResnick  as author too and we sort of compress this into this EIP if we want to include this, whatever the ACD thinks is the best approach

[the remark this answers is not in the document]
g11tech: removed from the PR

EIPS/eip-7742.md
ralexstokes: ah, I see what's going on now, thanks!

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+  # note: multiplication of diff with EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS is before interger 
+  # division by target_blobs_per_block to preserve sensitivity
+  return (normalized_parent_excess_blob_gas + ((parent.blob_gas_used - target_blob_gas) * EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS) // target_blobs_per_block)
+
+def get_base_fee_per_blob_gas(header: Header) -> int:
+  return fake_exponential(
+    MIN_BASE_FEE_PER_BLOB_GAS,
+    header.excess_blob_gas,
+    BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED
+    )
```
ralexstokes, line 122: formatting
```suggestion
  return fake_exponential(
      MIN_BASE_FEE_PER_BLOB_GAS,
      header.excess_blob_gas,
      BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED
  )
```

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+```
 
-Otherwise, the specification of EIP-4844 is not changed. For example, blob base fee accounting and excess blob gas tracking occur in the exact same way.
+Rest of the [EIP-4844](./eip-4844.md) specification is not changed
```
ralexstokes, line 125:
```suggestion
The rest of the [EIP-4844](./eip-4844.md) specification is not changed.
```

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+can be derived by taking the `target_blobs_per_block` from the EL block header and multiplying by `GAS_PER_BLOB`.
 
-When verifying a block, execution clients **MUST** ensure the target blob count in the block header matches the one
+Validity of these values is guaranteed from the consensus layer, much like how withdrawals are handled. Hence when
+verifying a block, execution clients **MUST** ensure the target blobs per block in the block header matches the one
 provided by the consensus client.
 
-For a genesis block with no existing parent, the value should be set according to the agreed specification for the
-target blob count given by that genesis block's protocol rule set.
+Furthermore `get_base_fee_per_blob_gas` as specified by [EIP-4844](./eip-4844.md) is as modified below:
```
ralexstokes, line 98:
```suggestion
Furthermore `calc_excess_blob_gas` and `get_base_fee_per_blob_gas` as specified by [EIP-4844](./eip-4844.md) is modified as below:
```

EIPS/eip-7742.md
```diff
@@ -13,7 +13,7 @@ requires: 4844
 
 ## Abstract
 
-Update blob maximum and target verification from [EIP-4844](./eip-4844.md).
+Update blob maximum, target, and blob gas fee computation from [EIP-4844](./eip-4844.md).
```
ralexstokes, line 16:
```suggestion
Update the verification of blob maximum and target, and blob gas fee computation from [EIP-4844](./eip-4844.md).
```

EIPS/eip-7742.md
```diff
@@ -33,22 +33,26 @@ This EIP also changes how the EL sources the current blob target value for two r
...
 
 #### Maximum blobs per block
 
 The blob maximum is verified in the CL node and the EL inherits this verification during the consistency check of the
 versioned hashes corresponding to each blob as specified by the Engine API. Because of this, the strict check specified
 by EIP-4844 is unnecessary.
 
-#### Target amount of blobs per block
+For optimistically synced blocks from peer nodes, EL may further assume that the entire ancestor EL peer synced chain
+blocks correctly obeys this condition by the canonicality and correctness of the block they backfilling from as instructed
```
ralexstokes, line 52:
```suggestion
blocks correctly obeys this condition by the canonicality and correctness of the block they backfill from as instructed
```

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
 Even if the EL keeps a fixed target value based on the max, removing the max implies the EL would not know what the target
 value should be. To address this lack of information, this EIP proposes the CL sends the current target value to the EL
 with each provided payload over the Engine API. The EL block header will also need to be extended with this target value
-to preserve the security of optimistic sync.
+(`target_blobs_per_block`) to preserve the security of optimistic sync.
+
+#### Normalized `excess_gas`
+
+With the changing target, we also need to scale the `BLOB_BASE_FEE_UPDATE_FRACTION` from [EIP-4844](./eip-4844.md)
```
ralexstokes, line 68:
```suggestion
To account for future target increases, we also need to scale the `BLOB_BASE_FEE_UPDATE_FRACTION` from [EIP-4844](./eip-4844.md)
```

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+accordingly to bound the price jumps by +-12.5%. But this introduces some irregularity whenever target gets updated because
+excess blob gas (accumulated via old target) is now effectively scaled down by this changed fraction in the fee calculations.
+
+To mitigate this irregularity, we now save a _normalized_ excess gas with respect to a fixed
+`EXCESS_BLOB_GAS_NORMALIZATION_TARGET` and use a corresponding `BLOB_BASE_FEE_UPDATE_FRACTION_NORMALIZED` in the base fee
```
ralexstokes, line 73: `EXCESS_BLOB_GAS_NORMALIZATION_TARGET` here does not match the defined constant with `_BLOB` suffix.

I think its cleaner to just call it `EXCESS_BLOB_GAS_NORMALIZATION_TARGET` but in either case the EIP should be consistent.

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+calculations.
 
 ## Specification
 
-### Block structure and validity
+| Constant | Value | Remarks|
+| - | - | - |
+| `OLD_TARGET_BLOBS_PER_BLOCK` | `3` | blob target corresponding to [EIP-4844](./eip-4844.md) |
```
ralexstokes, line 80: nitpick, but I think better to call this `TARGET_BLOBS_PER_BLOCK_4844` as "old" is a relative term and may not age as well across protocol development

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+    # integer division by OLD_TARGET_BLOBS_PER_BLOCK to preserve sensitivity
+    normalized_parent_excess_blob_gas = (parent.excess_blob_gas * EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS) // OLD_TARGET_BLOBS_PER_BLOCK
+    target_blobs_per_block = OLD_TARGET_BLOBS_PER_BLOCK
+  else
+    normalized_parent_excess_blob_gas = parent.excess_blob_gas
+    target_blobs_per_block = parent.target_blobs_per_block
+
+  # note: multiplication of diff with EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS is before interger 
+  # division by target_blobs_per_block to preserve sensitivity
+  return (normalized_parent_excess_blob_gas + ((parent.blob_gas_used - target_blob_gas) * EXCESS_BLOB_GAS_NORMALIZATION_TARGET_BLOBS) // target_blobs_per_block)
```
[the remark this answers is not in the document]
g11tech: what if someone applies BODMAS? wanted to imply multiplication first and int division later for deterministic results

EIPS/eip-7742.md
g11tech: yea was remanent from an old iteration, will fix

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
-### Block processing
 
-Upon activating this EIP (i.e. before processing any transactions),
-the verification of the blob maximum as given in EIP-4844 can be skipped. Concretely, this means any logic relating
-to `MAX_BLOB_GAS_PER_BLOCK` as given in EIP-4844 can be deprecated.
-Additionally, any reference to `TARGET_BLOB_GAS_PER_BLOCK` from EIP-4844 can be derived by taking the `target_blobs_per_block` from the CL and multiplying by `GAS_PER_BLOB` as given in EIP-4844.
+```python
+def calc_excess_blob_gas(parent: Header) -> int:
+  # normalize parent's excess blob gas if this block was fork block
+  if(parent.timestamp < FORK_TIMESTAMP)
```
ralexstokes, line 104:
```suggestion
  if parent.timestamp < FORK_TIMESTAMP:
```

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+    # note multiplication of parent excess blob gas with EXCESS_BLOB_GAS_NORMALIZATION_TARGET is before
+    # integer division by TARGET_BLOBS_PER_BLOCK_4844 to preserve sensitivity
+    normalized_parent_excess_blob_gas = parent.excess_blob_gas * EXCESS_BLOB_GAS_NORMALIZATION_TARGET // TARGET_BLOBS_PER_BLOCK_4844
+    target_blobs_per_block = TARGET_BLOBS_PER_BLOCK_4844
+  else
```
ralexstokes, line 109:
```suggestion
  else:
```

EIPS/eip-7742.md
```diff
@@ -57,43 +61,78 @@ the blob max is desirable to reduce rigidity in this protocol parameter.
...
+    normalized_parent_excess_blob_gas = parent.excess_blob_gas
+    target_blobs_per_block = parent.target_blobs_per_block
+
+  # note: multiplication of diff with EXCESS_BLOB_GAS_NORMALIZATION_TARGET is before interger 
+  # division by target_blobs_per_block to preserve sensitivity
```
ralexstokes, line 114:
```suggestion
  # division by target_blobs_per_block to preserve sensitivity
  target_blob_gas = target_blobs_per_block * GAS_PER_BLOB
```

petertdavies:
I think the problem with this EIP is that it splits the blob fee pricing between the EL and the CL. The `target_blobs_per_block` is set on the CL, but the update fraction is set on the EL. In practice the two values are tightly linked in practice.

I think it would make the most sense for the CL to tell the EL `blob_base_fee` and `max_blobs_per_block` and do all the math on the CL. That way the CL can change blobonomics as it sees fit without affecting the EL.

ralexstokes:
Decided on [ACDC #146](https://github.com/ethereum/pm/issues/1200) to table this until Fusaka.

Can likely close this PR.

g11tech:
closing in favor of a combined full proposal here: https://github.com/ethereum/EIPs/pull/9062

Magkoooh (approved):

Magkoooh (approved):

PR #9055, opened 2024-11-25
Update EIP-6206: clarify wording on JUMPF (EOF)

pdobacz:
`Update EIP-6206: clarify "JUMPF prepares the delta` raised in https://github.com/ethereum/solidity/pull/15550#issuecomment-2494801436

The other commit found when re-reading. The [megaspec](https://github.com/ipsilon/eof/blob/main/spec/eof.md?plain=1#L286) already has this "iff" formulation

### DIFF
--- a/EIPS/eip-6206.md
+++ b/EIPS/eip-6206.md
@@ -52,7 +52,7 @@ Let the definition of `type[i]` be inherited from [EIP-4750](./eip-4750.md) and 
   * either `type[current_section_index].outputs` MUST be greater or equal `type[target_section_index].outputs`,
   * or `type[target_section_index].outputs` MUST be `0x80`
 * The stack height validation at `JUMPF` depends on whether the target section is non-returning:
-  * `JUMPF` into returning section (`type[target_section_index].outputs` does not equal `0x80`):  `stack_height_min` and `stack_height_max` MUST be equal to `type[current_section_index].outputs + type[target_section_index].inputs  - type[target_section_index].outputs`. This means that target section can output less stack elements than the original code section called by the top element on the return stack, if the current code section leaves the delta `type[current_section_index].outputs - type[target_section_index].outputs` element(s) on the stack.
+  * `JUMPF` into returning section (`type[target_section_index].outputs` does not equal `0x80`):  `stack_height_min` and `stack_height_max` MUST be equal to `type[current_section_index].outputs + type[target_section_index].inputs  - type[target_section_index].outputs`.
   * `JUMPF` into non-returning section (`type[target_section_index].outputs` equals `0x80`): `stack_height_min` MUST be greater than or equal to `type[target_section_index].inputs`.
 * Stack overflow check at `JUMPF`: `stack_height_max` MUST be less than or equal to `1024 - types[target_section_index].max_stack_height + types[target_section_index].inputs`.
 * `JUMPF` is considered terminating instruction, i.e. does not have successor instructions in code validation and MAY be final instruction in the section. 
@@ -60,7 +60,7 @@ Let the definition of `type[i]` be inherited from [EIP-4750](./eip-4750.md) and 
 
 `CALLF` instruction validation is extended to include the rule:
 
-* Code section is invalid in case an immediate argument `target_section_index` of any `CALLF` targets a non-returning section, i.e. `type[target_section_index]` equals `0x80`.
+* Code section is invalid in case an immediate argument `target_section_index` of any `CALLF` targets a non-returning section, i.e. `type[target_section_index].outputs` equals `0x80`.
 
 #### Non-returning status validation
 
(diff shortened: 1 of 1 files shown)


shemnon (approved):

EIPS/eip-6206.md
```diff
@@ -52,26 +52,26 @@ Let the definition of `type[i]` be inherited from [EIP-4750](./eip-4750.md) and
...
 *Note: This implies that section containing only `JUMPF`s into non-returning sections is non-returning itself.*
 
 ## Rationale
 
 ### Allowing `JUMPF` to section with less outputs
 
-As long as `JUMPF` prepares the delta `type[current_section_index].outputs - type[target_section_index].outputs` stack elements before changing code sections, it is possible to jump to a section with less outputs than was originally entered via `CALLF`. This will reduce duplicated code as it will allow compilers more flexibility during code generation such that certain helpers can be used generically by functions, regardless of their output values.
+A code section with a `JUMPF` can, just before that instruction, have a delta of `type[current_section_index].outputs - type[target_section_index].outputs` output elements on the operand stack (i.e. counted on top of the target section inputs!). This makes it possible to `JUMPF` to a section with less outputs than those of the current section. This will reduce duplicated code as it will allow compilers more flexibility during code generation such that certain helpers can be used generically by functions, regardless of their output values.
```
gumb0, line 74: > (i.e. counted on top of the target section inputs!).

This is quite confusing, output elements are _below_ target section inputs, right? I would just remove the part in parenthesis.

gumb0, line 74:
```suggestion
As long as code section with `JUMPF` has the delta `type[current_section_index].outputs - type[target_section_index].outputs` stack elements before changing code sections, it is possible to `JUMPF` to a section with less outputs than those of the current section. This will reduce duplicated code as it will allow compilers more flexibility during code generation such that certain helpers can be used generically by functions, regardless of their output values.
```

EIPS/eip-6206.md
```diff
@@ -52,26 +52,26 @@ Let the definition of `type[i]` be inherited from [EIP-4750](./eip-4750.md) and
   * either `type[current_section_index].outputs` MUST be greater or equal `type[target_section_index].outputs`,
   * or `type[target_section_index].outputs` MUST be `0x80`
 * The stack height validation at `JUMPF` depends on whether the target section is non-returning:
-  * `JUMPF` into returning section (`type[target_section_index].outputs` does not equal `0x80`):  `stack_height_min` and `stack_height_max` MUST be equal to `type[current_section_index].outputs + type[target_section_index].inputs  - type[target_section_index].outputs`. This means that target section can output less stack elements than the original code section called by the top element on the return stack, if the current code section leaves the delta `type[current_section_index].outputs - type[target_section_index].outputs` element(s) on the stack.
```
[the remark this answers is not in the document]
pdobacz: Heh, OK, I think you're right again... having spent some time with the rationale section, I'm starting to see it's maybe rationale section repeating the spec, and this sentence is ok to stay here as clarification...

EIPS/eip-6206.md
pdobacz: ok @gumb0 PTAL now. I brought back the sentence in the spec you've requested, but gave the rationale piece a less redundant sound and hopefully more to the point of why we're having this.

EIPS/eip-6206.md
```diff
@@ -52,26 +52,28 @@ Let the definition of `type[i]` be inherited from [EIP-4750](./eip-4750.md) and
...
 *Note: This implies that section containing only `JUMPF`s into non-returning sections is non-returning itself.*
 
 ## Rationale
 
 ### Allowing `JUMPF` to section with less outputs
 
-As long as `JUMPF` prepares the delta `type[current_section_index].outputs - type[target_section_index].outputs` stack elements before changing code sections, it is possible to jump to a section with less outputs than was originally entered via `CALLF`. This will reduce duplicated code as it will allow compilers more flexibility during code generation such that certain helpers can be used generically by functions, regardless of their output values.
+An alternative rule for `JUMPF` stack validation could require the target section's outputs to be exactly equal to the current section's outputs. Under such rule, a particular target section (a shared "helper" piece of code) would only "match" sections (requiring some shared "helper" code to execute before returning) *of the same outputs*.
```
gumb0, line 74:
```suggestion
An alternative rule for `JUMPF` stack validation could require the target section's outputs to be exactly equal to the current section's outputs. Under such rule, a particular target section (a shared "helper" piece of code) would only "match" sections (requiring some shared "helper" code to execute before returning) *with the same number of outputs*.
```

not sure actually if the second sentence here adds anything to understanding

pdobacz: it is a bit redundant, but describing the exact problem with the alternative helps (at least me) to follow the reasoning more easily - it introduces the second sentence of the (now) second paragraph about code gen. It also speaks to the concept of helpers from the Motivation section, bridging it to the gist of the "delta spec".

So the narrative of the 4 sentences is:
- what is the alternative spec
- why is it bad
- what is the chosen spec
- why is it good

gumb0 (approved):

EIPS/eip-6206.md
cameel: It does sound clearer to me so I think it's an improvement.

PR #9058, opened 2024-11-27
Update EIP-7801: Eip 7801 bitmask

garyschulte:
Update specification to use bitmask rather than an updating bitlist, so nodes will not need to update their ENR as the blockchain progresses across shard boundaries.  

Also change shard definition to 100k per 1M block span.  This makes data availability easy to reason about, and straightforward for a human to determine.

### DIFF
--- a/EIPS/eip-7801.md
+++ b/EIPS/eip-7801.md
@@ -53,7 +53,7 @@ This EIP could be implemented without a new protocol version by adding a new fie
 
 The bitmask approach provides a flexible means to represent and retain block data while committing to future spans. This mechanism aligns with the pruning proposed in EIP-4444, while ensuring that historical and future data spans remain available across the network.
 
-The bitmask approach is already used in the Consensus Layer for attestation subnets, making it a familiar and efficient method for representing data spans. Additionally, committing to future spans ensures better predictability and stability for data locality.
+A similar bitlist approach is already used in the Consensus Layer for attestation subnets, making it a familiar and efficient method for representing data spans. Additionally, committing to future spans ensures better predictability and stability for data locality.
 
 ## Backwards Compatibility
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7801.md
```diff
@@ -1,7 +1,7 @@
 ---
 eip: 7801
 title: eth/70 - Sharded Blocks Protocol
-description: Replaces block range with a bitlist representing 1-million-block spans in the handshake, with probabilistic shard retention
+description: Replaces block range with a bitmask representing 100,000-block spans in the handshake, committing to future block ranges
 author: Ahmad Bitar (@smartprogrammer93) <smartprogrammer@windowslive.com>, Giulio Rebuffo (@Giulio2002)
```
smartprogrammer93, line 5: Please add yourself as author if you would like

EIPS/eip-7801.md
```diff
@@ -13,45 +13,47 @@ requires: 7642
 
 ## Abstract
 
-This EIP introduces a method enabling an Ethereum node to communicate its available block spans via a bitlist, where each bit represents a 1-million-block span. Nodes use this bitlist to signal which spans of historical data they store, enabling peers to make informed decisions about data availability when requesting blocks. This aims to improve network efficiency by providing a probabilistic snapshot of data locality across nodes.
+This EIP introduces a method enabling an Ethereum node to communicate its available block spans via a bitmask, where each bit represents a 100_000-block span within each 1_000_000 block range of chain history. Nodes use this bitmask to signal which spans of historical data they store and commit to storing future spans as they are created.  This enables peers to make informed decisions about data availability without first connecting and querying for availability. The bitmask repeats every 1 million blocks, making for a somewhat simple to reason about probability of data availability.
```
smartprogrammer93, line 16:
```suggestion
This EIP introduces a method enabling an Ethereum node to communicate its available block spans via a bitmask, where each bit represents a 100_000-block span within each 1_000_000 block range of chain history. Nodes use this bitmask to signal which spans of historical data they store and commit to storing future spans as they are created. This enables peers to make informed decisions about data availability without first connecting and querying for availability. The bitmask repeats every 1 million blocks, making for a somewhat simple to reason about probability of data availability.
```

EIPS/eip-7801.md
```diff
@@ -13,45 +13,47 @@ requires: 7642
...
 
 ### ENR variation
 
-This EIP could be implemented without a new protocol version by adding a new field to the Ethereum Node Record (ENR) to advertise the bitlist. This would allow nodes to advertise their available block spans without requiring a handshake. However, this approach would not provide the same level of assurance as the handshake, as the ENR is not authenticated. Also, it would make the discovery process much simpler, as nodes just would need to take a look at the ENR to determine if a peer has the data they are looking for.
+This EIP could be implemented without a new protocol version by adding a new field to the Ethereum Node Record (ENR) to advertise the bitmask. This would allow nodes to advertise their available block spans without requiring a handshake. However, this approach would not provide the same level of assurance as the handshake, as the ENR is not authenticated. Also, it would make the discovery process much simpler, as nodes just would need to take a look at the ENR to determine if a peer has the data they are looking for.
```
smartprogrammer93, line 50:
```suggestion
This EIP could be implemented without a new protocol version by adding a new field to the Ethereum Node Record (ENR) to advertise the bitmask. This would allow nodes to advertise their available block spans without requiring a handshake. However, this approach would not provide the same level of assurance as the handshake, as the ENR is not authenticated. Also, it would make the discovery process much simpler, as nodes would only need to check the ENR to determine if a peer has the data they need.
```

smartprogrammer93 (approved):

PR #9062, opened 2024-11-30
Update EIP-7762: Move to Draft

nerolation:
As discussed in [ACDC 146](https://github.com/ethereum/pm/issues/1200), we may want to split the blob target increase from the base fee mechanism changes. Those include:

* Increasing the constant min base fee and reseting excess gas
* Normalizing excess gas over target blob gas
* Automate blob base fee update fraction updates
* (optional) Introduce scaling factor for excess gas to ensure symmetric base fee scaling

As of yesterday's call, it looks like we won't ship these updates in Pectra, but in Fusaka.

### DIFF
--- a/EIPS/eip-7762.md
+++ b/EIPS/eip-7762.md
@@ -1,10 +1,10 @@
 ---
-eip: 7762
-title: Increase MIN_BASE_FEE_PER_BLOB_GAS
-description: Adjust the MIN_BASE_FEE_PER_BLOB_GAS to speed up price discovery on blob space
-author: Max Resnick (@MaxResnick)
-discussions-to: https://ethereum-magicians.org/t/eip-7762-increase-min-base-fee-per-blob-gas/20949
-status: Review
+eip: 9999
+title: Adjust blob base fee mechanism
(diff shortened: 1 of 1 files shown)


EIPS/eip-7762.md
```diff
@@ -13,56 +13,98 @@ requires: 4844
...
 
 ```diff
 + MIN_BASE_FEE_PER_BLOB_GAS  =  2**25
 - MIN_BASE_FEE_PER_BLOB_GAS = 1
 ```
 
-## Rationale
-
-The current MIN_BASE_FEE_PER_BLOB_GAS is 1 wei. This is many orders of magnitude lower than the prevailing price of blobs when blobs enter price discovery. Whenever demand for blobs exceeds supply, blobs enter price discovery, but traversing the 8 orders of magnitude between 1 wei and the point where elasticity of demand starts to decrease takes a long time.
+Furthermore, to avoid sudden changes in the blob base fee, the excess gas MUST be reset to 0 at the time of the fork.
```
marioevz, line 33: This part alone requires a lot of testing: The t8n interface does not support a way of us telling "this is the fork block", so to work around it, we need to send hard-coded values in EEST at fork time, which requires framework changes.

PR #9090, opened 2024-12-05
Update EIP-7762: add excess gas reset

adietrichs:
Reset the excess blob gas to 0, to avoid a blob basefee spike.

Todo (in followup PR): add rationale.


MaxResnick (approved):
LGTM

PR #9094, opened 2024-12-05
Update EIP-7600: Add EIPs 7623 & 7691

timbeiko:
EIP-7691 was SFI'd on [ACDC#146](https://github.com/ethereum/pm/issues/1200) and EIP-7623 was SFI'd on [ACDE#201](https://github.com/ethereum/pm/issues/1197)

PR #9095, opened 2024-12-05
Update EIP-7691: Move to Review

timbeiko:
Currently blocking https://github.com/ethereum/EIPs/pull/9094


SamWilsn (changes requested):
You'll want to replace the `<!-- todo -->` comments with the actual values before moving into Review.

timbeiko:
@SamWilsn that comment was for the fork block, which we don't include until it gets set in ACD. Is the PR now OK to merge? Want to make sure we can properly reflect the Pectra scope in https://github.com/ethereum/EIPs/pull/9094

g11tech (dismissed):
i think we need to have another tag to track things to be filled before moving to final 

may be <-- TBA --> or <-- TBF --> can be used to move to review

SamWilsn (dismissed):

SamWilsn (approved):
I guess `PECTRA_FORK_EPOCH` can stay undefined :rofl:

PR #9120, opened 2024-12-09
Update EIP-7251: replace check_fee with get_fee check

protolambda:
There is no `check_fee` defined in the EIP, the reference in the text was probably referring to older pseudo code.

This changes the reference to `msg.value >= get_fee()`, a short version of the pseudo code in the EIP, part of `add_consolidation_request`:
```
    # Verify sufficient fee was provided.
    fee = get_fee()
    require(msg.value >= fee, 'Insufficient value for fee')
```


lightclient (approved):

lightclient:
thank you!

PR #9121, opened 2024-12-09
Update EIP7623 to reflect correct invalid cost

jochem-brouwer:
EIP 7623 currently marks transactions as "invalid" (these cannot be included in a block) for `21000 + TOTAL_COST_FLOOR_PER_TOKEN * tokens_in_calldata`. This does not reflect the case if a contract is being created, which should also bump this cost.

### DIFF
--- a/EIPS/eip-7623.md
+++ b/EIPS/eip-7623.md
@@ -60,7 +60,7 @@ tx.gasUsed = {
     )
 ```
 
-Any transaction with a gas limit below `21000 + TOTAL_COST_FLOOR_PER_TOKEN * tokens_in_calldata` is considered invalid. This limitation exists because transactions must cover the floor price of their calldata without relying on the execution of the transaction. There are valid cases where `gasUsed` will be below this floor price, but the floor price needs to be reserved in the transaction gas limit.
+Any transaction with a gas limit below `21000 + isContractCreation * (32000 + InitCodeWordGas * words(calldata)) + TOTAL_COST_FLOOR_PER_TOKEN * tokens_in_calldata` is considered invalid. This limitation exists because transactions must cover the floor price of their calldata without relying on the execution of the transaction. There are valid cases where `gasUsed` will be below this floor price, but the floor price needs to be reserved in the transaction gas limit.
 
 ## Rationale
 


nerolation:
Putting the contract-related cost into the pre-execution validation would cause transactions to be marked as invalid that would have actually been possible to execute.
Note that a tx either pays the floor cost, OR it pays the standard price for execution, calldata + contract-related things. So the floor cost already ensures that enough gas is available for the tx to not run out of gas after execution.
We didn't add access and authorization list costs because it belongs to execution and storage access cost which are part of evm gas costs here.
If I got you wrong, could you elaborate with an example, e.g. _tx x has y amount of calldata, deploys a contract and uses z amount of gas for evm_.

jochem-brouwer:
I think my main confusion comes from this text:

> Any transaction with a gas limit below `21000 + TOTAL_COST_FLOOR_PER_TOKEN * tokens_in_calldata` is considered invalid. This limitation exists because transactions must cover the floor price of their calldata without relying on the execution of the transaction. There are valid cases where `gasUsed` will be below this floor price, but the floor price needs to be reserved in the transaction gas limit.

If you mean that this is an extra check beside the "normal" check, for instance if a tx creates a contract it needs at least a gas limit of 21k + 32k gas to be even considered for inclusion. 

The simplest examples would be a transaction which would create a contract with no data. If I follow the current text it seems to imply that a 21k gas limit is OK here (because there are 0 `tokens_in_calldata`), but this is (obviously?) incorrect.

I think the text might need elaboration that this is an extra check. For instance, EELS has currently (in an open PR) implemented this specific "min gas limit" rule as: https://github.com/ethereum/execution-specs/pull/966/files#diff-1ead0e19bcf5b72f6e0ff1f07a9d44a25a738020803d5ea42839da1e3d58a8c5R400

I should likely edit this PR and include an elaboration rule here, upon thought I agree that including contract creation cost (+ other costs like 2930 access list cost / 7702 authority list cost) does not belong here, but rather this extra check that the "minimum tx gas limit" is thus the maximum of those two calculations?

nerolation:
Lgtm

nerolation (approved):

PR #9124, opened 2024-12-10
Update EIP-4762: clarify change in behaviour for CALLCODE cost if value transfer

lu-pinto:
I think the removal of CALLCODE cost for value transfer needs to be clarified and I didn't find any reference to it in the EIP. I remember that Besu bumped into this change and we actually applied the change to previous forks as well causing the CI pipeline to fail later on.

Even though it makes sense to not charge anything since CALLCODE transfers to itself, that's not the case with previous forks so this is a change in behaviour which IMO is worth documenting. 

cc @gballet @jsign

### DIFF
--- a/EIPS/eip-4762.md
+++ b/EIPS/eip-4762.md
@@ -193,6 +193,7 @@ Remove the following gas costs:
  * Increased gas cost of `CALL` if it is nonzero-value-sending
  * [EIP-2200](./eip-2200.md) `SSTORE` gas costs except for the `SLOAD_GAS`
  * 200 per byte contract code cost
+ * static `CALLCODE` cost if it is nonzero-value-sending
 
 Reduce gas cost:
 


NaNOteh (approved):

gballet (commented):
I have a few issues with this PR:

 * Why is `CALLCODE` special, all other static costs are also removed ( and have been since 2929/2930, this is not a 4762-specific change)
 * What other forks are you referring to ? Do you mean that value costs for value transfers no longer apply? We could clarify this if it makes sense, but what is the connection with the static costs?

lu-pinto:
> I have a few issues with this PR:
> 
>     * Why is `CALLCODE` special, all other static costs are also removed ( and have been since 2929/2930, this is not a 4762-specific change)
> 
>     * What other forks are you referring to ? Do you mean that value costs for value transfers no longer apply? We could clarify this if it makes sense, but what is the connection with the static costs?

OK maybe static is not the right word, we can remove that. But current mainnet charges 9000 gas for a value transfer in CALLCODE and in EIP4762 these charges have been removed, that's what I meant.

gballet (approved):
LGTM, we discussed it and agreed that this was we should do.

PR #9129, opened 2024-12-12
Add blob config eip

lightclient:
Adds a new object to the EL genesis chain config:

of the shape:

```json
"blobSchedule": {
  "prague": { "target": 6, "max": 9 },
  "osaka":   { "target": 12, "max": 16},
  ...
}
```

### DIFF
--- a/EIPS/blob-config.md
+++ b/EIPS/blob-config.md
@@ -1,0 +1,69 @@
+---
+title: Add blob schedule to execution client configuration files
+description: Include a per-fork schedule of max and target blob counts in client configuration files
+author: lightclient (@lightclient)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-12-12
+---
+
+
+## Abstract
+
+Add a new object to client configuration files `blobSchedule` which lists the
+target blob count per block and max blob count per block for each fork.
+
+## Motivation
+
+- ensure there is a way to dynamically adjust the target and max blob counts per
+  block
+- avoid complex handshake over engine API
+
+## Specification
+
+Extend the client configuration files with the object `blobSchedule` with the
+following shape:
+
+```json
+"blobSchedule": {
+  "cancun": {
+    "target": 3,
+    "max": 6
+  },
+  "prague": {
+    "target": 6,
+    "max": 9
+  }
+}
+```
+
+When there is no explicit configuration for the current fork, use the last
+specified fork value. If last value is specified, set both to zero.
+
+## Rationale
+
+Although maintaining the target and max blob just in the consensus client is
+desirable, we that the reality is execution clients need these values for
+various core protocol and peripheral activities. For example, the
+`eth_feeHistory` RPC method returns a field `blobGasUsedRatio` that does require
+the max, even though the core protocol doesn't specifically need such value.
+Passing this value over the engine API every block seem overkill so we believe a
+configuration value is a good middle ground.
+
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Test Cases
+
+TODO
+
+## Security Considerations
+
+Needs discussion.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE.md).


EIPS/blob-config.md
```diff
@@ -0,0 +1,69 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7840
```

EIPS/blob-config.md
```diff
@@ -0,0 +1,69 @@
...
+title: Add blob schedule to execution client configuration files
+description: Include a per-fork schedule of max and target blob counts in client configuration files
+author: lightclient (@lightclient)
+discussions-to: https://ethereum-magicians.org/t/add-blob-schedule-to-execution-client-configuration-files/22182
```
abcoathup, line 5:
```suggestion
discussions-to: https://ethereum-magicians.org/t/eip-7840-add-blob-schedule-to-execution-client-configuration-files/22182
```
Added assigned number to Eth Magicians

EIPS/eip-7840.md
```diff
@@ -0,0 +1,70 @@
...
+Passing this value over the engine API every block seem overkill so we believe a
+configuration value is a good middle ground.
+
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Test Cases
+
+TODO
```
g11tech, line 62: <-- TODO --> tag for tracking via bot

EIPS/eip-7840.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Security Considerations
+
+Needs discussion.
```
g11tech, line 66: <-- TODO --> tag

EIPS/eip-7840.md
```diff
@@ -0,0 +1,70 @@
...
+}
+```
+
+When there is no explicit configuration for the current fork, use the last
+specified fork value. If last value is specified, set both to zero.
+
+## Rationale
+
+Although maintaining the target and max blob just in the consensus client is
+desirable, we that the reality is execution clients need these values for
```
jochem-brouwer, line 49: This sentence does not read well

MarekM25 (commented):
Nethermind team is okay with this change. Signaling support 👍

EIPS/eip-7840.md
lightclient: fixed thanks

siladu:
Besu is ok with this change to get us through Prague and have a prototype implementation ready https://github.com/hyperledger/besu/pull/8042

marioevz:
I prototyped the test changes required in this PR: https://github.com/ethereum/execution-spec-tests/pull/1040

Feedback is appreciated.

klkvr:
Should we configure update fraction in this object as well? Given that EIP-7691 defines different update fractions for Cancun and Prague it seems that it is reasonable to handle it in the same way as target/max blob count?

g11tech:
i think in interop call was discussed that upgrade fraction should also be provided with the config butt it only matters w.r.t. to handle the max swing on the asymmetry schedule (drop is bigger on empty blobs compared to full blobs block)

7691 already modifies that for 6/9 so not sure if we need to provide it here. although there could be later EIPs to deal with assymetry (like normalization) so we don't really need it.

parithosh:
> i think in interop call was discussed that upgrade fraction should also be provided with the config butt it only matters w.r.t. to handle the max swing on the asymmetry schedule (drop is bigger on empty blobs compared to full blobs block)
> 
> 7691 already modifies that for 6/9 so not sure if we need to provide it here. although there could be later EIPs to deal with assymetry (like normalization) so we don't really need it.

However, without the update fraction being specified in the genesis files - we can't properly test future blob increases on testnets (atleast not with proper gas markets). E.g if we set 10/20 as the blob target/max on pectra (in order to do some load testing), we will have a situation where the wrong update fraction is applied for this blob limit. Allowing us to specify everything in the genesis file will fix this.

g11tech:
> However, without the update fraction being specified in the genesis files - we can't properly test future blob increases on testnets (atleast not with proper gas markets). E.g if we set 10/20 as the blob target/max on pectra (in order to do some load testing), we will have a situation where the wrong update fraction is applied for this blob limit. Allowing us to specify everything in the genesis file will fix this.

this EIP itself is baffling, 

1. so do we expect node runners to be able to play with these values for their clients like how they signal change to the gaslimit? if this is the case one should only be able to change their "soft max" locally which only impose limit on amx blobs their client can include

2. or this is just a testing tool? in that case why is this an EIP, it will never be activated in mainnet?

if this

parithosh:
> > However, without the update fraction being specified in the genesis files - we can't properly test future blob increases on testnets (atleast not with proper gas markets). E.g if we set 10/20 as the blob target/max on pectra (in order to do some load testing), we will have a situation where the wrong update fraction is applied for this blob limit. Allowing us to specify everything in the genesis file will fix this.
> 
> this EIP itself is baffling,
> 
>     1. so do we expect node runners to be able to play with these values for their clients like how they signal change to the gaslimit? if this is the case one should only be able to change their "soft max" locally which only impose limit on amx blobs their client can include
> 
>     2. or this is just a testing tool? in that case why is this an EIP on core track, it will never be activated in mainnet? may be this could be an informational EIP for client testnet tooling that goes into "living status" and not "final" status

For 1. I don't imagine node runners using any signaling as the canonical files for mainnet are coded into the client implementations. This is mostly just for testnets and testing purposes that we care about specifying the values instead of just hardcoding it.
2. I would agree that its an informational, but its @lightclient 's EIP so i'd let him justify that bit

terencechain (approved):

EIPS/eip-7840.md
```diff
@@ -0,0 +1,70 @@
+---
+eip: 7840
+title: Add blob schedule to EL config files
+description: Include a per-fork schedule of max and target blob counts in client configuration files
+author: lightclient (@lightclient)
+discussions-to: https://ethereum-magicians.org/t/add-blob-schedule-to-execution-client-configuration-files/22182
+status: Draft
+type: Standards Track
+category: Core
```
g11tech, line 9:
```suggestion
type: Informational
```

timbeiko:
@g11tech can we please merge this now that the track has been changed, as we've already accepted to include it in the fork. Thanks!

jochem-brouwer:
@timbeiko Side-topic, it would be new (?) for informational EIPs to become part of a hardfork spec? (Because this is not part of consensus and thus does not require a fork)

timbeiko:
@jochem-brouwer we agreed to put non-Core EIPs in hardfork spec for clarity (e.g. PeerDAS isn't a Core EIP), but I haven't formalized that anywhere. Need to open a PR to EIP-7723 to mention it 😄

PR #9136, opened 2024-12-14
Update EIP-7702: Add security consideration about use of storage

frangio:
Also applies a minor suggestion posted in https://github.com/ethereum/EIPs/pull/8969 after it was merged.


frangio:
This is unfortunately tripping into linting/CI errors from the EIP to ERC migration.

lightclient (commented):
LGTM - thanks!

PR #9143, opened 2024-12-16
Update EIP-7702: make chain id u256

lightclient:
There has been a comment that there are efforts which will expect chain id to be u256. I think we should make this change to 7702. https://ethereum-magicians.org/t/eip-7702-set-eoa-account-code/19923/326?u=matt


lightclient:
agree to go in for devnet-5 on acd 202

PR #9144, opened 2024-12-16
Update EIP-2935: bring up to date with sys contract impl

lightclient:
This makes a few updates based on audits of system contract code here: https://github.com/lightclient/sys-asm/blob/main/src/execution_hash/main.eas

* revert instead of return 0 when request is out of bounds
* require `get` input to be 32 bytes
* set history server window to 8191 instead of 8192 - [rationale same as eip-4788](https://eips.ethereum.org/EIPS/eip-4788#size-of-ring-buffers)


EIPS/eip-2935.md
```diff
@@ -26,11 +26,11 @@ A side benefit of this approach could be that it allows building/validating proo
...
 | Parameter | Value |
 | - | - |
-| `FORK_TIMESTAMP` | TBD |
-| `BLOCKHASH_SERVE_WINDOW` | `256` |
-| `HISTORY_SERVE_WINDOW` | `8192` |
-| `SYSTEM_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe` |
-| `HISTORY_STORAGE_ADDRESS` | `0x0aae40965e6800cd9b1f4b05ff21581047e3f91e`|
+| `FORK_TIMESTAMP`          | TBD |
+| `BLOCKHASH_SERVE_WINDOW`  | `256` |
+| `HISTORY_SERVE_WINDOW`    | `8192` |
```
marioevz, line 31:
```suggestion
| `HISTORY_SERVE_WINDOW`    | `8191` |
```
Update according to PR description, but I have not verified this value.

EIPS/eip-2935.md
```diff
@@ -61,8 +61,8 @@ The history contract has two operations: `get` and `set`. The `set` operation is
 It is used from the EVM for looking up block hashes.
 
 * Callers provide the block number they are querying in a big-endian encoding.
-* If calldata is bigger than 2^64-1, revert.
-* For any output outside the range of [block.number-`HISTORY_SERVE_WINDOW`, block.number-1] return 0.
+* If calldata is not 32 bytes, revert.
+* For any request outside the range of [block.number-`HISTORY_SERVE_WINDOW`, block.number-1], revert.
```
g11tech, line 65: this is not the current BLOCKHASH (with window 256) behavior so we are actually diverting from the behavior ?

EIPS/eip-2935.md
```diff
@@ -234,6 +202,12 @@ def process_block_hash_history(block: Block, state: State):
 
 The first option is recommended until the Verkle fork, to stay consistent with [EIP-4788](./eip-4788.md) and to issues for misconfigured networks where this EIP is activated but history contract hasn't been deployed. The recommendation may be reconsidered at the Verkle fork if filtering the system contract code chunks is deemed too complex.
 
+### Size of ring buffers
+
+The ring buffer data structure is sized to hold 8191 hashes. In other system contracts a prime ring buffer size is chosen in because using a prime as the modulus ensures that no value is overwritten until the entire ring buffer has been saturated and thereafter, each value will be updated once per iteration, regardless of if some slot are missing or the slot time changes. However, in this EIP the block number is the value in the modulo operation and it only ever increases by 1 each iteration. Which means we can be confident that the ring buffer will always remain saturated.
```
g11tech, line 207:
```suggestion
The ring buffer data structure is sized to hold 8191 hashes. In other system contracts a prime ring buffer size is chosen in because using a prime modulus ensures that the entire ring buffer gets saturated. However, in this EIP the block number only ever increases by 1 each iteration fully saturating the buffer before wrapping around.
```

EIPS/eip-2935.md
```diff
@@ -234,6 +202,12 @@ def process_block_hash_history(block: Block, state: State):
...
+
+For consistency with other system contracts, we have decided to retain the buffer size of 8191. Given the current mainnet values, 8191 roots provides about a day of coverage. This also gives users plenty of time to make a transaction with a verification against a specific hash and get the transaction included on-chain.
```
g11tech, line 209:
```suggestion
However for consistency with other system contracts, we have decided to retain the buffer size of 8191. Given the current mainnet values, 8191 roots provides about a day of coverage. This also gives users plenty of time to make a transaction with a verification against a specific hash and get the transaction included on-chain.
```

gballet (changes requested):
@lightclient could you please share a link to the audit? We're especially interested in figuring out the `REVERT` on out-of-bounds part.

The behavior of returning 0 was initially intended, as there was no ring buffer in initial versions of the spec, and any request for a block hash before the fork was to be 0. The size of the ring buffer might be extended in the future (especially if the block time changes). Now, this is not a big deal per se, as executing the system contract's bytecode becomes "legacy" with eip-7709. Still, it'd be nice to know why an out-of-bounds read returning 0 is a security problem in this particular case.

gballet (commented):
I had a round of discussions and while we consider it less than ideal to pick a buffer size for no direct reason than matching a design choice in _another_ contract, we won't block Prague for that, since it has limited impact provided the justification is written in the eip.

gballet:
But @lightclient please address the change requests before merging

EIPS/eip-2935.md
lightclient: mentioned this elsewhere, but I think this divergence is okay because i) it matches other system contracts and ii) it is pretty straightforward to create an adapter around this system contract to mimic the BLOCKHASH behavior when it is used to back the actual op.

EIPS/eip-2935.md
```diff
@@ -74,89 +74,57 @@ It is used from the EVM for looking up block hashes.
...
+calldatasize
+sub
+push1 0x42
 jumpi
-
-// check if input > 8 byte value and revert if this isn't the case
-// the check is performed by comparing the biggest 8 byte number with
-// the call data, which is a right-padded 32 byte number.
-push8 0xffffffffffffffff
 push0
```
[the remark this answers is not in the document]
lightclient: I removed all these comments because the assembled code comes from the command:

`disease --code 0x$(geas src/consolidations/main.eas) | sed -E '/^[[:space:]]*$/d; s/^[[:space:]]*[0-9a-fA-F]+:[[:space:]]+//; s/[[:space:]]*# selector\(.*\)$//'`

all the code info is in the sysasm: https://github.com/lightclient/sys-asm/blob/main/src/execution_hash/main.eas

I'd probably hold off on code comments in the eip until this is going to last call.

EIPS/eip-2935.md
```diff
@@ -74,89 +74,57 @@ It is used from the EVM for looking up block hashes.
...
 return
-
-// 0x53: revert
 jumpdest
 push0
 push0
 revert
-
-// 0x57: set op - sstore the input to number-1 mod 8192
```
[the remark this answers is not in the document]
lightclient: I removed all these comments because the assembled code comes from the command:

`disease --code 0x$(geas src/consolidations/main.eas) | sed -E '/^[[:space:]]*$/d; s/^[[:space:]]*[0-9a-fA-F]+:[[:space:]]+//; s/[[:space:]]*# selector\(.*\)$//'`

all the code info is in the sysasm: https://github.com/lightclient/sys-asm/blob/main/src/execution_hash/main.eas

I'd probably hold off on code comments in the eip until this is going to last call.

g11tech:
@lightclient there are some merge conflicts here i guess

EIPS/eip-2935.md
```diff
@@ -26,10 +26,10 @@ A side benefit of this approach could be that it allows building/validating proo
 
 | Parameter | Value |
 | - | - |
-| `BLOCKHASH_SERVE_WINDOW` | `256` |
-| `HISTORY_SERVE_WINDOW` | `8192` |
-| `SYSTEM_ADDRESS` | `0xfffffffffffffffffffffffffffffffffffffffe` |
-| `HISTORY_STORAGE_ADDRESS` | `0x0aae40965e6800cd9b1f4b05ff21581047e3f91e`|
+| `BLOCKHASH_SERVE_WINDOW`  | `256` |
+| `HISTORY_SERVE_WINDOW`    | `8192` |
```
g11tech, line 30:
```suggestion
| `HISTORY_SERVE_WINDOW`    | `8191` |
```

g11tech (approved):

PR #9152, opened 2024-12-18
Update EIP-7702: remove chain_id in authorization tuple

fjl:
This is a counterproposal to #9143. Since the chain ID is already available in the context of the transaction, there is no need to specify it again in each authorization tuple. To preserve the special case of an authorization which is valid on all chains (`chain_id == 0`), a new boolean field `unchained` is added to the tuple in place of `chain_id`.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -70,12 +70,12 @@ At the start of executing the transaction, after incrementing the sender's nonce
 3. `authority = ecrecover(keccak(MAGIC || rlp([auth_chain_id, address, nonce])), y_parity, r, s]`
     * 3a. The `s` value must be less than or equal to `secp256k1n/2`, as specified in [EIP-2](./eip-2.md).
 4. Add `authority` to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
-5. Verify the code of `authority` is either empty or already delegated.
+5. Verify the code of the `authority` account is either empty or already delegated.
 6. Verify the nonce of `authority` is equal to `nonce`. In case `authority` does not exist in the trie, verify that `nonce` is equal to `0`.
 7. Add `PER_EMPTY_ACCOUNT_COST - PER_AUTH_BASE_COST` gas to the global refund counter if `authority` exists in the trie.
 8. Set the code of `authority` to be `0xef0100 || address`. This is a delegation designation. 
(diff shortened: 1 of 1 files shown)


smartprogrammer93 (approved):
Makes sense

EIPS/eip-7702.md
```diff
@@ -41,18 +41,15 @@ We introduce a new [EIP-2718](./eip-2718.md) transaction, "set code transaction"
...
 
-The `authorization_list` is a list of tuples that store the address to code which the signer desires to execute in the context of their EOA. The transaction is considered invalid if the length of `authorization_list` is zero.
-
-The transaction is also considered invalid when any field in an authorization
-tuple cannot fit within the following bounds:
+`authorization_list` is a list of tuples that store the address to code which the signer desires to execute in the context of their EOA. The transaction is considered invalid if the length of `authorization_list` is zero. It is also considered invalid when any field in an authorization tuple exceeds the following bounds:
 
 ```python
-assert auth.chain_id < 2**64
+assert auth.unchained == 0 || auth.unchained == 1
```
jochem-brouwer, line 52: This specific check looks like the `yParity` check (where `yParity` is also either 0 or 1). 
```suggestion
assert auth.unchained < 2**8
```

There should be an added check for each tuple to validate if the `auth.unchained` is 0 or 1 (if not, skip the tuple).

gakonst (commented):
From Reth side we don't think this makes full sense / is an improvement.

With this idea:
* When I'm signing, I need to sign `auth_chain_id` which I need to pull from the transaction context, depending  on whether the `unchained` flag is set or not.
* This seems more complex for not changing any functionality, or improving security.

I understand that this is a counterproposal to U256 chainId because plausibly U256 as as an auth parameter makes the auth list big in terms of data?

fjl:
I'm proposing this just because it's weird to duplicate a potentially large chainID in every authorization tuple.
The chainID value will always be zero or equal to the tx chainID.

From an implementation point of view, the chainID is already an input to the signing function. This does not change with the PR, you still have to supply the chainID when signing an authorization. We just leave out the chainID in the encoded authorization object that goes into the transaction, and replace it with a single bit instead, which says if chainID was zero.

gurukamath:
Would this leave any room for replaying a chain specific authorisation tuple on a different chain?

fjl:
No, because the signed-over data still contains the chainID, and it will be verified with the chainID of the executing chain.

EIPS/eip-7702.md
lightclient: Yes, I think this line should be reverted. This section has to do with decoding txs and I think 1 byte is the highest resolution we want here. I think many languages represent boolean as 1 byte value despite it only needing 1 bit of information.

We would verify `unchained` is 0 or 1 in the tuple validation.

mScott224 (approved):

PR #9163, opened 2024-12-19
Update EIP-7607: Add EIP-7666 (EVM-ify the identity precompile) to proposed-for-inclusion in Fusaka

vbuterin:
We need to start having a pipeline to retire precompiles that are  outdated and underused, to reduce the protocol complexity and consensus risk over time.

The easiest precompile to start removing is the identity precompile. EIP-7666 removes the precompile, and replaces it with a piece of EVM code that has equivalent functionality (except naturally it costs more gas). If this approach works for this case, we can later start also applying it to other precompiles (eg. Blake, RIPEMD-160) that see very little usage.

### DIFF
--- a/EIPS/eip-7607.md
+++ b/EIPS/eip-7607.md
@@ -41,6 +41,8 @@ Definitions for `Scheduled for Inclusion`, `Considered for Inclusion` and `Propo
 
 ### Proposed for Inclusion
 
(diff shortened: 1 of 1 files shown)


timbeiko (approved):
First formal `PFI` PR 🎉 !

sambacha:
@pcaversaccio your prayers have been answered

PR #9186, opened 2024-12-24
Update EIP-1485: Fix typographical errors in Documentation

Pricstas:
This pull request resolves typographical errors across multiple EIP documents to improve readability and grammatical accuracy. The following changes have been made:

1. **`eip-1485.md`**:
   - Corrected "a implementation" to "an implementation".

2. **`eip-158.md`**:
   - Fixed "a address" to "an address".

3. **`eip-1884.md`**:
   - Updated "a inherently" to "an inherently".

4. **`eip-1959.md`**:
   - Corrected "an hardfork" to "a hardfork".

5. **`eip-2294.md`**:
   - Changed "a incompatible" to "an incompatible".

6. **`eip-234.md`**:
   - Fixed "a refetch" to "an refetch".

7. **`eip-2474.md`**:
   - Corrected "an change" to "a change".

8. **`eip-747.md`**:
   - Updated "a asset" to "an asset".

9. **`eip-7808.md`**:
   - Fixed "an transaction-type" to "a transaction-type".

These edits affect the following files:
- `eip-1485.md`
- `eip-158.md`
- `eip-1884.md`
- `eip-1959.md`
- `eip-2294.md`
- `eip-234.md`
- `eip-2474.md`
- `eip-747.md`
- `eip-7808.md`

### Notes:
- These changes only affect comments and documentation within the EIPs.
- No functional changes are introduced.

---

### Checklist:
- [x] Verified grammar corrections for accuracy.
- [x] Targeted PR against the correct branch.
- [x] Followed repository contribution guidelines.


wighawag (approved):

Pandapip1 (approved):

holiman (approved):

Pandapip1:
I can't do anything about this PR. I'm not an editor anymore.

EIPS/eip-234.md
```diff
@@ -28,7 +28,7 @@ A client (dApp) who needs reliable notification of both log additions (on new bl
...
...'t* the block they want the logs for and is instead for a block that was re-orged in (and may not be fully reconciled with the internal client state).  This can be partially worked around by looking at the resulting logs themselves and identifying whether or not they are for the block hash requested.  However, if the result set is an empty array (no logs fetched) then the client is in a situation where they don't know what block the results are for.  The results could have been legitimately empty (bloom filter can yield false positives) for the block in question, or they could be receiving empty logs for a block that they don't know about.  At this point, there is no decision the client can make that allows them a guarantee of recovery.  They can assume the empty logs were for the correct block, but if they weren't then they will never try to fetch again.  This creates a problem if the block was only transiently re-orged out because it may come back before the next block poll so the client will never witness the reorg.  They can assume the empty logs were for the wrong block, a refetch them, but they may continue to get empty results putting them right back into the same situation.
```
g11tech, line 31:
```suggestion
Once the user has a reliable stream of blocks, they can then look at the bloom filter for the new block and if the block *may* have logs of interest they can fetch the filtered logs for that block from the node.  The problem that arises is that a re-org may occur between when the client receives the block and when the client fetches the logs for that block.  Given the current set of filter options, the client can only ask for logs by block number.  In this scenario, the logs they get back will be for a block that *isn't* the block they want the logs for and is instead for a block that was re-orged in (and may not be fully reconciled with the internal client state).  This can be partially worked around by looking at the resulting logs themselves and identifying whether or not they are for the block hash requested.  However, if the result set is an empty array (no logs fetched) then the client is in a situation where they don't know what block the results are for.  The results could have been legitimately empty (bloom filter can yield false positives) for the block in question, or they could be receiving empty logs for a block that they don't know about.  At this point, there is no decision the client can make that allows them a guarantee of recovery.  They can assume the empty logs were for the correct block, but if they weren't then they will never try to fetch again.  This creates a problem if the block was only transiently re-orged out because it may come back before the next block poll so the client will never witness the reorg.  They can assume the empty logs were for the wrong block, and refetch them, but they may continue to get empty results putting them right back into the same situation.
```

g11tech: the suggested change is:
```
They can assume the empty logs were for the wrong block, and refetch them
```

instead of `a refetch them`

SamWilsn (approved):

PR #9193, opened 2024-12-26
Add EIP: Private key deactivation and reactivation

colinlyguo:
Introduce a new precompiled contract to enable Externally Owned Accounts (EOAs) to deactivate and reactivate their private keys.

### DIFF
--- a/EIPS/eip-xxxx.md
+++ b/EIPS/eip-xxxx.md
@@ -1,0 +1,201 @@
+---
+title: Private key deactivation and reactivation
+description: Introduce a new precompiled contract to enable Externally Owned Accounts (EOAs) to deactivate and reactivate their private keys.
+author: Liyi Guo (@colinlyguo)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-12-27
+requires: 20, 2612, 7701, 7702
+---
+
+## Abstract
+
+This EIP introduces a precompiled contract that enables EOAs with delegated control to smart contracts via [EIP-7702](./eip-7702) to deactivate/reactivate their private keys. This design does not require additional storage fields or account state changes. By leveraging delegated code, reactivation can be performed securely through mechanisms such as social recovery.
+
+## Motivation
+
+[EIP-7702](./eip-7702) enables EOAs to gain smart contract capabilities, but the private key of the EOA still retains full control over the account.
+
+With this EIP, EOAs can fully migrate to smart contract wallets, while retaining recovery options with reactivation. The flexible deactivate/reactivate design also paves the way for native account abstraction. e.g., [EIP-7701](./eip-7701).
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Parameters
+
+| Constant                          | Value                |
+|-----------------------------------|----------------------|
+| `PRECOMPILE_ADDRESS`              | `0xTBD`              |
+| `PRECOMPILE_GAS_COST`             | `5000` (tentative)   |
+
+### Delegated code encoding
+
+The deactivation status is encoded by appending or removing the `0x00` byte at the end of the delegated code. The transitions between two states are as follows:
+
+- Active state: `0xef0100 || address`, the private key is active and can sign transactions.
+- Deactivated state: `0xef0100 || address || 0x00`, the private key is deactivated and cannot sign transactions.
+
+### Precompiled contract
+
+A new precompiled contract is introduced at address `PRECOMPILE_ADDRESS`. For each call, it consumes `PRECOMPILE_GAS_COST` gas, and the precompiled contract executes the following steps:
+- The precompiled contract checks that the caller is an EOA with delegated code (i.e., its account code begins with the prefix `0xef0100`, as defined in [EIP-7702](./eip-7702)). If the code does not conform to the required prefix, the contract MUST terminate without making any state changes.
+- The precompile determines the current state of the delegated code based on its byte length:
+    - If the delegated code is 24 bytes (`0xef0100 || address || 0x00`), it removes the last byte (`0x00`), transitioning to the active state (`0xef0100 || address`).
+    - If the delegated code is 23 bytes (`0xef0100 || address`), it appends `0x00`, transitioning to the deactivated state (`0xef0100 || address || 0x00`).
+- The updated delegated code is saved as the new account code for the EOA.
+
+
+### Transaction validation
+If the account is verified as an EOA with a delegated code (begins with the prefix `0xef0100`), transactions signed by the private key MUST be rejected if the delegated code is in the deactivated state (i.e., `24` bytes long).
+
+### Gas cost
+No changes to the base transaction gas cost (`21000`) are required, as the additional valid check for the deactivation status is minimal compared with [EIP-7702](./eip-7702) transactions. Thus it is reasonable to include them in the base gas cost.
+
+## Rationale
+
+### Using a precompiled contract
+Alternative methods for implementing this feature include:
+- Adding a new transaction type: A new transaction type could deactivate/reactive EOA private keys. This would complicate reactivation, as the contract would need to serve as the authorizer for reactivation, increasing protocol complexity.
+- Deploying a regular smart contract: A regular deployed contract could track the `deactivated` status of each `address`. This approach would break the base transaction gas cost of `21000` in transaction validation, as accessing the `deactivated` status would require additional address and storage lookups, increasing gas usage.
+
+### In-protocol reactivation
+This approach ensures maximum compatibility with future migrations. EOAs can reactivate their private keys, delegate their accounts to an [EIP-7701](./eip7701) contract, and then deactivate their private keys again. This avoids the limitations of upgradable contracts. e.g., to remove legacy proxy contracts when EOF contracts become available, thereby reducing gas overhead, one can reactivate the EOA and delegate to an EOF proxy contract.
+
+### `5000` Gas `PRECOMPILE_GAS_COST`
+The `5000` gas cost is sufficient to cover validation, computation, and storage updates for the delegated code.
+
+### Alternative EOA migration approach
+One alternative migration approach involves using a hard fork to edit all existing and new EOAs to upgradable smart contracts using EOA's ECDSA signatures. Users can then upgrade these smart contracts to achieve more granular permission control. However, this approach is incompatible with EOAs that have already delegated to smart contracts, as it overwrites the existing smart contract implementations. The EIP aims to fill this migration gap.
+
+### Avoiding delegated code prefix modification
+This EIP appends a byte (`0x00`) to the delegated code instead of modifying the prefix (`0xef0100`) of [EIP-7702](./eip-7702) to ensure forward compatibility. If future prefixes such as `0xef0101` are introduced, changing the prefix (e.g., to `0xef01ff`) makes it unclear which prefix to restore upon reactivation.
+
+### Avoiding account state changes
+Another alternative is to add a new field to the account state to store the `deactivated` status. However, this approach complicates the account state. It also brings changes in the account trie structure and RLP encoding used in networking, which complicates the implementation.
+
+### Forwards compatibility for removing EOAs
+After all existing and future EOAs have been migrated to smart contracts. It's natural and easy to deprecate this EIP:
+- Removing the precompiled contract.
+- Removing validation of the deactivation status since all EOAs are smart contracts.
+- The appended `0x00` byte can be optionally removed from the delegated code.
+
+## Backwards Compatibility
+
+This EIP maintains backwards compatibility with existing EOAs and contracts.
+
+## Test Cases
+
+```python
+# Initialize the state database and precompiled contract
+state_db = StateDB()
+precompiled_contract = PrecompiledContract()
+
+# Test 1: Valid caller with active state
+valid_caller = "0x1234"
+delegated_address = bytes.fromhex("112233445566778899aabbccddeeff0011223344")
+delegated_code = PrecompiledContract.DELEGATED_CODE_PREFIX + delegated_address
+state_db.set_code(valid_caller, delegated_code)  # Active state
+assert state_db.get_code(valid_caller) == delegated_code  # Verify initial state
+
+# Toggle to deactivated state
+precompiled_contract.run(valid_caller, state_db)
+assert state_db.get_code(valid_caller) == delegated_code + b"\x00"  # Verify deactivated
+
+# Toggle back to active state
+precompiled_contract.run(valid_caller, state_db)
+assert state_db.get_code(valid_caller) == delegated_code  # Verify reactivated
+
+# Test 2: Invalid caller without delegated code
+invalid_caller = "0x5678"  # No delegated code
+assert state_db.get_code(invalid_caller) == b""  # Verify initial state is empty
+
+# Run contract on invalid caller (should do nothing)
+precompiled_contract.run(invalid_caller, state_db)
+assert state_db.get_code(invalid_caller) == b""  # State remains unchanged
+
+# Test 3: Caller with invalid prefix
+invalid_prefix_caller = "0x9999"
+invalid_code = bytes.fromhex("ab01") + delegated_address  # Invalid prefix
+state_db.set_code(invalid_prefix_caller, invalid_code)
+precompiled_contract.run(invalid_prefix_caller, state_db)
+assert state_db.get_code(invalid_prefix_caller) == invalid_code  # State remains unchanged
+```
+
+## Reference Implementation
+
+```python
+class PrecompiledContract:
+    """
+    Precompiled contract for toggling the activation status of EOAs
+    based on their delegated code.
+    """
+    DELEGATED_CODE_PREFIX = bytes.fromhex("ef0100")  # Prefix for delegated code
+    ACTIVE_CODE_LENGTH = 23  # Length of code in active state
+    DEACTIVATED_CODE_LENGTH = 24  # Length of code in deactivated state
+
+    def run(self, caller, state_db):
+        """
+        Toggles the delegated code of the caller between active and deactivated states.
+
+        Parameters:
+        - caller: The address calling the contract.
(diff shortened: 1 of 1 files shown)


EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,201 @@
+---
```
abcoathup, line 1:
```suggestion
---
eip: 7851
```

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,201 @@
...
+title: Private key deactivation and reactivation
+description: Introduce a new precompiled contract to enable Externally Owned Accounts (EOAs) to deactivate and reactivate their private keys.
+author: Liyi Guo (@colinlyguo)
+discussions-to: <URL>
```
abcoathup, line 5: Please create a discussions topic in Eth Magicians using the template: https://ethereum-magicians.org/c/eips/5

colinlyguo: thanks. updated in [314ed2b](https://github.com/ethereum/EIPs/pull/9193/commits/314ed2b98598f5ad51c5c1a4277686ca232c1df9).

EIPS/eip-xxxx.md
colinlyguo: updated in the same commit as https://github.com/ethereum/EIPs/pull/9193#discussion_r1898155717.

colinlyguo: got it. updated in commit [314ed2b](https://github.com/ethereum/EIPs/pull/9193/commits/314ed2b98598f5ad51c5c1a4277686ca232c1df9).

EnyiflexWire (approved):

jochem-brouwer:
I don't see how this works with 7702 though. If I have the private key of account A, then if I "deactivate" sending txs from A then I can still auth-sign from address B to A, thus removing the "deactivation" from A?

Should this add an extra check to 7702 when checking the auth-tuples? If the `authority` has the `00` deactivation byte then skip this authority?

EIPS/eip-7851.md
```diff
@@ -0,0 +1,223 @@
...
+The deactivation status is encoded by appending or removing the `0x00` byte at the end of the delegated code. The transitions between two states are as follows:
+
+- Active state: `0xef0100 || address`, the private key is active and can sign transactions.
+- Deactivated state: `0xef0100 || address || 0x00`, the private key is deactivated and cannot sign transactions.
+
+### Precompiled contract
+
+A new precompiled contract is introduced at address `PRECOMPILE_ADDRESS`. For each call, it consumes `PRECOMPILE_GAS_COST` gas, and the precompiled contract executes the following steps:
+
+- The precompiled contract checks that the caller is an EOA with delegated code (i.e., its account code begins with the prefix `0xef0100`, as defined in [EIP-7702](./eip-7702)). If the code does not conform to the required prefix, the contract MUST terminate without making any state changes.
```
jochem-brouwer, line 46: "Terminate" here means just "stop" the call to this precompile without making any state changes? (One could also "revert", thus the item put on stack when using the `CALL*` opcodes to this precompile would put 0 on stack if these conditions are not met (so no state changes are done?))

EIPS/eip-7851.md
```diff
@@ -0,0 +1,223 @@
...
+- The precompile determines the current state of the delegated code based on its byte length:
+  - If the delegated code is 24 bytes (`0xef0100 || address || 0x00`), it removes the last byte (`0x00`), transitioning to the active state (`0xef0100 || address`).
+  - If the delegated code is 23 bytes (`0xef0100 || address`), it appends `0x00`, transitioning to the deactivated state (`0xef0100 || address || 0x00`).
+- The updated delegated code is saved as the new account code for the EOA.
```
jochem-brouwer, line 50: I think a rule should be added here. In case that we are updating the delegated code, but we are running in "static" mode (so a call chain after using `STATICCALL`) this should abort (revert, no state-changes).

colinlyguo:
> I don't see how this works with 7702 though. If I have the private key of account A, then if I "deactivate" sending txs from A then I can still auth-sign from address B to A, thus removing the "deactivation" from A?
> 
> Should this add an extra check to 7702 when checking the auth-tuples? If the `authority` has the `00` deactivation byte then skip this authority?

nice catch. added the extra check you mentioned in the draft. pls refer to this commit: [0f2aaba](https://github.com/ethereum/EIPs/pull/9193/commits/0f2aabaad752324ad8bab51a123213ca290a1c10)

EIPS/eip-7851.md
colinlyguo: yes. you're right! I added the `STATICCALL` error case in this commit: [0f2aaba](https://github.com/ethereum/EIPs/pull/9193/commits/0f2aabaad752324ad8bab51a123213ca290a1c10)

EIPS/eip-7851.md
colinlyguo: Thanks for the clarification, the error case of precompile is updated as: "Returns a precompile contract error and consumes all provided gas" (refer to [0f2aaba](https://github.com/ethereum/EIPs/pull/9193/commits/0f2aabaad752324ad8bab51a123213ca290a1c10)).

The handling logic is consistent with other precompiles mentioned in https://www.evm.codes/precompiled. Also, consuming all gas for invalid calls (reducing DoS attack vectors) looks reasonable.

For the error return, I searched client implementation in [geth](https://github.com/ethereum/go-ethereum/blob/v1.14.12/core/vm/instructions.go#L756-L762) and found it handled outside by `CALL`, `STATICCALL`, `CALLCODE`, `DELEGATECALL`. When [precompiles return an error](https://github.com/ethereum/go-ethereum/blob/v1.14.12/core/vm/evm.go#L210-L244), it will be returned to `CALL`, and `CALL` will set return value as `0`. But formal documents would help clarify this for sure.

colinlyguo:
> ### File `EIPS/eip-7851.md`
> Requires 1 more reviewers from @g11tech, @lightclient, @SamWilsn

Any review from the review list to check if the pr can be considered as a draft? so that it can be further discussed.

EIPS/eip-7851.md
```diff
@@ -0,0 +1,245 @@
...
+eip: 7851
+title: EOA private key deactivation/reactivation
+description: Introduce a new precompiled contract for EOAs with delegated code to deactivate or reactivate private keys.
+author: Liyi Guo (@colinlyguo)
+discussions-to: https://ethereum-magicians.org/t/eip-7851-eoa-private-key-deactivation-reactivation/22344
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-12-27
+requires: 20, 2612, 7701, 7702
```
jochem-brouwer, line 11:
```suggestion
requires: 7702
```

I don't see how ERC-20, EIP 2612, and 7701 are required here. If you would require 7701 this means this EIP can only be active once EOF is also active.

EIPS/eip-7851.md
```diff
@@ -0,0 +1,245 @@
...
+- Updates caller's delegated code based on length:
+  - 23 bytes (`0xef0100 || address`): Appends `0x00` to deactivate private key authorization.
+  - 24 bytes (`0xef0100 || address || 0x00`): Removes last byte `0x00` to activate private key authorization.
+- Saves updated code as caller's new account code.
+
+### Transaction validation
+
+If the account is verified as an EOA with a delegated code (begins with the prefix `0xef0100`), the following validations MUST be performed:
+
+- Transactions signed by the private key MUST be rejected if the delegated code is in the deactivated state (i.e., `24` bytes long).
```
jochem-brouwer, line 59: I think extra clarification here is necessary. What does "rejected" mean? I am assuming that "rejected" here means that the transactions which do not meet this rule can be included in the block, however any state changes like EVM execution, publishing blobs, or authorizing (EIP-7702) are not done and instead the transaction immediately exits. However, this does mean that these state changes are still done:

- Nonce update
- Paying for gas (so balance update)

Also, for this rejection: is the entire gas limit consumed or only the intrinsic gas (which includes for instance the calldata fee)

EIPS/eip-7851.md
colinlyguo: oh I see. thanks for the clarification here.

EIPS/eip-7851.md
```diff
@@ -0,0 +1,245 @@
...
+- Any [EIP-7702](./eip-7702) authorization from an authority with deactivated delegated code MUST be considered invalid and skipped.
+
+### Gas Cost
+
+No changes to the base transaction gas cost (`21000`) are required, as the additional valid check for the deactivation status is minimal. Thus it is reasonable to consider the overhead covered by the base gas cost.
```
jochem-brouwer, line 64: The check if the current account is deactivated actually is more expensive than a base transaction due to the nature of the state trie. To read an account from the trie, one gets these items:

- Nonce
- Balance
- `codeHash`
- `storageHash`

In order to read the actual code (to figure out if the account is delegated and thus to also perform the private key deactivation check) we have to actual read from the database again (to find what code `codeHash` points to). Since this is a disk read this is expensive and since this is now necessary for all transactions this might be a reason to increase the base transaction cost.

colinlyguo: it makes sense. I missed this (`code` read cost).

In the first draft before opening this pr, the proposal plans to add a new bool (1 byte) field in the account state, so that it can be read together with `nonce`, `balance`, `codeHash`, and `storageHash`.

this may add some complexities in backward compatibility discussions while keeping the basic cost of transactions cheaper. do you think it makes sense to change the implementation of the draft to this method?

EIPS/eip-7851.md
```diff
@@ -0,0 +1,245 @@
...
+- Returns a precompile contract error, consumes all gas, and no state changes are made if:
+  - Gas is insufficient.
+  - Called via `STATICCALL` (i.e. in a read-only context).
+  - Caller is not an EOA with delegated code (prefix `0xef0100` as per [EIP-7702](./eip-7702)).
+- Updates caller's delegated code based on length:
+  - 23 bytes (`0xef0100 || address`): Appends `0x00` to deactivate private key authorization.
+  - 24 bytes (`0xef0100 || address || 0x00`): Removes last byte `0x00` to activate private key authorization.
+- Saves updated code as caller's new account code.
+
+### Transaction validation
```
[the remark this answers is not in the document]
colinlyguo: reported.

EIPS/eip-7851.md
jochem-brouwer: @colinlyguo This extra field in state will make this rather complex, because you are adding an extra field to the account state. This also means there will be logic introduced when enabling this EIP: each time if an account is hit, there will be logic to see if this account is "upgraded" to have this boolean in state. If this is not the case, then it could either add this flag to the state, or it could also do it only when necessary (so, if a transaction is sent, there will first be a check if the private key is deactivated. To check, first see if there is a boolean field at all: if this is not the case, then we can know for sure that this is an "old" account state and the private key is thus active. We can then add this field to the state. But this mechanism of checking and adding these fields will make the test vectors for this very big and it will also add a lot of somewhat boilerplate "upgrade" code to ensure the old accounts will update to the new ones (with this private key enabled flag). I would not recommend it.

I do recall in the past there have been discussions of using the `nonce` field. One could, for instance, use the highest bit of the nonce for this flag. One could use this flag to see if the accounts private key is disabled. What is handy here is that the nonce data is already in the account, so no extra disk lookups are necessary (when compared to putting it in the code).

However some nonce-limiting EIPs are around: https://eips.ethereum.org/EIPS/eip-3338, https://eips.ethereum.org/EIPS/eip-4803, putting a flag in nonce would likely throw away the benefits/rationales of these EIPs (4803 ensures that it can be decoded as int).

colinlyguo: agree. introduced a new field in the account state will add a lot of unnecessary "upgrade" code, which should be avoided.

adding a flag in `nonce` looks like a good idea.

for [EIP-3338](https://eips.ethereum.org/EIPS/eip-3338) which limits the nonce <= 2^52, using the highest bit of 2^52 seems not to make a large difference: (i) the number is still large, 2^51 = 2.25 * 10^15. (ii) compared with 2^64-1, it's also quite small as 2^52.

To limit the range of future nonce, can move the "deactivated bit" based on the length of the nonce after an upgrade.

while this will also introduce a problem (maybe?) that historical nonce in some reverted transactions (those included but reverted since the EOA is deactivated) will exceed future limited value ("languages that only have floating point number support" may not be happy to see that).

For this, I have come up with 2 solutions:
(i) reject transactions in the "signature validation phase" if the highest bit of its nonce is 1.
(ii) encode the "deactivated bit" in the lowest bit of the "nonce" field.

EIPS/eip-7851.md
colinlyguo: added clarifications in this commit: [4ff1e76](https://github.com/ethereum/EIPs/pull/9193/commits/4ff1e76241ae136f4f2c79ed6c0a03d1e28bee46).

also removing the basic 21000 gas discussion, since it seems to be unrelated now.

EIPS/eip-7851.md
ADNANALISALHADABY: Ok

ADNANALISALHADABY (commented):
EIPS/eip-xxxx.md

EIPS/eip-7851.md
```diff
@@ -0,0 +1,247 @@
...
+- Deploying a regular smart contract: A regular deployed contract could track the deactivated status of each address. However, invoking this contract, which executes bytecode and accesses storage, during deactivation and reactivation, would be more expensive than using a precompiled contract. Additionally, this approach "leaks" the design of a (widely used) programming language into the Ethereum core protocol. While it poses no obvious security risks, it is not ideal from a design perspective.
+
+### In-protocol reactivation
+
+This approach ensures maximum compatibility with future migrations. EOAs can reactivate their private keys, delegate their accounts to an [EIP-7701](./eip-7701) contract, and then deactivate their private keys again. This avoids the limitations of contract upgrades. e.g. to remove legacy proxy contracts (reducing gas overhead) when upgrading to EOF contracts, one can reactivate the EOA and redelegate it to an EOF proxy contract.
```
Thegaram, line 78: I see no better alternative, but this definitely sounds risky. Once a privkey has been deactivated most users would not take care to keep it secure. If it's leaked an attacker could use it after reactivation.

A powerful solution would be to enable delegation through a precompile. Then the account code could implement the required checks before "upgrading" its own code.

EIPS/eip-7851.md
```diff
@@ -0,0 +1,247 @@
...
+        if addr not in self.accounts:
+            self.accounts[addr] = {}
+        self.accounts[addr]["code"] = value
+```
+
+## Security Considerations
+
+### Additional checks during transaction validation
+
+The deactivation status is determined by checking the length of the delegated code. This check introduces an additional storage read and check comparable to account state read and checks, such as nonce validation. It is integrated into the transaction validity check of the transaction pool and consensus rules, ensuring that invalid transactions are filtered out before propagation and execution.
```
Thegaram, line 223: > This check introduces an additional storage read

Since we read the account anyway for the nonce, I don't think this requires an additional read, so there's no additional workload for the txpool.

EIPS/eip-7851.md
Thegaram, line 78: > EOAs can reactivate their private keys

The code cannot know if its privkey is currently deactivated or not, right? (Since `extcodehash` only returns the delegation prefix.) Would this be an issue?

EIPS/eip-7851.md
jochem-brouwer: There is an extra storage read, because the code itself is not directly stored in the account: only the code hash is stored in the account. So an extra storage read is necessary to get the actual code of the account (and not the hash).

EIPS/eip-7851.md
colinlyguo: I think it's OK not to know the specific status, each time the contract invokes the precompile, it just toggles the status, i.e. from `deactivated` to `activated`, or from `activated` to `deactivated`.

EIPS/eip-7851.md
```diff
@@ -0,0 +1,247 @@
+---
+eip: 7851
+title: EOA private key deactivation/reactivation
+description: Introduce a new precompiled contract for EOAs with delegated code to deactivate or reactivate private keys.
```
g11tech, line 4:
```suggestion
description: Introduce a new precompiled contract for EOAs with delegated code to deactivate or reactivate delegation.
```

EIPS/eip-7851.md
```diff
@@ -0,0 +1,247 @@
+---
+eip: 7851
+title: EOA private key deactivation/reactivation
```
g11tech, line 3:
```suggestion
title: EOA delegation deactivation/reactivation
```

EIPS/eip-7851.md
```diff
@@ -0,0 +1,247 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-12-27
+requires: 7702
+---
+
+## Abstract
+
+This EIP introduces a precompiled contract that enables Externally Owned Accounts (EOAs) with delegated control to smart contracts via [EIP-7702](./eip-7702) to deactivate or reactivate their private keys. This design does not require additional storage fields or account state changes. By leveraging delegated code, reactivation can be performed securely through mechanisms such as social recovery.
```
g11tech, line 16:
```suggestion
This EIP introduces a precompiled contract that enables Externally Owned Accounts (EOAs) with delegated control to smart contracts via [EIP-7702](./eip-7702) to deactivate or reactivate their delegation. This design does not require additional storage fields or account state changes. By leveraging delegated code, reactivation can be performed securely through mechanisms such as social recovery.
```

EIPS/eip-7851.md
g11tech, line 3:
```suggestion
title: Delegated EOA's private key deactivation/reactivation
```
may be doesn't fit the title length constraint...

colinlyguo: yes. While the added `delegated` seems necessary, what about `deactivate/reactivate of delegated EOA's key`?

g11tech: yes sounds good

g11tech (approved):
lgtm for draft

EIPS/eip-7851.md
colinlyguo: for complementary, I found that nodes already load code during transaction validation due to EIP-3607 and EIP-7702, so adding a length check is negligible.

code loading in geth implementation: https://github.com/ethereum/go-ethereum/blob/0e1a19da76d1761fdb3db78112a9d0e3571d43a4/core/state_transition.go#L323

and transaction validation in `txpool` still needs getting `code`, which is comparable to `nonce` and `balance` checks.

PR #9217, opened 2025-01-06
EIP-2537: Update test files to latest gas schedule

jochem-brouwer:
This PR updates the test files linked in EIP-2537 to reflect:

- The G1MUL/G2MUL precompiles are removed
- The gas schedule of (nearly) all precompiles is updated

These files pass the tests on EthereumJS https://github.com/ethereumjs/ethereumjs-monorepo/pull/3807 and the BLS precompiles also pass the latest Pectra tests: https://github.com/ethereum/execution-spec-tests/releases/tag/pectra-devnet-5%40v1.1.0

If another client could double-check that the modified files are correct, that would be great :smile: :+1:


g11tech:
@jochem-brouwer could you have @marioevz give it a review/approve before I merge it?

marioevz:
I checked the differences and found a couple of things:
- The gas in our JSON files was outdated, but this was a no-op because EEST has the formulas built-in and disregarded the value in the JSON.
- The *mul_* files were not deleted from the EEST repository, rather they were redirected to be executed by the MSM precompile.

I updated the JSON files in EEST with the ones in this PR and the generated files were identical.

Nevertheless I will open a PR in EEST to update them just for them to be up to date, although we will keep the *mul_* JSON files.

jochem-brouwer:
It is actually a good point regarding the `mul` files to redirect those to MSM. I should do so as well, but for clarity I think it is better (in here) to put those inputs/outputs into the MSM file to make it clear it should be ran by that precompile? (Since the `mul` precompile does not exist anymore)

jochem-brouwer:
I have updated this PR, @marioevz could you please re-check? I have now kept in the `mul` files (both failing and succeeding ones) and left a comment in the test file readme to ensure people know these should now be executed by the MSM precompiles.

jochem-brouwer:
I wanted to keep the original formatting in the json files, but my VSCode keeps messing up the indentation which is why the diff here looks super huge (only the `Gas` lines should be changed here!).

marioevz (approved):
All looks good:

- I've updated the vectors included in [EEST](https://github.com/ethereum/execution-spec-tests/pull/1063) to match the ones in https://github.com/ethereum/EIPs/pull/9217/commits/3e23068fb94cfccf3d9638943eb63a31c259e80b
- I added a new cross-check to verify that the gas values in the JSON match the values calculated by EEST
- All tests and cross-checks passed

jochem-brouwer:
@g11tech this is merge ready :smile: :+1:

g11tech (approved):
lgtm

PR #9228, opened 2025-01-09
Update EIP-7702: Update description

Thegaram:
The word *permanently* was added in 52c5677 but this seems to be outdated. In the latest version of EIP-7702, `SET_CODE_TX_TYPE` transactions can set, update, or clear an account's delegation designation.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -1,7 +1,7 @@
 ---
 eip: 7702
 title: Set EOA account code
-description: Add a new tx type that permanently sets the code for an EOA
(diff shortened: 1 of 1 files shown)


lightclient:
Thanks for this. I added this specifically because people felt it was less clear that 7702 added the delegation permanently. When it was originally proposed it was a temporary set code operation.

Thegaram:
> people felt it was less clear that 7702 added the delegation permanently

Makes sense. But the word _permanently_ suggests that the delegation cannot be changed/removed, at least that would be my interpretation. Maybe _persistently_ (i.e. stored in state) would be more accurate.

PR #9237, opened 2025-01-13
Update EIP-7642: Move to Draft

fjl:
Here I am updating the eth/69 proposal with two changes:

- Simplifying the receipts encoding. The original EIP proposed a new format for receipts with the bloom filter removed. It did that by mirroring the complex encoding for typed receipts which is also used by Ethereum consensus. Since the network protocol now introduces a deviation from the encoding used by consensus, there is no need to make it so complex. 

  Ultimately all receipt types contain the same four fields. So with this update, I am changing the network encoding to be the same for all receipt types.

- I am adding a new field `earliestBlock` in the `Status` message. This is somewhat similar to EIP-7542, which proposed adding a `blockRange`.

### DIFF
--- a/EIPS/eip-7642.md
+++ b/EIPS/eip-7642.md
@@ -91,12 +91,11 @@ CPU intensive. The bandwidth gains amount to roughly 530GiB per syncing node or 
 95GiB snappy compressed.
 
 In Ethereum consensus, the encoding of receipts differs between legacy transactions and
-typed transactions. Typed transaction receipts are technically 'opaque' and have the data
-wrapped in a byte array. However, all receipt types ultimately contain the same four
-fields. With the removal of the bloom filter, the networking protocol now deviates from
-the encoding used by consensus, and there is no need to replicate the weird and expensive
-encoding used there. The proposed receipt encoding is just a flat list of the required
-data fields.
+typed transactions. Typed transaction receipts are 'opaque' and have the data wrapped in a
+byte array. However, all receipt types ultimately contain the same four fields. With the
+removal of the bloom filter, the networking protocol now deviates from the encoding used
+by consensus, and there is no need to replicate the weird and expensive encoding used
+there. The proposed receipt encoding is just a flat list of the required data fields.
 
 ## Backwards Compatibility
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7642.md
```diff
@@ -1,7 +1,7 @@
 ---
 eip: 7642
-title: eth/69 - Drop pre-merge fields
-description: Drop unnecessary fields after the merge
+title: eth/69 - history expiry and simpler receipts
+description: Adds history serving window and removes bloom filter in receipt
 author: Marius van der Wijden (@MariusVanDerWijden), Felix Lange <fjl@ethereum.org>
```
lightclient, line 5:
```suggestion
author: Marius van der Wijden (@MariusVanDerWijden), Felix Lange (@fjl)
```
this allows you to merge changes to the eip via the bot

fjl: Yeah I know but I want to discuss a bit before merging.

jflo (approved):
Besu team has no objections.

smartprogrammer93:
@fjl any reason not to reopen EIP-7542.  it helps by also letting you know the latest block a node has. so syncing nodes can let you know what is the latest block they have.
Additionally, having this information only in the status message means that if the node ever updates the range of blocks it has, there is no way for peers to know about this update. with EIP-7542 there is a message nodes can send and get a response of the updated range.

benaadams (approved):
Generally supportive of this; however would prefer the block range, which would allow easier syncing from a syncing client

e.g. where the peer is not fully sync'd itself and syncing blocks in a full sync manner going forwards; so it may have some interesting blocks, but earliestBlock by itself will give a false impression of block availability

fjl:
OK, I can add a notification about available range.

EIPS/eip-7642.md
```diff
@@ -1,7 +1,7 @@
...
 discussions-to: https://ethereum-magicians.org/t/eth-70-drop-pre-merge-fields-from-eth-protocol/19005
 status: Stagnant
```
g11tech, line 7: dont want to update the status?

benaadams (approved):

smartprogrammer93 (approved):

g11tech (approved):

MariusVanDerWijden (approved):

PR #9238, opened 2025-01-13
Update EIP-4762: Fix references to CALLCODE and wording around its value-transfer costs

lu-pinto:
This is to correct a previous PR that was merged without applying these changes. I also found what I believe some missing references for CALLCODE since wildcarded `*CALL` does not cover this case.


gballet (approved):
LGTM

PR #9240, opened 2025-01-14
Update EIP-7840: Add BaseFeeUpdateFraction

MarekM25:

--

 - The build passes.

### DIFF
--- a/EIPS/eip-7840.md
+++ b/EIPS/eip-7840.md
@@ -24,7 +24,7 @@ target blob count per block and max blob count per block for each fork.
 
 ## Specification
 
-Extend the client configuration files with the object `blobSchedule` with the
+Extend client configuration files with the object `blobSchedule`, which has the
 following shape:
 
 ```json
@@ -32,18 +32,19 @@ following shape:
   "cancun": {
     "target": 3,
     "max": 6,
-    "baseFeeUpdateFraction": 3338477
+    "updateFraction": 3338477
   },
   "prague": {
     "target": 6,
     "max": 9,
-    "baseFeeUpdateFraction": 5007716
+    "updateFraction": 5007716
   }
 }
 ```
 
-Clients **MUST** provide a way to configure target, max and baseFeeUpdateFraction per-fork.
-This EIP does not define the behavior when the configuration is missing; clients are free to choose how to handle this situation.
+Clients must configure the target, max and updateFraction per-fork. The behavior
+when the configuration is missing or incomplete for a fork is undefined. Clients
+are free to choose how to handle this situation.
 
 ## Rationale
 
@@ -53,7 +54,7 @@ for various activities. For example, the `eth_feeHistory` RPC method returns a
 field `blobGasUsedRatio` that does require the max, even though the core
 protocol doesn't specifically need such value. Passing this value over the
 engine API every block seem overkill so we believe a configuration value is a
-good middle ground. Additionally, the `baseFeeUpdateFraction` parameter was added to adjust the responsiveness of blob gas pricing per fork.
+good middle ground. Additionally, the `updateFraction` parameter was added to adjust the responsiveness of blob gas pricing per fork.
 
 ## Backwards Compatibility
 
--- a/EIPS/eip-7840.md
+++ b/EIPS/eip-7840.md
@@ -32,17 +32,17 @@ following shape:
   "cancun": {
     "target": 3,
     "max": 6,
-    "updateFraction": 3338477
+    "baseFeeUpdateFraction": 3338477
   },
   "prague": {
     "target": 6,
     "max": 9,
-    "updateFraction": 5007716
+    "baseFeeUpdateFraction": 5007716
   }
 }
 ```
 
-Clients must configure the target, max and updateFraction per-fork. The behavior
+Clients must configure the target, max and baseFeeUpdateFraction per-fork. The behavior
 when the configuration is missing or incomplete for a fork is undefined. Clients
 are free to choose how to handle this situation.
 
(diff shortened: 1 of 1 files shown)


tanishqjasoria (approved):

g11tech (approved):
lgtm from ethjs

siladu (approved):
LGTM from Besu

jochem-brouwer:
> When there is no explicit configuration for the current fork, use the last specified fork value. If no last value is specified, set both to zero.

(From the current EIP). This "talks" about "both", maybe for clarification mention that then the fraction is undefined (cannot divide by 0).

I also think for completeness it would be great if in the Rationale there is an addition that this new param also allows to tweak the gas adjustments.

somnathb1:
LGTM

rkrasiuk (approved):

MarekM25:
> Member

@jochem-brouwer Thanks! Regarding defaults without configuration, I think the correct behavior for clients should be to throw an exception. In this way, we will notice quickly misconfigured tests/tooling etc. So, if 4844 is enabled and there's no blob configuration, let's throw an exception. I believe the EIP shouldn't force clients to handle defaults in a specific way, and we have soft consensus on either throwing an exception or using zeroes/undefined values.

Another option could be falling back to the Cancun blob parameters.

yperbasis (approved):

jochem-brouwer:
So this raises some questions about what is defined regarding the configuration and what is not defined. In 4844, `MAX_BLOB_GAS_PER_BLOCK` and `TARGET_BLOB_GAS_PER_BLOCK` are in the EIP, but the question is if these should also be interpreted as "default configuration" for 4844. That is: if the supplied config in this EIP (7840) does not contain a `cancun` field, should it fall back on 4844 or should it throw? The same goes for EIP-7691, it has `MAX_BLOB_GAS_PER_BLOCK` and `TARGET_BLOB_GAS_PER_BLOCK` defined in the EIP, should we also consider this the default values in here (7840)? 

The default configuration of the blob schedule should be:

```json
"blobSchedule": {
  "cancun": {
    "target": 3,
    "max": 6,
    "baseFeeUpdateFraction": 3338477
  },
  "prague": {
    "target": 6,
    "max": 9,
    "baseFeeUpdateFraction": 5007716
  }
}
```

(i.e. this is as-defined currently in this EIP). A small note: we now define `target`/`max` representing the count of blobs, in order to calculate the gas and to apply the formulas in 4844, we have to thus multiply with `GAS_PER_BLOB` (from 4844) to get the relevant `MAX_BLOB_GAS_PER_BLOCK` and `TARGET_BLOB_GAS_PER_BLOCK` parameters to use in the formulas defined in 4844.

Configuration behavior
----------------------

The question now is, how should clients handle this EIP / the configuration? Note that 4844 (blobs) somewhat define the configuration parameters for Cancun and 7691 (blob schedule upgrade) define them for Prague. However since these are different EIPs than this one, and in order to be explicit, let's self-contain this EIP:

- The configuration above is the "default" configuration
- It is possible to override those values (so for instance a network on Prague with 12 max blobs)

The rules for overriding holds, so if there is an override JSON which only defines `cancun` for instance, then `prague` will use the params read from the default JSON above, but `cancun` will use the override.

If there is a fork (for instance: `amsterdam`) which is active but not defined in the default JSON (above), then the configuration should throw.

This also means that for future forks we should update this EIP to reflect the correct "base" JSON.

jochem-brouwer:
> @jochem-brouwer Thanks! Regarding defaults without configuration, I think the correct behavior for clients should be to throw an exception. In this way, we will notice quickly misconfigured tests/tooling etc. So, if 4844 is enabled and there's no blob configuration, let's throw an exception. I believe the EIP shouldn't force clients to handle defaults in a specific way, and we have soft consensus on either throwing an exception or using zeroes/undefined values.
> 
> Another option could be falling back to the Cancun blob parameters.

Yes, I agree. Clients should explicitly provide a blob schedule. If Cancun / 4844 is active but there is no field in the config, then throw. If Prague / 7691 is active but there is no field in the config, then also throw. For me this would also hold for future forks to explicitly define the schedule in the config.

siladu:
This is an informational EIP and thus I don't think we should specify default/error behaviour here. If we do, then clients are not required to follow it so it doesn't achieve the goal of aligning behaviour. I think the preferred approach is to leave it up the clients.

If there is value in throwing for the user of a particular client that's fine, but what is the value in cross-client alignment on this?

Current Besu behaviour is to not throw, but instead default to the appropriate fork's hardcoded values (as specified in EIP-4844 and EIP-7691).

Using the blobSchedule then becomes a way to override mainnet values with something non-standard, mainly for testing purposes.

EIPS/eip-7840.md
```diff
@@ -30,17 +31,20 @@ following shape:
...
+    "max": 9,
+    "baseFeeUpdateFraction": 5007716
   }
 }
 ```
 
-When there is no explicit configuration for the current fork, use the last
-specified fork value. If no last value is specified, set both to zero.
+Clients **MUST** explicitly configure blob parameters in the configuration for Ethereum mainnet.
+If there is no explicit configuration for the current fork, the client should use the last specified fork value.
```
jochem-brouwer, line 46: I would add here that the "explicit configuration" means: `being explicit means that for every fork where we do change the configuration, we must configure it—including Cancun.`

EIPS/eip-7840.md
```diff
@@ -30,17 +31,20 @@ following shape:
...
-    "max": 9
+    "max": 9,
+    "baseFeeUpdateFraction": 5007716
   }
 }
 ```
 
-When there is no explicit configuration for the current fork, use the last
-specified fork value. If no last value is specified, set both to zero.
+Clients **MUST** explicitly configure blob parameters in the configuration for Ethereum mainnet.
```
jochem-brouwer, line 45: Using "explicitly configure" and "explicit configuration" might lead to confusion, I think we are talking about slightly different things here. The point here is that the `blobSchedule` should be present in config (?) and it should be valid (the Cancun entry should be there)

jochem-brouwer (commented):
Two comments to clarify the text

jochem-brouwer (approved):
LGTM :+1: :smile:

lightclient (approved):
I would have preferred something shorter like `updateFraction`, but I'm late to the party here so will accept as is :)

PR #9241, opened 2025-01-14
Add EIP: Delayed Execution Layer State Root

charlie-paradigm:
This EIP implements a delayed state root. It is ready for draft status and discussion.

### DIFF
--- a/EIPS/eip-7862.md
+++ b/EIPS/eip-7862.md
@@ -24,7 +24,7 @@ The execution layer (EL) `state_root` contained within the `ExecutionPayload` tr
 
 The EL `state_root` must be computed by block builders (and verified by relays, validators, and light clients) in near-real-time. This computation accounts for a large fraction of block production and verification time, especially in an MEV-Boost environment. This overhead is also a challenge for efforts to enable real-time ZK proving of the chain.
 
-By delaying the EL `state_root` reference by one block, we can remove EL `state_root` calculation overhead from the critical path of block production. Instead, clients can pipeline calculation of block `n-1`'s EL `state_root` during the idle slot time of block `n`.
+By delaying the EL `state_root` reference by one block, we can remove EL `state_root` calculation overhead from the critical path of block production. Instead, clients can pipeline calculation of block `n-1`'s EL `state_root` during the idle slot time before block `n`.
 
 This change will lower latency at the tip, allowing us to increase throughput and simplify the block production pipeline. It can also significantly accelerate the timeline of real-time ZK proving Ethereum.
 
--- a/EIPS/eip-7862.md
+++ b/EIPS/eip-7862.md
@@ -26,7 +26,7 @@ The EL `state_root` must be computed by block builders (and verified by relays, 
 
 By delaying the EL `state_root` reference by one block, we can remove EL `state_root` calculation overhead from the critical path of block production. Instead, clients can pipeline calculation of block `n-1`'s EL `state_root` during the idle slot time of block `n`.
 
-This change will lower latency at the tip, allowing us to increase throughput and simplify the block production pipeline. It can also significantly accelerate the timeline real-time ZK proving Ethereum.
+This change will lower latency at the tip, allowing us to increase throughput and simplify the block production pipeline. It can also significantly accelerate the timeline of real-time ZK proving Ethereum.
 
 ## Specification
 
@@ -54,7 +54,7 @@ This change will lower latency at the tip, allowing us to increase throughput an
        uint256 timestamp;
        bytes32 parentHash;
        address feeRecipient;
-       bytes32 stateRoot;  // now references the post-state root of block (n-1)
+       bytes32 stateRoot;  // now references *this* block's pre-state i.e. the post-state root of block (n-1)
      }
      ```
    
@@ -107,7 +107,7 @@ Upon the hard fork that activates this proposal (call this activation block `F`)
 1. **Reduce Latency at the Tip**
    
    - Proposers and builders no longer must compute and finalize the new state root within the same slot.  
-   - This speeds up block propagation, as peers can quickly validate the rest of the block without waiting on the state root calculation.
+   - This speeds up block validation, as validators can sign after checking the execution of the block and that its pre-state root matches their pre-computed expectation (i.e., without needing to wait to calculate a new state root).
 
 3. **Increase Throughput**
    
--- a/EIPS/eip-7862.md
+++ b/EIPS/eip-7862.md
@@ -1,7 +1,7 @@
 ---
-eip: XXXX
-title: "Delayed Execution Layer State Root"
-description: "Delays the ExecutionPayload's state root reference by one block."
+eip: 7862
+title: Delayed Execution Layer State Root
+description: Delays the ExecutionPayload's state root reference by one block.
 author: Charlie Noyes <charlie@paradigm.xyz>, Dan Robinson <dan@paradigm.xyz>, Justin Drake <justin@ethereum.org>, Toni Wahrstätter <toni@ethereum.org>
 discussions-to: https://ethereum-magicians.org/t/eip-XXXX/, https://ethresear.ch/t/proposal-delay-stateroot-reference-to-increase-throughput-and-reduce-latency/20490
 status: Draft
--- a/EIPS/eip-7862.md
+++ b/EIPS/eip-7862.md
@@ -12,7 +12,7 @@ created: 2024-12-23
 
 ## Abstract
 
-This EIP proposes that, for each block `n`, the `state_root` in block `n`'s `ExecutionPayload` references the **post-state of block `n-1`**. As a result, block `n` no longer needs to include or verify the newly computed state root for block `n`’s own transactions. This removes the expensive state root computation from the block-propagation critical path, reducing end-to-end latency for block production and potentially increasing throughput.
+This EIP proposes that, for each block `n`, the `state_root` in block `n`'s `ExecutionPayload` references the **post-state of block `n-1`**. As a result, block `n` no longer needs to compute & include or verify the post state root for block `n`. This removes the expensive state root computation from the block-propagation critical path, reducing end-to-end latency for block production and potentially increasing throughput.
 
 ## Motivation
 
--- a/EIPS/eip-xxxx.md
+++ b/EIPS/eip-xxxx.md
@@ -1,0 +1,136 @@
+---
+eip: XXXX
+title: "Delayed Execution Layer State Root"
+description: "Delays the ExecutionPayload's state root reference by one block."
+author: Charlie Noyes <charlie@paradigm.xyz>, Dan Robinson <dan@paradigm.xyz>, Justin Drake <justin@ethereum.org>, Toni Wahrstätter <toni@ethereum.org>
+discussions-to: https://ethereum-magicians.org/t/eip-XXXX/, https://ethresear.ch/t/proposal-delay-stateroot-reference-to-increase-throughput-and-reduce-latency/20490
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-12-23
+---
+
+## Abstract
+
+This EIP proposes that, for each block `n`, the `state_root` in block `n`'s `ExecutionPayload` references the **post-state of block `n-1`**. As a result, block `n` no longer needs to include or verify the newly computed state root for block `n`’s own transactions. This removes the expensive state root computation from the block-propagation critical path, reducing end-to-end latency for block production and potentially increasing throughput.
+
+## Motivation
+
+Currently, every Ethereum block includes two `state_root`. 
+
+The consensus layer (CL) `state_root` contained within the `BeaconBlock` tracks CL state (e.g., validator entries and exits).
+
+The execution layer (EL) `state_root` contained within the `ExecutionPayload` tracks the effect of transaction execution (e.g., account balances, code, etc.). 
+
+The EL `state_root` must be computed by block builders (and verified by relays, validators, and light clients) in near-real-time. This computation accounts for a large fraction of block production and verification time, especially in an MEV-Boost environment. This overhead is also a challenge for efforts to enable real-time ZK proving of the chain.
+
+By delaying the EL `state_root` reference by one block, we can remove EL `state_root` calculation overhead from the critical path of block production. Instead, clients can pipeline calculation of block `n-1`'s EL `state_root` during the idle slot time of block `n`.
+
+This change will lower latency at the tip, allowing us to increase throughput and simplify the block production pipeline. It can also significantly accelerate the timeline real-time ZK proving Ethereum.
+
+## Specification
+
+### Execution Layer Changes
+
+1. **`ExecutionPayload` State Root Field**  
+   - **Before** (in the current post-Merge design):  
+     ```solidity
+     struct ExecutionPayload {
+       uint256 blockNumber;
+       uint256 timestamp;
+       bytes32 parentHash;
+       address feeRecipient;
+       ...
+       bytes32 stateRoot;  // references the post-state root of *this* block's transactions
+       ...
+     }
+     ```
+   - **After** (conceptual):  
+     ```solidity
+     struct ExecutionPayload {
+       uint256 blockNumber;
+       uint256 timestamp;
+       bytes32 parentHash;
+       address feeRecipient;
+       ...
+       bytes32 stateRoot;  // now references the post-state root of block (n-1)
+       ...
+     }
+     ```
+    - In other words, block `n`’s `ExecutionPayload.stateRoot` must now point to the **execution-layer post-state of block `n-1`**, rather than referencing the new state after block `n`’s own transactions.
+
+2. **Block Header Validation**  
+   - **Remove** the rule requiring each block’s `stateRoot` to match the **post-state** produced by the transactions in *that same block*:  
+   - **Instead**, require that `payload_n.stateRoot == post_state(n-1)`:
+     ```python
+     def validate_execution_payload(payload_n, post_state_of_block_n_minus_1):
+         # Was:
+         #   assert compute_state_root(transactions_of_block_n) == payload_n.stateRoot
+         #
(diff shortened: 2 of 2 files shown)


EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,136 @@
...
+
+Currently, every Ethereum block includes two `state_root`. 
+
+The consensus layer (CL) `state_root` contained within the `BeaconBlock` tracks CL state (e.g., validator entries and exits).
+
+The execution layer (EL) `state_root` contained within the `ExecutionPayload` tracks the effect of transaction execution (e.g., account balances, code, etc.). 
+
+The EL `state_root` must be computed by block builders (and verified by relays, validators, and light clients) in near-real-time. This computation accounts for a large fraction of block production and verification time, especially in an MEV-Boost environment. This overhead is also a challenge for efforts to enable real-time ZK proving of the chain.
+
+By delaying the EL `state_root` reference by one block, we can remove EL `state_root` calculation overhead from the critical path of block production. Instead, clients can pipeline calculation of block `n-1`'s EL `state_root` during the idle slot time of block `n`.
```
terencechain, line 27: > idle slot time of block `n`.

I think this is `n-1`. At `n-1`, the block arrives around the 4th second of the slot, you verify everything except the state root, and then you have 8 more seconds in `n-1` to verify the state root

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,136 @@
...
+         # Was:
+         #   assert compute_state_root(transactions_of_block_n) == payload_n.stateRoot
+         #
+         # Now:
+         assert payload_n.stateRoot == post_state_of_block_n_minus_1
+     ```
+   - The actual new state root resulting from block `n` will only appear in block `n+1`.
+
+3. **Proactive Computation of Post-state Root**:
+   - To fully realize latency benefits, clients should compute the post-state root of block `n-1` during the idle time of slot `n`.
```
terencechain, line 76: > during the idle time of slot `n`. 

Same. I think this would be `n-1`, not `n`

EIPS/eip-xxxx.md
zzzzzzch: I think in this design the EL clients are calculating the state root of block `n-1` while building block `n` at the same time, and this process occurs during the last 8 seconds of block `n-1`. Is this correct?

g11tech: yea it should read:

Instead, clients can pipeline calculation of block `n`'s EL `state_root` during the idle slot time of block `n`.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,136 @@
...
+     ```
+   - **After** (conceptual):  
+     ```solidity
+     struct ExecutionPayload {
+       uint256 blockNumber;
+       uint256 timestamp;
+       bytes32 parentHash;
+       address feeRecipient;
+       ...
+       bytes32 stateRoot;  // now references the post-state root of block (n-1)
```
g11tech, line 56:
```suggestion
       bytes32 stateRoot;  // now references *this* block's pre-state i.e. the post-state root of block (n-1)
```
just a suggestion for more clarity

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,136 @@
...
+```solidity
+  ExecutionPayload(F).stateRoot = post_state(F-1)
+```
+- **Block `F+1`** will reference the post-state of block `F`, continuing the new delayed root scheme from that point onward.
+
+## Rationale
+
+1. **Reduce Latency at the Tip**  
+   - Proposers and builders no longer must compute and finalize the new state root within the same slot.  
+   - This speeds up block propagation, as peers can quickly validate the rest of the block without waiting on the state root calculation.
```
fradamt, line 100: It doesn't actually speed up block propagation, because forwarding a block doesn't require fully verifying it, the main check is just the proposer's signature

EIPS/eip-xxxx.md
g11tech, line 56: i have another Q: 

since the block builders will have the stateroot because they can't include a tx before executing a previous one because of previous txs affecting any prestate of the tx, so why should we not include the stateroot in it

since the logs/requests etc are all dependent on the correct execution. Ofcourse the attesters can still do delayed verification. 

may we we just add a pre-stateroot field because proofs of correct execution results against this block might require this block's post stateroot

Thegaram: > since the block builders will have the stateroot because they can't include a tx before executing a previous one because of previous txs affecting any prestate of the tx, so why should we not include the stateroot in it

They only need the state, not the state root.

Currently they'd need to compute the state root before completing the block building process. Now they don't need to compute it eagerly (they can do it async).

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,136 @@
...
+2. **Block Header Validation**  
+   - **Remove** the rule requiring each block’s `stateRoot` to match the **post-state** produced by the transactions in *that same block*:  
+   - **Instead**, require that `payload_n.stateRoot == post_state(n-1)`:
+     ```python
+     def validate_execution_payload(payload_n, post_state_of_block_n_minus_1):
+         # Was:
+         #   assert compute_state_root(transactions_of_block_n) == payload_n.stateRoot
+         #
+         # Now:
+         assert payload_n.stateRoot == post_state_of_block_n_minus_1
```
Thegaram, line 71: Does this mean that if a future protocol upgrade changes EVM behavior, then nodes that did not upgrade would fork off the network with 1 block delay? (Though I don't think this is an issue.)

omerfirmak: I guess that is true only if the protocol change results in a mismatched state but doesn't change the execution flow of the transactions.

Thegaram: Right, so for instance adding a new tx type could result in a fork exactly at the upgrade block.

But adding a new opcode or precompile, or changing the gas schedule, could result in a fork one block later.

EIPS/eip-xxxx.md
g11tech: yes I get that, but having post stateroot in block is a good thing for creating proofs  merkle and ZK against the current block header itself.

yes it would matter how hard it is to create a stateroot, my supposition is : its not really that expensive and doesn't add any real latency but then depends on how big the block is, but if its vialble we should include it

EIPS/eip-7862.md
```diff
@@ -0,0 +1,136 @@
...
+2. **Block Header Validation**  
+   - **Remove** the rule requiring each block’s `stateRoot` to match the **post-state** produced by the transactions in *that same block*:  
+   - **Instead**, require that `payload_n.stateRoot == post_state(n-1)`:
+     ```python
+     def validate_execution_payload(payload_n, post_state_of_block_n_minus_1):
+         # Was:
+         #   assert compute_state_root(transactions_of_block_n) == payload_n.stateRoot
+         #
+         # Now:
+         assert payload_n.stateRoot == post_state_of_block_n_minus_1
```
g11tech, line 71: the current block execution validations also validates receipts, execution requests root, header has logbloom etc so many other fields which are the result of block execution and get validated so I am assuming that those validations stay in?

EIPS/eip-xxxx.md
Thegaram: Sorry I'm not sure I understand. Do you suggest to include both pre state root and post state root of the current block in its header?

> having post stateroot in block is a good thing for creating proofs merkle and ZK against the current block header itself

Do you mean verifying blocks with a zk proof instead of re-executing them? That's not necessarily in conflict with this EIP, nodes can still receive the state root together with the proof (not as part of the header but on some other channel).

EIPS/eip-xxxx.md
charlie-paradigm: I think it might be clearer to specify post-state root, rather than the parameter.

"Instead, clients can pipeline calculation of block `n-1`'s EL **post-state** root during the idle slot time of block `n`."

Would you guys be happy with this?

To clarify what happens is that:

When n-1 arrives, it contains the EL pre-state root of n-1.

While waiting for block n to arrive, the client can calculate the post-state root of n-1 (equivalently, pre-state of n).

When block n arrives, it can verify that it contains the correct post-state of n-1 (equivalently, pre-state of n).

EIPS/eip-xxxx.md
charlie-paradigm: I don't think that I agree. 

You could say "compute the pre-state root of n during the idle slot time of n"

and that might be clearer?

Muaytie23 (approved):

EIPS/eip-7862.md
charlie-paradigm: Yes

jochem-brouwer: I did not realize this until this comment, because I first assumed that this could mean that tx execution could happen after publishing the block. But this is not possible, since we still need receipt roots.

Are there any numbers how much the time is of calculating the state root? This is a simplification, but I would assume that most clients would execute txs and write state changes to a cache. Only at the end of running the txs this cache is flushed to disk in order to calculate the new state root. I would honestly have no idea what the ratio is between executing the transactions and flushing the new cache to disk / calculating the root and would love to see some numbers on this.

EIPS/eip-7862.md
```diff
@@ -0,0 +1,149 @@
...
+created: 2024-12-23
+---
+
+## Abstract
+
+This EIP proposes that, for each block `n`, the `state_root` in block `n`'s `ExecutionPayload` references the **post-state of block `n-1`**. As a result, block `n` no longer needs to include or verify the newly computed state root for block `n`’s own transactions. This removes the expensive state root computation from the block-propagation critical path, reducing end-to-end latency for block production and potentially increasing throughput.
+
+## Motivation
+
+Currently, every Ethereum block includes two `state_root`.
```
jochem-brouwer, line 19: I find the wording of this statement incorrect, EL blocks have only one state root, and this is the state root of Ethereum's "world state" (MPT of accounts)

EIPS/eip-7862.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Motivation
+
+Currently, every Ethereum block includes two `state_root`. 
+
+The consensus layer (CL) `state_root` contained within the `BeaconBlock` tracks CL state (e.g., validator entries and exits).
+
+The execution layer (EL) `state_root` contained within the `ExecutionPayload` tracks the effect of transaction execution (e.g., account balances, code, etc.). 
+
+The EL `state_root` must be computed by block builders (and verified by relays, validators, and light clients) in near-real-time. This computation accounts for a large fraction of block production and verification time, especially in an MEV-Boost environment. This overhead is also a challenge for efforts to enable real-time ZK proving of the chain.
```
jochem-brouwer, line 25: I'd love to see some numbers here, what is this large fraction? If splitting the time between receiving a block, executing the txs, and calculating the root, what would these numbers be?

EIPS/eip-7862.md
```diff
@@ -0,0 +1,149 @@
...
+
+1. **Light clients**
+   
+   - Must accept that a block’s final state root is not published until the subsequent block, adding up to a slot’s worth of latency (e.g. 12 seconds).  
+   - In practice, many light client protocols already tolerate slight delays in proof availability.
+   - Notably, many other blockchain networks already operate on a delayed state root and it has worked fine. One such network, Cosmos, is a multichain architecture based entirely on light-client interoperability with a delayed state root.
+
+3. **Reliant Contracts**
+
+   - We are not aware of any common contract types that assume the availability of same-block state roots.
```
jochem-brouwer, line 145: Contracts cannot read a block state root, so I'm not sure if this is relevant?

EIPS/eip-xxxx.md
terencechain: I looked at this again and think what you have is fine. If we want to be clearer, we could rephrase it as the following: 

"As a proposer/builder for `n`, assuming it receives a block for `n-1` during slot `n-1`, to fill in the `state_root` field, they should compute the post-state root at `n-1` during the idle time of `n-1`"

I agree with what you said here: https://github.com/ethereum/EIPs/pull/9241#discussion_r1919081235

EIPS/eip-xxxx.md
g11tech: yes I mean to specify both because the block builder will have the final execution results anyway when its done adding txs to the block.

for the proof, it needs to be anchored in post state root. Although receipts should have the entire execution traced but thats not true of system updates which are purely reflected in stateroot.

EIPS/eip-xxxx.md
g11tech: so all we are delaying for the attesting/verifying validator is the stateroot calc? how significant is that as percentate of block building time?

we could infact delay the entire execution for the attesting/verifying validator (which means they can execute/verify bigger blocks on the commodity hardware while the specialized builders can build a gas chugging block

terencechain: I don't think we can delay attesting unless we do eip7732

EIPS/eip-xxxx.md
g11tech: ok now I understand that this EIP is trying to save post state root compute time even for builder. is that correct?

if so then as I asked in a previous comment how significant this time would be (so benchmarks against some typical blocks with 30m 60m... 100m 1g gas i guess) on a latest commondity hardware (which one would expect least of block builders)

I guess that would bring forward the proper justification of the EIP

EIPS/eip-7862.md
g11tech: yea  agree with @jochem-brouwer  we should see what is the benchmark delay for batched state root calculation

EIPS/eip-7862.md
```diff
@@ -0,0 +1,149 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-7862-delayed-execution-layer-state-root/22559
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-12-23
+---
+
+## Abstract
+
+This EIP proposes that, for each block `n`, the `state_root` in block `n`'s `ExecutionPayload` references the **post-state of block `n-1`**. As a result, block `n` no longer needs to include or verify the newly computed state root for block `n`’s own transactions. This removes the expensive state root computation from the block-propagation critical path, reducing end-to-end latency for block production and potentially increasing throughput.
```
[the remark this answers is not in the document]
g11tech: this clarifies to me what exactly EIP is proposing.

EIPS/eip-7862.md
g11tech: the EIP is talking about the full block as it references `ExecutionPayload`

EIPS/eip-7862.md
g11tech: for purposes of draft EIP can we add a <-- TODO benchmark --> here?

EIPS/eip-7862.md
```diff
@@ -0,0 +1,149 @@
...
+
+Currently, every Ethereum block includes two `state_root`. 
+
+The consensus layer (CL) `state_root` contained within the `BeaconBlock` tracks CL state (e.g., validator entries and exits).
+
+The execution layer (EL) `state_root` contained within the `ExecutionPayload` tracks the effect of transaction execution (e.g., account balances, code, etc.). 
+
+The EL `state_root` must be computed by block builders (and verified by relays, validators, and light clients) in near-real-time. This computation accounts for a large fraction of block production and verification time, especially in an MEV-Boost environment. This overhead is also a challenge for efforts to enable real-time ZK proving of the chain.
+
+By delaying the EL `state_root` reference by one block, we can remove EL `state_root` calculation overhead from the critical path of block production. Instead, clients can pipeline calculation of block `n-1`'s EL `state_root` during the idle slot time of block `n`.
```
g11tech, line 27:
```suggestion
By delaying the EL `state_root` reference by one block, we can remove EL `state_root` calculation overhead from the critical path of block production. Instead, clients can pipeline calculation of block `n-1`'s EL `state_root` during the idle slot time before block `n` arrival.
```

EIPS/eip-7862.md
```diff
@@ -0,0 +1,149 @@
...
+
+This change will lower latency at the tip, allowing us to increase throughput and simplify the block production pipeline. It can also significantly accelerate the timeline of real-time ZK proving Ethereum.
```
g11tech, line 29: how can it accelerate real-time ZK proving?

EIPS/eip-xxxx.md
g11tech: so we still execute the block before attesting -- only stateroot is not required.

since the effects of EIP 4788 and 2935, withdrawals, block tips coinbase rewards will be only reflected in state root (lets say no txs uses them and hence won't be in reflected in receipts etc) the block execution can't be fully verified unless we have a root/commitment to those changes and hence a validator could vote on a bad/buggy block execution.

Ideally we we should have a root of entire statediff/witnesses as block property which will sort of help close out the execution issue (and could be much faster than stateroot compute which would require harddisk accesses which is what this EIP will actually optimize in my opinion)

EIPS/eip-xxxx.md
g11tech: for purposes of accepting EIP draft just add a `<-- TODO benchmark -->` comment

EIPS/eip-xxxx.md
fradamt: It should be during the idle time of slot `n-1`. Slots work like this @charlie-paradigm:
- 0-4s: block propagation
- 4-12s: attestation aggregation

Between 0-4s you'd still get a block, execute it and attest to it, you just wouldn't be required to compute a state root yet. The computation should then be done between 4-12s of the same slot, no? E.g. the proposer/builder of slot n would want to be done with this by the time it needs to propose, which is *the beginning* of slot n.

EIPS/eip-7862.md
charlie-paradigm: We get more of the slot time to calculate it async, rather than a proof needing to be made after block building is complete (probably in <100ms)

g11tech (approved):

PR #9243, opened 2025-01-15
Update EIP-7691: clarify that all 3 EL constants change in Prague

pdobacz:
Otherwise the doc would say:

> "The execution clients would continue to use `MAX_BLOB_GAS_PER_BLOCK` (...) for the cancun fork"

while `MAX_BLOB_GAS_PER_BLOCK` is given a new value, which threw me off for a second.


nerolation (commented):
lgtm, waiting for 1 more review.

pdobacz:
Validation complains about the presence of reference to the peerDAS EIP-7594, which is now stagnant. I think it is fine to drop the specific EIP reference, but I defer to the authors.

PR #9246, opened 2025-01-15
Add EIP: Block-level Warming

nerolation:
This EIP proposes to keep accessed addresses and storage keys warm over the duration of a block.

### DIFF
--- a/EIPS/eip-7863.md
+++ b/EIPS/eip-7863.md
@@ -2,8 +2,8 @@
 eip: 7863
 title: Block-level Warming
 description: Warm addresses and storage keys over the duration of a block
-author: Toni Wahrstätter (@nerolation), Alex Stokes (@ralexstokes), Ansgar Dietrichs (@adietrichs)
-discussions-to: ""
+author: Toni Wahrstätter (@nerolation), Jochem Brouwer (@jochem-brouwer), Alex Stokes (@ralexstokes), Ansgar Dietrichs (@adietrichs)
+discussions-to: https://ethereum-magicians.org/t/eip-7863-block-level-warming/22572
 status: Draft
 type: Standards Track
 category: Core
@@ -30,8 +30,9 @@ As of Jan. 2025, the empirically messured efficiency gains are around 5%.
 ### Mechanics
 
 When a storage slot is accessed within a block:
-1. The first access to a slot in a block incurs the cold access cost as of [EIP-2929](./eip-2929).
-2. All subsequent accesses to the same slot within the same block incur only the warm access cost as of [EIP-2929](./eip-2929).
+
+1. The first access to a slot in a block incurs the cold access cost as of [EIP-2929](./eip-2929.md).
+2. All subsequent accesses to the same slot within the same block incur only the warm access cost as of [EIP-2929](./eip-2929.md).
 3. The warm/cold status resets at block boundaries
 
 ### Block Processing
@@ -41,14 +42,14 @@ When a storage slot is accessed within a block:
 2. For each transaction in the block:
    * Before processing storage access:
      * Check if touched address is in `block_level_accessed_addresses`
-     * If yes: charge `GAS_WARM_ACCESS`
+     * If yes: charge `WARM_STORAGE_READ_COST`
      * If no:
-       * Charge `GAS_COLD_ACCOUNT_ACCESS`
+       * Charge `COLD_ACCOUNT_ACCESS_COST`
        * Add address to `block_level_accessed_addresses`
      * Check if storage key is in `block_level_accessed_storage_keys`
-     * If yes: charge `GAS_WARM_ACCESS`
+     * If yes: charge `WARM_STORAGE_READ_COST`
      * If no:
-       * Charge `GAS_COLD_SLOAD`
+       * Charge `COLD_SLOAD_COST`
        * Add storage key to `block_level_accessed_storage_keys`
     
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+
+As of Jan. 2025, the empirically messured efficiency gains are around 5%.
+
+
+## Specification
+
+### Mechanics
+
+When a storage slot is accessed within a block:
+1. The first access to a slot in a block incurs the cold access cost as of [EIP-2929](./eip-2929).
```
jochem-brouwer, line 33:
```suggestion
1. The first access to a slot in a block incurs the cold access cost as of [EIP-2929](./eip-2929.md).
```

jochem-brouwer: Except if it has been added to the access list. This raises the question also what to do with the access lists: if a tx has an address/storage slot in the access list, do we put it at the warm access of the block level before executing that transaction? (So, even if the transaction does not access the slot/address at all, is this still added to the block warm addresses/slots?)

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+2. All subsequent accesses to the same slot within the same block incur only the warm access cost as of [EIP-2929](./eip-2929).
```
jochem-brouwer, line 34:
```suggestion
2. All subsequent accesses to the same slot within the same block incur only the warm access cost as of [EIP-2929](./eip-2929.md).
```

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+3. The warm/cold status resets at block boundaries
+
+### Block Processing
+
+1. At the start of each block:
+   * Initialize two empty sets `block_level_accessed_addresses` and `block_level_accessed_storage_keys`
+2. For each transaction in the block:
+   * Before processing storage access:
+     * Check if touched address is in `block_level_accessed_addresses`
+     * If yes: charge `GAS_WARM_ACCESS`
```
jochem-brouwer, line 44: This constant is not defined in EIP-2929, I think what is meant (will also suggest the params for similar params here)

```suggestion
     * If yes: charge `WARM_STORAGE_READ_COST`
```

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+     * If no:
+       * Charge `GAS_COLD_ACCOUNT_ACCESS`
```
jochem-brouwer, line 46:
```suggestion
       * Charge `COLD_ACCOUNT_ACCESS_COST`
```

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+       * Add address to `block_level_accessed_addresses`
+     * Check if storage key is in `block_level_accessed_storage_keys`
+     * If yes: charge `GAS_WARM_ACCESS`
```
jochem-brouwer, line 49:
```suggestion
     * If yes: charge `WARM_STORAGE_READ_COST`
```

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+     * If no:
+       * Charge `GAS_COLD_SLOAD`
```
jochem-brouwer, line 51:
```suggestion
       * Charge `COLD_SLOAD_COST`
```

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+
+When a storage slot is accessed within a block:
+1. The first access to a slot in a block incurs the cold access cost as of [EIP-2929](./eip-2929).
+2. All subsequent accesses to the same slot within the same block incur only the warm access cost as of [EIP-2929](./eip-2929).
+3. The warm/cold status resets at block boundaries
+
+### Block Processing
+
+1. At the start of each block:
+   * Initialize two empty sets `block_level_accessed_addresses` and `block_level_accessed_storage_keys`
```
jochem-brouwer, line 40: I know this is rather implicit, but the `block_level_accessed_storage_keys` is rather an item of a Map from addresses to the Set of accessed storage keys of that address.

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+
+The implementation modifies the block execution process to maintain block-level sets of accessed addresses and storage slots. 
+
+#### Block-Level Storage Management
+
+```python
+def apply_body(...):
+    # Initialize block-level tracking sets
+    block_level_accessed_addresses = set()
+    block_level_accessed_storage_keys = set()
```
jochem-brouwer, line 65: Should likely be a map from addresses to the accessed storage keys per address?

EIPS/eip-9999.md
jochem-brouwer: Never mind, from the code below it is clear these should be added.

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+    # Add block-level pre-accessed slots
+    preaccessed_addresses.add(env.block_level_accessed_addresses)
+    preaccessed_storage_keys.add(env.block_level_accessed_storage_keys)
+    
+    # Handle access lists from transaction
+    ...
+```
+
+This adds the block-level-accessed addresses and storage keys to the preaccessed addresses and storage keys.
+As a result, from the perspective of a transaction, block-level accessed addresses and storage keys are treated the same as precompiles or the coinbase address.
```
jochem-brouwer, line 99: What if the transaction itself reverts? In EIP-2929 we have the logic that if a sub-call reverts, all the accessed addresses/slots are now cold. In case of a transaction failure, what happens here? Are all the accessed addresses/slots in this specific addresses also removed? Will put a point at the conclusion of this review if we should consider removing this revert logic altogether since that would fit within the scope/idea of this PR.

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+    )
+```
+
+The message call processing tracks accessed addresses and storage keys during execution, which are then propagated back up to the transaction level and ultimately to the block level.
+
+## Rationale
+
+The proposal builds on several key observations:
+
+1. **Caching Efficiency**: Modern Ethereum clients already implement sophisticated caching mechanisms at the block level. Extending address and storage key warming to match this caching behavior better aligns gas costs with actual computational costs.
```
jochem-brouwer, line 116: Right, this caching point would also show why I think we need to remove the EIP-2929 reverting mechanism. The "cold" gas cost represents the expensive price we have to pay if we have to read the storage slot from disk. Even if a sub-call reverts, then most (all (?)) clients should have kept this in memory and should not read this again from disk. Therefore, only the first access should charge the cold cost, irrelevant if it invalidates this sub-call / transaction later.

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+
+2. **Backward Compatibility**: The worst-case scenario for any transaction remains unchanged - it will never pay more than the current cold access cost for its first access to a slot.
+
+3. **First Access Warms For All**: Under the proposed mechanism, the first transaction in a block that accesses and warms multiple addresses or storage slots bears the entire cost of warming, even if subsequent transactions benefit from those already-warmed slots without incurring additional costs. This approach is deemed acceptable because early execution within a block—where warming typically occurs—is generally occupied for transactions placed by professional builders aiming for top-of-block execution. Since the resulting cost discrepancy is relatively minor, a more complex cost-sharing model is avoided in favor of maintaining simplicity.
```
jochem-brouwer, line 120: This is an interesting comment, since competition will thus now also be much more dependent upon what storage was added and what not. Therefore to pack more logic / more transactions this thus means that transactions which interact with already warm addresses/slots are more preferred than those who are not. (This would be an interesting research also, because this could theoretically mean that (in an extreme example) for block A we will only see txs with contract Y, and then for block B only those for block Z, but this obviously also depends upon the goal/optimization target of the builder)

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+An alternative to the one-warms-for-all mechanism involves distributing the warming costs across all accesses within a block. In this approach, each transaction must at least be able to pay the cold access cost. Once all transactions in the block are executed, the total warming costs are evenly distributed, and transaction originators are refunded accordingly.
```
jochem-brouwer, line 121: From an economic perspective this is fair, but:

- This is currently about `(2100 - 100) = 2000` cost for storage and `(2600 - 100) = 2500` for accounts, thus we would have to distribute these 2k/2.5k gas costs over these txs, and those are only per slot/account once
- This seems super complex
- How to deal with accounting errors, for instance 3 accounts pay the 2k cost for a storage? 
- This would likely mean that on the gas accounting when executing, each tx will have to pay the cold gas upfront
- This will thus resort back to the original 2929 pricing where on the tx-level everything is cold, and will thus not lead to measured empirical efficiency of 5%

EIPS/eip-9999.md
```diff
@@ -0,0 +1,150 @@
...
+
+4. **Alternative Block Warming Windows**: Instead of applying warming at the block level, more advanced multi-block warming approaches can be considered. Potential options include:
```
jochem-brouwer, line 123: I think first moving to "block-warm" state instead of "tx-warm" state is preferred. But this is certainly something to think about. The main complexity problem I see with this idea is that this ring buffer needs to be saved somewhere, and it needs to be part of the consensus, such that clients once shut down and starting up again can lookup whether accounts/storage are warm or not.

jochem-brouwer (commented):
I love this EIP! :smile: 

I think in general the idea here is that we only charge the cold gas cost once, and this is used to reflect when we pull this storage from disk. If that is the case, then I also propose we remove the logic from 2929  which reverts the journal of warm slots in case a sub-call reverts. This would make the implementation of 2929 for this fork likely much easier (since the revert mechanism does not have to be coded) and we thus now have that cold load costs for an account/slot are at most only charged once per blob.

EIPS/eip-9999.md
nerolation: Yeah true, although, one could also concat contract+storage key and treat the resulting bytes as a unique storage key

EIPS/eip-9999.md
nerolation: This is a very good point. I agree that it might make sense to keep slots warm even if the tx reverts. Will add it to the rational for now

EIPS/eip-9999.md
nerolation: Yeah the ring buffer adds some complexity for sure.
@pipermerriam also had a comment regarding additional complexity for statelessness [here](https://ethresear.ch/t/block-level-warming/21452/2?u=nero_eth).

EIPS/eip-9999.md
nerolation: I'm not a fan of the refund-approch neither. To me it feels like a clear pareto improvement even without doing refunds.

EIPS/eip-9999.md
nerolation: Put it into the rational section for now but might be worth to directly adopt "reverts don't unwarm slots" in the EIP

EIPS/eip-9999.md
nerolation: yeah, this can be further clarified in the EIP. Access lists would then warm for all consecutive txs too.

g11tech:
how different is this from https://eips.ethereum.org/EIPS/eip-3584 ? if not much can we revive the same EIP?

nerolation:
> how different is this from https://eips.ethereum.org/EIPS/eip-3584 ? if not much can we revive the same EIP?

I think they are different enough to keep separate. 3584 is more to share the access list mechanism over the block while  this EIP is about doing this without using access lists though a first-access-pays mechanism + no arguments in the header. I can see a realistic world with both of those EIPs being shipped.

EIPS/eip-9999.md
pipermerriam: I don't see any problems with this as long as it only remains single-block.

g11tech (approved):

PR #9250, opened 2025-01-17
Update EIP-7702: Clarify CODE* operations behavior

gumb0:
Clarify that `CODE*` operate on the code being executed, and not on the delegation designator.

(Tests' expectations agree with this.)

This clarification is motivated by confusion observed in this document: https://hackmd.io/5kcrdKgaQHSjrQ9MtAlBfg?comment=37baf7a2-fd97-4e3a-9972-e4a86c92ce8b

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -91,6 +91,8 @@ The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3
 
 For example, `EXTCODESIZE` would return `23` (the size of `0xef0100 || address`), `EXTCODEHASH` would return `keccak256(0xef0100 || address)`, and `CALL` would load the code from `address` and execute it in the context of `authority`.
 
+`CODESIZE` and `CODECOPY` instructions operate on executable code, as before. *Note that in a delegated execution `CODESIZE` and `CODECOPY` produce different result comparing to `EXTCODESIZE` and `EXTCODECOPY` of execution target.*
+
 In case a delegation designator points to a precompile address, retrieved code is considered empty and `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL` instructions targeting this account will execute empty code, i.e. succeed with no execution given enough gas.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7702.md
```diff
@@ -90,6 +90,8 @@ The delegation designation uses the banned opcode `0xef` from [EIP-3541](./eip-3
 
 For example, `EXTCODESIZE` would return `2` (the size of `0xef01`) instead of `23` which would represent the delegation designation, `EXTCODEHASH` would return `0xeadcdba66a79ab5dce91622d1d75c8cff5cff0b96944c3bf1072cd08ce018329` (`keccak256(0xef01)`), and `CALL` would load the code from `address` and execute it in the context of `authority`.
 
+`CODESIZE` and `CODECOPY` instructions operate on executable code, as before. *Note that in a delegated execution `CODESIZE` and `CODECOPY` produce different result comparing to `EXTCODESIZE` and `EXTCODECOPY` of execution target.*
```
jochem-brouwer, line 93: I did not realize this before, but this is the first time we have a case where `CODECOPY`/`CODESIZE` vs `EXTCODECOPY`/`EXTCODESIZE` on itself (`ADDRESS`) returns something different, right? (Except in the context of a contract creation frame)

jochem-brouwer (approved):

EIPS/eip-7702.md
shemnon: Yes, but it also answers the question "Am I being run as a delegate" - which is a useful security feature.

frangio: I don't think this is true. CODECOPY copies the code that is dynamically running whereas EXTCODECOPY on ADDRESS copies the account code, these can already differ prior to EIP-7702 due to DELEGATECALL or CALLCODE.

lightclient (approved):
sgtm

gumb0:
@lightclient Not sure what's the CI failure about

jochem-brouwer:
Can we retrigger? (Maybe this comment does it?)

jochem-brouwer (approved):

PR #9259, opened 2025-01-22
Add EIP: Flow Control Wallet Call Capability

SamWilsn:


EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+status: Draft
+type: Standards Track
+category: Interface
+created: 2025-01-17
+requires: 5792
+---
+
+## Abstract
+
+This proposal extends [EIP-5792](./eip-5792.md) to allow dapps to downgrade their required atomicity guarantees and control the behaviour after a failed/reverted call. It introduces the batch-scope concept of `strict` vs. `loose` atomicity, where a `strict` batch remains atomic in the face of chain reorgs and a `loose` batch does not; and the per-call ability to continue after a failed/reverted call (`continue`) or stop processing (`halt`.)
```
Marchhill, line 16:
```suggestion
This proposal extends [EIP-5792](./eip-5792.md) to allow dapps to downgrade their required atomicity guarantees and control the behaviour after a failed/reverted call. It introduces the batch-scope concept of `strict` vs. `loose` atomicity, where a `strict` batch remains atomic in the face of chain reorgs and a `loose` batch does not; and the per-call ability to continue after a failed/reverted call (`continue`) or stop processing (`halt`).
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+#### Rollback
+
+A rollback is informally defined as "causing no meaningful changes on chain." A
+rolled back batch only makes gas accounting and bookkeeping (eg. nonce)
+changes. In other words, a rollback is the default behaviour of EIP-5792 when
+a call fails.
+
+#### Critical Calls
+
+A critical call is a call the causes the entire batch to rollback on failure,
```
Marchhill, line 178:
```suggestion
A critical call is a call that causes the entire batch to rollback on failure,
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+
+### Concepts
+
+#### Call Failure
+
+<!-- TODO: is a "call failure" the same as a "revert"? -->
+
+#### Rollback
+
+A rollback is informally defined as "causing no meaningful changes on chain." A
```
Marchhill, line 171: Is there any reason to break up lines like this? A paragraph could be on one line and use line wrapping, not sure what the standard is here

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+by setting batch-scope `atomicity` to `strict`, `loose`, or `none` respectively.
+Strict may also be enabled by omitting `atomicity` entirely.
+
+Strict atomicity is simply naming the default behaviour of EIP-5792: calls
+within a single batch MUST be contiguous and applied atomically (or the batch
+rolled back.)
+
+Loose atomicity, on the other hand, is a weaker guarantee. In the event of a
+block reorg, any number of calls from the batch MAY appear on chain (possibly
+interspersed with other transactions.) If there are no block reorgs, loose
```
Marchhill, line 195:
```suggestion
interspersed with other transactions). If there are no block reorgs, loose
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+atomicity MUST provide the same guarantees as strict.
+
+The none level of atomicity only provides the guarantee that the calls appear on
+chain in the order they are in the batch. Any number of calls from the batch
+MAY appear on chain (possibly interspersed with other transactions.)
```
Marchhill, line 200:
```suggestion
MAY appear on chain (possibly interspersed with other transactions).
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+> * MUST NOT await for any calls to be finalized to complete the batch
+> * MUST submit multiple calls as an atomic unit in a single transaction
+> * MAY revert all calls if any call fails
+> * MUST not execute any further calls after a failed call
+> * MAY reject the request if one or more calls in the batch is expected to
+>   fail, when simulated sequentially
+
+##### Added Requirements
+
+The wallet:
```
Marchhill, line 233:
```suggestion
The following requirements are imposed on wallets:
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+
+
+#### `wallet_getCallsStatus`
+
+When `wallet_getCallsStatus` is called with a batch identifier corresponding to
+a batch submitted with the batch-scope `flow-control` capability enabled, the
+following changes override the behaviour defined in EIP-5792. Note that:
+
+* There are no changes when called with a batch without this capability enabled;
+  and
```
Marchhill, line 299:
```suggestion

```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+
+##### Added Requirements
+
+###### Capabilities
+
+The returned capabilities object:
+
+* MUST contain a `flow-control` key set to exactly the boolean `true`.
+    * It may be tempting to include additional detail about the status of
+      individual calls here, but don't. Instead use a multi-status capability
```
Marchhill, line 328: Wording isn't very formal, maybe something like 'additional details on the status of individual calls SHOULD NOT be included'

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+This proposal modifies some of the status codes for use with EIP-5792's
+`GetCallsResult.status` field, and introduces the following new codes:
+
+| Code   | Description           |
+|--------|-----------------------|
+| `102`  | Partially Executed    |
+| `207`  | Partial Success       |
+
+An "included" call, in this section, is defined as having either been
+successfully or unsuccessfully executed. A call that has been recorded on chain,
```
Marchhill, line 375:
```suggestion
executed, even if execution was unsuccessful. A call that has been recorded on chain,
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+###### `100` Pending
+
+Status `100` MUST NOT be returned if any calls in the batch have been included
+on chain.
+
+###### `102` Partially Executed
+
+Status `102` SHALL be returned only when all of the following are true:
+
+ * At least one call in the batch has been included on chain; and
```
Marchhill, line 393:
```suggestion
 * At least one call in the batch has been included on chain.
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+contain receipts for all transactions with calls that have been included.
+
+Note that a receipt capturing a failed call does not mean the call will
+ultimately fail. Wallets can resubmit calls (eg. with a higher gas limit), and
+the call may be executed successfully eventually.
+
+###### `200` Confirmed
+
+Status `200` MUST NOT be returned if any calls in the batch failed (including
+batch rollback, and the `on-failure` modes `halt`/`continue`.)
```
Marchhill, line 406:
```suggestion
batch rollback, and the `on-failure` modes `halt`/`continue`).
```

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+
+###### `207` Partial Success
+
+Status `207` SHALL be returned only when all of the following are true:
+
+* At least one call in the batch has been included and succeeded;
```
Marchhill, line 412:
```suggestion
* At least one call in the batch has been included and succeeded.
```
Same for rest of list, for consistency with other lists

EIPS/eip-draft_wallet-call-flow-control.md
```diff
@@ -0,0 +1,550 @@
...
+  The current placeholder is acceptable for a draft.
+
+  TODO: Remove this comment before submitting
+-->
+
+TBD
+
+## Backwards Compatibility
+
+<!--
```
Marchhill, line 522: can be removed

Marchhill (approved):
A few nitpicks

EIPS/eip-draft_wallet-call-flow-control.md
SamWilsn: I like a narrow line width, but we accept both wrapped and unwrapped lines.

Remaong89 (approved):

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### RPC Interface
+
+The following subsections are modifications to the API endpoints from EIP-5792.
+
+If a request does not match the schema defined below, the wallet MUST reject the request with an error code of 0 <!-- TODO -->.
```
g11tech, line 30: i would prefer to create a params/constants table and refer them by their param/constant name here. brings it out as well as confine the TODO to a particualr section.

similarly with other TODOS

g11tech:
small nit and linter/markup fixes required

lukasrosario:
any particular reason this isn't an ERC @SamWilsn ? other capabilities are ERCs and 5792 technically does specify capabilities should be ERCs

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+
+#### Atomicity Levels
+
+This proposal introduces three atomicity levels: strict, loose, and none; enabled
+by setting batch-scope `atomicity` to `strict`, `loose`, or `none` respectively.
+Strict may also be enabled by omitting `atomicity` entirely.
+
+Strict atomicity is simply naming the default behaviour of EIP-5792: calls
+within a single batch MUST be contiguous and applied atomically (or the batch
+rolled back.)
```
lukasrosario, line 191: this is effectively `atomicBatch` right?

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+requires: 5792
+---
+
+## Abstract
+
+This proposal extends [EIP-5792](./eip-5792.md) to allow dapps to downgrade their required atomicity guarantees and control the behaviour after a failed/reverted call. It introduces the batch-scope concept of `strict` vs. `loose` atomicity, where a `strict` batch remains atomic in the face of chain reorgs and a `loose` batch does not; and the per-call ability to continue after a failed/reverted call (`continue`) or stop processing (`halt`).
+
+## Motivation
+
+While the base EIP-5792 specification works extremely well for smart contract wallets, it does not allow the expression of the full range of flow control options that wallets can implement. For example, a dapp may only be submitting a batch for gas savings and not care about whether all calls are reverted on failure. A wallet may only be able to offer a limited form of atomicity through block builder backchannels, but that may be sufficient for a trading platform.
```
jakubuid, line 20: > a dapp may only be submitting a batch for gas savings and not care about whether all calls are reverted on failure
Why wouldn't dapp care? What's the user flow in this case?

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+    "from": "0xd46e8dd67c5d32be8058bb8eb970870f07244567",
+    "chainId": "0x01",
+    "calls": [
+      {
+        "to": "0xd46e8dd67c5d32be8058bb8eb970870f07244567",
+        "value": "0x182183",
+        "data": "0xfbadbaf01",
+        "capabilities": {
+            "flow-control": {
+                "on-failure": "continue"
```
jakubuid, line 111: what exactly is a wallet supposed to do in this case when the flow fails? what does continue mean? like to continue with the execution of the next calls? what if there's only one call in the array?

EIPS/eip-7867.md
SamWilsn: Yep!

EIPS/eip-7867.md
SamWilsn: Something like a token dispersement or several unrelated swaps. For example, I could be swapping several ERC-20 tokens into ETH. I just want as much of that ETH as possible, and I'll resubmit any failed swaps.

EIPS/eip-7867.md
SamWilsn: Hopefully that's defined well enough later in the document (and let me know if it isn't), but a failed call with `"on-failure": "continue"` is just slipped, and the next call is executed.

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+    reasons. This error code MUST NOT be used for other rejection reasons.
+* MUST reject (with error code 0<!-- TODO -->) batches containing at least one
+  critical call if the batch requests an atomicity level the wallet cannot
+  provide for any reason other than user rejection.
+  * Wallets supporting `strict` but not `loose` SHOULD NOT reject `loose`
+    batches and SHOULD instead upgrade the request to strict atomicity.
+  * Note that a batch with exactly one call _always_ satisfies the requirements
+    of strict atomicity.
+* MUST reject (with error code 0<!-- TODO -->) batches containing unsupported
+  `on-failure` modes.
```
jakubuid, line 282: We should rather set some default mode instead of forcing a wallet to reject such a batch. The reasoning is that on mobile devices MUST reject indicates that the wallet should reject such request immediately which is bad UX. Imagine a mobile wallet in the background that is triggered by a dapp aka it receives the batch request, a wallet is put to the foreground and immediately redirects back to a dapp without any information to the user. It should rather default to some mode and continue the flow.

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+The none level of atomicity only provides the guarantee that the calls appear on
+chain in the order they are in the batch. Any number of calls from the batch
+MAY appear on chain (possibly interspersed with other transactions).
+
+### Behaviour
+
+#### `wallet_sendCalls`
+
+The wallet MUST reject `wallet_sendCalls` requests with error code 0
+<!-- TODO --> where both:
```
jakubuid, line 207: As mentioned above, this creates a bad UX on mobile devices. We should rather default to one agreed behaviour to keep the user flow consistent instead of letting a wallet auto-reject such requests:
https://github.com/ethereum/EIPs/pull/9259/files#r1948151108

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+    * It may be tempting to include additional detail about the status of
+      individual calls here, but don't. Instead use a multi-status capability
+      defined elsewhere.
+
+###### Receipts
+
+The returned `receipts` array:
+
+* MUST NOT contain more than one receipt for the same transaction.
+* SHOULD NOT contain receipts for transactions without a call from the requested
```
jakubuid, line 338: Why not MUST NOT? Is it allowed to add a receipt for a call that was not icluded in the batch? If yes, why, and what is the use case?

EIPS/eip-7867.md
SamWilsn: > We should rather set some default mode instead of forcing a wallet to reject such a batch.

This is unacceptable behaviour. If the dapp requests a mode, it _has_ to get that mode. Imagine asking for "rollback" and getting "continue". That could be catastrophic.

> Imagine a mobile wallet in the background that is triggered by a dapp aka it receives the batch request, a wallet is put to the foreground

Does the wallet need to be in the foreground to reject the request? If so, how does the UX flow work for unsupported RPC endpoints, like for example `eth_signTransaction` on MetaMask? This doesn't seem unique to just this proposal.

> immediately redirects back to a dapp without any information to the user

The standard doesn't say anything about not displaying information to the user. The wallet is free to display a message to the user before redirecting back to the dapp.

EIPS/eip-7867.md
SamWilsn: I went with "MUST NOT" for things that would break/be unsafe, and SHOULD NOT for things that would be weird but not dangerous.

EIPS/eip-7867.md
jakubuid: I meant more like the scenario where the dapp requests the mode, but the wallet doesn't support that mode.  The proposal suggests the rejection, which is a bad UX on mobile, hence reasoning for some default behaviour that mitigates that.

Yes, on mobile, a wallet must be foregrounded to receive the request. So, in the MetaMask case (and not only), they support the CAIP-25 handshake, which defines the methods, chains, and events supported by a wallet, so dapp knows before sending the request if a given method is supported by a wallet or not. Hence, there's no problem with automatic rejection on mobile.

Now, when I'm writing this, I realized that the wallet could use the CAIP-25 handshake (sessionProperties) to define what modes are supported, so the dapp could only request those that the wallet supports.

jakubuid: @SamWilsn 
I think it might be useful to add a section explaining how wallets can define flow-control capabilities via CAIP-25 that would mitigate the above. Something similar to the section in [7811](https://github.com/ethereum/ERCs/pull/709/files#diff-be675f3ce6b6aa5616dd1bccf5e50f44ad65775afb967a47aaffb8f5eb51b849R203)

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+```
+
+##### Example Response
+
+```js
+{
+    "0x1": {
+        "flow-control": {
+            "loose": ["halt", "continue"],
+            "strict": ["continue"]
```
jakubuid, line 157: Shouldn't this define `on-failure` flow control for `call-scope` and `atomicity` flow control for `batch-scope`?

Not sure how to interpret this part:
```
 "loose": ["halt", "continue"],
 "strict": ["continue"]
```

EIPS/eip-7867.md
```diff
@@ -0,0 +1,553 @@
...
+When `flow-control` is present in the batch-scope `capabilities`, the following
+changes override the behaviour specified in EIP-5792.
+
+##### Removed Requirements
+
+These requirements defined in EIP-5792 are removed:
+
+> The wallet:
+>
+> * MUST NOT await for any calls to be finalized to complete the batch
```
g11tech, line 224: some lint fixes required here

EIPS/eip-7867.md
SamWilsn: This snippet is saying the wallet supports `on-failure` modes of `halt` and `continue` when using `loose` atomicity, but only `continue` when using `strict`.

EIPS/eip-7867.md
```diff
@@ -0,0 +1,565 @@
...
+* MUST NOT rollback the batch if zero critical calls (as defined above) fail.
+  * In other words, if the only failures are non-critical, the successful calls
+    have to appear on chain.
+* MUST NOT execute a call (or ever allow a call to be executed) more than once.
+
+###### Flow Control
+
+* MUST treat a missing call-scope `flowControl` capability as equivalent to
+  setting `onFailure` to `rollback`.
+* MUST treat a missing call-scope `onFailure` mode as equivalent to `rollback`.
```
azf20, line 259: I think this is another argument for request-level `onFailure` settings - as an example I don't think `loose` and `none` are necessarily capable of `rollback`?

EIPS/eip-7867.md
```diff
@@ -0,0 +1,565 @@
...
+  [EIP-7702](./eip-7702.md) set code transaction.)
+  * Note that this only applies to user rejections specifically because of
+    atomicity. It does not change the behaviour for batches rejected for other
+    reasons. This error code MUST NOT be used for other rejection reasons.
+* MUST reject (with error code `UNSUPPORTED_LEVEL`) batches containing at least
+  one critical call if the batch requests an atomicity level the wallet cannot
+  provide for any reason other than user rejection.
+  * Wallets supporting `strict` but not `loose` SHOULD NOT reject `loose`
+    batches and SHOULD instead upgrade the request to strict atomicity.
+  * Note that a batch with exactly one call _always_ satisfies the requirements
```
azf20, line 279: For such calls any specification of flow control capabilities is effectively ignored, as it's necessarily `strict / rollback`?

EIPS/eip-7867.md
```diff
@@ -0,0 +1,565 @@
...
+###### Receipts
+
+The returned `receipts` array:
+
+* MUST NOT contain more than one receipt for the same transaction.
+* SHOULD NOT contain receipts for transactions without a call from the requested
+  batch.
+* MUST contain exactly one receipt capturing each successful call.
+    * Multiple calls MAY be captured in one receipt, but the successful
+      execution of one call MUST NOT be captured by multiple receipts.
```
azf20, line 342: is it even possible for a successful call to be captured by multiple receipts?

EIPS/eip-7867.md
```diff
@@ -0,0 +1,565 @@
...
+    * Given two calls (_A_ and _B_) in a batch, the following are non-exhaustive
+      example combinations of calls-per-receipt. Each `(...)` is a receipt from
+      a single transaction.
+        * Valid Examples:
+            * `[(successful A, successful B)]`
+            * `[(successful A), (successful B)]`
+            * `[(successful A, unsuccessful B), (successful B)]`
+            * `[(unsuccessful A), (successful A), (successful B)]`
```
azf20, line 350: how are unsuccessful calls represented? obviously the receipt as a whole has a status, but you can have failed calls in a successful receipt (requiring tracing to track down what actually happened) - and I don't see where in the [call result](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md?plain=1#L164) it's identified one way or the other

EIPS/eip-7867.md
```diff
@@ -0,0 +1,565 @@
...
+Unlike a plain EOA, a shielded mempool can provide additional guarantees about
+transaction atomicity. In this example, the wallet does not offer the
+`onFailure` mode of `halt` when using `none` atomicity, but does when using
+`loose`.
+
+```js
+{
+    "0x1": {
+        "flowControl": {
+            "none": [ "rollback", "continue" ]
```
azf20, line 488: how do you rollback in `none`?

azf20 (commented):
A couple of comments / questions.

EIPS/eip-7867.md
```diff
@@ -0,0 +1,565 @@
...
+<!-- TODO: is a "call failure" the same as a "revert"? -->
+
+#### Rollback
+
+A rollback is informally defined as "causing no meaningful changes on chain." A
+rolled back batch only makes gas accounting and bookkeeping (eg. nonce)
+changes. In other words, a rollback is the default behaviour of EIP-5792 when
+a call fails.
+
+#### Critical Calls
```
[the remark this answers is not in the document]
SamWilsn: I'm defining critical here so I can use:

- "critical call" as short-form for call with `"onFailure": "rollback"` and;
-  "non-critical call" as short-form for the other modes.

EIPS/eip-7867.md
SamWilsn: I'm not sure I find those more clear? What is a bundle? They also seem tied to particular implementations. I can imagine a multisig where each signature is a single transaction and only the last one executes the calls. You'd have multiple transactions, applied atomically.

EIPS/eip-7867.md
cryptotavares: I believe the options should be `halt` and `continue`. Right @SamWilsn?

EIPS/eip-7867.md
azf20: I think in that example I would only expect the last transaction to show up in the result (the signature transactions aren't really relevant to the app's request).

Fair that it's still not super clear. I guess I prefer these as it's more explicit about how the calls take place, i.e.
transaction = all in one transaction
bundle = in a bundle of transactions ([Flashbots bundles as a reference](https://docs.flashbots.net/flashbots-auction/advanced/understanding-bundles)), implying that transactions will be executed as a group, unless there's a reorg
none = no constraints

g11tech (approved):

PR #9267, opened 2025-01-25
Update EIP-7805: Try fixing HTML rendering of bullets

jsign:
I was reading [this section from the rendered version](https://eips.ethereum.org/EIPS/eip-7805#execution-layer) and there' seems to be a glitch:
image

It looks weird that point is 1. again.

But if I look at the [rendered Markdown](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7805.md), it looks fine:
image

I saw the Markdown linter suggest fixing some stuff that _maybe_ explains the problem?
I can't know if this fixes it until the full HTML is rendered (since the markdown preview was already showing the expected).

Maybe there's a way to render the EIP locally and check? or just merge the PR and see if it's fixed.

### DIFF
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -95,11 +95,11 @@ For each transaction `T` in ILs, perform the following:
 
 1. Check whether `T` is present in `B`. If `T` is present, then jump to the next transaction, else continue with next step.
 
-2. Validate `T` against `S`. 
+2. Validate `T` against `S`.
 
-  - If `T` is invalid, then continue to the next transaction. 
-  
-  - If `T` is valid, terminate process and assert block `B` as invalid.
+   - If `T` is invalid, then continue to the next transaction.
+
+   - If `T` is valid, terminate process and assert block `B` as invalid.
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7805.md
```diff
@@ -95,11 +95,11 @@ For each transaction `T` in ILs, perform the following:
...
+2. Validate `T` against `S`.
 
-  - If `T` is invalid, then continue to the next transaction. 
-  
-  - If `T` is valid, terminate process and assert block `B` as invalid.
+   - If `T` is invalid, then continue to the next transaction.
+
+   - If `T` is valid, terminate process and assert block `B` as invalid.
 
 3. Execute `T` on state `S`. Assert that the execution of `T` fails.
```
jsign, line 104: BTW, under which 2. sub-bullet do we end up in step 3.?

Shouldn't the second sub-bullet say moving to the next step (i.e. 3.) and do the assertion?

SamWilsn:
I'm _very_ quickly going through typo fix PRs. Unfortunately this PR has a conflict, so I'm going to close it, but feel free to create a new one if the issue still exists.

PR #9283, opened 2025-01-28
Update EIP-7823: Use decimal numbers for input length.

rodiazet:
This PR fixes formatting of inputs length values.


EIPS/eip-7823.md
```diff
@@ -57,44 +57,44 @@ Since MODEXP was introduced in the Byzantium hard fork, an analysis has been con
 
 | `input_of_BASE` | count |
```
jochem-brouwer, line 58: Are these bits or bytes? `input_of_BASE`

rodiazet: Good catch. Thank you. https://github.com/ethereum/EIPs/pull/9284

PR #9286, opened 2025-01-29
Update EIP-7686: Clarify the size of `request_type`

gumb0:
I think it's unfortunate, that the size of `request_type` is never specified.


gumb0:
@lightclient please take a look

mkalinin (approved):

lightclient (approved):

PR #9293, opened 2025-01-30
Update EIP-7748: Fix code-chunking conversion unit counting

jsign:
This PR fixes a mistake in the Python code describing how code-chunking counts towards the CONVERSION_STRIDE limit.


EIPS/eip-7748.md
```diff
@@ -206,7 +205,6 @@ def state_convert(state: State, stride: int):
             
             for chunk_num in range(len(chunked_code)):
                 state_set_codechunk(state, address, chunk_num, chunked_code[chunk_num])
-                n += 1
```
jsign, line 209: See the `Account code chunking done in a single step` rationale section.

g11tech: makes sense lgtm

gballet (approved):
LGTM. In theory, we would still want to have a granularity of one verkle tree leaf, but this would make it more complex to handle, which is why we never did it in the conversion code in geth. It's a good thing to do the same thing in the spec.

g11tech (approved):

PR #9315, opened 2025-02-03
Add EIP: Enhanced RETURN opcodes

jhweintraub:
EIP to replace `RETURN` with 4 more specific `RETURN` opcodes which allow defining where to return from explicitly instead of first requiring copying to memory.

### DIFF
--- a/EIPS/eip-7877.md
+++ b/EIPS/eip-7877.md
@@ -13,22 +13,22 @@ requires: 6, 1153, 3855
 
 ## Abstract
 
-This EIP specifies a series of new return opcodes which allow the user to specify which data location to return from
+This EIP specifies a series of new `RETURN` opcodes which allow the user to specify which data location to return from
 instead of defaulting to returning from memory.
 
 ## Motivation
 
-With the introduction of transient storage, many smart contracts have begun to store data using the new transient opcodes to optimize for gas usage, whereby a callback 
-involves returning the data previously stored transiently. However, tthe current `RETURN` opcode only allows for returning sequential bytes in memory. This requires
-developers to incur additional gas overhead by manually writing data from transient storage to memory before returning, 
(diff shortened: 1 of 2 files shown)


EIPS/eip-7877.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Abstract
+
+This EIP specifies a series of new return opcodes which allow the user to specify which data location to return from
+instead of defaulting to returning from memory.
+
+## Motivation
+
+With the introduction of transient storage, many smart contracts have begun to store data using the new transient opcodes to optimize for gas usage, whereby a callback 
+involves returning the data previously stored transiently. However, tthe current `RETURN` opcode only allows for returning sequential bytes in memory. This requires
```
cll-gg, line 22:
```suggestion
involves returning the data previously stored transiently. However, the current `RETURN` opcode only allows for returning sequential bytes in memory. This requires
```

EIPS/eip-7877.md
```diff
@@ -0,0 +1,90 @@
...
+
+There is precedent for these changes.
+
+EIP-3855: Introducing `PUSH0` opcode to optimize away alternative methods of pushing 0 onto the stack.
+
+EIP-6: Renaming `SUICIDE` to `SELFDESTRUCT` without changing functionality.
+
+## Backwards Compatibility
+
+There are no backwards compatibility concerns, as `MRETURN` will utilize the same gas cost and opcode as `RETURN` does now. Due to EOF, it is suggested that these changes be
```
g11tech, line 81: do you intend to make this EIP `require` EOF parent EIP (that lists all EOF eips)? if so then add it in require section

jhweintraub: no because EOF is not strictly requried to implement but I assumed that due to the complexity of EOF that any opcode changes would be considered until after that implementation is finished.

PR #9292, opened 2025-01-30
Update EIP-7607: Proposing 7805 for fusaka

soispoke:
**ATTENTION: ERC-RELATED PULL REQUESTS NOW OCCUR IN [ETHEREUM/ERCS](https://github.com/ethereum/ercs)**

--

When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #9320, opened 2025-02-04
proposes EIP-7805 (FOCIL) to be included in the Fusaka hardfork

jflo:
The upcoming hardfork (Fulu / Osaka, or Fusaka) should include censorship resistance in the form of FOCIL.

The EIP is already implemented well enough by clients that it has its own devnet. While edge cases are still being discovered, the specs seem stable, and implementation is well understood by multiple client teams.


jflo:
Closing since it duplicates #9292

PR #9323, opened 2025-02-06
Move EIP-7623 to Last Call

nerolation:


gumb0:
I have one nitpicky suggestion to clarify the spec https://github.com/ethereum/EIPs/pull/9350

nerolation:
I will add this clarification, since we're not changing the value, which becomes clearer that way:
https://github.com/ethereum/EIPs/compare/master...nerolation:EIPs:7623_add_reference

gumb0:
> I will add this clarification, since we're not changing the value, which becomes clearer that way: [master...nerolation:EIPs:7623_add_reference](https://github.com/ethereum/EIPs/compare/master...nerolation:EIPs:7623_add_reference)
> 
> @gumb0, lmk what you think!

I'd still mention the value for the reference, like "Let `INITCODE_WORD_COST` be `2` as defined in [EIP-3860](./eip-3860.md)."

Note also how in some EIPs we define previously known constants in the table https://eips.ethereum.org/EIPS/eip-7620#parameters

nerolation:
I will add the value, good point yeah! Thanks

nerolation:
Done here:
https://github.com/ethereum/EIPs/pull/9351#event-16247846559

PR #9376, opened 2025-02-18
Add EIP: Delayed execution

nerolation:
This PR proposes delayed execution as described in this ethresearch post:
https://ethresear.ch/t/delayed-execution-and-skipped-transactions/21677

Find Specs here:
https://github.com/ethereum/execution-specs/compare/master...fradamt:execution-specs:delayed-exec-basic

### DIFF
--- a/EIPS/eip-7886.md
+++ b/EIPS/eip-7886.md
@@ -3,13 +3,12 @@ eip: 7886
 title: Delayed execution
 description: Make blocks statically verifiable by charging coinbase for inclusion costs upfront
 author: Francesco D`Amato (@fradamt), Toni Wahrstätter (@nerolation)
-discussions-to: "https://ethereum-magicians.org/t/eip-xxxx-delayed-execution/22890"
+discussions-to: https://ethereum-magicians.org/t/eip-7886-delayed-execution/22890
 status: Draft
 type: Standards Track
 category: Core
-created: "2025-02-18"
+created: 2025-02-18
 ---
-
 
 ## Abstract
 
@@ -17,14 +16,9 @@ This proposal makes (execution) blocks statically verifiable through minimal che
 
 ## Motivation
 
-A key advantage of the skipping mechanism is that it enables **asynchronous block validation**: a block can be considered valid even if it contains invalid transactions, because those transactions end up skipping execution without invalidating the entire block. 
+A key advantage of this proposal is that it enables **asynchronous block validation**. Currently, blocks must be fully executed before validators can attest to their validity. This requirement creates a bottleneck in the consensus process, as attestors must wait for execution results before committing their votes.
 
-## Overview
-
-Enabling delayed execution by making the block's validity statically verifiable requires two things:
-
-1. **Deferred execution outputs**: all header fields that commit to execution outputs are deferred by one slot. For example `state_root` and `receipt_root` become `pre_state_root` (similarly to [EIP-7862](./eip-7862.md)) and `parent_receipt_root`, the root of the state and receipt trie obtained after executing the block's parent. The same applies to the General Purpose Execution Layer Requests from [EIP-7685](./eip-7685.md): the requests are deferred by one slot and the requests in the CL must correspond to the `parent_requests_hash` in the EL header. However, this alone would only defer the computation of these execution outputs (mainly of the state root) rather than the actual execution, because verifying transaction validity would still require executing. 
-2. **Upfront payment of inclusion costs by `COINBASE`**: in addition, we need to be able to skip (no-op) invalid transactions without invalidating the whole block. Right now, this is not possible because of the free-DA problem: as soon as we include a transaction into a block, it must pay for its data footprint. By charging the inclusion cost of all transactions upfront from the block's `COINBASE`, it is possible to skip transactions that are found to be invalid at execution time, because the protocol has already been compensated for the inclusion.
+By introducing a mechanism where invalid transactions are skipped rather than invalidating the entire block, execution is no longer an immediate requirement for validation. Instead, a block’s validity can be determined based on its structural correctness and the upfront payment of inclusion costs by the `COINBASE`. This allows attestation to happen earlier, independent of execution, potentially enabling higher block gas limits.
 
 ## Specification
 
@@ -69,7 +63,6 @@ class Header:
     s: U256
 ```
 
-
 The following function is used to recover the signer’s address, intended to be `COINBASE`, from the signed block header:
 
 ```python
@@ -108,7 +101,6 @@ def recover_header_signer(
     )
 ```
 
-By signing over the header, the `COINBASE` address explicitly accepts responsibility for the upfront inclusion costs *of this block*. Therefore, the recovered address MUST equal the block's `COINBASE`. The `COINBASE`'s commitment is protected from replay attacks, because the header is a commitment to the block, so the signature only serves as an authorization for the exact block for which the `COINBASE` has agreed to take responsibility.
 
 ```python
 header_signer = recover_header_signer(
@@ -145,7 +137,7 @@ def check_transaction_static(
     chain_id: U64,
 ) -> Address:
     
-    ...   
+    ... 
     
     if isinstance(tx, (FeeMarketTransaction, BlobTransaction, SetCodeTransaction)):
     if tx.max_fee_per_gas < tx.max_priority_fee_per_gas:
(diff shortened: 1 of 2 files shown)


EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+---
+
+
+## Abstract
+
+This proposal makes (execution) blocks statically verifiable through minimal checks that only require the previous state, but no execution of the block's transactions, allowing validators attest to a block's validity without completing its execution. We allow transactions to be skipped when invalid at execution time, without invalidating the whole block. To ensure that even skipped transactions pay for their resources, the `COINBASE` pays for all inclusion costs upfront (base cost, calldata and blobs), and recovers the costs only when transactions are successfully executed.
+
+## Motivation
+
+A key advantage of the skipping mechanism is that it enables **asynchronous block validation**: a block can be considered valid even if it contains invalid transactions, because those transactions end up skipping execution without invalidating the entire block.
```
SamWilsn, line 20: You should probably answer the question: why is asynchronous block validation a desirable property?

EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+2. **Upfront payment of inclusion costs by `COINBASE`**: in addition, we need to be able to skip (no-op) invalid transactions without invalidating the whole block. Right now, this is not possible because of the free-DA problem: as soon as we include a transaction into a block, it must pay for its data footprint. By charging the inclusion cost of all transactions upfront from the block's `COINBASE`, it is possible to skip transactions that are found to be invalid at execution time, because the protocol has already been compensated for the inclusion.
+
+## Specification
```
SamWilsn, line 29:
```suggestion
## Specification

### Overview

Enabling delayed execution by making the block's validity statically verifiable requires two things:

1. **Deferred execution outputs**: all header fields that commit to execution outputs are deferred by one slot. For example `state_root` and `receipt_root` become `pre_state_root` (similarly to [EIP-7862](./eip-7862.md)) and `parent_receipt_root`, the root of the state and receipt trie obtained after executing the block's parent. The same applies to the General Purpose Execution Layer Requests from [EIP-7685](./eip-7685.md): the requests are deferred by one slot and the requests in the CL must correspond to the `parent_requests_hash` in the EL header. However, this alone would only defer the computation of these execution outputs (mainly of the state root) rather than the actual execution, because verifying transaction validity would still require executing. 
2. **Upfront payment of inclusion costs by `COINBASE`**: in addition, we need to be able to skip (no-op) invalid transactions without invalidating the whole block. Right now, this is not possible because of the free-DA problem: as soon as we include a transaction into a block, it must pay for its data footprint. By charging the inclusion cost of all transactions upfront from the block's `COINBASE`, it is possible to skip transactions that are found to be invalid at execution time, because the protocol has already been compensated for the inclusion.
```

EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+2. **Upfront payment of inclusion costs by `COINBASE`**: in addition, we need to be able to skip (no-op) invalid transactions without invalidating the whole block. Right now, this is not possible because of the free-DA problem: as soon as we include a transaction into a block, it must pay for its data footprint. By charging the inclusion cost of all transactions upfront from the block's `COINBASE`, it is possible to skip transactions that are found to be invalid at execution time, because the protocol has already been compensated for the inclusion.
```
SamWilsn, line 27: This content seems like it belongs more in the Rationale section because it's explaining _why_ certain choices were made within this document.

EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+
+## Specification
+
+### Change to the header structure
+
+#### Deferred fields 
+
+In order to be verifiable before execution, the header cannot commit to any execution output. In particular, we need to defer these fields: `state_root, receipt_root, bloom, gas_used, requests_hash`. We replace them with the equivalent execution outputs from the parent block.
```
SamWilsn, line 35: > In order to be verifiable before execution, the header cannot commit to any execution output.

A bit of rationale in the specification section is fine (this is at the limit) but try to keep the Specification section purely about the changes that need to be made and not about the reasoning behind the changes.

EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+
+### Change to the header structure
+
+#### Deferred fields 
+
+In order to be verifiable before execution, the header cannot commit to any execution output. In particular, we need to defer these fields: `state_root, receipt_root, bloom, gas_used, requests_hash`. We replace them with the equivalent execution outputs from the parent block. 
+
+#### Coinbase signature over the header
+
+We include a signature from `COINBASE` over the rest of the header in the header , so that the `COINBASE` address can authorize the upfront payment of inclusion costs.
```
SamWilsn, line 39: Similarly, this is better placed in the Rationale section.

EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+                header.blob_gas_used,
+                header.excess_blob_gas,
+                header.parent_beacon_block_root,
+                header.parent_requests_hash,
+            )
+        )
+    )
+```
+
+By signing over the header, the `COINBASE` address explicitly accepts responsibility for the upfront inclusion costs *of this block*. Therefore, the recovered address MUST equal the block's `COINBASE`. The `COINBASE`'s commitment is protected from replay attacks, because the header is a commitment to the block, so the signature only serves as an authorization for the exact block for which the `COINBASE` has agreed to take responsibility.
```
SamWilsn, line 111: Little bit more rationale in here. I'm going to stop commenting on this; you get the idea.

EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+if coinbase != header_signer:
+    raise InvalidBlock
+```
+
+If `COINBASE != header_signer`, the block MUST be considered invalid.
+
+
+### Static block validation
+
+We split up a block's validation from its execution. In the ethereum/execution-specs, static validation is done in `validate_block`, after which a block is guaranteed to be valid and can be attested to, while execution remains within `apply_body`. In `validate_block`, we do some formal checks, as well as:
```
SamWilsn, line 127: > In the ethereum/execution-specs, static validation is done in `validate_block` [...]

You should link explicitly to lines in a specific commit, in case things move around or are refactored in EELS.

EIPS/eip-7886.md
```diff
@@ -0,0 +1,400 @@
...
+        raise InvalidBlock
+    if block.header.blob_gas_used != blob_gas_used:
+        raise InvalidBlock
+    
+    return sender_addresses
+```
+
+### Block execution
+
+This logic is implemented into the ethereum/execution-specs, in `apply_body`.
```
SamWilsn, line 246: Again, specific link is preferred.

EIPS/eip-7886.md
```diff
@@ -0,0 +1,399 @@
...
+In order to be verifiable before execution, the header cannot commit to any execution output. In particular, we need to defer these fields: `state_root, receipt_root, bloom, gas_used, requests_hash`. We replace them with the equivalent execution outputs from the parent block. 
+
+#### Coinbase signature over the header
+
+We include a signature from `COINBASE` over the rest of the header in the header , so that the `COINBASE` address can authorize the upfront payment of inclusion costs.
+
+The final header structure then is: 
+
+```python
+class Header:
```
rkrasiuk, line 38: `Deferred` comment on the field is a little bit confusing given that these `parent_*` fields are validated immediately

EIPS/eip-7886.md
```diff
@@ -0,0 +1,399 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-7886-delayed-execution/22890
+status: Draft
+type: Standards Track
+category: Core
+created: 2025-02-18
+---
+
+## Abstract
+
+This proposal makes (execution) blocks statically verifiable through minimal checks that only require the previous state, but no execution of the block's transactions, allowing validators attest to a block's validity without completing its execution. We allow transactions to be skipped when invalid at execution time, without invalidating the whole block. To ensure that even skipped transactions pay for their resources, the `COINBASE` pays for all inclusion costs upfront (base cost, calldata and blobs), and recovers the costs only when transactions are successfully executed.
```
rkrasiuk, line 15: one unmentioned consideration here is that this will require EL clients to bring back or introduce signer management for local payload building

rkrasiuk (commented):
very cool. would also be great to see in the consideration section what potential challenges there are with increasing the delay window (coinbase balance decrease and gas limit are two that immediately come to mind)

g11tech (approved):

PR #9378, opened 2025-02-18
Update EIP-7594: include cell proofs in network wrapper of blob txs

fradamt:
See [this doc](https://hackmd.io/@jimmygchen/HkUpFliYJx) for context

### DIFF
--- a/EIPS/eip-7594.md
+++ b/EIPS/eip-7594.md
@@ -47,7 +47,7 @@ cell_proofs = [cell_proof_0, cell_proof_1, ...]
 
 The `tx_payload_body`, `blobs` and `commitments` are as in EIP-4844, while the `proofs` field is replaced by `cell_proofs`, and a `wrapper_version` is added. These are defined as follows:
 
-- `wrapper_version` - one byte indicating which version of the wrapper is used. For the current one, it is set to `1`.  If `wrapper_version` is 0 the transaction must be decoded as a EIP-4844 transaction
+- `wrapper_version` - one byte indicating which version of the wrapper is used. For the current one, it is set to `1`.
 - `cell_proofs` - list of cell proofs for all `blobs`, including the proofs for the extension indices, for a total of `CELLS_PER_EXT_BLOB` proofs per blob (`CELLS_PER_EXT_BLOB` is the number of cells for an extended blob, defined [in the consensus specs](https://github.com/ethereum/consensus-specs/tree/9d377fd53d029536e57cfda1a4d2c700c59f86bf/specs/fulu/polynomial-commitments-sampling.md#cells)) 
 
 Note that, while `cell_proofs` contain the proofs for all cells, including the extension cells, the blobs themselves are sent without being extended (`CELLS_PER_EXT_BLOB / 2` cells per blob). This is to avoid sending redundant data, which can quickly be computed by the receiving node.
--- a/EIPS/eip-7594.md
+++ b/EIPS/eip-7594.md
@@ -29,9 +29,36 @@ Nodes find and maintain a diverse peer set and sample columns from their peers t
 
 A node can reconstruct the entire data matrix if it acquires at least 50% of all the columns. If a node has less than 50%, it can request the necessary columns from the peer nodes.
 
-The detailed specifications are on [ethereum/consensus-specs](https://github.com/ethereum/consensus-specs/blob/b4188829b32139916127827c64ba17c923e66c3c/specs/_features/eip7594/das-core.md).
+The detailed specifications are on [ethereum/consensus-specs](https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu).
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Networking
+
+This EIP introduces cell KZG proofs, which are used to prove that a KZG commitment opens to a cell at the given index. This allows downloading only specific cells from a blob, while still ensuring data integrity with respect to the corresponding KZG commitment, and is therefore a key component of data availability sampling. However, computing the cell proofs for a blob is an expensive operation, which a block producer would have to repeat for many blobs. Since proof verification is much cheaper than proof computation, and the proof size is negligible compared to cell size, we instead require blob transaction senders to compute the proofs themselves and include them in the network wrapper for blob transactions.
+
+To this end, during transaction gossip responses (`PooledTransactions`), the EIP-4844 wrapper of the EIP-2718 `TransactionPayload` of the blob transaction is modified to:
+
+```
+rlp([tx_payload_body, blobs, commitments, cell_proofs])
+
+cell_proofs = [cell_proof_0, cell_proof_1, ...]
+```
+
+The `tx_payload_body`, `blobs` and `commitments` are as in EIP-4844, while the `proofs` field is replaced by `cell_proofs`, defined as follows:
+- `cell_proofs` - lists of cell proofs for all `blobs`, including the proofs for the extension indices, for a total of `CELLS_PER_EXT_BLOB` (the number of cells for an extended blob, defined [in the consensus specs](https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/polynomial-commitments-sampling.md#cells)) proofs per blob
+
+Note that, while `cell_proofs` contain the proofs for all cells, including the extension cells, the blobs themselves are sent without being extended. This is to avoid sending redundant data, which can quickly be computed by the receiving node.
+In other words, `cell_proofs[i * CELLS_PER_EXT_BLOB + j]` is the proof for cell `j` of `compute_cells(blobs[i])`, where `compute_cells(blob)` outputs all cells of `blob`, including the extension cells.
+
+The node MUST validate `tx_payload_body` and verify the wrapped data against it. To do so, ensure that:
+
+- There are an equal number of `tx_payload_body.blob_versioned_hashes`, `blobs` and `commitments`.
+- `cell_proofs` contains exactly `CELLS_PER_EXT_BLOB * len(blobs)` cell proofs
+- The KZG `commitments` hash to the versioned hashes, i.e. `kzg_to_versioned_hash(commitments[i]) == tx_payload_body.blob_versioned_hashes[i]`
+- The KZG `commitments` match the corresponding `blobs` and `cell_proofs`. This requires computing the extension cells for all `blobs`, and verifying all `cell_proofs`. (Note: all cell proofs can be batch verified at once)
+
+
 
 ## Rationale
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7594.md
```diff
@@ -29,10 +29,38 @@ Nodes find and maintain a diverse peer set and sample columns from their peers t
...
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
+### Networking
+
+This EIP introduces cell KZG proofs, which are used to prove that a KZG commitment opens to a cell at the given index. This allows downloading only specific cells from a blob, while still ensuring data integrity with respect to the corresponding KZG commitment, and is therefore a key component of data availability sampling. However, computing the cell proofs for a blob is an expensive operation, which a block producer would have to repeat for many blobs. Since proof verification is much cheaper than proof computation, and the proof size is negligible compared to cell size, we instead require blob transaction senders to compute the proofs themselves and include them in the network wrapper for blob transactions.
+
+To this end, during transaction gossip responses (`PooledTransactions`), the EIP-4844 wrapper of the EIP-2718 `TransactionPayload` of the blob transaction is modified to:
```
daniellehrner, line 40: Reviewing Besu's code we currently have no way of changing the definition of a transaction type at a hardfork block. It would be much easier to introduce a new transaction type and make the current 4844 transactions invalid. That change would be straightforward in our case, it would be way less error prone and faster to be implemented.

dapplion: The transaction RLP is the same. Does your TX definition include what verification function to use?

EIPS/eip-7594.md
```diff
@@ -29,10 +29,38 @@ Nodes find and maintain a diverse peer set and sample columns from their peers t
...
+The detailed specifications are on [ethereum/consensus-specs](https://github.com/ethereum/consensus-specs/tree/9d377fd53d029536e57cfda1a4d2c700c59f86bf/specs/fulu/).
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
+### Networking
+
+This EIP introduces cell KZG proofs, which are used to prove that a KZG commitment opens to a cell at the given index. This allows downloading only specific cells from a blob, while still ensuring data integrity with respect to the corresponding KZG commitment, and is therefore a key component of data availability sampling. However, computing the cell proofs for a blob is an expensive operation, which a block producer would have to repeat for many blobs. Since proof verification is much cheaper than proof computation, and the proof size is negligible compared to cell size, we instead require blob transaction senders to compute the proofs themselves and include them in the network wrapper for blob transactions.
```
jimmygchen, line 38: There was a suggestion from Lion and also brought up by Nethermind during the PeerDAS breakout to have the _broadcasting_ EL node to translate the eip4844 blob transaction into the new network form (with cell proofs), i.e. compute cell proofs after receiving blob transaction via RPC, and propagate the updated transaction to p2p network - it seems reasonable to me and would have no impact to end users, even if we change the cell size in the future.

EIPS/eip-7594.md
jflo: Yes, but it is not protocol-schedule aware. We don't currently have a means of changing that rule during and upgrade.

jimmygchen: Just to add that L2 teams that's given input so far are happy with making changes on their side. 

I think @marcindsobczak also had some input on this on the last breakout call, any thoughts on the above comment?

Mini198962: Mina198934@gmail.com

EIPS/eip-7594.md
Mini198962: > There was a suggestion from Lion and also brought up by Nethermind during the PeerDAS breakout to have the _broadcasting_ EL node translate the eip4844 blob transaction into the new network form (with cell proofs), i.e. compute cell proofs after receiving blob transaction via RPC, and propagate the updated transaction to p2p network - it seems reasonable to me and would have no impact to end users, even if we change the cell size in the future.

EIPS/eip-7594.md
marcindsobczak: From what I understand the only change will be in proofs - there will be `n * m` instead of `m` (`m - number of blobs, n - number of cells in blob`), so structure will be the same. Right now we are validating if number of blobs is equal to number of proofs, we will need to change it and accept txs in two forms - old and new. So validate `number of proofs == n*m || m`

What I think is the most reasonable here:
1. In all ELs p2p start accepting new form of blob txs without dropping support for old format - just accept both. Keep responding to `GetPooledTransactions` with old format at this point - just be able to handle a new format. Check readiness in test environment.
2. Prepare internally for changes required by new format, like blob pool adjustments and "translating" rpc txs with old format to the new one. Probably we need to keep both forms of proofs in blob pools at this point.
3. After the fork start sending txs by p2p in a new format - all peers will be ready as hard fork requires clients version update and all clients were already ready to accept it. In p2p only change will be in `PooledTransactions` messages, so should be easy to switch at fork
4. Some time after the fork drop p2p support for old format and remove all stale stuff from blob pools etc.

I haven't started implementing it yet, so all above is just a theory at this point. Planning to start it next week.

Another ideas:
- new tx type
- new eth protocol version with override of `PooledTransactions`

EIPS/eip-7594.md
```diff
@@ -29,10 +29,38 @@ Nodes find and maintain a diverse peer set and sample columns from their peers t
...
+
+To this end, during transaction gossip responses (`PooledTransactions`), the EIP-4844 wrapper of the EIP-2718 `TransactionPayload` of the blob transaction is modified to:
+
+```
+rlp([tx_payload_body, blobs, commitments, cell_proofs])
```
g11tech, line 43:
```suggestion
rlp([tx_payload_body, blobs, commitments, [], cell_proofs])
```
why can't we have this, this will deterministically indicate that the proofs are cell proofs, and while we are doing transition even proofs can be sent instead of `[]` along with cell proofs?

EIPS/eip-7594.md
g11tech: I think this is easily resolvable by https://github.com/ethereum/EIPs/pull/9378/files#r1969600029

EIPS/eip-7594.md
marcindsobczak: It would require changes in RLP serialization. With a simple replacement of proofs -> cell_proofs, the RLP structure remains unchanged - only the number of elements differs, which is validated after serialization.
If we decide to modify the RLP structure, introducing a new transaction type would be a better approach. However, with a new transaction type, we could also eliminate proofs altogether - bringing us full circle.

dapplion: > @fjl said he would prefer to have a version byte in the new format

Rationale?

fradamt: My understanding: he wants to keep both wrapper types around during the transition and for some time after, and prefers to have a version to differentiate between the two, also for forward compatibility with changing the wrapper again in the future. Maybe @fjl can say more

EIPS/eip-7594.md
lightclient: I think it's better for this to be done by the tx sender. Blobs are sent by sophisticated users, so I think it's reasonable to expect them to do the proofing. We'll probably support it in the RPC just to help bootstrap things and be useful on devnet / testnets, but I don't think we want it to be the recommended route.

EIPS/eip-7594.md
lightclient: Seems like he likes:

```
rlp([tx_payload_body, type_prefix, blobs, commitments, cell_proof])
```

fjl: Hey, sorry for the indirect reporting. We discussed this during in-person meetings and that's where the suggestion to add a type came up. Let me try to explain it further.

Transactions are often validated outside of the fork schedule, i.e. there has to be a way to verify the basic validity of a transaction without knowing the block in which it will be included. This is especially true since we accept these transactions on p2p and RPC endpoints. In order to make validity checks work, we need to be aware of the blob proof type of the transaction after parsing it from RLP. **The encodings of both types of proof need to be distinct.**

So this is why I am proposing to add a proof version number (I suggest `2`) into the payload. It's one additional byte in the RLP encoding. I am also proposing the following handling of proof types 1 (current) and 2 (cell):

- When receiving an encoded transaction with proof type 2 before the PeerDAS fork activates, we reject it. Only type 1 proofs are accepted into the pool.
- At the time of fork activation, the implementer has two options:
  - Drop all blob transactions from mempool.
  - Convert all proofs to type two in the background
- After the fork has activated, only type 2 proofs are accepted via p2p relay. 
- On RPC (`eth_sendRawTransaction`), type 1 proofs may still be accepted and will be auto-converted by the node.
  At implementer discretion, this facility can be deprecated later when users have switched to new client libraries that can create type two proofs.

fjl: I think a separate tx type is not required, and it's in fact a bad solution. Changing the blob proof type does not change the transaction hash, and the transactions will look the same in blocks. So they should be same type.

fradamt: I added a `wrapper_version` field. If you don't like the name feel free to change it.

Other than that, do you think the EIP should mention anything else? I would perhaps only make it more explicit that the previous wrapper version should not be accepted in tx gossip after the fork, and also point to the Engine API spec and note that also there we are only going to accept the latest version with cell proofs.

The conversion of the existing pool and the accepting txs over RPC seems out of scope for the EIP since it's up to implementers.

fjl: It's a good question how we should define the acceptance on p2p. Like I said before, this is not usually tied to a fork block boundary. Only the engine API is. So we definitely need to define the behavior of the engine API here. 

The ideas around conversion and gossip acceptance need to be explained somehow, just maybe not in the Specification section.

0x00101010: should we make proofVersion = 0 as default and 1 as cell? This could make decoding easier, and more backward compatible?

EIPS/eip-7594.md
```diff
@@ -29,10 +29,39 @@ Nodes find and maintain a diverse peer set and sample columns from their peers t
...
+
+```
+rlp([tx_payload_body, wrapper_version, blobs, commitments, cell_proofs])
+
+cell_proofs = [cell_proof_0, cell_proof_1, ...]
+```
+
+The `tx_payload_body`, `blobs` and `commitments` are as in EIP-4844, while the `proofs` field is replaced by `cell_proofs`, and a `wrapper_version` is added. These are defined as follows:
+
+- `wrapper_version` - one byte indicating which version of the wrapper is used. For the current one, it is set to `2`.
```
dapplion, line 50:
```suggestion
- `wrapper_version` - one byte indicating which version of the wrapper is used. For the current one, it is set to `1`.  If `wrapper_version` is 0 the transaction must be decoded as a EIP-4844 transaction
```

g11tech: +1

PR #9381, opened 2025-02-19
Update EIP-7805: expand on EL changes

Marchhill:
- Add Engine API changes from https://github.com/ethereum/execution-apis/pull/609
- Clarify IL building process is unspecified in EL section
- Correct statements about block being invalid if IL not satisfied
- Clarify some details of EL / engine API changes

### DIFF
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -85,7 +85,7 @@ When validators receive ILs from the P2P network, they perform a series of valid
 
 ### Execution Layer
 
-On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we attempt to execute each valid transaction from ILs that was not present in the payload. If one of those transactions executes successfully, then an error is returned to the CL. Although the block is valid the CL will not attest to it.
+On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we check whether any transaction from ILs, that is not already present in the payload, could be validly included. If that is the case for any transaction, then an error is returned to the CL. Although the block is valid, the CL will not attest to it.
 
 Let `B` denote the current block.
 Let `S` denote the execution state following the execution of the last transaction in `B`.
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -115,7 +115,7 @@ The full Engine API changes can be found in the [Execution APIs](https://github.
 
 #### IL Building
 
-By default, ILs are built by selecting raw transactions from the public mempool, ordered by priority fees, up to the IL’s maximum size in bytes of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB` per IL. Additional rules can be optionally applied to maximize censorship resistance, such as prioritizing valid transactions that have been pending in the mempool the longest. Implementations are free to choose their own ordering rules.
+The rules for building ILs are left up to the implementers discretion. Possible strategies include randomly choosing from the transaction pool, ordering by priority fee, time spent pending in the mempool, etc.
 
 ### Consensus Layer
 
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -103,13 +103,31 @@ For each transaction `T` in ILs, perform the following:
 
 3. Execute `T` on state `S`. Assert that the execution of `T` fails.
 
-If `B` is full, the process terminates. Also note that we do not need to reset the state to `S`, since the only way for a transaction to alter the state is for it to execute successfully, in which case the block is invalid, and so the block will not be applied to the state.
+If at any point `B` is full (i.e. less than 21000 remaining gas), the process terminates. Also note that we do not need to reset the state to `S`, since the only way for a transaction to alter the state is for it to execute successfully, in which case the block is invalid, and so the block will not be applied to the state.
+
+#### Engine API Changes
 
 We make the following changes to the engine API:
 
-- Add `engine_getInclusionList` endpoint to retrieve an IL from the `ExecutionEngine`
-- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer
-- Modify `engine_forkchoiceUpdated` endpoint to include a field in the payload attributes for transactions in ILs determined by the proposer
+- Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`.
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update the IL that should be used to build the block. This takes as an argument an 8 byte `payloadId` of the ongoing payload build process, along with the IL itself.
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
+
+The IL is always encoded as an array of RLP encoded transactions in hex form.
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
+
+#### IL Building
+
+The IL should be built using the following rules:
+
+First order all transactions from the pool by priority fee, and then by timestamp in the event of a tie. For each transaction `T` in the sorted list:
+
+1. If adding `T` would make the IL exceed `MAX_BYTES_PER_INCLUSION_LIST` then skip this transaction.
+
+2. Otherwise add this transaction to the IL.
+
+3. If there is less than `100` bytes remaining then the IL is considered full and the process terminates.
 
 ### Consensus Layer
 
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -101,7 +101,7 @@ For each transaction `T` in ILs, perform the following:
 
 3. Execute `T` on state `S`. Assert that the execution of `T` fails.
 
-If at any point `B` is full, the process terminates. Also note that we do not need to reset the state to `S`, since the only way for a transaction to alter the state is for it to execute successfully, in which case the block is invalid, and so the block will not be applied to the state.
+If `B` is full, the process terminates. Also note that we do not need to reset the state to `S`, since the only way for a transaction to alter the state is for it to execute successfully, in which case the block is invalid, and so the block will not be applied to the state.
 
 #### Engine API Changes
 
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -110,7 +110,7 @@ If at any point `B` is full (i.e. less than 21000 remaining gas), the process te
 We make the following changes to the engine API:
 
 - Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`.
-- Add `engine_updatePayloadWithInclusionListV1` endpoint to update the IL that should be used to build the block. This takes as an argument an 8 byte `payloadId` of the ongoing payload build process, along with the IL itself.
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8 byte `payloadId` of the ongoing payload build process, along with the IL itself.
 - Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
 
 The IL is always encoded as an array of RLP encoded transactions in hex form.
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -85,7 +85,7 @@ When validators receive ILs from the P2P network, they perform a series of valid
 
 ### Execution Layer
 
-On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we check whether any transaction from ILs, that is not already present in the payload, could be validly included. If that is the case for any transaction, then an error is returned to the CL. Although the block is valid, the CL will not attest to it.
+On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we check whether any transaction from ILs, that is not already present in the payload, could be validly included (i.e. nonce and balance checks pass). If that is the case for any transaction, then an error is returned to the CL. Although the block is valid, the CL will not attest to it.
 
 Let `B` denote the current block.
 Let `S` denote the execution state following the execution of the last transaction in `B`.
@@ -111,7 +111,7 @@ We make the following changes to the engine API:
 - Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8-byte `payloadId` of the ongoing payload build process, along with the IL itself.
 - Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the IL committee member. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
 
-The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/572c54cd9d003f700bb5a57ce201606adbc8fc1c/src/engine/experimental/eip7805.md) GitHub repository.
 
 #### IL Building
 
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -14,7 +14,7 @@ created: 2024-11-01
 
 FOCIL implements a robust mechanism to preserve Ethereum’s censorship resistance properties by guaranteeing timely transaction inclusion.
 
-FOCIL (**Fo**rk-choice enforced **I**nclusion **L**ists) is built in a few simple steps:
+FOCIL (**Fo**rk-**c**hoice enforced **I**nclusion **L**ists) is built in a few simple steps:
 
 - In each slot, a set of validators is selected as inclusion list (IL) committee members. Each member builds and gossips one IL according to their subjective view of the mempool.
 - The proposer and all attesters of the next slot monitor, store and forward available ILs.
@@ -89,20 +89,19 @@ On the execution layer, an additional check is introduced for new payloads. Afte
 
 Let `B` denote the current block.
 Let `S` denote the execution state following the execution of the last transaction in `B`.
+Let `gas_left` be the gas remaining after execution of B.
 
 For each transaction `T` in ILs, perform the following:
 
 1. Check whether `T` is present in `B`. If `T` is present, then jump to the next transaction, else continue with next step.
 
-2. Validate `T` against `S`. 
+2. Check whether `B` has enough remaining gas to execute `T`. If `T.gas` > `gas_left`, then jump to the next transaction, else continue with next step.
+
+3. Validate `T` against `S`.
 
   - If `T` is invalid, then continue to the next transaction. 
   
-  - If `T` is valid, terminate process and return an error.
-
-3. Execute `T` on state `S`. Assert that the execution of `T` fails.
-
-If `B` is full, the process terminates. Also note that the state `S` must be restored if a transaction from the IL is successfully executed against it.
+  - If `T` is valid, terminate process and return an `INVALID_INCLUSION_LIST` error.
 
 #### Engine API Changes
 
@@ -110,13 +109,13 @@ We make the following changes to the engine API:
 
 - Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`.
 - Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8-byte `payloadId` of the ongoing payload build process, along with the IL itself.
-- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the IL committee member. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
 
 The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
 
 #### IL Building
 
-The rules for building ILs are left to the discretion of implementers. For instance, they may select transactions from the public mempool in various ways such as at random, by priority fee, or based on how long they have been pending. This process continues until the IL reaches its maximum size of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB`.
+The rules for building ILs are left to the discretion of implementers. For instance, they may select transactions from the public mempool in various ways such as at random, by priority fee, or based on how long they have been pending. The IL has a maximum size of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB` for all of the unserialised transactions.
 
 ### Consensus Layer
 
--- a/EIPS/eip-7805.md
+++ b/EIPS/eip-7805.md
@@ -115,7 +115,7 @@ The full Engine API changes can be found in the [Execution APIs](https://github.
 
 #### IL Building
 
-The rules for building ILs are left to the discretion of implementers. For instance, they may select transactions from the public mempool in various ways such as at random, by priority fee, or based on how long they have been pending. The IL has a maximum size of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB` for all of the unserialised transactions.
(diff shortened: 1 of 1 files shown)


EIPS/eip-7805.md
```diff
@@ -103,13 +103,31 @@ For each transaction `T` in ILs, perform the following:
...
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update the IL that should be used to build the block. This takes as an argument an 8 byte `payloadId` of the ongoing payload build process, along with the IL itself.
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
+
+The IL is always encoded as an array of RLP encoded transactions in hex form.
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
+
+#### IL Building
+
+The IL should be built using the following rules:
```
jihoonsong, line 122: Actually we don't impose IL ordering rules. Some clients may prefer to pick transactions based on priority fee while others may choose different heuristic algorithms. We expect more diverse approaches on IL construction would foster censorship resistance.

Marchhill: That makes sense, but conflicts with another part of the EIP:

> By default, ILs are built by selecting raw transactions from the public mempool, ordered by priority fees

I can update both places

EIPS/eip-7805.md
```diff
@@ -105,11 +103,21 @@ For each transaction `T` in ILs, perform the following:
...
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8 byte `payloadId` of the ongoing payload build process, along with the IL itself.
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
+
+The IL is always encoded as an array of RLP encoded transactions in hex form.
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
+
+#### IL Building
+
+By default, ILs are built by selecting raw transactions from the public mempool, ordered by priority fees, up to the IL’s maximum size in bytes of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB` per IL. Additional rules can be optionally applied to maximize censorship resistance, such as prioritizing valid transactions that have been pending in the mempool the longest. Implementations are free to choose their own ordering rules.
```
jihoonsong, line 120: There is no default rule. While ordering and selecting transactions by priority fee is probably the most anticipated strategy, it should remain as optional, too.

Marchhill: This was moved from elsewhere in the EIP but maybe should be changed

jihoonsong: Where did you get it from?

jihoonsong: Regarding adding this to the EIP, I want to ask other people's opinions. cc @soispoke @terencechain

Marchhill: previously it was here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7805.md#il-committee-members

jihoonsong: Ah, right, my bad. I prefer to keep it as is and have only engine API changes.

Marchhill: imo it's nice to have it here so the EL changes are contained in this section, easier to know what needs to be implemented. You think having an extra section here is bad for readability?

jihoonsong: From the big picture perspective, yes. The original version provides a guidance on how IL committee members could construct their ILs when a reader reads the role of IL committee member. This change removes it and hence brings ambiguity compared to before.

Marchhill: I made a change to mention IL building in that earlier section but then expand on it in the EL section, wdyt?

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
...
-- Modify `engine_forkchoiceUpdated` endpoint to include a field in the payload attributes for transactions in ILs determined by the proposer
+- Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`.
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8 byte `payloadId` of the ongoing payload build process, along with the IL itself.
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
+
+#### IL Building
+
+The rules for building ILs are left up to the implementers discretion. Possible strategies include randomly choosing from the transaction pool, ordering by priority fee, time spent pending in the mempool, etc.
```
jihoonsong, line 119:
```suggestion
The rules for building ILs are left to the discretion of implementers. For instance, they may select raw transactions from the public mempool in various ways such as at random, by priority fee, or based on how long they have been pending. This process continues until the IL reaches its maximum size of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB`.
```

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
...
+
+#### Engine API Changes
 
 We make the following changes to the engine API:
 
-- Add `engine_getInclusionList` endpoint to retrieve an IL from the `ExecutionEngine`
-- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer
-- Modify `engine_forkchoiceUpdated` endpoint to include a field in the payload attributes for transactions in ILs determined by the proposer
+- Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`.
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8 byte `payloadId` of the ongoing payload build process, along with the IL itself.
```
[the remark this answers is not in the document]
Marchhill: 'bytes' doesn't sound right to me here, but I think there should be a hyphen: https://ell.stackexchange.com/questions/93379/do-you-say-2-byte-or-2-bytes

soispoke (commented):
lgtm

soispoke (approved):

soispoke (commented):
it should just be validity (nonce and balance) checks, not execution

EIPS/eip-7805.md
```diff
@@ -85,8 +85,7 @@ When validators receive ILs from the P2P network, they perform a series of valid
 
 ### Execution Layer
 
-On the execution layer, the block validity conditions are extended such that, after all of the transactions in the block have been executed, we attempt to execute each valid transaction from ILs that was not present in the block.
-If one of those transactions executes successfully, then the block is invalid.
+On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we attempt to execute each valid transaction from ILs that was not present in the payload. If one of those transactions executes successfully, then an error is returned to the CL. Although the block is valid the CL will not attest to it.
```
fradamt, line 88:
```suggestion
On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we check whether any transaction from ILs, that is not already present in the payload, could be validly included. In other words, we validate the transaction against the post-state of the block, and also check whether there is sufficient gas left to include it. If one of those transactions could be included, then an error is returned to the CL. Although the block is valid, the CL will not attest to it.
```

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
 
   - If `T` is invalid, then continue to the next transaction. 
   
-  - If `T` is valid, terminate process and assert block `B` as invalid.
+  - If `T` is valid, terminate process and return an error.
 
 3. Execute `T` on state `S`. Assert that the execution of `T` fails.
```
fradamt, line 103:
```suggestion
```

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
 
   - If `T` is invalid, then continue to the next transaction.
```
fradamt, line 99:
```suggestion
  - If `T` is invalid against `S`, then continue to the next transaction. 
```

EIPS/eip-7805.md
```diff
@@ -85,8 +85,7 @@ When validators receive ILs from the P2P network, they perform a series of valid
 
 ### Execution Layer
 
-On the execution layer, the block validity conditions are extended such that, after all of the transactions in the block have been executed, we attempt to execute each valid transaction from ILs that was not present in the block.
-If one of those transactions executes successfully, then the block is invalid.
+On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we attempt to execute each valid transaction from ILs that was not present in the payload. If one of those transactions executes successfully, then an error is returned to the CL. Although the block is valid the CL will not attest to it.
 
 Let `B` denote the current block.
```
fradamt, line 90: Let `B` denote the current block.
Let `S` denote the execution state following the execution of the last transaction in `B`.
Let `gas_left` be the gas remaining after execution of `B`.

For each transaction `tx` in ILs, perform the following:

1. Check whether `tx` is present in `B`. If `tx` is present, then jump to the next transaction, else continue with next step.
2. Validate `tx` against `S`.

  - If  `tx` is invalid or `tx.gas > gas_left`, then continue to the next transaction. 
  - If `tx` is valid and `tx.gas <= gas_left`, terminate process and return an error.

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
 
   - If `T` is invalid, then continue to the next transaction. 
   
-  - If `T` is valid, terminate process and assert block `B` as invalid.
+  - If `T` is valid, terminate process and return an error.
 
 3. Execute `T` on state `S`. Assert that the execution of `T` fails.
 
-If `B` is full, the process terminates. Also note that we do not need to reset the state to `S`, since the only way for a transaction to alter the state is for it to execute successfully, in which case the block is invalid, and so the block will not be applied to the state.
+If `B` is full, the process terminates. Also note that the state `S` must be restored if a transaction from the IL is successfully executed against it.
```
fradamt, line 105: I'd remove these two lines. I don't think it makes sense to talk about `B` being full, since it's only full if it uses exactly the block gas limit. It makes sense to talk about this tx per tx explicitly, by comparing `tx.gas` and `gas_left` like done above.  I also don't think it makes sense to talk about `S` being restored since we're never doing any execution.

fradamt, line 105:
```suggestion
```

I don't think it makes sense to talk about `B` being full, since it's only full if it uses exactly the block gas limit. It makes sense to talk about this tx per tx explicitly, by comparing `tx.gas` and `gas_left` like done above. 

 I also don't think it makes sense to talk about `S` being restored since we're never doing any execution.

EIPS/eip-7805.md
jihoonsong: I think it's more precise to view `gas_left` as a separate value rather than as a part of the execution state.

---

Let `B` denote the current block.
Let `S` denote the execution state following the execution of the last transaction in `B`.
Let `gas_left` be the gas remaining after execution of `B`.

For each transaction `tx` in ILs, perform the following:

1. Check whether `tx` is present in B. If `tx` is present, then jump to the next transaction, else continue with next step.

2. Check whether `B` has enough remaining gas to execute `tx`. If `tx.gas` > `gas_left`, then jump to the next transaction, else continue with next step.

3. Let `balance` and `nonce` be the balance and nonce of `tx.origin` derived from `S` respectively. If the required funds for `tx` > `balance` or `tx.nonce` < `nonce`, then jump to the next transaction, else terminate process and return an error.

EIPS/eip-7805.md
jacobkaufmann: does it make sense to say "the block is valid" here if we return an error? although the meaning is clear to me, we could clarify a bit. not a blocker for approval though.

a further nit: it's nice to try to keep one sentence per line in the markdown for easier review.
I know that we are inconsistent on that in this EIP though.

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
...
+#### Engine API Changes
 
 We make the following changes to the engine API:
 
-- Add `engine_getInclusionList` endpoint to retrieve an IL from the `ExecutionEngine`
-- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer
-- Modify `engine_forkchoiceUpdated` endpoint to include a field in the payload attributes for transactions in ILs determined by the proposer
+- Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`.
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8-byte `payloadId` of the ongoing payload build process, along with the IL itself.
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
```
jacobkaufmann, line 113:
```suggestion
- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the validator. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
```

or "attester"? it is not the proposer that supplies the IL that should be satisfied by the proposer's payload.

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
...
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
```
jacobkaufmann, line 115: I don't think we can provide external links.

jacobkaufmann (commented):
@Marchhill can we address the comments?

EIPS/eip-7805.md
jihoonsong: Or we could use "the IL committee members"

EIPS/eip-7805.md
```diff
@@ -99,17 +98,25 @@ For each transaction `T` in ILs, perform the following:
...
+
+#### Engine API Changes
 
 We make the following changes to the engine API:
 
-- Add `engine_getInclusionList` endpoint to retrieve an IL from the `ExecutionEngine`
-- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer
-- Modify `engine_forkchoiceUpdated` endpoint to include a field in the payload attributes for transactions in ILs determined by the proposer
+- Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`.
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8-byte `payloadId` of the ongoing payload build process, along with the IL itself.
```
jochem-brouwer, line 112: I'm confused by this. I might be totally getting the CL <-> EL communication wrong though, but I cannot wrap my head around it currently.

`newPayload` has a proposed new parameter for the IL. `newPayload` is used to verify the payload (but does not return a `payloadId`). The `payloadId` is returned by `forkchoiceUpdated`, but only if the CL is proposing a block (then CL will send the extra `payloadAttributes` param to instruct EL to start building). So `newPayload` can be used to verify if the block is valid given the IL, but this can only be done after the build process has finished (`forkchoiceUpdated` with a `payloadAttributes` returns `payloadId`, then `getPayload` to get the payload and then to verify `newPayload` with the IL). This also means that only after the build process has started (via `forkchoiceUpdated`) the `updatePayloadWithInclusionList` can be called (this should interrupt the process and start over by now including the IL). 

I think that these changes should be made instead:

`engine_newPayloadV5`

params: copy from https://github.com/ethereum/execution-apis/blob/3779ff972be60cff5fec93e9b51b0ec24aa52e2d/src/engine/prague.md#engine_newpayloadv4 and include `v. inclusionList` param. This should also verify the `MaxBytesPerInclusionList` check (and spec how this size is measured, for instance the RLP of these txs?)

`engine_forkchoiceUpdatedV4`

Builds on `forkchoiceUpdatedV3` https://github.com/ethereum/execution-apis/blob/3779ff972be60cff5fec93e9b51b0ec24aa52e2d/src/engine/cancun.md#engine_forkchoiceupdatedv3 and adds `PayloadAttributesV4` with the `inclusionList` as extra attribute (this should also verify the `MaxBytesPerInclusionList` I think)

`engine_updatePayloadWithInclusionListV1`

I think this can be removed. To update the IL in the build process, just call `forkchoiceUpdatedV4` again. The `forkchoiceState` will (should?) stay the same in the build process, so it is clear that  the build process should be started over with these new `payloadAttributes` (new IL)

`engine_getInclusionListV1` 

Keep as currently specced

The main point here is that when building the block we should be aware of the IL. Otherwise the verification process is: build the block, then verify it (`newPayload`). The build process returns the `payloadId` and the only way currently to set the IL is via the `updatePayloadWithInclusionList`, but this needs the `payloadId`.

jochem-brouwer (commented):
I think that IL txs should not be executed (this is a DoS vector) but only validated if they could be included (points also made by other reviewers). 

I also am confused by the engine API spec, but I might be completely messing up the entire EL <-> CL communication process here :sweat_smile: Have proposed changes to it.

EIPS/eip-7805.md
jihoonsong: @jochem-brouwer Updated the execution-api PR, thanks!

EIPS/eip-7805.md
jihoonsong: 0.
Regarding `engine_updatePayloadWithInclusionListV1`, the main reason I've added was because I wasn't sure how EL devs would feel about overloading forkchoice attribute. We can remove it and add `engine_forkchoiceUpdatedV4`, if that's what EL devs prefer. We just haven't had enough feedback from EL devs thus far.

Another reason of having `engine_updatePayloadWithInclusionListV1` was that it's easier to implement and we wanted to make faster progress. The extra amount of work of merging it into `engine_forkchoiceUpdatedV4` can be justified given where we're stand at the moment IMHO.

1.
Block building is one thing and block verification is another. A block builder doesn't need to verify their block. They're the one who builds a block according to IL constraints. Attesters are the one who verify a block and attest to it, if it's legit.

jochem-brouwer: Could you then show me how the current flow would go for building the block? How would the builder know the IL? I'd assume `forkchoiceUpdated`, and then `updatePayloadWithInclusionList` with the `payloadId` returned from the `forkchoiceUpdated`? This would thus chain 2 calls - right? I'd then really suggest to put this in `forkchoiceUpdated` :thinking:

EIPS/eip-7805.md
Marchhill: @jacobkaufmann I think it's important to specify that the block is valid here, this affects which code path we take in the EL

EIPS/eip-7805.md
jihoonsong: > I'd assume forkchoiceUpdated, and then updatePayloadWithInclusionList with the payloadId returned from the forkchoiceUpdated?

Yes, this is current implementation.

It's not guaranteed to have all the ILs until 9 seconds into the slot. So if we were to use `forkchoiceUpdatedV4` and make EL call only once, you need to wait until 9 seconds into the slot before initiating the payload building process. I'd like to learn more about how people would think about this change.

I don't have a strong opinion on either way. Thank you for your feedback and I'd encourage more people to give us feedback to find a better solution :)

EIPS/eip-7805.md
Marchhill: @SamWilsn would it be possible to add execution api as an allowed external link?

EIPS/eip-7805.md
Marchhill: @jihoonsong @fradamt is it sufficient to check the nonce and balance? When we are talking about validating against S are we not simulating execution against S?

EIPS/eip-7805.md
jochem-brouwer: I disagree, the block builder can already start building the block immediately, it can just call `forkchoiceUpdatedV4` and we thus assume the IL list is empty. If the IL list gets updated, then either the `updatePayloadWithInclusionList` could be called (if this is the approach) or if we insert it as param for `forkchoiceUpdated` then it would call it with exactly the same arguments, but with the updated IL.

I would assume that block builders start building the block immediately. I am not sure, but I think it is likely if the node is not censoring that in most cases, if it has received all ILs (or time to receive new ILs is over) then I think it is likely that the block already consists of most of these txs.

But this is a good point, I am not super familiar with the block building process which clients implement, and it would also be worth to study what "big block builders" implement and how they might run into build problems with the new ILs.

Note that the EIP-1559 fee market will target the blocks to be full on average of the `gasTarget` which is half the `gasLimit`. We can thus assume that a "normal block" is 50% full. If the IL then contains transactions which are NOT in the block yet, but are valid (sufficient balance, correct nonce, willing to pay enough for the `baseFee` and enough gas left in the block), then these can be added on top of the existing block (which was already built). On the other hand, if the block is rather full then the txs might not fit in anymore (due to gas limits). But this block is thus per this spec compliant with the ILs and the ELs will accept these blocks (and the CLs will thus attest).

jihoonsong: If you add a new engine API, which is `forkchoiceUpdatedV4 ` and make EL call twice, what's the improvement?

jochem-brouwer: The freedom that you can start the building process at any moment. The version with `updatePayloadWithInclusionList` makes it mandatory to start the build process with the empty list. Then one has to update/interrupt the build process with the inclusion list by calling `updatePayloadWithInclusionList`

jihoonsong: I don't think so. You have the same freedom in `updatePayloadWithInclusionList` approach. In your scenario, the only difference is whether the new engine API is `forkchoiceUpdatedV4` or `updatePayloadWithInclusionList`.

Also, the payload building usually finishes in 100ms. You don't interrupt the building process, you just initiate another round with ILs.

The only case `forkchoiceUpdatedV4` approach is better is when you initiate the payload building process after 11 seconds into the previous slot. Given that most of blocks play timing games, I started to think that this would be better in that sense.

EIPS/eip-7805.md
soispoke: lgtm

soispoke (approved):
if it looks good to everyone I'm fine merging the PR

EIPS/eip-7805.md
```diff
@@ -85,31 +85,37 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the IL committee member. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/58b019ea5ee4a84d9767402145b9a88789f0359e/src/engine/experimental/eip7805.md) GitHub repository.
```
jihoonsong, line 114: I'm not sure if we can add a link to `execution-apis` but if we can, could you please change the link with this latest one? https://github.com/ethereum/execution-apis/blob/572c54cd9d003f700bb5a57ce201606adbc8fc1c/src/engine/experimental/eip7805.md

jihoonsong (commented):
Once we deal with the link to `execution-apis`, it looks good to me.

nerolation:
Just some minor feedback: It would be nice revamp the EL section: Instead of using the current (scientific) notation, it would be cleaner to stick to the EL specs. For example, the variable gas_left is already defined in the specs, thus interpreted differently, and what you actually want is gas_available. Maybe have a pseudo code function or directly put a `is_includable(state, transaction)` function into the specs and the EIP.

EIPS/eip-7805.md
gurukamath: I think we may need more checks to make sure that a transaction is includable. As an example, 7702 `SetCodeTransactions` that have empty authorisation lists are invalid.

See [comment](https://github.com/ethereum/execution-specs/pull/1214#discussion_r2102402039).

Marchhill:
> Just some minor feedback: It would be nice revamp the EL section: Instead of using the current (scientific) notation, it would be cleaner to stick to the EL specs. For example, the variable gas_left is already defined in the specs, thus interpreted differently, and what you actually want is gas_available. Maybe have a pseudo code function or directly put a `is_includable(state, transaction)` function into the specs and the EIP.

Makes sense. I think for now will get this merged as is, but you could make another PR on top of this to change

EIPS/eip-7805.md
Marchhill: made a PR to allow links to execution api https://github.com/ethereum/EIPs/pull/9809

PR #9396, opened 2025-02-22
Update EIP-7805: IL building algorithm

Marchhill:
Suggests building IL using nondeterminism and bias, explained [here](https://ethereum-magicians.org/t/eip-7805-committee-based-fork-choice-enforced-inclusion-lists-focil/21578/6?u=marchhill)


EIPS/eip-7805.md
```diff
@@ -85,31 +87,47 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the IL committee member. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/572c54cd9d003f700bb5a57ce201606adbc8fc1c/src/engine/experimental/eip7805.md) GitHub repository.
+
+#### IL Building
+
+The rules for building ILs are left to the discretion of implementers; a diverse set of strategies can strengthen censorship resistance. Transactions from the public mempool could be selected based on several factors such as:
+
+- Time spent pending in the mempool.
+- Nondeterminism: Introducing randomness can reduce overlap between ILs, leading to a larger IL when they are combined.
```
soispoke, line 124: could you just move these two items (nondeterminism and bias) after byte size?

EIPS/eip-7805.md
```diff
@@ -40,7 +40,9 @@ This section outlines the workflow of FOCIL, detailing the roles and responsibil
 - **`Slot N`, `t=0 to 8s`**:
 IL committee members construct their ILs by including transactions pending in the public mempool, and broadcast them over the P2P network after processing the block for `slot N` and confirming it as the head. If no block is received by `t=7s`, they should run `get_head` and build and release their ILs based on the node’s local head.
 
-  By default, ILs are built by selecting raw transactions from the public mempool, ordered by priority fees, up to the IL’s maximum size in bytes of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB` per IL. Additional rules can be optionally applied to maximize censorship resistance, such as prioritizing valid transactions that have been pending in the mempool the longest.
+IL committee members may follow different strategies for constructing their ILs as discussed in [IL Building](#il-building).
+
+Each committee member is assigned an id from `0` to `f`. This is derived from the validator index, the lowest validator index is mapped to `0` and the highest to `f`.
```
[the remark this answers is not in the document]
Marchhill: Ok, will define the validator id along with the engine rpc method

EIPS/eip-7805.md
```diff
@@ -85,31 +85,46 @@ When validators receive ILs from the P2P network, they perform a series of valid
 
 ### Execution Layer
 
-On the execution layer, the block validity conditions are extended such that, after all of the transactions in the block have been executed, we attempt to execute each valid transaction from ILs that was not present in the block.
-If one of those transactions executes successfully, then the block is invalid.
+On the execution layer, an additional check is introduced for new payloads. After all of the transactions in the payload have been executed, we check whether any transaction from ILs, that is not already present in the payload, could be validly included (i.e. nonce and balance checks pass). If that is the case for any transaction, then an error is returned to the CL. Although the block is valid, the CL will not attest to it.
```
jochem-brouwer, line 88: Besides `nonce` and `balance` check other checks have to be passed as well. For instance: does it allow to pay the `baseFee`? (Other tx validity checks which are independent of the block environment (e.g. `baseFee` which depends on the current block) such as `tx.gaslimit` is sufficient (pays intrinsic fee) should be checked as well. (This is a note: the balance/nonce should be the most prominent checks but obviously other params have to be checked as well -> this is ok to be motivated later, but a mention of this here would be more clear)

EIPS/eip-7805.md
```diff
@@ -85,31 +85,46 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
 
 Let `B` denote the current block.
 Let `S` denote the execution state following the execution of the last transaction in `B`.
+Let `gas_left` be the gas remaining after execution of B.
```
jochem-brouwer, line 92: I think here the definition of `gas_left` could be made a bit more clear. It is clear that this is the gas left in the block (`block.gasLimit - cumulativeGasUsed` where `cumulativeGasUsed` is the gas spent by all txs in the block). (This is also a note, I think from context it is clear enough, but could be made more explicit)

EIPS/eip-7805.md
```diff
@@ -85,31 +85,46 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
 
 For each transaction `T` in ILs, perform the following:
 
 1. Check whether `T` is present in `B`. If `T` is present, then jump to the next transaction, else continue with next step.
 
-2. Validate `T` against `S`. 
+2. Check whether `B` has enough remaining gas to execute `T`. If `T.gas` > `gas_left`, then jump to the next transaction, else continue with next step.
+
+3. Validate `T` against `S` by checking the nonce and balance of `T.origin`.
```
jochem-brouwer, line 100: This is clear, but I would mention previously that `S` is thus the state **before** processing the withdrawals (this could lead that the tx sender now has enough balance to execute tx as without processing the withdrawal)

EIPS/eip-7805.md
```diff
@@ -85,31 +85,46 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
 
   - If `T` is invalid, then continue to the next transaction.
```
jochem-brouwer, line 102: (invalid implicilty also thus checks things like `baseFee` check against `maxFeePerGas`)

EIPS/eip-7805.md
```diff
@@ -85,31 +85,46 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
+#### Engine API Changes
 
 We make the following changes to the engine API:
 
-- Add `engine_getInclusionList` endpoint to retrieve an IL from the `ExecutionEngine`
-- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the proposer
-- Modify `engine_forkchoiceUpdated` endpoint to include a field in the payload attributes for transactions in ILs determined by the proposer
+- Add `engine_getInclusionListV1` endpoint to retrieve an IL from the `ExecutionEngine`. It takes as an argument the IL committee member id, where the id is a value from `0`-`f` derived from ordering by validator index (i.e. lowest index has id `0`).
+- Add `engine_updatePayloadWithInclusionListV1` endpoint to update a payload with the IL that should be used to build the block. This takes as an argument an 8-byte `payloadId` of the ongoing payload build process, along with the IL itself.
+- Modify `engine_newPayload` endpoint to include a parameter for transactions in ILs determined by the IL committee member. If the IL is not satisfied an `INVALID_INCLUSION_LIST` error must be returned.
```
jochem-brouwer, line 112: I disagree with the current engine API proposal, see discussion here: https://github.com/ethereum/EIPs/pull/9381#discussion_r2096501253

EIPS/eip-7805.md
```diff
@@ -85,31 +85,46 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
+
+The full Engine API changes can be found in the [Execution APIs](https://github.com/ethereum/execution-apis/blob/572c54cd9d003f700bb5a57ce201606adbc8fc1c/src/engine/experimental/eip7805.md) GitHub repository.
+
+#### IL Building
+
+The rules for building ILs are left to the discretion of implementers; a diverse set of strategies can strengthen censorship resistance. Transactions from the public mempool could be selected based on several factors such as:
+
+- Time spent pending in the mempool.
+- Priority fee.
+- Byte size: prioritising smaller transactions.
```
jochem-brouwer, line 122: Why would we prioritize smaller txs? "Big" txs are already invalid in the ILs because of the limit `MAX_BYTES_PER_INCLUSION_LIST`. So I'd argue that big txs have more priority over smaller txs, because smaller txs are more probable to get added to the ILs (especially as the "last tx" to get added to fill the "remainder" of the IL)

jochem-brouwer (commented):
Some minor comments regarding clarification, plus have linked my thoughts on the current idea of changes to the engine API and a disagreement on how to construct ILs :smile: :+1:

EIPS/eip-7805.md
jihoonsong: The withdrawals happen before transactions execution. See https://ethereum.github.io/consensus-specs/specs/electra/beacon-chain/#block-processing

EIPS/eip-7805.md
Marchhill: I think `gas_left` is quite clear as this is the variable used in the execution specs, so you wouldn't need to do this subtraction to calculate

EIPS/eip-7805.md
jochem-brouwer: No, withdrawals happen after tx execution, see the EIP:

https://eips.ethereum.org/EIPS/eip-4895#state-transition

> The `withdrawals` in an execution payload are processed **after** any user-level transactions are applied.

EIPS/eip-7805.md
jochem-brouwer: Ah ok, yes I see, I had not thought of this type of attack. For "bigger" txs which do not get included, because of the suggestion of "time spent pending in the mempool" this would at some point thus get prioritzed and picked by one of the IL creators of the committee?

jochem-brouwer:
> > Some minor comments regarding clarification, plus have linked my thoughts on the current idea of changes to the engine API and a disagreement on how to construct ILs 😄 👍
> 
> Thanks for feedback! Just to be clear, this PR is just to add the IL building stuff. Most of these comments apply to the [original PR](https://github.com/ethereum/EIPs/pull/9381) it's built on so will make any changes there that aren't related to IL building

Ah ok! You could (for the diff) also point the original PR to only present the relevant diff here and not the changes of the OG PR also :smile: :+1:

EIPS/eip-7805.md
```diff
@@ -85,31 +85,46 @@ When validators receive ILs from the P2P network, they perform a series of valid
...
+- Nondeterminism: Introducing randomness can reduce overlap between ILs, leading to a larger IL when they are combined. Note that some overlap is desirable, otherwise the 1 of N honesty assumption would be undermined.
+- Bias: Picking some transactions more often than others in order to further reduce overlap between ILs. This should be done based on committee member id, for example the member with id `a` may be more likely to pick transactions with hashes that start with `a`.
+- The number of times a transaction has been included in reorged payloads.
+
+The IL has a maximum size of `MAX_BYTES_PER_INCLUSION_LIST = 8 KiB` for all of the unserialised transactions.
```
jochem-brouwer, line 127: I'm slightly confused how this should be measured. The max list size is 8 KiB, so we have to encode the txs in a list... so some kind of serialization has to happen, right?

Marchhill: Fixed in other PR, should be RLP encoded I think

PR #9412, opened 2025-02-25
Update EIP-7620: Move to Review

pdobacz:
EOFCREATE and TXCREATE (and also the soon-to-be-obsolete EOF creation txs) must obey EIP-161 and set the nonce of the newly created account to 1 before executing initcode


pdobacz:
We are inclined to take on a different approach - for behavior which doesn't change from legacy CREATE/CREATE2 to EOF/TXCREATE we should just mention that, without being overly meticulous in listing the exact steps. This is in order to not obfuscate the _important_ pieces of the spec. This is similar to how the EIP-7069 (EXT*CALL) is phrased.

At the same time, all of these behaviors will need be covered by EEST tests, c.f. https://github.com/ethereum/execution-spec-tests/pull/1270 https://github.com/ethereum/execution-spec-tests/pull/1271

PR #9415, opened 2025-02-26
Update EIP-7864: Finalize hash function selection to Poseidon2

GarmashAlex:
eip-7864: Keep hash function as TBD with detailed analysis

This PR updates the hash function section of the unified binary tree state EIP to:

1. Keep the hash function selection as TBD while providing detailed analysis of the candidates:
   - BLAKE3: Used in reference implementation for experimentation
   - Keccak: Already used in Ethereum, well-studied security
   - Poseidon2: Excellent for ZK proofs, pending security analysis

2. Make the specification hash-function agnostic by:
   - Updating the node merkelization section to use a generic hash function H
   - Clarifying that the reference implementation uses BLAKE3 temporarily
   - Adding context about the ongoing EF cryptography initiative for Poseidon2

3. Update the TODO section to include finalizing the hash function selection

This approach allows EL clients to experiment with the EIP using BLAKE3 while research continues on the optimal hash function choice.

### DIFF
--- a/EIPS/eip-7864.md
+++ b/EIPS/eip-7864.md
@@ -14,34 +14,54 @@ created: 2025-01-20
 
 Introduce a new binary state tree, intended to replace the hexary patricia trees. Account and storage tries are merged into a single tree with 32-byte keys, which also contains contracts code. Account data is broken into independent leaves which  are grouped by 256 in order to provide some locality.
 
-Note: the hash function used in the current draft is not final, other potential candidates are Keccak or Poseidon2. For the latter, an extra section on how to encode 32-bytes->[FiniteField] will be created. **Do not** assume Blake3 is a final decision. <-- TODO finalize hash function -->
+Note: This EIP uses the Poseidon2 hash function for all tree operations. The function provides optimal performance for zero-knowledge proofs while maintaining strong security properties.
+
+### Poseidon2 Parameters
+
+The Poseidon2 hash function is instantiated with the following parameters:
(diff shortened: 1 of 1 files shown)


jsign:
@GarmashAlex, the decision about the hash function to be used will remain TBD for a while.

Regarding Poseidon2 security, there's currently an EF cryptography [initiative](https://www.poseidon-initiative.info/), which is trying to assess if it's considered safe. While that is happening, the current proposal of BLAKE3 is to reduce friction in EL that want to experiment with implementing the EIP -- the hash function can be switched with almost zero impact on existing implementations.

Even if Posiedon2 is considered safe (or safe-ish+workarounds), another discussion is which field should be used. Your proposal, BN254 scalar field, isn't obvious compared with 31-bit field elements or even binary proofs. So, this is another angle that requires further discussion and probably waiting until this EIP is closer to being accepted to have more information about proving system progress.

jsign:
@GarmashAlex, I see you updated the changes based on my comment, looks good.
Do you mind adjusting the PR description to summarize the current PR change (instead of the previous removed one)?

GarmashAlex:
@jsign  Corrected

jsign (approved):

jsign:
@GarmashAlex, ah do you mind editing also the PR title?

GarmashAlex:
@jsign  Yes i will change it soon

GarmashAlex:
@jsign like this?

PR #9417, opened 2025-02-27
Update EIP-7620/7698/7873 - Defer EOF creation rules to legacy

pdobacz:
Makes note in all the EIPs concerning EOF contract creation, that whenever not explicitly stated, legacy rules should be had in mind.


shemnon (commented):
We should drop the  in 7873.  I'd do an inline comment but github won't let me.

PR #9429, opened 2025-02-28
Fix TODO: Add Shanghai fork check for code access list in EVM

DeVikingMark:
Implements the TODO in core/vm/evm.go by adding a check for Shanghai fork activation before adding an address to the code access list. This change ensures EIP-5027 functionality is only enabled when the Shanghai fork is active.


SamWilsn:
This'll require approval from the author @qizhou

qizhou:
Hi, EIP-5027 is superseded by EIP-7907.  We will not support EIP-5027 in the future.

PR #9447, opened 2025-03-04
Update EIP-1: fix typo in exec spec tests link

maximevtush:
typo in the example URL


SamWilsn:
`execution-specs` is distinct from `execution-spec-tests` and we allow links to both.

PR #9450, opened 2025-03-04
Update EIP-2242: add Test Cases and Implementation for EIP-2242 (Transaction Postdata)

crStiv:
- Comprehensive test cases covering functionality, gas calculation, error handling, RLP encoding, and compatibility
- Reference implementation in Go with transaction structure updates, RLP encoding/decoding, gas calculation, and validation logic

### DIFF
--- a/EIPS/eip-2242.md
+++ b/EIPS/eip-2242.md
@@ -10,20 +10,25 @@ created: 2019-08-16
 ---
 
 ## Simple Summary
+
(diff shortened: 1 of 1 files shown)


g11tech:
since this PR changes the EIP text in substantial manner, it can't be merged by without the review of the author that it doesn't meaningfully change the EIP, but this might be unlikely since the EIP is stagnant

so does the author wants to champion and push for this EIP? if yes, then recommend to get in touch with the author to help them rechampion

PR #9453, opened 2025-03-05
Update EIP-6110: add filtering by deposit event sig hash

mkalinin:


abcoathup:
@mkalinin minor typo “deisgn”

PR #9457, opened 2025-03-05
Update EIP-template: Add chain-specific considerations

Marchhill:
Added a section to the EIP template to address any chain-specific considerations.

The aim is to reduce the likelihood of incidents like [the recent one on Sepolia](https://blog.ethereum.org/2025/03/05/sepolia-pectra-incident), which are caused by minor differences between chains. If EIP authors address these considerations then such problems could be caught much earlier.

### DIFF
--- a/eip-template.md
+++ b/eip-template.md
@@ -81,6 +81,21 @@ TBD
 
 No backward compatibility issues found.
 
+## Chain-specific Considerations
+
+<!--
+
+  This section is optional.
+
+  All EIPs that require changes for non-mainnet chains must include a section describing these changes. This includes considerations that must be made for testnets, Gnosis chain, and EVM-compatible L2s.
+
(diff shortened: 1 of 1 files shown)


benaadams (approved):

rolfyone (approved):

Maks199010 (approved):

g11tech (approved):

SamWilsn:
Neither am I. Probably because it's the template, honestly.

This'll also require changes in EIP-1, and in eipw.

SamWilsn (changes requested):
Sorry, perhaps I wasn't clear. To merge this pull request, you'll also have to update the text of EIP-1 and the configuration for eipw.

Would probably be best to do a Call for Input for this as well.

felixthecat420 (approved):

PR #9474, opened 2025-03-11
Update EIP-7873: EOF - flag to include caller in Creator Contract

pdobacz:
It has been raised [here](https://github.com/ethereum/EIPs/pull/9391#issuecomment-2678451413) that the EOF Creator Contract would be a better default factory, if it allowed for the CALLER to be included in the salt (and newly created account's address) by default, only with the option to exclude it explicitly if needed.

Follow the thread linked above for details, but the rationale is that such Creator Contract is better versed to be useful over the long term, rather than just being a bootstrapping contract. Without this change, such bootstrap-only Creator Contract would need to be abandoned in practice for a more feature-rich factory. There are concerns that tooling will have a hard time to adapt to many competing factories, and as such, it makes sense to make the Creator Contract a sensible default from day 1.

The counterargument is that the simpler the Creator Contract is, the better, as its logic comes at a premium of extra attack vector, extra audit surface and general complexity

---

That being said, I did my best to represent the intended behavior in the pseudo code, but **I am very likely to have been very wrong**. This PR should be an invitation to discuss the pros and cons of the proposed change - thereby tagging @frangio @pcaversaccio @cameel @charles-cooper  @kuzdogan who have up to this point participated in the discussions.

But of course, if anyone can point out an error or an attack which this contract would be susceptible to, or just that it doesn't work at all :sweat_smile: , please LMK.

### DIFF
--- a/EIPS/eip-7873.md
+++ b/EIPS/eip-7873.md
@@ -168,7 +168,9 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the 
     if lt(size, 65) { revert(0, 0) }
 
     // copy tx_initcode_hash, salt, unsafe_flag and init_data to memory to hash.
-    // note that the unsafe_flag is included to prevent pre-image attacks on the two paths to calculate salt below.
+    // note that the unsafe_flag is included to prevent hash collisions between the two branches:
+    //   - unsafe_flag == 0, caller's address is appended by the code
+    //   - unsafe_flag == 1, caller's address is sent via init_data
     calldatacopy(0, 0, size)
 
     // mask out the 31 bytes that follow the flag in input data
--- a/EIPS/eip-7873.md
+++ b/EIPS/eip-7873.md
@@ -160,36 +160,42 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the 
 
     /// tx_initcode_hash and salt are 32 bytes wide, unsafe_flag is 1 byte wide, init_data can be any width.
     /// If unsafe_flag is 0x00, `CALLER` is included in the salt and impacts the target address of the created account.
-    /// If unsafe_flag is non-zero, `CALLER` is excluded. Deployment by any sender will result in same target address. 
+    /// If unsafe_flag is non-zero, `CALLER` is excluded, with 32 zero-bytes used instead. Deployment by any sender will result in same target address. 
     /// Deployments using non-zero unsafe_flag are susceptible to front-running.
 
     // init_data.length can be 0, but the first 65 bytes are mandatory
     let size := calldatasize()
     if lt(size, 65) { revert(0, 0) }
 
-    // copy tx_initcode_hash, salt, unsafe_flag and init_data to memory to hash.
-    // note that the unsafe_flag is included to prevent hash collisions between the two branches:
-    //   - unsafe_flag == 0, caller's address is appended by the code
-    //   - unsafe_flag == 1, caller's address is sent via init_data
-    calldatacopy(0, 0, size)
+    // copy tx_initcode_hash and salt to memory to hash.
+    // note that the unsafe_flag doesn't need to be included. If set to non-zero, 32 zero-bytes
+    // are used instead of the CALLER address bytes, which prevents hash collisions between
+    // deployments using unsafe_flag and those not using it.
+    calldatacopy(0, 0, 64)
 
     // mask out the 31 bytes that follow the flag in input data
     let unsafe_flag := byte(0, calldataload(64))
 
     if iszero(unsafe_flag) {
-        // Include sender into salt just after init_data
-        mstore(size, caller())
-        let final_salt := keccak256(0, size + 32)
+        // store caller in memory to hash, just after salt
+        mstore(64, caller())
     } else {
-        // Exclude sender from salt - susceptible to front-running
-        let final_salt := keccak256(0, size)
+        // exclude sender from salt - susceptible to front-running
+        // assuming no bytes were written to memory word at 64 and it's all zeros
     }
 
-    let tx_initcode_hash := calldataload(0)
     let init_data_size := sub(size, 65)
 
+    // copy init_data to memory to hash, just after caller (or zeros)
+    calldatacopy(96, 65, init_data_size)
+
+    // final_salt = keccak256(tx_initcode_hash | salt | caller_or_zeros | init_data)
+    let final_salt := keccak256(0, 96 + init_data_size)
+
+    let tx_initcode_hash := calldataload(0)
+
     // reuse init_data which has been already copied to memory above
-    let ret := txcreate(tx_initcode_hash, callvalue(), final_salt, 65, init_data_size)
+    let ret := txcreate(tx_initcode_hash, callvalue(), final_salt, 96, init_data_size)
 
     if iszero(ret) {
         let ret_data_size := returndatasize()
--- a/EIPS/eip-7873.md
+++ b/EIPS/eip-7873.md
@@ -155,10 +155,10 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the 
 
 ```solidity
 {
-    /// Takes [unsafe_flag][tx_initcode_hash][salt][init_data] as input,
+    /// Takes [tx_initcode_hash][salt][unsafe_flag][init_data] as input,
     /// creates contract and returns the address or failure otherwise.
 
-    /// unsafe_flag is 1 byte wide. tx_initcode_hash and salt are 32 bytes wide, init_data can be any width.
+    /// tx_initcode_hash and salt are 32 bytes wide, unsafe_flag is 1 byte wide, init_data can be any width.
     /// If unsafe_flag is 0x00, `CALLER` is included in the salt and impacts the target address of the created account.
     /// If unsafe_flag is non-zero, `CALLER` is excluded. Deployment by any sender will result in same target address. 
     /// Deployments using non-zero unsafe_flag are susceptible to front-running.
@@ -167,12 +167,12 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the 
     let size := calldatasize()
     if lt(size, 65) { revert(0, 0) }
 
-    // copy unsafe_flag, tx_initcode_hash, salt and init_data to memory to hash.
+    // copy tx_initcode_hash, salt, unsafe_flag and init_data to memory to hash.
     // note that the unsafe_flag is included to prevent pre-image attacks on the two paths to calculate salt below.
     calldatacopy(0, 0, size)
 
     // mask out the 31 bytes that follow the flag in input data
-    let unsafe_flag := calldataload(0) & (0xff << 248)
+    let unsafe_flag := calldataload(64) & (0xff << 248)
 
     if iszero(unsafe_flag) {
         // Include sender into salt just after init_data
@@ -183,7 +183,7 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the 
         let final_salt := keccak256(0, size)
     }
 
-    let tx_initcode_hash := calldataload(1)
+    let tx_initcode_hash := calldataload(0)
     let init_data_size := sub(size, 65)
 
     // reuse init_data which has been already copied to memory above
--- a/EIPS/eip-7873.md
+++ b/EIPS/eip-7873.md
@@ -172,11 +172,11 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the 
     calldatacopy(0, 0, size)
 
     // mask out the 31 bytes that follow the flag in input data
-    let unsafe_flag := calldataload(64) & (0xff << 248)
+    let unsafe_flag := byte(0, calldataload(64))
 
     if iszero(unsafe_flag) {
         // Include sender into salt just after init_data
-        mstore(msize(), caller())
+        mstore(size, caller())
         let final_salt := keccak256(0, size + 32)
     } else {
         // Exclude sender from salt - susceptible to front-running
--- a/EIPS/eip-7873.md
+++ b/EIPS/eip-7873.md
@@ -155,21 +155,39 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the 
 
 ```solidity
 {
-    /// Takes [tx_initcode_hash][salt][init_data] as input,
-    /// creates contract and returns the address or failure otherwise
+    /// Takes [unsafe_flag][tx_initcode_hash][salt][init_data] as input,
+    /// creates contract and returns the address or failure otherwise.
 
-    // init_data.length can be 0, but the first 2 words are mandatory
+    /// unsafe_flag is 1 byte wide. tx_initcode_hash and salt are 32 bytes wide, init_data can be any width.
+    /// If unsafe_flag is 0x00, `CALLER` is included in the salt and impacts the target address of the created account.
+    /// If unsafe_flag is non-zero, `CALLER` is excluded. Deployment by any sender will result in same target address. 
+    /// Deployments using non-zero unsafe_flag are susceptible to front-running.
+
+    // init_data.length can be 0, but the first 65 bytes are mandatory
     let size := calldatasize()
-    if lt(size, 64) { revert(0, 0) }
+    if lt(size, 65) { revert(0, 0) }
 
-    calldatacopy(0, 0, size)  // copy tx_initcode_hash, salt and init_data to memory to hash
-    let final_salt := keccak256(0, size)
+    // copy unsafe_flag, tx_initcode_hash, salt and init_data to memory to hash.
+    // note that the unsafe_flag is included to prevent pre-image attacks on the two paths to calculate salt below.
+    calldatacopy(0, 0, size)
 
-    let tx_initcode_hash := calldataload(0)
-    let init_data_size := sub(size, 64)
+    // mask out the 31 bytes that follow the flag in input data
+    let unsafe_flag := calldataload(0) & (0xff << 248)
+
+    if iszero(unsafe_flag) {
+        // Include sender into salt just after init_data
+        mstore(msize(), caller())
+        let final_salt := keccak256(0, size + 32)
+    } else {
+        // Exclude sender from salt - susceptible to front-running
+        let final_salt := keccak256(0, size)
+    }
+
+    let tx_initcode_hash := calldataload(1)
+    let init_data_size := sub(size, 65)
 
     // reuse init_data which has been already copied to memory above
-    let ret := txcreate(tx_initcode_hash, callvalue(), final_salt, 64, init_data_size)
+    let ret := txcreate(tx_initcode_hash, callvalue(), final_salt, 65, init_data_size)
 
     if iszero(ret) {
         let ret_data_size := returndatasize()


EIPS/eip-7873.md
```diff
@@ -155,21 +155,39 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
 
 ```solidity
 {
-    /// Takes [tx_initcode_hash][salt][init_data] as input,
-    /// creates contract and returns the address or failure otherwise
+    /// Takes [unsafe_flag][tx_initcode_hash][salt][init_data] as input,
```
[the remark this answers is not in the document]
pdobacz: good point, let me fix that and see if no hairy issues arise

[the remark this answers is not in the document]
pdobacz: Done 38d4a9d5, pls double check

EIPS/eip-7873.md
```diff
@@ -155,21 +155,39 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
+    let unsafe_flag := calldataload(64) & (0xff << 248)
+
+    if iszero(unsafe_flag) {
+        // Include sender into salt just after init_data
+        mstore(msize(), caller())
+        let final_salt := keccak256(0, size + 32)
+    } else {
+        // Exclude sender from salt - susceptible to front-running
+        let final_salt := keccak256(0, size)
+    }
```
frangio, line 184: Just an idea, this seems equivalent:

```suggestion
    if iszero(unsafe_flag) {
        // Include sender into salt just after init_data
        mstore(msize(), caller())
    }

    let final_salt := keccak256(0, msize())
```

EIPS/eip-7873.md
```diff
@@ -155,21 +155,39 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
+    // note that the unsafe_flag is included to prevent pre-image attacks on the two paths to calculate salt below.
+    calldatacopy(0, 0, size)
+
+    // mask out the 31 bytes that follow the flag in input data
+    let unsafe_flag := calldataload(64) & (0xff << 248)
+
+    if iszero(unsafe_flag) {
+        // Include sender into salt just after init_data
+        mstore(msize(), caller())
+        let final_salt := keccak256(0, size + 32)
```
frangio, line 180: I think there is a bug here: `msize()` is rounded up to multiple of 32, but `init_data` may not be a multiple, so this code results in some padding between `init_data` and sender, and since the hash uses `size` it does not actually include the full address in the final salt.

EIPS/eip-7873.md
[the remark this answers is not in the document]
frangio: @shemnon's suggestion above implies that `if lt(size, 65) { revert(0, 0) }` should be changed to `if lt(size, 32) { revert(0, 0) }`, so that all of `[tx_initcode_hash][salt][init_data]` become optional arguments with default zero/empty.

EIPS/eip-7873.md
```diff
@@ -155,21 +155,39 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
+    // init_data.length can be 0, but the first 65 bytes are mandatory
+    let size := calldatasize()
+    if lt(size, 65) { revert(0, 0) }
+
+    // copy tx_initcode_hash, salt, unsafe_flag and init_data to memory to hash.
+    // note that the unsafe_flag is included to prevent pre-image attacks on the two paths to calculate salt below.
+    calldatacopy(0, 0, size)
+
+    // mask out the 31 bytes that follow the flag in input data
+    let unsafe_flag := calldataload(64) & (0xff << 248)
```
frangio, line 175:
```suggestion
    let unsafe_flag := byte(0, calldataload(64))
```

EIPS/eip-7873.md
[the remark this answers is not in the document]
frangio: That's not sufficient though. If `32 < calldatasize < 65` you'd want to bump `size` up to 65 so that the caller is correctly appended.

EIPS/eip-7873.md
```diff
@@ -155,21 +155,39 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
+        let final_salt := keccak256(0, size)
+    }
 
     let tx_initcode_hash := calldataload(0)
-    let init_data_size := sub(size, 64)
+    let init_data_size := sub(size, 65)
 
     // reuse init_data which has been already copied to memory above
-    let ret := txcreate(tx_initcode_hash, callvalue(), final_salt, 64, init_data_size)
+    let ret := txcreate(tx_initcode_hash, callvalue(), final_salt, 65, init_data_size)
```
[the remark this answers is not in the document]
pdobacz: yeah, definitely pseudo-code. I knew this was maybe a stretch on the syntax, but don't want also to overthink it.

EIPS/eip-7873.md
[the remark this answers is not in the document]
pdobacz: I didn't take that suggestion went that far, that is, the caller would still need to explicitly send the 33 zero bytes (at a discount). That is, the advantage was more of a elegance standpoint. Re-reading now I think you're right and requiring only 32 bytes was the point.

This raises a question whether we want to rely on calldata behavior like that? It also kinda makes the whole thing more complex - as you've already kinda noticed. :thinking:

pdobacz:
Actually, I have an issue with this change - even with this, the Creator Contract will not provide a good default factory, as it will only work once for each sender. In order for subsequent deploys to work for the sender, they would need to provide unique salts. I'm not sure if we're not going down a false security (false UX advantage) path here.

And storing a map between CALLER address and nonce, and adding yet another flag, seems to bring a lot of additional complexity in, or would you say it is warranted?

frangio:
Absolutely would not add a nonce map here.

>  In order for subsequent deploys to work for the sender, they would need to provide unique salts.

Not sure I see the issue with this, can you elaborate?

frangio:
Ah I think you mean it won't be a good replacement for creation transactions because those do allow multiple deployments of the same code. It seems easy to fix that by generating a unique random salt by default unless the developer requests a "deterministic deployment", which is an option libraries already tend to have to enable deployment via CREATE2 factories.

EIPS/eip-7873.md
```diff
@@ -155,21 +155,39 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
 
-    calldatacopy(0, 0, size)  // copy tx_initcode_hash, salt and init_data to memory to hash
-    let final_salt := keccak256(0, size)
+    // init_data.length can be 0, but the first 65 bytes are mandatory
+    let size := calldatasize()
+    if lt(size, 65) { revert(0, 0) }
+
+    // copy tx_initcode_hash, salt, unsafe_flag and init_data to memory to hash.
+    // note that the unsafe_flag is included to prevent pre-image attacks on the two paths to calculate salt below.
+    calldatacopy(0, 0, size)
```
frangio, line 172: FWIW, I believe this implements @shemnon's suggestion.

```suggestion
    // revert if tx_initcode_hash not provided.
    if lt(size, 32) {
        revert(0, 0)
    }

    // copy parameters to memory to hash.
    // note that the unsafe_flag is included to prevent pre-image attacks on the two paths to calculate salt below.
    calldatacopy(0, 0, size)

    // increase size if salt or unsafe_flag not provided, defaulting to zero.
    if lt(size, 65) {
        size := 65
    }
```

pdobacz: I had it drafted similar, but did the "zero-padding-defaulting" on the `calldatacopy` step, i.e. I expand `size` one step earlier. More expensive but maybe safer.

Anyway, by "moot" you mean to keep it as is (no zero padding by default, require 65 bytes)? I think simpler is better.

EIPS/eip-7873.md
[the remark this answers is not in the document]
shemnon: Good point on the discouraging zero salt.  

Perhaps we should consider asking tooling that sends these deployments to use some function on the the sender's nonce for a salt unless explicitly specified?

shemnon:
For the salt issue, as I said in a nested comment I think we should encourage tooling like foundry and hardhat to use the sending accounts nonce (or some hash on that nonce) as the salt when the user does not specify a salt.  This puts the burden of nonced deployment onto the developer tooling, which is actually better set up to check for conclicts.

EIPS/eip-7873.md
frangio: Yes, let's just require 65 bytes. This comment should be resolved.

gumb0: agree this seems too complex without much benefit, I would require all arguments

EIPS/eip-7873.md
```diff
@@ -155,21 +155,41 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
+    //   - unsafe_flag == 1, caller's address is sent via init_data
+    calldatacopy(0, 0, size)
+
+    // mask out the 31 bytes that follow the flag in input data
+    let unsafe_flag := byte(0, calldataload(64))
+
+    if iszero(unsafe_flag) {
+        // Include sender into salt just after init_data
+        mstore(size, caller())
+        let final_salt := keccak256(0, size + 32)
```
chfast, line 182: This is insecure. Even if you include the address of the caller this can be front-runned by:
- setting unsafe_flag to 1,
- appending the caller's address to the input.

I think the better approach is to have a fixed placeholder of 32-bytes for the caller's address before the input data. If `unsafe_flag` is set leave the placeholder all zeros, otherwise write caller's address there.

frangio: Alternatively, always hash over `size + 32` but only write the sender in the last word if `unsafe_flag` is nonzero.

pdobacz: `unsafe_flag` is included in the hashing, so that front-running won't work (will land in different target address), and there's a comment laying out this exact reason (also raised on discord and explained).

> // note that the unsafe_flag is included to prevent hash collisions between the two branches

Am I maybe missing sth?

EIPS/eip-7873.md
```diff
@@ -155,21 +155,41 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
+    if lt(size, 65) { revert(0, 0) }
+
+    // copy tx_initcode_hash, salt, unsafe_flag and init_data to memory to hash.
+    // note that the unsafe_flag is included to prevent hash collisions between the two branches:
+    //   - unsafe_flag == 0, caller's address is appended by the code
+    //   - unsafe_flag == 1, caller's address is sent via init_data
+    calldatacopy(0, 0, size)
+
+    // mask out the 31 bytes that follow the flag in input data
+    let unsafe_flag := byte(0, calldataload(64))
```
pdobacz, line 177: I had another thought on this - maybe it's better to require the unsafe_flag be just one lowest bit of that byte, so that the other bits can be possibly repurposed for other flags? The contract is not upgradable (nor planned to be re-deployed by state change), but... maybe this tiny amount forward compatibility would be useful?

This of course opens another can of worms: should we revert on higher bits set?

EIPS/eip-7873.md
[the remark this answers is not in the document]
pdobacz: in the other thread we landed at a conclusion that we still want the contract to require a 65-byte input to proceed. So we're keeping the proposed ordering but just for the sake of elegance of starting the input with a non-zero argument of `tx_initcode_hash`

pdobacz:
Given that we've spent time discussing the execution of the change (adding a flag), and have insight into the complexity and issues we run into - what is the general feel on the general topic - feature rich, default and long-term vs simple?

I think the complexity is manageable, I expected it to be worse, even though there's still some questions to resolve. OTOH I feel it's rather unlikely that the Creator Contract will remain in use in the long term enough to warrant the work we're putting in now.

EIPS/eip-7873.md
```diff
@@ -155,21 +155,41 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
 {
-    /// Takes [tx_initcode_hash][salt][init_data] as input,
-    /// creates contract and returns the address or failure otherwise
+    /// Takes [tx_initcode_hash][salt][unsafe_flag][init_data] as input,
+    /// creates contract and returns the address or failure otherwise.
 
-    // init_data.length can be 0, but the first 2 words are mandatory
-    let size := calldatasize()
-    if lt(size, 64) { revert(0, 0) }
+    /// tx_initcode_hash and salt are 32 bytes wide, unsafe_flag is 1 byte wide, init_data can be any width.
```
pdobacz, line 161: New topic: shouldn't we ABI encode the arguments and make the API more future-proof? The reason to do this would be to achieve an open standard for inputs to an EOF factory, so that any ERC-driven factories can adhere to and extend.

This would have the benefit for the tooling, which would (maybe) not need to have an entirely separate path to encode arguments for the Creator Contract and the other factories. E.g. it could abstract the choice of factory and some "extra" params out and leave it up to the user.

Now, coming up with such a standard future-proof API is a huge task, so I'm playing devil's advocate here - if we can't realistically offer that, it might be less likely Creator Contract will be used long-term.

---

A minimalist hack **just to illustrate the idea** would be to have:

```
[tx_initcode_hash][salt][unsafe_flag][abi_encoded_extra:bytes][init_data]
```

where `abi_encoded_extra:bytes` would be assumed to hold ABI encoded `bytes` holding inside opaque bytes with extra arguments to any future extensions. `abi_encoded_extra` would be required, but otherwise ignored by the Creator Contract (except decoding the length of bytes to know how much input to require and ignore). `abi_encoded_extra` would be exempt from salt hashing. Vanilla usage would be to send 32 zero bytes meaning an empty `bytes`.

gumb0:
What is the motivation for keeping unsafe option and not making every deployment non-frontrunnable? Saving gas for those that don't care about front-running?

EIPS/eip-7873.md
gumb0: I also thought about ABI, and convenience for tooling might be a good argument in favor (also, would it be easier to call from Solidity?). But devising some general API for factories is going too far, definitely out of scope.

chfast:
> What is the motivation for keeping unsafe option and not making every deployment non-frontrunnable? Saving gas for those that don't care about front-running?

Universal deployments on compatible chains without maintaining a private key.

EIPS/eip-7873.md
pdobacz: > I also thought about ABI, and convenience for tooling might be a good argument in favor (also, would it be easier to call from Solidity?). But devising some general API for factories is going too far, definitely out of scope.

What about the "flipped" argument that not having a future-proof ABI/API contradicts the notion that Creator Contract will be used in the long term (which we're basing off of to introduce `unsafe_flag` in the first place here)?

frangio:
> OTOH I feel it's rather unlikely that the Creator Contract will remain in use in the long term enough to warrant the work we're putting in now.

What makes you think the Creator Contact will not remain in use with this change? As far as I can tell it is quite feature complete given the requirements for factories we've seen. The only thing it doesn't provide is deterministic addresses independent of `init_code_hash` and `init_data`, but I believe this is a small minority of use cases.

EIPS/eip-7873.md
frangio: No that sounds right... If a tx is "frontrun" as @chfast suggests, it should result in a different deployed address.

EIPS/eip-7873.md
frangio: I really don't think that non-standard encoding will be an issue for tooling. A shared abstraction covering other kinds of factories can be built even if different factories require different encoding.

EIPS/eip-7873.md
```diff
@@ -155,21 +155,41 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
-    calldatacopy(0, 0, size)  // copy tx_initcode_hash, salt and init_data to memory to hash
-    let final_salt := keccak256(0, size)
+    // init_data.length can be 0, but the first 65 bytes are mandatory
+    let size := calldatasize()
+    if lt(size, 65) { revert(0, 0) }
+
+    // copy tx_initcode_hash, salt, unsafe_flag and init_data to memory to hash.
+    // note that the unsafe_flag is included to prevent hash collisions between the two branches:
+    //   - unsafe_flag == 0, caller's address is appended by the code
+    //   - unsafe_flag == 1, caller's address is sent via init_data
```
frangio, line 173: This is a little weird. Nothing forces the caller to include their address in `init_data`. In most cases that use non-zero unsafe flag the deployment should be totally independent of the caller. Additionally, the flag can be any non-zero value.

EIPS/eip-7873.md
frangio: The forward compatible option would be to force `unsafe_flag` to be `0x00` or `0x01` and revert on other values.

I also find that tempting. Using RJUMPV it would be very simple to express in the code. So I'd be in favor of this change.

pdobacz: > The forward compatible option would be to force `unsafe_flag` to be `0x00` or `0x01` and revert on other values.
> 
> I also find that tempting. Using RJUMPV it would be very simple to express in the code. So I'd be in favor of this change.

Ah, right, if we revert on unsafe_flag > 0x01 then no need to worry about hashing. I find this an extra advantage of such restricted approach - otherwise unsafe_flag's higher 7 bits become de-facto salt, which is a bit weird.

@gumb0 can you elaborate on your reservations towards this (you've :-1:ed )?

gumb0: To me it feels like overthinking and unnecessary logic complication for speculative benefit. Current code with branching on `iszero(unsafe_flag)` looks fine, for hashing I would do Pawel's suggestion to hash 0 address instead of caller for unsafe mode and don't hash the flag.

EIPS/eip-7873.md
gumb0: I think this comment means that without hashing `unsafe_flag` the deployment with `unsafe_flag == 0` collides with deployment with `unsafe_flag == 1` and caller's address in `init_data`

frangio: Ok, that makes more sense.

EIPS/eip-7873.md
pdobacz: I pushed the proposed alternative to 0b55a4f5. Note that this is slightly more complex actually, as long as we keep the earlier suggestion to place `unsafe_flag` as third argument. We can debate whether or not to keep it there.

Overall it has a nice property of cutting out the address malleability via various non-zero values of `unsafe_flag` byte...

EIPS/eip-7873.md
pdobacz: Resolved, see other thread and let's continue there

EIPS/eip-7873.md
```diff
@@ -155,21 +155,47 @@ At the start of the block in which [EIP-3540](./eip-3540.md) activates, set the
...
+
+    // mask out the 31 bytes that follow the flag in input data
+    let unsafe_flag := byte(0, calldataload(64))
+
+    if iszero(unsafe_flag) {
+        // store caller in memory to hash, just after salt
+        mstore(64, caller())
+    } else {
+        // exclude sender from salt - susceptible to front-running
+        mstore(64, 0)
```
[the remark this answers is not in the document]
pdobacz: Just explicitness (and me being forgetful). I'll switch the mstore to just a suitable comment

EIPS/eip-7873.md
pdobacz: @gumb0 @chfast LMK if you're happy with the latest revision and :heavy_check_mark: if yes, so we can merge this if there's also consensus on the overall direction.

chfast: I have more ideas in https://github.com/ipsilon/eof/issues/172#issuecomment-2732366390, but maybe better to make another PR to present my new ideas :)

So I think this is fine to merge.

chfast:
BTW, is this interface compatible with `abi.encodePacked()`?

frangio:
Yes! This is exactly what one gets by doing `abi.encodePacked(bytes32, bytes32, bool, bytes)`. That's nice.

PR #9493, opened 2025-03-18
Add EIP: eth_config JSON-RPC Method

shemnon:
Initial draft of the EIP

### DIFF
--- a/EIPS/eip-0000.md
+++ b/EIPS/eip-0000.md
@@ -1,5 +1,5 @@
 ---
-eip: XXXX
+eip: 0000
 title: eth_config JSON-RPC Method
 description: A JSON-RPC method that describes the configuration of the current and next fork
 author: Danno Ferrin (@shemnon)
--- a/EIPS/eip-0000.md
+++ b/EIPS/eip-0000.md
@@ -23,7 +23,9 @@ The most significant occurrence was during the Pectra activation on the Holešky
 By providing an RPC method that allows clients to report key configuration variables before the next hard fork, operations teams can gain greater confidence that clients are correctly configured and prepared for upcoming forks.
 
 ### Target Audience and Use Cases
+
 This method is intended for node operators, validator teams, and network monitoring tools to verify client readiness for upcoming forks. Use cases include:
+
 - Automated pre-fork validation scripts comparing `eth_config` outputs across nodes.
 - Manual checks by validator operators to ensure alignment with fork specifications.
 - Debugging by client developers to identify configuration mismatches.
@@ -142,6 +144,7 @@ Clients supporting pre-Cancun forks MAY return partial or non-standard configura
 ### Sample Configs
 
 Hoodi Prague Config
+
 ```JSON
 {
   "activationTime": 1742999832,
@@ -181,6 +184,7 @@ Hoodi Prague Config
 ```
 
 Hoodi Cancun Config
+
 ```JSON
 {
   "activationTime": 0,
--- a/EIPS/eip-0000.md
+++ b/EIPS/eip-0000.md
@@ -71,7 +71,9 @@ Future forks may add, adjust, remove, or update fields. The respective changes M
 
 #### `activationTime`
 
-The fork activation timestamp, represented as a JSON number in Unix epoch seconds (UTC). For the "current" configuration, this reflects the actual activation time; for "next," it is the scheduled time if known, or null if unscheduled.
+The fork activation timestamp, represented as a JSON number in Unix epoch seconds (UTC). For the "current" configuration, this reflects the actual activation time; for "next," it is the scheduled time.
+
+Activation time is required. If a fork is activated at genesis the value `0` is used. If the fork is not scheduled to be activated or its activation time is unknown it should not be in the rpc results.
 
 #### `blobsSchedule`
 
@@ -79,11 +81,13 @@ The blob configuration parameters for the specific fork, as defined in the genes
 
 #### `chainId`
 
-The chain ID of the current network, presented as a string with an unsigned `0x`-prefixed hexadecimal number, with all leading zeros removed. This specification does not support chains without a chain ID or with a chain ID of zero.
+The chain ID of the current network, presented as a string with an unsigned `0x`-prefixed hexadecimal number, with all leading zeros removed. This specification does not support chains without a chain ID or with a chain ID of zero. 
+
+For purposes of canonicalization this value must always be a string.
 
 #### `precompiles`
 
-A representation of the active precompile contracts for the fork. If a precompile is replaced by an on-chain contract, it is not included.
+A representation of the active precompile contracts for the fork. If a precompile is replaced by an on-chain contract, or removed, then it is not included.
 
 This is a JSON object where the members are the 20-byte `0x`-prefixed hexadecimal addresses of the precompiles (with zeros preserved), and the values are agreed-upon names for each contract, typically specified in the EIP defining that contract.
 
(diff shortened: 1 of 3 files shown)


EIPS/eip-config.md
```diff
@@ -0,0 +1,302 @@
...
+
+Future forks MUST define the list of system contracts in their meta-EIPs.
+
+## Rationale
+
+### Why Enumerate Precompiles? (And in General, Why Track a Particular Config Item?)
+
+The purpose of this specification is to enable nodes to advertise, prior to a fork, that they have the correct configurations loaded and ready. Past testnet and mainnet forks have revealed clients with incorrect precompile sets, chain IDs, deposit contract addresses, and other configuration errors.
+
+For precompiles in particular there has been discussion about removing or replaceing precomiles in future forks, so a full enumeration of precompiles will reflect removal.
```
marioevz, line 107:
```suggestion
For precompiles in particular there has been discussion about removing or replacing precompiles in future forks, so a full enumeration of precompiles will reflect removal.
```

marioevz (commented):
Looks really good to me. I would like to give it a try to generate the config in EEST to see how easily it can be implemented and distributed for clients or, ideally, generate an automatic hive test to be added to the dashboard.

EIPS/eip-config.md
```diff
@@ -0,0 +1,303 @@
...
+
+Future forks may add, adjust, remove, or update fields. The respective changes MUST be defined in their respective meta-EIPs.
+
+#### `activationTime`
+
+The fork activation timestamp, represented as a JSON number in Unix epoch seconds (UTC). For the "current" configuration, this reflects the actual activation time; for "next," it is the scheduled time if known, or null if unscheduled.
+
+#### `blobsSchedule`
+
+The blob configuration parameters for the specific fork, as defined in the genesis file. This is a JSON object with three members —  `basefeeUpdateFraction`, `max`, and `target` — all represented as JSON numbers.
```
jochem-brouwer, line 76:
```suggestion
The blob configuration parameters for the specific fork, as defined in the genesis file. This is a JSON object with three members —  `baseFeeUpdateFraction`, `max`, and `target` — all represented as JSON numbers.
```

jochem-brouwer (commented):
One style nit after first read, looks good so far! However I think we will come up with some extra params and some situations which might happen in the future. 

What would also be great is if this becomes a canonical thing, then these `eth_config` relevant info (like params and their definitions) should be added to Core EIPs as well :)

EIPS/eip-config.md
```diff
@@ -0,0 +1,303 @@
...
+    "0x0000000000000000000000000000000000000008": "BN256_PAIRING",
+    "0x0000000000000000000000000000000000000009": "BLAKE2F",
+    "0x000000000000000000000000000000000000000a": "KZG_POINT_EVALUATION",
+    "0x000000000000000000000000000000000000000b": "BLS12_G1ADD",
+    "0x000000000000000000000000000000000000000c": "BLS12_G1MULTIEXP",
+    "0x000000000000000000000000000000000000000d": "BLS12_G2ADD",
+    "0x000000000000000000000000000000000000000e": "BLS12_G2MULTIEXP",
+    "0x000000000000000000000000000000000000000f": "BLS12_PAIRING",
+    "0x0000000000000000000000000000000000000010": "BLS12_MAP_FIELD_TO_CURVE",
+    "0x0000000000000000000000000000000000000011": "BLS12_MAP_FIELD_TO_CURVE"
```
jochem-brouwer, line 171: Ah another one 0x10 and 0x11 are the same precompile :thinking:

EIPS/eip-config.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Abstract
+
+This document describes an RPC method that provides node-relevant configuration data for the current and next fork.
+
+## Motivation
+
+Throughout Ethereum's history, there have been multiple instances where a client was not correctly configured for an upcoming hard fork, causing it to fall out of consensus when the fork boundary was crossed. Most incidents have been minor, such as a single client forking the chain in proof-of-work or having its blocks orphaned in proof-of-stake.
+
+The most significant occurrence was during the Pectra activation on the Holesky testnet. Four out of six clients on the network had an incorrect configuration for the deposits contract. Instead of being orphaned, the incorrect chain was justified, and the side effects persist on Holesky even after reaching finality.
```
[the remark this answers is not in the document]
shemnon: fixed.  Samples migrated to Hoodi

EIPS/eip-config.md
shemnon: Fixed.

Also, some other BLS names did not align with EIP names, updated those.

EIPS/eip-0000.md
```diff
@@ -0,0 +1,307 @@
...
+    "WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS": "0x00000961ef480eb55e80d19ad83579a64c007002"
+  }
+}
+```
+
+Hoodi Cancun Config
+
+```JSON
+{
+  "activationTime": 0,
```
bomanaps, line 190: We should replace `activationTime: 0` with `null` to indicate unknown or unscheduled fork activation times, avoiding ambiguity with the Unix epoch and aligning with Ethereum's timestamp conventions, or possibly add a comment to that.

EIPS/eip-0000.md
```diff
@@ -0,0 +1,307 @@
...
+
+Nested structures are easier to read, while flat structures are easier to merge. Since this specification uses full configurations for the current and next forks, merging is unnecessary, making readability the priority.
+
+### Serving Data Not Specified in genesis.json
+
+Some reported values are specification-level constants, which many clients do not include in their configuration files. However, certain EIP constants (e.g., the deposit contract) have become variables in testnets, necessitating their inclusion.
+
+### JSON as Config Format
+
+JSON was chosen for its ubiquity, machine and human readability, and the existence of a standardized canonical form via [RFC-8785](https://www.rfc-editor.org/rfc/rfc8785). YAML lacks a standard canonical form. No Ethereum software uses XML for configuration, and adopting it would increase every client’s library size. An invented format is possible but would require definition and standardization within this specification.
```
bomanaps, line 130: How should implementations reconcile the requirement for RFC-8785 canonical JSON (which specifies numbers in simplest form) with the representation of chain IDs as hex strings with '0x' prefixes in the configuration objects?

shemnon: What the RFC is talking about is unneeded decimals or exponential notation.  `10` vs. `10.0` vs `1e2`.  For the purpose of canonicalization he chain ID is a string.

EIPS/eip-0000.md
shemnon: Geth and Besu at least differentiate between null and 0.  Null means that the for is not to be activated directly and zero means it is activated at genesis.

If there is an unknown or unscheduled fork there is no configuration object.  The current fork is known and the next fork would then be represented by a `null` configuration object.

bomanaps: Can we add a comment close to that in order to clarify this?

EIPS/eip-0000.md
shemnon: Added a comment to `chainId` to re-enforce it is a string

https://github.com/ethereum/EIPs/pull/9493/commits/5c664013dc6d2a53a476c5926789fc47d45aac15

g11tech (approved):

EIPS/eip-7910.md
```diff
@@ -0,0 +1,311 @@
...
+
+For purposes of canonicalization this value must always be a string.
+
+#### `precompiles`
+
+A representation of the active precompile contracts for the fork. If a precompile is replaced by an on-chain contract, or removed, then it is not included.
+
+This is a JSON object where the members are the 20-byte `0x`-prefixed hexadecimal addresses of the precompiles (with zeros preserved), and the values are agreed-upon names for each contract, typically specified in the EIP defining that contract.
+
+For Cancun, the contract names are (in order): `ECREC`, `SHA256`, `RIPEMD160`, `ID`, `MODEXP`, `BN256_ADD`, `BN256_MUL`, `BN256_PAIRING`, `BLAKE2F`, `KZG_POINT_EVALUATION`.
```
rubo, line 94: Unfortunately, I'm late to the party, but these `BN256_` things should be `BN254_` instead. We have three different names for the same thing -- `alt_bn128`, `BN254`, and `BN256` (why?) -- and that's pretty confusing and inconsistent.

shemnon: Post a PR and some link as to why it's the proper name.  The PR hasn't gone into last call neither has it gone final, so there is still time.

The original EIPs just call it ADD/ECADD, MUL/ECMUL, [EIP-196](https://eips.ethereum.org/EIPS/eip-196) or no name ([EIP-197](https://eips.ethereum.org/EIPS/eip-197)).  But both call it alt_bn128 in the tiles.

Before calling it BN254 I'd want to see some evidence of usage.

rubo: [Here you go](https://github.com/ethereum/EIPs/pull/10029)

OBITOONDEADO1MZ (approved):

PR #9494, opened 2025-03-19
Update EIP-7873: Add ASE compatibility to Creator Contract

chfast:
This changes how the caller address is included in the final salt in the Creator Contract: it is now left-padded with zeros to 32-bytes to make it compatible with Address Space Expansion.
Moreover, if unsafe flag is on the 32-byte placeholder for the caller address is filled with all 1s. Fix fixes the case when the caller address is 0 address.


gumb0:
> This changes how the caller address is included in the final salt in the Creator Contract: it is now left-padded with zeros to 32-bytes to make it compatible with Address Space Expansion.

It was already left-padded, wasn't it?

chfast:
> It was already left-padded, wasn't it?

Oh yes, I confused myself.

PR #9500, opened 2025-03-20
Update EIP-7702: Update eip-7702.md

pochenai:
> Without EIP-7702, this situation **can only ever** arise in the **topmost** execution layer of a transaction.

However, proxy contracts can also lead to cases where `msg.sender == tx.origin` besides the topmost execution layer of a transaction, meaning this was never a strict invariant. EIP-7702 further expands such scenarios, affecting smart contracts that rely on require(msg.sender == tx.origin) checks.


SamWilsn:
> proxy contracts can also lead to cases where `msg.sender == tx.origin` besides the topmost execution layer of a transaction

Can you elaborate?

PR #9502, opened 2025-03-21
Update EIP-3540: Change kind_data from 0x04 to 0xff

pdobacz:
Align with eof-devnet-1 change https://github.com/ipsilon/eof/issues/173. Goes in hand with https://github.com/ethereum/execution-spec-tests/pull/1339


kuzdogan (commented):
Just, should the `metadata_section` be `0x04`? I guess keeping it `0x05` and skipping the `0x04` value is a bit better for backwards compatibility?

kuzdogan:
I also don't have a preference. Just wanted to raise this.

PR #9508, opened 2025-03-21
Update EIP-7002: Add system call gas aspects and directive for handling failures

somnathb1:


EIPS/eip-7002.md
```diff
@@ -706,6 +706,10 @@ Note: the above code reverts if the fee is too high, the fee can change signific
...
 
+### System Call failure
+
+Although the likelihood of a failed system call to SYSTEM_ADDRESS is extremely low, the behavior in such cases is well-defined: the block is marked as invalid. However, if the failure results from processing a transaction within the block, the public mempool may still retain the transaction even after the block is invalidated. This can result in the offending transaction being included again, potentially causing one or more subsequent slots to go without valid blocks. To mitigate this, we recommend that the block producer implementation shuffle their transaction set to increase the chances of producing a valid block, without the offending transaction(s). The block producer implementation and/or the mempool should be aware of system call failure scenarios to enable this behavior.
```
jochem-brouwer, line 711: Same here, `SYSTEM_ADDRESS` does not (should not) have code, but `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS` has the relevant code here.

EIPS/eip-7002.md
```diff
@@ -226,7 +226,7 @@ def reset_withdrawal_requests_count():
     sstore(WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS, WITHDRAWAL_REQUEST_COUNT_STORAGE_SLOT, 0)
 ```
 
-Each withdrawal request must appear in the EIP-7685 requests list in the order they are returned from `dequeue_withdrawal_requests()`.
+Each withdrawal request must appear in the EIP-7685 requests list in the order they are returned from `dequeue_withdrawal_requests()`. If, however, at the end of processing an execution block the System Call to the `SYSTEM_ADDRESS` contract fails, the block must be considered invalid.
```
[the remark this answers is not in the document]
somnathb1: Yes, thanks!

mkalinin:
Looks good to me in general. A question: what are potential ways for a system call to fail? If the call is anyhow parametrized then we must specify those parameters in the EIP to make EL clients agree on them and avoid potential splits

somnathb1:
Currently it's not parameterized. System call can fail if
- The contract is bad or a transaction triggered an unexpected behaviour
- The contract/syscall runs out of gas (30M, say)
- Erroneous/Differing EL implementation - currently we don't guard against this because it's trivial

> Looks good to me in general. A question: what are potential ways for a system call to fail? If the call is anyhow parametrized then we must specify those parameters in the EIP to make EL clients agree on them and avoid potential splits

mkalinin:
> The contract/syscall runs out of gas (30M, say)

30M sounds like a parameter. I think it worth adding it to the EIP

somnathb1:
> > The contract/syscall runs out of gas (30M, say)
> 
> 30M sounds like a parameter. I think it worth adding it to the EIP

It's only the chain's block gas limit. And yes, we should specify what the System Call looks like in terms of gas limit.

mkalinin:
> It's only the chain's block gas limit. And yes, we should specify what the System Call looks like in terms of gas limit.

Right, i suppose a system call doesn’t count into gas used by block transactions? Why should it be limited by 30M?

somnathb1:
> Right, i suppose a system call doesn’t count into gas used by block transactions? Why should it be limited by 30M?
- An upper ceiling to avoid block delay
- A common theme in system calls
- We can get rid of that, but it may conflict with how it's implemented in practice in EL clients

mkalinin:
> We can get rid of that, but it may conflict with how it's implemented in practice in EL clients

I think we should rather make it explicit in all EIPs with gas limited system calls

somnathb1:
> I think we should rather make it explicit in all EIPs with gas limited system calls

Added the definitions, and also added invalidate condition for when there is nothing deployed at the `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS`. This has the pro that if clients have differing configs, this will cause a stir. But the downside is a chain cannot operate without this.

I.m.o the post-pectra readiness always involves deploying this address. And absence of the contract that has associated system calls should not be considered a valid scenario or be "failed silently" with

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
-Specifically, the functionality is defined in pseudocode as the function `read_withdrawal_requests()`:
+Each withdrawal request must appear in the EIP-7685 requests list in the exact order returned by dequeue_withdrawal_requests().
+
+Additionally, the system call and subsequent block processing **MUST** conform to the following:
+
+* The call has a dedicated gas limit of `30_000_000`.
+* Gas consumed by this call does not count against the block’s overall gas usage.
+* Both the gas limit assigned to the call and the gas consumed **MUST** be excluded from any checks against the block’s gas limit.
+* The call does not follow [EIP-1559](./eip-1559.md) fee burn semantics—value **MUST NOT** be transferred as part of this call.
+* If there is no code at `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS`, the corresponding block **MUST** be marked invalid.
```
jochem-brouwer, line 161: The problem here is (I think this was also raised by @marioevz) that if there is no code at the fork block, then the block is thus invalid (at the fork block). Since now every block which is being produced is invalid, it is impossible to deploy the withdrawal system contract.

The suggestion which I would take here is to just follow the path as in the normal call, except there is no code at the address. Therefore, the returndata is the empty bytes, and therefore also the request data itself is empty, so it will not alter the requestsHash but it will ensure that the block is valid and it is possible to deploy the withdrawals contract.

somnathb1: Pectra (or EIP-7002 specifically) should not be activated before being Pectra (or, here, EIP-7002) ready. This is not a problem, it's deliberate. 

As i highlighted on the call, one way to then achieve chain continuation is pushing the timestamp of the fork block, deploy the contract and continue. The chain would stop on the pre-fork block, ready for this change

jochem-brouwer: Right, sorry, you indeed raised this point during the call, I forgot.

I'm not sure if this process works. So, first we set fork at time X but do not recognize the system contract is not deployed. Then, at time X all blocks produced are invalid. So now we re-coordinate a fork time later for clients, we deploy this new fork time (how does this work in practice? Nodes will get un-stuck not all at the same time, but they will get un-stuck if they now update the client to the later fork time X). So now we have a chain which rolls forward, we deploy the system contract, and then we succesfully fork (this could work but I'm not sure what would happen if clients get un-stuck piece-by-piece? Is that a problem?)

jochem-brouwer: Actually I think it would be nice if you put this point somewhere, since this is a super good point, it only allows the chain to move forward once the contract is deployed and this must happen at the fork block if it isn't there (I did not get this directly from this point, I assumed if the chain gets stuck it would just be doomed unless the fork would be rescheduled, which is not the case)

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
 * The contract's queue is updated based on withdrawal requests dequeued and the withdrawal requests queue head/tail are reset if the queue has been cleared (`dequeue_withdrawal_requests()`)
 * The contract's excess withdrawal requests are updated based on usage in the current block (`update_excess_withdrawal_requests()`)
 * The contract's withdrawal requests count is reset to 0 (`reset_withdrawal_requests_count()`)
 
-Specifically, the functionality is defined in pseudocode as the function `read_withdrawal_requests()`:
+Each withdrawal request must appear in the EIP-7685 requests list in the exact order returned by dequeue_withdrawal_requests().
+
+Additionally, the system call and subsequent block processing **MUST** conform to the following:
+
+* The call has a dedicated gas limit of `30_000_000`.
```
mkalinin, line 157:
```suggestion
* The call **MUST** have a dedicated gas limit of `30_000_000`.
```

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
+* Gas consumed by this call does not count against the block’s overall gas usage.
```
mkalinin, line 158:
```suggestion
* Gas consumed by this call **MUST NOT** count against the block’s overall gas usage.
```

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
+* Both the gas limit assigned to the call and the gas consumed **MUST** be excluded from any checks against the block’s gas limit.
+* The call does not follow [EIP-1559](./eip-1559.md) fee burn semantics—value **MUST NOT** be transferred as part of this call.
+* If there is no code at `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS`, the corresponding block **MUST** be marked invalid.
+* If the call to the contract fails or returns an error, the block **MUST** be invalidated
```
mkalinin, line 162:
```suggestion
* If the call to the contract fails or returns an error, the block **MUST** be invalidated.
```

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
 
-Specifically, the functionality is defined in pseudocode as the function `read_withdrawal_requests()`:
+Each withdrawal request must appear in the EIP-7685 requests list in the exact order returned by dequeue_withdrawal_requests().
+
+Additionally, the system call and subsequent block processing **MUST** conform to the following:
+
+* The call has a dedicated gas limit of `30_000_000`.
+* Gas consumed by this call does not count against the block’s overall gas usage.
+* Both the gas limit assigned to the call and the gas consumed **MUST** be excluded from any checks against the block’s gas limit.
+* The call does not follow [EIP-1559](./eip-1559.md) fee burn semantics—value **MUST NOT** be transferred as part of this call.
```
mkalinin, line 160:
```suggestion
* The call does not follow [EIP-1559](./eip-1559.md) fee burn semantics — value **MUST NOT** be transferred as part of this call.
```

EIPS/eip-7002.md
somnathb1: Already says  `**MUST** conform` above, this line is just descriptive.

mkalinin: I would prefer to use MUST NOT instead of “does not". But is this is just my preference and otherwise it reads clear then i am happy to leave “does not” as is

somnathb1: We can leave "does not" as is, it's very clear and impossible to confuse with any hint of optionality in this context.

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
+At the end of processing any execution block starting from the `FORK_BLOCK` (i.e. after processing all transactions and after performing the block body withdrawal requests validations), the implementation **MUST** invoke a call to the contract at `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS` as `SYSTEM_ADDRESS` with no calldata. This invocation triggers the following behavior:
 
 * The contract's queue is updated based on withdrawal requests dequeued and the withdrawal requests queue head/tail are reset if the queue has been cleared (`dequeue_withdrawal_requests()`)
 * The contract's excess withdrawal requests are updated based on usage in the current block (`update_excess_withdrawal_requests()`)
 * The contract's withdrawal requests count is reset to 0 (`reset_withdrawal_requests_count()`)
 
-Specifically, the functionality is defined in pseudocode as the function `read_withdrawal_requests()`:
+Each withdrawal request must appear in the EIP-7685 requests list in the exact order returned by `dequeue_withdrawal_requests()`.
+
+Additionally, the system call and subsequent block processing **MUST** conform to the following:
```
lightclient, line 155: What do you mean by "subsequent block processing"?

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
 
 ##### System Call
 
-At the end of processing any execution block starting from the `FORK_BLOCK` (i.e. after processing all transactions and after performing the block body withdrawal requests validations), call the contract as `SYSTEM_ADDRESS` and perform the following:
+At the end of processing any execution block starting from the `FORK_BLOCK` (i.e. after processing all transactions and after performing the block body withdrawal requests validations), the implementation **MUST** invoke a call to the contract at `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS` as `SYSTEM_ADDRESS` with no calldata. This invocation triggers the following behavior:
```
lightclient, line 147:
```suggestion
At the end of processing any execution block starting from the `FORK_BLOCK` (i.e. after processing all transactions and after performing the block body withdrawal requests validations), call `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS` as `SYSTEM_ADDRESS` with no calldata. The invocation triggers the following:
```

Not really using RFC 2119 vocabulary in this EIP so should keep the style the same.

EIPS/eip-7002.md
lightclient: also you can just copy to text from 4788 so the behavior and style of the eip is identical:

https://github.com/ethereum/EIPs/blob/c6bc36304bdef3ddce9ede45f0c7cb5901bb7e07/EIPS/eip-4788.md?plain=1#L238-L243

lightclient (commented):
change generally LGTM, left some nits

do we also need to update 7251?

EIPS/eip-7002.md
somnathb1: > What do you mean by "subsequent block processing"?

Remaining part of processing of the block in which the call is made. There are post block generation and validation checks.

If it's not clear, because of the word "subsequent", we can say "and the corresponding block processing"

somnathb1: > also you can just copy to text from 4788 so the behavior and style of the eip is identical:

It is mostly copied from there, yes, with some changes

EIPS/eip-7002.md
somnathb1: That's true, but for the sake of additional clarification, it is okay. Do you have a stronger reason to fall back to the more informal language here?

somnathb1:
> change generally LGTM, left some nits
> 
> do we also need to update 7251?

Planning to create a cc PR for 7251 after this one gets argued about and merged.

EIPS/eip-7002.md
somnathb1: Many of the `**MUST**`'s were also suggested by @mkalinin . I am okay with either. Just want to be as clear as possible. If it is indicated that less stricter language is also as clear, we can go for that

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
+At the end of processing any execution block starting from the `FORK_BLOCK` (i.e. after processing all transactions and after performing the block body withdrawal requests validations), call `WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS` as `SYSTEM_ADDRESS` with no calldata. The invocation triggers the following:
 
 * The contract's queue is updated based on withdrawal requests dequeued and the withdrawal requests queue head/tail are reset if the queue has been cleared (`dequeue_withdrawal_requests()`)
 * The contract's excess withdrawal requests are updated based on usage in the current block (`update_excess_withdrawal_requests()`)
 * The contract's withdrawal requests count is reset to 0 (`reset_withdrawal_requests_count()`)
 
-Specifically, the functionality is defined in pseudocode as the function `read_withdrawal_requests()`:
+Each withdrawal request must appear in the EIP-7685 requests list in the exact order returned by `dequeue_withdrawal_requests()`.
+
+Additionally, the system call and subsequent block processing **MUST** conform to the following:
```
somnathb1, line 155:
```suggestion
Additionally, the system call and the processing of that block must conform to the following:
```

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
 * The contract's withdrawal requests count is reset to 0 (`reset_withdrawal_requests_count()`)
 
-Specifically, the functionality is defined in pseudocode as the function `read_withdrawal_requests()`:
+Each withdrawal request must appear in the EIP-7685 requests list in the exact order returned by `dequeue_withdrawal_requests()`.
+
+Additionally, the system call and the processing of that block must conform to the following:
+
+* The call has a dedicated gas limit of `30_000_000`.
+* Gas consumed by this call does not count against the block’s overall gas usage.
+* Both the gas limit assigned to the call and the gas consumed **MUST** be excluded from any checks against the block’s gas limit.
```
somnathb1, line 159:
```suggestion
* Both the gas limit assigned to the call and the gas consumed is excluded from any checks against the block’s gas limit.
```

EIPS/eip-7002.md
```diff
@@ -144,13 +144,24 @@ When the input to the contract is length zero, interpret this as a get request f
...
+* The call does not follow [EIP-1559](./eip-1559.md) fee burn semantics — value **MUST NOT** be transferred as part of this call.
```
somnathb1, line 160:
```suggestion
* The call does not follow [EIP-1559](./eip-1559.md) fee burn semantics — no value to be transferred as part of this call.
```

mkalinin (approved):

PR #9523, opened 2025-03-24
Add EIP: SSZ ProgressiveByteList

etan-status:
Move SSZ ProgressiveByteList to its own EIP so that it can be worked on independently of the eth_getLogs filter usage.

Plans are to work with @wemeetagain to generalize it beyond byte lists, and to simplify the parameters with `CAPACITY = unbounded` as well as `COMMON_RATIO = 4`. That way, in EIPs such as EIP-6404 and EIP-6466 it is no longer necessary to define arbitrary bounds to byte lists, and in consensus we can use such an `UnboundedList[T]` construct to replace proposer / attester slashing lists (with a shared maximum allowed len), and could use it for attestations list that keeps changing across forks.

First step is to move this out to its own EIP with Zsolt as author, then do the followup work with me and @wemeetagain in separate PRs.

### DIFF
--- a/EIPS/eip-####.md
+++ b/EIPS/eip-####.md
@@ -1,0 +1,67 @@
+---
+title: SSZ ProgressiveByteList
+description: New SSZ type to improve efficiency for short lists
+author: Zsolt Felföldi (@zsfelfoldi)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: Core
+created: 2025-03-24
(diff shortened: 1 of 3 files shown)


zsfelfoldi (approved):

g11tech (approved):

eip-automerger:

 - Contains new file EIPS/eip-7916.md

PR #9017, opened 2024-11-04
Add EIP: SSZ execution blocks

etan-status:

PR #6466, opened 2023-02-08
Add EIP-6466: SSZ receipts root

etan-status:
When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #9524, opened 2025-03-24
Update EIP-7607: Add "remove bloom filters" (EIP 7668) to PFI Fusaka

jochem-brouwer:
Adds "remove bloom filters" https://eips.ethereum.org/EIPS/eip-7668 to PFI Fusaka.


g11tech (approved):
huge +1

ncitron:
If we wanted to do this it might make sense to make some of the data structures a bit more snark friendly. In my explorations on building out an implementation of this, I have noticed that the current zkVMs are much faster at SHA2 then Keccak (SHA3). It may make sense to explore replacing the receipts merkle tree with SHA2 and the block header hashing as well.

It might also make sense to just use SSZ structures here, such as making the header and receipts list an SSZ object which uses SHA2 anyways. This is mostly encompassed by #6466 and #9017.

jochem-brouwer:
@ncitron I think your proposal is to make the transactionsTrie and receiptsTrie to still (?) use MPT, but instead of Keccak256 use SHA2 to hash the trie?

This is beyond the scope of this EIP though. 

If you want these EIPs https://github.com/ethereum/EIPs/pull/6466 https://github.com/ethereum/EIPs/pull/9017 considered then move them to PFI for Fusaka (but do that today, deadline to add PFIs is today :) See: https://ethereum-magicians.org/t/eip-7607-fusaka-meta-eip/18439 )

g11tech:
yes those defintely makes sense but will blowup the scope of this EIP and would/should be independent. how about doing an EIP to change the hashing for receipts and suggest poseidon even? easy for ACD to decide piecemeal if they want something

timbeiko (approved):

timbeiko:
@jochem-brouwer there's a merge conflict that needs to be addressed

jochem-brouwer:
I fear this will happen with every PR which PFIs something and when another PR which also PFIs somethng gets merged, right? Will fix it.

jochem-brouwer:
Interesting, so you approved and the merge commit itself got approved and auto-merged as well, despite having merge conflicts (which I solved in the merge commit). I would assume it is then also possible to "smuggle" in extra contents which would need a re-approval :thinking:

PR #9529, opened 2025-03-25
Update EIP-5792: Propose changes to EIP-5792 atomicity

cryptotavares:
## Goal
Improve the current API to support, in an expressive and simple way, both atomic and non-atomic batch requests.

## Context
The atomic topic has been heavily discussed within the latest 5792 calls. And it was agreed that in order to foster broader adoption of the proposed wallet APIs specified in this spec, it should support both `atomic` and `non-atomic` batches.

## Changes
Introduce a top level `atomic` property across the different API methods (`wallet_sendCalls`, `wallet_getCallsStatus`, `wallet_getCapabilities`).

Wallets express the `atomic` values it supports via `wallet_getCapabilities`.

~~The `strict`, `loose` and `none` atomicity concepts are directly taken from the [flow-control capability](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7867.md) proposed by @SamWilsn.~~


EIPS/eip-5792.md
```diff
@@ -88,6 +88,7 @@ type SendCallsParams = {
   id?: string;
   from?: `0x${string}`;
   chainId: `0x${string}`; // Hex chain id
+  atomicity: "atomic" | "non-atomic" | "any";
```
lukasrosario, line 91: nits:
- wondering if should be `strict`, `loose`, `any`. seems a little more descriptive? @SamWilsn curious if you have strong opinions on naming here
- also wondering if it should be camel case, given these are largely used in js and camel case is more idiomatic

EIPS/eip-5792.md
```diff
@@ -318,6 +324,34 @@ The capabilities below are for illustrative purposes.
 }
 ```
 
+### `atomicity` Capability
+
+Like the illustrative examples given above and other capabilities to be defined in future EIPs, the capability to specify how calls delivered via the above-defined methods should be executed can be expressed  as `atomic` or `non-atomic`.
+
+This capability is expressed separately on each chain and should be interpreted as a guarantee only for batches of transactions on that chain
```
lukasrosario, line 331: is this supposed to be how the wallet expresses the values for `atomicity` it supports? if so think we could reword to make a bit clearer, lmk if you had something else in mind though

EIPS/eip-5792.md
cryptotavares: So I initially considered those options (in order to better integrate with the [flow-control capability](https://github.com/ethereum/EIPs/blob/c941b5c4519dcf24f39a1bc9163ed28cc5ac5a11/EIPS/eip-7867.md)). However, I thought that `atomic` should cover both `strict` and `loose` (for 99% percent of users those are implementation details... which can be fined tuned via the flow-control capability).

I agree that we can probably drop the `non-atomic` option. I initially included it because, although I am not sure why an App would want to explicitly execute calls sequentially, I also did not want close that door.

cryptotavares: Also agree on the camel case.. updated 👍

EIPS/eip-5792.md
cryptotavares: Yes that was the idea...
I have kept the details on the how the flow is controlled, because I believe that those are implementation details that can be better expressed via the flow-control capability. I have prioritised simplicity over configuration. Meaning that the Wallet can implement the both the `atomic` and `non-atomic` option as it sees fit. As long as it respects the atomicity for the calls execution.

Re-worded it a bit. Let me know what you think.

EIPS/eip-5792.md
```diff
@@ -88,6 +88,7 @@ type SendCallsParams = {
   id?: string;
   from?: `0x${string}`;
   chainId: `0x${string}`; // Hex chain id
+  atomicity: "atomic" | "nonAtomic" | "any";
```
wilsoncusack, line 91: Seems like it would be nice to have atomicPreferred for when it's preferred but not required

EIPS/eip-5792.md
cryptotavares: On having only the `atomic` value vs having `strict` and `loose`, I think that there'a tradeoff there. For once, having more explicit values would simplify the `wallet_getCallsStatus`, as we can specify the expected `receipts` for each option. At the same time, It feels that we are drilling to much into implementation details that are better extracted to its own capability.

Let me know your thoughts @lukasrosario and @SamWilsn.

EIPS/eip-5792.md
cryptotavares: Would it ever not be preferred? 🤔 Both from Wallet and App perspective..

The idea for `any`, is for when an App does not care about execution, but really just wants to send a batch of transactions to the wallet.

Also I have not added the option for the Wallet to support both non-atomic and atomic execution for the same chain ([see here](https://github.com/cryptotavares/EIPs/blob/cryptotavares/atomicity-proposal/EIPS/eip-5792.md#atomicity-capability)). The reason was exactly that if `atomic` is available then the calls will be executed atomically.
Do you think that there's a case where:
* A wallet would prefer sequential execution on a chain where it supports atomic execution?
* An App would prefer sequential execution when it can ask for atomic execution?

And if yes, do you feel that it should be included in the spec, or that it is specific enough that it could be its own capability (defined as a different erc)?

EIPS/eip-5792.md
```diff
@@ -318,6 +324,38 @@ The capabilities below are for illustrative purposes.
...
+### `atomicity` Capability
+
+Like the illustrative examples given above and other capabilities to be defined in future EIPs, the `atomicity` capability specifies how the wallet will execute the batches of transactions requested through `wallet_sendCalls` method.
+
+The only valid JSON-RPC values for this capability's only member, `atomicity`, are `"atomic"` and `"nonAtomic"`:
+
+* `atomic` means that the wallet will execute the calls as part of a single transaction.
+* `nonAtomic` means that the wallet will execute the each call as an individual transaction.
+
+This capability is expressed separately on each chain and should be interpreted as a guarantee only for batches of transactions on that chain.
```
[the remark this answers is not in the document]
cryptotavares: Do you think that there's a case where a Wallet would prefer to submit a batch sequentially when atomic is available?
See my comment here - https://github.com/ethereum/EIPs/pull/9529#discussion_r2012813371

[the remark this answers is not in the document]
azf20: Agree with @cryptotavares - and if the use case is (for example) related to the size / number of calls being too large for one transaction I don't think it's worth introducing that edge case (i.e. the call should fail & the app should batch calls on its side)

EIPS/eip-5792.md
```diff
@@ -328,7 +366,11 @@ We also debated whether the methods should be called `wallet_sendTransaction`, `
 
 ### Call Execution Atomicity
 
-The `wallet_sendCalls` method accepts an array of `calls` and by default requires that these calls be executed as part of a single transaction. That said, we expect that in some cases app developers might want to submit batched calls if and only if they can be executed non-atomically. This would enable EOA wallets to accept multiple calls as well over the same interface. A definition of this behaviour may be proposed as a separate capability ERC.
+The `wallet_sendCalls` method accepts an array of `calls`. However, this proposal does not require that these calls be executed as part of a single transaction. It enables EOA wallets to express their ability to execute calls as part of a single transaction (atomically) or as part of multiple transactions (non-atomically). It also enables Apps to express their preference for how calls must be executed (in case a wallet supports both atomic and non-atomic execution).
```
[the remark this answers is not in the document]
azf20: How can there be revert protection across multiple transactions? Seems like that is hard to generalise?

EIPS/eip-5792.md
```diff
@@ -328,7 +366,11 @@ We also debated whether the methods should be called `wallet_sendTransaction`, `
...
+
+The `atomicity` special capability that was made an integral part of this specification in order promote expressiveness. And due to its importance to reduce ambiguity between the App and the Wallet, the capability is expressed as a top-level field in the `wallet_sendCalls` request and the `wallet_getCallsStatus` response.
```
azf20, line 371: I think as well as expressiveness, this is to facilitate adoption by EOAs (for example where 7702 isn't available or a user hasn't enabled it for their account)

EIPS/eip-5792.md
```diff
@@ -151,9 +153,11 @@ Note that the `receipts` objects of this method's response is a strict subset of
...
-* If a wallet executes multiple calls **non-atomically** through some `capability` defined elsewhere, `wallet_getCallsStatus` MUST return an object with a `receipts` field that contains **an array of receipts** for all transactions containing batch calls that were included onchain. This includes the batch calls that were included on-chain but eventually reverted.
+* If a wallet executes multiple calls **atomically** in a single transaction, `wallet_getCallsStatus` MUST return an object with a `receipts` field that contains a single transaction receipt, corresponding to the transaction in which the calls were included. it also MUST be explicit and return a `atomicity` field set to `"atomic"`.
+* If a wallet executes multiple calls **non-atomically**, `wallet_getCallsStatus` MUST return an object with a `receipts` field that contains **an array of receipts** for all transactions containing batch calls that were included onchain. This includes the batch calls that were included on-chain but eventually reverted. It also MUST be explicit and return a `atomicity` field set to `"nonAtomic"`.
```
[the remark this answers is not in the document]
matthewwalsh0: This is in reference to "loose" atomicity where a wallet could use `eth_sendBundle` for example and multiple transactions could be submitted "atomically", at least in 99.9% of cases.

EIPS/eip-5792.md
cryptotavares: Updated the proposal to include loose atomicity. The `strict`, `loose` and `none` atomicity concepts are directly taken from the [flow-control capability](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7867.md) proposed by @SamWilsn

EIPS/eip-5792.md
```diff
@@ -222,6 +236,7 @@ The `id` batch identifier is a unique 64 bytes represented as a hex encoded stri
   "chainId": "0x01",
   "id": "0x00000000000000000000000000000000000000000000000000000000000000000e670ec64341771606e55d6b4ca35a1a6b75ee3d5145a99d05921026d1527331",
   "status": 200,
+  "atomicity": "strict";
```
frangio, line 239:
```suggestion
  "atomicity": "strict",
```

EIPS/eip-5792.md
```diff
@@ -43,6 +43,8 @@ These fields are set inside this capability's entry in the `capabilities` object
 Each entity in the `calls` field may contain an optional `capabilities` object.
 This object allows the applications to attach a capability-specific metadata to individual calls.
 
+The `atomicity` field specifies how the wallet should handle the batch of calls. Despite not being a capability, and being a top-level field in the request, the wallet supported values can be sourced via `wallet_getCapabilities`.
```
frangio, line 46: > Despite not being a capability

This is inconsistent with the later section titled "`atomicity` Capability". Should this just say something like the following instead?

```suggestion
The `atomicity` field specifies how the wallet should handle the batch of calls. It is a built in capability and the wallet supported values can be sourced via `wallet_getCapabilities`.
```

EIPS/eip-5792.md
```diff
@@ -53,10 +55,16 @@ The wallet:
 * MUST send the calls from the address specified in the request's `from` property, if provided
   * If `from` is not provided the wallet SHOULD present the user with an opportunity to view and select the `from` address during confirmation
 * MUST NOT await for any calls to be finalized to complete the batch
-* MUST submit multiple calls as an atomic unit in a single transaction
 * MUST NOT send any calls from the request if the user rejects the request
-* MAY revert all calls if any call fails
-* MUST not execute any further calls after a failed call
+* When atomicity is set to `strict`:
+  * MUST execute all calls atomically in a single transaction.
```
SamWilsn, line 60:
```suggestion
  * MUST execute all calls atomically—either all calls execute successfully or no material effects appear on chain.
  * MUST execute all calls contiguously, meaning no other transactions/calls may be interleaved with the calls from this batch.
```

I would _really_ like to avoid the "single transaction" language. There's no reason to require a single transaction. You can get atomicity over multiple transactions. The trivial example would be a multi sig where each signature is submitted in a transaction and the final signature triggers execution.

EIPS/eip-5792.md
```diff
@@ -53,10 +55,16 @@ The wallet:
...
+* When atomicity is set to `loose`:
+  * MAY execute each call as a single transaction, as part of a bundle that is executed atomically.
+  * MAY execute all calls atomically in a single transaction.
+* When atomicity is set to `none`:
+  * MAY execute each call as a single transaction.
+  * MAY execute each call as a single transaction, as part of a bundle that is executed atomically.
+  * MAY execute all calls sequentially without any atomicity guarantees.
```
SamWilsn, line 67: This doesn't specify the behaviour on failed transactions.

EIPS/eip-5792.md
cryptotavares: I might be missing something, but I do not see how a multi sig would trigger an atomic execution through multiple transactions. Nevertheless I agree with you. We should not restrict the spec to `strict = single transaction`. We should leave it broad enough so that the space can evolve unconstrained.
An example is the `sendBundle` announced by Linea team. That is a new rpc method that they are implementing and will achieve `strict` atomicity through multiple transactions (as Linea has single slot finality there are no reorgs). So a batch is only included if all transactions of the batch are executed properly (can be included in the block and don't revert).

EIPS/eip-5792.md
```diff
@@ -328,7 +376,11 @@ We also debated whether the methods should be called `wallet_sendTransaction`, `
...
+
+The `atomicity` special capability that was made an integral part of this specification in order promote expressiveness and facilitate adoption (from both wallets and apps). And due to its importance to reduce ambiguity between the App and the Wallet, the capability is expressed as a top-level field in the `wallet_sendCalls` request and the `wallet_getCallsStatus` response.
+
+We initially proposed that multiple calls must be executed atomically, but after some debate we ultimately decided this was too opinionated. Instead, we are including a specification for an `atomicity` capability. This allows for EOA wallets to accept multiple calls and still gives developers the option to only submit batched calls if they are executed atomically.
```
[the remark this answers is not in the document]
cryptotavares: Agree. Seem my comment here - https://github.com/ethereum/EIPs/pull/9529/files#r2022512583

EIPS/eip-5792.md
```diff
@@ -151,9 +161,12 @@ Note that the `receipts` objects of this method's response is a strict subset of
...
+* If a wallet executes multiple calls **atomically** in a single transaction, `wallet_getCallsStatus` MUST return an object with a `receipts` field that contains a single transaction receipt, corresponding to the transaction in which the calls were included. It also MUST be explicit and return an `atomicity` field set to `"strict"`.
+* If a wallet executes multiple calls **atomically** in a multiple transactions as part of a bundle, `wallet_getCallsStatus` MUST return an object with a `receipts` field that contains **an array of receipts** for all transactions. It also MUST be explicit and return a `atomicity` field set to `"loose"`.
+* If a wallet executes multiple calls **non-atomically**, `wallet_getCallsStatus` MUST return an object with a `receipts` field that contains **an array of receipts** for all transactions containing batch calls that were included onchain. This includes the batch calls that were included on-chain but eventually reverted. It also MUST be explicit and return a `atomicity` field set to `"none"`.
```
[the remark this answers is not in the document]
cryptotavares: The benefit was to try to make the reasoning about the receipts structure simple and comprehensive enough for dapps. Reduce the complexity and any ambiguity that might exist. However I feel that we can make it less restrictive, as I agree with your previous comment about `strict` being executed as multiple transactions. At that point, then the receipts field can have:
* a single or multiple transactions for `strict` atomicity
* multiple transactions for `loose` and `none`

EIPS/eip-5792.md
```diff
@@ -340,7 +392,7 @@ Wallets that do not support the methods defined here SHOULD return error respons
 
 ## Security Considerations
 
-App developers MUST NOT assume that all calls will be sent in a single transaction if they were submitted to a wallet with a capability defining how to execute the calls non-atomically. The exact behavior of non-atomically executed calls is to be defined in separate ERCs.
+App developers MUST NOT assume that all calls will be sent in a single transaction if they were submitted to a wallet with a `atomicity` set to `nonAtomic` or another capability defining how to execute the calls non-atomically.
```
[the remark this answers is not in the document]
cryptotavares: From a previous version of the proposal 😅
Removing it.

EIPS/eip-5792.md
```diff
@@ -53,10 +55,13 @@ The wallet:
...
-* MUST submit multiple calls as an atomic unit in a single transaction
 * MUST NOT send any calls from the request if the user rejects the request
-* MAY revert all calls if any call fails
-* MUST not execute any further calls after a failed call
+* When atomic is set to `true`:
+  * MUST execute all calls atomically—either all calls execute successfully or no material effects appear on chain.
+  * MUST execute all calls contiguously, meaning no other transactions/calls may be interleaved with the calls from this batch.
+* When atomic is set to `false`:
+  * SHOULD execute all calls sequentially without any atomicity/contiguity guarantees.
+  * MAY upgrade to `atomic: true`.
```
jakubuid, line 64: How does the upgrade from atomic `true` to `false` work?

EIPS/eip-5792.md
```diff
@@ -318,6 +329,39 @@ The capabilities below are for illustrative purposes.
...
+Like the illustrative examples given above and other capabilities to be defined in future EIPs, the `atomic` capability specifies how the wallet will execute the batches of transactions requested through `wallet_sendCalls` method.
+
+The only valid JSON-RPC values for this capability's only member, `atomic`, are `true` or `false`:
+
+* `true` means that the wallet will execute the calls atomically and contiguously. 
+* `false` means that the wallet does not provide any atomicity or contiguity guarantees.
+
+This capability is expressed separately on each chain and should be interpreted as a guarantee only for batches of transactions on that chain.
+
+If the `atomic` capability is not present for a given chain, and unless it is explicitly overriden by anothe `capability` (not in scope of this EIP), it means that the wallet does not support batching for that given chain.
```
jakubuid, line 343:
```suggestion
If the `atomic` capability is not present for a given chain, and unless it is explicitly overriden by another `capability` (not in scope of this EIP), it means that the wallet does not support batching for that given chain.
```

EIPS/eip-5792.md
lukasrosario: nit: wording is a little confusing here & makes it sound like the default should be to downgrade if a wallet is capable of atomicity. would flip it around & be a little more explicit:
```
...
* When atomic is set to `false`:
  * MAY execute all calls sequentially without any atomicity/contiguity guarantees.
  * If a wallet that is capable of providing atomicity guarantees receives a batch with `atomic: false`, it MAY execute the batch atomically.
```

EIPS/eip-5792.md
```diff
@@ -340,7 +388,7 @@ Wallets that do not support the methods defined here SHOULD return error respons
 
 ## Security Considerations
 
-App developers MUST NOT assume that all calls will be sent in a single transaction if they were submitted to a wallet with a capability defining how to execute the calls non-atomically. The exact behavior of non-atomically executed calls is to be defined in separate ERCs.
+Regardless of the `atomic` value specified, App developers MUST NOT assume that all calls will be sent in a single transaction.
```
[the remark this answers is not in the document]
drortirosh: The definition above states:
> * MUST execute all calls atomically—either all calls execute successfully or no material effects appear on chain.
Can you give an example of an implementation of "atomic:true" that matches that assert, and still is not on a single transaction ?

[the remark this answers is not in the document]
cryptotavares: @drortirosh see my comment [here](https://github.com/ethereum/EIPs/pull/9529#discussion_r2022512583)

EIPS/eip-5792.md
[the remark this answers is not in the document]
cryptotavares: see my comment https://github.com/ethereum/EIPs/pull/9529#discussion_r2022512583

EIPS/eip-5792.md
```diff
@@ -318,6 +330,39 @@ The capabilities below are for illustrative purposes.
...
+type AtomicCapability = boolean;
+```
+
+
+#### `wallet_getCapabilities` Example Return Value including `atomicBatch`
+
+```json
+{
+  "0x2105": {
+    "atomic": true
```
bumblefudge, line 358: sorry if i missed something, but why are these
`"atomic": true`
instead of 
`"atomic": { "supported": true }`
like the other examples above?

bumblefudge (changes requested):
see in-line nit about shape of "atomic" capability

EIPS/eip-5792.md
cryptotavares: The other capabilities highlighted in the spec are just examples of what we might have, but they are not real capabilities defined anywhere. Also there is no specified type for the capability itself (currently is specified as `any`). Only for the [`getCapabilities` response](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md#wallet_getcapabilities-rpc-specification):
```js
type GetCapabilitiesResult = Record<`0x${string}`, <Record<string, any>>; // Hex chain id
```
I opted for a simple boolean just for simplicity.

Do you think we should change the capability type from `any` to be something more structured?
My personal opinion is that we should keep it unconstrained, because adopting a capability will be an intended action. So app and wallet will need to specifically develop and implement such capability. And if that is the case, then we should, at least for now, not constrain capability authors with a specified schema/structure.

What do you think?

EIPS/eip-5792.md
cryptotavares: Do you think that we need to specify the behaviour for failed transactions submitted sequentially?

I was thinking that we could leave that for the wallets to decide. Meaning that in the simplest form, an app that does not care about atomic execution, will need to go through the receipts and figure out the batch was executed. I would leave further specifications to a different capability (like the [flow-control](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7867.md) one).

EIPS/eip-5792.md
lukasrosario: i think we should keep the `GetCapabilitiesResult` type as is but agree we should conform to `"atomic": { "supported": true }`. i dont think that format needs to be a strict requirement, but i dont see why we'd stray from it

lukasrosario (approved):
lfg

PR #9538, opened 2025-03-25
Update EIP-7668: Move to Draft

acolytec3:
Adds me as a new author for purposes of championing this for [inclusion in Fusaka](https://github.com/ethereum/EIPs/pull/9524).  Will also refine implementation details


g11tech:
if you have an proposed changes for this EIP that you want to champion, would be good to add those as well

EIPS/eip-7668.md
```diff
@@ -12,23 +12,23 @@ created: 2024-03-31
...
+In practice, this mechanism is far too slow. Almost all dapps that access history end up doing so not through RPC calls to an Ethereum node (even a remote-hosted one), but through centralized extra-protocol services.  Not too mention, in a chain with relatively full blocks, the false positive rate in bloom filters has increased to the point of making bloom filters almost useless.
 
-This EIP proposes recognizing that reality, and removing bloom filters from the protocol. Applications that need history querying would be encouraged to develop and use decentralized protocols that create provable log indexes using eg. ZK-SNARKs and incrementally-verifiable computation.
+This EIP proposes recognizing that reality, and removing bloom filters from the protocol since it places an unnecessary compute burden on execution clients computing what is for all prati. Applications that need history querying would be encouraged to develop and use decentralized protocols that create provable log indexes using eg. ZK-SNARKs and incrementally-verifiable computation.
```
jochem-brouwer, line 23: This sentence seems incomplete?

EIPS/eip-7668.md
```diff
@@ -12,23 +12,23 @@ created: 2024-03-31
...
 
 ## Specification
 
-The logs bloom of an execution block is now required to be empty (ie. 0 bytes long). The logs bloom of a transaction receipt is now required to be empty (ie. 0 bytes long).
+The `logsbloom` field of the execution block header and the `logsbloom` field of the transaction receipt is now required to be `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` (ie. 32 bytes filled with `ff`).
```
jochem-brouwer:
```suggestion
The `logsBloom` field of the execution block header and the `logsBloom` field of the transaction receipt is now required to be `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` (ie. 256 bytes filled with `ff`).
```

jochem-brouwer: I guess cut it off (the 0xffff..ff), because this is wayyy too long.

Also, I think that the original (empty bytes it instead of writing the 1s) is better for storage and performance reasons. I know we wanted to write the all-bits-1 version because then the bloom filter would be positive for any query, but I think this can be handled at the application side (RPC). We could even wrap the RPC such that it returns the 0xffff..ffff as `logsBloom` for instance, or just even better really deprecate `logsBloom` and thus writing empty bytes.

We could also consider completely removing the `logsBloom` field from the receipts. This would save some extra bytes when hashing the receipt for the `receiptsTrie`.

EIPS/eip-7668.md
```diff
@@ -12,23 +12,23 @@ created: 2024-03-31
...
 
 ## Rationale
 
-This is a minimally disruptive way to remove the need to handle blooms from clients. A future EIP can later clean up by removing this field entirely, along with other fields that have been deprecated.
+This is a minimally disruptive way to remove the need to compute blooms from clients while not requiring any breaking consensus changes.
```
jochem-brouwer, line 31: Well, this is a breaking consensus change, since it forces the bloom filter to be a constant which in almost all practical cases will not be the produced bloom filter when executing the block.

EIPS/eip-7668.md
```diff
@@ -12,23 +12,23 @@ created: 2024-03-31
...
+Require the bloom filters in an execution block, including at the top level and in the receipt object, to be hardcoded to `0xffff...ff`.
 
 ## Motivation
 
 Logs were originally introduced to give applications a way to record information about onchain events, which decentralized applications (dapps) would be able to easily query. Using bloom filters, dapps would be able to quickly go through the history, identify the few blocks that contained logs relative to their application, and then quickly identify which individual transactions have the logs that they need.
 
-In practice, this mechanism is far too slow. Almost all dapps that access history end up doing so not through RPC calls to an Ethereum node (even a remote-hosted one), but through centralized extra-protocol services.
+In practice, this mechanism is far too slow. Almost all dapps that access history end up doing so not through RPC calls to an Ethereum node (even a remote-hosted one), but through centralized extra-protocol services.  Not too mention, in a chain with relatively full blocks, the false positive rate in bloom filters has increased to the point of making bloom filters almost useless.
```
jochem-brouwer, line 21:
```suggestion
In practice, this mechanism is far too slow. Almost all dapps that access history end up doing so not through RPC calls to an Ethereum node (even a remote-hosted one), but through centralized extra-protocol services.  Not to mention, in a chain with relatively full blocks, the false positive rate in bloom filters has increased to the point of making bloom filters almost useless.
```

One could also point to the fact that the gas limit will likely rise to 60M, which means more txs are packed in the block and which would thus bloat the filter even more because of more gas available for LOGs.

jochem-brouwer (commented):
Some comments!

EIPS/eip-7668.md
acolytec3: Are you saying replace the current field with a static byte array of 256 bytes of 0 or just an empty array?   I'm not confident on the implications for the RLP encoding of a tx receipt or block header.  I guess maybe nothing if we say as of the fork block, the logsBloom field becomes a `0` or `[]`, right?

EIPS/eip-7668.md
acolytec3: Well, breaking on the level of "we're changing how this field is computed" but breaking the structure of the values that are RLP encoded into the serialized form of the header and receipt, right?

EIPS/eip-7668.md
Speedtraders199: Then complet it so it can balance

acolytec3: > This sentence seems incomplete?

Yup, sorry, typo-d it.  Have updated with my latest commit

EIPS/eip-7668.md
jochem-brouwer: If we input the empty bytes for the `logsBloom` starting at fork block then the RLP structure will not change (besides that the field is now not the constant 256 bytes long, but now 0. This is not a problem, we have variable-length fields like `extraData` which can be something between 0-32 bytes). If we would input it as empty list then this would (from my perspective) change the structure somewhat. I'd say empty bytes is the way to go. So empty bytes is not the empty list `[]`, it is a field of length 0. (This is possible, txs with a `to` field of 0 bytes are CREATE txs)

EIPS/eip-7668.md
```diff
@@ -20,7 +20,7 @@ Logs were originally introduced to give applications a way to record information
 
 In practice, this mechanism is far too slow. Almost all dapps that access history end up doing so not through RPC calls to an Ethereum node (even a remote-hosted one), but through centralized extra-protocol services.  Not to mention, in a chain with relatively full blocks, the false positive rate in bloom filters has increased to the point of making bloom filters almost useless.
 
-This EIP proposes recognizing that reality, and removing bloom filters from the protocol since it places an unnecessary compute burden on execution clients computing what is for all prati. Applications that need history querying would be encouraged to develop and use decentralized protocols that create provable log indexes using eg. ZK-SNARKs and incrementally-verifiable computation.
+This EIP proposes recognizing that reality, and removing bloom filters from the protocol since it places an unnecessary compute burden on execution clients computing essentially used bytes. Applications that need history querying would be encouraged to develop and use decentralized protocols that create provable log indexes using eg. ZK-SNARKs and incrementally-verifiable computation.
```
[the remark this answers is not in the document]
acolytec3: Ugh, i can't type today.  Will revise my language.

EIPS/eip-7668.md
```diff
@@ -12,23 +12,23 @@ created: 2024-03-31
...
+This EIP proposes recognizing that reality, and removing bloom filters from the protocol since it places an unnecessary compute burden on execution clients who are currently computing essentially unused and unusable bloom filters and shipping them around the network. This will become even more acute if or when block level gas limits are increased as is currently being proposed in other EIPs. Applications that need history querying would be encouraged to develop and use decentralized protocols that create provable log indexes using eg. ZK-SNARKs and incrementally-verifiable computation.
 
 ## Specification
 
-The logs bloom of an execution block is now required to be empty (ie. 0 bytes long). The logs bloom of a transaction receipt is now required to be empty (ie. 0 bytes long).
+The `logsbloom` field of the execution block header and the `logsbloom` field of the transaction receipt are now required to be an empty bytes array.
```
jochem-brouwer, line 27:
```suggestion
The `logsBloom` field of the execution block header and the `logsBloom` field of the transaction receipt are now required to be an empty bytes array.
```

EIPS/eip-7668.md
```diff
@@ -2,7 +2,7 @@
 eip: 7668
 title: Remove bloom filters
 description: Remove bloom filters from the execution block
-author: Vitalik Buterin (@vbuterin)
+author: Vitalik Buterin (@vbuterin), Andrew Day (@acolytec3)
 discussions-to: https://ethereum-magicians.org/t/eip-7653-remove-bloom-filters/19447
 status: Stagnant
```
g11tech, line 7:
```suggestion
status: Draft
```

g11tech (approved):
lgtm since @jochem-brouwer and @acolytec3 from ethjs team wants to champion this.

would still require @vbuterin 's approval

PR #9542, opened 2025-03-26
Update EIP-7916: Remove bound parameter and use constant scaling factor

etan-status:
To simplify the usage of `ProgressiveList[T]`, make it unbounded by mixing in a 0 on the final non-empty layer, and choose 4 as constant scaling factor to simplify implementation logic. Further generalize the structure to support arbitrary element type `T` and expand motivation and rationale sections.


wemeetagain (approved):
@etan-status and I have been refining this over the past week.
LGTM

EIPS/eip-7916.md
```diff
@@ -38,30 +67,108 @@ ProgressiveByteList[3000, 4]
      \/
     ROOT
 
-V1: ByteVector[32]
-V2: ByteVector[128]
-V3: ByteVector[512]
-V4: ByteVector[2048]
-V5: ByteVector[512]
+V1: Vector[T, 1]
```
zsfelfoldi, line 70: Here the vector capacities are the same as the number of chunks in each subtree. This is only true if vector elements have a 32 byte size or are stored as `hash_tree_root`. Maybe an explanatory comment is enough here that this is an assumption and the vectors have fixed chunk count, not fixed element capacity.

EIPS/eip-7916.md
```diff
@@ -38,30 +67,108 @@ ProgressiveByteList[3000, 4]
...
 
 ## Backwards Compatibility
 
-The new SSZ type does not interfere with any existing types.
+`ProgressiveList[T]` is a new SSZ type, coexisting with `List[T, N]` and other types without conflict. Its `List`-equivalent serialization ensures compatibility with existing serializers.
 
 ## Security Considerations
 
-None
+- Resource limits: The `uint32` length-prefix caps serialization of variable-length subsections at 4GB, but practical limits (e.g., 10MB libp2p messages) apply. Implementations SHOULD enforce context-specific bounds.
```
zsfelfoldi, line 170: Actually, why define it as `uint32`? I mean it has little practical importance, but the length is a full chunk and encoded in little endian anyways, so it could just as well be `uint256`. Practical limits should still be enforced and they will be typically lower than 4G elements but since the type definition says it's a limitless list, I feel like it would just look better.

zsfelfoldi (approved):
See comments, otherwise LGTM.

EIPS/eip-7916.md
etan-status: The `uint32` is just how offsets get encoded of variable-length sections within a complex type. Have updated the wording to better reflect that.

EIPS/eip-7916.md
etan-status: Thanks, yes, that's true. Have added a comment.

PR #9543, opened 2025-03-26
Add EIP: Blob base fee bounded by execution cost

anderselowsson:


smartprogrammer93 (approved):
Yes, lets do it

EIPS/eip-0000.md
```diff
@@ -0,0 +1,99 @@
+---
+eip: 0000
```
g11tech, line 2: eip 7918

g11tech (approved):

PR #9547, opened 2025-03-26
Update EIP-7607: Add EIP-7919 to Fusaka PFI

etan-status:


timbeiko (approved):
Thanks for the whitespace trims 😄

etan-status:
There's also still Holesky vs Hoodi ;)

PR #9558, opened 2025-03-27
Add EIP: CALL_GAS_SPLIT opcode

rakita:
Introduce a notion of splitting of gas to limit gas that is send to the EXT*CALL


EIPS/call_gas_split.md
```diff
@@ -0,0 +1,64 @@
...
+
+With future multidimensional gas models, this instruction could potentially restrict all gas types.
+
+## Specification
+
+A new instruction `CALL_GAS_SPLIT` is introduced. This instruction:
+
+1. Pops one item from the stack representing a gas split ratio.
+2. Floors this value to `65535` (u16::MAX).
+3. Calculates the gas limit by using the **current** gas by spliting it by floored value as `split_gas=(gas*value)/65535`.And preserves this value until the end of the call.
```
jochem-brouwer, line 33: To avoid confusion I'd use a different var name as `value` here, since `value` is also used as parameter in legacy CALLs to denote how much eth is send. Suggestion: `fraction`.

EIPS/call_gas_split.md
```diff
@@ -0,0 +1,64 @@
...
+By default, the gas split is set to `63/64` to match the existing behavior defined in [EIP-150](./eip-150.md).
+
+With future multidimensional gas models, this instruction could potentially restrict all gas types.
+
+## Specification
+
+A new instruction `CALL_GAS_SPLIT` is introduced. This instruction:
+
+1. Pops one item from the stack representing a gas split ratio.
+2. Floors this value to `65535` (u16::MAX).
```
jochem-brouwer, line 32: I think it should exceptionally halt if the popped item is higher than this (to avoid unintended behavior)

EIPS/call_gas_split.md
```diff
@@ -0,0 +1,64 @@
...
+3. Calculates the gas limit by using the **current** gas by spliting it by floored value as `split_gas=(gas*value)/65535`.And preserves this value until the end of the call. 
+4. The preserved value should be used in all `EXT*CALL` operations to set gas limit as `max(split_gas, gas_left)`.
+
+If this value is not set, the EVM interpreter MUST maintain the `63/64` rule that was previously used as default.
```
jochem-brouwer, line 36: I'm assuming that the gas forwarded is still subject to the 63/64 rule. If this would not be the case, then the call stack depth limit of 1024 can be reached (this is also a DoS vector being able to cheaply create new call frames)

EIPS/call_gas_split.md
```diff
@@ -0,0 +1,64 @@
...
+
+A key problem with always forwarding all available gas is that a malicious callee could consume all gas during execution (even in case of a revert), potentially causing denial of service. By allowing the caller to restrict the forwarded gas to a specific portion, this risk can be mitigated.
+
+By default, the gas split is set to `63/64` to match the existing behavior defined in [EIP-150](./eip-150.md).
+
+With future multidimensional gas models, this instruction could potentially restrict all gas types.
+
+## Specification
+
+A new instruction `CALL_GAS_SPLIT` is introduced. This instruction:
```
jochem-brouwer, line 29: What is the gas cost of this instruction?

EIPS/call_gas_split.md
jochem-brouwer: When is `split_gas` calculated? After deducting the gas cost of the `CALL_GAS_SPLIT` opcode?

jochem-brouwer:
```suggestion
3. Calculates the gas limit by using the **current** gas by spliting it by floored value as `split_gas=(gas*value)//65535`.And preserves this value until the end of the call. 
```

Suggestion to floor the value to avoid non-integer arithmetic

EIPS/call_gas_split.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Rationale
+
+Several alternatives were considered for implementing gas split functionality:
+
+Adding gas split parameters directly to `EXT*CALL` call instructions: While possible, this would require setting the split for each call individually. Having a standalone instruction allows setting the split once for multiple calls.
+
+Hardcoding the gas split value as immediate: This would be restrictive for dynamic scenarios where the number of calls is only known at runtime.
+
+Using a stack item for `CALL_GAS_SPLIT`: This approach was chosen as it provides the most flexibility, allowing dynamic adjustment of the gas split ratio.
```
jochem-brouwer, line 46: This idea is nice, but between EXT*CALLs gas in the contract will also be consumed. So if you would split up current gas in 3 portions, then likely the third portion of gas "wanted" is much higher than the available gas. I would also not know how to solve this though :sweat:

jochem-brouwer (commented):
Some initial thoughts :)

EIPS/call_gas_split.md
rakita: Makes sense, will add that

EIPS/call_gas_split.md
rakita: If you have three calls, i would make the gas split in four parts. Last part would be reserved for current contract

EIPS/call_gas_split.md
rakita: It was briefly mentioned that 63/64 rule stayed in EOF calls to not disrupt the current pattern and have yet another test vector that we need to test. Having discussion on lifting this would be beneficial, I don't have strong opinion on it rn.

EIPS/call_gas_split.md
rakita: Have put it to `3` as it pops the item checks if it is less than u16::MAX, calculates with u64 a split and sets this split to the Interpreter.

EIPS/call_gas_split.md
rakita: Have added a line that this is up for discussion, it really depends on implementations if they can support big call stack, if it becomes troublesome, we can remove it.

EIPS/eip-99999.md
```diff
@@ -0,0 +1,70 @@
+---
+eip: 99999
```
g11tech, line 2: 7926

rakita:
Closed as obselete

PR #9561, opened 2025-03-27
Update EIP-7873: set tx type to a currently unused (?) first RLP byte

jochem-brouwer:
We should not use 0x05 as tx type for the InitcodeTransaction, because this is used as MAGIC byte when signing EIP-7702 authorities: https://eips.ethereum.org/EIPS/eip-7702#parameters

As far as I am aware, 0x06 is not used as magic byte or "RLP escape byte".

### DIFF
--- a/EIPS/eip-7873.md
+++ b/EIPS/eip-7873.md
@@ -33,7 +33,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 | Constant | Value |
 | - | - |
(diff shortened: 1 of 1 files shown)


pdobacz:
Makes sense and LGTM. We (Ipsilon) can update the EIP-7873 EEST, evmone and Megaspec PRs. Can we get a 2nd pair of eyes approve @axic @chfast @gumb0 @shemnon ?

pdobacz (approved):
There's :+1: from other people too, approving. Thank you!

PR #9574, opened 2025-03-31
Update EIP-7702: Update Parameter to be more specific

shemnon:
Update the `MAGIC` parameter to be specific to `SET_CODE_TX` scope.


lightclient:
I'm not really sure the need for this? The magic is already context specific to the EIP. The tx type is more specific because it is filling in the definition for eip-2718.

shemnon:
It's to give the parameter a globaly unique name.  I'm looking into gathering all such specified parameters in all the EIPs and this one stood out as lacking clarity in it's name.  MAGIC means one thing in 7702 and another thing in code headers, where `0xef00` and `0xef01` are the magic values for EOF and code delegation prefixes.

montana64:
Coinsult_Solaxy_0xe0...CF48_Audit.pdf

PR #9580, opened 2025-04-01
Add EIP: Block-Level Access Lists

nerolation:


benaadams:
For parallelization would be better if read and write are distinguished (if in write list doesn't need to be in read list as SSTORE both reads and writes)

This is as multiple txs that only read can happen at same time; is only ones that write that enforce a specific ordering of tx order

nerolation:
> For parallelization would be better if read and write are distinguished (if in write list doesn't need to be in read list as SSTORE both reads and writes)

This is already done by having an optional second value for the tuple of values. If there is only one, it's a read. This is early design though and not something I've spent much time further optimizing for practical and worst case use.

octavio12345300 (approved):
List

EIPS/eip-9999.md
```diff
@@ -0,0 +1,92 @@
...
+created: 2025-03-31
+---
+
+## Abstract
+
+This EIP proposes adding block-level access lists. By including a complete list of all addresses and storage keys accessed during a block, along with their initial values, in the block header, we can improve execution efficiency and enable better parallelization.
+
+## Motivation
+
+Currently, transactions without an explicit transaction access list cannot be efficiently parallelized, as the execution engine cannot determine in advance which addresses and storage slots will be accessed. While transaction-level access lists exist, they are not enforced, making it difficult to optimize execution pipelines.
```
jochem-brouwer, line 19: The scope of this EIP seems to be parallelization by usage of access lists. The focus seems to be storage slots. Note that other things can also change, namely: account code and account balance. This is definitely needed for parallelization, think of an easy example where Tx 1 edits the balance of account A, and Tx 2 does (code) branch X if A has balance < C or branch Y if A has balance >= C. For this parallel execution balances would also need to be tracked.

EIPS/eip-9999.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Block Header Modification
+
+Introduce a new field in the block header:
+
+```python
+class BlockBody:
+    ...
+    # Each access → ordered list of tx indices and associated values
+    block_level_accesses: Dict[
```
jochem-brouwer, line 35: For the parallelization (looking at storage only) we need to know:

- Which address is being read
- Which storage slot is being read
- Which storage slot is being written to

We need to know the first tx to read(/write) a storage slot (writing to storage implicitly means reading the storage (due to gas calculation we need to know the current value)). For the first read/write we thus need to know the initial value. After the first write we also need to know the final value of that write (so the value in the slot after this "first tx" is done). Note that even if the entire tx reverts, but it does read the storage, it should still be added to the list as being the first tx to read the slot.

I propose this structure:

```
block_level_accesses = List[
    Tuple(Address, List(Tuple(StorageKey, List[Txs], List[StorageValues])))
]
```

This is an ordered list. The first time a storage slot is being read, the address is being added to the list, as well as the Tx index (note: tx index this is one-based, not zero-based, zero has a special value, more on this later). This adds the initial value to the `List[StorageValues]`. In case the value changes at the end of this tx, also add the new value to the `List[StorageValues]`. Note that the storage list is also ordered in the tx order, therefore the first item in the `List[Txs]` in the `List(Tuple(StorageKey` list will either be equal than or higher than the previous tuple item in the list.

In the case that the address and the storage are already in the list AND it is being written to, this thus means that a tx previously has read or written the storage value. In that case, if this tx also ends up writing a different value to the storage slot, add the final item for this storage slot to the list (as well as the tx index).

Note these layouts for the list and what we can deduce:

`(Key, [1], [1])` -> The tx with index 0 (note: one-based!) has read `Key` and the value was 1. The net end result is that the value stays 1 in this slot.
`(Key, [1], [1,2])` -> The tx with index 0 has read `Key` and has written `2` to it (this is thus the value of the storage after the tx has finished)
`(Key, [1,2], [1,2,1])` -> The tx with index 0 has read `Key` and written `2` to it. The tx with index 1 has written `1` to it.
`(Key, [1,2], [1,2])` -> The tx with index 0 has read `Key`, but the end result after this tx is that it is still the initial value (1). The tx with index 1 has however written 2 to it, which is thus the storage value after tx with index 1.

**Why 1-based**

Because we also need to take into account system calls! EIP-2935 for instance writes a blockhash to a ring buffer contract PRIOR to all txs. So, we need to take into account these storage read/writes as well (actually in this case we only care about the write). Therefore we reserve tx index 0 for all system calls PRIOR to the tx. This thus also means that for any system calls which edit state AFTER all txs are ran should also be added. This will thus have the index `tx-index` in the block + 2 (tx-index being 0 based here. So if a block has 1 tx then index 0 is pre-tx-access, index 1 is tx 0, and index 2 are system edits after the txs).

EIPS/eip-9999.md
```diff
@@ -0,0 +1,92 @@
...
+
+```
+
+Transactions must not read or write to any address or storage key not declared in the block-level access list. Additionally, the block is invalid if the access list contains entries for addresses or storage keys that were not actually accessed, or if the recorded values deviate from those observed during execution.
+
+## Rationale
+
+### Block Size Considerations
+
+Including access lists increases block size, potentially impacting network propagation times. Empirically, typical access lists are around 30 KiB, with occasional peaks up to 120 KiB. Efficient encoding and potential limits on access list sizes should be considered.
```
jochem-brouwer, line 70: Block headers are supposed to be small. A hash of the list (either a MPT trie or whatever trie and take the root hash of this) should be the value in the header and the actual data should be in the block body (like transactions or withdrawals)

EIPS/eip-9999.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Security Considerations
+
+### Validation Overhead
+
+Nodes must verify that the access lists in the block header accurately reflect the actual accesses during execution. Discrepancies should result in block invalidation.
+
+### Block Size
+
+Including comprehensive access lists may increase block size, potentially impacting network propagation times. Empirical analysis shows typical access lists would have had around 30 KiB, with occasional peaks up to 120 KiB.
```
jochem-brouwer, line 88: We need a worst case scenario for this, and this will also depend on the data structure of this access list. Likely the worst case scenario is either one tx doing SLOADs over random keys or the same tx doing SSTOREs to these random keys. Cold SLOAD is 2100 and cold SSTORE for a slot which has nonzero value prior the transaction is 5k (but this SSTORE will thus write the initial + final value to the list). So napkin analysis says that

SLOAD: `Math.ceil(30M/2.1k)` = 14286 keys, so key+value of 32 bytes would yield 914.304 KiB
SSTORE: `Math.ceil(30M/5k)` = 6000 keys, so key+value+new value of 32 bytes would yield 576 KiB. 

This is of course without any encoding and maybe there are other ways to get larger values (EIP 2930 access lists maybe to make those SLOADs/SSTOREs slightly cheaper)

EIPS/eip-9999.md
```diff
@@ -0,0 +1,92 @@
...
+
+    if computed_access_list != block.header.block_level_access_list:
+        raise InvalidBlock("Mismatch in block-level access list or values.")
+
+```
+
+Transactions must not read or write to any address or storage key not declared in the block-level access list. Additionally, the block is invalid if the access list contains entries for addresses or storage keys that were not actually accessed, or if the recorded values deviate from those observed during execution.
+
+## Rationale
+
```
jochem-brouwer, line 67: Since we now have an access list for address+storage at the block level, can we now drop EIP-2929 logic entirely? So treat all accounts+storage as "warm"?

EIPS/eip-9999.md
```diff
@@ -0,0 +1,92 @@
...
+This EIP proposes adding block-level access lists. By including a complete list of all addresses and storage keys accessed during a block, along with their initial values, in the block header, we can improve execution efficiency and enable better parallelization.
+
+## Motivation
+
+Currently, transactions without an explicit transaction access list cannot be efficiently parallelized, as the execution engine cannot determine in advance which addresses and storage slots will be accessed. While transaction-level access lists exist, they are not enforced, making it difficult to optimize execution pipelines.
+
+By enforcing access lists at the block level, we shift the responsibility of declaring storage accesses to the block builder. This enables the execution engine to efficiently parallelize transaction execution, knowing in advance the exact scope of storage interactions for each transaction. 
+
+With access lists at the block level, transactions can be scheduled and executed in parallel, allowing for performance improvements.
```
jochem-brouwer, line 23: For this parallelization topic, has the re-introduction of intermediate state roots in the tx receipts also be studied/considered? If we consider a bulky block builder then this should be fine, any additional data reads should be fine for the bulky block builder (but should be prevented on local/home nodes)

I'm not entirely sure what impact this could have on parallelization, but for instance one could prove a single tx with merkle proofs against the previous tx receipt root and use this as some kind of witness or help speeding up things (this intermediate tx state root is kind of a shower thought :sweat_smile: )

jochem-brouwer (commented):
I love the work on this! Left some lengthy thoughts. In general, I think that besides storage slots we also need to track accesses to balances of accounts, and we also need to know if new contracts are created. Such that if Tx A creates contract Z, then if Tx B calls Z it thus has to "wait" before Tx A has executed. The nice property of https://eips.ethereum.org/EIPS/eip-6780 is that we know that if a contract is creatd, it will now stay there, and can't be deleted anymore :smile: :+1: 

General other question: is there also a discussion chat about this topic?

EIPS/eip-9999.md
jochem-brouwer: Note: also assumes 30M block gas limit, but we are likely making this higher, so this is also too low likely.

EIPS/eip-9999.md
```diff
@@ -0,0 +1,92 @@
...
+def state_transition(block: Block) -> None:
+    computed_access_list = {}
+
+    for idx, tx in enumerate(block.transactions):
+        accessed_items = execute_transaction(tx)  # returns {(address, slot): (value_before, optional_value_after)}
+        for key, val in accessed_items.items():
+            computed_access_list[key] = (idx, val)
+
+    if computed_access_list != block.header.block_level_access_list:
+        raise InvalidBlock("Mismatch in block-level access list or values.")
```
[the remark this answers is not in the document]
nerolation: The specifications in the current EIP use the EELS, but I agree that this point could be made clearer and implemented in closer alignment with the EIP.

EIPS/eip-9999.md
nerolation: Yeah agree, this is needed. We could have addresses mapped to an ordered list of events which are tuples containing the tx index and another tuple with the pre and post balance:

```python
class BlockBody:
    ...
    # Mapping each address to a list of balance diff events.
    # Each event is a tuple containing:
    #   - The transaction index that effected the change.
    #   - The balance diff
    balance_diffs: Dict[
        Address,
        List[Tuple[int, Value]]
    ]
    ```

File: EIPS/eip-9999.md (hunk 3 above)
Comment (nerolation) [reply EIPS/eip-9999.md:70]: Yeah agree with that. Let's put the list in the body and a commitment in the header.

File: EIPS/eip-9999.md (hunk 4 above)
Comment (nerolation) [reply EIPS/eip-9999.md:88]: Yeah, I'm working on a worst-case scenario analysis right now, but I think the random SSLOAD case will likely be the worst case (though, haven't looked into changes caused by balance diffs or contract code yet).

For random SLOADs, you would have a random address + a random storage key, acting as another 2x.

This is an interesting optimization problem because with pre and post values + balance diffs, maybe random calls sending 1 wei might be even more worst-case.

File: EIPS/eip-9999.md (hunk 5 above)
Comment (nerolation) [reply EIPS/eip-9999.md:67]: I'm not entirely sure about that because if we treat everything as warm, we also allow many more reads and writes which again affects the size of the access list. Maybe we should treat everything cold and focus on moving towards a better, block-level pricing mechanism for storage.

File: EIPS/eip-9999.md [hunk 8]
```diff
@@ -0,0 +1,92 @@
+---
+eip: 9999
```
g11tech, line 2:
```suggestion
eip: 7928
```

EIPS/eip-7928.md
```diff
@@ -0,0 +1,171 @@
...
+## Abstract
+
+This EIP proposes adding Block-Level Access Lists (BALs). By including a complete list of all addresses and storage keys accessed during a block, along with their post-execution values for writes, we enable parallel disk reads and parallel transaction validation. This improves execution efficiency and accelerates block verification, potentially allowing for future gas limit increases.
+
+## Motivation
+
+Currently, transactions without an explicit transaction access list cannot be efficiently parallelized, as the execution engine cannot determine in advance which addresses and storage slots will be accessed. While transaction-level access lists exist via EIP-2930, they are not enforced, making it difficult to optimize execution pipelines.
+
+By enforcing access lists at the block level, we shift the responsibility of declaring storage accesses to the block builder. This enables to efficiently parallelize both disk reads and transaction execution, knowing in advance the exact scope of storage interactions for each transaction.
+
```
jsign, line 22: Might be useful to write a sentence with some TL;DR of your research on tx parallelization opportunity in mainnet? Given you did the research with real data, I think can give a less theoretical (i.e., stronger) argument on why parallelizing is possible. (e.g., in mainnet blocks between A and B, usually X% of txs are parallelizable, etc)

EIPS/eip-7928.md
```diff
@@ -0,0 +1,171 @@
...
+## Backwards Compatibility
+
+This proposal requires changes to the block structure that are not backwards compatible and require a hard fork.
+
+## Security Considerations
+
+### Validation Overhead
+
+Nodes must verify that the access lists and balance diffs in the block accurately reflect the actual accesses and balance changes during execution. Discrepancies MUST result in block invalidation.
```
jsign, line 159: Worth mentioning the BAL should be canonical?
e.g., can't contain more info than needed.

bomanaps (commented):
Implementation of parallelization is mentioned in your motivation but not detailed in specification.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Abstract
+
+This EIP proposes adding Block-Level Access Lists (BALs). By including a complete list of all addresses and storage keys accessed during a block, along with their post-execution values for writes, we enable parallel disk reads and parallel transaction validation. This improves execution efficiency and accelerates block verification, potentially allowing for future gas limit increases.
+
+## Motivation
+
+Currently, transactions without an explicit transaction access list cannot be efficiently parallelized, as the execution engine cannot determine in advance which addresses and storage slots will be accessed. While transaction-level access lists exist via EIP-2930, they are not enforced, making it difficult to optimize execution pipelines.
+
+By enforcing access lists at the block level, we shift the responsibility of declaring storage accesses to the block builder. This enables to efficiently parallelize both disk reads and transaction execution, knowing in advance the exact scope of storage interactions for each transaction.
```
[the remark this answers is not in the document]
nerolation: yeah agree, 2 good points!

EIPS/eip-7928.md
```diff
@@ -0,0 +1,171 @@
...
+```python
+# Type aliases
+Address = ByteVector(20)
+StorageKey = ByteVector(32)
+StorageValue = ByteVector(32)
+TxIndex = uint16
+
+# Constants; chosen to support a 630m block gas limit
+MAX_TXS = 30_000
+MAX_SLOTS = 300_000
```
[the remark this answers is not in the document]
nerolation: yeah agree, I think we can raise those constants even more since the max of a list doesn't increase the list's size in SSZ.

EIPS/eip-7928.md
nerolation: agree, good point, will add.

EIPS/eip-7928.md
nerolation: agree, will add!

EIPS/eip-7928.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Motivation
+
+Currently, transactions without an explicit transaction access list cannot be efficiently parallelized, as the execution engine cannot determine in advance which addresses and storage slots will be accessed. While transaction-level access lists exist via EIP-2930, they are not enforced, making it difficult to optimize execution pipelines.
+
+By enforcing access lists at the block level, we shift the responsibility of declaring storage accesses to the block builder. This enables to efficiently parallelize both disk reads and transaction execution, knowing in advance the exact scope of storage interactions for each transaction.
+
+BALs map transactions to `(address, storage key, value)` tuples and include balance diffs. This approach facilitates parallel disk reads and parallel execution, reducing execution time to `parallel IO + parallel EVM` and improving overall network performance.
```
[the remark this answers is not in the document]
nerolation: yeah agree, will move those parts

octavio12345300 (approved):

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+For writes, each `SlotAccess` MUST contain an ordered list of transaction indices that accessed this key, and the final storage value after the last access. Transactions with writes that do not change the storage value MUST NOT contain a `value_after`.
+For contract deployments, the `value_after` MUST contain the runtime bytecode of the newly deployed contract.
+
+For reads, each `SlotAccess` MUST contain an empty `accesses` list.
+
+The `BalanceDiffs` structure tracks every address with a balance change, including transaction senders, recipients, and the block's coinbase address. 
+Each entry MUST include the transaction index and the signed balance delta per address for each transaction.
+11 bytes are sufficient to represent the total ETH supply.
+
+The `NonceDiffs` structure MUST record the post-block nonce values for all `CREATE2` deployer accounts and the deploayed contracts in the block.
```
jochem-brouwer, line 96: For CREATE and CREATE2 the nonce is bumped of an account. This is a good point because for CREATE we need the account nonce in order to calculate the to-be-created address (so the new account in which the intitcode of the CREATE runs). It currently says `post-block`, I assume you mean `post-tx` here?

We can also likely optimize this structure a bit, since CREATE runs in a touched account we can make an optional field which updates the nonce and includes the txs win which this nonce is updated. (I am pretty new to SSZ: is it possible to create optional/null fields/values in the data structure?). So for these nonce updates we know that:

- The account is already in the access list (or if its not, add it, but I can't think of a practical situation where this is possible)
- The nonce update list should thus only include the total nonce bump and the txs where this happens.

All other cases should be covered as "default behavior" e.g. tx nonce updates, or if for a contract the nonce update is not included it thus stays the same.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+        raise InvalidBlock("Mismatch in block-level access list.")
+        
+    if not validate_balance_diffs(block.balance_diffs, computed_balance_diffs):
+        raise InvalidBlock("Mismatch in balance diffs.")
+        
+    if not validate_nonce_diffs(block.nonce_diffs, computed_nonce_diffs):
+        raise InvalidBlock("Mismatch in nonce diffs.")
+```
+
+The BAL MUST be complete and accurate. It MUST NOT contain too few entries (missing accesses) or too many entries (spurious accesses). Any missing or extra entries in the access list, balance diffs, or nonce diffs SHALL result in block invalidation.
```
jochem-brouwer, line 140: I'm not sure what the correct term is but I think "complete" covers it: all items in the list are touched within the block, and items in the block which are touched are also in the list.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+            if is_write:
+                access_entry["value_after"] = value
+            
+            computed_access_list[(address, slot)].append(access_entry)
+        
+        # Process balance changes
+        for address, delta in balance_changes:
+            if address not in computed_balance_diffs:
+                computed_balance_diffs[address] = []
+            computed_balance_diffs[address].append({"tx_index": idx, "delta": delta})
```
jochem-brouwer, line 127: All looks fine but we can likely take out a lot of items for default behavior. (Of course we should still account for the worst case scenario).

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+11 bytes are sufficient to represent the total ETH supply.
+
+The `NonceDiffs` structure MUST record the post-block nonce values for all `CREATE2` deployer accounts and the deploayed contracts in the block.
+
+### State Transition Function
+
+Modify the state transition function to validate the block-level access lists:
+
+```python 
+def state_transition(block: Block) -> None:
```
jochem-brouwer, line 103: Looks fine but we can likely take out a lot of items for default behavior. (Of course we should still account for the worst case scenario).

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Rationale
+
+### BAL Design Choice
+
+We chose the BAL variant for several key reasons:
+
+1. **Balance between size and parallelization benefits**: BALs enable both parallel disk reads and parallel EVM execution while keeping the size manageable.
+
+2. **Storage value inclusion**: Including post-execution values for write operations facilitates state reconstruction during syncing, enabling faster chain catch-up.
```
jochem-brouwer, line 154: Wait, WOW this is actually a super good point! So this is super related to the diff layer of the snap sync! Wow I had not realized this, will keep this in mind, maybe most of the logic we want here is already available at the devp2p layer (snap sync).

Note: the snap sync diff layer only diffs at the block level, not at tx level. So with the snap sync diff layer we do not get the benefit of the tx parallelization.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+3. **Balance diffs**: Including balance diffs ensures correct handling of execution that depends on account balances, crucial for parallel transaction execution.
+
+4. **Nonce diffs**: Providing pre-block nonce values for transaction senders enables validators to verify nonce ordering without sequential execution.
```
jochem-brouwer, line 158: Ok, maybe I got the nonce part here wrong. Note that for transaction 0 in the block, which has nonce N in the transction and is sent from A, we should fetch from disk account A and verify that it's nonce is N. We now also know that after this tx it is N+1. There is (currently) no way to edit the nonce during the tx (but with 7702 there will be, one could call into an account and then CREATE/CREATE2 there bumping the nonce). So nonce updates should be very predictable.

The ways to update a nonce:

- Send a transaction
- Execute a CREATE/CREATE2 operation
- Change the delegation of an account via 7702 authority lists

I think only (2) is unpredictable. The remainder can be pre-fetched and checked from disk.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+5. **Efficiency**: An analysis of historical blocks (random sample of 1000 blocks between 22,195,599 and 22,236,441) shows that BALs would have had an average size of around 57 KiB with balance diffs adding only 9.6 KiB on average, which is a reasonable overhead given the performance benefits.
+
+6. **Transaction independence**: Analysis of the same block sample revealed that approximately 60-80% of transactions within a block are fully independent from one another, meaning they access disjoint sets of addresses and storage slots.
```
jochem-brouwer, line 162: I can believe this is correct but does not sound directly believable, since the majority of the txs call into super popular contracts (thus they share calls to the same address i.e. not disjoint, a call to Uniswap for instance). (But, in practice it should thus not touch any storage which previous txs have also touched)

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Block Size Considerations
+
+Including access lists increases block size, potentially impacting network propagation times. Based on analysis of historical blocks:
+
+- Average BAL size over 1,000 blocks was around 57 KiB (SSZ-encoded)
+- Average balance diff size was approximately 9.6 KiB
+- Worst-case BAL size for consuming the entire block gas limit with storage access operations would be approximately 0.93 MiB
```
jochem-brouwer, line 170: These statements are fine but should be motivated. This motivation itself could be an appendix (can include other files in EIPS and link to those! See for instance appendices/supplements of EIP 2537 (BLS precompiles): https://github.com/ethereum/EIPs/blob/master/assets/eip-2537/test-vectors.md

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+This proposal requires changes to the block structure that are not backwards compatible and require a hard fork.
+
+## Security Considerations
+
+### Validation Overhead
+
+Nodes MUST verify that the access lists and balance diffs in the block accurately reflect the actual accesses and balance changes during execution. Discrepancies MUST result in block invalidation.
+
+### Block Size
```
jochem-brouwer, line 189: This will definitely be the major point and there should be a very thorough analysis on what the worst case will be. If it is less than X KiB on average then this is very nice, but if the worst-case scenario produces a 100MiB block then this will never get accepted. (So analysis should be included, and should convince everyone that all cases are covered and thus the mentioned worst case scenario is thus really the worst case scenario)

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+Including comprehensive access lists, balance diffs and nonce values increases block size, potentially impacting network propagation times. However, as noted in the rationale section, the overhead is reasonable given the performance benefits, with typical BALs averaging around 67 KiB in total.
+
+### State Sync Benefits
+
+The inclusion of post-execution values for writes in BALs provides an additional benefit for state syncing. Nodes can use these values to reconstruct state without processing all transactions, verifying correctness by comparing the derived state root to the head block's state root.
```
jochem-brouwer, line 195: Yes this is an awesome point which I did not realize so far..! Something to think about a bit more! :thinking:

jochem-brouwer (commented):
Some more comments! :smile: :+1:

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Motivation
+
+Currently, transactions without an explicit transaction access list cannot be efficiently parallelized, as the execution engine cannot determine in advance which addresses and storage slots will be accessed. While transaction-level access lists exist via [EIP-2930](./eip-2930.md), they are not enforced, making it difficult to optimize execution pipelines.
+
+We propose enforcing access lists at the block level and shifting the responsibility of their creation to the block builder. This enables to efficiently parallelize both disk reads and transaction execution, knowing in advance the exact scope of storage interactions for each transaction.
+
+BALs map transactions to `(address, storage key, value)` tuples and include balance diffs. This approach facilitates parallel disk reads and parallel execution, reducing execution time to `parallel IO + parallel EVM` and improving overall network performance.
```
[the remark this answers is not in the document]
nerolation: yeah good point, will add that

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Specification
+
+### Block Structure Modification
+
+We introduce three new components in the block body:
+
+1. A Block Access List (BAL) that maps transactions to accessed addresses and storage slots, transaction indices, and post-execution values for writes.
```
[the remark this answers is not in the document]
nerolation: yeah agree. The intention was to "spell out" the aggregatable SSZ object, how it's design right now. But I see how this is confusing.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+MAX_ACCOUNTS = 300_000
+
+# SSZ containers
+class PerTxAccess(Container):
+    tx_index: TxIndex
+    value_after: StorageValue # value in state after the last access
+
+class SlotAccess(Container):
+    slot: StorageKey
+    accesses: List[PerTxAccess, MAX_TXS] # empty for reads
```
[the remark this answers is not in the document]
nerolation: Interestingly, you might not even need the indix.
At time of execution, you would have the value in cache (no matter if from IO or from another value_after), so, for reads, we actually don't care about the index of the transaction.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+class AccountNonce(Container):
+    address: Address  # account address
+    nonce_after: Nonce  # nonce value after block execution
+
+NonceDiffs = List[AccountNonceDiff, MAX_TXS]
+```
+
+The `BlockAccessList` is a deduplicated list of accessed addresses. For each address, it MUST contain a list of accessed storage keys. 
+For writes, each `SlotAccess` MUST contain an ordered list of transaction indices that accessed this key, and the final storage value after the last access. Transactions with writes that do not change the storage value MUST NOT contain a `value_after`.
+For contract deployments, the `value_after` MUST contain the runtime bytecode of the newly deployed contract.
```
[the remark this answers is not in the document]
nerolation: Oh, yeah, thanks!

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+For reads, each `SlotAccess` MUST contain an empty `accesses` list.
+
+The `BalanceDiffs` structure tracks every address with a balance change, including transaction senders, recipients, and the block's coinbase address.
```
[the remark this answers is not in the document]
nerolation: Theoretically, yes but practically almost no transaction pays exactly the gas used specified and most get some gas left back.
For the coinbase, >90% of block (those that come over mevboost) contain a MEV-Boost payment which is a transaction to the feeRecipient (the proposer) account. Thus, in 9 out of 10 cases, the coinbase account wouldn't be the proposer but the builder (since the builder address is set as coinbase). Searchers usually pay builders using direct calls to the block.coinbase.
The case with recipients might occur more often and will definetely bring some efficiency gains.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+The `BlockAccessList` is a deduplicated list of accessed addresses. For each address, it MUST contain a list of accessed storage keys. 
+For writes, each `SlotAccess` MUST contain an ordered list of transaction indices that accessed this key, and the final storage value after the last access. Transactions with writes that do not change the storage value MUST NOT contain a `value_after`.
+For contract deployments, the `value_after` MUST contain the runtime bytecode of the newly deployed contract.
+
+For reads, each `SlotAccess` MUST contain an empty `accesses` list.
+
+The `BalanceDiffs` structure tracks every address with a balance change, including transaction senders, recipients, and the block's coinbase address. 
+Each entry MUST include the transaction index and the signed balance delta per address for each transaction.
+11 bytes are sufficient to represent the total ETH supply.
```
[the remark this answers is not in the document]
nerolation: Will fix!

EIPS/eip-7928.md
nerolation: yeah the STF is still a rough draft. I will update it as soon as I get to spec'ing it.

EIPS/eip-7928.md
nerolation: Yeah right, the only thing that snap sync gives you that such state diffs don't is a proof that the updated state corresponds to the root in the block. However, we don't get actual execution proofs.
The state diffs can be applied during sync and we need at least post-block values for catching up.

EIPS/eip-7928.md
nerolation: Right, this is an artifact from an older version (where pre-executio values were considered). This point just needs to be removed.

EIPS/eip-7928.md
nerolation: same address != same (address, storage slot) though, and while many read the same slots, they don't write.
You sending WETH to me is disjoint from X sending WETH to Y.

EIPS/eip-7928.md
nerolation: The analysis I did in the shared post were already realistic and conservative. As soon as we settle on a design, I will start more thorougly analysing it. 
Still, I'm very certain we will be able to keep worst-case sizes below 1 MiB and the goal should be to stay below the calldata max, not because that's a magical boundary but more because we won't create another source of size-related bugs.

EIPS/eip-7928.md
```diff
@@ -0,0 +1,199 @@
...
+
+# Constants; chosen to support a 630m block gas limit
+MAX_TXS = 30_000
+MAX_SLOTS = 300_000
+MAX_ACCOUNTS = 300_000
+
+# SSZ containers
+class PerTxAccess(Container):
+    tx_index: TxIndex
+    value_after: StorageValue # value in state after the last access
```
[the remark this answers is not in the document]
nerolation: see comment above

EIPS/eip-7928.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Backwards Compatibility
+
+This proposal requires changes to the block structure that are not backwards compatible and require a hard fork.
+
+## Security Considerations
+
+### Validation Overhead
+
+Nodes MUST verify that the access lists and balance diffs in the block accurately reflect the actual accesses and balance changes during execution. Discrepancies MUST result in block invalidation.
```
g11tech, line 194: this is not a security consideration, move to specification segment

EIPS/eip-7928.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Block Size
+
+Including comprehensive access lists, balance diffs and nonce values increases block size, potentially impacting network propagation times. However, as noted in the rationale section, the overhead is reasonable given the performance benefits, with typical BALs averaging around 67 KiB in total.
```
g11tech, line 198:
```suggestion
Including comprehensive access lists, balance diffs and nonce values increases block size, potentially opening up a grieving attack vector impacting network propagation times and blockchain liveness. However, as noted in the rationale section, the overhead is reasonable given the performance benefits, with typical BALs averaging around 67 KiB in total.
```

EIPS/eip-7928.md
g11tech: move in motivation/spec section but out of security consideration section

EIPS/eip-7928.md
```diff
@@ -93,9 +93,9 @@ The `code` field in `AccountAccess` is optional and MUST be present only when a
 
 For reads, each `SlotAccess` MUST contain an empty `accesses` list.
 
-The `BalanceDiffs` structure tracks every address with a balance change, including transaction senders, recipients, and the block's coinbase address. 
+The `BalanceDiffs` structure tracks every address with a balance change, including transaction senders, recipients, and the block's coinbase address. Touched accounts without balance changes MUST be omitted.
 Each entry MUST include the transaction index and the signed balance delta per address for each transaction.
-11 bytes are sufficient to represent the total ETH supply.
+12 bytes are sufficient to represent the total ETH supply.
```
jochem-brouwer, line 98: Maybe for completeness `2^(12*8) / 1e18 / 1e6` (`1e18` = 1 eth) = approx 79228 million ETH (which is much higher than current supply of ~121 million.

Also note: we have to encode the balance diffs as integer. So one of the bits is used as sign in the balance diffs. For 11 bytes we thus have `11*8-1 = 87` bits, which yields `2^87/1e18/1e6` = 154 million. Which is close-ish to the total supply currently. (Of course in practice we will not see a balance diff anywhere of the total supply, but if we are well above it then we can state that balance diffs thus always fit, without exceptions)

EIPS/eip-7928.md
```diff
@@ -30,7 +30,7 @@ We introduce three new components in the block body:
 
 1. A Block Access List (BAL) that maps transactions to accessed addresses and storage slots, transaction indices, and post-execution values for writes.
 2. Balance Diffs that track every address touched by value transfers along with the balance deltas.
-3. Nonces that record the pre-block nonces of all transaction senders within the block.
+3. Nonces that record the post-block nonces of contracts using `CREATE` or `CREATE2` within the block.
```
jochem-brouwer, line 33: I think we need pre-block nonces for the contracts which use CREATE, since the address calculation algorithm uses the nonce of the contract then. So we will need it in the BAL to calculate the target address. CREATE2 will also update the nonce, so if a contract first CREATE2 and then CREATEs then we would also need the pre-block nonce of that contract. (We can optimize the BALs later to remove nonce entries which we do not need. For instance if a contract CREATE2s we do not need the nonce)

For completeness here the two algorithms to calculate the addresses for CREATE/CREATE2

CREATE: `return keccak256(RLP.encode([from, nonce])).subarray(-20)`

CREATE2: `keccak256(concatBytes(hexToBytes('0xff'), from, salt, keccak256(initCode)))`

In both: `from` is the current address of the contract doing CREATE(2), and for CREATE2 the `salt` is an item popped from stack. The `initCode` is the code (read from evm memory) to be ran once attempting to create the contract

jochem-brouwer (commented):
Two comments on the latest changes :smile: :+1:

nerolation:
Great, incorporated all the feedback - thanks everyone!

Would be great to have it merged now. It's ready for a first draft.

g11tech (approved):

Muaytie666 (approved):

Muaytie666 (approved):

Cassieminkus1 (approved):

Mrana5398 (commented):
Hello

PR #9589, opened 2025-04-03
Update EIP-5920: add clarification to ban opcode in STATIC mode

jochem-brouwer:
This PR updates the PAY opcode PR such that it is self-consistent: currently, there is no rule about what happens in a `STATIC` context (`STATIC` context/flag: https://eips.ethereum.org/EIPS/eip-214#specification). It is clear that using PAY in a `STATIC` context is usually not allowed (it changes state), but what about the case if the value sent is 0? I have followed the way that CALL handles this: if the value is zero, it is fine: if the value is nonzero and we are in a `STATIC` context then exceptionally abort.


EIPS/eip-5920.md
```diff
@@ -45,6 +45,7 @@ A new opcode is introduced: `PAY` (`0xfc`), which:
 - Pops two values from the stack: `addr` then `val`.
 - Charges the gas cost detailed below.
 - Exceptionally halts if `addr` has any of the high 12 bytes set to a non-zero value (i.e. it does not contain a 20-byte address).
+- Exceptionally halts if `val` is nonzero if the `STATIC` flag of the current call frame is `true`
```
shemnon, line 48: I don't think we should complicate it with wether `val` is used or not.  There are better non-writing ways to "warm" a target address.

```suggestion
- halt with exceptional failure if the current frame is in `static-mode`.
```

jochem-brouwer: Ok, this is also fine by me. I copied the behavior of call, which allows calls if no value is sent (this is of course used in static contexts to call libraries for instance). I also used `STATIC` flag which is the original EIP terminology. But I think `static-mode` is also clear for every EVM developer.

jochem-brouwer: Actually, indeed, also for the warm use case: `BALANCE` is clearly used to read, and `PAY` is clearly used to write. So `PAY` should not be used in a static context.

EIPS/eip-5920.md
```diff
@@ -42,6 +42,7 @@ Currently, to send ether to an address requires you to call into that address, w
 
 A new opcode is introduced: `PAY` (`0xfc`), which:
 
+- Halt with exceptional failure if the current frame is in `static-mode`.
```
wjmelements, line 45: This diction matches EIP-7069 and so this version of the text should be approved.

wjmelements (approved):

jochem-brouwer:
Can we merge this?  :smile: :+1: 

@Pandapip1 @xinbenlv @SamWilsn

jochem-brouwer:
This is now part of https://github.com/ethereum/EIPs/pull/9590

PR #9590, opened 2025-04-03
Update EIP-5920: add status stack return item / improve gas schedule

jochem-brouwer:
This PR updates EIP-5920:

- Add extra motivation why this is necessary to prevent over-spending gas in Prague (for ETH transfers to 7702-delegates EOAs)
- Clarifies the opcode rules by explicitly banning it in `static-mode` (in a `STATICCALL` frame)
- Pushes an item to stack to signal succesful PAYs or not (only unsuccessful case is if the `val` is higher than current balance). Success is 0. Failure case is 1. Therefore this is also future-proof if there are other reasons in the future to signal for failure. Compilers can now check for `0` for success, any non-zero is thus unsuccessful.
- Updates the EIP with clarifications and makes certain sentences "more correct" (obviously up to review)


EIPS/eip-5920.md
```diff
@@ -43,10 +43,12 @@ Currently, to send ether to an address requires you to call into that address, w
 A new opcode is introduced: `PAY` (`0xfc`), which:
 
 - Pops two values from the stack: `addr` then `val`.
-- Charges the gas cost detailed below.
 - Exceptionally halts if `addr` has any of the high 12 bytes set to a non-zero value (i.e. it does not contain a 20-byte address).
-- Transfers `val` wei from the current target address to the address `addr`.
+- Charges the gas cost detailed below.
 - Marks `addr` as warm (adding `addr` to `accessed_addresses`).
+- Transfers `val` wei from the current target address to the address `addr`.
+- Pushes `1` to the stack if the transfer succeeds, or `0` if it fails.
```
pdobacz, line 50: This is consistent with `*CALL` behavior in legacy EVM, but not with the planned `EXT*CALL` opcodes in EOF [see 7069](https://eips.ethereum.org/EIPS/eip-7069), which has richer codes and returns 0 on success.

Would PAY's return codes differ when called in EOF vs legacy?

jochem-brouwer: No, good point, I will change. But then I will mark this EIP as EOF only

jochem-brouwer: I forgot to do this, sorry. I have just pushed a commit which should reflect the changes in the way that they reflect what happens if you CALL with value and the value is more than the current balance OR similar for EXTCALL. Could you double check the values I push to stack?

jochem-brouwer: I have also decided against making it EOF-only. If we want that, then that should be a different PR.

EIPS/eip-5920.md
```diff
@@ -43,10 +43,17 @@ Currently, to send ether to an address requires you to call into that address, w
...
+- Charges the gas cost detailed below.
 - Marks `addr` as warm (adding `addr` to `accessed_addresses`).
+- Transfers `val` wei from the current target address to the address `addr`.
+- Push a value to the stack signalling the success or failure of the `PAY` operation
+  - This in inspired by the behavior of the `CALL` or `EXTCALL` opcodes. It is used to signal success or failure of the payment.
+- The transfer only fails if `val` is higher than the current balance of the account. In that case, balances are not changed.
+  - If the `PAY` opcode is used in a legacy context:
+    - Pushes `1` to the stack if the transfer succeeds, or `0` if it fails.
+  - If the `PAY` opcode is used in the EOF context:
+    - Pushes `0` to the stack if the transfer succeeds, or `2` if it fails.
```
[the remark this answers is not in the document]
jochem-brouwer: Aha, ok I was not aware of this EXTCALL rule. This is different from CALL I think because that would OOG and that would thus put 2 on stack. I will change.

I should join the EOF implementation calls, I will try to do so in the future and let my presence know if I'm there (or not).

I can see why the behavior between legacy/EOF contracts being different is a bit awkward. What was the result of the call? Keep the stack return value for legacy/EOF consisntent? And what should this then push to stack?

I think because this is a new opcode we can indeed make the return value consistent. I think the EOF way is preferred (I do not recall the exact reasoning but I think it was the branching rule that we want to put 0 on stack for success so if we RJUMPI later the "happy" case will not RJUMP). 

I will address the 2->1 change but I will await the discussion on if we want to align legacy/EOF here (I could not find a decision in the magicians post you linked)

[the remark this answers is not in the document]
pdobacz: > I will address the 2->1 change but I will await the discussion on if we want to align legacy/EOF here (I could not find a decision in the magicians post you linked)

Not a decision, just argument for alignment:

> Frangio thinks differing impls based on context is a bad idea, and that users will just need to keep in mind what is happening.

EIPS/eip-5920.md
```diff
@@ -68,7 +75,7 @@ The gas cost for `PAY` is the sum of the following:
 
 ### Argument order
 
-The order of arguments mimics that of `CALL`, which pops `addr` before `val`. Beyond consistency, though, this ordering aids validators pattern-matching MEV opportunities, so `PAY` always appears immediately after `COINBASE`.
+The order of arguments mimics that of `CALL`, which pops `addr` before `val`. Beyond consistency, though, this ordering aids validators pattern-matching MEV opportunities, so `PAY` always appears immediately after `COINBASE`. The choice for pushing `1` or `0` to the stack is also made to be consistent with the `CALL` opcode. For EOF, it pushes either `0` or `2` to the stack to be consistent with the `EXTCALL` opcode.
```
gumb0, line 78:
```suggestion
The order of arguments mimics that of `CALL`, which pops `addr` before `val`. Beyond consistency, though, this ordering aids validators pattern-matching MEV opportunities, so `PAY` always appears immediately after `COINBASE`. The choice for pushing `1` or `0` to the stack is also made to be consistent with the `CALL` opcode. For EOF, it pushes either `0` or `1` to the stack to be consistent with the `EXTCALL` opcode.
```

EIPS/eip-5920.md
```diff
@@ -2,7 +2,7 @@
 eip: 5920
 title: PAY opcode
 description: Introduces a new opcode, PAY, to send ether to an address without calling any of its functions
-author: Gavin John (@Pandapip1), Zainan Victor Zhou (@xinbenlv), Sam Wilson (@SamWilsn)
+author: Gavin John (@Pandapip1), Zainan Victor Zhou (@xinbenlv), Sam Wilson (@SamWilsn), Jochem Brouwer (@jochem-brouwer)
```
jochem-brouwer, line 5: I have added myself here because I will push for this in Fusaka.

EIPS/eip-5920.md
```diff
@@ -42,11 +43,19 @@ Currently, to send ether to an address requires you to call into that address, w
...
 
+- Halt with exceptional failure if the current frame is in `static-mode`.
 - Pops two values from the stack: `addr` then `val`.
-- Charges the gas cost detailed below.
 - Exceptionally halts if `addr` has any of the high 12 bytes set to a non-zero value (i.e. it does not contain a 20-byte address).
-- Transfers `val` wei from the current target address to the address `addr`.
+- Charges the gas cost detailed below.
 - Marks `addr` as warm (adding `addr` to `accessed_addresses`).
+- Transfers `val` wei from the current address to the address `addr`, only if the current address has a balance higher than or equal to `val`.
+- Push a value to the stack signalling the success or failure of the `PAY` operation
```
pcaversaccio, line 52: Having a `success` flag could be beneficial; in Vyper the built-in [`send`](https://docs.vyperlang.org/en/stable/built-in-functions.html#send) function could be modified in the following way:

```py
send(to: address, value: uint256, revert_on_failure: bool = True) -> None
```

If `revert_on_failure=False` it returns an optional `success` Boolean, which the user can `assert` himself _or_ can e.g. push to some return array in the function he uses. Example:

```vy
struct BatchValue:
    target: address
    allow_failure: bool
    value: uint256

@internal
@payable
def _multicall_value(data: DynArray[BatchValue, 255]) -> DynArray[bool, 255]:
    value_accumulator: uint256 = empty(uint256)
    for batch: BatchValue in data:
        success: bool = empty(bool)
        msg_value: uint256 = batch.value
        value_accumulator = value_accumulator + msg_value
        if batch.allow_failure:
            success = send(batch.target, msg_value, revert_on_failure=False)
        else:
            success = True
            send(batch.target, msg_value)
        results.append(success)
    assert msg.value == value_accumulator, "multicall: value mismatch"
    return results
```

jochem-brouwer:
Upon some feedback have to make some slight changes to this proposal, so have converted back to draft.

jochem-brouwer:
Have changed the stack return value and the gas costs, please review :smile: :+1: Thanks!

EIPS/eip-5920.md
```diff
@@ -55,20 +65,18 @@ The gas cost for `PAY` is the sum of the following:
...
     - Otherwise, `COLD_ACCOUNT_ACCESS_COST`.
-- Does `addr` exist or is `val` zero?
-    - If yes to either, zero;
-    - Otherwise, `GAS_NEW_ACCOUNT`.
-- Is `val` zero?
-    - If yes, zero;
-    - Otherwise, `GAS_CALL_VALUE`.
+- Is `val` non-zero?
+  - Charge `GAS_CALL_VALUE`
+  - Is `addr` non-existant?
```
SanLeo461, line 70:
```diff
-  - Is `addr` non-existant?
+  - Is `addr` non-existent?
```

EIPS/eip-5920.md
```diff
@@ -55,20 +65,18 @@ The gas cost for `PAY` is the sum of the following:
...
+      - Charge `GAS_NEW_ACCOUNT`
 
 `PAY` cannot be implemented on networks with empty accounts (see [EIP-7523](./eip-7523.md)).
 
 ## Rationale
 
 ### Argument order
 
-The order of arguments mimics that of `CALL`, which pops `addr` before `val`. Beyond consistency, though, this ordering aids validators pattern-matching MEV opportunities, so `PAY` always appears immediately after `COINBASE`.
+The order of arguments mimics that of `CALL`, which pops `addr` before `val`. Beyond consistency, though, this ordering aids validators pattern-matching MEV opportunities, so `PAY` always appears immediately after `COINBASE`. The choice for pushing `1` or `0` to the stack is also made to be consistent with the `CALL` opcode. For EOF, it pushes either `0` or `1` to the stack to be consistent with the `EXTCALL` opcode.
```
SanLeo461, line 79: > The choice for pushing `1` or `0` to the stack is also made to be consistent with the `CALL` opcode. For EOF, it pushes either `0` or `1` to the stack to be consistent with the `EXTCALL` opcode.

This is an extremely confusing sentence with regards to earlier
>   - If the operation succeeds, push `0`
>  - If the operation fails, push `1`
>  - This is different from normal stack push value of `*CALL*` opcodes, which signals `0` on failure or `1` on success
>    - It is choosen here to signal `0` for success, and `1` for failure
>      - It is only possible here to push `1` for a balance failure
>      - In the future, it could be possible that for other reasons (such as accounts not directly accepting ETH) the PAY operation fails
>      - Therefore, nonzero stack return values are reserved for failure reasons. Checking for the `0` stack value for success is thus future-proof.

In non-EOF, does this mean that `1` is success, `0` is failure, and anything else is also failure?
I'm in favour of standardizing as `0` is success and anything else is failure, both within and without EOF.

jochem-brouwer: Sorry, this sentence had to be removed with the new stack update succes -> 0 and failure -> nonzero. There (should) now also no be different behavior for EOF in the EIP.

jochem-brouwer: https://github.com/ethereum/EIPs/pull/9590/commits/f836701170d3c1a95eba01344ac03f22cf210592

EIPS/eip-5920.md
jochem-brouwer: Thanks, https://github.com/ethereum/EIPs/pull/9590/commits/018b296ff252f7d02f0f4dff99dfb8c8eb133592

EIPS/eip-5920.md
```diff
@@ -19,10 +19,13 @@ This EIP introduces a new opcode, `PAY`, taking two stack parameters, `addr` and
...
+- First of all, it opens a reentrancy attack vector, as the recipient can call back into the sender. More generally, the recipient can unilaterally execute arbitrary state changes, limited only by the gas forwarded, which is not desirable from the point of view of the sender.
 - Secondly, it opens a DoS vector. Contracts which want to send ether must be cognizant of the possibility that the recipient will run out of gas or revert.
-- The `EXTCALL` opcode does not provide a way to restrict gas.
-- Finally, the `CALL` opcode is needlessly expensive for simple ether transfers, as it requires the memory and stack to be expanded, the recipient's full data including code and memory to be loaded, and finally needs to execute a call, which might do other unintentional operations. Having a dedicated opcode for ether transfers solves all of these issues, and would be a useful addition to the EVM.
+- The `EXTCALL` opcode (introduced in [EOF](./eip-7069)) does not provide a way to restrict gas, which means sending ether via `EXTCALL` is subject to the reentrancy attack vector.
```
SanLeo461, line 24: For me at least, reentrancy is the least of my concern when sending ether without PAY. We know how to stop re-entrancy attacks easily now, either with the CEI pattern, or with re-entrancy guards.
The bigger concern IMO is being locked out of making certain types of systems, e.g. an auction denominated in ether, where the last highest bid gets returned to the corresponding bidder.
This is because a malicious bidder can just make a bid from a gas-griefing contract, which causes the amount of gas required to send ether to them to rise above the block gas limit.

This is often seen as an anti-pattern, and people recommend using a "pull-type" mechanism for such a thing, but from a UX perspective, users are not familiar at all with this pattern and would be beyond confused by it.

This doesn't just apply to an auction obviously, there are many usecases where you want user A's transaction to send ether to user B without being subject to gas-griefing, which I would cite as a key motivator, personally.

EIPS/eip-5920.md
```diff
@@ -42,11 +45,20 @@ Currently, to send ether to an address requires you to call into that address, w
...
 - Marks `addr` as warm (adding `addr` to `accessed_addresses`).
+- Transfers `val` wei from the current address to the address `addr`, only if the current address has a balance higher than or equal to `val`.
+- Push a value to the stack signalling the success or failure of the `PAY` operation
+  - If the operation succeeds, push `0`
+  - If the operation fails, push `1`
+  - This is different from normal stack push value of `*CALL*` opcodes, which signals `0` on failure or `1` on success
+    - It is choosen here to signal `0` for success, and `1` for failure
+      - It is only possible here to push `1` for a balance failure
+      - In the future, it could be possible that for other reasons (such as accounts not directly accepting ETH) the PAY operation fails
+      - Therefore, nonzero stack return values are reserved for failure reasons. Checking for the `0` stack value for success is thus future-proof.
```
SanLeo461, line 61: Probably make it clearer that the thinking is that we could preserve other nonzero return values (2, 3, etc.) to be for different reasons why the PAY operation failed.

EIPS/eip-5920.md
SanLeo461: Nevermind, didn't notice L23 in my review, maybe can just expand on L23 my point

EIPS/eip-5920.md
jochem-brouwer: How would you write this sentence? Is it also unclear in the context of this sentence? The line above this one is currently: "In the future, it could be possible that for other reasons (such as accounts not directly accepting ETH) the PAY operation fails" (which is where the "Therefore" reflects on). But I'd love to improve this such that the idea/message is clear :smile: :+1:

EIPS/eip-5920.md
jochem-brouwer: Yes, this lock mechanism is indeed a concern. If I would edit the previous sentence to:

"Secondly, it opens a DoS vector. Contracts which want to send ether must be cognizant of the possibility that the recipient will run out of gas or revert. Therefore, one must be careful with the amount of gas which is being forwarded, since all this gas could be consumed, leaving not much gas to use after attempting to send ether."

(This also nicely set ups the reasoning about the EXTCALL not being able to restrict gas forwarded :smile: )

EIPS/eip-5920.md
SanLeo461:
```
- It is chosen here to signal `0` for success, and `>=1` for failure
  - If  push `1` for a balance failure
  - (List any other possible cases, e.g. the currently discussed address high-bytes invalid vs. return stack value 2)
  - In the future, it could be possible that for other reasons (such as accounts not directly accepting ETH) the PAY operation fails
  - Therefore, other nonzero stack return values are reserved for failure reasons. Checking for the `0` stack value for success is thus future-proof.
```
Keyword here is `other nonzero`, instead of just `nonzero`, makes it clearer

Also note typo on L58 `choosen` -> `chosen`

jochem-brouwer: Have changed :smile: :+1:

SamWilsn (approved):

PR #9601, opened 2025-04-04
Update EIP-5792: Add errors and update atomicity capability

lukasrosario:
- update `atomic` capability to support 3 values:
  - `supported`: the wallet is currently able to and will submit batches atomically
  - `ready`: the wallet is able to upgrade to `supported` pending approval from the user
  - `unsupported`: the wallet cannot support calls atomically nor can does it have a mechanism for upgrading
  - updated `wallet_sendCalls` constraints accordingly
- added error codes
Screenshot 2025-04-04 at 2 44 33 PM


EIPS/eip-5792.md
```diff
@@ -358,17 +363,34 @@ type AtomicCapability = {
...
+### Error Codes
+
+The following error codes are defined for the methods specified in this EIP. These errors follow the JSON-RPC 2.0 specification for error objects, which require a `code` and `message` field.
+
+| Code | Message | Description | Related RPCs |
+|------|---------|-------------|--------------|
+| 4001 | User Rejected Request | The user rejected submitting the batch of calls (from [EIP-1193](./eip-1193.md)) | `wallet_sendCalls` |
+| 4100 | Unauthorized | The specified address is not connected or not in the wallet (from [EIP-1193](./eip-1193.md)) | `wallet_sendCalls`, `wallet_getCapabilities` |
+| 5700 | Unsupported non-optional capability | This wallet does not support a capability that was not marked as optional | `wallet_sendCalls` |
+| 5710 | Unsupported chain id | This wallet does not support the specified chain id | `wallet_sendCalls` |
```
jxom, line 386: Don't we already have `4901: ChainDisconnectedError` on EIP-1193?

EIPS/eip-5792.md
```diff
@@ -358,17 +363,34 @@ type AtomicCapability = {
...
+| 5720 | Malformed request | The wallet cannot parse this request (e.g., missing 0x prefix, leading zeros in chain id, request fails schema validation) | `wallet_sendCalls`, `wallet_getCallsStatus`, `wallet_showCallsStatus`, `wallet_getCapabilities` |
+| 5730 | Duplicate ID | There is already a bundle submitted with this id | `wallet_sendCalls` |
+| 5740 | Unknown bundle id | This bundle id is unknown / has not been submitted | `wallet_getCallsStatus`, `wallet_showCallsStatus` |
+| 5750 | Bundle too large | The call bundle is too large for the wallet to process | `wallet_sendCalls` |
+| 5760 | Atomic-ready wallet rejected upgrade | The wallet can support atomicity after an upgrade, but the user rejected the upgrade | `wallet_sendCalls` |
+| 5770 | Atomicity not supported | The wallet does not support atomic execution but the request requires it | `wallet_sendCalls` |
+
```
jxom, line 393: IMO, this seems like it should be in a separate ERC.

lukasrosario: i feel like the spec would be kind of incomplete without errors, so would prefer to keep it here but am open to opening another ERC to tie them to other RPCs. i feel like it would be odd that if a new wallet looking to implement would have to look elsewhere to know how to handle what will likely be pretty common errors.

think most of us can agree we can do a better job around errors throughout ethereum ecosystem, so want to make it as easy as possible to align around them.

of course am open to changing my mind though if other folks feel the same way! so would be good for the rest of the group to weigh in

EIPS/eip-5792.md
```diff
@@ -290,6 +293,7 @@ If any of these supplemental expressions of capabilities are contradicted by cap
 Capabilities are returned in key/value pairs, with the key naming a capability and a value conforming to a shape defined for that name, in an object keyed to the relevant [EIP-155](./eip-155.md) `chainId` expressed in hexadecimal notation.
 Capabilities are nested in per-chain objects because wallets may support different capabilities across multiple chains authorized in a given session.
 Capabilities that the wallet supports on all the chains SHOULD be included only once, using a special `chainID` value `"0x0"`, and SHOULD NOT be repeated in nested per-chain objects.
+If a wallet does not support a chain that is requested in the `wallet_getCapabilities` method, the wallet MUST NOT respnond with an error, and MUST instead not include a key for the unsupported chain(s) in the response.
```
cryptotavares, line 296:
```suggestion
If a wallet does not support a chain that is requested in the `wallet_getCapabilities` method, the wallet MUST NOT respond with an error, and MUST instead not include a key for the unsupported chain(s) in the response.
```

EIPS/eip-5792.md
```diff
@@ -358,17 +363,34 @@ type AtomicCapability = {
...
   "0x2105": {
     "atomic": {
-      "supported": true
+      "status": "supported"
     }
   },
   "0x14A34": {
     "atomic": {
-      "supported": false
+      "supported": "unsupported"
```
cryptotavares, line 371:
```suggestion
      "status": "unsupported"
```

EIPS/eip-5792.md
```diff
@@ -358,17 +363,34 @@ type AtomicCapability = {
...
 ```
 
+### Error Codes
+
+The following error codes are defined for the methods specified in this EIP. These errors follow the JSON-RPC 2.0 specification for error objects, which require a `code` and `message` field.
+
+| Code | Message | Description | Related RPCs |
+|------|---------|-------------|--------------|
+| 4001 | User Rejected Request | The user rejected submitting the batch of calls (from [EIP-1193](./eip-1193.md)) | `wallet_sendCalls` |
+| 4100 | Unauthorized | The specified address is not connected or not in the wallet (from [EIP-1193](./eip-1193.md)) | `wallet_sendCalls`, `wallet_getCapabilities` |
```
cryptotavares, line 384: [nit]: should the description match the EIP-1193 description for these errors?

EIPS/eip-5792.md
```diff
@@ -358,17 +363,34 @@ type AtomicCapability = {
...
+| 5700 | Unsupported non-optional capability | This wallet does not support a capability that was not marked as optional | `wallet_sendCalls` |
+| 5710 | Unsupported chain id | This wallet does not support the specified chain id | `wallet_sendCalls` |
+| 5720 | Malformed request | The wallet cannot parse this request (e.g., missing 0x prefix, leading zeros in chain id, request fails schema validation) | `wallet_sendCalls`, `wallet_getCallsStatus`, `wallet_showCallsStatus`, `wallet_getCapabilities` |
```
cryptotavares, line 387: Could we keep the `malformed request` error out of this list? It seems that this would be covered by the [json rpc spec `-32700 Parse Error`](https://www.jsonrpc.org/specification#error_object).

lukasrosario: is that the intention of that error? my understanding of that would be if there was invalid *json*, eg
```
{
 method: 'wallet_sendCalls' // missing comma, eg JSON.parse(...) would throw (i think)
 params: ...
}
```

jxom: I thought we already had -32000: InvalidInputError: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md#error-codes

cryptotavares: @lukasrosario Yes you are right. The Parse Error is a different thing, but what about the standard -32602: Invalid Params? Or even the -32000: Invalid Input one suggested above by @jxom?

lukasrosario: how widely adopted are those codes ? i see the eip is stagnant. will not die on this hill but i dont recall seeing these codes much & i think it would be a little odd to build off a technically stagnant EIP + those codes are negative lol

EIPS/eip-5792.md
bumblefudge: would the `(from [EIP-1193](./eip-1193.md))` be sent as part of the message? i assume not but, might be clearer to move that to another column

EIPS/eip-5792.md
bumblefudge: i think the "nonstandard" ones defined in that EIP are used by MM so at the very least colliding with them would be confusing to some devs that use MM

lukasrosario: updated to use -3602 invalid params

PR #9557, opened 2025-03-27
Update EIP-7692: add eof-devnet-2 EIPs

shemnon:
* Update the EOF meta EIP with devnet-2 schedule
* Add devnet-1 EIPs to SFI for Fuska meta-eip
* Add devnet-2 EIPs to CFI for Fusaka meta-eip

**ATTENTION: ERC-RELATED PULL REQUESTS NOW OCCUR IN [ETHEREUM/ERCS](https://github.com/ethereum/ercs)**

--

When opening a pull request to submit a new EIP, please use the suggested template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

We have a GitHub bot that automatically merges some PRs. It will merge yours immediately if certain criteria are met:

 - The PR edits only existing draft PRs.
 - The build passes.
 - Your GitHub username or email address is listed in the 'author' header of all affected PRs, inside <triangular brackets>.
 - If matching on email address, the email address is the one publicly listed on your GitHub profile.

PR #9615, opened 2025-04-08
Update EIP-7692: EOF - update scope of EOF with EIP-7873

pdobacz:
Makes the list of EIPs align with current plan for eof-devnet-1


shemnon:
see https://github.com/ethereum/EIPs/pull/9557

PR #9623, opened 2025-04-10
Update EIP-7892: Add blob count per tx limit via blobSchedule

flcl42:
### Adds blob count per tx limit via blobSchedule

Let's assume 72blob txs are allowed, in such case maximum tx size exceeds 10MiB.
- Tx pool requires to be adjusted for the worst case scenario of big load of such transactions, it is solvable by evicting enough low cost txs in case of limited storage, but such design can be complex to implement efficiently and may mean an additional dos vector
- Propagation is slower for such in case of limited throughput, still manageable. 3-4 seconds for the case 10MB/s, 10k nodes, 50 nodes connected
- Validation of network form of such tx requires 72+ms, which is not a big obstacle, still good to consider
- Another concern is spending the entire blob space by a single actor, which seems a bit unfair, still it is probably better not to additionally regulate blob space market

It does not seem like there is an obvious point of failure, especially if we spend more time on tx pool improvements. But in sum it effects confidence in ability to handle extreme load of such big transactions.
In case if the proposal is relevant, there are some ideas what can be the initial limit for 48/72 case:
- the most active rollups spend 1-2 blobs on average, let's 10x it in the nearest fork = 12 blobs
- more manageable tx message size seems to be like 2-3MiB, which = 18-20 blobs
- the safest option is to allow previous limit like = 9 blobs
Can it be like 12-15?


EIPS/eip-7594.md
```diff
@@ -60,7 +64,25 @@ The node MUST validate `tx_payload_body` and verify the wrapped data against it.
...
+### Configuration
+
+`blobSchedule` is required to be extended with an additional optional field `maxPerTx`:
+
+```json
+"blobSchedule": {
+  ...
+  "osaka": {
+    ...
+    "maxPerTx" : 6
```
MariusVanDerWijden:
```suggestion
    "maxBlobsPerTx" : 6
```

flcl42: Sounds better. `maxPerTex` was selected because we have field named `max` in `blobSchedule`

EIPS/eip-7594.md
```diff
@@ -60,7 +64,25 @@ The node MUST validate `tx_payload_body` and verify the wrapped data against it.
...
+### Configuration
+
+`blobSchedule` is required to be extended with an additional optional field `maxPerTx`:
+
+```json
+"blobSchedule": {
+  ...
+  "osaka": {
+    ...
+    "maxBlobsPerTx" : 6
```
macfarla:
```suggestion
    "maxPerTx" : 6
```

flcl42: 👍  Fixed at line 69 instead

jwasinger (approved):

fjl (approved):

benaadams:
Do we also need to invalidate block if tx is higher?

barnabasbusa:
I suggest we target these changes for [7892](https://eips.ethereum.org/EIPS/eip-7892) instead of 7594

EIPS/eip-7594.md
```diff
@@ -60,7 +64,25 @@ The node MUST validate `tx_payload_body` and verify the wrapped data against it.
...
+  }
+}
+```
+
+The field defines maximal allowed blob count per transaction.
```
Marchhill:
```suggestion
The field defines the maximum number of blobs allowed per transaction.
```

EIPS/eip-7594.md
```diff
@@ -60,7 +64,25 @@ The node MUST validate `tx_payload_body` and verify the wrapped data against it.
...
+
+Clients **MUST** consider this setting during validating transactions received from the network, during block building and validation.
```
Marchhill:
```suggestion
Clients **MUST** consider this setting when validating transactions received from the network, and when building or validating a block.
```

EIPS/eip-7594.md
```diff
@@ -60,7 +64,25 @@ The node MUST validate `tx_payload_body` and verify the wrapped data against it.
 - The KZG `commitments` hash to the versioned hashes, i.e. `kzg_to_versioned_hash(commitments[i]) == tx_payload_body.blob_versioned_hashes[i]`
 - The KZG `commitments` match the corresponding `blobs` and `cell_proofs`. This requires computing the extension cells for all `blobs`, and verifying all `cell_proofs`. (Note: all cell proofs can be batch verified at once)
 
+### Configuration
+
+`blobSchedule` is required to be extended with an additional optional field `maxBlobsPerTx`:
```
Marchhill:
```suggestion
`blobSchedule` is extended with an additional optional field `maxBlobsPerTx`:
```

EIPS/eip-7594.md
```diff
@@ -21,6 +21,8 @@ DAS is a method of scaling data availability beyond the levels of [EIP-4844](./e
 
 Providing additional data availability helps bring scale to Ethereum users in the context of layer 2 systems called "roll-ups" whose dominant bottleneck is layer 1 data availability.
 
+Limiting the number of blobs per transaction allows transaction pool to not become a new bottleneck and to be scaled independently of consensus layer.
```
Marchhill:
```suggestion
Limiting the number of blobs per transaction prevents the transaction pool from becoming a bottleneck, allowing it to be scaled independently of consensus layer.
```

Marchhill: does this mean scaling the txpool independently of the consensus layer?

EIPS/eip-7892.md
```diff
@@ -12,7 +12,7 @@ requires: 7840
 
 ## Abstract
 
-This EIP introduces **Blob Parameter Only (BPO) Hardforks**, a lightweight mechanism for incrementally scaling Ethereum’s blob capacity through targeted hard forks that modify only blob-related parameters: `blob target`, `blob limit`, and `baseFeeUpdateFraction`. Unlike traditional hard forks, which require extensive coordination and introduce broader protocol changes, BPO forks enable rapid, low-overhead scaling of blob capacity in response to **real-world demand and network conditions**.
+This EIP introduces **Blob Parameter Only (BPO) Hardforks**, a lightweight mechanism for incrementally scaling Ethereum’s blob capacity through targeted hard forks that modify only blob-related parameters: `blob target`, `blob limit`, and `baseFeeUpdateFraction`. It also introduces an additional optional limit `blob limit per tranaction`. Unlike traditional hard forks, which require extensive coordination and introduce broader protocol changes, BPO forks enable rapid, low-overhead scaling of blob capacity in response to **real-world demand and network conditions**.
```
Marchhill:
```suggestion
This EIP introduces **Blob Parameter Only (BPO) Hardforks**, a lightweight mechanism for incrementally scaling Ethereum’s blob capacity through targeted hard forks that modify only blob-related parameters: `blob_target`, `blob_limit`, and `baseFeeUpdateFraction`. It also introduces an additional optional limit `blob_limit_per tx`. Unlike traditional hard forks, which require extensive coordination and introduce broader protocol changes, BPO forks enable rapid, low-overhead scaling of blob capacity in response to **real-world demand and network conditions**.
```
Would it be possible to change to `base_fee_update_fraction`, or change the other variables to camelCase for consistency?

EIPS/eip-7892.md
```diff
@@ -117,6 +121,7 @@ The parameters and schedules above are purely illustrative. Actual values and sc
 - Execution and consensus clients **MUST** share consistent BPO fork schedules.
 - The slot number in the EL's `blobSchedule` **MUST** align with the start of the epoch specified in the consensus layer configuration.
 - The `max` field in the EL's `blobSchedule` **MUST** equal the `MAX_BLOBS_PER_BLOCK` value in the consensus layer configuration.
+- The `maxBlobsPerTx` field is optional and defaults to value of `max` when omitted. It's not significant for the consensus layer.
```
Marchhill:
```suggestion
- The `maxBlobsPerTx` field is optional and defaults to value of `max` when omitted. It is not used by the consensus layer.
```
Should it be `max` or `MAX_BLOBS_PER_BLOCK`?

EIPS/eip-7892.md
```diff
@@ -182,6 +187,10 @@ No changes are needed in this interaction, but it is noted that the response pay
 
 No changes are required to topic structure or configuration. However, all topics will automatically rotate at a BPO fork due to changes in their `ForkDigestValue` component.
 
+#### Tx pool considerations
```
Marchhill:
```suggestion
#### Transaction Pool considerations
```

EIPS/eip-7892.md
```diff
@@ -182,6 +187,10 @@ No changes are needed in this interaction, but it is noted that the response pay
...
+
+A bpo fork may allow less blobs per block or per transaction than previous one, which means that at some point transactions with high blob count located in the pool may become unusable. Such trasnactions **MUST** be stopped from propagation and should be considred for eviction from tx pool.
```
Marchhill:
```suggestion
A BPO fork may decrease the number of blobs allowed per block or transaction. This means that transactions in the pool with a high blob count may become ineligible for inclusion. Such transactions **MUST NOT** be propagated and should be considered for eviction from the transaction pool.
```

EIPS/eip-7892.md
```diff
@@ -74,16 +75,19 @@ To facilitate these changes on the execution layer, each fork in the `blobSchedu
   "osaka": {
     "target": 9,
     "max": 12,
+    "maxBlobsPerTx": 9,
```
Marchhill: Inconsistent with name `blob_limit_per_tx `?

barnabasbusa:
```suggestion
    "target": 6,
     "max": 9,
    "maxBlobsPerTx": 6,
```

barnabasbusa:
> Do we also need to invalidate block if tx is higher?

We probably should tbh.  And it probably should be part of the spec too.

PR #9632, opened 2025-04-12
Update EIP-7702: Add checks for signatures in authorization lists

nethoxa:
This PR adds some checks that are being done by all EL clients to the signature values of authorization lists, which are not stated in the EIP. As such, developers following the spec may have issues, as they would be working with signatures that are invalid at the EL layer. As an example, see [Geth](https://github.com/ethereum/go-ethereum/blob/80753ba14787ba82954106d7b9025e3a623b96b0/crypto/crypto.go#L277) (this is called when checking the auth signatures [here](https://github.com/ethereum/go-ethereum/blob/80753ba14787ba82954106d7b9025e3a623b96b0/core/types/tx_setcode.go#L119))

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -100,6 +100,9 @@ following:
 3. `authority = ecrecover(msg, y_parity, r, s)`
     * `msg = keccak(MAGIC || rlp([chain_id, address, nonce]))`
     * `s` value must be less than or equal to `secp256k1n/2`, as specified in [EIP-2](./eip-2.md).
+    * `r` value must be less than `secp256k1n`.
+    * `s` and `r` values must be non-zero.
+    * `y_parity` value must be either `0` or `1`.
 4. Add `authority` to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
 5. Verify the code of `authority` is either empty or already delegated.
 6. Verify the nonce of `authority` is equal to `nonce`. In case `authority` does


EIPS/eip-7702.md
```diff
@@ -100,6 +100,9 @@ following:
 3. `authority = ecrecover(msg, y_parity, r, s)`
     * `msg = keccak(MAGIC || rlp([chain_id, address, nonce]))`
     * `s` value must be less than or equal to `secp256k1n/2`, as specified in [EIP-2](./eip-2.md).
+    * `r` value must be less than `secp256k1n`.
+    * `s` and `r` values must be non-zero.
+    * `y_parity` value must be either `0` or `1`.
```
jochem-brouwer, line 105: I think these checks are implicit in the `ecrecover` method but I think it is great to add this for clarification

nethoxa: Yeah, they are implicit, but we have found bugs in some EL clients like Besu that accepted values diff from the expected ones, leading to a chain split. Because of that, I think being more verbose here could be helpful

jochem-brouwer: Could you point me to the PR which fixes this? This sounds like there is likely something going wrong with the wrapper to the underlying `ecrecover` method. One would expect if we would input for instance `r > secp256k1n` that ecrecover throws (?). This should thus "skip" this authority item. But by this comment I understand that somehow `ecrecover` returned an address? (Because this is the situation I would imagine to lead to a chain split, for instance `ecrecover` wrongly returning the 0 address and then Besu setting a delegate there?)

jochem-brouwer (approved):
Checked against linked Geth ref (from OP): https://github.com/ethereum/go-ethereum/blob/80753ba14787ba82954106d7b9025e3a623b96b0/crypto/crypto.go#L277
LGTM

EIPS/eip-7702.md
nethoxa: This was found in the contest, check https://github.com/hyperledger/besu/pull/8408/files#diff-3fda6d9a89c02cb420202501530dbd9ee5a92427f533ed4ec6baeedca0dde9c0. Even if they return an address, which is correct per the default implementation of ECDSA, they constraint it according to the EIPs, so even if an address was returned, it will not return it if the signature has invalid args. No issues at all.

nethoxa:
Just a bump on this as seems to have been a bit forgotten :)

PR #9633, opened 2025-04-12
Add EIP: Algorithmic Transaction Wrapper

SirSpudlington:
This EIP introduces a new wrapper type transaction to allow for alternative signature algorithms to sign off on an ethereum transactions.

### DIFF
--- a/EIPS/draft-eip-multi-algorithm-pki.md
+++ b/EIPS/draft-eip-multi-algorithm-pki.md
@@ -163,7 +163,7 @@ def calculate_penalty(signing_data: bytes, algorithm: int) -> int:
   return total_gas_pentalty
 ```
 
-The penalty MUST be added onto the `21000` base gas of each transaction BEFORE the transaction is processed. If the wrapped tx's `gas_limit` is less than to `21000 + calculate_penalty(signing_data, algorithm)` than the transaction MUST be considered invalid and MUST NOT be included within blocks.
+The penalty MUST be added onto the `21000` base gas of each transaction BEFORE the transaction is processed. If the wrapped tx's `gas_limit` is less than to `21000 + calculate_penalty(signing_data, algorithm)` than the transaction MUST be considered invalid and MUST NOT be included within blocks. This transaction also MUST inherit the intrinsics of the wrapped tx's fee structure (e.g. a wrapped EIP-1559 tx would behave as a EIP-1559 tx).
 
 ## Rationale
 
--- a/EIPS/draft-eip-multi-algorithm-pki.md
+++ b/EIPS/draft-eip-multi-algorithm-pki.md
@@ -1,0 +1,169 @@
+---
+title: Transactions backed with non-secp256k1 PKI algorithms
+description: This EIP adds a standardized way to support additional algorithms to back security of accounts and transactions.
+author: James Kempton (@SirSpudlington)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: Core
+created: 2025-04-12
+requires: 2718, 2930, 1559, 4844
+---
+
+## Abstract
+
+This EIP introduces a new wrapper type transaction to allow for an alternative signature algorithm to sign off on an ethereum transaction.
+
+## Motivation
+
+As quantum computers are getting more advanced and several new post-quantum (PQ) algorithms have been designed. These algorithms all contain drawbacks such as large key sizes (>1KiB), large signature sizes or long verification times. These issues make them more expensive to compute and store than the current secp256k1 curve in use (as of 2025-04-12).
+
+This EIP provides a future-proofed solution to the several algorithms by adding a standardized way to represent alternative algorithms within a transaction.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Parameters
+
+| Constant | Value |
+| - | - |
+| `ALG_TX_TYPE` | `Bytes1(0x04)` |
+| `GAS_PER_ADDITIONAL_VERIFICATION_BYTE`| `16` |
+
+### Algorithmic Transaction
+
+This EIP introduces a [eip-2718](./eip-2718.md) transaction with a `TransactionType` of `ALG_TX_TYPE` and a `TransactionPayload` of the RLP serialization of the `AlgTransactionPayloadBody` defined below:
+
+`[alg_type, signature_info, parent]`
+
+The field `alg_type` is a unsigned 8-bit integer (uint8) that represents the algorithm used to sign the transaction in the `parent` field. This EIP defines only a single value that could be in this value which is `0` for the default secp256k1 curve.
+
+The `signature_info` info field contains information required to verify the signature of the transaction in the `parent` field. This is a byte-array of arbitrary length, which would be passed to the verifier.
+
+The `parent` field contains another [eip-2718](./eip-2718.md) Typed Transaction Envelope, which MUST support all existing `TransactionType`s e.g. `0x1` Optional access list transactions, `0x2` [eip-1559](./eip-1559.md) transactions and `0x3` Shard blob transactions. In addition, all legacy transactions with a `TransactionType` of `> 0x7f`, the only exception to this rule is the `Algorithmic Transaction` itself, which MUST NOT be double-wrapped. These transactions all contain `y_parity`, `r`, `s` values, which MUST be set to `Bytes0()` if wrapped in a `AlgTransactionPayloadBody`, all other values MUST be identical, along with any signing functions proposed by the transactions.
+
+The Algorithmic Transaction MUST NOT generate a transaction reciept with a `TransactionType` of `ALG_TX_TYPE`, it MUST generate a transaction receipt of the transaction it is wrapping (the tx in the `parent` field). Implementations MUST 
+
+### Algorithm specification
+
+Further algorithms MUST be specified via an additional EIP.
+
+Each type of algorithm MUST specify the `ALG_TYPE`, `MAX_SIZE` and `GAS_PENALTY` fields, the `GAS_PENALTY` field MUST only account for verification costs, not storage nor signing.
+
+New algorithms MUST also specify how to recover & verify a valid address (`bytes20`) from the `signature_info` field inside the transaction, the verification function MUST follow the following function signature:
+
+`def verify(signature_info: bytes, parent_hash: bytes32) -> boolean, bytes20`
+
+The verify function MUST return `(false, 0x0)` if there was an error recovering a valid address from the signature, otherwise the function MUST return `true` and the address of the signer.
+
+### The default secp256k1 `0x0` algorithmic type
+
+| Constant | Value |
+| - | - |
+| `ALG_TYPE` | `Bytes1(0x0)` |
+| `GAS_PENALTY`| `0` |
+| `MAX_SIZE` | `65` |
+
+```python
+def verify(signature_info: bytes, parent_hash: bytes32) -> boolean, bytes20:
+  assert(len(signature_info) == 65)
+  v, r, s = int.from_bytes(signature_info[0:32], 16), int.from_bytes(signature_info[32:64], 16), signature_info[64]
+
+  # This assumes `ecrecover` is identical to the `ecrecover` function in solidity.
+  signer = ecrecover(parent_hash, v, r, s)
+
+  return (signer != 0x0, signer[:20])
+```
+
+### Verification
+
+Implementations MUST NOT gossip transactions that have a `len(tx.signature_info)` of more than `alg.MAX_SIZE`.
+
+The validity/processing of the transaction is completed by the function defined below:
+```python
+
+def process_transaction(tx: Transaction, from_address: bytes20 = None, start_gas = 21000):
+  match tx:
+    # Verification for other transactions, if `from_address != None`
+    # then the verifier MUST NOT attempt to validate the `y_parity`, `r`, `s`
+    # parameters. Additionally, the verification function MUST start the initial gas value
+    # at `start_gas`.
+    ...
+
+    ALG_TX:
+      assert(from_address == None)          # Ensure no double-wrapping
+      assert(Algorithms[alg_type] != None)  # `Algorithms` is a dictionary containing every defined algorithm
+
+      alg = Algorithms[alg_type]
+
+      assert(len(tx.signature_info) <= alg.MAX_SIZE)
+
+      valid, from_address = alg.verify(tx.signature_info, calculate_signing_hash(wrapped)) # calculate_signing_hash is defined within the wrapped transaction's EIP.
+      assert(valid)
+
+      process_transaction(tx.parent, from_address, 21000 + calculate_penalty(tx.signature_info, alg))
+
+```
+
+### Gas calculation
+
+All transactions that use more resources than the secp256k1 curve suffer an additional penalty. This penalty MUST be calculated as follows:
+```python
+def calculate_penalty(signing_data: bytes, algorithm: int) -> int:
+  gas_penalty_base = max(len(signing_data) - 65, 0) * GAS_PER_ADDITIONAL_VERIFICATION_BYTE
+  total_gas_pentalty = gas_penalty_base + ALGORITHMS[algorithm].GAS_PENALTY
+  return total_gas_pentalty
+```
+
+The penalty MUST be added onto the `21000` base gas of each transaction BEFORE the transaction is processed.
+
+## Rationale
+
+### Setting `y_parity`, `r`, `s` values to zero rather than removing them
+
+Keeping the `y_parity`, `r`, `s` values inside the transactions keeps the previous parsing, verification and processing logic the same and allows for minimal modification to the other specifications while ensuring future proof operation.
+
+### Opaque `signature_info` type
+
+As each algorithm has unique properties, i.e. signature recovery and key sizes. A bytearray of indeterminate size would be able to hold every permutation of every possible key and signature, this does lead to a DOS vector which the [Gas penalties](#gas-penalties) section solves.
+
+### Gas penalties
+
+Since having multiple different algorithms results in multiple different signature sizes, and verification costs. Hence, every signature algorithm that is more expensive than the default curve secp256k1, incurs an additional gas penalty, this is to discourage the use of overly expensive algorithms.
+
+The `GAS_PER_ADDITIONAL_VERIFICATION_BYTE` value being `16` was taken from the calldata cost of a transaction, as it is a similar datatype.
+
+### Redundant `0` type
+
+The default type is not designed for general use, but is mostly for prototyping and testing with client teams, as the resultant reciept, logs and state change would be the same as a non-wrapped transaction. It also allows implementations to always wrap transactions without removing secp256k1 curve.
+
+### Not specifing account key-sharing / migration
+
+Allowing a single account to share multiple keys creates a security risk as it depletes the security to the weakest algorithm. This is also out of scope for this EIP and could be implemented via a future EIP.
+
+### Keeping a similar address rather than introducing a new address format
+
+While adding a new address format for every new algorithm would ensure that collisions never happen and that security is not bound by the lowest common denominator in algorithmic security. However, the amount of changes that would have to be made and backwards compatibility issues would be to vast to warrent this.
+
+## Backwards Compatibility
+
+Non-EIP-<TODO-PUT-EIP-NUMBER-HERE> transactions will still be included within blocks and will be treated as the default secp256k1 curve. Therefore there would be no backwards compatibility issues will processing other transactions. However, as a new [eip-2718](./eip-2718.md) transaction has been added non-upgraded clients would not be able to process these transactions, and blocks that include these.
+
+## Test Cases
+
+These test cases do not involve processing other types of transactions. Only the wrapping, unwrapping and verification of transactions.
+
+TODO, must be done before EIP enter review stage.
+
+## Security Considerations
+
+Allowing more ways to potentially create transactions for a single account may decrease overall security for that specific account, however this is partially mitigated by the increase in processing power required to trial all algorithms. Even still, adding additional algorithms may need further discussing to ensure that the security of the network is not compromised.
+
+Having `signature_info` be of no concrete type is that there is a chance that an algorithms logic is specified or implemented incorrectly, which could lead to in the best case, invalid blocks, or at worst, the ability for anyone to sign a fraudulent transaction for any account. This security consideration is delegated to the algorithms specification, therefore care must be taken when writing these algorithm specifications.
+
+Further security considerations need discussion.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE.md).
--- a/EIPS/draft-eip-multi-algorithm-pki.md
+++ b/EIPS/draft-eip-multi-algorithm-pki.md
@@ -12,11 +12,11 @@ requires: 2718, 2930, 1559, 4844
 
 ## Abstract
 
-This EIP introduces a new wrapper type transaction to allow for alternative signature algorithms to sign off on an ethereum transactions.
+This EIP introduces a new wrapper type transaction to allow for alternative signature algorithms to sign off on ethereum transactions.
 
 ## Motivation
 
-As quantum computers are getting more advanced and several new post-quantum (PQ) algorithms have been designed. These algorithms all contain drawbacks such as large key sizes (>1KiB), large signature sizes or long verification times. These issues make them more expensive to compute and store than the current secp256k1 curve in use (as of 2025-04-12).
+As quantum computers are getting more advanced, several new post-quantum (PQ) algorithms have been designed. These algorithms all contain drawbacks such as large key sizes (>1KiB), large signature sizes or long verification times. These issues make them more expensive to compute and store than the current secp256k1 curve in use (as of 2025-04-12).
 
 This EIP provides a future-proofed solution to the several algorithms by adding a standardized way to represent alternative algorithms within a transaction.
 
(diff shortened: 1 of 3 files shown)


EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Parameters
+
+| Constant | Value |
+| - | - |
+| `ALG_TX_TYPE` | `Bytes1(0x04)` |
```
jochem-brouwer, line 31: This is the same as an EIP 7702 transaction? I suggest using `0x07`: `0x05` is used as magic in authority lists (by EIP 7702) and `0x06` is currently reserved by EOFs TXCREATE tx

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,169 @@
...
+
+`[alg_type, signature_info, parent]`
+
+The field `alg_type` is a unsigned 8-bit integer (uint8) that represents the algorithm used to sign the transaction in the `parent` field. This EIP defines only a single value that could be in this value which is `0` for the default secp256k1 curve.
+
+The `signature_info` info field contains information required to verify the signature of the transaction in the `parent` field. This is a byte-array of arbitrary length, which would be passed to the verifier.
+
+The `parent` field contains another [eip-2718](./eip-2718.md) Typed Transaction Envelope, which MUST support all existing `TransactionType`s e.g. `0x1` Optional access list transactions, `0x2` [eip-1559](./eip-1559.md) transactions and `0x3` Shard blob transactions. In addition, all legacy transactions with a `TransactionType` of `> 0x7f`, the only exception to this rule is the `Algorithmic Transaction` itself, which MUST NOT be double-wrapped. These transactions all contain `y_parity`, `r`, `s` values, which MUST be set to `Bytes0()` if wrapped in a `AlgTransactionPayloadBody`, all other values MUST be identical, along with any signing functions proposed by the transactions.
```
jochem-brouwer, line 44: Ok, interesting, so thinking more about the design of this EIP, it essentially "takes out" the signature elements of the original txs and replaces it by this "choice" thing (the `alg_type`).

However, this also means that every original tx can now be sent over the network as the original tx, or as the Algorithmic Transaction with `alg_type` 0. I don't think we want two different representations of essentially the same object.

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,169 @@
...
+
+```python
+def verify(signature_info: bytes, parent_hash: bytes32) -> boolean, bytes20:
+  assert(len(signature_info) == 65)
+  v, r, s = int.from_bytes(signature_info[0:32], 16), int.from_bytes(signature_info[32:64], 16), signature_info[64]
+
+  # This assumes `ecrecover` is identical to the `ecrecover` function in solidity.
+  signer = ecrecover(parent_hash, v, r, s)
+
+  return (signer != 0x0, signer[:20])
```
jochem-brouwer, line 76: I think the address is the hash of the pubkey (which is returned by `ecrecover`) and then taking either the first or last 20 bytes (not sure which at the moment but should be easy to find out)

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+### Verification
+
+Implementations MUST NOT gossip transactions that have a `len(tx.signature_info)` of more than `alg.MAX_SIZE`.
```
jochem-brouwer, line 81: So these transactions are invalid? (The same as transactions where, for instance, the `to` field is not 0 or 20 bytes, but something else like 10, these also get rejected at the network level as being "invalid txs")

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,169 @@
...
+      valid, from_address = alg.verify(tx.signature_info, calculate_signing_hash(wrapped)) # calculate_signing_hash is defined within the wrapped transaction's EIP.
+      assert(valid)
+
+      process_transaction(tx.parent, from_address, 21000 + calculate_penalty(tx.signature_info, alg))
+
+```
+
+### Gas calculation
+
+All transactions that use more resources than the secp256k1 curve suffer an additional penalty. This penalty MUST be calculated as follows:
```
jochem-brouwer, line 111: I would not specify this struct rule in this EIP, but rather make this "gas penalty" reflect the costs of the recovery algorithm. For instance if something is 10x more CPU intensive than `secp256k1` then that should definitely be included in the gas costs.

jochem-brouwer: Oh, nvm, that's already in there :smile:

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Gas penalties
+
+Since having multiple different algorithms results in multiple different signature sizes, and verification costs. Hence, every signature algorithm that is more expensive than the default curve secp256k1, incurs an additional gas penalty, this is to discourage the use of overly expensive algorithms.
+
+The `GAS_PER_ADDITIONAL_VERIFICATION_BYTE` value being `16` was taken from the calldata cost of a transaction, as it is a similar datatype.
+
+### Redundant `0` type
+
+The default type is not designed for general use, but is mostly for prototyping and testing with client teams, as the resultant reciept, logs and state change would be the same as a non-wrapped transaction. It also allows implementations to always wrap transactions without removing secp256k1 curve.
```
jochem-brouwer, line 139: I would convert this EIP to be a general template for introducing different signature algorithms on currently supported txs and I would not include this "default" behavior where the 0 algorithm is the "default" secp256k1. This would add it to consensus and this is unnecessary. 

For testing, either a test-EIP or a supplemental file here could be written explaining the test cases for this tx, for instance wrapping secp256k1 in the Algorithmic Tx.

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,169 @@
...
+
+While adding a new address format for every new algorithm would ensure that collisions never happen and that security is not bound by the lowest common denominator in algorithmic security. However, the amount of changes that would have to be made and backwards compatibility issues would be to vast to warrent this.
+
+## Backwards Compatibility
+
+Non-EIP-<TODO-PUT-EIP-NUMBER-HERE> transactions will still be included within blocks and will be treated as the default secp256k1 curve. Therefore there would be no backwards compatibility issues will processing other transactions. However, as a new [eip-2718](./eip-2718.md) transaction has been added non-upgraded clients would not be able to process these transactions, and blocks that include these.
+
+## Test Cases
+
+These test cases do not involve processing other types of transactions. Only the wrapping, unwrapping and verification of transactions.
```
jochem-brouwer, line 155: Right, but in the testing of these transactions this means in practice we just doubled the entire test size. Because for each "old" tx we now also need to test if the algorithmic variant works exactly the same (if thus the `arg_type` of 0 is included)

jochem-brouwer (commented):
I like the idea to already think about how we can plugin new signature schemes but I am not sure if this is the "best" way to do it. I think a more cleaner approach would be to already have this feature in the transactions themselves (so the signing is a "module" and can either take the form of secp256k1 or something new). But, this is obviously not possible so we need another mechanism to swap it out for something else and this is already a good start :smile: :+1:

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: I have reworded this so it is less ambigious, as it attempts to ensure that the exact opposite happens, so no test cases are needed for every other transaction test case.

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: They are only invalid if `len(tx.signature_info) > `alg.MAX_SIZE`, as some algorithms may have variable signatures or variable additional data required for verification.

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: Just looked it up, that is correct and the `[:20]` would be redundant.

SirSpudlington:
> I think a more cleaner approach would be to already have this feature in the transactions themselves (so the signing is a "module" and can either take the form of secp256k1 or something new)

I completely agree, it would be nicer to have signing modules embedded in transactions, but as you said, it would break backwards compatibility in almost every way.

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: Solved with commit `d6e2650f494b476f9eba306d6bf1892e37f1dfca`

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: Solved with commit `d6e2650f494b476f9eba306d6bf1892e37f1dfca`

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,216 @@
...
+
+`[alg_type, signature_info, parent]`
+
+The field `alg_type` is a unsigned 8-bit integer (uint8) that represents the algorithm used to sign the transaction in the `parent` field. This EIP does not define algorithms for use with this transaction type but a structure so they remain interoperable.
+
+The `signature_info` info field contains information required to verify the signature of the transaction in the `parent` field. This is a byte-array of arbitrary length, which would be passed to the verification function.
+
+The `parent` field contains another [eip-2718](./eip-2718.md) Typed Transaction Envelope, which MUST support all existing `TransactionType`s. In addition, all legacy transactions with a `TransactionType` of `> 0x7f` MUST also be supported, the only exception to this rule is the `Algorithmic Transaction` itself, which MUST NOT be placed within itself. These transactions all contain `y_parity`, `r`, `s` values, which MUST be set to `Bytes0()` if wrapped in a `AlgTransactionPayloadBody`, all other values MUST be unchanged from their original values.
```
bomanaps, line 44: In your specification it states all legacy transactions with `TransactionType > 0x7f` must be supported, but doesn't clarify how this interacts with future transaction types?

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,216 @@
...
+```python
+def verify(signature_info: bytes, parent_hash: bytes32) -> boolean, bytes20:
+  assert(len(signature_info) == 65)
+  v, r, s = int.from_bytes(signature_info[0:32], 16), int.from_bytes(signature_info[32:64], 16), signature_info[64]
+
+  # This assumes `ecrecover` is identical to the `ecrecover` function in solidity.
+  signer = ecrecover(parent_hash, v, r, s)
+
+  return (signer != 0x0, signer)
+```
```
bomanaps, line 108: This  unpacks signature components incorrectly (the byte ranges don't align with typical r, s, v structure), the verify function extracts v,r,s but with incorrect byte ordering and lengths.
This is what I mean

The byte ranges and ordering for v, r, s are incorrect:

Standard Ethereum signature format has r (32 bytes), s (32 bytes), and v (1 byte)
But your code assigns:

v = first 32 bytes (should be r)
r = next 32 bytes (should be s)
s = final byte (should be v)




Then we have the `int.from_bytes1` function usage which is also wrong:

You're useing base 16 as the second parameter `(int.from_bytes(signature_info[0:32], 16))`
In Python, the second parameter should be byte order ('big' or 'little'), not the base

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,216 @@
...
+      assert(Algorithms[alg_type] != None)  # `Algorithms` is a dictionary containing every defined algorithm
+
+      alg = Algorithms[alg_type]
+
+      assert(len(tx.signature_info) <= alg.MAX_SIZE)
+
+      valid, from_address = alg.verify(tx.signature_info, calculate_signing_hash(wrapped)) # calculate_signing_hash is defined within the wrapped transaction's EIP.
+      assert(valid)
+
+      process_transaction(tx.parent, from_address, 21000 + calculate_penalty(tx.signature_info, alg))
```
bomanaps, line 150: What happens if  `from_address` returned by the verification function doesn't match an expected address for the transaction?

bomanaps (commented):
Lastly penalty calculation adds gas before transaction processing, but how do you clarify how this interacts with user-specified gas limits and transaction fee structures?

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: This would never occur, If a different address is returned then it was signed by a different public key.

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: I removed & restructured these as `int.from_bytes()` was not needed as solidity takes bytes32. That was more of an amalgamation of rust's `u32::from_le_bytes()` and python's `int.from_bytes`

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: I have added the additional specification regarding future transaction types.

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,218 @@
+---
+title: Support for transactions signed with non-secp256k1 algorithms
+description: This EIP adds a standardized way to support additional algorithms to signed transactions
```
SamWilsn, line 3:
```suggestion
description: Introduces a new transaction type allowing alternative transaction signature algorithms
```

This still isn't perfect, but it's going in the right direction. Feel free to play with this and not take this suggestion literally. The main concern is that we want the description and title to be as information-dense as possible.

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,218 @@
...
+status: Draft
+type: Standards Track
+category: Core
+created: 2025-04-12
+requires: 2718, 2930, 1559, 4844
+---
+
+## Abstract
+
+This EIP introduces a new wrapper type transaction to allow for alternative signature algorithms to sign off on ethereum transactions.
```
SamWilsn, line 15: Your abstract certainly satisfies the brief requirement, but I feel it is lacking some technical meat. The abstract should give the reader a high-level (but technical) overview of _how_ the proposal accomplishes its goals and not just state what those goals are. Here you could maybe mention a bit about what "wrapper" means, and how it relates to other transaction types/encodings.

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,218 @@
...
+| `ALG_TX_TYPE` | `Bytes1(0x07)` |
+| `GAS_PER_ADDITIONAL_VERIFICATION_BYTE`| `16` |
+
+### Algorithmic Transaction
+
+This EIP introduces a new [eip-2718](./eip-2718.md) transaction with a `TransactionType` of `ALG_TX_TYPE` and a `TransactionPayload` of the RLP serialization of the `AlgTransactionPayloadBody` defined below:
+
+`[alg_type, signature_info, parent]`
+
+The field `alg_type` is a unsigned 8-bit integer (uint8) that represents the algorithm used to sign the transaction in the `parent` field. This EIP does not define algorithms for use with this transaction type but a structure so they remain interoperable.
```
SamWilsn, line 40: > but a structure so they remain interoperable

This bit is rationale. Having a bit of rationale embedded in the specification section is fine, but I wanted to point it out in case you'd like to move it to that section.

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,218 @@
...
+
+The `signature_info` info field contains information required to verify the signature of the transaction in the `parent` field. This is a byte-array of arbitrary length, which would be passed to the verification function.
+
+The `parent` field contains another [eip-2718](./eip-2718.md) Typed Transaction Envelope, while MUST be able to contain every possible `TransactionType`s, including legacy transactions with a `TransactionType` of `> 0x7f`, the only exception to this rule is the `Algorithmic Transaction` itself, which MUST NOT be placed within itself. These transactions all contain `y_parity`, `r`, `s` values, which MUST be set to `Bytes0()` if wrapped in a `AlgTransactionPayloadBody`, all other values MUST be unchanged from their original values.
```
SamWilsn, line 44:
```suggestion
The `parent` field contains another [eip-2718](./eip-2718.md) Typed Transaction Envelope, which MUST be able to contain every possible `TransactionType`, including legacy transactions with a `TransactionType` of `> 0x7f`, but the only exception to this rule is the `Algorithmic Transaction` itself, which MUST NOT be placed within itself. These transactions all contain `y_parity`, `r`, `s` values, which MUST be set to `Bytes0()` if wrapped in a `AlgTransactionPayloadBody`, all other values MUST be unchanged from their original values.
```

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,218 @@
...
+
+The `GAS_PENALTY` field MUST only account for verification costs, not storage nor signing.
+
+New algorithms MUST also specify how to recover & verify a valid address (`bytes20`) from the `signature_info` field inside the transaction, the verification function MUST follow the following function signature:
+
+`def verify(signature_info: bytes, parent_hash: bytes32) -> boolean, bytes20`
+
+The verify function MUST return `(false, 0x0)` if there was an error recovering a valid address from the signature, otherwise the function MUST return `true` and the address of the signer.
+
+### Example EIP for adding the secp256k1 curve
```
SamWilsn, line 69: I'd move this template into your `assets` directory and link to it. It gets a little confusing to have an EIP embedded inside an EIP.

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,218 @@
...
+
+~~~md
```
SamWilsn, line 71: That's interesting. I've always used

`````markdown
````markdown
content
````
`````

File: EIPS/draft-eip-multi-algorithm-pki.md [hunk 17]
```diff
@@ -0,0 +1,218 @@
...
+
+~~~
+
+### Verification
+
+Implementations MUST consider transactions invalid where `len(tx.signature_info) > alg.MAX_SIZE`.
+
+The validity/processing of the transaction is completed by the function defined below:
+```python
+
```
SamWilsn, line 132:
````suggestion

```python
````

File: EIPS/draft-eip-multi-algorithm-pki.md [hunk 18]
```diff
@@ -0,0 +1,218 @@
...
+Non-EIP-<TODO-PUT-EIP-NUMBER-HERE> transactions will still be included within blocks and will be treated as the default secp256k1 curve. Therefore there would be no backwards compatibility issues will processing other transactions. However, as a new [eip-2718](./eip-2718.md) transaction has been added non-upgraded clients would not be able to process these transactions nor blocks that include these transactions.
+
+## Test Cases
+
+These test cases do not involve processing other types of transactions. Only the wrapping, unwrapping and verification of these transactions without interfacing with the `parent` tx held inside the main tx.
+
+All the following test cases use the parameters from the [example eip](#example-eip-for-adding-the-secp256k1-curve) listed above.
+
+TODO, must be done before EIP enter review stage.
```
SamWilsn, line 206:
```suggestion
<!-- TODO, must be done before EIP enter review stage. -->
```

EIPS/draft-eip-multi-algorithm-pki.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Security Considerations
+
+Allowing more ways to potentially create transactions for a single account may decrease overall security for that specific account, however this is partially mitigated by the increase in processing power required to trial all algorithms. Even still, adding additional algorithms may need further discussing to ensure that the security of the network would not be compromised.
+
+Having `signature_info` be of no concrete type creates a chance that an algorithms logic could be specified or implemented incorrectly, which could lead to, in the best case, invalid blocks, or at worst, the ability for anyone to sign a fraudulent transaction for any account. This security consideration is delegated to the algorithms specification, therefore care must be taken when writing these algorithm specifications to avoid critical security flaws.
+
+Further security considerations need discussion.
```
SamWilsn, line 214:
```suggestion
<!-- TODO: Further security considerations need discussion. -->
```

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: I could not figure out how to originally do this, the first thing I found when googling was `~~~`

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: I padded the abstract out a bit more

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: I removed it, it was not that important.

EIPS/draft-eip-multi-algorithm-pki.md
SirSpudlington: Moved

SirSpudlington:
@SamWilsn The HTMLProofer is returning an error with `This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15`

EIPS/eip-7932.md
```diff
@@ -0,0 +1,260 @@
+---
+eip: 7932
+title: Algorithmic Transactions
```
g11tech, line 3:
```suggestion
title: Algorithmic Transaction Wrapper
```

assets/eip-7932/template-eip.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Example EIP to add secp256k1 curve as an algorithmic type
+description: Example EIP to add secp256k1 curve as an algorithmic type
```
g11tech, line 3: no need to add a template eip, if you have a signature scheme to propose, ass another EIP

EIPS/eip-7932.md
```diff
@@ -0,0 +1,260 @@
...
+
+The `signature_info` field contains information required to verify the signature of the transaction in the `parent` field. This is a byte-array of arbitrary length, which would be passed to the verification function.
+
+The `parent` field contains another serialized [EIP-2718](./eip-2718.md) Typed Transaction Envelope (i.e. the `parent` field type is `bytes`), which MUST be able to contain every possible `TransactionType`, including legacy transactions with a `TransactionType` of `> 0x7f`, but the only exception to this rule is the `Algorithmic Transaction` itself, which MUST NOT be placed within itself.
+
+These `parent` transactions all contain `y_parity`, `r`, `s` values, which MUST be set to `Bytes0()` if wrapped in a `AlgTransactionPayloadBody`, there are two exceptions to this rule however:
+
+- If the `alg_type` is the NULL algorithm, the signature field MUST be left unchanged.
```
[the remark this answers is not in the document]
SirSpudlington: I added a link to where the NULL algorithm is specified, it is to fix edge cases when a 7702 transaction is signed by secp256k1 but the `authorization_list` contains values of non-secp256k1 transactions. I added a link to the section where this is specified.

EIPS/eip-7932.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Algorithmic Transaction
+
+This EIP introduces a new [EIP-2718](./eip-2718.md) transaction with a `TransactionType` of `ALG_TX_TYPE` and a `TransactionPayload` of the RLP serialization of the `AlgTransactionPayloadBody` defined below:
+
+`[alg_type, signature_info, parent, additional_info]`
+
+The field `alg_type` is an unsigned 8-bit integer (uint8) that represents the algorithm used to sign the transaction in the `parent` field. This EIP does not define algorithms for use with this transaction type; however, it does specify a NULL algorithm (0xFF) which MUST trigger implementations to verify transaction fields.
+
+The `signature_info` field contains information required to verify the signature of the transaction in the `parent` field. This is a byte-array of arbitrary length, which would be passed to the verification function.
```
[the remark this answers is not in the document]
SirSpudlington: This is inheriantly not done as the precompile and verification function are two distinct things. The precompile calls the verification function. I added a note to the top of the specification to outline this.

EIPS/eip-7932.md
```diff
@@ -0,0 +1,260 @@
...
+
+- If the `alg_type` is the NULL algorithm, the signature field MUST be left unchanged.
+- If the transaction type is that of a legacy transaction, where `y_parity` MUST be equal to the chain ID value, the transactions signing data MUST also be calculated the same way as [EIP-155](./eip-155.md) specifies
+
+All other transaction values MUST be unchanged from their original values.
+
+The `additional_info` field only needs to be populated if additional protocol level signatures are required such as [EIP-7702](./eip-7702.md)'s `authorization_list`. This field MUST contain the RLP serialization of a list of the signatures `[alg_type, signature_info]` and MUST be repeated for every NON-NULL signature inside the transaction. The order of which MUST be the same as the signatures appear in the `parent` tx. This signature MUST also be checked to ensure that `alg_type` is known and is not the NULL algorithm (0xFF), `len(signature_info) <= alg.MAX_SIZE` and then MUST also be verified using the `verify` function for the specific algorithm.
```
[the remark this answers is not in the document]
SirSpudlington: No, the `additional_info` array is not signed by `singature_info`. This is because this data cannot be modified without invalidating the transaction.

assets/eip-7932/template-eip.md
SirSpudlington: This template EIP is added for initial test cases, and will not be included for production environments.

SamWilsn (approved):

PR #9653, opened 2025-04-16
Update EIP-7702: Update `nonce` validation for consistency with EIP-2681

fselmo:
Validate authority nonce as `< 2**64 - 1` as in EIP-2681. The `nonce` has the added implication that it is incremented at when processing a transaction, as well as an auth now. This prevents the `nonce` from ever having the value of ``2**64`` after processing a successful authority.


fselmo:
This invalidates the tx, rather than just the authorization when processed. Nothing to see here.

PR #9656, opened 2025-04-17
Update EIP-5450: Simplify stack overflow checks

gumb0:
This removes one redundant rule.


pdobacz (approved):
Makes sense, but I'd like to see more people review this.

chfast (approved):

PR #9658, opened 2025-04-17
Add EIP: RLP Execution Block Size Limit

Giulio2002:

--

 - The build passes.


EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,75 @@
+---
+eip: xxxx
```
lightclient, line 2: 7934

benaadams (approved):

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
+---
+eip: 7934
+title: RLP Execution Block Size Limit
+description: Introduce a protocol-level cap on the maximum RLP-encoded block size to 10 MB, including a 512 KB margin for beacon block size.
+author: Giulio Rebuffo (@Giulio2002), Ben Adams (@benaadams), Storm Slivkoff (@sslivkoff)
+discussions-to: https://ethereum-magicians.org/t/eip-xxx-add-bytesize-limit-to-blocks/23589
```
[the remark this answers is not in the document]
Giulio2002: done

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
+---
+eip: 7934
+title: RLP Execution Block Size Limit
+description: Introduce a protocol-level cap on the maximum RLP-encoded block size to 10 MB, including a 512 KB margin for beacon block size.
```
jochem-brouwer, line 4: I think the correct unit to be used in this document is `KiB` and `MiB`. A MegaByte is `1e6` bytes, and a MebiByte is `2^20 = 1,048,576` bytes. See https://simple.wikipedia.org/wiki/Mebibyte (KiB = 1024 bytes, KB = 1000 bytes)

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+Additionally, blocks exceeding 10 MB are not propagated by the consensus layer's (CL) gossip protocol, potentially causing network fragmentation or denial-of-service (DoS) conditions.
+
+By imposing a protocol-level limit on the RLP-encoded block size, Ethereum can ensure enhanced resilience against targeted attacks on block validation times. Adding an additional margin of 512 KB explicitly accommodates beacon block sizes, ensuring compatibility across network components.
+
+## Specification
+
+### Block Size Cap
+
+- Introduce constants:
+  - `MAX_BLOCK_SIZE` set to **10 MB (10,485,760 bytes)**
```
jochem-brouwer, line 33:
```suggestion
  - `MAX_BLOCK_SIZE` set to **10 MiB (10,485,760 bytes)**
```

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+  - `MARGIN` set to **512 KB (524,288 bytes)**
```
jochem-brouwer, line 34:
```suggestion
  - `MARGIN` set to **512 KiB (524,288 bytes)**
```

EIPS/eip-7934.md
jochem-brouwer, line 4:
```suggestion
description: Introduce a protocol-level cap on the maximum RLP-encoded block size to 10 MiB, including a 512 KiB margin for beacon block size.
```

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+- Any RLP-encoded block exceeding `MAX_RLP_BLOCK_SIZE` must be considered invalid.
+
+Thus add the following check to the Ethereum protocol:
+
+```python
+MAX_BLOCK_SIZE = 10_485_760  # 10 MB
+MARGIN = 524_288  # 512 KB
+MAX_RLP_BLOCK_SIZE = MAX_BLOCK_SIZE - MARGIN
+
+# if true, the block is invalid and should be rejected/not get built
```
jochem-brouwer, line 45: I think first the block gets built, and then it gets RLP encoded. So only after building/encoding it is known if the block size is too large (?). Except of course when decoding the RLP to get the contents of the block, then we can reject immediately.

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+created: 2025-04-16
+---
+
+## Abstract
+
+This proposal introduces a protocol-level cap on the maximum RLP-encoded execution block size to 10 megabytes (MB), which includes a margin of 512 KB to account for beacon block sizes.
+
+## Motivation
+
+Currently, Ethereum does not enforce a strict upper limit on the encoded size of blocks. This lack of constraint can result in:
```
jochem-brouwer, line 19: I think we should carefully define what is a "block" here. Because for EL we could pick the block header, the block itself (header + block body (so this includes transactions and withdrawals for instance)). Or would it be some serialization of whatever gets sent over `engine_newPayload` (so some kind of RLP serialization of `ExecutionPayloadV_n`?)

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+  - `MARGIN` set to **512 KB (524,288 bytes)**
+  - `MAX_RLP_BLOCK_SIZE` calculated as `MAX_BLOCK_SIZE - MARGIN`
+- Any RLP-encoded block exceeding `MAX_RLP_BLOCK_SIZE` must be considered invalid.
+
+Thus add the following check to the Ethereum protocol:
+
+```python
+MAX_BLOCK_SIZE = 10_485_760  # 10 MB
+MARGIN = 524_288  # 512 KB
+MAX_RLP_BLOCK_SIZE = MAX_BLOCK_SIZE - MARGIN
```
jochem-brouwer, line 43: Can we edit the variable names?

```suggestion
GOSSIP_UPPER_LIMIT = 10_485_760  # 10 MiB
SAFETY_MARGIN = 524_288  # 512 KiB
MAX_RLP_BLOCK_SIZE = GOSSIP_UPPER_LIMIT - SAFETY_MARGIN
```

I think this is somewhat closer to what the "purpose" is of these variables (but far from perfect, would love to hear some other suggestions what could be even better here)

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+
+# if true, the block is invalid and should be rejected/not get built
+def exceed_max_rlp_block_size(block: Block) -> bool:
+    return len(rlp.encode(block)) > MAX_RLP_BLOCK_SIZE
```
jochem-brouwer, line 47: This method is not referenced anywhere (although it is clear what it means and in what context it should be invoked). We would also need a check when the block is incoming over the network layer: immediately reject it if it is larger than the max size.

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Rationale
+
+### Why 10 MB?
+
+A cap of 10 MB aligns with the gossip protocol constraint in Ethereum's consensus layer (CL). An additional 512 KB margin explicitly accounts for beacon block sizes, ensuring compatibility and consistent block propagation across the network. Blocks significantly larger than 10 MB will not be broadcast by the CL, potentially leading to network fragmentation or denial-of-service scenarios.
+
+## Backwards Compatibility
+
+This change is **not backward-compatible** with any blocks larger than the newly specified size limit. Validators and miners will need to ensure their block construction logic strictly respects this limit.
```
jochem-brouwer, line 68: This should be part of a hardfork, because based on the definition of the `MAX_RLP_BLOCK_SIZE` the validation rules could then reject canonical chains if there exist a block which is larger than that size.

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+
+1. **Block Creation**: Validators must ensure the total RLP-encoded size of any produced block does not exceed `MAX_RLP_BLOCK_SIZE`.
+2. **Block Validation**: Nodes must reject blocks whose RLP-encoded size exceeds `MAX_RLP_BLOCK_SIZE`.
+
+### Protocol Adjustment
+
+- All Ethereum client implementations must integrate this size check as part of block validation and propagation.
+- This limit applies independently of gas-related metrics.
+
+## Rationale
```
jochem-brouwer, line 60: Is it currently possible to create a block which is larger than 10 MiB? Trivially one tx filled with zero calldata? Does this change in Prague (lower/higher?) compared to Cancun? (assuming gas limit of 30M)

jochem-brouwer (commented):
Some comments, main one I think MiB and KiB should be used instead of MB and KB :) I also don't completely understand the 512 KiB margin, why is this added? If the block fits within the network limit it should be fine, right? Why this extra margin?

EIPS/eip-7934.md
benaadams: Hits around 100MGas

EIPS/eip-7934.md
benaadams: The transactions when being build have their rlp data; so isn't too hard to go from that to overall block size (EL block, CL contribution is unknown to the EL, hence margin)

benaadams:
> If the block fits within the network limit it should be fine, right? Why this extra margin?

Is only EL block the EL can reason about not the EL block with CL wrapper

jochem-brouwer:
> Is only EL block the EL can reason about not the EL block with CL wrapper

Ah right, sorry, I'm a "pure EL" dev so I completely forgot that CL adds extra "fields" in order to propagate the block. Another novice question: can we prove that this safety margin is sufficient enough? Or are there cases where we use all the reserved block space for the block, and the added CL fields will now grow the data packet over 10 MiB? Is there a worst case scenario where this is possible? (Is it possible to set a safe margin limit such that we can theoretically never go over the 10 MiB)

benaadams:
> Another novice question: can we prove that this safety margin (512 KiB) is sufficient enough?

It was quoted as worst case of 128KiB in Belatrix https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/p2p-interface.md#gossipsub

>  Considering that BeaconBlock max size is on the order of 128 KiB in the worst case

Don't know what it is now; hence going for 512KiB

Also is an EL devp2p limit of 10MiB (- framing) after which peer will be disconnected if they gossip it https://github.com/ethereum/devp2p/blob/master/caps/eth.md

> The practical limits for the eth protocol are lower, typically 10 MiB. If a received message is larger than the limit, the peer should be disconnected.

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+This proposal introduces a protocol-level cap on the maximum RLP-encoded execution block size to 10 megabytes (MB), which includes a margin of 512 KB to account for beacon block sizes.
+
+## Motivation
+
+Currently, Ethereum does not enforce a strict upper limit on the encoded size of blocks. This lack of constraint can result in:
+
+1. **Network Instability**: Extremely large blocks slow down propagation and increase the risk of temporary forks and reorgs.
+2. **DoS Risks**: Malicious actors could generate exceptionally large blocks to disrupt network performance.
+
+Additionally, blocks exceeding 10 MB are not propagated by the consensus layer's (CL) gossip protocol, potentially causing network fragmentation or denial-of-service (DoS) conditions.
```
benaadams, line 24: Looks like EL devp2p will disconnect peers that gossip messages over 10MiB also

> The practical limits for the eth protocol are lower, typically 10 MiB. If a received message is larger than the limit, the peer should be disconnected.

https://github.com/ethereum/devp2p/blob/master/caps/eth.md#ethereum-wire-protocol-eth

g11tech:
lemme know when good for merging since there are a good amount of open technical comments

EIPS/eip-7934.md
Giulio2002: yes - hence **not backward compatible**

Giulio2002:
> lemme know when good for merging since there are a good amount of open technical comments

I think good for merging

changeschung:
Thank you for searching out
It really matters to me

EIPS/eip-xxxx.md
changeschung: Can you explain?

EIPS/eip-7934.md
changeschung: Okay any one if you sincerely can explain something to me ease

dajuguan:
I have a naive question: with the current gas limit, block sizes typically don't exceed 3 MiB. In what scenario would a block size reach 10 MiB?

benaadams:
> I have a naive question: with the current gas limit, block sizes typically don't exceed 3 MiB. In what scenario would a block size reach 10 MiB?

Around 100MGas block where the block was a single tx that was all zeros

EIPS/eip-7934.md
Giulio2002: I will close this as it is an impl detail and kind of trivial

EIPS/eip-7934.md
Giulio2002: this is just the spec for the check on state transition. how you implement it is your an implementation detail

Giulio2002:
@g11tech ready for merging. addressed everything

EIPS/eip-7934.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Block Size Cap
+
+- Introduce constants:
+  - `MAX_BLOCK_SIZE` set to **10 MiB (10,485,760 bytes)**
+  - `MARGIN` set to **512 KiB (524,288 bytes)**
+  - `MAX_RLP_BLOCK_SIZE` calculated as `MAX_BLOCK_SIZE - MARGIN`
+- Any RLP-encoded block exceeding `MAX_RLP_BLOCK_SIZE` must be considered invalid.
+
+Thus add the following check to the Ethereum protocol:
```
g11tech, line 38: should be good idea to list where all these checks should be enforced unless speced out somewhere else

g11tech (approved):
added a nit but still approving, see if you would want to address it in your followup PRs which shouldn't require editor's approval (if in draft)

noelportillo (commented):
Are they going for the new ones

PR #9669, opened 2025-04-22
Update EIP-7873: Handling of EOF containers in existing TX types.

shemnon:
Revise the spec so that the behavior of Type 0, 1, and 2 transactions
remains unchanged when the input data starts with EF or contains a valid
 EOF container. (Do not parse and execute the EF instruction).

### DIFF
--- a/EIPS/eip-7873.md
+++ b/EIPS/eip-7873.md
@@ -81,7 +81,7 @@ tx.gasUsed = (
 
 Under transaction validation rules `initcodes` are not validated for conforming to the EOF specification. They are only validated when accessed via `TXCREATE`. This avoids potential DoS attacks of the mempool. If during the execution of an `InitcodeTransaction` no `TXCREATE` instruction is called, such transaction is still valid.
 
-Legacy creation transactions (any transactions with empty `to`) that start with the two bytes `EF00` will be invalid. They will not be parsed into EOF containers and they will not be executed as legacy bytecode.
+Other creation transactions that support contract creation (specifically type 0 "Frontier," type 1 "AccessList," type 2 "FeeMarket" transactions with an empty `to` field) will not attempt to parse EOF containers in their `input` field and will execute the code as non-EOF code. This will result in immediately executing the `INVALID`(`0xEF`) instruction (see [EIP-141](./eip-141.md)) and halting.
 
 
 #### RLP and signature
@@ -175,7 +175,7 @@ This change poses no risk to backwards compatibility, as it is introduced at the
 
 The transactions of the new type are invalid until this change activates.
 
-Contract creation options do not change for legacy bytecode.
+Contract creation options do not change for legacy bytecode, including how existing transactions behave when encountering code that may look like an EOF container.
 
 ## Security Considerations
 


EIPS/eip-7873.md
```diff
@@ -175,7 +175,7 @@ This change poses no risk to backwards compatibility, as it is introduced at the
 
 The transactions of the new type are invalid until this change activates.
 
-Contract creation options do not change for legacy bytecode.
+Contract creation options do not change for legacy bytecode, including how existing transactions behave when encountering code that may look like an EOF container.
```
pdobacz, line 178:
```suggestion
Contract creation options do not change for legacy bytecode, including how existing transactions with `to: nil` behave when encountering code in their calldata that may look like an EOF container.
```

Otherwise
> encountering code that may look like an EOF container.

is ambiguous.

pdobacz (commented):
Just a comment on the wording here, leaving the discussion on the issue to the EthMagicians thread for 7873

EIPS/eip-7873.md
```diff
@@ -167,6 +167,18 @@ Relying on the EOF creation transactions as the alternative solution makes it im
...
+
+One impact that would be seen by this would be block builders, as they would need to ensure that the transactions that are invalid are not included in a block. There is precedent for this, as [EIP-7623](./eip-7623.md) established a new floor for the gas limit in transactions, prior to its adoption the limit was lower.  However, block builders have an existing check on gas limits and transactions and that change was adjusting the formula and constants.
+
+We should also consider the treatment [EIP-7702](./eip-7702) delegation designations receive if they show up as code in a contract creation transaciton.  The transaction is valid, is executed as plain bytecode, ad the first operation executed is `0xEF`, which as per [EIP-141](eip-141) is the designated invalid opcode and would cause the whole transction to abort.
```
chfast, line 178: EF is not the designated invalid opcode, FE is.

EIPS/eip-7873.md
```diff
@@ -167,6 +167,18 @@ Relying on the EOF creation transactions as the alternative solution makes it im
...
+
+Letting the inputdata of a type 0, 1, and 2 contract creation transaction execue as plain EVM code irrespective of the magic `0xEF00` byes keeps a consistent behavior in client code and keeps block builders from having to update their logic in older transaction types.
```
chfast, line 180:
```suggestion
Letting the inputdata of a type 0, 1, and 2 contract creation transaction execue as plain EVM code irrespective of the magic `0xEF00` bytes keeps a consistent behavior in client code and keeps block builders from having to update their logic in older transaction types.
```

EIPS/eip-7873.md
```diff
@@ -81,7 +81,7 @@ tx.gasUsed = (
 
 Under transaction validation rules `initcodes` are not validated for conforming to the EOF specification. They are only validated when accessed via `TXCREATE`. This avoids potential DoS attacks of the mempool. If during the execution of an `InitcodeTransaction` no `TXCREATE` instruction is called, such transaction is still valid.
 
-Legacy creation transactions (any transactions with empty `to`) that start with the two bytes `EF00` will be invalid. They will not be parsed into EOF containers and they will not be executed as legacy bytecode.
+Other creation transactions that support contract creation (specifically type 0 "Frontier," type 1 "AccessList," type 2 "FeeMarket" transactions with an empty `to` field) will not attempt to parse EOF containers in their `input` field and will execute the code as non-EOF code. This will result in immediately executing the `INVALID`(`0xEF`) instruction (see [EIP-141](./eip-141.md)) and halting.
```
chfast, line 84:
```suggestion
Other creation transactions that support contract creation (specifically type 0 "Frontier," type 1 "AccessList," type 2 "FeeMarket" transactions with an empty `to` field) will not attempt to parse EOF containers in their `input` field and will execute the code as non-EOF code. This will result in immediately executing the undefined `0xEF` instruction and halting.
```

pdobacz (approved):
Content-wise LGTM. Thank you for the rationale section.

We likely would want to pair it with an update to https://eips.ethereum.org/EIPS/eip-7620#execution-semantics which removes the extra rule for CREATE/CREATE2, causing it to similarly fail hard. The reason is: consistency, simplicity of client implementation and testing (less edge cases).

UPDATE: decided against pairing the changes, i.e. #9688 got closed

pdobacz:
@shemnon @chfast here is the accompanying PR to EIP-7620, PTAL and let's consider merging both at the same time https://github.com/ethereum/EIPs/pull/9688

chfast (approved):

PR #9684, opened 2025-04-23
Update EIP-7748: address special case of eip-7610 accounts

gballet:


EIPS/eip-7748.md
```diff
@@ -144,7 +144,13 @@ def conversion_move_to_next_account(state: State):
...
         # Move to the next account
         state._conversion_curr_account.address = address
-        state._conversion_curr_account.phase = StoragePhase("0x00")
+        curacc = state._conversion_curr_account
+
+        # Skip translating storage if the account has an empty code hash.
+        if curacc.nonce == 0 and len(curacc.code) == 0 and curacc.state_root != empty_state_root:
+            state._conversion_curr_account.phase = AccountDataPhase()
+        else:
+            state._conversion_curr_account.phase = StoragePhase("0x00")
```
jsign, line 153: I would leave this code it is and only do your change below in:

```
# EIP-161 should not be converted.
if account_exists_and_is_empty(state, curr_account.address):
            state.conversion_move_to_next_account()
            continue
```
As in, change to refer to `EIP-7610` and adjust the method.

jsign: But was the conclusion to delete them? Or only not migrate the storage?

PR #9687, opened 2025-04-23
Add EIP: 64-bit mode EVM opcodes (EVM64)

sorpaas:
([Rendered](https://github.com/sorpaas/EIPs/blob/sp-64bit-mode/EIPS/eip-7937.md))

## Abstract

This EIP introduces multibyte opcodes prefixed by `C0` for 64-bit arithmetic (`C001`-`C00B`), comparison (`C010`-`C015`), bitwise (`C016`-`C019`), stack (`DUP*` and `SWAP*`) and flow (`C056` and `C057` in "legacy", or `C0E1` and `C0E2` in EOFv1) operations.

## Motivation

Not all computations in EVM can utilize the full 256-bit integer width. It can therefore be beneficial to have a "64-bit mode" to avoid unnecessary cycles. This EIP uses a "prefix" opcode `C0`, essentially forming multibyte opcodes to avoid polluting the EVM opcode space too much.

## Rationale

When a smart contract uses the 64-bit mode, it's expected that once entered, it will want to stay in 64-bit mode, and only exit to non-64-bit mode when the computationally intensive function is finished. This EIP is designed particularly with this fact in mind.

All 64-bit opcodes only operates on the 64-bit value. It totally discards the rest 192 bits. The interpreter only needs to ensure that when it exits into non-64-bit mode and next time when a value result is read, that value has the first 192 bits reset to zero. The EVM interpreter can therefore use a typed stack for optmization:

```haskell
type StackItem = Value U256 | Value64 U64
```

The typed stack can also be implemented as a bitmap for memory alignment.

For all inputs of 64-bit opcodes, it will either read in a `Value`, when it'll only take the last 64 bits, or a `Value64`, which is what is needed. It then always outputs a `Value64`. After exiting into non-64-bit mode and upon a `Value64` is read, the interpreter then translate the value back to 256-bit `Value` by extending the first 192 bits with zero.

The 64-bit mode does not contain any opcodes which depends on the value's endianness, therefore `Value64` can also be stored in the optimal endianness of the architecture.


clarus:
I think the Move's VM can be a good source of inspiration, as it is also a stack-based VM used for smart contracts, and they have various sizes of integers.

Links to Move's implementation:

- AST: https://github.com/move-language/move-sui/blob/82e94dfbe71d78f2e74463654a52d1041d738b28/crates/move-binary-format/src/file_format.rs#L1177
- Values on the stack: https://github.com/move-language/move-sui/blob/82e94dfbe71d78f2e74463654a52d1041d738b28/crates/move-vm-types/src/values/values_impl.rs#L40 (there are also data structures in addition of integer types)
- Type-checker: https://github.com/move-language/move-sui/blob/82e94dfbe71d78f2e74463654a52d1041d738b28/crates/move-bytecode-verifier/src/type_safety.rs#L459
- Interpreter: https://github.com/move-language/move-sui/blob/82e94dfbe71d78f2e74463654a52d1041d738b28/crates/move-vm-runtime/src/interpreter.rs#L813

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-9687-64-bit-mode-evm-operations/23794
+status: Draft
+type: Standards Track
+category: Core
+created: 2025-04-23
+---
+
+## Abstract
+
+This EIP introduces multibyte opcodes prefixed by `C0` for 64-bit arithmetic (`C001`-`C00B`), comparison (`C010`-`C015`), bitwise (`C016`-`C019`), stack (`DUP*` and `SWAP*`) and flow (`C056` and `C057` in "legacy", or `C0E1` and `C0E2` in EOFv1) operations.
```
jochem-brouwer, line 15: Note that these multibyte opcodes are expensive to deposit. Each byte costs 200 gas to deposit. So although this bytecode layout is nice, it will mean that users will pay more to deposit this code. Also note that there is a limit of size of a smart contract, so if someone would use mostly these 64-bit opcodes in a contract that would thus mean the maximum size of that contract is effectively halved (half of operations are available when compared to non-64-bit opcodes)

jochem-brouwer: Aha, but I did not get this immediately (might be something to add for clarity): the "normal" opcodes are re-used here, but if these are thus prefixed by the `C0` byte they run in 64-bit mode. So this `C0` is basically an "optional opcode flag". This is very creative, I like it, but please note that this will cost more to deposit, and will also limit the total program size (doubles cost, halves code size, in the extreme cases of only using 64bit).

Since it is likely (?) that multiple 64-bit operations are chained after each other, would you consider a variant where one can set the 64-bit mode on and off? (Either two opcodes, or one opcode which pops from stack?)

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Motivation
+
+Not all computations in EVM can utilize the full 256-bit integer width. It can therefore be beneficial to have a "64-bit mode" to avoid unnecessary cycles. This EIP uses a "prefix" opcode `C0`, essentially forming multibyte opcodes to avoid polluting the EVM opcode space too much.
```
jochem-brouwer, line 19: Which computations cannot utilize the full 256 bit width?

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Prefix opcode behavior
+
+This EIP uses the prefix opcode `C0`, and it only occupies this single EVM opcode space. Upon the interpreter encountering opcode `C0`, it MUST continue to seek the next byte in code. It then executes things in "64-bit mode", based on the second byte, described below. If the execution is successful, then the interpreter MUST increase `PC` by 2 (instead of 1).
+
+If the second byte is not a valid 64-bit mode operation, then the interpreter MUST OOG.
+
```
jochem-brouwer, line 30: Might have to note that 0xC05B (5B is JUMPDEST) would mark that JUMPDEST as invalid location to jump to (because it's an immediate byte for the 0xC0 64-bit mode)

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+### General 64-bit mode behavior
+
+In 64-bit mode, all operations only work on the lowest 64-bit of each stack value. The first 192-bit is discarded. When a result value is pushed back onto the stack, then it MUST ensures that observable effects will see that the first 192-bit is zero. Note that here it's not necessary for an interpreter to reset the first 192-bit to zero every time -- if the next opcode is still in 64-bit mode, then the first 192-bit is still unobservable. We discuss the full details in the "rationale" section. The interpreter only needs to reproduce the full 256-bit value upon entering non-64-bit mode. If the full computational heavy part can be written in pure 64-bit mode, then this can result in noticable performance gain.
```
jochem-brouwer, line 33: "lowest" and "first" could be confusing, I think if you change "first 192-bit" to "topmost 192 bits" this is more clear.

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+* `G_MID64`: 4
+
+### Arithmetic opcodes
+
+The 64-bit mode arithmetic opcodes are defined the same as non-64-bit mode, except that it only operates on the last 64-bits. In the below definition, `a`, `b`, `N` is `a mod 2^64`, `b mod 2^64` and `N mod 2^64`.
+
+* ADD (`C001`) and SUB (`C003`): `a op b mod 2^64`, gas cost `G_VERYLOW64`.
+* MUL (`C002`), DIV (`C004`), SDIV (`C005`), MOD (`C006`), SMOD (`C007`), SIGNEXTEND (`C00B`): `a op b mod 2^64`, gas cost `G_LOW64`.
+* ADDMOD (`C008`), MULMOD (`C009`): `a op b % N mod 2^64`, gas cost `G_MID64`.
+* EXP (`C00A`): `a EXP b mod 2^64`, gas cost is the same as non-64-bit mode.
```
jochem-brouwer, line 51: The EXP gas costs are dynamic: it depends on the exponent, and is not a standard gas cost (not sure if this is meant by this sentence to just copy that gas cost)

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Comparison and bitwise opcodes
+
+The 64-bit mode comparison and bitwise opcodes are defined the same as non-64-bit mode, except that they only operates on the last 64-bits.
```
jochem-brouwer, line 55: Here "last 64 bits" is used. I think if you make it clear that you are talking about big endianness representation of the bits then it is clear what is meant by "highest" and "lowest".

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+
+* LT (`C010`), GT (`C011`), SLT (`C012`), SGT (`C013`), EQ (`C014`), AND (`C016`), OR (`C017`), XOR (`C018`): `a op b mod 2^64`, gas cost `G_VERYLOW64`
```
jochem-brouwer, line 57: I think for these operations it should  be made extra clear that prior of entering the 64-bit mode the top 192 bits are cleared (set to 0). Otherwise with SHR you could move topmost bits inside the low 64-bit fields before doing `mod 2^64`, thus "smuggling" high bits into the low-bit mode.

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+* ISZERO (`C015`), NOT (`C019`): `op a mod 2^64`, gas cost `G_VERYLOW64`
+
+Note that:
+
+* 64-bit EQ (`C014`) may return true for two different integers because it'll only compare the last 64-bits.
```
jochem-brouwer, line 62: Ok, this is confusing, regarding my point of SHR just now. How does SHR work if there are topmost bits? This seems to imply I can smuggle top bits into my 64 bit mode? Is this intended?

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+* 64-bit ISZERO (`C015`) may return true for non-zero 256-bit integers as long as the last 64 bits are zero.
+* BYTE (`1A`) does not have 64-bit mode because it affects endianness.
```
jochem-brouwer, line 64: You can express  BYTE as `(x >> (248 - i * 8)) & 0xFF` where we take the `i`th byte of `x`. This does not use any operation with endianness, so I think BYTE can be defined in the 64 bit variant as well (?)

EIPS/eip-7937.md
```diff
@@ -0,0 +1,145 @@
...
+
+This EIP also recommends that we reserve `C0`-`CF` for prefix (modes) opcodes. For example, an additional modes OVERFLOW can be envisioned that changes the behavior of arithmetic opcodes from wrapping to overflow OOG, which can help to reduce, for example, the extra cycles needed for `SafeMath`.
+
+## Backwards Compatibility
+
+This EIP introduces a new (prefix) opcode `C0`. `C0` was previously an invalid opcode that has little usage, and thus the backward compatibility issues are minimal.
+
+## Test Cases
+
+To be added.
```
jochem-brouwer, line 131: Also as examples these would be helpful, I am somewhat confused by how certain operations would work if the original value has most bits of the 256 available set (instead of only the available 64). For instance operations like shift right (SAR).

jochem-brouwer (commented):
Some comments! :smile: :+1:

EIPS/eip-7937.md
jochem-brouwer: Upon a first read of this EIP I did not completely get what would happen with the topmost bits. It now seems clear, we have to ignore them. Does this also mean that once these topmost 192 bits come "out" of the 64 bit opcodes, they are thus still unchaged?

So if I put 0xFF..00000001 in and then add for instance 0x00..01, this would yield 0xFF..0002 after the addition. So once I use a "normal" opcode the topmost bits are thus unchanged? I should thus see all these 64 bit operations to run on the 64 bits and ignore everything else? (I think this is stated in the EIP but I somehow did not get this at the first pass).

In your example above, I get the point, but shouldn't it be `000000ff f000000f >> 000000ff 00000001`? If we group two hex numbers (one hex is 0-f, so can contain 16 numbers = 4 bits, so 2 hex = 8 bits = one byte) together we see that the 8 hex numbers represent 32 bits. (so stack item examples are 128 bits, not 256)

EIPS/eip-7937.md
jochem-brouwer: Yes, great approach, usually EIPs are underspecified instead of overspecified so this approach would be great. For the per-opcode specification can't you "just" run the same opcode as the 256-bit variant, but before the opcode runs you take the relevant stack items and `mod 2^64` them to remove the topmost bits. This is now your input to the opcode. Now you evaluate the opcode, and for the output you have to `mod 2^64` it again. What rests is to add the original "topmost bits" (so topmost 192 bits) back into the stack item (I think you can just ADD / XOR the topmost bits in)

EIPS/eip-7937.md
jochem-brouwer: Great, thanks for the further explanation, I got a bit confused by the interpreter which (for optimization) does not directly clear out the topmost 192 bits. But the "naive" / simple way to do this is just clear out the topmost bits before doing any operation in 64-bit mode (and later optimize). Got it! :smile: :+1:

EIPS/eip-7937.md
```diff
@@ -48,7 +48,7 @@ The 64-bit mode arithmetic opcodes are defined the same as non-64-bit mode, exce
 * ADD (`C001`) and SUB (`C003`): `a op b mod 2^64`, gas cost `G_VERYLOW64`.
 * MUL (`C002`), DIV (`C004`), SDIV (`C005`), MOD (`C006`), SMOD (`C007`), SIGNEXTEND (`C00B`): `a op b mod 2^64`, gas cost `G_LOW64`.
 * ADDMOD (`C008`), MULMOD (`C009`): `a op b % N mod 2^64`, gas cost `G_MID64`.
-* EXP (`C00A`): `a EXP b mod 2^64`, gas cost is the same as non-64-bit mode.
+* EXP (`C00A`): `a EXP b mod 2^64`, gas cost is the same as non-64-bit mode EXP.
```
jochem-brouwer, line 51: If the gas cost is the same, why would I use the 64 bit variant?

lightclient (approved):

PR #9688, opened 2025-04-24
Update EIP-7620: Handling of EOF containers in CREATE/CREATE2

pdobacz:
This is a mirror to the [update to EIP-7873](https://github.com/ethereum/EIPs/pull/9669) where legacy creation txs with EOF initcode are not treated specially anymore. Here CREATE/CREATE2 attempts to deploy from EOF initcontainers stops to be treated specially.

### DIFF
--- a/EIPS/eip-7620.md
+++ b/EIPS/eip-7620.md
@@ -53,9 +53,9 @@ If the code is legacy bytecode, any of these instructions result in an *exceptio
 ### Execution Semantics
 
 - The instructions `CREATE`, `CREATE2` are made obsolete and rejected by validation in EOF contracts. They are only available in legacy contracts.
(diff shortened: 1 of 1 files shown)


chfast (approved):

shemnon:
I think we missed the window on spec changes that are not driven by ACD.  So lets close this and just accept the small amount of friction it creates.

pdobacz:
> Before I make a final move on these, I'll make sure by stripping down [ethereum/evmone#1205](https://github.com/ethereum/evmone/pull/1205), that the friction we're accepting is not huge.

Friction in the EVM (`evmone`) is OK - we just have the EOF initcode check coded twice - once for CREATE/CREATE2 instruction, second for legacy creation tx.

PR #9695, opened 2025-04-25
Update EIP-7748: refactor eip7610 account handling code

gballet:
Addresses some comments that arrived after the previous PR auto-merged.

Also fixes some linter issues.

### DIFF
--- a/EIPS/eip-7748.md
+++ b/EIPS/eip-7748.md
@@ -170,7 +170,10 @@ def state_convert(state: State, stride: int):
         curr_account = state.get_conversion_account()
         
         # Skip translating storage if the account has an empty code hash.
-        if cur_account.nonce == 0 and len(cur_account.code) == 0 and cur_account.state_root != empty_state_root:
+        # Skip storage conversion for accounts with 0 nonce and empty code.
+        # This covers the 28 EIP-7610 accounts, but would also covers all
+        # pre-eip161 accounts on other chains.
+        if cur_account.nonce == 0 and len(cur_account.code) == 0:
             state._conversion_curr_account.phase = AccountDataPhase()
         
         # Account storage.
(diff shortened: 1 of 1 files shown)


EIPS/eip-7748.md
```diff
@@ -175,10 +169,9 @@ def state_convert(state: State, stride: int):
     while n < stride and not state._conversion_finished:
         curr_account = state.get_conversion_account()
         
-        # EIP-161 should not be converted.
-        if account_exists_and_is_empty(state, curr_account.address):
-            state.conversion_move_to_next_account()
-            continue
+        # Skip translating storage if the account has an empty code hash.
+        if cur_account.nonce == 0 and len(cur_account.code) == 0 and cur_account.state_root != empty_state_root:
```
jsign, line 173: I think `and cur_account.state_root != empty_state_root` is not necessary since it is already handled below.

jsign (approved):
LGTM, left a comment.

EIPS/eip-7748.md
gballet: true

gballet: actually I don't think this is handled below, because then:

 - on L177 `curr_account.phase` is `StoragePhase`
 - on L181 `trie` is _not_ `None` since the state root isn't empty
 - so the leaves will be converted, which is not what we want.

gballet:
```suggestion
        # Skip storage conversion for accounts with 0 nonce and empty code.
        # This covers the 28 EIP-7610 accounts, but would also covers all
        # pre-eip161 accounts on other chains.
        if cur_account.nonce == 0 and len(cur_account.code) == 0:
```

jsign (approved):

PR #9698, opened 2025-04-27
Add EIP: Exponential Gas Limit Increase

dankrad:


LukaszRozmej:
I'm all for raising gas limit, by quite a lot. But I'm against committing to aggressive schedules just for the sake of it. All increases should either be manually agreed upon or if more automatic than not too aggressive. Security first.

potuz:
Is this really an "informational" EIP? it changes the default behavior of node operators from an "opt-in" an increase to an "opt-out".

S1nus:
LGTM

EIPS/eip-9698.md
```diff
@@ -0,0 +1,66 @@
...
+
+The exponential growth model ensures gradual but significant increases, allowing the network to adapt while targeting ambitious throughput goals.
+
+## Backwards Compatibility
+
+The change is non-consensus and backward compatible. Clients not implementing the EIP will continue to behave as before. Only the default behavior changes; manual configuration remains supported.
+
+## Security Considerations
+
+A rapid increase in the gas limit may stress less-optimized nodes and increase block propagation times. However, the exponential schedule with very gradual increments per epoch gives node operators and developers ample time to adapt and optimize.
```
jochem-brouwer, line 61: If the majority of node operators and developers adapt this schedule, then it will be hard to shut this down. We'd need 50%+ of block proposers who would opt-out of this schedule in order to either stay at a constant gas limit or decrease it (due to the gas limit voting mechanism).

EIPS/eip-9698.md
```diff
@@ -0,0 +1,66 @@
...
+
+The current gas limit mechanism relies on miner/operator voting, which lacks coordination and predictability. While flexible, this approach can lead to stagnation or overly cautious increases. By introducing a predictable exponential growth pattern as a client default, this EIP encourages a sustainable and transparent gas limit trajectory, aligned with expected advancements in hardware and protocol efficiency.
+
+## Specification
+
+### Schedule
+
+Let `G0 = 50,000,000` be the gas limit at the activation epoch. Let the activation epoch be Ethereum beacon chain **epoch 369017**, which corresponds to approximately June 1, 2025.
+
+Let `t` be the number of beacon chain epochs since the activation epoch (Epoch 0).
```
jochem-brouwer, line 26: I think we need a `t` and a `t_0` here where `t_0` is the activation epoch and `t` is the current epoch. Then for mainnet, `t_0` is 369017. Then `G(t) = round(G0 * m^(t-t_0))` and is only valid if `t >= t_0`

EIPS/eip-9698.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Motivation
+
+The current gas limit mechanism relies on miner/operator voting, which lacks coordination and predictability. While flexible, this approach can lead to stagnation or overly cautious increases. By introducing a predictable exponential growth pattern as a client default, this EIP encourages a sustainable and transparent gas limit trajectory, aligned with expected advancements in hardware and protocol efficiency.
+
+## Specification
+
+### Schedule
+
+Let `G0 = 50,000,000` be the gas limit at the activation epoch. Let the activation epoch be Ethereum beacon chain **epoch 369017**, which corresponds to approximately June 1, 2025.
```
jochem-brouwer, line 24: This does not reflect current `G0`, which would be `36_000_000` (approx)

jochem-brouwer (commented):
Some comments, I think some formulas or calculations/constants are not 100% correct or have missing variables

jochem-brouwer:
Also to @LukaszRozmej point:

> I'm all for raising gas limit, by quite a lot. But I'm against committing to aggressive schedules just for the sake of it. All increases should either be manually agreed upon or if more automatic than not too aggressive. Security first.

I think the current 100x in 4 years is too aggressive and I think we should approach it in steps. The mechanism here (exponential increase over epochs) is fine. But I'd rather see a 2x increase in the next 6 months for instance (and yes this is at the slow, and very conservative side). Main point: let's do this in steps before we have the majority of the nodes committed to this schedule and we will have a problem when we find out the limit is getting too high (then we thus need to coordinate with all nodes to opt-out and shut it down again)

EIPS/eip-9698.md
```diff
@@ -0,0 +1,66 @@
...
+
+Let `t` be the number of beacon chain epochs since the activation epoch (Epoch 0).
+
+To reach a 10x increase over 164,250 epochs (2 years), the gas limit at epoch `t` is:
```
SanLeo461, line 28: There are currently discussions to reduce slot time, which assuming epochs remain 32 slots in length, would reduce the 2 year timeline if implemented
I suppose if that happens, we'd just have to make sure to update this EIP accordingly? Maybe add a note for what happens in that case.

EIPS/eip-9698.md
```diff
@@ -0,0 +1,66 @@
+---
+eip: 9698
```
g11tech, line 2:
```suggestion
eip: 7938
```

EIPS/eip-9698.md
dankrad: Builders can currently override the gas vote, right?

So an intervention with a small number of parties can probably get an emergency stop done quite quickly.

EIPS/eip-9698.md
dankrad: Yeah it intentionally kicks off with a step increase because we are just "catching up" to what we know is already safe and wanted to do before Pectra (but couldn't due to the max payload problem).

dankrad:
> 2x increase in the next 6 months

This schedule actually only increases by 1.77x over 6 months (`10^(1/4)`).

nlgripto (approved):

lightclient (approved):

EIPS/eip-9698.md
terencechain: I think by default the consensus layer client will check the builder's gas limit, and if the gas limit in the header does not match the local computation, the consensus layer client will rebuild the block locally

kavehtehrani:
Finally a renewed focus on L1. Thanks Dankrad!

PR #9710, opened 2025-04-30
Update EIP-7702: Remove trie concept from the vm specification

JereSalo:
## Motivation
- Remove concept of trie from the EVM specification in EIP-7702 and replace it with "emptiness" for simplicity.

## Rationale
- I think it would be good to abstract the EVM from the trie implementation. The EVM should just know if an account is empty or not. This simplification has been done almost 9 years ago and we should stick to it. For clients that don't implement forks older than Tangerine Whistle it adds unnecessary complexity because they'd need to track if an account is empty but still exists in the trie, something that's not possible in these scenarios.

From [EIP-158](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-158.md)
"Whenever the EVM checks if an account exists, emptiness is treated as equivalent to nonexistence. Particularly, note that this implies that, once this change is enabled, there is no longer a meaningful difference between emptiness and nonexistence from the point of view of EVM execution."

## Possible Counter Argument:
The only counter argument that I can think of for this is that in the case of some clients it may happen that there are empty accounts in the trie but I still think it is better to treat those as non-existent in the EVM.

### DIFF
--- a/EIPS/eip-7702.md
+++ b/EIPS/eip-7702.md
@@ -112,9 +112,8 @@ following:
 4. Add `authority` to `accessed_addresses`, as defined in [EIP-2929](./eip-2929.md).
 5. Verify the code of `authority` is empty or already delegated.
 6. Verify the nonce of `authority` is equal to `nonce`.
-   * If `authority` does not exist in the trie, verify `nonce` is equal to `0`.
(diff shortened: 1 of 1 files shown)


tomip01 (approved):

lightclient (approved):
Seems reasonable to me. Empty is not the best, most well defined term in EIP language, so if someone comes here for the future to understand - please look at EIP-161 and EIP-7523.

PR #9712, opened 2025-05-01
Update EIP-3416: fix typo in BASE_FEE formula

famouswizard:


g11tech (approved):
might still require author's approvl so may be we can just force merge this apparent typo fix @SamWilsn

PR #9740, opened 2025-05-09
Update EIP-7600: Move to Final

timbeiko:
With Pectra live on mainnet, the Meta EIP should be moved to Final. Note that this will require all included EIPs to be moved to final: 

- [x] [EIP-2537](./eip-2537.md): Precompile for BLS12-381 curve operations  
- [x] [EIP-2935](./eip-2935.md): Save historical block hashes in state  
- [x] [EIP-6110](./eip-6110.md): Supply validator deposits on chain  
- [x] [EIP-7002](./eip-7002.md): Execution layer triggerable exits  
- [x] [EIP-7251](./eip-7251.md): Increase the MAX_EFFECTIVE_BALANCE  
- [x] [EIP-7549](./eip-7549.md): Move committee index outside Attestation  
- [x] [EIP-7623](./eip-7623.md): Increase calldata cost  
- [x] [EIP-7685](./eip-7685.md): General purpose execution layer requests  
- [x] [EIP-7691](./eip-7691.md): Blob throughput increase  
- [x] [EIP-7702](./eip-7702.md): Set EOA account code  
- [x] [EIP-7840](./eip-7840.md): Add blob schedule to EL config files (Informational)  
- [x] [EIP-7642](./eip-7642.md): eth/69 - Drop pre-merge fields (Networking)

### DIFF
--- a/EIPS/eip-7600.md
+++ b/EIPS/eip-7600.md
@@ -7,7 +7,7 @@ discussions-to: https://ethereum-magicians.org/t/eip-7600-hardfork-meta-prague-e
 status: Final
 type: Meta
 created: 2024-01-18
-requires: 2537, 2935, 6110, 7002, 7251, 7549, 7623, 7685, 7691, 7702, 7840
+requires: 2537, 2935, 6110, 7002, 7251, 7549, 7623, 7685, 7691, 7702
 ---
 
 ## Abstract
--- a/EIPS/eip-7600.md
+++ b/EIPS/eip-7600.md
@@ -4,8 +4,7 @@ title: Hardfork Meta - Pectra
 description: EIPs included in the Prague/Electra Ethereum network upgrade.
 author: Tim Beiko (@timbeiko)
 discussions-to: https://ethereum-magicians.org/t/eip-7600-hardfork-meta-prague-electra/18205
-status: Last Call
-last-call-deadline: 2025-04-01
+status: Final
 type: Meta
 created: 2024-01-18
 requires: 2537, 2935, 6110, 7002, 7251, 7549, 7623, 7685, 7691, 7702, 7840
@@ -17,9 +16,9 @@ This Meta EIP lists the EIPs formally Scheduled for Inclusion in the Prague/Elec
 
 ## Specification
 
-A formal definition for `Scheduled for Inclusion` can be found in [EIP-7723](./eip-7723.md).
+A formal definition for `Included` can be found in [EIP-7723](./eip-7723.md).
 
-### EIPs Scheduled for Inclusion  
+### Included EIPs
 
 #### Core EIPs
 
@@ -33,7 +32,6 @@ A formal definition for `Scheduled for Inclusion` can be found in [EIP-7723](./e
 * [EIP-7685](./eip-7685.md): General purpose execution layer requests 
 * [EIP-7691](./eip-7691.md): Blob throughput increase
 * [EIP-7702](./eip-7702.md): Set EOA account code
-
 
 #### Other EIPs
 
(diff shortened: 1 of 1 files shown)


ralexstokes:
for 2537: https://github.com/ethereum/EIPs/pull/9743

jochem-brouwer:
Hi @timbeiko, small side-note, the "rule" that `Included` EIPs for the upgrade should be marked as `Final` is not part of https://eips.ethereum.org/EIPS/eip-7723 :wink: 

(I am assuming that 7723 is the current "draft spec" of the process, please correct me if I'm wrong)

EIPS/eip-7600.md
```diff
@@ -4,8 +4,7 @@ title: Hardfork Meta - Pectra
 description: EIPs included in the Prague/Electra Ethereum network upgrade.
 author: Tim Beiko (@timbeiko)
 discussions-to: https://ethereum-magicians.org/t/eip-7600-hardfork-meta-prague-electra/18205
-status: Last Call
-last-call-deadline: 2025-04-01
+status: Final
 type: Meta
 created: 2024-01-18
 requires: 2537, 2935, 6110, 7002, 7251, 7549, 7623, 7685, 7691, 7702, 7840
```
poojaranjan, line 10:
```suggestion
requires: 2537, 2935, 6110, 7002, 7251, 7549, 7623, 7685, 7691, 7702
```

poojaranjan: There are 10 Core EIPs `Included` in Pectra. [EIP-7840](https://eips.ethereum.org/EIPS/eip-7840) is an `Informational` EIP. I missed fixing this in the previous [PR](https://github.com/ethereum/EIPs/pull/9726).

g11tech: but we should still require it imo since its coordinated with the hardfork

poojaranjan: Given that EIP-1 defines Informational EIPs as optional, I'd be reluctant to add it in the "Requires" section. However, the proposal is added in the "Specification" section under "Other EIPs" along with Networking EIP [EIP-7642](https://eips.ethereum.org/EIPS/eip-7642).

poojaranjan: Historically, Upgrade Meta EIPs have only included Core EIPs. 

As per my understanding, the definition of **Core EIPs** includes proposals that are either considered for a network upgrade or added asynchronously to the mainnet protocol.

While we’ve seen **Networking EIPs** deployed to mainnet in the past, they haven’t been treated as part of the upgrade process nor listed in the “Required” section of Meta EIPs.

I believe the key difference between content in a Meta EIP and an Informational EIP is that the Ethereum community may choose to follow Informational EIPs, but the process and information in Meta EIPs are expected to be followed—they can’t be ignored. **Including an Informational EIP in the “Required” section could unintentionally blur this line.**

I am fine with Informational EIP being part of the Meta EIP document to add related information. But, reluctant to include it under “Required” without broader consensus.

Happy to raise this at an upcoming EIPIP meeting for discussion. If there’s general agreement to evolve the process, I think we should also explore the following questions:

- What fundamentally distinguishes Meta and Informational EIPs today?
- Should we revise the definition of Informational EIPs if they’re no longer strictly optional?
- If we’re open to including Informational EIPs, why not include Networking EIPs?
- Should we retroactively update Meta EIPs to reflect Networking EIPs from past upgrades?
- Should we measure and upgrade with the highest number of Core EIPs or including Informational plus Meta, or any other supporting EIPs at the time?

timbeiko: We did something similar for [Shapella](https://github.com/ethereum/execution-specs/blob/8dbde99b132ff8d8fcc9cfb015a9947ccc8b12d6/network-upgrades/mainnet-upgrades/shanghai.md). I think we should stick with it, optimizing the Meta EIP to communicate the scope of the fork as comprehensively as possible. 

All the questions you raise are important; I agree we should have a deeper conversation about the categories for EIPs, but don't want to block this PR on it 😄

PR #9762, opened 2025-05-12
Update EIP-7793: Change TXINDEX to ASSERTTXINDEX

Marchhill:
Change to assertion so transaction index can be statically determined.

### DIFF
--- a/EIPS/eip-7793.md
+++ b/EIPS/eip-7793.md
@@ -31,7 +31,7 @@ The opcode takes a single input from the stack, an offset into calldata. From th
 The `ASSERTTXINDEX` instruction will revert under the following conditions:
 
 - The current transaction index does not match the expected value loaded from calldata.
-- If `PC != 1` or the instruction executed at `PC == 0` was not a `PUSH*`.
+- If `gasUsed > 3` (to allow a `PUSH*` or similar instruction before to provide an argument).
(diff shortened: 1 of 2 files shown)


EIPS/eip-7793.md
```diff
@@ -12,37 +12,45 @@ created: 2024-10-17
...
+The new opcode aims to improve support for encrypted mempools. Transactions in an encrypted mempool are ordered while the transactions are encrypted, before being decrypted and included onchain at the top of the block. If the builder does not respect the order when including the decrypted transactions then they could frontrun decrypted transactions. The new opcode can be used to make this impossible; if a decrypted transaction is not included in the correct order it will revert.
 
 ## Specification
 
-If `block.timestamp >= TBD` a new precompiled contract `TXINDEX` shall be created at address `TBD`.
+If `block.timestamp >= TBD`, a new opcode `ASSERTTXINDEX` shall be created at `TBD`. It takes as an argument the specified transaction index as an offset into calldata.
+
+### Input
+
+Takes as input a 32 byte index into calldata, matching the semantics of `CALLDATALOAD`.
```
smartprogrammer93, line 27: clarify where this input comes from. it should be an immediate in the bytecode imo. instead of being read from the stack or the memory. not sure if that is possible in the EVM or not. double check

PR #9776, opened 2025-05-15
EIP-7870: Lowers requirements for a "Full Node".md

MicahZoltu:


dankrad:
I don't think this is an appropriate change for this EIP, as it isn't only for current usage, but also in expectation of future L1 scaling: https://ethereum-magicians.org/t/hardware-and-bandwidth-recommendations-for-full-nodes-and-validators/22675/8

MicahZoltu:
(replied in Magicians thread, we can continue conversation there until there is agreement on what the point of this EIP is)

PR #9784, opened 2025-05-16
Update EIP-6110: ensure ints are decoded as uints

jochem-brouwer:
This change clarifies that the python `int.from_bytes` should use `signed=False`. Therefore, it will never return a negative number. Note, that `signed=False` is the default setting for `int.from_bytes`. This is thus a clarification to avoid decoding it as signed int. It is now thus decoded as `uint`.

### DIFF
--- a/EIPS/eip-6110.md
+++ b/EIPS/eip-6110.md
@@ -89,11 +89,11 @@ def is_valid_deposit_event_data(deposit_event_data: bytes) -> bool:
     if len(deposit_event_data) != 576:
         return False
 
-    pubkey_offset = int.from_bytes(deposit_event_data[0:32], byteorder='big')
-    withdrawal_credentials_offset = int.from_bytes(deposit_event_data[32:64], byteorder='big')
-    amount_offset = int.from_bytes(deposit_event_data[64:96], byteorder='big')
-    signature_offset = int.from_bytes(deposit_event_data[96:128], byteorder='big')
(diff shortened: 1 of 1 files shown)


mkalinin (approved):
Thank you! Note that this PR doesn’t change the spec but rather make it clearer, thus fine to be merged after the EIP has been move to Final.

jochem-brouwer:
Thank you! Thanks for the note that this indeed does not change the spec, it only clarifies it. Would be great if the EIP team could merge this, thanks :smile: :+1:

lightclient (approved):

EIPS/eip-6110.md
mustafacryptolife, line 1: [](url)

mustafacryptolife (commented):
[](url)

EIPS/eip-6110.md
```diff
@@ -89,11 +89,11 @@ def is_valid_deposit_event_data(deposit_event_data: bytes) -> bool:
     if len(deposit_event_data) != 576:
         return False
 
-    pubkey_offset = int.from_bytes(deposit_event_data[0:32], byteorder='big')
```
Nithin200487, line 92: 92 -     pubkey_offset  = 2fb6edcb8eb5822279dcd5a0fa92f7f83909d610.from_Eth(deposit_event_,data[0:32],  ethorfer= 'big')

PR #9790, opened 2025-05-18
Move eip-7932 to review

SirSpudlington:


SirSpudlington:
Need to address comments first

PR #9792, opened 2025-05-19
Add EIP: Sorted Key-Value Store with Iteration for the EVM

keyvank:


EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,99 @@
...
+        bytes memory key,
+        bytes memory value
+    );
+}
+```
+
+### Notes
+
+* **Lexicographical Sorting**: Keys are sorted using byte-wise lexicographical order.
+* **Gas Efficiency**: All operations are designed to be efficient. The underlying Ethereum database (e.g., LevelDB or RocksDB) natively supports key iteration and sorted access.
```
s1na, line 66: One thing you have to note is that the keys are hashed when inserting an item in the trie. This means the sorted access doesn't come out of the box.

keyvank: Yup! A separate storage is probably needed since a kv-store may have keys of different sizes (even longer than 32 bytes)

keyvank: I just wonder how a "storage proof" would look like in that case? Since we definitely cant store a sorted kv store within a patricia trie

bomanaps, line 66: There are some important technical details missing here — for example, storage limits like the maximum key size and maximum value size. Also, gas costs: a table showing the base and per-byte costs for set, get, remove, and next() would be helpful.

bomanaps:
I have a question, you return a bytes32 iterator — how is this maintained between calls? My point is, if the iterator contains just a position pointer, it creates serious implementation challenges across different client implementations.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,99 @@
...
+created: 2025-05-19
+---
+
+## Simple Summary
+
+This EIP proposes the introduction of a native Key-Value (KV) store in the EVM that would allow contracts to efficiently store, access, and ***iterate*** over key-value pairs in a ***sorted*** manner. This would enhance the capabilities of smart contracts by allowing better data management, particularly for use cases that involve ordered data, such as ranking systems or order-books.
+
+## Abstract
+
+Currently, the EVM provides basic storage functionalities for smart contracts, primarily in the form of mappings (key-value pairs). However, mappings are unordered, and there is no built-in support for iterating over them in a sorted order. This proposal suggests adding a new functionality to the EVM to support a sorted KV store, enabling smart contracts to:
```
g11tech, line 19: mappings is a solidity concept, slot storage has 32 byte key index and is conceptually ordered

EIPS/eip-xxxx.md
g11tech: keys are of fixed size in slit storage, doesn't matter how they are stored (hashed or not), you can iterate them as 32 byte numbers

keyvank: We can't mix the data we want to index with the contract storage data. As you mentioned, the keys in contract storage are often the result of a hash, so contract storage entries would be scattered randomly among our indexed data, making the whole approach ineffective. Therefore, a separate storage system is needed.

EIPS/eip-xxxx.md
keyvank: Slot storage is still some kind of KV-store (Thus a mapping). Even though the slot-storage items are ordered in LevelDB/RocksDB side, the contracts still can't iterate through them.

EIPS/eip-xxxx.md
```diff
@@ -0,0 +1,99 @@
...
+discussions-to: https://ethereum-magicians.org/t/eip-xxxx-sorted-key-value-store-with-iteration-for-the-evm/24267
+status: Draft
+type: Standards Track
+category: Core
+created: 2025-05-19
+---
+
+## Simple Summary
+
+This EIP proposes the introduction of a native Key-Value (KV) store in the EVM that would allow contracts to efficiently store, access, and ***iterate*** over key-value pairs in a ***sorted*** manner. This would enhance the capabilities of smart contracts by allowing better data management, particularly for use cases that involve ordered data, such as ranking systems or order-books.
```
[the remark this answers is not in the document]
keyvank: Fixed

EIPS/eip-xxxx.md
keyvank: Done. Although let's keep the gas-costs TBD for now

s1na: > We can't mix the data we want to index with the contract storage data. As you mentioned, the keys in contract storage are often the result of a hash, so contract storage entries would be scattered randomly among our indexed data, making the whole approach ineffective. Therefore, a separate storage system is needed.

I don't think the alternative is workable. There is a reason why the keys are hashed in the trie. It's to avoid targeted attacks for making a part of the trie very deep. The tree which contains the storage should be roughly uniform. Not saying that the feature you are seeking is unworkable, just this solution might not pan out.

EIPS/eip-xxxx.md
dajuguan: The gas cost would be incredibly high if many keys are iterated, as they involve random reads and are not easily parallelizable at database. I question the practicality of this use case.

keyvank: @dajuguan You won't randomly read, you'll perform a SEEK operation to jump to the point of database you would like to read and iterate from there.

EIPS/eip-xxxx.md
keyvank: @s1na Yes there has to be a way!

EIPS/eip-7948.md
```diff
@@ -0,0 +1,139 @@
...
+- **Off-Chain Solutions**: Currently, many developers use off-chain solutions, such as oracles or indexing services, to maintain sorted data. However, these solutions increase complexity and reliance on external infrastructure.
+- **Custom Sorting Logic**: Developers could manually implement sorting within contracts, but this requires complex logic and would be costly in terms of gas fees. The native support for sorted KV stores would streamline this process.
+
+## Backwards Compatibility
+
+The new functionality would not break existing smart contracts but would add a new method of managing key-value pairs. It is fully compatible with the existing Ethereum Virtual Machine (EVM) and can coexist with current storage models such as mappings.
+
+## Security Considerations
+
+Since the values in Indexed Storage can be of arbitrary size, a key associated with a large value could result in unexpectedly high gas costs when read. This could lead to the caller unintentionally paying excessive gas fees. To mitigate this risk, one solution would be to limit values to 32 bytes, similar to the fixed-size keys, ensuring more predictable gas costs and preventing excessive storage usage.
```
g11tech, line 133: on system level, every storage is read as 32 bytes, its upto compilers to constuct high level data structures.

so for what you are proposing is the security consideration is: that since keys are stored in a sequence it will lead to an unbalanaced state tree with higher costs to access those keys

EIPS/eip-xxxx.md
pochenai: Yes, but actually my concern shares with @s1na. Currently, keys of mapping slots are hashed to ensure a balanced tree, and access patterns in underlying DB are random reads. If you're proposing an alternative storage model for this EIP, I have two questions:
1. How do you plan to maintain a balanced tree structure without hashing the keys?
2. This approach wouldn't be backward compatible, correct? There would be no way to index or access existing mappings under the old database schema?

Dhy2928 (approved):

Dhy2928 (approved):

Dhy2928 (approved):

PR #9799, opened 2025-05-20
Update EIP-7782: Move to Draft

benaadams:

### DIFF
--- a/EIPS/eip-7782.md
+++ b/EIPS/eip-7782.md
@@ -38,6 +38,19 @@ SLOT_SCHEDULE:
 
 The parameters and schedules above are purely illustrative. Actual values and schedules are beyond the scope of this specification.
 
+### Adjustment to gas and blob limits
+
+The first execution block after the fork needs to specify half the previous gas limit. With this fork, all gas limit settings are reinterpreted as "gas per 12 seconds". I.e., if the user configuration or otherwise default gas limit vote was 36,000,000, the client should now vote for 18,000,000. All values are rounded down to integers.
+
+The blob target and limit are also halved, rounded down to the nearest integer.
+
+### Attestation deadlines
+
+All clients need to be reconfigured so that deadlines can be configured in milliseconds for finer granularity. Currently the slot is split into three equal intervals, with proposal happening at time 0, and attestations and aggregations being sent at one thirds and two thirds of the slot time. We change this to a new uneven schedule, in order to give sufficient time for block propagation, which takes the most time:
+ * Block proposal at 0 ms
+ * Attestations at 3000 ms
+ * Aggregates at 4500 ms
+
 ## Rationale
 
 This proposal balances user experience, economic efficiency, and network stability:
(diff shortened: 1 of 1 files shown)


EIPS/eip-7782.md
```diff
@@ -1,173 +1,58 @@
...
-- **Increased Frequency**:
-
-  - **Block Proposals**: Validators will propose blocks more frequently due to shorter slots.
-  - **Attestations**: Validators will need to perform attestation duties at shorter intervals.
-
-- **Resource Management**:
-
-  - **Client Performance**: Optimize validator clients to handle increased operational demands without excessive resource consumption.
-  - **Bandwidth Usage**: Ensure that validators with lower bandwidth capacities can handle the increased frequency of messages.
+The parameters and schedules above are purely illustrative. Actual values and schedules are beyond the scope of this specification.
```
dankrad, line 39:
```suggestion
The parameters and schedules above are purely illustrative. Actual values and schedules are beyond the scope of this specification.

### Adjustment to gas and blob limits

The first execution block after the fork needs to specify half the previous gas limit. With this fork, all gas limit settings are reinterpreted as "gas per 12 seconds". I.e., if the user configuration or otherwise default gas limit vote was 36,000,000, the client should now vote for 18,000,000. All values are rounded down to integers.

The blob target and limit are also halved, rounded down to the nearest integer.

### Attestation deadlines

All clients need to be reconfigured so that deadlines can be configured in milliseconds for finer granularity. Currently the slot is split into three equal intervals, with proposal happening at time 0, and attestations and aggregations being sent at one thirds and two thirds of the slot time. We change this to a new uneven schedule, in order to give sufficient time for block propagation, which takes the most time:
 * Block proposal at 0 ms
 * Attestations at 3000 ms
 * Aggregates at 4500 ms
```

EIPS/eip-7782.md
```diff
@@ -1,173 +1,73 @@
...
+### Attestation deadlines
 
-- **Resource Management**:
+All clients need to be reconfigured so that deadlines can be configured in milliseconds for finer granularity. Currently the slot is split into three equal intervals, with proposal happening at time 0, and attestations and aggregations being sent at one thirds and two thirds of the slot time. We change this to a new uneven schedule, in order to give sufficient time for block propagation, which takes the most time:
 
-  - **Client Performance**: Optimize validator clients to handle increased operational demands without excessive resource consumption.
-  - **Bandwidth Usage**: Ensure that validators with lower bandwidth capacities can handle the increased frequency of messages.
+- Block proposal at 0 ms
+- Attestations at 3000 ms
+- Aggregates at 4500 ms
```
dankrad, line 54:
```suggestion
- Aggregates at 4500 ms

### Issuance and staking

This EIP should not affect the issuance and rewards per time. As the issuance is currently implemented on a per slot basis, it has to be halved to keep the issuance and staking rewards per time constant:

```
BASE_REWARD_FACTOR = 32 # previously 64
```
```

lightclient (approved):

PR #9801, opened 2025-05-20
Add EIP: Genesis File Format

jflo:
JSON schema for genesis.json


EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+
+
+```json
+
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$defs": {
+  	"hexOrDecimal48": {
+    		"type": "string",
+    		"pattern": "^(0x[0-9a-fA-F]{1,12}|[0-9]{1,12})$"
```
jochem-brouwer, line 81: Is this correct? Seems like 12 0-9 digits are a decimal48? This would fit in a hex48 but not in a decimal48, would need more numbers (15 numbers? `2^48`?)

EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+  	},
+    "address": {
+      "type": "string",
+      "pattern":  "^0x[0-9a-fA-F]{40}$"
+    },
+    "hash": {
+      "type": "string",
+      "pattern": "^0x[0-9a-fA-F]{64}$"
```
jochem-brouwer, line 89: Do we allow uppercase hash here?

EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$defs": {
+  	"hexOrDecimal48": {
+    		"type": "string",
+    		"pattern": "^(0x[0-9a-fA-F]{1,12}|[0-9]{1,12})$"
+  	},
+    "address": {
+      "type": "string",
+      "pattern":  "^0x[0-9a-fA-F]{40}$"
```
jochem-brouwer, line 85: Do we allow uppercase addresses here? (Could be for checksums?)

EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+  "properties": {
+    "config": {
+      "type": "object",
+      "properties": {
+        "chainId": { "type": "integer" },
+        "homesteadBlock": { "type": "integer" },
+        "daoForkBlock": { "type": "integer" },
+        "eip150Block": { "type": "integer" },
+        "eip155Block": { "type": "integer" },
+        "eip158Block": { "type": "integer" },
```
jochem-brouwer, line 104: We need to fix these fork names and not use the eip names. 

EIP150 is TangerineWhistle
EIP155 is SpuriousDragon
EIP158 is never shipped (but EIP161 is, and was shipped in SpuriousDragon).

So let's add `tangerineWhistleBlock` and `spuriousDragonBlock`

EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+        "constantinopleBlock": { "type": "integer" },
+        "petersburgBlock": { "type": "integer" },
+        "istanbulBlock": { "type": "integer" },
+        "muirGlacierBlock": {"type": "integer"},
+        "berlinBlock": { "type": "integer" },
+        "londonBlock": { "type": "integer" },
+        "arrowGlacierBlock": { "type": "integer" },
+        "grayGlacierBlock": { "type": "integer" },
+        "terminalTotalDifficulty": { "type": "integer" },
+        "mergeNetsplitBlock": { "type": "integer"},
```
jochem-brouwer, line 115: This needs documentation because this is confusing. I remember I researched this a month or so back but I now don't completely recall what this was. I think this had to do with a forkID to signal if you are merge-ready or not.

EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+  },
+  "additionalProperties": true
+}
+```
+
+## Rationale
+
+There are a growing number of EIPs that propose improvements to how a network is configured at genesis:
+
+[Add Blob Schedule to EL Config File](eip-7840)
```
jochem-brouwer, line 176:
```suggestion
[Add blob schedule to EL config files](eip-7840)
```

EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+
+[Blob Paramter Only Hardforks](eip-7892)
```
jochem-brouwer, line 178:
```suggestion
[Blob Parameter Only Hardforks](eip-7892)
```

EIPS/eip-8888.md
```diff
@@ -0,0 +1,194 @@
...
+
+[eth_config JSON-RPC method](eip-7910)
```
jochem-brouwer, line 180:
```suggestion
[eth_config JSON-RPC Method](eip-7910)
```

jochem-brouwer (commented):
Some small comments.

I noticed in the schema we allow both uppercase and lowercase hexadecimal, I'm ok with this, but should we not make the spec rather strict and just only use lowercase?

EIPS/eip-8888.md
jochem-brouwer: Also seems like the decimals are prepended with `0x`?

jflo: great catch. tested with 0x800000000000 and 140737488355328 i.e. 2^48-1

EIPS/eip-8888.md
jflo: They've been added, but didn't remove eip15*Block, since Geth needs 'em:

https://github.com/ethereum/go-ethereum/blob/master/params/config.go#L391

jochem-brouwer: Right, yes, I see. In this specific case I think we need to change the format. This format now introduces 3 forks (where there should be 2 -> 155/158 are split up), the name is incorrect (EIP158), and they are not indexed like others by their hardfork name.

Here is an excerpt from the mapper in hive:

```jq
    "eip150Block": env.HIVE_FORK_TANGERINE|to_int,
    "eip150Hash": env.HIVE_FORK_TANGERINE_HASH,
    "eip155Block": env.HIVE_FORK_SPURIOUS|to_int,
    "eip158Block": env.HIVE_FORK_SPURIOUS|to_int,
```

https://github.com/ethereum/hive/blob/874e7bc198afc224742706862ea40856a01de73b/clients/go-ethereum/mapper.jq#L42

EIPS/eip-8888.md
jochem-brouwer: For mainnet this block seems to be unset: https://github.com/ethereum/go-ethereum/blob/63740b7aca732f6c204c56a9e4b4932f5251e4e0/params/config.go#L60 

So maybe we can remove this

EIPS/eip-8888.md
```diff
@@ -0,0 +1,196 @@
...
+
+The canonical genesis file MUST be a JSON object with the following top-level fields:
+
+### Top-Level Fields
+
+| Field           | Description                                                     |
+|-----------------|-----------------------------------------------------------------|
+| `config`        | Chain configuration object.                                     |
+| `alloc`         | Map of addresses to pre-allocated balances and/or code/storage. |
+| `nonce`         | Hex-encoded nonce (8 bytes).                                    |
```
jochem-brouwer, line 30: Have to check if this gets padded to 8 bytes (I think it does not)

EIPS/eip-8888.md
```diff
@@ -0,0 +1,195 @@
+---
+eip: 8888
+title: Genesis File Format
+description: schema for genesis.json files
```
jochem-brouwer, line 4:
```suggestion
description: Schema for `genesis.json` files
```

MustardseedX (commented):
Ok

EIPS/eip-8888.md
```diff
@@ -0,0 +1,195 @@
...
+
+### `alloc` Object
+
+The `alloc` field defines the initial state at genesis. It maps addresses (as lowercase hex strings) to the following object:
+
+| Field          | Description                                     |
+|----------------|-------------------------------------------------|
+| `balance`      | decimal balance in wei.                         |
+| `code`         | Hex-encoded EVM bytecode.                       |
+| `nonce`        | decimal value.                                  |
```
jochem-brouwer, line 66: We should limit this field to `uint64`, see https://eips.ethereum.org/EIPS/eip-2681

EIPS/eip-8888.md
jflo: Yes, i think we should per https://eips.ethereum.org/EIPS/eip-55

EIPS/eip-8888.md
jflo: geth allows both

PR #9804, opened 2025-05-22
Add EIP-xxxx: Encode chain id with transaction hash

microbecode:
Suggestion for a new standard.

This standard proposes a way to encode the combination of a chain ID and a transaction hash into one string.

This PR is switched over from the ERC repo: https://github.com/ethereum/ERCs/pull/1036

### DIFF
--- a/EIPS/eip-encode-idhash.md
+++ b/EIPS/eip-encode-idhash.md
@@ -1,0 +1,48 @@
+---
+title: Encode chain id with transaction hash
+description: A standard for encoding a chain ID and transaction hash into a unique string format
+author: Lauri Peltonen (@microbecode)
+discussions-to: https://ethereum-magicians.org/t/a-new-standard-for-encoding-chain-id-transaction-hash/23782
+status: Draft
+type: Standards Track
+category: ERC
+created: 2025-05-10
+requires: 155
+---
+
+## Abstract
+
+This standard proposes a way to encode the combination of a chain ID and a transaction hash into one string.
+
+## Motivation
+
+Looking up a transaction by its hash always requires the context of the chain - a transaction hash alone is not enough to identify the used chain. If the chain information is included in the string itself, finding the right chain for the transaction is easy.
+
+Such strings can then be used, for example, in a forwarder service that forwards to the correct blockchain explorer.
+
+## Specification
+
+The encoded string has three components:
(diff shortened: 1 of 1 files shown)


microbecode:
~Waiting for an EIP number.~

Waiting to be merged as draft.

EIPS/eip-encode-idhash.md
```diff
@@ -0,0 +1,49 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+## Rationale
+
+The chain ID is the most important detail when routing queries based on this standard and is therefore the first element in the string. The transaction hash is the second most important element.
+
+The suffix `tx` is used to differentiate from, for example, addresses. Without the `tx` it would remain unclear whether an encoded string refers to an address, a transaction hash or something else.
+
+## Security Considerations
```
mustafacryptolife, line 43: 0031687571922
M Saleh

microbecode: wait, what?

EIPS/eip-7950.md
```diff
@@ -0,0 +1,50 @@
...
+requires: 155
+---
+
+## Abstract
+
+This standard proposes a way to encode the combination of a chain ID and a transaction hash into one string.
+
+## Motivation
+
+Looking up a transaction by its hash always requires the context of the chain - a transaction hash alone is not enough to identify the used chain. If the chain information is included in the string itself, finding the right chain for the transaction is easy.
```
jochem-brouwer, line 20: I would include here that for non-legacy type transactions the `chainId` is included in the transaction object, which thus uniquely maps the data to a chain-id unique hash. The same applies to EIP-155 replay protected transactions. But, obviously, this map works only from data => hash and not the other way around.

(This caused slight confusion in my head on why to include chainId "in" the hash (since this is "part" of it), but this is not the case, it adds the chain id as identifer to the hash to also make the "modified hash" => chain id map possible (side-by-side the other way around))

microbecode: Thanks for the feedback!

Good point about the transaction object. A lot of places may include the chain id, but nobody really wants to store a whole transaction object in places just to keep the chain id reference.

I added now a mention. Thanks again!

g11tech (approved):

PR #9810, opened 2025-05-23
Update EIP-695: Update JSON-RPC hex value encoding reference to ethereum.org

GarmashAlex:
Replaced the outdated link to the Ethereum Wiki for JSON-RPC hex value encoding in EIP-695 with the current and maintained documentation on ethereum.org. This ensures that readers are directed to an up-to-date and reliable source for the encoding specification. No other content was changed.

### DIFF
--- a/EIPS/eip-695.md
+++ b/EIPS/eip-695.md
@@ -90,7 +90,7 @@ Implementers should take care to implement `eth_chainId` correctly and promote i
 
 ## Reference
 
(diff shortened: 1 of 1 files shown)


g11tech (approved):

SamWilsn:
It would be better to reproduce the rules here (respecting any relevant licenses) instead of linking, but this'll do for now.

PR #9811, opened 2025-05-23
Update EIP-7864: Add test cases and reference link to EIP-7864 Test Cases section

VolodymyrBg:
Replaced the TODO in the Test Cases section with practical test scenarios for the unified binary tree: added basic cases (insertion, update, deletion, merkelization, edge cases) and a link to the reference implementation. Included a note that final test vectors will be published after the hash function is finalized.


mustafacryptolife:
[]()

PR #9818, opened 2025-05-25
Update EIP-7892: harden BPO forks spec.

raulk:
This is a minimal changeset on top of the existing BPO hardforks spec to solve some loose ends. 

Concretely, this adjusts the following at the consensus layer:
- Assigns a `fork_version` to pure BPO forks.
- Updates `compute_fork_version()` to take BPO forks into account.
- Specifies ENR field mapping to account for upcoming BPO forks.
- Notes that p2p topics will roll over.

Furthermore, this makes a broader adjustment to avoid future ambiguity:
- We now require that all future blob parameter changes be performed through blob schedule configuration—even those introduced by regular (named) hardforks.

---

While thinking through this, I realized there may be value in formalizing the broader concept of "parametric forks" by defining how they interact with regular hard forks, establishing a process for bringing additional parameters under this management mechanism, and iterating on client config schemas (https://github.com/eth-clients) to introduce time-based triggers as a top-level organizational axis. Perhaps the subject of a future proposal.

### DIFF
--- a/EIPS/eip-7892.md
+++ b/EIPS/eip-7892.md
@@ -86,17 +86,16 @@ To facilitate these changes on the execution layer, the `blobSchedule` object sp
 
 ### Consensus layer configuration
 
-On the consensus layer, a new parameter is added to the configuration. The `FORK_VERSION` is only specified here for pure BPO forks. Regular forks have their own fork version specified in source, even if they alter blob parameters.
+A new `BLOB_PARAMETER_SCHEDULE` field is added to consensus layer configuration, containing a sequence of entries representing blob parameter changes after `ELECTRA_FORK_EPOCH`. This replaces the deprecated `MAX_BLOB_PER_BLOCK{_FORKNAME}` convention for future forks, while preserving existing entries for previous forks.
+
+Entry rules:
+
+- One entry per fork that changes blob parameters (regular or BPO forks).
+- `EPOCH` and `MAX_BLOBS_PER_BLOCK` required for all entries.
+- `FORK_VERSION` required only for BPO forks (regular forks specify this in source).
 
 ```yaml
 BLOB_PARAMETER_SCHEDULE:
-  # These backported entries are commented out to clarify that existing fork logic does not need to be modified.
-  # However, some clients may prefer to read blob parameters for all forks (including historical ones) from this
-  # unified configuration.
-  # - EPOCH: 269568     ## Deneb (backported)
-  #   MAX_BLOBS_PER_BLOCK: 6
-  # - EPOCH: 364032     ## Electra (backported)
-  #   MAX_BLOBS_PER_BLOCK: 9
   - EPOCH: 400000     ## A future anonymous BPO fork carrying a fork version
     FORK_VERSION: 0x09000000
     MAX_BLOBS_PER_BLOCK: 24
--- a/EIPS/eip-7892.md
+++ b/EIPS/eip-7892.md
@@ -43,17 +43,17 @@ The key motivations for BPO forks are as follows:
 
 ### Definition
 
-BPO hardforks are defined as hardforks that only change certain protocol parameters at a designated point in time without requiring client-side code changes. The new parameters are activated instantly, with no transition function. These hardforks are consensus-breaking, and therefore receive a `ForkVersion` in the consensus layer.
+BPO hardforks are defined as hardforks that only change blob-related protocol parameters at a designated point in time through configuration, without requiring client-side code changes. Activation of new parameters is instant.
 
-### Managed parameters
+### Blob schedule configuration
 
-The following blob-related parameters are now managed by parametric configuration:
+The following blob-related parameters are now managed by the blob schedule configuration:
 
 - **Blob Target (`blob_target`)**: The expected number of blobs per block.
 - **Blob Limit (`blob_limit`)**: The maximum number of blobs per block.
 - **Blob Base Fee Update Fraction (`baseFeeUpdateFraction`)**: Determines how blob gas pricing adjusts per block.
 
-To ensure consistency, when a regular hardfork changes any of these parameters, it MUST do so by adding an entry to the blob schedule.
+To ensure consistency, when a regular hardfork changes any of these parameters, it MUST do so by adding an entry to the blob schedule configuration.
 
 ### Execution layer configuration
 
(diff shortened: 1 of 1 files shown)


EIPS/eip-7892.md
```diff
@@ -39,12 +41,22 @@ The key motivations for BPO forks are:
 
 ## Specification
 
-BPO forks are a special class of hard fork which **only modifies any of the following** blob-related parameters:
+### Parametric hardforks
+
+BPO forks are a category of **parametric hardforks**, which only change certain protocol parameters at a designated point in time without requiring client-side code changes. The new parameters are activated instantly, with no transition function. These hardforks are consensus-breaking, and therefore receive a `ForkVersion` in the consensus layer.
```
tersec, line 46: This ... isn't a true statement. "Parametric hardforks" are not a thing (yet).

raulk: Removed this to avoid confusion.

EIPS/eip-7892.md
```diff
@@ -72,24 +84,75 @@ To facilitate these changes on the execution layer, the `blobSchedule` object sp
...
+  
+  # Add blob schedule entries that define fork versions (therefore representing BPO forks).
+  bpo_forks = [
+      (entry.epoch, entry.fork_version) 
+      for entry in blob_schedule 
+      if entry.fork_version is not None
+  ]
+  forks.extend(bpo_forks)
+  
+  forks.sort(reverse=True)
```
jtraglia, line 140: Let's order `forks` chronologically, so Altair is first and Electra is last.

Currently, this results in: `bpo3`, `bpo2`, `bpo1`, `altair`, `bellatrix`, ..., `electra` which is wrong.

It should be `bpo3`, `bpo2`, `bpo1`, `electra`, `deneb`, ..., `altair` right?

EIPS/eip-7892.md
```diff
@@ -72,24 +84,75 @@ To facilitate these changes on the execution layer, the `blobSchedule` object sp
...
 ```
 
 The parameters and schedules above are purely illustrative. Actual values and schedules are beyond the scope of this specification.
 
-### Requirements
+**Requirements:**
+
+- Execution and consensus clients **MUST** share consistent BPO fork schedules.
+- The timestamp in EL's `blobSchedule` **MUST** align with the start of the epoch specified in the consensus layer configuration.
+- The `max` field in `blobSchedule` **MUST** equal the `MAX_BLOBS_PER_BLOCK` value in the consensus layer configuration.
```
jtraglia, line 115:
```suggestion
- The `max` field in the EL's `blobSchedule` **MUST** equal the `MAX_BLOBS_PER_BLOCK` value in the consensus layer configuration.
```

EIPS/eip-7892.md
```diff
@@ -72,24 +84,75 @@ To facilitate these changes on the execution layer, the `blobSchedule` object sp
...
+    MAX_BLOBS_PER_BLOCK: 72
 ```
 
 The parameters and schedules above are purely illustrative. Actual values and schedules are beyond the scope of this specification.
 
-### Requirements
+**Requirements:**
+
+- Execution and consensus clients **MUST** share consistent BPO fork schedules.
+- The timestamp in EL's `blobSchedule` **MUST** align with the start of the epoch specified in the consensus layer configuration.
```
jtraglia, line 114:
```suggestion
- The timestamp in the EL's `blobSchedule` **MUST** align with the start of the epoch specified in the consensus layer configuration.
```

EIPS/eip-7892.md
jtraglia: Also, this method is not going to work properly in the future when we add more regular forks. If we append a value to `forks` then, for example, it will be: `bpo3`, `bpo2`, `bpo1`, `gloas`, `fulu`, `electra`, `deneb`, ..., `altair` which is wrong. Because `gloas` is probably newer than `bpo3`.

And this reminds me. Let's add `fulu` to the list here.

EIPS/eip-7892.md
```diff
@@ -72,24 +84,75 @@ To facilitate these changes on the execution layer, the `blobSchedule` object sp
...
-```
-BLOB_SCHEDULE:
-  - EPOCH: 348618
+A new `BLOB_PARAMETER_SCHEDULE` field is added to consensus layer configuration, containing a sequence of entries representing blob parameter changes after `ELECTRA_FORK_EPOCH`. This replaces the deprecated `MAX_BLOB_PER_BLOCK{_FORKNAME}` convention for future forks, while preserving existing entries for previous forks.
+
+Entry rules:
+
+- One entry per fork that changes blob parameters (regular or BPO forks).
+- `EPOCH` and `MAX_BLOBS_PER_BLOCK` required for all entries.
+- `FORK_VERSION` required only for BPO forks (regular forks specify this in source).
```
jtraglia, line 95: Personally, I'm not a fan of `FORK_VERSION` being optional. It just feels error prone and a little confusing. I would rather it be listed twice, then deduplicated in `compute_fork_version`.

For example:

```
GLOAS_FORK_EPOCH: 440000
GLOAS_FORK_VERSION: 0x0B000000

BLOB_SCHEDULE:
  ...
  - EPOCH: 440000  ## GLOAS_FORK_EPOCH
    MAX_BLOBS_PER_BLOCK: 72
    FORK_VERSION: 0x0B000000  ## GLOAS_FORK_VERSION

Comment (rolfyone) [reply EIPS/eip-7892.md:95]: i know it breaks old deployments but imo it shouldn't be optional in the spec at the very least

Comment (raulk) [reply EIPS/eip-7892.md:95]: I also prefer verbosity, but expected the duplication to not be welcome. Can certainly add it.

File: EIPS/eip-7892.md [hunk 6]
```diff
@@ -59,37 +71,91 @@ To facilitate these changes on the execution layer, the `blobSchedule` object sp
...
+A new `BLOB_PARAMETER_SCHEDULE` field is added to consensus layer configuration, containing a sequence of entries representing blob parameter changes after `ELECTRA_FORK_EPOCH`.
+
+Entry rules:
+
+- One entry per fork that changes blob parameters (regular or BPO forks).
+- `EPOCH` and `MAX_BLOBS_PER_BLOCK` required for all entries.
+- `FORK_VERSION` required only for BPO forks (regular forks specify this in source).
+
+```yaml
+BLOB_PARAMETER_SCHEDULE:
```
rolfyone, line 98: this name is still different... last consensus-spec entry was `BLOB_SCHEDULE`

raulk: Restored in https://github.com/ethereum/EIPs/pull/9840.

PR #9819, opened 2025-05-26
Add EIP: EVM64 - EOF support

sorpaas:
This splits EIP-7937 (Core document for EVM64) to have EOF support (code validation, `RJUMPI` and `RJUMPV` opcodes) defined separately.

### DIFF
--- a/EIPS/eip-7957.md
+++ b/EIPS/eip-7957.md
@@ -25,7 +25,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 We define the following gas cost constant:
 
-* `G_RJUMPIV64`: 3
+| Name | Gas |
+|------|-----|
+| `G_RJUMPIV64` | `3` |
 
 At EOF contract creation time as defined in [EIP-3670](./eip-3670.md), if the opcode `C0` is encountered and it is not part of PUSH opcode's data, then the interpreter MUST validate that:
 


EIPS/eip-evm64-eof.md
```diff
@@ -0,0 +1,65 @@
...
+* `G_RJUMPIV64`: 3
+
+If EOFv1 mode is entered, then an additional validation step is added. If the opcode `C0` is encountered and it is not part of PUSH opcode's data, then the interpreter MUST validate that:
+
+* The next opcode exists.
+* The next opcode is one of the valid 64-bit mode opcode described above.
+
+For flow operations RJUMPI and RJUMPV, the 64-bit mode has following changes:
+
+* For `RJUMPI64` (0xc0e1), the condition popped from stack is only read for the last 64 bits. Gas cost is `G_RJUMPIV64`.
```
jochem-brouwer, line 36: RJUMPI checks if the condition is zero and then does not jump. If it is nonzero, it will jump to the encoded relative offset. The same goes for RJUMPV except that RJUMPV has one byte to encode how many entries there are in this table (so max 256 entries). The `case` variable from the stack is popped and if this is larger than the amount of cases encoded, it will then not jump, otherwise it will read the relative jump from the table. See also https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4200.md#specification

EIPS/eip-evm64-eof.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+We define the following gas cost constant:
+
+* `G_RJUMPIV64`: 3
+
+If EOFv1 mode is entered, then an additional validation step is added. If the opcode `C0` is encountered and it is not part of PUSH opcode's data, then the interpreter MUST validate that:
```
jochem-brouwer, line 29: So this is validated at container deploy time, and not every time we enter an EOF contract? I assume this should be added to the validation logic of https://eips.ethereum.org/EIPS/eip-3670, but this is not entirely clear here

EIPS/eip-evm64-eof.md
```diff
@@ -0,0 +1,65 @@
...
+
+* The next opcode exists.
+* The next opcode is one of the valid 64-bit mode opcode described above.
```
jochem-brouwer, line 32: There is nothing defined above, except something about RJUMPI and RJUMPV. I'm assuming that EOF cannot have only RJUMPI/RJUMPV in 64-bit mode (but also opcodes like ADD for instance). Also note that RJUMPI/RJUMPV likely do not need special rules (will write that at appropriate line)

EIPS/eip-evm64-eof.md
jochem-brouwer: Also note from spec:

> The immediate argument `relative_offset` is encoded as a 16-bit signed (two's-complement) big-endian value. 

So the relative jump offset also fits in the 64bit mode. So also all (up to 256) RJUMPV table entries will fit in the 64bit mode.

jochem-brouwer (commented):
Hey there! I think for these specific opcodes the 64-bit mode is not necessary. But I might also have missed something here or in my understanding of the RJUMP* opcodes :smile: :+1:

jochem-brouwer:
Never mind! I just realized I'm wrong here, you should ignore the topmost bits. So if the condition has bits set higher than the 64bits then indeed it is >0 in the 256 bit world but not in the 64 bits. Disregard my comments!

EIPS/eip-evm64-eof.md
jochem-brouwer: Nvm, apply the 64bit mask on the possible 256 number and then read the condition. This is indeed different.

EIPS/eip-evm64-eof.md
```diff
@@ -0,0 +1,65 @@
...
+
+For flow operations RJUMPI and RJUMPV, the 64-bit mode has following changes:
+
+* For `RJUMPI64` (0xc0e1), the condition popped from stack is only read for the last 64 bits. Gas cost is `G_RJUMPIV64`.
+* For `RJUMPV64` (0xc0e2), the case popped from stack is only read for the last 64 bits. Gas cost is `G_RJUMPIV64`.
+
```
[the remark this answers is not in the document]
jochem-brouwer: Nvm, its in the original EIP under Optimization Assumptions. This would thus mean we would encode our 256 bit stack items as 4 64-bit items, right? I have not enough low-level knowledge here, I could imagine this is faster, but am a bit concerned because we support both 256 bit and 64 bit operations, so really wondering how this would look like in practice (and if this 4-stack 64-bit representation is as fast or slower than the "normal" 256bit evm)

EIPS/eip-evm64-eof.md
```diff
@@ -0,0 +1,65 @@
...
+For detailed rationale discussion, please see the core EVM64 definition [EIP-7937](./eip-7937.md).
+
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Test Cases
+
+To be added.
```
g11tech: add <--TODO--> markers here and eveerywhere else for bots to flag it on review+ stages

sorpaas: Added.

EIPS/eip-7957.md
```diff
@@ -0,0 +1,70 @@
...
+
+EOF defines a stricter stricter code validation rule to improve efficiency. Due to EVM64 using multibyte opcode (the mode opcode `C0`), a small adaptation is needed. This EIP also additionally defines a 64-bit mode `RJUMPI` and `RJUMPV` to be 64-bit.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+We define the following gas cost constant:
+
+* `G_RJUMPIV64`: 3
```
g11tech: its better to form a table for all constants here including the codes for various rjumps, and also put 3 in `` as well as other constant values so that its more visible

g11tech: @sorpaas let me know if you want to make this change?

sorpaas: Yeah sorry. I missed this!

sorpaas: Done!

PR #9821, opened 2025-05-26
Add EIP: EVM64 - Little endian opcodes

sorpaas:
This defines `BYTE64`, `MLOAD64`, `MSTORE64` and `PUSH*64`.

### DIFF
--- a/EIPS/eip-7958.md
+++ b/EIPS/eip-7958.md
@@ -1,5 +1,5 @@
 ---
-eip: 9821
+eip: 7958
 title: EVM64 - Little endian opcodes
 description: Little endian opcodes `BYTE64`, `MLOAD64`, `MSTORE64`, and `PUSH*64` for EVM64
 author: Wei Tang (@sorpaas)


jochem-brouwer:
Aha, ok, yes this clarifies a lot. I think my confusion also comes from writing stuff in JavaScript which is obviously very far away from the low-level code. I also realized (or: had to look up) that most systems run in little endian and considering that, you are right, then indeed the big endian would first need to be converted.

I think if you could write in your post above in the EIP it would clarify a lot :smile: :+1: On the other hand, this might also be common knowledge and I come from a different "world" where this is not obvious and thus confusing.

EIPS/eip-evm64-le.md
```diff
@@ -0,0 +1,73 @@
...
+
+For detailed rationale discussion, please see the core EVM64 definition [EIP-7937](./eip-7937.md).
+
+## Backwards Compatibility
+
+No backward compatibility issues found.
+
+## Test Cases
+
+To be added.
```
g11tech: add <--TODO--> markets for bot to flag this on review+ stages

sorpaas: Added.

g11tech (approved):

PR #9822, opened 2025-05-26
Update EIP-7937: Fix backward compatibility of JUMPDEST

sorpaas:

--

 - The build passes.

### DIFF
--- a/EIPS/eip-7937.md
+++ b/EIPS/eip-7937.md
@@ -74,7 +74,7 @@ For flow operations JUMP and JUMPI, the behavior is as follows:
 
 * `JUMP` will only read the last 64 bits from the stack value. The rest 192 bits are discarded without reading. Gas cost is `G_MID64`.
 * `JUMPI` will only read the last 64 bits from the stack as destination, and the condition check will only read the last 64 bits. Gas cost is `G_HIGH64`.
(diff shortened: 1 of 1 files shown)


jochem-brouwer:
Ah great, saw this from your post on magicians :smile: :+1: 

Note that JUMPDEST pops/pushes no stack values so it could technically also be a valid `JUMPDEST64`. But, this approach thus does not change these specific cases in current contracts. Before, `C0` would OOG because Invalid opcode, and now `C05B` (5B = jumpdest) also OOGs, but still marks the `5B` as valid jumpdest so does not change that.

Two small other comments regarding the discussions: I would motivate the whole low-endian and the uint64 stuff by linking to this part of `uint256` https://github.com/holiman/uint256/blob/b2bde2d424303d4c61c00d8a2a75e4b671b4f1f2/uint256.go#L16, which shows it is also internally there represented as low-endian 4 64bit units. (I had to find this for myself, have seen `uint256` a lot in Geth, but never looked deep into this. Today I learned uint256 is actually 4 low-endian 64bit :smile: :+1: ). This would place a bit more in context why this is logical/natural especially when you come (like me) from a high level language.

Last thing I spotted, this version of EIP still has `G_RJUMPIV64` defined but not used.

